PR #324, opened 2020-07-03
Dynamic kernel: lstm

ce1adon:
LSTM hidden state update kernels are used by LSTM paths and this change will affect relevant paths.


TejashShah:
@ce1adon The changes look straightforward and good.  Could you please gather perf and compile time # before and after this change?

ce1adon:
@TejashShah Got it.

ce1adon:
@mwootton
Please review

TejashShah:
@ce1adon Code changes all look good. Just want to check the compilation time diff.

Since Auxiliary time is not implemented for non-conv driver command, you could do the following: 

```
diff --git a/src/solver.cpp b/src/solver.cpp
index 1c01267..99ec63a 100644
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -81,7 +81,13 @@ void PrecompileSolutions(const Handle& h, const std::vector<ConvSolution>& sols)
     }
     // Precompile the kernels in parallel, but dont add them to the cache
+    struct timeval start, end;
+    gettimeofday(&start, NULL);
      std::vector<Program> programs = PrecompileKernels(h, kernels);
+    gettimeofday(&end, NULL);
+    if(programs.size() > 0)
+        MIOPEN_LOG_I2((end.tv_sec * 1000000 + end.tv_usec) -
+                      (start.tv_sec * 1000000 + start.tv_usec));
```

ce1adon:
No obvious change at driver cmd end. Please see relative files.

src/ocl/rnn_util_ocl.cpp
```diff
@@ -77,6 +88,8 @@ void LSTMForwardHiddenStateUpdate(const Handle& handle,
         auto kernel = kernels.front();
         kernel(cx,
                reserve_space,
+               static_cast<int>(hy_h),
```
atamazov, line 91: These vars are already of type int, isn't it?

ce1adon: Yes, good catch.

ce1adon:
> LGTM.
> 
> Please update year in License Header of each file you modified.
> 
> e.g. 2019-20 for the following file
> 
> https://github.com/ROCmSoftwarePlatform/MIOpen/blob/22942cdb615f7afde7ca92e9e1fdc80c67b8e2d3/src/ocl/rnn_util_ocl.cpp#L5

Done

src/kernels/MIOpenRNNHiddenStateUpdate.cl
```diff
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright (c) 2019 Advanced Micro Devices, Inc.
+ * Copyright (c) 2020 Advanced Micro Devices, Inc.
```
TejashShah, line 5: The license header includes all years one has modified the file.  So, you should specify 2019-20 as 2019 was the year when it was first authored.

ce1adon: @daniellowell Hey Daniel, what is the license format we are using now?

src/ocl/rnn_util_ocl.cpp
```diff
@@ -185,10 +190,21 @@ void LSTMBackwardHiddenStateUpdate(const Handle& handle,
 
     size_t max_active_threads = handle.GetMaxComputeUnits() * handle.GetWavefrontWidth() * 32;
 
+    size_t total_work = max_batch * hy_h;
+
+    size_t RD_BLCK = (total_work >= 4 * max_active_threads && hy_h % 4 == 0)
+                         ? 4
+                         : ((total_work >= 2 * max_active_threads && hy_h % 2 == 0) ? 2 : 1);
+
+    size_t total_item   = std::max(size_t(total_work / RD_BLCK), size_t(1));
```
TejashShah, line 199: c++-style cast i.e. static_cast<>

https://softwareengineering.stackexchange.com/questions/50442/c-style-casts-or-c-style-casts/50457

```
C-style and function-style casts

And finally, we have the C-style and function-style casts —

(type) object; // c-style cast
type(object); // function-style cast

These two are functionally equivalent and perform the followings in that order until one succeeds —

    const_cast
    static_cast (ignoring access restrictions)
    static_cast , then const_cast
    reinterpret_cast
    reinterpret_cast, then const_cast

It’s better not to use these two because of the fact that they can invoke reinterpret_cast, unless you’re 100% sure static_cast will succeed. Even then, it’s better to explicitly use static_cast.
```

src/kernels/MIOpenRNNHiddenStateUpdate.cl
TejashShah: @daniellowell Please provide your input. In past, the licenses I have seen includes all the years since the file was authored.

daniellowell: Year the file was authored.

src/ocl/rnn_util_ocl.cpp
ce1adon: Please find the post-merging commit here for this comment
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/349/commits/8a39e547b5bd62cb47cc99e2eb71fd90eda5ea74

Issue #299, opened 2020-06-17
Hybrid mode with iGEMM hit detection


daniellowell:
Currently we have PR #178. Which has this "hybrid" behavior:
1. When running Find() MIOpen will first check the Find-Db
2. If there is a hit, then that solver will be selected for compilation and execution
3. If there is a miss, then the "normal" Find() behavior kicks in.

In order to reduce compilation time down further, we propose this tweak in the "hybrid" mode:
1. When running Find() MIOpen will first check the Find-Db
2. If there is a hit, then that solver will be selected for compilation and execution even if the kernel is a iGEMM HIP source kernel
3. If there is a miss, then the "normal" Find() mode kicks in with the HIP source iGEMM kernels bypassed.

This would imply we can distinguish ASM versions of iGEMM from HIP versions and that we have a complement of ASM iGEMM solvers so there is no significant performance decrease. However, since a miss in the Find-Db implies a miss in the performance database, I do not think there will be a performance drop.

atamazov:
> If there is a miss, then the "normal" Find() mode kicks in with the HIP source iGEMM kernels bypassed.

That would bypass also "dynamic" HIP kernels. Do we want this?

atamazov:
Shall we see how much can we get from existing measures, -- Hybrid Find, precompiled kernels, hip-clang (which seems way faster than hcc) -- and then decide about importance of this.

daniellowell:
> That would bypass also "dynamic" HIP kernels. Do we want this?

No, what I meant was that asm iGEMM “dynamic” kernels would run, but HIP source kernels would not be.

atamazov:
...So this would implicitly depend on how "dynamic" kernels are written. I think it is better to explicitly specify which Solutions should be skipped during Hybrid Find fallback (that leverages normal Find). This is clear design (and easy to implement).

atamazov:
[Minor side effect] This (skipping solvers during Hybrid Find fallback) may affect _potential_ Immediate mode performance. I am saying "potential" because this happens when find-db does NOT contain relevant record, so Immediate mode would by default hit its fallback (now GEMM). However Immediate mode may reveal better performance after normal Find (and after Hybrid mode) than after this modified Hybrid. Although the side effect is negligible, it indicates a quirky design.

carlushuang:
This mode seems also imply a **db-record-first** strategy. I think it is OK and easy to understand. If there is some _potential_ performance drop, just add them into DB?

[...]

PR #317, opened 2020-06-29
dynamic igemm wrw

shaojiewang:
Dynamic igemm for wrw path：
1. Dynamic implicit gemm kernels for backward weights;
2. With option to do reduction after gridwise gemm compution;
3. Can be used in case ((n*ho*wo%4==0)&&(c*y*x%32==0)&&(k%32==0));
4. Some cases is more than 10 times faster than exited one.
Test results is recorded here: [performance_data_sheet](https://amdcloud-my.sharepoint.com/:x:/r/personal/shaowang_amd_com/_layouts/15/Doc.aspx?sourcedoc=%7B48DC1965-EF0F-45D4-8125-DB619F2AACE3%7D&file=implicitGemm-wrw-v4r1-06-18-2020-MI60.xlsx&wdOrigin=OFFICECOM-WEB.MAIN.REC&action=default&mobileredirect=true) (refer: sheet "comparison" column "best isa with reduction")

PR #272, opened 2020-06-09
[igemm_dynamic] v4r1 bwd dynamic kernel

carlushuang:
v4r1 backward_data dynamic kernel.
This is to enable dynamic kernel in bwd direction, with v4r1 backward_data stratagy(gridwise_gemm). Although tensor contraction seems a MUST in dynamic kernels(due to its reduce in index calculation), let me re-add tensor-contraction version in future PR.

PR #326, opened 2020-07-06
Fast Hybrid mode of iGEMM

zjing14:
- A implementation of issue #299 
- Add the `skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage` attribut into `ExecutionContext`
- Add `IsOptimizedHybrid()` to set the attribute to `true`. `IsHybrid()` to return true for both plain `Hybrid` and `OptimizedHybrid` modes
- Modify `IsApplicable` of all hip-based implicitGemm solvers. If the attribute is true, skip the solver
- Set `OptimizedHybrid` mode as default


atamazov:
The proposed design has couple drawbacks.
- (1) It does not affect the result of GWSS for the Solvers that implement the `GetWorkspaceSize()` method, so the library might request an excessive workspace from the user's program.
- (2) The design is algorithm-based which does not comply to the [proposal](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/299#issuecomment-645692843)

I have different design in mind.
- (A) Add a new boolean attribute to the `ExecutionContext`: `skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage`. This ugly name tells anything (and BTW denotes [quirky architecture](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/299#issuecomment-645698370)).
  - The new attribiute is similar to the exising `disable_perfdb_access` 
- (B) Set this attribute to `true` when the new "Optimized Hybrid" Find mode is active.
- (C) Each Solver that takes long time to build and has narrow coverage should be modified: its `IsApplicable()` should return `false` when `skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage` is `true. Other solvers stay intact.

An attentive readers would note that “my” design does not _literally_ match my own architectural proposal. However, if they look more closely, they could see that this design is actually a linear transformation of the proposal.

atamazov:
My recent comment edited.

zjing14:
@atamazov Thanks for your feedback. Supporting `GetWorkspaceSize` should be simple to implement. Looks like the main problem is using lists or not. Both solutions are good. The key is as @JehandadKhan said how we have one location with all the information instead of checking `IsApplicable` of each solver. We can discuss it during the meeting and then make a confirmation.

atamazov:
@zjing14 You are welcome, But please don't let words to fool yourself. ___One___ location is Solver. List is ___the second___ location. If we are going to add new lists for each new small feature, then where we'll end up?

zjing14:
@atamazov Just double-check if I understand your proposal and confirm with the implementation details.
1. Add the `skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage` attribut into `SolverBase`
2. Add the "Optimized Hybrid" Find mode that goes through a list of to-disable solvers and set their attribute to `true`
3. Modify the `IsApplicable` to check the attribute.

atamazov:
@zjing14 Good questions!
> 1. Add the `skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage` attribut into `SolverBase`

No, to `ExecutionContext`.

> 2. Add the "Optimized Hybrid" Find mode...

Yes, the support for `OptimizedHybrid` should be added to the `FindMode` class. I recommend updating existing `IsHybrid()` to return `true` for both plain Hybrid and OptimizedHybrid modes and introducing a new function, `IsOptimizedHybrid()`. That would simplify code changes necessary in other places.

> ...that goes through a list of to-disable solvers and set their attribute to `true`

There is no need to create a new list. "OptimizedHybrid" should differ from plain "Hybrid" only in case of find-db miss: it sets `skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage` in the context to `true`. This requires modifications in convolutionocl.cpp and in convolution.cpp. I recommend searching for `IsHybrid` in the sources to find places that require modifications. Examples:

<details><summary><b><i>Modified fragment of FindConvFwdAlgorithm()</i></b></summary><hr>
(Comments removed to save space)

```C++
const miopen::FindMode fm;
bool use_immediate_solution = false;
miopenConvSolution_t sol;
if((fm.IsFast() || fm.IsHybrid()) && !use_winograd_only)
{
    size_t count;
    GetForwardSolutions(handle, wDesc, xDesc, yDesc, 1, &count, &sol);
    use_immediate_solution = (count > 0) && !(fm.IsHybrid() && sol.time < 0);
}

if(use_immediate_solution)
{
    CompileForwardSolution(handle, wDesc, xDesc, yDesc, sol.solution_id);
    const auto id = solver::Id(sol.solution_id);
    perf_db.push_back(
        {id.GetAlgo(conv::Direction::Forward), id.ToString(), sol.time, sol.workspace_size});
}
else
{
    ctx.skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage
        = fm.IsOptimizedHybrid(); /// <<<------------------------------------------ ADDED STATEMENT
    perf_db = UserFindDbRecord::TryLoad(handle, problem, [&](DbRecord& record) {
        DirConvFindCore(handle,
                        xDesc,
                        x,
                        wDesc,
                        w,
                        yDesc,
                        y,
                        workSpace,
                        workSpaceSize,
                        *this,
                        exhaustiveSearch,
                        record,
                        ctx, /// <<<---------------------- PASSING CONTEXT TO SearchForAllSolutions
                        use_winograd_only);
    });
}
```
<hr></details>

<details><summary><b><i>Modified fragment of ForwardGetWorkSpaceSize()</i></b></summary><hr>
(Comments removed to save space)

```C++
auto ctx = ConvolutionContext{xDesc, wDesc, yDesc, *this, conv::Direction::Forward};
...
const miopen::FindMode fm;
while(fm.IsFast() || fm.IsHybrid())
{
    size_t count;
    miopenConvSolution_t sol;
    GetForwardSolutions(handle, wDesc, xDesc, yDesc, 1, &count, &sol);
    if(count < 1 || (fm.IsHybrid() && sol.time < 0))
    {
        ctx.skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage
            = fm.IsOptimizedHybrid(); /// <<<-------------------------------------- ADDED STATEMENT
        break; // Fall down to Normal Find.
    }
    MIOPEN_LOG_I2(sol.workspace_size);
    return sol.workspace_size;
}

ctx.SetupFloats();
ctx.do_search             = false;
ctx.disable_perfdb_access = true;
const size_t direct_workspace = ForwardBackwardDataGetWorkSpaceSizeDirect(ctx);
const size_t implicit_gemm_workspace = ForwardBackwardGetWorkSpaceSizeImplicitGemm(ctx);
const size_t workspace_size_scgemm = ForwardBackwardDataGetWorkSpaceSizeSCGemm(handle, ctx);
...
```
<hr></details>

Notice that some of the required modifications are ___more complex___, because there are still some irregularities in the code (not all Solvers/Invokers are ready yet (((

> 3. Modify the `IsApplicable` of corresponding solvers to check the attribute.

Yes. Each Solver that “knows” about itself that it has narrow coverage and its build takes a lot of time should check the new attribute and tell the caller "I am not applicable".

Hope this helps.

pfultz2:
It does seem a little hackish to use context variable to traverse a different list

atamazov:
@pfultz2 
> It does seem a little hackish to use context variable to traverse a different list

Which list do you mean?

zjing14:
@atamazov Thanks very much for the clarification. Now, I think I understand. Will finish a modification soon.

zjing14:
Hi all, I modified the code following @atamazov comments. Please take a look. If the implementation is OK, I will set the optimized_hybrid to default.

src/convolution.cpp
```diff
@@ -447,7 +447,11 @@ std::size_t ConvolutionDescriptor::ForwardGetWorkSpaceSize(Handle& handle,
         miopenConvSolution_t sol;
         GetForwardSolutions(handle, wDesc, xDesc, yDesc, 1, &count, &sol);
         if(count < 1 || (fm.IsHybrid() && sol.time < 0))
+        {
+            ctx.skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage =
```
JehandadKhan, line 451: Please shorten this variable name and preferably make this an `int` to indicate cost estimate if @atamazov agrees. If not at least rename it to something simpler.

src/find_controls.cpp
```diff
@@ -194,6 +194,7 @@ const char* ToCString(const FindMode::Values mode)
     case FindMode::Values::Normal: return "NORMAL";
     case FindMode::Values::Fast: return "FAST";
     case FindMode::Values::Hybrid: return "HYBRID";
+    case FindMode::Values::OptimizedHybrid: return "OPTIMIZED_HYBRID";
```
JehandadKhan, line 197: Why add a new mode, this looks like a heuristic which should just be part of the "Hybrid" mode. @atamazov comments ?

atamazov: This heuristic has [negative side effects](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/299#issuecomment-645698370) and [looks like a patch](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/299#issuecomment-646140162) from the architectural POV. That is why I would like to keep "plain" Hybrid.

src/convolution.cpp
atamazov: Cost estimate (or a build time?) is a good idea that allows for more flexibility. But we need also indication of _coverage_ to make a decision (skip or use). Also I am not sure that we need this flexibility right now. Let's keep this idea in mind for the future.

atamazov: WRT naming, see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/326#issuecomment-654483612:

> ...`skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage`. This ugly name tells anything (and BTW denotes [quirky architecture](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/299#issuecomment-645698370)).

src/ocl/convolutionocl.cpp
```diff
@@ -789,6 +794,9 @@ void ConvolutionDescriptor::FindConvFwdAlgorithm(Handle& handle,
     }
     else
     {
+        ctx.skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage =
```
atamazov, line 797: Please make similar change in `FindConvBwdDataAlgorithm()`

zjing14:
@atamazov @daniellowell  Do we need to set `optimized_hyrid` mode as default?

doc/src/find_and_immediate.md
```diff
@@ -166,14 +166,15 @@ MIOpen provides a set of Find modes which are used to accelerate the Find calls.
...
+- `HYBRID`, or `3`: Hybrid Find: Checks the [Find-Db](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/finddb.html) for an entry. If there is a Find-Db hit, use that entry. If there is a miss, use the existing Find machinery. Slower start-up times than Fast Find, but no GPU performance drop.
+- `OPTIMIZED_HYBRID`, or `4`, or unset `MIOPEN_FIND_MODE`: Optimized hybrid Find: Checks the [Find-Db](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/finddb.html) for an entry. If there is a Find-Db hit, use that entry. If there is a miss, use the existing Find machinery with skipping slow-compiling kernels. Faster start-up times than Hybrid Find, may have GPU performance drop.
 
- As of MIOpen 2.6, the default mode is set to `HYBRID` mode as default. To run the full `NORMAL` Find mode, set the environment as:
+ As of MIOpen 2.6, the default mode is set to `OPTIMIZED_HYBRID` mode as default. To run the full `NORMAL` Find mode, set the environment as:
```
atamazov, line 172: 2.7 ?

src/ocl/convolutionocl.cpp
atamazov: Please resolve this.

zjing14: Fixed.

doc/src/find_and_immediate.md
zjing14: Fixed.

zjing14:
Here are numbers of renext101, bs=128. @daniellowell @atamazov @JehandadKhan  

  | Wall time (sec) | Throughput (img/sec)
-- | -- | --
Normal | 388 | 126
Hybrid | 336 | 126
Optimized_Hybrid | 175 | 88

atamazov:
> Here are numbers of renext101, bs=128.

Looks like these configs are not included into system-find-db.

JehandadKhan:
> Looks like these configs are not included into the system-find-db.

Possibly, @zjing14 do you have commit [16d87c1](https://github.com/ROCmSoftwarePlatform/MIOpen/commit/16d87c1952e41b54b36fe6cb59a2a40cc3275934) merged in to your branch ? 

Thats the one with the latest tuning.

zjing14:
> > Looks like these configs are not included into the system-find-db.
> 
> Possibly, @zjing14 do you have commit [16d87c1](https://github.com/ROCmSoftwarePlatform/MIOpen/commit/16d87c1952e41b54b36fe6cb59a2a40cc3275934) merged in to your branch ? 
> 
> Thats the one with the latest tuning. 

I merged the last develop.

JehandadKhan:
@zjing14 Your branch's commit history does not include the commit I mentioned above.

zjing14:
> @zjing14 Your branch's commit history does not include the commit I mentioned above.

Yes. But in the test, I merged the latest develop.

doc/src/find_and_immediate.md
```diff
@@ -166,14 +166,15 @@ MIOpen provides a set of Find modes which are used to accelerate the Find calls.
...
-- `HYBRID`, or `3`, or unset `MIOPEN_FIND_MODE`: Hybrid Find: Checks the [Find-Db](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/finddb.html) for an entry. If there is a Find-Db hit, use that entry. If there is a miss, use the existing Find machinery. Slower start-up times than Fast Find, but no GPU performance drop.
+- `HYBRID`, or `3`: Hybrid Find: Checks the [Find-Db](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/finddb.html) for an entry. If there is a Find-Db hit, use that entry. If there is a miss, use the existing Find machinery. Slower start-up times than Fast Find, but no GPU performance drop.
+- `OPTIMIZED_HYBRID`, or `4`, or unset `MIOPEN_FIND_MODE`: Optimized hybrid Find: Checks the [Find-Db](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/finddb.html) for an entry. If there is a Find-Db hit, use that entry. If there is a miss, use the existing Find machinery with skipping slow-compiling kernels. Faster start-up times than Hybrid Find, may have GPU performance drop.
```
pfultz2, line 170: I think this would be better to be called `FAST_HYBRID` rather than `OPTIMIZED_HYBRID`. The word "optimize" seems to imply that there would be not GPU performance drop. I also would phrase it as:

> If there is a miss, use the existing Find machinery with a reduced search space. Faster start-up times than Hybrid Find, but may have GPU performance drop.

doc/src/find_and_immediate.md
```diff
@@ -166,14 +166,15 @@ MIOpen provides a set of Find modes which are used to accelerate the Find calls.
...
 
- As of MIOpen 2.6, the default mode is set to `HYBRID` mode as default. To run the full `NORMAL` Find mode, set the environment as:
+ As of MIOpen 2.7, the default mode is set to `OPTIMIZED_HYBRID` mode as default. To run the full `NORMAL` Find mode, set the environment as:
```
pfultz2, line 172: I dont think it would be a good idea to make this the default mode as this can "corrupt" the find_db with slower solvers. I think we would need a way to annotate entries in find_db with an attribute to know how likely it is to be correct. This way in `HYBRID` mode it can know whether to run a normal find mode or not when an entry is found.

daniellowell: I wasn't planning on making it default any time soon. @zjing14 you might want to modify the documentation and make sure this is not default.

daniellowell: @pfultz2 Perhaps this design make thing too complicated internally. It might be better to have this mode as our initial try for an immediate mode fall back.

zjing14: Set the "hybrid" mode back to default.

doc/src/find_and_immediate.md
zjing14: Changed.

zjing14:
Here are updated numbers of ReNext101, bs=128 with tuning (MIOPEN_FIND_ENFORCE=3). @daniellowell @atamazov @JehandadKhan  

  | Wall time (sec) | Throughput (img/sec)
-- | -- | --
Normal | 379  | 153
Hybrid | 214 | 153
Fast_Hybrid | 217 | 154

daniellowell:
Adding `onHold` until we merge in #272 and #317

zjing14:
Numbers of resnext101, bs128 with different FIND_MODE on Vega20 (MI60).


  | Wall time | Throughput (img/sec)
-- | -- | --
Normal | 7m32.903s | 77.25
Hybrid | 2m37.938s | 77.08
Fast_Hybrid | 2m37.859s | 77.12

TejashShah:
@zjing14 Merge conflict with develop.

zjing14:
Numbers of resnext101, bs128 for different MIOPEN_FIND_MODE on Vega20 (MI60) with empty find-db.

Find mode | Wall time | Throughput (img/sec)
-- | -- | --
Normal | 7m37.145s | 78.9
Fast | 1m54.875s | 35.9
Hybrid | 7m38.975s | 78.7
Fast_Hybrid | 4m21.852s | 78.0

dagamayank:
@pfultz2 @atamazov @daniellowell @JehandadKhan Pl review.

TejashShah:
@zjing14 Just retrigger the build. 

> 
> 
> @pfultz2 @atamazov @daniellowell @JehandadKhan Pl review.

Needs review

atamazov:
@JehandadKhan Thanks for your opinion (although I have the opposite one).

doc/src/find_and_immediate.md
atamazov: I would propose "weird hybrid" ;)

PR #313, opened 2020-06-24
Embed binary databases

JehandadKhan:
This PR embeds the `miopen.db` file as a statically linked library inside MIOpen. This avoids installing files in `<install prefix>/share` and `<install prefix>/miopen/share` directories. 

Since `miopen.db` is a SQLite file, this PR also adds the SQLite memvfs module to MIOpen. This module allows loading a SQLite database from memory, removing the file I/O requirement. The memvfs module for SQLite is built as a separate target and linked against as well. 

The CMake module responsible for linking data files creates a map (`miopen_data`) which holds the start and end pointers to multiple files and can be read from at run time. At runtime MIOpen reads the location of the files from the above mentioned map and passes that pointer to the memvfs module which loads the SQLite database from the supplied pointer.

PR #329, opened 2020-07-07
Disable installation of system DB and convert associated warnings to Info

JehandadKhan:


JehandadKhan:
@daniellowell  Tested on gfx906 for resent configs with SysDB access off as well as disabling access to Sys and User Dbs.

JehandadKhan:
Please note that these changes would be over ridden by #313 and are therefore temporary.

Issue #330, opened 2020-07-07
FP16 Transpose3d Conv throws runtime error


rohithkrn:
Environment: gfx908, rocm3.5

Error log:
```
gridwise_convolution_backward_data_implicit_gemm_v1r1_ncdhw_kczyx_nkdhw.cpp:97:9: error: use of undeclared identifier 'CK_PARAM_TUNABLE_GEMM_A_BLOCK_COPY_DST_DATA_PER_WRITE_GEMM_M'
        CK_PARAM_TUNABLE_GEMM_A_BLOCK_COPY_DST_DATA_PER_WRITE_GEMM_M;
        ^
gridwise_convolution_backward_data_implicit_gemm_v1r1_ncdhw_kczyx_nkdhw.cpp:122:9: error: use of undeclared identifier 'CK_PARAM_TUNABLE_GEMM_B_BLOCK_COPY_DST_DATA_PER_WRITE_GEMM_N'
        CK_PARAM_TUNABLE_GEMM_B_BLOCK_COPY_DST_DATA_PER_WRITE_GEMM_N;
        ^
2 errors generated when compiling for gfx908.
[...]

PR #331, opened 2020-07-08
[igemm]prevent v1r1 bwd 3d fp16 until we use v2 gridwise

asleepzzz:
fix issue #330 
v1r1 bwd 3d FP16/BF16 can't use now until using v2 gridwise to implement it


atamazov:
All tests passed, let's move on?

asleepzzz:
> All tests passed, let's move on?

ok,please help to approve

PR #334, opened 2020-07-08
Dynamic kernel: dropout

ce1adon:
Dropout is used by dropout and rnn/lstm/gru paths and this change will affect relevant paths.

### DIFF
--- a/src/ocl/dropoutocl.cpp
+++ b/src/ocl/dropoutocl.cpp
@@ -271,23 +271,16 @@ void DropoutDescriptor::DropoutForward(const Handle& handle,
     std::string kernel_name  = "DropoutForward";
 
     std::string network_config =
-        "fwd-" + std::string(xDesc.GetType() == miopenHalf ? "fp16-" : "fp32-") + "dim" +
-        std::to_string(in_len[0]) + "x" + std::to_string(in_len[1]) + "x" +
-        std::to_string(in_len[2]) + "x" + std::to_string(in_len[3]) + "x" +
-        std::to_string(in_len[4]) + "-xstr" + std::to_string(in_str[0]) + "x" +
-        std::to_string(in_str[1]) + "x" + std::to_string(in_str[2]) + "x" +
-        std::to_string(in_str[3]) + "x" + std::to_string(in_str[4]) + "-ystr" +
-        std::to_string(out_str[0]) + "x" + std::to_string(out_str[1]) + "x" +
-        std::to_string(out_str[2]) + "x" + std::to_string(out_str[3]) + "x" +
-        std::to_string(out_str[4]) + "-dropout" + std::to_string(dropout) + "-seed" +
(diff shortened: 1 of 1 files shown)


TejashShah:
@ce1adon Does our unit test suite cover dropout layer? Just want to make sure that no functional impact on it.

ce1adon:
@TejashShah 
Yes, we have dedicated dropout test and three rnn related tests for dropout.

TejashShah:
@ce1adon Thanks. Maybe you are probably aware - @zjing14 mentioned that to measure the rough estimation of compiler time difference for before and after changes, you could use -wall time flag in MIOpenDriver.

Wall-clock Time Forward Conv. Elapsed: 2.3396 ms, **Auxiliary API calls: 4436.12 ms** (GWSS: 19.1602)

The auxiliary API calls includes compiler time. 

I will see if we can gather only compilation time stats by making code changes in hip build utils and dump that info when env variable is set.

ce1adon:
@TejashShah 
Sure. Thank you! Keep me posted.

atamazov:
@TejashShah 
> @ce1adon Thanks. Maybe you are probably aware - @zjing14 mentioned that to measure the rough estimation of compiler time difference for before and after changes, you could use -wall time flag in MIOpenDriver.
> 
> The auxiliary API calls includes compiler time.

I am afraid that Aux Wall time is implemented only for convolutions.

ce1adon:
@atamazov 
Thanks for the info! I've requested framework guys to evaluate the impact, which will also give us some clue about the perf change.

daniellowell:
@ce1adon Let me know when it is ok to merge this PR.

ce1adon:
> @ce1adon Let me know when it is ok to merge this PR.

@TejashShah Any thoughts?

ce1adon:
@mwootton

TejashShah:
> 
> 
> > @ce1adon Let me know when it is ok to merge this PR.
> 
> @TejashShah Any thoughts?

Code changes all look good. Just want to check the compilation time diff.

Since Auxiliary time is not implemented for non-conv driver command, you could do the following:

diff --git a/src/solver.cpp b/src/solver.cpp
index 1c01267..99ec63a 100644
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -81,7 +81,13 @@ void PrecompileSolutions(const Handle& h, const std::vector<ConvSolution>& sols)
     }
     // Precompile the kernels in parallel, but dont add them to the cache
+    struct timeval start, end;
+    gettimeofday(&start, NULL);
      std::vector<Program> programs = PrecompileKernels(h, kernels);
+    gettimeofday(&end, NULL);
+    if(programs.size() > 0)
+        MIOPEN_LOG_I2((end.tv_sec * 1000000 + end.tv_usec) -
+                      (start.tv_sec * 1000000 + start.tv_usec));

PR #335, opened 2020-07-09
Cellfft

XiuYuLi:
Cellfft is called tiling-fft in cudnn, but some differents and it`s faster 2x than cudnn`s tiling-fft at most with some cases.

The kernel generator :
    https://github.com/ROCmSoftwarePlatform/micore_generator/tree/master/micoreBingen_asmtext

Another:XiuYu.Li, Xiuyu.Li@amd.com

Updated perf data for several E2E model is here: [CellFFT-MI50-perf-data](https://amdcloud-my.sharepoint.com/:x:/g/personal/xiaoyazh_amd_com/EZMAEddQqY9HshiJJXFjuaMBiaCA4Gr5chdo6g0-Xm8T2Q?e=L8wgcs)


XiuYuLi:
Add 'cellfft' algo for convolution

atamazov:
Please let us know if you need assistance with formatting.

> The files "cellfft_gfx900.hsaco" && "cellfft_gfx906.hsaco" && "cellfft_gfx908.hsaco" in src/kernels will be upload later

Why you would like to use pre-built binary code objects?

We already have FFT algo in the library. Why would you like to add a new algorithm for "cell FFT"? Is it very distinct from conventional FFT?

Notion of an "algorithm" in MIOpen (RFC): https://github.com/AMDComputeLibraries/MLOpen/issues/1753.

carlushuang:
Hi @atamazov , for the question about *Why would you like to add a new algorithm for "cell FFT"* I can provide the answer. basically currently FFT in MIOpen is using one kernel doing the FFT-conv, here xiuyu's `CellFFT` is indeed like `multi-path wino`, that using multiple kernel to do the FFT-conv. this is a technique called `tiling-based` algorithm.

And in cudnn's world, forward for example:

|CUDNN      |wino-conv     |fft-conv     |
| :------------- | :----------: | -----------: |
|no tiling-based |CONVOLUTION_FWD_ALGO_WINOGRAD  |CONVOLUTION_FWD_ALGO_FFT   |
|tiling-based   |CONVOLUTION_FWD_ALGO_WINOGRAD_NONFUSED |CONVOLUTION_FWD_ALGO_FFT_TILING  |


So for miopen equivalent, xiuyu is providing the tiling-based FFT equivalent algorithm

XiuYuLi:
> Please let us know if you need assistance with formatting.
> 
> > The files "cellfft_gfx900.hsaco" && "cellfft_gfx906.hsaco" && "cellfft_gfx908.hsaco" in src/kernels will be upload later
> 
> Why you would like to use pre-built binary code objects?
> 
> We already have FFT algo in the library. Why would you like to add a new algorithm for "cell FFT"? Is it very distinct from conventional FFT?
> 
> Notion of an "algorithm" in MIOpen (RFC): [AMDComputeLibraries/MLOpen#1753](https://github.com/AMDComputeLibraries/MLOpen/issues/1753).

Hi, atamazov
1  I used '/ / clang format off / on' to avoid the problem of clang format check, but it doesn't seem to work completely. Maybe I have to use clang-format
2  Because kernels of 'cellfft' comes from micore_generator(https://github.com/ROCmSoftwarePlatform/micore_generator), there are too many kernels, so it need time to convert to asm .text (the kernel in micore generator is also an assembly, just a C++ style assembly, and I don't think asm .text has more advantages in readability and maintainability)
3 Yes, I thought about using existing FFT algorithm in miopen, but the implementation difference is too big, and the integration will be very troublesome. The difference is that cellfft is faster and faster than miopen's existing FFT

XiuYuLi:
> Hi @atamazov , for the question about _Why would you like to add a new algorithm for "cell FFT"_ I can provide the answer. basically currently FFT in MIOpen is using one kernel doing the FFT-conv, here xiuyu's `CellFFT` is indeed like `multi-path wino`, that using multiple kernel to do the FFT-conv. this is a technique called `tiling-based` algorithm.
> 
> And in cudnn's world, forward for example:
> 
> CUDNN	wino-conv	fft-conv
> no tiling-based	CONVOLUTION_FWD_ALGO_WINOGRAD	CONVOLUTION_FWD_ALGO_FFT
> tiling-based	CONVOLUTION_FWD_ALGO_WINOGRAD_NONFUSED	CONVOLUTION_FWD_ALGO_FFT_TILING
> So for miopen equivalent, xiuyu is providing the tiling-based FFT equivalent algorithm

No, the non-tiling FFT is still multipass, but more two pass:
Non-tiling:
    FFT R2C data transf ->A
    FFT R2C filter transf ->B
    Transpose A->a
    Transpose B->b
    CGEMM c, a, b
    FFTC2R c->output

Tiling:
    FFTR2C_data&perm in LDS ->a
    FFTR2C_filter&perm in LDS ->b
    cgemm a,b ->c
    perm&FFTC2R c->output

atamazov:
> 2 Because kernels of 'cellfft' comes from micore_generator(https://github.com/ROCmSoftwarePlatform/micore_generator), there are too many kernels, so it need time to convert to asm .text (the kernel in micore generator is also an assembly, just a C++ style assembly, and I don't think asm .text has more advantages in readability and maintainability)

I just looked into https://github.com/ROCmSoftwarePlatform/micore_generator/blob/master/micoreBingen/source/fft/sfft16x16/sfft16x16_c2r_g3x3.cpp. It seems like micore implements a kind of its own assembler. Let me omit the argumentation (and possible debates). I highly recommend converting this thing to normal assembly code.

atamazov:
> And in cudnn's world, forward for example...

MIOpen's algorithm and CUDNN's algorithm are substantially different things. It is the MIOpen's Solver/Solution that is the closest counterpart to the CUDNN's algorithm.

In the future (Find 2.0 API) we are going to demote the algorithms to simple attributes of Solutions.

> CellFFT is indeed like multi-path wino...

Multi-Pass Winograd kernels are implementing the Winograd algorithm. MP Winograd's computational properties (for example, accuracy) match the ones of the "single-pass" Winograd. Therefore we did not add any special algorithm for Multi-Pass Winograd.

> that using multiple kernel to do the FFT-conv. this is a technique called tiling-based algorithm.

If computational properties of thie new tile-based FFT (cell FFT) match the ones of the "conventional" FFT, then please re-use existing algorithm.

Some explanations: https://github.com/AMDComputeLibraries/MLOpen/issues/1753#issuecomment-496262114

atamazov:
It seems like we need to refactor FFT to the Solver/Solution/Invoker architecture, in order to open path for the new FFT Solution (this one). @DrizztDoUrden is the owner of this work.

atamazov:
Of course, if computational properties of the tile-based FFT ___substantially differ___ from "conventional" FFT, then we need to introduce the new algorithm.

atamazov:
> I highly recommend converting this thing to normal assembly code.

...Or produce assembly source text (instead of a binary) at least.

FuAsZero:
> It seems like we need to refactor FFT to the Solver/Solution/Invoker architecture, in order to open path for the new FFT Solution (this one). @DrizztDoUrden is the owner of this work.

Okay, after @DrizztDoUrden refactor FFT to the Solver/Solution/Invoker architecture, Xiuyu will put this FFT solution under that.

FuAsZero:
> I highly recommend converting this thing to normal assembly code.
> 
> ...Or produce assembly source text (instead of a binary) at least.

It is doable, Xiuyu can update generator to print asm text as well as current binary.
This will cost some time, can be parallel with FFT refactor.

atamazov:
> Okay, after @DrizztDoUrden refactor FFT to the Solver/Solution/Invoker architecture, Xiuyu will put this FFT solution under that.

Good. Please do not hesitate to ask @DrizztDoUrden for assistance, if needed.

atamazov:
> > I highly recommend converting this thing to normal assembly code.
> > ...Or produce assembly source text (instead of a binary) at least.
> 
> It is doable, Xiuyu can update generator to print asm text as well as current binary.
> This will cost some time, can be parallel with FFT refactor.

Currently the library does not support assembly path for _generated_ kernels, so there is some extra work for me and/or @DrizztDoUrden as well. Thus _conversion_ is preferred from my POV; we have some ASM experts who could help with this.

However I do not know how important/urgent is this new "cell FFT" thing. @daniellowell Could you please shed some light?

carlushuang:
A side note here, the cell-fft kernel in this PR is all dynamic, as prefered in  https://github.com/ROCmSoftwarePlatform/MIOpen/issues/337

atamazov:
> A side note here, the cell-fft kernel in this PR is all dynamic, as prefered in #337

Very good!

FuAsZero:
> > > I highly recommend converting this thing to normal assembly code.
> > > ...Or produce assembly source text (instead of a binary) at least.
> > 
> > 
> > It is doable, Xiuyu can update generator to print asm text as well as current binary.
> > This will cost some time, can be parallel with FFT refactor.
> 
> Currently the library does not support assembly path for _generated_ kernels, so there is some extra work for me and/or @DrizztDoUrden as well. Thus _conversion_ is preferred from my POV; we have some ASM experts who could help with this.

Xiuyu will provide .s file, like dynamic_igemm.

daniellowell:
@XiuYuLi This PR is for fp32 only?

daniellowell:
@XiuYuLi Jenkins failures within your module:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/cellfft_part0/52/pipeline

XiuYuLi:
> @XiuYuLi This PR is for fp32 only?

Yes, now fp32 only

atamazov:
Am I dreaming or we really would like to add two asm source files, ~200 000 lines each?

daniellowell:
I actually don't have a problem with the number of lines of code, since the asm code is autogenerated via an additional tool.
Having said that @XiuYuLi this feature needs to be designed in a such a way to provide a reasonable method of debugging the assembly kernels when a problem arises (as it always does).

@XiuYuLi Also, please provide the link to the assembly code generation repo at the top of this PR.

XiuYuLi:
> Am I dreaming or we really would like to add two asm source files, ~200 000 lines each?

XiuYuLi:
> Am I dreaming or we really would like to add two asm source files, ~200 000 lines each?

We plan to reduce the size of *.s in the next pr (add support for gfx908)

XiuYuLi:
> I actually don't have a problem with the number of lines of code, since the asm code is autogenerated via an additional tool.
> Having said that @XiuYuLi this feature needs to be designed in a such a way to provide a reasonable method of debugging the assembly kernels when a problem arises (as it always does).
> 
> @XiuYuLi Also, please provide the link to the assembly code generation repo at the top of this PR.

Ok, has been uploaded：https://github.com/ROCmSoftwarePlatform/micore_generator/tree/master/micoreBingen_asmtext

atamazov:
> Okay, after @DrizztDoUrden refactor FFT to the Solver/Solution/Invoker architecture, Xiuyu will put this FFT solution under that.

Good. Let's have separate algorithm for now, for development purposes, and wait for have FFT solver.

src/cellfft/source/cellfft_get_kernel.cpp
```diff
@@ -0,0 +1,488 @@
+/*******************************************************************************
```
atamazov, line 1: Is there a reason to put this code to a separate module? If this is used only in one module (in cellfft solver, I guess), please put it into that module directly.

src/CMakeLists.txt
```diff
@@ -462,6 +464,18 @@ if( MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "HIP")
...
+list(APPEND MIOpen_Source
+    cellfft/include/cellfft_get_kernel.hpp
+    cellfft/include/cellfft_op.hpp
+    cellfft/include/cellfft_param.hpp
+    cellfft/source/cellfft_get_kernel.cpp
+    cellfft/source/cellfft_op.cpp
+    cellfft/source/cellfft_param.cpp
+    include/miopen/conv/invokers/cellfft.hpp
+    conv/invokers/cellfft.cpp
+    solver/conv_cellfft.cpp
```
atamazov, line 476: Invokers are used in one module only, i.e. in the cellfft solver. There is no need to have separate module for invoker, it can be included directly into the solver. Likewise, all other modules and respective interfaces (.cpp and .hpp files). AFAICS all host code can be moved into solver (and made that code non-global). This would open path for strong compiler optimizations, for example and cleans global namespace.

src/convolution.cpp
```diff
@@ -464,6 +467,8 @@ std::size_t ConvolutionDescriptor::ForwardGetWorkSpaceSize(Handle& handle,
 
     const size_t implicit_gemm_workspace = ForwardBackwardGetWorkSpaceSizeImplicitGemm(ctx);
 
+    const size_t cellfft_workspace = cellfft::get_auxbuf_size(ctx);
```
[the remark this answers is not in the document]
atamazov: Of course, this is Ok for now, ___if___ we plan to move cellfft under FFT algorithm. When we do that, this line will be reverted automatically, altogether with all changes in other common files.

src/cellfft/source/cellfft_get_kernel.cpp
```diff
@@ -0,0 +1,488 @@
...
+#define START_R2C_X   (START_C2R_S+22)
+#define START_R2C_Xg  (START_R2C_X+6)
+#define START_C2R_X   (START_R2C_Xg+8)
+#define START_R2C     (START_C2R_X+4)
+#define START_R2C_PAD (START_R2C+96)
+#define START_C2R     (START_R2C_PAD+48)
+
+static const char* g_knames[]=
+{
+    "sfft4x4_r2c_perm_s3x3",
```
atamazov, line 44: AFAICS we have hundreds of kernels in one .s file. After assembly, we have a large code object, but only a few kernels will be used. I'm afraid we are wasting time for loading the excessively big module. Is it possible to split the asm source file? Have you considered how this would affect performance?

driver/conv_driver.hpp
```diff
@@ -2944,7 +2953,7 @@ int ConvDriver<Tgpu, Tref>::VerifyForward()
     auto tolerance = GetDefaultTolerance();
     // iGemm's deviation is higher than other algorithms.
     // The reason is most likely different order of computations.
-    if(is_fwd_igemm)
+    if(is_fwd_igemm || is_fwd_cellfft)
```
atamazov, line 2956: Is it really so that cellfft needs bigger tolerance?

atamazov (changes requested):

atamazov:
@DrizztDoUrden What is ETA for FFT solver/invoker?

DrizztDoUrden:
Not too soon, sadly. I hope to finish Winograd asap, and then I'll have my leave (I think it's 24.08 - 09.09). After that, I'll soon finish run and measure elimination. At that point, I would be able to resurrect the FFT branch and create PR. So I'd hope to make it one and a half months from now approximately.

src/cellfft/include/cellfft_get_kernel.hpp
```diff
@@ -0,0 +1,48 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2019 Advanced Micro Devices, Inc.
```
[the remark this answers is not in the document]
XiuYuLi: Ok!

src/cellfft/source/cellfft_get_kernel.cpp
XiuYuLi: I`ll put it into conv_cellfft.cpp directly

driver/conv_driver.hpp
XiuYuLi: According to our test of hundreds of cases, FWD and BWD can achieve the same accuracy as GEMM in most cases (1e-6), but in some cases it is 1e-5; wrw accounts for about half of each. Therefore, for convenience, the tolerance is set to 1e-5

src/cellfft/source/cellfft_get_kernel.cpp
XiuYuLi: Cellfft is the most matching combination from a large number of kernels. It is difficult to group and split separately, which will make the problem very complicated. The size binary object (if include gfx908) is about 1.6M,so I think it will only slightly affect the initialization time

driver/conv_driver.hpp
atamazov: > ...the same accuracy as GEMM...

Hmm... Do you mean GEMM or Implicit GEMM (iGEMM)?

> ...but in some cases it is 1e-5...

Have you carefully checked that cellfft works properly in these cases, especially for Forward and Backward directions?

src/cellfft/source/cellfft_get_kernel.cpp
atamazov: > It is difficult to group and split separately, which will make the problem very complicated.

Hmm... I do not see any need to _group_ kernels. One source file per kernel should be fine. If you have some kernel generation tool, that shouldn't be a problem, right?

> The size binary object (if include gfx908) is about 1.6M,so I think it will only slightly affect the initialization time

I tend to agree with you. However I would prefer to really check this. I do not know if ROCm is good with big files containing many small kernels.

driver/conv_driver.hpp
XiuYuLi: > > ...the same accuracy as GEMM...
> 
> Hmm... Do you mean GEMM or Implicit GEMM (iGEMM)?
> 
> > ...but in some cases it is 1e-5...
> 
> Have you carefully checked that cellfft works properly in these cases, especially for Forward and Backward directions?

Yes, we checked, even in these cases, the accuracy of most elements is still <1e-6(the driver result display *e-8), the accuracy > 1e-5(driver display result *e-7) occurs when the tile of 32x32 is selected, because the larger the tile, the lower the accuracy (but the accuracy of 32x32 also meets the requirement of < 1e-6 in many cases) and i test with direct method of cpu in one seperate program also, this my compare accuracy code:

bool check( const float* a, const float* b, int n )
{
double s0=0;
double s1=0;
double ref=0;
for( int i=0; i<n; ++i ){
double ai=(double)a[i];
double bi=(double)b[i];
double d=ai-bi;
double dd=dd;
double aa=2.0ai*ai;
double x=dd/aa;
if(x>ref){ ref=x; }
s0+=dd;
s1+=aa;
}
return (sqrt(s0/s1)<1e-5);
}

XiuYuLi: > > ...the same accuracy as GEMM...
> 
> Hmm... Do you mean GEMM or Implicit GEMM (iGEMM)?
> 
> > ...but in some cases it is 1e-5...
> 
> Have you carefully checked that cellfft works properly in these cases, especially for Forward and Backward directions?

Yes, Direct method or gemm method

Kirpich30000: @XiuYuLi what error metric is computed by ```check()``` ? It is not a RMSE. Anyway the value returned by ```check()``` is highly influenced by the input data distribution.

atamazov: [off-topic] @XiuYuLi Please do not use editing others' comments for your replies. Basically nobody receive notifications about edits, so we didn't read your response and I saw it by accident.

atamazov: @Kirpich30000 FYI answer by @XiuYuLi:
> I'm not sure, but the 'check()' function uses a commonly used method for checking the accuracy and correctness of the results. In fact, we not only rely on the return value of the 'check()', also output it to a file for viewing. Of course, it doesn't matter. Whether it's in the miopen driver or in my own test program, it's all passed.As for the accuracy of the result, it is what it should be

atamazov: Reply to the [following comment](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r474983224) from @XiuYuLi :
> I think the driver display '*e-x' and the tolerance is not same means. I can't confirm now because it's just for testing convenience, but the value of tolerance in test.cpp whether be changed or not , it has no effect on the calculation results

atamazov: @XiuYuLi Let me ask again the following question (from [this comment](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r474983224)):
- ___Does driver fail verification without these adjustments? On which configs?___

daniellowell:
@atamazov Your review is stale.

daniellowell:
@XiuYuLi can you run through a full model to see what end-to-end performance improvements we are seeing with this PR?
Reach out to me via email if you need details on how to do this.

carlushuang (dismissed):
LGTM

XiuYuLi:
> @XiuYuLi can you run through a full model to see what end-to-end performance improvements we are seeing with this PR?
> Reach out to me via email if you need details on how to do this.

Ok

src/conv/invokers/cellfft.cpp
```diff
@@ -0,0 +1,256 @@
...
+#include <miopen/handle.hpp>
+#include <miopen/kernel.hpp>
+#include <miopen/tensor.hpp>
+#include <boost/any.hpp>
+
+// clang-format off
+namespace miopen {
+static void lk_cgemm( const Handle& handle, const Kernel& kern, const solver::cellfft_param_t& p, void* c, void* a, void* b, float alpha )
+{
+    float coef=alpha*(p.id==0?0.00390625f:0.0009765625f);
```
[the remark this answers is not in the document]
atamazov: @carlushuang This is ternary op that can return one of two magic numbers: 0.00390625 and 0.0009765625.

@XiuYuLi I am second to the question -- where these numbers came from? We need to know to maintain this code.

src/CMakeLists.txt
atamazov: This is not resolved.

src/cellfft/source/cellfft_get_kernel.cpp
atamazov: > However I would prefer to really check this. I do not know if ROCm is good with big files containing many small kernels.

Any results of the check?

driver/conv_driver.hpp
atamazov: This is not resolved.

src/conv/invokers/cellfft.cpp
```diff
@@ -0,0 +1,256 @@
...
+#include <miopen/conv/invokers/cellfft.hpp>
+#include <miopen/conv/data_invoke_params.hpp>
+#include <miopen/conv/wrw_invoke_params.hpp>
+#include <miopen/errors.hpp>
+#include <miopen/handle.hpp>
+#include <miopen/kernel.hpp>
+#include <miopen/tensor.hpp>
+#include <boost/any.hpp>
+
+// clang-format off
```
atamazov, line 36: This should be removed everywhere, except cases when clang-format spoils readability.

atamazov (changes requested):

src/conv/invokers/cellfft.cpp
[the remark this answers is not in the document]
XiuYuLi: > @carlushuang This is ternary op that can return one of two magic numbers: 0.00390625 and 0.0009765625.
> 
> @XiuYuLi I am second to the question -- where these numbers came from? We need to know to maintain this code.

@atamazov This number is 1/(cell_size*cell_size), the cell_size is 16 or 32, the reason why i used constant number instead '1/(cell_size*cell_size)' is :
    Cellfft is to transplant part of the code of deepcore(https://github.com/XiuYuLi/deepcore_source_code) to GCN, and deeply optimizes the characteristics of GCN. There are 5 cell sizes(in cellfft, the tile size=cell_size-filter_size+1) in deepcore:8x8, 16x16, 32x32, 64x64 and 128x128; 64x64 and 128x128 must be compatible with flatcgemm, but flatcgemm cannot be accelerated with xdlops, so it is not necessary to support it. With wino (8x8 tile) of F (6,3), cellfft of 8x8 is even more unnecessary. So only two cell sizes and use constant is simpled.
    Thanks!

src/conv/invokers/cellfft.cpp
XiuYuLi: This can be unchanged now? we can solve it together in the next PR about 908, because CI consumes too much time

src/conv/invokers/cellfft.cpp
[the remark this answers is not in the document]
atamazov: Thanks. Please put this explanation into C++ comment into the source.

src/conv/invokers/cellfft.cpp
atamazov: I would prefer to not merge incorrectly formatted code.

atamazov:
The following (yet superficial) comments are not resolved yet:
- [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#issuecomment-676513261
  - Blocked by FFT Solver work, see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#issuecomment-657640940
- [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r473148114
- [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r474320719
- [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r473163884
- [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r488153134
- [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r487145974

Also I need to make deeper review of this PR and some testing to evaluate performance impact.

@daniellowell  Are we in hurry?

src/conv/invokers/cellfft.cpp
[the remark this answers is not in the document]
XiuYuLi: Added comments

src/conv/invokers/cellfft.cpp
XiuYuLi: The changes have been completed: Removed all / / clang format *, only in conv_ cellfft.cpp A small part of the code in uses once

XiuYuLi:
> The following (yet superficial) comments are not resolved yet:
> 
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#issuecomment-676513261)
>   
>   * Blocked by FFT Solver work, see [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#issuecomment-657640940)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r473148114)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r474320719)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r473163884)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r488153134)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r487145974)
> 
> Also I need to make deeper review of this PR and some testing to evaluate performance impact.
> 
> @daniellowell Are we in hurry?

@atamazov @daniellowell Can we start as an independent algo now so that we can add supporting for gfx908 as soon as possible？

src/solver/conv_cellfft.cpp
```diff
@@ -47,517 +46,350 @@
...
 {
-    "sfft4x4_r2c_perm_s3x3",
-    "sfft4x4_r2c_perm_s5x5",
-    "sfft4x4_r2c_qerm_s3x3",
-    "sfft4x4_r2c_qerm_s5x5",
-    "sfft5x5_r2c_perm_s3x3",
-    "sfft5x5_r2c_perm_s5x5",
-    "sfft5x5_r2c_qerm_s3x3",
-    "sfft5x5_r2c_qerm_s5x5",
+    "sfft4x4_r2c_perm_s3x3", "sfft4x4_r2c_perm_s5x5",
```
atamazov, line 49: I think you want to insert `//clang-format off` here and revert to linear list.

XiuYuLi: > I think you want to insert `//clang-format off` here and revert to linear list.

Have changed.

src/include/miopen/conv/invokers/cellfft.hpp
```diff
@@ -0,0 +1,40 @@
...
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+* SOFTWARE.
+*
+*******************************************************************************/
+// clang-format off
+#ifndef GUARD_MIOPEN_CELLFFT_HPP
+#define GUARD_MIOPEN_CELLFFT_HPP
+
+#include <miopen/invoker.hpp>
+#include "../../../../solver/cellfft/include/cellfft_param.hpp"
```
DrizztDoUrden, line 31: Can you move cellfft_param.hpp to the include directory so you won't have to use `../` stuff?

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,603 @@
...
+        kid = START_C2R_S + 11 * p.id + (cc0 ? 8 : (p.cnx > p.cny ? 0 : 4)) + (nmax >> 1) - 1;
+    }
+    size_t gdx = (p.m + (1 << shx) - 1) >> shx;
+    const std::vector<size_t> block{256, 1, 1};
+    const std::vector<size_t> grid{gdx << 8, p.n, 1};
+    std::ostringstream options;
+    GenerateClangDefsym(options, "ROCM_METADATA_VERSION", ctx.rmv.UseV3() ? 5 : 4);
+    return KernelInfo{options.str(), block, grid, file_name, g_knames[kid]};
+}
+static void get_solution(const ConvolutionContext& ctx, ConvSolution& sol, const cellfft_param_t& p)
```
DrizztDoUrden, line 514: I'd suggest renaming it to something like `fill_kernel_build_parameters`. Or `FillKernelBuildParameters`. Most of the library functions are camel-cased, but I don't think it is enforced.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,603 @@
...
+#include <miopen/gcn_asm_utils.hpp>
+#include <miopen/solver.hpp>
+#include <miopen/env.hpp>
+#include <miopen/kernel_build_params.hpp>
+#include <miopen/conv/tensors.hpp>
+#include <boost/any.hpp>
+#include <miopen/conv/invokers/cellfft.hpp>
+
+// clang-format off
+#define START_R2C_S   0
```
DrizztDoUrden, line 38:
```suggestion
constexpr size_t start_r2c_s = 0;
```

That would provide a nicely typed value. The same applies to the rest of the defines.

DrizztDoUrden (changes requested):

src/include/miopen/conv/invokers/cellfft.hpp
XiuYuLi: I think i can move it into solver/ directory because this file only used by cellfft?

src/solver/conv_cellfft.cpp
XiuYuLi: Ok

src/include/miopen/conv/invokers/cellfft.hpp
DrizztDoUrden: You can make it `#include <cellfft/param.hpp>` if you wish to distinguish it from other headers even more. And it has cellfft in its name anyway :)

XiuYuLi: Ok..., I'll wait for other people's review results to save CI resources

atamazov: The cellft_param module (.hpp/.cpp) is used only in cellfft.cpp. What is the purpose of detaching of cellfft_param?

XiuYuLi: @atamazov :
The reason for the separation is that when the optimal solver is selected in the initialization of a model, the host parameters of each layer can be cached without repeated calculation. Although it is uncertain whether miopen will support it, this possibility is not ruled out

DrizztDoUrden: If I understand you correctly, they effectively get cached inside of an invoker.

atamazov: > The reason for the separation is that when the optimal solver is selected in the initialization of a model, the host parameters of each layer can be cached without repeated calculation. Although it is uncertain whether miopen will support it, this possibility is not ruled out

Right now entities declared in celfft_param.hpp are used only in one module, and there is no need to detach it. Let's put it all together. This wouldn't cut off future possibilities. When/if it is needed somewhere else, we can easily refactor, -- by moving the necessary functionality into separate module.

atamazov: Usage of `../../..` stuff in #include directives is not allowed anyway.

XiuYuLi: > > The reason for the separation is that when the optimal solver is selected in the initialization of a model, the host parameters of each layer can be cached without repeated calculation. Although it is uncertain whether miopen will support it, this possibility is not ruled out
> 
> Right now entities declared in celfft_param.hpp are used only in one module, and there is no need to detach it. Let's put it all together. This wouldn't cut off future possibilities. When/if it is needed somewhere else, we can easily refactor, -- by moving the necessary functionality into separate module.

Ok, i`ll changed and do same for flexgemm

asroy:
@XiuYuLi Is the kernel generator still here?

https://github.com/ROCmSoftwarePlatform/micore_generator/tree/master/micoreBingen_asmtext

XiuYuLi:
> @XiuYuLi Is the kernel generator still here?
> 
> https://github.com/ROCmSoftwarePlatform/micore_generator/tree/master/micoreBingen_asmtext

Yes, still here

src/include/miopen/conv/invokers/cellfft.hpp
atamazov: Good!

FuAsZero:
> The following (yet superficial) comments are not resolved yet:
> 
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#issuecomment-676513261)
>   
>   * Blocked by FFT Solver work, see [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#issuecomment-657640940)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r473148114)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r474320719)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r473163884)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r488153134)
> * [ ]  [#335 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/335#discussion_r487145974)
> 
> Also I need to make deeper review of this PR and some testing to evaluate performance impact.
> 
> @daniellowell Are we in hurry?

Artem, thanks for review suggestion. 
How about your progress of performance impact testing?

daniellowell:
@XiuYuLi @atamazov @FuAsZero I believe we will need to add this solver to the approved "dynamic kernel" list. Non-tunable and ASM fits.

This can be a follow up PR.

daniellowell:
>@daniellowell Are we in hurry?

We need this in this year. It will be a nice fill in for the dynamic kernel list. In that case we should have it in by ROCm 4.0 at the latest.

atamazov:
> 
> 
> > It seems like we need to refactor FFT to the Solver/Solution/Invoker architecture, in order to open path for the new FFT Solution (this one). @DrizztDoUrden is the owner of this work.
> 
> Okay, after @DrizztDoUrden refactor FFT to the Solver/Solution/Invoker architecture, Xiuyu will put this FFT solution under that.

Now we have FFT Solver! Please go ahead and integrate cellfft into fresh develop. 

If you see too many merge conflicts, then perhaps it will be easier to revert all except the new Solver and then make some small changes to integrate it to FFT algorithm.

atamazov:
Converted to draft. Please do the following:
- merge recent develop into this. 
- integrate cellfft into existing FFT algo.

Until that, please do not run this on Jenkins to save resources. Thanks.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+    const Handle& h, const Kernel& kern, const cellfft_param_t& p, void* dst, const void* src)
+{
+    uint32_t ldr = p.bnx * p.bny * (p.dir == 0 ? p.k : p.n);
+    h.Run(kern)(dst, p.ldb, p.bbks, src, p.n, ldr);
+}
+static void lk_fft2d_r2c_grid_perm(
+    const Handle& h, const Kernel& kern, const cellfft_param_t& p, void* dst, const void* src)
+{
+    uint32_t grid = (p.grid_y << 16) | p.grid_x;
+    uint32_t tile = (p.tile_y << 16) | p.tile_x;
```
atamazov, line 659: Are we quite sure that grid and tile wouldn't exceed 65535? The same wrt other `<< 16` computations.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+    uint32_t pv  = ctx.pad_h;
+    p.anx        = ctx.in_width;
+    p.any        = ctx.in_height;
+    p.bnx        = ctx.kernel_size_w;
+    p.bny        = ctx.kernel_size_h;
+    p.cnx        = ctx.out_width;
+    p.cny        = ctx.out_height;
+    if((p.dir = ctx.direction.IsForward() ? 0 : 1) != 0)
+    {
+        pu = p.bnx - pu - 1;
```
atamazov, line 513: Use ctx.GetBackwardPadH/W

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+    uint32_t bny;
+    uint32_t cnx;
+    uint32_t cny;
+    uint32_t pad_l;
+    uint32_t pad_r;
+    uint32_t pad_t;
+    uint32_t pad_b;
+    uint32_t nbanks;
+    uint32_t id;
+    uint32_t dir;
```
atamazov, line 411: Why not enum for `dir`?

atamazov (changes requested):
The solver is overloaded with magic numbers (each one is an implicit dependence). The code will be difficult to maintain. Please try to get rid of them as much as possible.

jane-zxy:
> The solver is overloaded with magic numbers (each one is an implicit dependence). The code will be difficult to maintain. Please try to get rid of them as much as possible.

@atamazov 
Hi Atam,
thanks for all your review, I'll update the pr later, and will re-request your review after I do corresponding changes.

src/convolution.cpp
```diff
@@ -467,6 +467,7 @@ std::size_t ConvolutionDescriptor::ForwardGetWorkSpaceSize(Handle& handle,
     const size_t workspace_size_winograd = ForwardBackwardDataGetWorkSpaceSizeWinograd(ctx);
     const size_t direct_workspace        = ForwardBackwardDataGetWorkSpaceSizeDirect(ctx);
     const size_t implicit_gemm_workspace = ForwardBackwardGetWorkSpaceSizeImplicitGemm(ctx);
+    const size_t workspace_size_fft      = ForwardBackwardDataGetWorkSpaceSizeFFT(ctx);
```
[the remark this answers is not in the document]
jane-zxy: Hi Atam,
i think this change is necessary, since inside of following "if MIOPEN_USE_GEMM", there are a max func, pls review.
        if(miopen::any_of(GetConvDilations(), [](auto v) { return v > 1; }))
        {
            return std::max({workspace_size_fft,
                             workspace_size_gemm,
                             direct_workspace,
                             implicit_gemm_workspace,
                             workspace_size_winograd});
        }

src/convolution.cpp
```diff
@@ -937,6 +936,7 @@ ConvolutionDescriptor::BackwardWeightsGetWorkSpaceSize(Handle& handle,
     const size_t workspace_size = std::max({BackwardWeightsGetWorkSpaceSizeImplicitGemm(ctx),
                                             BackwardWeightsGetWorkSpaceSizeWinograd(ctx),
                                             BackwardWeightsGetWorkSpaceSizeDirect(ctx),
+                                            ForwardBackwardDataGetWorkSpaceSizeFFT(ctx),
```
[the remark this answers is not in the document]
jane-zxy: yes, you are right, I've add a new func to do this: BackwardWeightsGetWorkSpaceSizeFFT(), pls review the latest code.

src/idiv.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+// Find the position of the least significant bit set to 1 in a 32 bit integer
+static inline uint32_t bfls(uint32_t n)
+{
+    n = n | (n >> 0x01);
+    n = n | (n >> 0x02);
+    n = n | (n >> 0x04);
+    n = n | (n >> 0x08);
+    n = n | (n >> 0x10);
+    return __builtin_popcount(n);
```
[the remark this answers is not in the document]
jane-zxy: previous code is implemented by Xiuyu, I use a simple and more readable implementation in latest code, pls review.

src/include/miopen/idiv.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <cstdint>
+
+namespace miopen {
+struct magic_t
+{
+    uint32_t m; // magic number
+    uint32_t s; // shift
+};
+// get magic&shift of faster integer division
```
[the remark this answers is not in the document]
jane-zxy: basically,
This is to faster the integer division for asm coding, which will use much less multipler instructions while we need divide a num.
I'm thinking of moving this only to solver conv_cellfft.cpp, since only asm level coding and cellfft use it now.

src/include/miopen/idiv.hpp
```diff
@@ -0,0 +1,40 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#ifndef GUARD_MIOPEN_IDIV_HPP
+#define GUARD_MIOPEN_IDIV_HPP
+
+#include <cstdint>
+
+namespace miopen {
```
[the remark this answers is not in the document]
jane-zxy: as I previouslly said, I'm thinking of moving this only to solver conv_cellfft.cpp, since only asm level coding and cellfft use it now.

src/idiv.cpp
```diff
@@ -0,0 +1,63 @@
...
+magic_t idiv_magic(uint32_t nmax, uint32_t d)
+{
+    magic_t magic = {1, 0};
+    if(d == 1)
+        return magic;
+    uint64_t nc    = ((nmax + 1) / d) * d - 1;
+    uint32_t nbits = bfls(nmax);
+    uint32_t r     = (nbits << 1) + 1;
+    magic.m        = -1;
+    magic.s        = -1;
```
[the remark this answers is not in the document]
jane-zxy: fixed as you said.

src/idiv.cpp
```diff
@@ -0,0 +1,63 @@
...
+    return __builtin_popcount(n);
+}
+
+namespace miopen {
+magic_t idiv_magic(uint32_t nmax, uint32_t d)
+{
+    magic_t magic = {1, 0};
+    if(d == 1)
+        return magic;
+    uint64_t nc    = ((nmax + 1) / d) * d - 1;
```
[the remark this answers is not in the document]
jane-zxy: added it

src/include/miopen/solver.hpp
```diff
@@ -1895,6 +1895,13 @@ struct gemm : SolverBase<ConvolutionContext>
     }
 };
 
+struct ConvCellfft : SolverBase<ConvolutionContext>
```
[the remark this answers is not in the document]
jane-zxy: yes, you are right, I think so. I've added it.

src/ocl/convolutionocl.cpp
```diff
@@ -3059,6 +3059,14 @@ void ConvolutionDescriptor::FindConvBwdWeightsAlgorithm(Handle& handle,
                     AlgorithmName{"miopenConvolutionBwdWeightsAlgoImplicitGEMM"};
                 EvaluateInvokers(handle, all, algorithm_name, network_config, invoke_ctx, record);
             }
+
+            // fft
+            if(!miopen::IsDisabled(MIOPEN_DEBUG_CONV_FFT{}))
```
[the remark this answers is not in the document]
jane-zxy: added.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+        {
+            sol.construction_params.push_back(get_kinfo_r2c_a(ctx, p, fname));
+            sol.construction_params.push_back(get_kinfo_r2c_b(ctx, p, fname));
+        }
+        sol.construction_params.push_back(get_kinfo_c2r_g(ctx, p, fname));
+    }
+}
+bool ConvCellfft::IsApplicable(const ConvolutionContext& ctx) const
+{
+    if(MIOPEN_BACKEND_OPENCL)
```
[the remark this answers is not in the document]
jane-zxy: changed as you said

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+        return false;
+    const auto name = ctx.GetStream().GetDeviceName();
+    if(name != "gfx900" && name != "gfx906")
```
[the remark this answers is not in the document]
jane-zxy: No, by now, cellfft only support gfx900 and gfx906, support for gfx908 is still under developing

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+        return false;
+    if((ctx.kernel_stride_w | ctx.kernel_stride_h | ctx.kernel_dilation_w | ctx.kernel_dilation_h |
+        ctx.group_counts) != 1)
```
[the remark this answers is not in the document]
jane-zxy: I change the implementation into simpler one, pls review.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+        return false;
+    if(!ctx.direction.IsForward())
+    {
+        int pu = ctx.kernel_size_w - ctx.pad_w - 1;
+        int pv = ctx.kernel_size_h - ctx.pad_h - 1;
+        if((pu < 0) || (pv < 0))
+            return false;
+    }
```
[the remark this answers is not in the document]
jane-zxy: changed as comments.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+{
+    if(MIOPEN_BACKEND_OPENCL)
+        return false;
+    const auto name = ctx.GetStream().GetDeviceName();
+    if(name != "gfx900" && name != "gfx906")
+        return false;
+    if((ctx.kernel_stride_w | ctx.kernel_stride_h | ctx.kernel_dilation_w | ctx.kernel_dilation_h |
+        ctx.group_counts) != 1)
+        return false;
+    if(!ctx.direction.IsForward())
```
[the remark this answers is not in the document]
jane-zxy: yes, you are right, I'll change it into IsBackwardData(). however previous code is functionally ok since this func is only called for fwd/bwd cases.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+    cellfft_param_t params{};
+    if(!ctx.direction.IsBackwardWrW())
+    {
+        build_cellfft_params(params, ctx);
+    }
+    else
+    {
+        build_cellfft_params_grad(params, ctx);
+    }
+    sol.workspce_sz = get_auxbuf_size(params);
```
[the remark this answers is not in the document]
jane-zxy: changed as review.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+    uint32_t any = ctx.in_height;
+    uint32_t fnx = ctx.kernel_size_w;
+    uint32_t fny = ctx.kernel_size_h;
+    uint32_t pu  = ctx.pad_w;
+    uint32_t pv  = ctx.pad_h;
+    if(!ctx.direction.IsForward())
+    {
+        pu = fnx - pu - 1;
+        pv = fny - pv - 1;
+    }
```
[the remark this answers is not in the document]
jane-zxy: changed as review.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+    const size_t gdx   = (p.m + 15) >> 4;
+    const std::vector<size_t> block{bdx, 1, 1};
+    const std::vector<size_t> grid{gdx * bdx, p.k, 1};
+    std::ostringstream options;
+    GenerateClangDefsym(options, "ROCM_METADATA_VERSION", ctx.rmv.UseV3() ? 5 : 4);
+    return KernelInfo{options.str(), block, grid, fname, g_knames[kid]};
+}
+static KernelInfo
+get_kinfo_r2c_xga(const ConvolutionContext& ctx, const cellfft_param_t& p, const std::string& fname)
+{
```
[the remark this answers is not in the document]
jane-zxy: @atamazov ,
Sorry, I don't understand your review, would you explain it in more detail? thanks.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+    "sfft5x5_c2r_perm_relu_v31",
+    "sfft5x5_c2r_perm_relu_v32"
+};
+// clang-format on
+
+static uint32_t choose_cgemm_id(uint32_t m, uint32_t n)
+{
+    uint32_t mi = (m + 31u) >> 5;
+    uint32_t ni = (n + 15u) >> 4;
+    return ((1 ^ (mi & 1)) * 3 + ((ni & 3) == 0 ? 2 : (1 ^ (ni & 1))));
```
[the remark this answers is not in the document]
jane-zxy: it's not magic number, I'll try to add the comments.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+}
+
+#define PSIZE(n, m) (((n) + (m)) & (~(m)))
+
+static uint32_t choose_optimal_cell_id(uint32_t anx, uint32_t any, uint32_t bnx, uint32_t bny)
+{
+    uint32_t id = 1, n = 0x7fffffff;
```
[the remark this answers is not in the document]
jane-zxy: yes, you are right.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+// clang-format on
+
+static uint32_t choose_cgemm_id(uint32_t m, uint32_t n)
+{
+    uint32_t mi = (m + 31u) >> 5;
+    uint32_t ni = (n + 15u) >> 4;
+    return ((1 ^ (mi & 1)) * 3 + ((ni & 3) == 0 ? 2 : (1 ^ (ni & 1))));
+}
+
+#define PSIZE(n, m) (((n) + (m)) & (~(m)))
```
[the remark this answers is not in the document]
jane-zxy: this code is previously implemented by xiuyu, I've changed into inline function and give it a readable name.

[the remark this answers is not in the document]
jane-zxy: changed into inline func as review.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+{
+    // coef=alpha*(1.0/(cell_size*cell_size)), cell_size=16 or 32
+    float coef = alpha * (p.id == 0 ? 0.00390625f : 0.0009765625f);
+    h.Run(kern)(c, p.lda, p.cbks, a, b, p.lda, p.ldb, p.m, p.n, p.k, p.abks, p.bbks, coef);
+}
+static void lk_fft2d_r2c_perm_a(
+    const Handle& h, const Kernel& kern, const cellfft_param_t& p, void* dst, const void* src)
+{
+    h.Run(kern)(
+        dst, p.lda, p.abks, src, (p.dir != 2 ? 0 : 0x80000000) | p.m, p.anx, p.aldx, p.aldy);
```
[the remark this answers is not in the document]
jane-zxy: this is kernel arg is used to combine p.m and p.dir information together, this will save SGPR's usage.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+}
+static void lk_fft2d_r2c_perm_b(
+    const Handle& h, const Kernel& kern, const cellfft_param_t& p, void* dst, const void* src)
+{
+    h.Run(kern)(
+        dst, p.ldb, p.bbks, src, (p.dir == 0 ? 0 : 0x80000000) | p.n, p.bnx, p.bldx, p.bldy);
```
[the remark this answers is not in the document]
jane-zxy: same as above.

src/solver/conv_cellfft.cpp
```diff
@@ -0,0 +1,1166 @@
...
+}
+static void lk_fft2d_r2c_perm_pad(
+    const Handle& h, const Kernel& kern, const cellfft_param_t& p, void* dst, const void* src)
+{
+    h.Run(kern)(dst,
+                p.lda,
+                p.abks,
+                src,
+                p.m,
+                (p.dir != 2 ? 0 : 0x80000000) | p.pad_l,
```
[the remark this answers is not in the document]
jane-zxy: same as above.

src/solver/conv_cellfft.cpp
jane-zxy: yes, we are pretty sure. it's also a way to save SGPR usage.

src/solver/conv_cellfft.cpp
jane-zxy: changed as review.

src/solver/conv_cellfft.cpp
jane-zxy: changed as a enum type.

jane-zxy:
> > The solver is overloaded with magic numbers (each one is an implicit dependence). The code will be difficult to maintain. Please try to get rid of them as much as possible.
> 
> @atamazov
> Hi Atam,
> thanks for all your review, I'll update the pr later, and will re-request your review after I do corresponding changes.

@atamazov ,
magic number usage for division will decrease 25 instructions into only 3 instructions, so it does bring great benefit.
However, we do think it will make the code difficult to maintain, so we may think about remove some of them later, but not in this version.
cellfft algorithm will be updated with comming support for gfx908, etc.

atamazov:
@jane-zxy Unfortunately, postponed review comments tend to stay unresolved for months (see #560 for example). Please ask @daniellowell for explicit approval of magic numbers, if you still would like to postpone.

> magic number usage for division will decrease 25 instructions into only 3 instructions, so it does bring great benefit.

Please do not manually optimize the host code except bottlenecks proven by measurements.

jane-zxy:
> @jane-zxy Unfortunately, postponed review comments tend to stay unresolved for months (see #560 for example). Please ask @daniellowell for explicit approval of magic numbers, if you still would like to postpone.
> 
> > magic number usage for division will decrease 25 instructions into only 3 instructions, so it does bring great benefit.
> 
> Please do not manually optimize the host code except bottlenecks proven by measurements.

@atamazov 
No, we don't want to postponed this review process,  we want to keep current magic numbers' usage, and Pls help to review this pr, thanks.

atamazov:
@jane-zxy 
> No, we don't want to postponed this review process, we want to keep current magic numbers' usage, and Pls help to review this pr, thanks.

Please clarify if usage of magic numbers approved by @daniellowell. Meanwhile I will go ahead with reviewing.

atamazov:
@carlushuang It would be nice if you explain the reason of your latest commit into this PR, 10815d0.

atamazov (changes requested):
Please add layout filtering, see #687

atamazov:
@carlushuang Please explain 10815d0 or revert it.

atamazov:
Merge conflicts

carlushuang:
@atamazov @XiuYuLi  sorry for delay. https://github.com/ROCmSoftwarePlatform/MIOpen/commit/10815d0ac3eb96fc582ea4fe0f2185292aa70c5d this work around a bug in kernel selection in backward weight. The `ctx.out_height`, `ctx.out_width` in wrw case is indeed input height/width. And When input h/w <= 5x5, current will select kernel like `sfft4x4_r2c_perm_s5x5` or `sfft4x4_r2c_perm_s3x3` for input tensor transform, and will cause computation error or crash. Currently I have not enough time to change the kernel selection logic according to this, so here just filter out the `IsApplicable()` to let wrw not work while input image is <= 5x5

carlushuang:
@atamazov  let me add this according to #687

atamazov:

JehandadKhan:
Closing this PR until the developer can address merge conflicts.

Issue #142, opened 2020-04-08
[test_conv2d] Memory access fault


shurale-nkn:
Error in test_conv2d from MIOpenConvBwdWrWS2.cl

System:
Vega 20
HIP
ROCm 3.0.6
checked commits: ce986b21b , 3a716fe, fcd5563, 481d6b9,  cea60643c, a02171c92e3

For reproduce:
Reproduceable in combination of BWD + WRW only. The error is present in all combinations, but only in this one it can be seen as "Memory access fault". 
```
./bin/test_conv2d --float --cmode conv  --group-count 1 --input 1, 1, 8, 8 --weights 1, 1, 2, 2 --pads_strides_dilations 1 1 1 1 1 1 --pmode same --disable-forward
```

```
MIOpen(HIP): Info [Handle] stream: 0x2352180, device_id: 0
MIOpen(HIP): Info [BackwardDataGetWorkSpaceSize] 
MIOpen(HIP): Info [HipGetHccVersionImpl] HCC base: 3.0.19493
MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv2_COv3, MIOpen version 2.4.0.7759-cea60643c
MIOpen(HIP): Info [ForwardGetWorkSpaceSize] 
MIOpen(HIP): Info [BackwardWeightsGetWorkSpaceSize] 
MIOpen(HIP): Info [FindSolutionImpl] ConvBinWinogradRxSf2x3 (db access disabled)
MIOpen(HIP): Info [GetPerformanceConfig] 1
MIOpen(HIP): Info [BackwardDataGetWorkSpaceSize] 
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] requestAlgoCount = 1, workspace = 1024
MIOpen(HIP): Info [TryLoad] Find-db regenerating.
MIOpen(HIP): Info [FindSolutionImpl] ConvBinWinogradRxSf2x3
MIOpen(HIP): Info [FindSolutionImpl] Perf Db: record not found for: ConvBinWinogradRxSf2x3
MIOpen(HIP): Info [GetPerformanceConfig] 1
MIOpen(HIP): Info [EvaluateInvokers] ConvBinWinogradRxSf2x3: miopenSp3AsmConv_group_20_5_23_M_stride1: 0.02224 < 3.40282e+38
[...]

PR #341, opened 2020-07-13
Prevent UB in all ASM & OCL kernels that do not anticipate asymmetric padding

atamazov:
Resolves ASM & OCL leftovers of #142.

Disables configs that have asymmetric padding in all Solvers that do not anticipate it. See https://github.com/ROCmSoftwarePlatform/MIOpen/issues/142#issuecomment-625941366 for details. 

Note that currently asymmetric padding is possible only for even filter sizes. However this PR affects Solvers that support odd filters as well. This is done in order to avoid adding new implicit dependencies (on how we define “asymmetric padding”).

### DIFF
--- a/src/solver/conv_asm_1x1u.cpp
+++ b/src/solver/conv_asm_1x1u.cpp
@@ -381,6 +381,8 @@ bool ConvAsm1x1U::IsApplicable(const ConvolutionContext& params) const
         return false;
     if(!params.Is2d())
         return false;
+    if(params.IsAsymmetricPadH() || params.IsAsymmetricPadW())
+        return false;
     if(!params.rmv.IsV2orV3())
         return false;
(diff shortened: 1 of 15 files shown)


daniellowell:
This sound like a bug.

atamazov:
Formally speaking, the bug is a reason of an issue. However I do not have specific issues on hand (and would like to avoid spending time for finding these). So this is not a bug fix, technically.

TejashShah:
@atamazov LGTM.

atamazov:
> @atamazov What about igemm solvers?

Out of scope of this PR

atamazov:
#142 updated with Q for @asroy

TejashShah:
LGTM to be merged.

atamazov:
All tests passed, ready for merge.

PR #343, opened 2020-07-13
[igemm] Adding flexible layout support

jerryyin:
* Modified tensor_api and descriptor for layout updates
* Updated conv_driver to set tensor layouts
* Updated problem_description to avoid key collision
* Enabled implicit_gemm to get layouts from tensors

cc @whchung @zjing14 @asroy 

--------------

Note: This is going to be the first PR in the series to support igemm kernel generation. This PR aims to pave the road for introducing igemm runtime generated kernels. I mark the PR for draft pull request so that we start with the API design and then the implementation details.

~~Specifically, I want to get feedback on the `miopenSet4dTensorDescriptorLayout` API.~~

~~- As a reference, Cuda's related solution is a `cudnnTensorFormat_t`, which includes either `NCHW` or `NHWC`.~~
~~This is less desirable because igemm kernels supports way more layout than the two. However, it is also valid point if we want to hide this capability from user. If we follow their approach, it will require a number of new APIs on setting and getting the descriptor.~~
~~A compromise is to still use the current API, but invent a new enum: `miopenTensorFormat_t`. Since formats needs to be able to convey the difference between input/filter/output tensors, this will ended up in a huge array of enums one day. This does not seem to be a clean solution either.~~


jerryyin:
FYI I cannot reproduce the failures from CI, [here](http://micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLibs/pipelines/MIOpen/branches/zyin-layouts/runs/1/nodes/459/steps/501/log/?start=0). Please let me know if there is better ways to capture those issues.

Build command
`CXX=/opt/rocm/llvm/bin/clang++ cmake -DMIOPEN_USE_COMGR=On -DBUILD_DEV=On -DCMAKE_BUILD_TYPE=debug -DMIOPEN_GPU_SYNC=On -DMIOPEN_TEST_FLAGS=--disable-verification-cache`

Test command
`CTEST_PARALLEL_LEVEL=4 MIOPEN_DEBUG_IMPLICIT_GEMM_NON_XDLOPS_INLINE_ASM=0 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0 make -j64 check`

In my local machine yielded result of
`100% tests passed, 0 tests failed out of 55`

driver/conv_driver.hpp
```diff
@@ -531,6 +535,9 @@ int ConvDriver<Tgpu, Tref>::GetandSetData()
 template <typename Tgpu, typename Tref>
 int ConvDriver<Tgpu, Tref>::AddCmdLineArgs()
 {
+    inflags.AddInputFlag("in_layout", 'I', "nchw", "Input Layout (Default=nchw)", "string");
+    inflags.AddInputFlag("out_layout", 'O', "nkhw", "Output Layout (Default=nkhw)", "string");
+    inflags.AddInputFlag("fil_layout", 'f', "kcyx", "Input Layout (Default=kcyx)", "string");
```
atamazov, line 540: What about 3D convolutions? "nchw" is inappropriate, for example.

src/include/miopen/conv/problem_description.hpp
```diff
@@ -161,13 +161,13 @@ struct ProblemDescription
...
     std::size_t GetInStrideW() const { return GetW5(GetSpatialDims(), in.GetStrides()); }
-    std::string GetInLayout() const { return "NCHW"; }
+    std::string GetInLayout() const { return in.GetLayout(); }
     std::size_t GetInElementSize() const { return GetTypeSize(GetInDataType()); }
 
     std::size_t GetInSize() const
     {
         // clang-format off
-        return (GetInLayout() == "NCHW")
+        return (GetInLayout() == "nchw")
```
atamazov, line 170: Why this change?

driver/conv_driver.hpp
jerryyin: The in_layout option allows user to use any string as layout. So if it needs to be 3D convolution "nchwd", for example, user can set it. Setting it to "nchw" "nkhw" "kcyx" just so that it got defaulted to a set of coherent characters.

Once we agreed on `miopenSet4dTensorDescriptorLayout` interface that allows user to set the layout string field of a tensor, I can update the interface such that:

- Drop the `4d` in its name
- Formalize the set of characters and allowed length for layout string

src/include/miopen/conv/problem_description.hpp
jerryyin: The is related with the previous change: since input layout get defaulted to the lower case "nchw", the comparison is using "nchw". This can be addressed by, say, we formalized the layout string to be all lower or upper case.

Though I don't think I fully understand why it was originally written to compare with the upper cased "NCHW". Since `GetInLayout()` already get hard-coded to `NCHW` what's the point of the comparison here? Is it aiming for some future non-packed data?

src/conv/problem_description.cpp
```diff
@@ -67,15 +67,26 @@ void ProblemDescription::BuildConfKey(std::string& conf_key) const
...
+    }
+
     ss << 'x' << EncodeDataTypesForKey(GetInDataType(), GetWeightsDataType(), GetOutDataType());
     ss << 'x' << PrintDHW('x', GetSpatialDims(), GetPadD(), GetPadH(), GetPadW());
     ss << 'x'
        << PrintDHW(
               'x', GetSpatialDims(), GetKernelStrideD(), GetKernelStrideH(), GetKernelStrideW());
     ss << 'x' << PrintDHW('x', GetSpatialDims(), GetDilationD(), GetDilationH(), GetDilationW());
     ss << 'x' << GetGroupCount();
-    ss << 'x' << (GetDirection() == Direction::Forward ? "1" : "0");
```
jerryyin, line 78: TODO: reenable this. Any objection for me to adopt the approach from `Serialize()` function?
`F`/`B`/`W` or `forward`/`backwarddata`/`backwardweights`

src/include/miopen/conv/problem_description.hpp
```diff
@@ -161,13 +161,13 @@ struct ProblemDescription
     std::size_t GetInStrideD() const { return GetD5(GetSpatialDims(), in.GetStrides()); }
     std::size_t GetInStrideH() const { return GetH5(GetSpatialDims(), in.GetStrides()); }
     std::size_t GetInStrideW() const { return GetW5(GetSpatialDims(), in.GetStrides()); }
-    std::string GetInLayout() const { return "NCHW"; }
+    std::string GetInLayout() const { return in.GetLayout(); }
```
atamazov, line 164: This function is used in `ProblemDescription::Serialize` that is used to generate database keys. Changing from "NCHW" to "nchw" would break find-db and, most likely, perf-db functionality.

src/conv/problem_description.cpp
```diff
@@ -96,6 +107,8 @@ void ProblemDescription::Serialize(std::ostream& stream) const
     stream << sep << PrintDHW('x', GetSpatialDims(), GetDilationD(), GetDilationH(), GetDilationW());
     stream << sep << GetBias();
     stream << sep << GetInLayout();
+    stream << sep << GetWeightsLayout();
```
atamazov, line 110: We must maintain this function backward compatible as it generates database keys. When layouts are default (NCHW-NKHW-KCYX) the modified function must return literally the same key as it does now.

/cc @JehandadKhan as this requirement is also important for Tuna, I think.

src/conv/problem_description.cpp
atamazov: Most likely , this is harmless, but I am not sure. Please discuss this with @DrizztDoUrden

src/include/miopen/conv/problem_description.hpp
jerryyin: Sounds good thanks for reminding. Looks like to be consistent I will have to adopt all upper letters for layouts.

src/conv/problem_description.cpp
jerryyin: Thanks. Will branch according to whether it is default layouts.

src/include/miopen/conv/problem_description.hpp
atamazov: > since input layout get defaulted to the lower case "nchw"

This change is undesired, please see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343#discussion_r456054013.

> Though I don't think I fully understand why it was originally written to compare with the upper cased "NCHW".

I am far from even moderate understanding. All this is legacy stuff, survived 4 years of refactoring and other changes. Originated from https://github.com/AMDComputeLibraries/MLOpen/pull/5/files#r456075197 4 years ago. It looks like dead code now.

And it will become a bug as soon as `GetInLayout()` will begin to return "NCDHW" for 3D tensors.

[Recommendation] Get rid of this in a separate PR.

jerryyin: Sounds good, will return here, but file a separate PR to return `GetInBatchSize() * GetInChannels() * GetInDepth() * GetInHeight() * GetInWidth() * GetInElementSize()` directly.

atamazov: My previous recommendation (getting rid of ternary) also applies to GetOutSize, GetWeightsSize

atamazov: > Sounds good, will revert here, but file a separate PR 

I recommend keeping this unchanged here, to avoid merge conflicts. Up to you of course.

src/conv/problem_description.cpp
JehandadKhan: We actually need it when parsing the database files to fix swapped elements by the problem descriptor.

jerryyin: Thanks for reminding, reverted to original 1/0

atamazov: > We actually need it when parsing the database files...

Database keys are produced by Serialize(), and direction is encoded there as F, B or W. NetworkConfig is used only in find-db, as a solution-specific, opaque payload.

DrizztDoUrden:
Alternative would likely be to make conv ctx to reuse serialization from conv::ProblemDescription, but our current serialization process doesn't allow post-deserialization handling, so that won't be trivial.

src/conv/problem_description.cpp
DrizztDoUrden: I think network confing string representation is only used in conjunction with algo name string containing direction. But it would be better to check that, I am not constantly looking through every change to the library.

jerryyin:
> Alternative would likely be to make conv ctx to reuse serialization from conv::ProblemDescription, but our current serialization process doesn't allow post-deserialization handling, so that won't be trivial.

No worries, It's been fairly straightforward to copy-o-past the solution to another place. I had it addressed in my last commit. I don't see a problem in the scope of this review. Though in the development process this has been very confusing and took me a while to realize why update in one problem description does not fix an issue I was having - Because there are two problem descriptions!

DrizztDoUrden:
I ll have to clean up all the code to finally move to the new structure, but that got postponed for a while already.

src/conv/problem_description.cpp
```diff
@@ -95,7 +110,16 @@ void ProblemDescription::Serialize(std::ostream& stream) const
...
+    if (GetInLayout() == "NCHW" && GetWeightsLayout() == "KCYX" && GetOutLayout() == "NKHW")
+    {
+        stream << sep << GetInLayout();
+    }else {
+        stream << sep << GetInLayout();
+        stream << sep << GetWeightsLayout();
+        stream << sep << GetOutLayout();
+    }
+    stream << sep << GetWeightsLayout();
+    stream << sep << GetOutLayout();
```
JehandadKhan, line 122: For backwards compatibility the output layout should not be serialized either.

jerryyin: Thanks for catching this! Think I accidentally forgot about those two lines when adding the if condition. Fixed in d0f3423

jerryyin:
Wasn't familiar with MIOpen development process, so gentle ping here. This PR has sit still for a while, what does it take to proceed with approval/merge? Is it blocked by CI failures? Let me know how to re-kick the CI for a new run.

daniellowell:
All please review this PR.

daniellowell:
@atamazov Please review this PR.

daniellowell:
@jerryyin Please merge the latest `develop` into your branch. This should resolve being stuck at the COMgr stage.

jerryyin:
@daniellowell Thanks, rebased to latest state of develop branch.

src/include/miopen/tensor.hpp
```diff
@@ -182,6 +185,8 @@ struct TensorDescriptor : miopenTensorDescriptor
     bool packed;
 
     miopenDataType_t type = miopenFloat;
+
+    std::string layout;
```
pfultz2, line 189: We should not be storing the layout as a string. The layout comes from the order of strides in the tensor descriptor.

include/miopen/miopen.h
```diff
@@ -538,6 +538,15 @@ MIOPEN_EXPORT miopenStatus_t miopenCreateTensorDescriptor(miopenTensorDescriptor
 MIOPEN_EXPORT miopenStatus_t miopenSet4dTensorDescriptor(
     miopenTensorDescriptor_t tensorDesc, miopenDataType_t dataType, int n, int c, int h, int w);
 
+/*! @brief Set the layout to a tensor descriptor
+ *
+ * API for setting the layout to a tensor descriptor
+ * @param tensorDesc Tensor descriptor type (input)
+ * @param layout     Tensor layout (input)
+ */
+MIOPEN_EXPORT miopenStatus_t miopenSetTensorDescriptorLayout(miopenTensorDescriptor_t tensorDesc,
```
pfultz2, line 547: We should not have this API. The layout can be set by using the strides or we can provide setters for the common layouts, such as `miopenSetCNHWTensorDescriptor` or `miopenSetNHWCTensorDescriptor`.

src/include/miopen/tensor.hpp
```diff
@@ -173,6 +173,9 @@ struct TensorDescriptor : miopenTensorDescriptor
 
     std::string ToString() const;
 
+    void SetLayout(const std::string& l);
+    std::string GetLayout() const { return this->layout; }
```
pfultz2, line 177: This is wrong as well. The layout should come from the order of the strides.

include/miopen/miopen.h
daniellowell: I don't understand the underlying reason why this should not be an API. The other options are good sure, but what is the actual objection?

pfultz2: The main problem is we have two ways to represent the layout. We can already specify CNHW layout with `miopenSetTensorDescriptor(tensorDesc, dataType, 4, {N, C, H, W}.begin(), {H*W, N*H*W, W, 1}.begin())`. Although this may not always be the most convenient for some users, which is why we should provide `miopenSetCNHWTensorDescriptor`. 

Using a string for the layout leads to us erroneously using the string to determine the layout instead of using the actual layout of the tensor. We could make the layout string and the actual layout of the tensor work, but this will double our test coverage. 

Also, MIGraphX will most likely specify the layout using the strides and not a layout string as it would be too cumbersome to infer.

src/include/miopen/tensor.hpp
pfultz2: Actually, the labels should be passed to `GetLayout` since the labels depend on the input type and not on the tensor layout. This can be implemented as:

```cpp
template <class Vector, class Op>
inline std::vector<int64_t> sort_permutation(const Vector& data, Op op)
{
    std::vector<std::int64_t> result(data.size());
    std::iota(result.begin(), result.end(), 0);
    std::sort(result.begin(), result.end(), [&](auto x, auto y) { return op(data[x], data[y]); });
    return result;
}

std::string TensorDescriptor::GetLayout(std::string labels) const
{
    auto result = labels;
    auto p = sort_permutation(strides, std::greater<>{});
    std::tranform(p.begin(), p.end(), result.begin(), [&](auto i) {
        return labels[i];
    });
    return result;
}
```

So then you can call it with `GetLayout("NCHW")` and if the layout is CNHW it will CNHW.

include/miopen/miopen.h
atamazov: +1 to Paul, I do not like strings here due to many reasons.

However I do not understand how `miopenSetTensorDescriptor(tensorDesc, dataType, 4, {N, C, H, W}.begin(), {H*W, N*H*W, W, 1}.begin())` can do the job. This doesn't look like a C code to me. Perhaps `{N, C, H, W}.begin()` and `{H*W, N*H*W, W, 1}.begin())` represent beginnings of buffers, but I do not see how the library can derive information about layout from two pointers. Maybe I just need to sleep some time? ;)

atamazov:
@jerryyin 
> Wasn't familiar with MIOpen development process, so gentle ping here. This PR has sit still for a while, what does it take to proceed with approval/merge? Is it blocked by CI failures?

We often postpone review until all CI failures fixed, because fixing the CI problems may require significant code changes and reviewers will have repeat their work after that.

Please look at CI failures. If your PR has nothing to do with them, please inform us and explicitly ask for reviewing.

Please also tell how you've tested your PR (in the topmost comment). For example, often I pull and re-test branches during a review process.

include/miopen/miopen.h
pfultz2: >  This doesn't look like a C code to me.

This is not C code its C++ code. The `{N, C, H, W}` creates an initializer list and `.begin()` returns a pointer to it, so it can be passed to the C API. I was just trying to shorten it to one line. Another way to write it would be(assuming `N`, `C`, `H`, and `W` are constants):

```cpp
std::vector<int> lens = {N, C, H, W};
std::vector<int> strides = {H*W, N*H*W, W, 1};
miopenSetTensorDescriptor(tensorDesc, dataType, 4, lens.data(), strides.data());
```

> but I do not see how the library can derive information about layout from two pointers.

It derives the layout from the strides. As I showed below with the `GetLayout` function, `sort_permutation` will return the permutation `0,1,2,3` for NCHW layout and in the example above for CNHW it will return the permutation `1,0,2,3`.

If we pass the labels in as a string to `GetLayout` for the default permutation(ie `0,1,2,3`) then it will put the labels in the order of its layout. So `GetLayout("NCHW")` will return `NCHW` with permutation of `0,1,2,3` and it will return `CNHW` for permutation of `1,0,2,3`.

jerryyin: Thanks for the comments guys. It makes total sense after looking through especially @pfultz2 comments and suggestions. Using strides to infer the layout does make it less ambiguous and make it able to take advantage the existing API as well. Will push a new commit to address this.

src/include/miopen/tensor.hpp
jerryyin: Nice, it gives client the flexibility to specify the layout character strings too. I like this implementation.

jerryyin:
@atamazov Thanks for reminding. We stand on the same page that reviewers should start reviewing only after fixing CI failures. Somehow I had the mis-understanding that CI wasn't fully functional due to `COMGR issues` but apparent it was fixed a while ago. 

Regarding with testing, it is less straightforward for this PR. I kept an `igemm-upstream` branch that had the integration of MIOpen and MLIR done. I was able to test a number of MIOpenDriver configs in different layouts in the context of that branch and things has been working fine. (No tuning related testing on my side, that might be how I missed the existing CI failures.) 

Considering that pushing all the changes in that branch at one go is messy and makes a huge reviewer effort, I decided to split that into a number of vertical PRs, just so that it is easier to review. Having said that, I'd be happy to share the instructions of how to test in the context of that branch, however, it might be too much to ask for a reviewer to do. What I will probably do is to also pull in the minimal amount of changes in `conv_driver` to make the PR test-able in its own context.

include/miopen/miopen.h
atamazov: @pfultz2 Thanks! it became clear to me in the morning, indeed :facepalm:

jerryyin:
Please hold off review this PR. WIP to make CI happy.

jerryyin:
Help needed: I can't figure out what make CI to consistently fail on two targets (in `test_conv2d` test case):
 - Hip clang debug target, in this [run](http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/zyin-layouts/16/pipeline)
 - gfx908 Hip debug target, in another [run](http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/zyin-layouts/17/pipeline)

For the first one, I have build the docker container from scratch and followed exactly on how CI builds and tests and all tests can pass just fine. For the second one, I tried to follow the process in a gfx906 but it always failed on building hcc. The ROCm version on the bare metal is 3.7 and is up to date. I don't know how to proceed further at this point.

src/conv/problem_description.cpp
```diff
@@ -77,6 +77,7 @@ void ProblemDescription::BuildConfKey(std::string& conf_key) const
               'x', GetSpatialDims(), GetKernelStrideD(), GetKernelStrideH(), GetKernelStrideW());
     ss << 'x' << PrintDHW('x', GetSpatialDims(), GetDilationD(), GetDilationH(), GetDilationW());
     ss << 'x' << GetGroupCount();
+    ss << 'x' << (GetDirection() == Direction::Forward ? "1" : "0");
```
jerryyin, line 88: TODO: remove the leftover from rebase

jerryyin:
CI has a full pass, [here](http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/zyin-layouts/28/pipeline).

All please continue the review. Thanks.

jerryyin:
Gentle ping @daniellowell @pfultz2. Let me know how to proceed with this PR.

daniellowell:
@atamazov @JehandadKhan Changes have been made please re-review.

DrizztDoUrden (dismissed):

src/include/miopen/tensor.hpp
```diff
@@ -173,6 +175,30 @@ struct TensorDescriptor : miopenTensorDescriptor
...
+    inline std::vector<int64_t> sort_permutation(const Vector& data, Op op) const
+    {
+        std::vector<std::int64_t> result(data.size());
+        std::iota(result.begin(), result.end(), 0);
+        std::sort(
+            result.begin(), result.end(), [&](auto x, auto y) { return op(data[x], data[y]); });
+        return result;
+    }
+
+    std::string GetLayout(std::string labels) const
```
atamazov, line 188: This involves creation/deletion of several containers, fillng, sorting etc. Not that fast as a simple getter should be. Shall we rename it to `ComputeLayout()`?

atamazov, line 188: Another consideration. When we support different layouts, layout comparisons (e.g. in `aSolver::IsApplicable()`) will happen quite often. Getting layouts is not very fast. Comparison of strings too. This could lead to some perf drop. Do you have a plan of optimization? Ideas?

src/include/miopen/tensor.hpp
```diff
@@ -173,6 +175,30 @@ struct TensorDescriptor : miopenTensorDescriptor
...
+    {
+        if(labels.size() != strides.size())
+        {
+            MIOPEN_THROW(
+                "Invalid labels size. Layout labels size must be equavalent to stride size");
+        }
+
+        auto result = labels;
```
atamazov, line 196: It seems like assignment is useless.

atamazov (changes requested):
(Sorry for delaying the review)

driver/conv_driver.hpp
```diff
@@ -531,6 +531,9 @@ int ConvDriver<Tgpu, Tref>::GetandSetData()
 template <typename Tgpu, typename Tref>
 int ConvDriver<Tgpu, Tref>::AddCmdLineArgs()
 {
+    inflags.AddInputFlag("in_layout", 'I', "NCHW", "Input Layout (Default=NCHW)", "string");
+    inflags.AddInputFlag("out_layout", 'O', "NCHW", "Output Layout (Default=NCHW)", "string");
+    inflags.AddInputFlag("fil_layout", 'f', "NCHW", "Input Layout (Default=NCHW)", "string");
```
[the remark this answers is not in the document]
jerryyin: Please refer to [this comment](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343#discussion_r474778988). Because of the reason there I decided that updating layout string to include non-NCHW is orthogonal to this review. Plus I don't feel strong about using the similar set of letters being a big issue here.

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -201,8 +201,6 @@ bool ConvMPBidirectWinograd<WinoDataH, WinoFilterH, WinoDataW, WinoFilterW>::IsA
         return false;
     }
 
-    assert(params.weights_layout.length() == 0); // _weights_layout is not supported yet
```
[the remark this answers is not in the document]
jerryyin: Sounds reasonable. Though I'd want more clarification in this check: In the original code, why is only `weights_layout` checked here (instead of three)?

I'm asking this because most probably `params.IsLayoutDefault()` will check all three layout on whether they are default, then it becomes an un-equivalent code change compared to original one.

src/include/miopen/tensor.hpp
jerryyin: I see this in two ways:

- It does get more expensive, but in very limited manner. The 4-letter layout string is very easy to manipulate and adds only linear time complexity here. I don't think this is a big problem leaving it as-is.

- On the other hand, we can certainly cache it to a member variable (such that only the first time `GetLayout()` called that it is actually computed, otherwise it is just returning cached copies.). I can certainly implement this. Though I want to make sure you guys are okay with it in the sense that:
  1. It stores duplicated information against strides, as pointed out from original review comments.
  2. It is not very "functional", per say.

src/include/miopen/tensor.hpp
jerryyin: This is the solution I copy-pasted from Paul's implementation. It copy constructed the `result` from `lables` (instead the alternative is to `push_back()` in the transform stage), I assume it is slightly faster than the alternative.

src/include/miopen/tensor.hpp
jerryyin: We can decide on the name depends on how next review comments goes. Should we keep the function untouched, I shall rename it to `Compute`. Should we decide to cache the result, I will keep as is.

pfultz2: I think renaming it to `ComputeLayout` makes sense since it needs to do a caculation.

pfultz2: > On the other hand, we can certainly cache it to a member variable (such that only the first time GetLayout() called that it is actually computed, otherwise it is just returning cached copies.). 

I dont think this is a good idea. `GetLayout`(or `ComputeLayout`) should be `const`. 

It would be better to store the layout string in the `ConvolutionContext` so that it only needs to be computed once.

src/include/miopen/tensor.hpp
pfultz2: Yes, this is to avoid calling `push_back` as incrementing a pointer should be faster.

src/include/miopen/tensor.hpp
```diff
@@ -173,6 +175,30 @@ struct TensorDescriptor : miopenTensorDescriptor
 
     std::string ToString() const;
 
+    template <class Vector, class Op>
+    inline std::vector<int64_t> sort_permutation(const Vector& data, Op op) const
```
pfultz2, line 179: This can be declared as `static`.

src/solver/conv_MP_bidirectional_winograd.cpp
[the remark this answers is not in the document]
atamazov: > why is only weights_layout checked here (instead of three)?

Because one line of useless code (comment) is enough to remind ;)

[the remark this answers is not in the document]
atamazov: > then it becomes an un-equivalent code change compared to original one.

Original code does not check anything ;)

src/include/miopen/tensor.hpp
jerryyin: @pfultz2 Yah, think we mean the same, wasn't very clear about where to cache the layout string in my original comments. 

Addressed in latest commit.

src/include/miopen/tensor.hpp
jerryyin: Done. Thanks

src/include/miopen/tensor.hpp
jerryyin: Done, now I have both `ComputeLayout` and `GetLayout`.

jerryyin:
FYI, the CI has some quite inconsistent result over the same code base again. I'm not sure its something that has been noticed. For example, an earlier run [(32) here](http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/zyin-layouts/32/pipeline/) complains that ROCm is not available on `gfx908 Hip debug`'s docker building step. It looks clear to me that different CI machines are configured differently which caused this inconsistency. My pr is hitting on this multiple times.

daniellowell:
@jerryyin We had 2 bad nodes in our CI. I've disabled them.

jerryyin:
Jenkins all passing [here](http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/zyin-layouts/40/pipeline). Let me know if you guys, @atamazov @pfultz2 have any further feedbacks. Thanks.

src/include/miopen/tensor.hpp
atamazov: Please add a comment that explains this. Otherwise the code looks like an error.

jerryyin: Done in f58abdaa7aa8b69d840066b9b20cae8ef58dc75c

driver/conv_driver.hpp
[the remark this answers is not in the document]
atamazov: This seems Ok as an intermediate solution. However, what is your future plans about this? I mean support for independent layouts for input, output and weights. Right now these defaults looks confusing.

I suggest opening a dedicated ticket, possibly labeled `request_for_comments`, for discussion and making/documenting design decisions.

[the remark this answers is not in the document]
jerryyin: I'm fine with adding this as an issue but not leading this effort. I'm willing to review and provide my feedback if there is a proposed solution, but I don't know what that would be at the moment. The primary reason has been twofold:

- I have very little idea about the history and previous design decisions for all the layout related implementation. This makes me a very poor owner for making a fundamental change to this solution, just because of how widely `buffer_info.hpp` has been included in MIOpen. I could be making (or propose) a dozen of changes not foreseeing the impact to existing code base. Therefore I'd like to scope my effort to the code base that is strictly igemm infrastructure where I could offer the most help.
- From the igemm perspective, I don't mind the actual string representation of layout - any 4-letter string will do the job. This does not impact the functionality of MIOpen existing code base - making it less of a priority. Readability-wise, it is not as good as NCHW-KCYX-NKHW, but each layout does come closely with a tensor so wouldn't be tricky to handle.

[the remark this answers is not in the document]
atamazov: > I have very little idea about the history and previous design decisions

Nobody has I guess ;) I would say "it just happened".

> I could be making (or propose) a dozen of changes not foreseeing the impact to existing code base. 

That is why the reviewers are here. Working together, we all learn.

> Therefore I'd like to scope my effort to the code base that is strictly igemm infrastructure where  I could offer the most help.

I'll keep in mind this. However, as you can see, this is often impossible.

#509 created.

atamazov:
@DrizztDoUrden Could you please have a look at this? It looks acceptable to me, but perhaps you have some substantial comments.

driver/conv_driver.hpp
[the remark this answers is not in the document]
jerryyin: Appreciate the feedback and for helping creating this issue. It is unfortunate that I still have couple of changes pending and I'm trying to make more progress in this. I'm trying to keep focused and have it wrapped up sooner rather than later as maintaining a working feature branch is also non-trivial. To give more details here, I'd prioritize on a functional igemm code generation, just so as to unblock the tuning implementation for igemm (Adding layout for all tensors will impact perfdb and tuna, another collaboration point with MIOpen developers.)

daniellowell:
@atamazov Your review is blocking this PR. Are there still outstanding issues?

daniellowell:
@JehandadKhan  If the changes have resolved your issues, please unblock this PR.

JehandadKhan (approved):
This PR can go in, in a subsequent PR i would add support for perf db to support these changes.

atamazov (approved):
Tested for performance. No regressions. LGTM.

pfultz2 (approved):

Issue #337, opened 2020-07-09
Dynamic Kernels Assignments


daniellowell:
### Purpose
This project exists to minimize our reliance on compile time parameterization in MIOpen's source kernels. The goal isn't to sacrifice performance, but rather determine a ways of reducing the compile time overhead of the first time iteration of neural networks using MIOpen. 

### Strategy
For some of these kernels the task is pretty straight forward; take the compile time parameters and move them into runtime parameters. In some cases this can be done without affecting performance. It may be the case, and often, that all compile time parameters may not be moved into the runtime without seriously affecting performance. In those cases we should identify those parameters that networks would change least frequently such that compiles are minimized. If remaining compile-time parameters do not reduce significantly the number of compiles, then it may be the case the the kernel should be converted to assembly code.

#### Data collection
- Compile time impact for various networks
- Performance impact of compile time modifications
- Data collection on solver usage

#### Structural
- [ ] Precompiled kernels for non-convolution kernels (@JehandadKhan) ROCm 3.8
- [ ] Tuna prepared for bin-cache generated kernels (@JehandadKhan, @alexandraBara, @cderb )
- [ ] Expand the methodology of ASM-iGEMM conversion to non-iGEMM kernels and non-convolutional kernels (@carlushuang, @shaojiewang , @jane-zxy , @fronteer )

#### Convolution Changes
Priority: HIGH
- [ ] Convolutions 
   - [ ] iGEMM HIP source refactor (@asroy) 7/24/2020	
   - [x] ASM-iGEMM (@carlushuang , @shaojiewang, @jane-zxy, @fronteer)
      - [x] ASM-iGEMM dynamic non-xDLOPs (ROCm 3.8)
      - [ ] ASM-iGEMM static xDLOPs fp32 (ROCm 3.10)
      - [ ] ASM-iGEMM static xDLOPs fp16/bfp16 (ROCm 4.0)

   - [ ] Make ASM iGEMM kernels only in hybrid MIOPEN_FIND_MODE (ISSUE: #299) (@zjing14 ) ROCm 3.8
   - [ ] Data collection for solver usage via Tuna (@ce1adon)
        - Dependent-tasks:
           - [ ] OpenCL legacy kernels --> reduce applicability range of solvers (@ce1adon , @JehandadKhan )
           - [ ] OpenCL legacy kernels --> delete unused solvers (@ce1adon , @JehandadKhan )
[...]

PR #347, opened 2020-07-15
add vector-load tune param to improve fp16 frwd

jane-zxy:
GemmN can be vector-load in some cases, and normally we choose maximum vector size, however, it's not the best solution for fp16 for following reason:
1.  increased VGPRs overhead;
2.  increased LDS overhead;
so make GemmBThreadDataPerRead_GemmN tunable, it will improve 7.3% average efficency for vectorizable cases. 
Detailed benefit gain is here: 
https://amdcloud-my.sharepoint.com/:x:/g/personal/xiaoyazh_amd_com/EZigTf_rnk1KnkfrWQDHj-0BK6fQT6orElK_j20W2xCTCA?e=fs5Tuo

any comments are welcome!


src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -81,6 +100,10 @@ bool PerformanceImplicitGemmForwardV4R4Xdlops::SetNextValue()
     {
         // list performance parameters in reverse order, in order for tuning to iterate over the
         // range in normal order
+#if CK_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM
+        if(!NextTwoPower<1, 8>(GemmBThreadDataPerRead_GemmN))
```
atamazov, line 104: This increases tuning time 4 times. Is it acceptable?

carlushuang:
This PR can indeed increase performance. However, compile time is increased due to one more tunable param introduced. As preferred in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/337, we may need consider reduce compile time first. @asroy need your dicision.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
jane-zxy: @atamazov , thanks for your comments.
This tune param is only valid for fp16's vectorizable case, also there's more restriction inside of "PerformanceImplicitGemmForwardV4R4Xdlops::CalculateGemmBBlockCopyPerformancePara", pls have a reference.

normally, for fp16's vectorizable cases, tuning time is 2 times of original time.

asroy:
@jane-zxy Please make whether to enable this extra tuning parameter a run-time choice controlled by env variable, and the default behavior is to not enable this extra tuning. You can see example here:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/46aa3d8ac5b569dfd537be5ae3fc35b694224c95/src/solver/implicitgemm_util.hpp#L13

atamazov:
> Please make whether to enable this extra tuning parameter a run-time choice controlled by env variable, and the default behavior is to not enable this extra tuning. 

Please note that the format of perf-config records should NOT depend on this. When new tuning param is disabled:
- its value should remain in the perf-config
- it should be locked at its default value during auto-tuning
- the function that checks validity of perf-configs should allow for values other than default, because these values may come from perf-db (that may contain records tuned when the new parameter is _enabled_).

If the above is not fulfilled, we will have perf-db compatibility problems.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
atamazov: Is it so that this implementation locks the value of this new param at some default value for all cases except "fp16 vectorizable"? And this is done in `PerformanceImplicitGemmForwardV4R4Xdlops::CalculateGemmBBlockCopyPerformancePara()`?

If yes, them why this is done there but not right here?

atamazov:
@carlushuang 
> However, compile time is increased due to one more tunable param introduced.

This is not actually so. Build (or compilation) time of the kernel remains the same. Only auto-tuning time is increased.

carlushuang:
> @carlushuang
> 
> > However, compile time is increased due to one more tunable param introduced.
> 
> This is not actually so. Build (or compilation) time of the kernel remains the same. Only auto-tuning time is increased.

Yes, sure

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
jane-zxy: the reason is that inside of 'SetNextValue', 'ConvolutionContext' is not accessible, so there's no way we can do seperate diverse for fp16.

atamazov: It seems like missing access to `ConvolutionContext` enforces us to use weird designs. However `ComputedContainer` is intended to ease the code (and the life of Solvers' developers).

We can make `ConvolutionContext` accessible in `SetNextValue()`.

What do you think?

/cc @shurale-nkn @asroy @ce1adon @zjing14 @daniellowell @DrizztDoUrden

jane-zxy: @atamazov , I agree, and adding convolutionContext into setNextValue would ease solver's tunning implementation, however this change would be framwork-level impacted. so I suggest finish this pr first and then open another pr to fix it, what do you think?

jane-zxy: @shurale-nkn @asroy @ce1adon @zjing14 @daniellowell @DrizztDoUrden
Could you pls comments on idea of making ConvolutionContext accessible in SetNextValue()? 
and since its impacts framework change, how we can fulfill it?

atamazov: Obviously there is no direct impact on the existing code. It just _opens ways_ to write new code better, and to refactor existing code to make it cleaner.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -440,7 +447,27 @@ PerformanceImplicitGemmForwardV4R4Xdlops::CalculateGemmBBlockCopyPerformancePara
         // calculate threadwise copy size
         auto data_per_thread_copy =
             std::max(1, (GemmKPerBlock * GemmNPerBlock * GemmKPack) / block_size);
-
+        if(miopen::IsEnabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM{}))
+        {
+            if(ctx.IsFp16())
```
asroy, line 452: There is divergence between fp16 and other data-type?

src/solver/implicitgemm_util.hpp
```diff
@@ -11,6 +11,12 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_XDLOPS)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_XDLOPS_EMULATE)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_IMPLICIT_GEMM_XDLOPS_INLINE_ASM)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_BLOCK_SYNC_LDS_WITHOUT_SYNC_VMEM)
+/* this fix is for fp16 xdlops vectorizable kernels due to followings, we may revisit this fix after
+  compiler fix:
+  1. compiler issues(25% impact)
+  2. LDS write performance(75% impact)
+*/
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM)
```
asroy, line 19: Please move it inside ```src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp```, and rename it to something like ```MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_FWD_V4R4_XDLOPS_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM```, because it's only for that solver

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -465,6 +492,18 @@ PerformanceImplicitGemmForwardV4R4Xdlops::CalculateGemmBBlockCopyPerformancePara
         // vector write into LDS
         DstDataPerWrite_GemmKPack = gcd(DstDataPerWrite_GemmKPack, data_per_thread_copy_gemmkpack);
 
+        if(miopen::IsEnabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM{}))
+        {
+            if(ctx.IsFp16())
```
asroy, line 497: Divergence between fp16 and other datatype. Is this intended?

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -81,6 +84,11 @@ bool PerformanceImplicitGemmForwardV4R4Xdlops::SetNextValue()
     {
         // list performance parameters in reverse order, in order for tuning to iterate over the
         // range in normal order
+        if(miopen::IsEnabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM{}))
```
asroy, line 87: @atamazov Please review if this implementaion satisfy requirement for perf-db.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
jane-zxy: yes, I tested on fp32 bfp16, there's no big improvement after adding this tunning parameter, the reason, i believe is fp16's the biggest flops, it make LDS operation relatively bottleneck.
so to avoid extra tunning time, i enable this parameter only for fp16.

src/solver/implicitgemm_util.hpp
jane-zxy: ok, I'll change it accordingly.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -465,6 +500,19 @@ PerformanceImplicitGemmForwardV4R4Xdlops::CalculateGemmBBlockCopyPerformancePara
         // vector write into LDS
         DstDataPerWrite_GemmKPack = gcd(DstDataPerWrite_GemmKPack, data_per_thread_copy_gemmkpack);
 
+        if(miopen::IsEnabled(
+               MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_FWD_V4R4_XDLOPS_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM{}))
+        {
+            if(ctx.IsFp16())
+            {
+                if((SrcDataPerRead_GemmN > 1) &&
```
asroy, line 508: What does this logic means? 
What is the purpose of this logic?
If we remove this logic, the code would not compile or it would run slow?

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
asroy: > It seems like missing access to `ConvolutionContext` enforces us to use weird designs. However `ComputedContainer` is intended to ease the code (and the life of Solvers' developers).
> 
> We can make `ConvolutionContext` accessible in `SetNextValue()`.
> 
> What do you think?

That's nice. Having accessing to ```ConvolutionContext``` anywhere is always good.

jane-zxy: @atamazov, @asroy  I've tried to make ConvolutionContext accessible in SetNextValue(), however other solver's SetNextValue() does not use ConvolutionContext, so there would be warning while compile. so code change can not pass the CI.
So I change to add a local value of m_isFp16 inside of PerformanceImplicitGemmForwardV4R4Xdlops, since it's only needed inside of this solver currently. 
Pls review the change.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -465,6 +506,11 @@ PerformanceImplicitGemmForwardV4R4Xdlops::CalculateGemmBBlockCopyPerformancePara
         // vector write into LDS
         DstDataPerWrite_GemmKPack = gcd(DstDataPerWrite_GemmKPack, data_per_thread_copy_gemmkpack);
 
+        if(!(data_per_thread_copy_gemmkpack > 0 && data_per_thread_copy_gemmk > 0 &&
```
jane-zxy, line 509: it's for clang tidy

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
jane-zxy: @asroy , please help to review, tunning is disable for non-fp16, and fp16 tunning filterring  is moved to IsFastToBeUsedForTuning, thanks.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -142,15 +160,16 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
         }
         else if(ctx.IsFp16())
         {
-            tmp = {256, 256, 8, 128, 128, 8, false, true};
-
+            tmp              = {256, 256, 8, 128, 128, 8, false, true, 1};
```
asroy, line 163: This heuristic would likely make the tuning parameter worse than before, because ```GemmBThreadDataPerRead_GemmN``` is always set to 1.

After discussion with @daniellowell, I think this would be an issue. Please call ```CalculateGemmBBlockCopyPerformancePara``` between line 230 and line 231 to determine what  ```GemmBThreadDataPerRead_GemmN``` should be.

asroy: @daniellowell 

@jane-zxy corrected me, that if when ```MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_FWD_V4R4_XDLOPS_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM=0```, which is true by default, then the heuristic tuning parameter will be the same as before. So this PR is fine

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -81,6 +91,12 @@ bool PerformanceImplicitGemmForwardV4R4Xdlops::SetNextValue()
...
         // range in normal order
+        if(miopen::IsEnabled(
+               MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_FWD_V4R4_XDLOPS_ADD_VECTOR_LOAD_GEMMN_TUNE_PARAM{}))
+        {
+            if(!NextTwoPower<1, 8>(GemmBThreadDataPerRead_GemmN))
+                break;
+        }
         if(!NextFlag<false, true>(GemmBThreadCopyMoreGemmKPack))
             break;
         if(!NextFlag<false, false>(GemmAThreadCopyMoreGemmK))
```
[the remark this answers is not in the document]
jane-zxy: Hi artem, 
what do you mean by 'semantics'? do you mean the usage scenario? if yes, pls refer to the comment of this pr.
However, I do see I should also deal with different kind of data type, for fp32, it's <1,4>, for fp16, it's <1,8> here. I'll fix it.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -120,6 +136,8 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
                 {
                     // list in reverse order of importance,
                     // and favor large GEMM
+                    if(!PreviousTwoPower<1, 8>(tmp.GemmBThreadDataPerRead_GemmN))
+                        break;
                     if(!PreviousTwoPower<1, 8>(tmp.GemmKPerBlock))
                         break;
                     if(!PreviousTwoPower<1, 4>(tmp.GemmKPack))
```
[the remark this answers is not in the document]
jane-zxy: yes, you are right, I'll fix it.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -111,7 +127,7 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
     auto get_euristic_config = [&](auto is_valid_func) {
         if(ctx.IsFp32())
         {
-            tmp = {256, 256, 8, 128, 128, 4, false, true};
+            tmp = {256, 256, 8, 128, 128, 4, false, true, 1};
```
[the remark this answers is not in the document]
jane-zxy: hi artem, 
I see there's some slice difference between fp32, fp16, bfp16:
1. initial value of tmp
2. value region of GemmKPack
2. value region of GemmBThreadDataPerRead_GemmN as well.
I don't know why, maybe for special consideration, @asroy 
since there're so many difference here, I would prefer to keep code as before.
3.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -111,7 +127,7 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
...
 
             bool all_visited = false;
```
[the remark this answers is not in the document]
jane-zxy: Hi artem,
yes, actually once is_valid_func() is true, we'll jump out of get_euristic_confg, and if 'all_visited=true', code is run into a dead loop, so I think there should be a 'break' here.
Detail pls refer to my new pr:

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
asroy: NextFlag<A, B> is same as NextTwoPower, it will loop over A -> B - > A...

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
asroy: line 239 and line 235 should not be the same,

line 235 is for 1st round check, and tmp should be IsReallyValid() && IsFast()

line 239 is for 2nd round check, if the 1st round fail to find a soluion. It should be IsReallyValid() only

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
asroy: > [Correctness] Range of `GemmKPack` for FP32 is `<1,4>`, but `<1,8>` for FP16/BF16. _**Should it be `<1,4>` for FP32 in `SetNextValue()` as well?**

Ideally yes, if SetNextValue takes ctx as argument. Currently, we cannot make the logic type dependent in SetNextValue().

We don't have to make it type-dependent, we just use the largest range for all data-type.

asroy:
@atamazov After double checking the comments, I think there is remaining issue we need to fix.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
atamazov: Ok.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
atamazov: I see, thanks.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
atamazov: This construction is ambiguous. `false` is the minimum and `false` is the maximum. How will the `NextFlag<min,max>()` function accept such a value? If the function would think that `false` on input is maximum, then it will increment it, and return an overflow indicator (`true`). If the function would think that `false` is minimum, it will increment it to the maximum, and does not report overflow to the caller (return `false`). To understand how all this works, one needs need to go into the implementation of `NextFlag()`.

However, this code is no different from a regular assignment. Moreover, there is no reason for doing this in in `SetNextValue()`. I would say this should be done exactly once, during initialization.

I already see that these constructs have proliferated in our code like rabbits. I think thast all of them must be eliminated.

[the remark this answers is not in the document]
jane-zxy: Hi atamazov,
Intentionally, NextFlag<> is similar as 'NextTwoPower', which will make a value to be set 'false' and 'true' while tunning. if it's NextFlag<false, false>, it will make a param untunable, but only set to 'false'. it's functionally same as just set its to 'false'.
However, I guess, original author, here, would try to make it tunable, and this it may be changed to tunable param someday how. 
@asroy , what do you think? leave it here or simply remove it from setNextValue?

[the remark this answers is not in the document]
atamazov: > if it's NextFlag<false, false>, it will make a param untunable, but only set to 'false'. 

Of course I know.

> it's functionally same as just set its to 'false'.

No, it is not the same because is also returns a wraparound indicator. The reader must read the definition of `NextFlag()` to find out which value the function would return in the `<false,false>` case.

> However, I guess, original author, here, would try to make it tunable, and this it may be changed to tunable param someday how.

This is over-design. 

There are an infinite number of "what if ..." cases in the potential future. The code that supports such cases has no immediate functional purpose. That confuses the reader who must guess why this was done. The more such code we have, the more difficult is to read and maintain the library. That as why we have to avoid this, or, in _exceptional_ cases, some explanatory comment must be added.

WRT this specific case, i.e. "idle" PerfConfig data members. Basically we should completely remove those. However, we can keep them, --- ___if it is highly likely___ that these will be needed in the future. This _exception_ is the simplest solution to make today's perf-db compatible with a future version of the code.

How this should be done?
- Put an explanatory comment near the definition of an "idle" member of PerformanceConfig.
- Initialize it but do not use in `Next...()`

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
atamazov: > I don't know why, 

But you should know because this is your PR. Please find out. When you figure it out, I will have no objection to leave the code as it is now (if you will want this).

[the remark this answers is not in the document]
jane-zxy: hi Artem,
thanks for your comments, I'll sync with chao to see if there's special consideration, if no, I would prefer to leave it here.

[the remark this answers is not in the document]
asroy: @atamazov FP32, FP16, BFp16 are not using exactly the same tuning parameter heuristic, merging them together would still require logic divergence. I prefer keeping them in separate blocks instead of mixing them.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
asroy: @jane-zxy You can follow @atamazov 's suggestion 1) remove this line from SetNextValue(), 2) comment in ```PerformanceImplicitGemmForwardV4R4Xdlops``` that this tuning parameter is a fixed value currently.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
atamazov: Ok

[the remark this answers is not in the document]
atamazov: @asroy 
> FP32, FP16, BFp16 are not using exactly the same tuning parameter heuristic

Note that it would be good for @jane-zxy to know the reasons of the above.

PR #348, opened 2020-07-15
Resnext101 tuning results 2.6

alexandraBara:
results for these networks:
Resnext101_32x4d_Pytorch_FP16_128
Resnext101_32x4d_Pytorch_FP16_256

on gfx908 HIP and removed invalidaded solvers from miopen.db and gfx908 HIP fdb


daniellowell:
Why is there only resnext here?

This is the set that should be tuned:
Resnet50_v1.5_Pytorch_FP16_256
Resnet50_v1.5_Pytorch_FP16_512
Resnext101_32x4d_Pytorch_FP16_128
Resnext101_32x4d_Pytorch_FP16_256

PR #349, opened 2020-07-15
Dynamic kernel: optensor

ce1adon:
OpTensor kernels are mainly used in convolutions with bias, RNN/LSTM/GRU paths

CC: @mwootton


src/kernels/MIOpenTensorKernels.cl
```diff
@@ -265,83 +206,33 @@ __kernel void OpTensorLeadingOnes(global MIOPEN_TYPE* a,
...
-        int o_h = (tid / c_w) % c_h;
-        int o_c = (tid / (c_w * c_h)) % c_c;
-        int o_n = tid / (c_w * c_h * c_c);
-
-        int index    = o_n * c_nstride + o_c * c_cstride + o_h * c_w + o_w;
-        c_off[index] = MIOPEN_TENSOR_OP(a_off[index] * alpha0, operand) + beta * c_off[index];
-    }
-#elif FIRST_NOT_ONE == 2 // bitmap = 1,1,1,0
-    int gid = get_group_id(0);
+    int gid = (bitmap == 15) ? get_global_id(0) : get_group_id(0);
```
TejashShah, line 209: You could use 0xF - just to hint it is bitmap rather than 15.

src/kernels/MIOpenTensorKernels.cl
```diff
@@ -395,102 +272,46 @@ __kernel void OpTensorLeadingOnesGeneric(global MIOPEN_TYPE* a,
...
-    {
-        int lid = get_local_id(0);
-        int o_c = gid % c_c;
-        int o_n = gid / c_c;
-
-        int bindex          = o_n * b_nstride + o_c * b_cstride;
+        int lid    = (bitmap == 15) ? 0 : get_local_id(0);
+        int lcl_sz = (bitmap == 15) ? work_per_wg : get_local_size(0);
+
+        int o_w = (bitmap & (1 << 0)) ? (gid % c_w) : 0;
```
TejashShah, line 285: You could replace 1 << 0  with just 0x1 - in case the compiler fails to optimize it.
You could replace 1 << 1  with just 0x2 - in case the compiler fails to optimize it. 

Personally, I find 0x1/0x2 easier to read than 1 << 0/ 1 << 1

src/kernels/MIOpenTensorKernels.cl
```diff
@@ -395,102 +272,46 @@ __kernel void OpTensorLeadingOnesGeneric(global MIOPEN_TYPE* a,
...
-    // num_wg: the number of workgroups should be launched
-    // MAX_NUM_WG: the maximum number of workgroups actually launched
-    for(; gid < num_wg; gid += MAX_NUM_WG)
-    {
-        int lid = get_local_id(0);
-        int o_c = gid % c_c;
-        int o_n = gid / c_c;
-
-        int bindex          = o_n * b_nstride + o_c * b_cstride;
+        int lid    = (bitmap == 15) ? 0 : get_local_id(0);
```
TejashShah, line 282: Since we are checking bitmap at so many different variable initializations, we could reduce it one if/else block. Essentially, reducing condition check you are making for each variable.

e.g. 
```
if(bimap == 0x1)
{
    ow = gid % c_w;
    oh = c_w;
}
else if(bimap == 0x1)
{
   ow =  ?
   oh = ?
}

```

src/ocl/tensorocl.cpp
```diff
@@ -174,8 +174,24 @@ void OpTensor3d(const Handle& handle,
...
-                     std::to_string(tensorOp);
+    network_config = std::to_string(bTensorDesc.GetType()) + "-" +
+                     std::to_string(aTensorDesc.GetType()) + "-" + std::to_string(tensorOp) + "-";
+
+    // for naive tensor ops
+    size_t RD_BLCK              = (clens[2] % 4 == 0) ? 4 : (clens[2] % 2 == 0) ? 2 : 1;
+    const std::string data_type = GetDataType(bTensorDesc.GetType());
+    const std::string READ_TYPE = (RD_BLCK == 1) ? data_type : data_type + std::to_string(RD_BLCK);
+
+    size_t total_work = std::max(size_t(clens[2] / RD_BLCK), size_t(1));
```
TejashShah, line 185: you don't need size_t cast operator for literal 1. Secondly, since we are using c++, for any casting operator us static_cast<size_t>(clen[2]/RD) rather than C-style.

src/ocl/tensorocl.cpp
```diff
@@ -210,18 +225,18 @@ void OpTensor3d(const Handle& handle,
                        long(Aoffset),
                        long(Boffset),
                        long(Coffset),
-                       int(clens[1]));
+                       long(total_work),
+                       long(total_work2),
+                       int(!float_equal(miopen_beta, 0.0)),
```
TejashShah, line 230: could we have used short in place of int to pass bool?

src/ocl/tensorocl.cpp
```diff
@@ -239,16 +254,20 @@ void OpTensor3d(const Handle& handle,
                        miopen_beta,
                        long(Aoffset),
                        long(Boffset),
-                       long(Coffset));
+                       long(Coffset),
+                       long(total_work),
+                       int(!float_equal(miopen_alpha0, 0.0)),
```
TejashShah, line 259: could we have used short in place of int to pass bool?

src/ocl/tensorocl.cpp
```diff
@@ -344,40 +340,17 @@ void OpTensor3d(const Handle& handle,
                 long(Aoffset),
                 long(Boffset),
                 long(Coffset),
-                int(clens[1]));
+                long(total_work),
+                long(total_work2),
+                int(!float_equal(miopen_beta, 0.0)),
```
TejashShah, line 345: could we have used short in place of int to pass bool?

src/kernels/MIOpenTensorKernels.cl
ce1adon: if/else block is what I try to avoid because it has negative impact on performance, while ternary operator has less impact.

src/ocl/tensorocl.cpp
ce1adon: Yes we can but it may not have much impact on performance, while int is easier to operate.

src/ocl/tensorocl.cpp
ce1adon: same as above

src/ocl/tensorocl.cpp
ce1adon: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/349#discussion_r456717071

ce1adon:
> Let's pass short in place of int for bool values.

https://github.com/ROCmSoftwarePlatform/MIOpen/pull/349#discussion_r456717071

src/kernels/MIOpenTensorKernels.cl
ce1adon: Will do

src/kernels/MIOpenTensorKernels.cl
ce1adon: I would prefer keep the current format as an indicator that this is a bitmap operation.

src/ocl/tensorocl.cpp
ce1adon: Will do

src/kernels/MIOpenTensorKernels.cl
TejashShah: It is a question of 1 if/else block vs quite a few ternary operators. In other words, consolidating a few conditional checks into a single one. I doubt this single #if else block will cause any thread divergence. Yes, it introduces conditional check but the same would be true for ternary operator.

TejashShah: Try this if it works on our compiler

```
A conditional of the form “if-then-else” generates branching. Use the select() function to replace these structures with conditional assignments that do not cause branching. For example:

if(x==1) r=0.5;
if(x==2) r=1.0;

becomes

r = select(r, 0.5, x==1);
r = select(r, 1.0, x==2);

Note that if the body of the if statement contains an I/O, the if statement cannot be eliminated.
```

https://rocmdocs.amd.com/en/latest/Programming_Guides/Opencl-optimization.html?highlight=%27bool%27

src/ocl/tensorocl.cpp
TejashShah: Ok, let's keep it int.

daniellowell:
@ce1adon Can you merge develop into your branch once again. There was another workaround for the COMgr phase.

src/kernels/MIOpenTensorKernels.cl
ce1adon: Some of the condition checks here are a little more complicated than simple if-then-else check. It may end up with several select function embedded in each other. The ternary should be as less expensive as the select function.

TejashShah: Ok. I think we can improve on using ternary operator these many times. I would take your word for granted that either select or a single if/else wouldn't have made performance better.

PR #358, opened 2020-07-23
[WIP] MP_bidirect_winograd

shurale-nkn:


src/buffer_info.cpp
```diff
@@ -75,98 +75,56 @@ MemLayout_t GetGroupConvLayout(MemLayout_t layout, bool IsDataBuffer)
 }
 
 BuffInfo::BuffInfo(
-    MemLayout_t layout, int nk, int c, int h, int w, int vec_c, int _g, int data_len_t)
+    MemLayout_t layout, int nk, int c, int h, int w, int vec_c, int _g, int _data_len_t)
```
Kirpich30000, line 78: It should be element size, not data len.

src/kernels/xform_bidirect_winograd_data.s
```diff
@@ -0,0 +1,32 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+xform_data=1
+xform_filter=0
+xform_output=0
+xform_mirror=0
```
Kirpich30000, line 29: It's not a good practice to have implicitly different interface to same type of entities. The `xform_mirror` by it's nature is a kernel argument (compile time). However for different transforms it is set in different places. The solver host code and .s files are different for filter transform and other transforms for no reason.

src/kernels/xform_bidirect_winograd_out.s
```diff
@@ -0,0 +1,31 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+xform_data=0
+xform_filter=0
+xform_output=1
+xform_mirror=0
```
Kirpich30000, line 29: see prev comment

src/mlo_dir_conv.cpp
```diff
@@ -149,7 +149,12 @@ static auto GetWindogradSolvers()
     return miopen::solver::SolverContainer<miopen::solver::ConvBinWinograd3x3U,
                                            miopen::solver::ConvBinWinogradRxSf3x2,
                                            miopen::solver::ConvBinWinogradRxSf2x3,
-                                           miopen::solver::ConvBinWinogradRxS>{};
+                                           miopen::solver::ConvBinWinogradRxS,
+                                           miopen::solver::ConvMPBidirectWinograd<2, 3>,
```
Kirpich30000, line 153: Did you see any case where F(2,3) shows better performance?

src/convolution.cpp
```diff
@@ -422,10 +422,12 @@ std::size_t ConvolutionDescriptor::ForwardGetWorkSpaceSize(Handle& handle,
...
     ctx.DetectRocm();
 
+    const size_t workspace_size_winograd = ForwardBackwardDataGetWorkSpaceSizeWinograd(ctx);
+
     if(IsWinograd3x3SupportedAndFast(ctx))
     {
-        MIOPEN_LOG_I2(0);
-        return 0;
+        MIOPEN_LOG_I2(workspace_size_winograd);
+        return workspace_size_winograd;
```
Kirpich30000, line 430: 3x3 Winograd didn't need any workspace, is it?

src/convolution.cpp
```diff
@@ -739,6 +753,32 @@ std::size_t ConvolutionDescriptor::ForwardBackwardDataGetWorkSpaceSizeDirect(
     }
 }
 
+std::size_t ConvolutionDescriptor::ForwardBackwardDataGetWorkSpaceSizeWinograd(
+    const miopen::ConvolutionContext& ctx) const
+{
+    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_WINOGRAD{}))
+        return 0;
+
+    try
```
Kirpich30000, line 762: Why should something like FindAllSolutions or GetWorkspaceSize throw any exception?

src/include/miopen/solver.hpp
```diff
@@ -1538,6 +1538,42 @@ struct ConvBinWinogradRxSFused : SolverBase<ConvolutionContext>
...
+    }
+
+    static std::string GetSolverKernelNames(int id)
+    {
+        static const std::string name_suffix =
+            '_' + std::to_string(WinoDataH) + '_' + std::to_string(WinoDataW) + '_' +
+            std::to_string(WinoFilterH) + '_' + std::to_string(WinoFilterW);
+        static const std::string names[3] = {
+            "miopenGcnAsmMPBidirectWinogradXformData" + name_suffix,
+            "miopenGcnAsmMPBidirectWinogradXformFilter" + name_suffix,
```
Kirpich30000, line 1564: Will the fwd and bwd filter transforms have the same name? Is it ok to have the same kernel name for different kernel binaries? Will the library work properly in that case?

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -0,0 +1,536 @@
...
+    if(!params.rmv.IsV3())
+        return false;
+    if(!params.Is2d())
+        return false;
+    if(params.direction.IsBackwardWrW())
+        return false;
+    if(!params.IsFp32())
+        return false;
+
+    if(!(StartsWith(name, "gfx8") || StartsWith(name, "gfx9")))
```
Kirpich30000, line 165: gfx8 was never tested with this kernel. It shouldn't be functional due to a gfx9 changes and I'm not sure it will ever assemble. Please do at least sanity checks with available hardware to check asic support.

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -0,0 +1,536 @@
...
+        && params.kernel_stride_h == params.kernel_stride_w
+        && params.kernel_dilation_w == 1
+        && params.kernel_dilation_h == 1
+        && params.batch_sz < std::pow(2, 16)
+        && params.n_inputs < std::pow(2, 16)
+        && params.n_outputs < std::pow(2, 16)
+        && params.out_height < std::pow(2, 16)
+        && params.out_width < std::pow(2, 16)
+        && params.bias == 0
+        && params.in_layout == "NCHW");
```
Kirpich30000, line 246: Are there checks for input dilation and output stride?

src/convolution.cpp
shurale-nkn: Yes, but you can guarantee that it's faster than MultiPass?

src/include/miopen/buffer_info.hpp
```diff
@@ -50,27 +51,117 @@ MemLayout_t GetGroupConvLayout(MemLayout_t layout, bool IsDataBuffer);
...
+}
+
+template <LPart_t... others>
+inline void FillLayoutStride(BuffInfo* b)
+{
+    FillNextLayoutStride<others..., LPart_t::LPart_begin>(b, 1);
+}
+
+} // namespace LayoutConstructor
+
```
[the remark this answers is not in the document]
shurale-nkn: please explain how it will look.

src/include/miopen/solver.hpp
shurale-nkn: Yes, same.
Yes, as with most tunable solvers.
Yes, it works fine.

src/include/miopen/buffer_info.hpp
```diff
@@ -50,27 +51,117 @@ MemLayout_t GetGroupConvLayout(MemLayout_t layout, bool IsDataBuffer);
...
 };
+
+enum class ConvWinoXformType
+{
+    // N_G_C_H_W,
+    N_1_CThTw_Xh_Xw,
+    N_GXhXw_C_Th_Tw
+};
+
 template <int WinoDataH, int WinoFilterH, int WinoDataW = WinoDataH, int WinoFilterW = WinoFilterH>
```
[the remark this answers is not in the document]
shurale-nkn: What?

src/include/miopen/buffer_info.hpp
```diff
@@ -50,27 +51,117 @@ MemLayout_t GetGroupConvLayout(MemLayout_t layout, bool IsDataBuffer);
 MemLayout_t GetMemLayout_t(const std::string& s);
 MemLayout_t GetSwappedNCLayout(MemLayout_t layout);
 
+// Parts of MemLayout
+enum class LPart_t
```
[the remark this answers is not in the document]
shurale-nkn: This enumeration describes the relationship between a dimension name and a numeric notation.
Part_begin, Part_end technical elements that limit the range of values used. It is necessary to expand this list and create a similar list for solutions based on Winograd algo.

src/buffer_info.cpp
```diff
@@ -75,98 +100,56 @@ MemLayout_t GetGroupConvLayout(MemLayout_t layout, bool IsDataBuffer)
 }
 
 BuffInfo::BuffInfo(
-    MemLayout_t layout, int nk, int c, int h, int w, int vec_c, int _g, int data_len_t)
+    MemLayout_t layout, int nk, int c, int h, int w, int vec_c, int _g, int _element_size)
 {
     if(!(vec_c != 0))
```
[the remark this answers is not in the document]
shurale-nkn: outdated and not used for a long time. Will be removed

src/buffer_info.cpp
```diff
@@ -75,98 +100,56 @@ MemLayout_t GetGroupConvLayout(MemLayout_t layout, bool IsDataBuffer)
 }
 
 BuffInfo::BuffInfo(
-    MemLayout_t layout, int nk, int c, int h, int w, int vec_c, int _g, int data_len_t)
+    MemLayout_t layout, int nk, int c, int h, int w, int vec_c, int _g, int _element_size)
```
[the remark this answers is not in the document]
shurale-nkn: outdated and not used for a long time. Will be removed

src/convolution.cpp
shurale-nkn: For uniformity of all GetWorkspaceSizes. 
In the future all functions must be merged into single one

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: && (params.kernel_stride_w == 1)
        && params.kernel_stride_h == params.kernel_stride_w
        && params.kernel_dilation_w == 1
        && params.kernel_dilation_h == 1

src/include/miopen/buffer_info.hpp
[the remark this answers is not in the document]
Kirpich30000:
```
FillStrides(int v3, int v2, int v1, size_t &stride3, size_t &stride2, size_t &stride1)
{
   stride1 = 1;
   stride2 = v1;
   stride3 = v2 * stride2;
   ...
}
```

Alternatively:
```
struct dim
{
   size_t size;
   size_t stride;
   size_t byte_stride;
}

FillStrides(dim &v3, dim &v2, dim &v1, size_t elem_size)
{
   v1.stride = 1;
   v2.stride = v1.size * v1.stride;
   v3.stride = v2.size * v2.stride;
   ...
}

src/convolution.cpp
Kirpich30000: Still, why should GetWorkspaceSize ever throw?

src/mlo_dir_conv.cpp
shurale-nkn: So, you want to save only  F(5,3) and F(6,3)?

src/include/miopen/buffer_info.hpp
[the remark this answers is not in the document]
shurale-nkn: about complexity look on this
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/conv/problem_description.hpp#L82

src/convolution.cpp
shurale-nkn: Some of them may throw. Some GEMMs have definitely done this before
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/6549d8e36aff494317cf3d4bc19597fb094c54d2/src/reducetensor.cpp#L286-L294

atamazov:
@shurale-nkn @Kirpich30000 Please assign priority label to this PR, thanks.

src/convolution.cpp
atamazov: @Kirpich30000 We know that when when 3x3 winograd is really fast, it is faster than all other algorithms. Therefore all other algorithms are skipped. However we are still trying _all_ applicable Winograds, and the one that needs workspace may win.

atamazov:
@Kirpich30000 Are all of your comments resolved?

src/convolution.cpp
atamazov: > Please add an issue about removing exceptions here. 

Yep, it looks like we do not trust in our own code ;) Perhaps we shall keep try/catch for Release builds, but allow the library to fail in Debug builds. This is common pattern that worth dedicated ticket.

> If some checks like inDescLengths.size() != outDescLengths.size() are needed it should be done in appropriate place and GetWorkspaceSize() doesn't look like appropriate place. 

I am second to this. Strange check, really ;)

atamazov: @Kirpich30000 After second thought:

> If some checks like inDescLengths.size() != outDescLengths.size() are needed it should be done in appropriate place and GetWorkspaceSize() doesn't look like appropriate place.

I am second to this but this is not relevant to this PR. `ForwardBackwardDataGetWorkSpaceSizeWinograd()` does not have inappropriate checks.

> Please add an issue about removing exceptions here.

When a solver is unable to solve a problem, it's `IsApplicable()` must return `false`. And `IsApplicable()` must not throw. Exception there means programming error that needs to be fixed. And try/catch here looks like we are hiding these errors.

However this is not exactly so. First of all, it is theoretically impossible to fully test the library. This means that we can deliver a library that contains elusive programming errors, including exceptions in `IsApplicable()`. The `try/catch` blocks in GWSS functions (a common pattern in the library) are intended to improve stability of the library in the event of such errors.

Second, we are not hiding errors, but lowering the level. Instead of termination, the library logs an Error onto console for Debug builds, and Warning for Releases.

If you insist, the only thing I have in mind is removing `try/catch` for Debug builds.

src/solver/conv_MP_bidirectional_winograd.cpp
atamazov: +1 We should't add new kernels for gfx8. Please see https://github.com/AMDComputeLibraries/MLOpen/issues/1934 for more info.

atamazov (commented):
I highly recommend removal of gfx8 support in this solver. This can be done in separate PR.

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: @atamazov gfx8 removed 13 days ago

shurale-nkn:
> I highly recommend removal of gfx8 support in this solver. This can be done in separate PR.

gfx8 support removed in commit c6c5ea4

atamazov:
Oops

PR #361, opened 2020-07-23
Embed Find DB 

JehandadKhan:
**This PR depends on #313 and should not be merged before it.**

This PR embeds all `*.fdb.txt` files as statically linked code objects inside MIOpen. This avoids installing files in `<install prefix>/share` and `<install prefix>/miopen/share` directories. 

The CMake module responsible for linking data files creates a map (`miopen_data`) which holds the start and end pointers to multiple files and can be read from at run time. At runtime MIOpen reads the location of the files from the above mentioned map, creates a `std::stringstream` from it so that the `readonlyramdb` code can prefetch it like it would any other file.


JehandadKhan:
@daniellowell  Description updated.

cmake/embed.cmake
```diff
@@ -0,0 +1,98 @@
+
```
daniellowell, line 1: Needs license

cmake/embed.cmake
```diff
@@ -0,0 +1,98 @@
...
+    list(LENGTH PARSE_OBJECTS OBJECTS_LEN)
+    if(NOT ${SYMBOLS_LEN} EQUAL ${OBJECTS_LEN})
+        message(FATAL_ERROR "Symbols and objects dont match: ${SYMBOLS_LEN} != ${OBJECTS_LEN}")
+    endif()
+    math(EXPR LEN "${SYMBOLS_LEN} - 1")
+
+    foreach(idx RANGE ${LEN})
+        list(GET PARSE_SYMBOLS ${idx} SYMBOL)
+        list(GET PARSE_OBJECTS ${idx} OBJECT)
+    # foreach(SYMBOL ${PARSE_SYMBOLS})
```
daniellowell, line 26: Delete comment.

src/CMakeLists.txt
```diff
@@ -632,13 +634,24 @@ rocm_export_targets(
...
-file(GLOB FIND_DB_FILES kernels/*.fdb.txt)
-install(FILES
-    kernels/miopen.db
-    ${FIND_DB_FILES}
- DESTINATION ${DATA_INSTALL_DIR}/db)
-
+if(NOT MIOPEN_EMBED_DB STREQUAL "")
+    include(embed)
+    set(CODE_OBJECTS)
+    list(APPEND CODE_OBJECTS "kernels/miopen.db")
```
daniellowell, line 641: Why is this line here?

src/readonlyramdb.cpp
```diff
@@ -72,44 +82,71 @@ static auto Measure(const std::string& funcName, TFunc&& func)
...
 
-            cache.emplace(key, CacheItem{n_line, contents});
+        constexpr bool isEmbedded = MIOPEN_EMBED_DB;
+        if(!testing_find_db_path_override() && isEmbedded)
+        {
+#if MIOPEN_EMBED_DB
+            boost::filesystem::path filepath(path);
+            const auto& it_p = miopen_data().find(filepath.filename().string() + ".o");
+            if(it_p == miopen_data().end())
+                MIOPEN_THROW(miopenStatusInternalError,
```
DrizztDoUrden, line 134: [Suggestion] I think we may want to log an error and just return without preloading anything because it is not a fatal error. But it may as well be left as is because it should not be allowed to happen in production anyway.

JehandadKhan: The later was my reasoning for making it fatal, since MIOpen is compiled specifically to embed the data for a particular architecture, if its not there, something is certainly wrong.

src/sqlite/memvfs.cpp
```diff
@@ -0,0 +1,557 @@
...
+    sqlite3_int64 sz;     /* Size of the file */
+    sqlite3_int64 szMax;  /* Space allocated to aData */
+    unsigned char* aData; /* content of the file */
+    int bFreeOnClose;     /* Invoke sqlite3_free() on aData at close */
+};
+
+/*
+** Methods for MemFile
+*/
+static int memClose(sqlite3_file*);
```
pfultz2, line 72: Will these symbols interfere if another application statically links in a memvfs file? Maybe we could namespace these functions or `#define` them to a different name for compilation. Even though its declared `static` it has `extern C` so it might make the symbol available across other TUs.

src/sqlite/memvfs.cpp
```diff
@@ -0,0 +1,557 @@
...
+#endif /* MEMVFS_TEST */
+
+#ifdef _WIN32
+__declspec(dllexport)
+#endif
+    /*
+    ** This routine is called when the extension is loaded.
+    ** Register the new VFS.
+    */
+    int sqlite3_memvfs_init(sqlite3* db, char** pzErrMsg, const sqlite3_api_routines* pApi)
```
pfultz2, line 535: Maybe we only need to change the name of this function?

cmake/embed.cmake
JehandadKhan: Fixed

cmake/embed.cmake
JehandadKhan: Fixed

src/CMakeLists.txt
JehandadKhan: Because the `MIOPEN_EMBED_DB` flag embeds both perf db and find db. This is the perf db part.

src/sqlite/memvfs.cpp
JehandadKhan: I have changed the name of the method below.

PR #305, opened 2020-06-22
Refactor igemm bwd xdlops v1r1

TejashShah:
Supports all datatypes in a single solver and supports all datatypes and group/non-group in a single kernel algorithm.

The full testing is underway. The initial testing showed verification passed

PR #362, opened 2020-07-27
Fix nan issue in bwd v1r1 kernel

TejashShah:
@asroy If we skip settensor, it causes nan in resnext50 as reported by @sunway513 . So, putting it back.

@asroy The assumption we had here in #305 https://github.com/ROCmSoftwarePlatform/MIOpen/pull/305#discussion_r455157461 seemed wrong, causing this nan.

### DIFF
--- a/src/conv/invokers/impl_gemm.cpp
+++ b/src/conv/invokers/impl_gemm.cpp
@@ -60,7 +60,7 @@ InvokerFactory MakeImplGemmDataInvokerFactory(const ConvolutionContext& ctx)
                     bool need_atomic_add = (stride_h < dilation_h * (y - 1) + 1) &&
                                            (stride_w < dilation_w * (x - 1) + 1);
                     bool every_pixel_is_written =
-                        (dilation_h == 1 && stride_h <= Y) && (dilation_w == 1 && stride_w <= X);
+                        (dilation_h == 1 && stride_h <= y) && (dilation_w == 1 && stride_w <= x);
                     bool need_set_zero = need_atomic_add || !(every_pixel_is_written);
                     bool need_cast     = need_atomic_add;
 
--- a/src/conv/invokers/impl_gemm.cpp
+++ b/src/conv/invokers/impl_gemm.cpp
@@ -57,7 +57,7 @@ InvokerFactory MakeImplGemmDataInvokerFactory(const ConvolutionContext& ctx)
                     const auto dilation_h = conv.GetConvDilations()[0];
                     const auto dilation_w = conv.GetConvDilations()[1];
 
-                    bool need_atomic_add = (stride_h < dilation_h * (y - 1) + 1) &&
+                    bool need_atomic_add = (stride_h < dilation_h * (y - 1) + 1) ||
                                            (stride_w < dilation_w * (x - 1) + 1);
                     bool every_pixel_is_written =
                         (dilation_h == 1 && stride_h <= y) && (dilation_w == 1 && stride_w <= x);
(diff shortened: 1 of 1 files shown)


sunway513:
Confirmed this PR can fix the Nan issue observed with TF Resnet50 fp16.

TejashShah:
> 
> 
> @TejashShah This change will call SetTensor (0) even if for 1x1 stride1 case. This is not what you would want.
> 
> Consider refactor the logic like this:
> 
> ```
> bool need_atomic_add = (stride_h < dilation_h * (y - 1) + 1) && (stride_w < dilation_w * (x - 1) + 1);
> bool every_pixel_is_written = (dilation_h == 1 && stride_h <= Y) && (dilation_w == 1 && stride_w <= X);
> bool need_set_zero = need_atomic_add || !(every_pixel_is_written);
> bool need_cast = need_atomic_add
> ```

We need SetTensor otherwise, it causes NaN

TejashShah:
If you see code below, v4r1 has applied the similar logic of determining when to invoke cast tensor or not.

asroy:
> We need SetTensor otherwise, it causes NaN

Do you need SetTensor for 1x1 stride=1?

asroy:
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/362#pullrequestreview-455995248

Please see this code if it makes sense or not

> If you see code below, v4r1 has applied the similar logic of determining when to invoke cast tensor or not.

I understand, I asked Kevin to implement like that to avoid unnecessary call of SetTensor

daniellowell:
This PR keeps failing in the same place. @TejashShah please take a look.

daniellowell:
The bug in this PR is reproducible locally:
```
MIOpen(HIP): Info2 [LoadBinary] Loading binary for: MIOpenSubTensorOpWithScalarKernel.cl ;args:  -DSUBTENSOR_OP_WITH_SCALAR=SUBTENSOR_OP_WITH_SCALAR_SET -DMIOPEN_USE_FP16=0 -DMIOPEN_USE_FP32=0 -DMIOPEN_USE_INT8=0 -DMIOPEN_USE_INT8x4=0 -DMIOPEN_USE_BFP16=1 -DMIOPEN_USE_INT32=0 -DMIOPEN_USE_RNE_BFLOAT16=1 -DWORK_LENGTH_0=65536 -mcpu=gfx908
MIOpen(HIP): Info2 [Prepare] SELECT kernel_blob, kernel_hash, uncompressed_size FROM kern_db WHERE (kernel_name = 'MIOpenSubTensorOpWithScalarKernel.cl.o') AND (kernel_args = ' -DSUBTENSOR_OP_WITH_SCALAR=SUBTENSOR_OP_WITH_SCALAR_SET -DMIOPEN_USE_FP16=0 -DMIOPEN_USE_FP32=0 -DMIOPEN_USE_INT8=0 -DMIOPEN_USE_INT8x4=0 -DMIOPEN_USE_BFP16=1 -DMIOPEN_USE_INT32=0 -DMIOPEN_USE_RNE_BFLOAT16=1 -DWORK_LENGTH_0=65536 -mcpu=gfx908');
MIOpen(HIP): Info2 [Measure] Db::FindRecord time: 1.35474 ms
Memory access fault by GPU node-5 (Agent handle: 0x26c1900) on address 0x7efc1fe07000. Reason: Page not present or supervisor privilege.
Aborted (core dumped)
root@6845bbe33ea6:/data/MIOpen/hcbuild#
```

TejashShah:
@daniellowell fp16/bfp16 bug is resolved. fp32 verification failure needs to be tracked down.

daniellowell:
> @daniellowell fp16/bfp16 bug is resolved. fp32 verification failure needs to be tracked down.

I expect you'll open an additional PR for the fp32 issue.

daniellowell:
@asroy @zjing14 Can you you re-review this PR?

Issue #365, opened 2020-07-30
3D pooling regression in TF unit tests with MIOpen from ROCm 3.7RC1


deven-amd:
Some of the TF unit tests are failing with the following error
```
MIOpen Error: /root/driver/MLOpen/src/ocl/pooling_ocl.cpp:142: 3D pooling doesn't support workspace index mask mode
2020-07-30 17:06:17.649098: E tensorflow/stream_executor/rocm/rocm_dnn.cc:4530] failed to enqueue forward pooling (before backward) on stream: miopenStatusUnknownError
```

The same tests work fine with older MIOpen versions (from ROCm 3.5). 

output with miopen logging enabled seems to point to
```
...
MIOpen(HIP): miopenStatus_t miopenPoolingForward(miopenHandle_t, const miopenPoolingDescriptor_t, const void *, const miopenTensorDescriptor_t, const void *, const void *, const miopenTensorDescriptor_t, void *, bool, void *, size_t){
MIOpen(HIP): 	handle = stream: 0x3de39c0, device_id: 0
MIOpen(HIP): 	poolDesc = miopenPoolingMax, 1, 1, 1, 0, 0, 0, 1, 1, 1, 
MIOpen(HIP): 	alpha = 0x7fbb33ffd160
MIOpen(HIP): 	xDesc = 1, 1, 3, 5, 4
MIOpen(HIP): 	x = 0x7fbad5a02d00
MIOpen(HIP): 	beta = 0x7fbb33ffd164
MIOpen(HIP): 	yDesc = 1, 1, 3, 5, 4
MIOpen(HIP): 	y = 0x7fbad5a03200
MIOpen(HIP): 	do_backward = 1
MIOpen(HIP): 	workSpace = 0x7fbad5a03100
MIOpen(HIP): 	workSpaceSize = 240
MIOpen(HIP): }
MIOpen(HIP): Command [Pooling_logging_cmd] ./bin/MIOpenDriver pool -d 3 -M 0 -n 1 -c 1 -D 3 -H 5 -W 4 -Z 1 -y 1 -x 1 -o 0 -p 0 -q 0 -s 1 -v 1 -u 1 -m max -F 1 -t 1
MIOpen Error: /root/driver/MLOpen/src/ocl/pooling_ocl.cpp:142: 3D pooling doesn't support workspace index mask mode
[...]

PR #366, opened 2020-07-30
set default workspaceIndexMode to image index for 3D pooling

ce1adon:
Fix for issue #365


daniellowell:
@ce1adon This is straight forward. Questions: Is this a breaking change for MIOpen? What was the previous behavior and what is the impact to PyTorch and Caffe2?

Issue #332, opened 2020-07-08
Dynamic Implicit GEMM should not have Performance Configs or RunAndMeasure methods.


JehandadKhan:
The Dynamic Implicit GEMM solvers are not tunable, but still have PerformanceConfig classes as well as RunAndMeasure Methods. 

This can result in unnecessary cycles spent in tuning.

atamazov:
Wait, it looks like this is tunable. Its `GetSolution()` needs `PerformanceConfig` on input: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/38d18b7716a96d229373747eeda4c1a864a349e3/src/solver/conv_asm_implicit_gemm_v4r1_dynamic.cpp#L565-L575

atamazov:
There are code quality issues (like unused boolean parameter of `GetSolution`, useless CO V3 related things, and ridiculous format of serialized `PerformanceConfig`), but I do not see why this is not tunable.

carlushuang:
Hi @atamazov , ultimately dynamic kernel should not need tunable. When I implement this solver I just keep it compatible with the non-dynamic version of HIP igemm, and I think I need to re-factor this to avoid confusing.

JehandadKhan:
Me and @carlushuang has a chat about this and he would factor this out in a future PR once he is done with other higher priority work.

atamazov:
OMG indeed, `GetSolutionBase()` does not use `PerformanceConfig` to configure kernel. How can we write this :facepalm: ...

carlushuang:
@atamazov sorry for the confusing, let me take the refactor into work.

aserio:
@carlushuang Is this issue related to an open pull request?

carlushuang:
> @carlushuang Is this issue related to an open pull request?

Hi, currently no PR is dealing with this issue. But there is another PR:https://github.com/ROCmSoftwarePlatform/MIOpen/pull/272 that changes several related invokers code. I prefer that PR merged then create another PR to deal with this issue, in case of code conflict :)

aserio:
@carlushuang, As PR#272 is now merged have you returned to this issue?

carlushuang:
@aserio Hi, yes let me take some time to fix this issue.

carlushuang:
@aserio Please check https://github.com/ROCmSoftwarePlatform/MIOpen/pull/368

aserio:
@carlushuang, Thanks for the update and your quick work!

PR #368, opened 2020-08-06
refactor fwd v4r1 dynamic solver to be non-tunable.

carlushuang:
fix issue https://github.com/ROCmSoftwarePlatform/MIOpen/issues/332, re-factor fwd v4r1 dynamic to be non-tunable, by remove PerformanceConfig classes and RunAndMeasure method


daniellowell:
@carlushuang is this a bug?

carlushuang:
> @carlushuang is this a bug?

Yes, the reason is that dynamic kernel should not be tunable (no need RunAndMeasure, and perfConfig), just heuristic is enough as described in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/332. So current solver of fwd dynamic need change in this PR

aserio:
@JehandadKhan or @atamazov  would either of you be able to review this issue? This is a fix for a high priority issue.

src/solver/conv_asm_implicit_gemm_v4r1_dynamic.cpp
```diff
@@ -573,16 +369,28 @@ static inline ConvSolution GetSolutionBase(const ConvolutionContext& ctx,
...
-    const ConvolutionContext& ctx, const PerformanceImplicitGemmV4R1Dynamic& config, bool) const
+ConvSolution ConvAsmImplicitGemmV4R1DynamicFwd::GetSolution(const ConvolutionContext& ctx) const
 {
-    return GetSolutionBase(ctx, config, AsmImplicitGemmV4R1);
+    auto tunables = GetImplicitGemmV4R1DynamicTunables();
+    auto it       = std::find_if(
+        tunables.begin(), tunables.end(), [&](auto tunable) { return tunable.IsValid(ctx); });
+
+    if(it == tunables.end())
+        MIOPEN_THROW("should not happen!");
```
JehandadKhan, line 379: Please add a descriptive error message.

src/solver/conv_asm_implicit_gemm_v4r1_dynamic.cpp
```diff
@@ -573,16 +369,28 @@ static inline ConvSolution GetSolutionBase(const ConvolutionContext& ctx,
...
-    const ConvolutionContext& ctx, const PerformanceImplicitGemmV4R1Dynamic& config, bool) const
+ConvSolution ConvAsmImplicitGemmV4R1DynamicFwd_1x1::GetSolution(const ConvolutionContext& ctx) const
 {
-    return GetSolutionBase(ctx, config, AsmImplicitGemmV4R1_1x1);
+    auto tunables = GetImplicitGemmV4R1DynamicTunables();
+    auto it       = std::find_if(
+        tunables.begin(), tunables.end(), [&](auto tunable) { return tunable.IsValid(ctx); });
+
+    if(it == tunables.end())
+        MIOPEN_THROW("should not happen!");
```
JehandadKhan, line 391: again

src/solver/conv_asm_implicit_gemm_v4r1_dynamic.cpp
carlushuang: OK, let me fix it

src/solver/conv_asm_implicit_gemm_v4r1_dynamic.cpp
carlushuang: OK, let me fix it

JehandadKhan: Please, specify `miopenStatusInternalError` as the error since this is something the user should not observe in the first place.

carlushuang: OK, added

src/solver/conv_asm_implicit_gemm_v4r1_dynamic.cpp
```diff
@@ -33,11 +33,40 @@
...
+        int, int, int, int, int, int, int, int, int, int, int, int, int, int, int, int);
+    bool IsValid(const ConvolutionContext& ctx) const;
+};
+
+static inline const std::vector<TunableImplicitGemmV4R1Dynamic>&
 GetImplicitGemmV4R1DynamicTunables()
 {
     // clang-format off
-    static const std::vector<PerformanceImplicitGemmV4R1Dynamic> tunables = {
+    static const std::vector<TunableImplicitGemmV4R1Dynamic> tunables = {
```
atamazov, line 69: What "tunable" means here? AFAICS each vector element (an instance of `TunableImplicitGemmV4R1Dynamic`) represents a name of some kernel. Perhaps these kernels are "pre-configured" versions of some tunable kernel, but neither of these kernels are tunable. I highly recommend using appropriate words in identifiers.

atamazov, line 69: Please add a comment saying that the order of elements is important for performance.

atamazov:
As far as I see, the Solver comprises a family of 14 kernels. Depending on the problem, one of these is selected for use in Solution. In other words, here we have not just one fully "dynamic" kernel, but 14 different ones.

What guarantees do we have that all these 14 kernels will get inside the pre-compiled kernel cache?

daniellowell:
>What guarantees do we have that all these 14 kernels will get inside the pre-compiled kernel cache?
Because these are ASM kernels we should not have them in the precompiled kernels package. I would like only those kernels that actually require compilation to exist there.

atamazov:
> Because these are ASM kernels we should not have them in the precompiled kernels package.

Note that there is no requirement that ASM kernels should not be put there.

> I would like only those kernels that actually require compilation to exist there.

This is weird, but acceptable, if the requirement for ASM kernels is like:

_There is no requirement to add all of ASM kernels into precompiled cache. Likewise, there is no requirement that ASM kernels shouldn't be added. In other words, an implementation is free to choose which ASM kernels should be added._

Issue #381, opened 2020-08-14
ResNext3D layer selected incorrectly large WSS causing OOM failure


daniellowell:
ROCm 3.8 Blocking issue.

JIRA tracking issue for ResNext3D:
http://ontrack-internal.amd.com/browse/SWDEV-246350

Issue is in default hybrid mode Caffe2 model is crashing with OOM. However, in normal find mode it passes. 

**Analysis**:
- Deleting the user Find-Db and running in default hybrid mode also fails with OOM.

- MIOPEN_LOG_LEVEL=6 for default hybrid mode shows 1.3GB is requested by layer:
```
MIOpen(HIP): Info2 [FindRecordUnsafe] Looking for key 256-1-14-14-3x3x3-256-1-14-14-4-1x1x1-1x1x1-1x1x1-0-NCHW-FP32-B_g256 in file /root/.config/miopen//gfx90878.HIP.2_6_0_8145-rocm-rel-3.7-19-c16087a4.ufdb.txt
MIOpen(HIP): Info2 [FindRecordUnsafe] Key match: 256-1-14-14-3x3x3-256-1-14-14-4-1x1x1-1x1x1-1x1x1-0-NCHW-FP32-B_g256
MIOpen(HIP): Info2 [FindRecordUnsafe] Contents found: miopenConvolutionBwdDataAlgoGEMM:gemm,0.20544,1387266048,rocBlas,<unused>
MIOpen(HIP): Info2 [Measure] Db::FindRecord time: 0.194715 ms
MIOpen(HIP): Info2 [BackwardDataGetWorkSpaceSize] 1387266048
```

- Normal mode is actually requesting 5MB:
```
MIOpen(HIP): Info2 [BackwardDataGetWorkSpaceSize] 5419008
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] requestAlgoCount = 1, workspace = 5419008
MIOpen(HIP): Info2 [FindRecordUnsafe] Looking for key 256-1-14-14-3x3x3-256-1-14-14-4-1x1x1-1x1x1-1x1x1-0-NCHW-FP32-B_g256 in file /root/.config/miopen//gfx90878.HIP.2_6_0_8145-rocm-rel-3.7-19-c16087a4.ufdb.txt
```

- Find-Db is populated after normal run with clean ufdb with ~the correct 5MB size~ the incorrect size (1.3GB).

- Driver command to reproduce the issue:
```
[...]

PR #382, opened 2020-08-15
[Fix] workspace size calc of gemm for group_conv

zjing14:
- Fixed a bug in GetWorkspaceSize that multiples group_conv twice inside and outside of BackwardDataGetWorkSpaceSizeGEMM (issue #381)


daniellowell:
@cderb We will have to rerun all group > 1 GEMM algo convolutions through find-db before we publish the find-db in a PR. Should be a fast run. 

You need to **cherry-pick** this PR into your tuning branch and rerun find-db for the above.

zjing14:
Just backwardData should be enough.

atamazov:
> Just backwardData should be enough.

And Forward Transposed.

PR #384, opened 2020-08-17
Force installation of SysDb when embedding Dbs

JehandadKhan:
* Address a bug in `src/CMakeLists.txt` where static linking would fail while embedding dbs
* Force installation of sys-db when embedding Dbs ( currently sys db installation may disable the embedding feature)


daniellowell:
How was this tested?

CMakeLists.txt
```diff
@@ -115,12 +115,16 @@ set(MIOPEN_INSTALL_CXX_HEADERS Off CACHE BOOL "Install MIOpen's C++ header inter
 
 
 # Embedded Build Configuration
-set(MIOPEN_EMBED_DB "" CACHE STRING "Semi-colon separated list of of architecture CU pairs to embed on-disk DBs in the binary. Example gfx906_60:gfx900_56")
+set(MIOPEN_EMBED_DB "" CACHE STRING "Semi-colon separated list of architecture CU pairs to embed on-disk DBs in the binary. Example gfx906_60;gfx900_56")
+option(MIOPEN_DISABLE_SYSDB  "Disable sys database access" ${MIOPEN_EMBED_BUILD})
+if(NOT MIOPEN_EMBED_DB STREQUAL "")
+    set(MIOPEN_DISABLE_SYSDB Off CACHE BOOL "" FORCE)
+endif()
```
pfultz2, line 122: It seems it would be better to write:

```cmake
if(NOT MIOPEN_EMBED_DB STREQUAL "")
    option(MIOPEN_DISABLE_SYSDB  "Disable sys database access" Off)
else()
    option(MIOPEN_DISABLE_SYSDB  "Disable sys database access" ${MIOPEN_EMBED_BUILD})
endif()
```

JehandadKhan: I was not sure if repeating the help string was good form, I will update it.

JehandadKhan:
> How was this tested?

By manually building with various CMake configs. If preferred I can add the flags to the Jenkins stage.

CMakeLists.txt
JehandadKhan: Fixed

PR #385, opened 2020-08-17
Disable faulty compiler option for global addressing using SGPR

asroy:
MIOpen is using "-amdgpu-enable-global-sgpr-addr" option for hip-clang compiler, but the feature has not been correctly implemented in compiler, and can cause GPU core dump in some cases (http://ontrack-internal.amd.com/browse/SWDEV-239555). Currently there is a workaround in MIOpen to prevent this failure. This PR remove this compiler flag and the workaround.

However, disabling the option can result in more than 20% regression in gfx908 fp16 cases.
http://ontrack-internal.amd.com/browse/SWDEV-239555?focusedCommentId=6260190&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6260190

A JIRA is opened to request proper implementation of SGPR offset for global addressing in compiler, so performance can be restored.
http://ontrack-internal.amd.com/browse/SWDEV-248564

### DIFF
--- a/src/hip/hip_build_utils.cpp
+++ b/src/hip/hip_build_utils.cpp
@@ -122,14 +122,13 @@ boost::filesystem::path HipBuild(boost::optional<TmpDir>& tmp_dir,
     params += MIOPEN_STRINGIZE(HIP_COMPILER_FLAGS);
     if(IsHccCompiler())
     {
-        env += std::string("KMOPTLLC=\"-mattr=+enable-ds128 -amdgpu-enable-global-sgpr-addr");
+        env += std::string("KMOPTLLC=\"-mattr=+enable-ds128 ");
         if(miopen::HipCompilerVersion() >= external_tool_version_t{2, 8, 0})
             env += " --amdgpu-spill-vgpr-to-agpr=0";
         env += '\"';
     }
     else if(IsHipClangCompiler())
     {
-        params += " -mllvm -amdgpu-enable-global-sgpr-addr";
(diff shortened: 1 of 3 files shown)


daniellowell:
So are we going to shoot MIOpen in the foot with this PR and hope the compiler team patches us up?

asroy:
@daniellowell Compiler is going to remove this option in future compiler, so we can no longer use it as long as it's still faulty (until they implemented it properly (http://ontrack-internal.amd.com/browse/SWDEV-248564)

But I prefer to hold this PR for a while and let things fly a bit, there has been some activity going on with SWDEV-248564, and I want to see how it goes.

atamazov:
Yes, the compiler team strongly recommends avoiding the use of implementation-dependent options (like this one).

I highly recommend looking ROCm (or compiler) version and deciding about this option based on that. IIRC we already have version detection machinery. Thus, the library will be backward compatible with the old compiler and, at the same time, will not use deprecated options with the newer one.

The hardest part is deciding in which versions this option should be enabled/disabled.

zjing14 (approved):

atamazov:
@daniellowell Still on hold?

atamazov:
Note that this PR should lead to regressions with some previous version(s) of ROCm (see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/385#issuecomment-677947221). But I do not know which version(s) affected.

atamazov:
> @daniellowell Still on hold?

Just in case -- The question was related to the `on_hold` label (which is still here). I had no intent to push this PR into develop.

PR #386, opened 2020-08-17
Remove repeated hip compiler args

JehandadKhan:
fixes an issue in the hip-config.cmake file where certain GPU related compiler arguments are supplied twice when using the hip-clang compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,7 +231,16 @@ if(MIOPEN_HIP_COMPILER MATCHES ".*clang\\+\\+$")
         PATHS /opt/rocm/llvm
     )
     # we are using hip-clang and need to remove repeated hip compiler args
-    string(REGEX REPLACE "-mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false" "" HIP_COMPILER_FLAGS "${HIP_COMPILER_FLAGS}")
+    string(REGEX MATCH " -mllvm -amdgpu-early-inline-all=(true|false)" MATCH_EARLY_INLINE "${HIP_COMPILER_FLAGS}")
+    if(MATCH_EARLY_INLINE)
+        string(REGEX REPLACE "${MATCH_EARLY_INLINE}" "" HIP_COMPILER_FLAGS "${HIP_COMPILER_FLAGS}")
+        message(STATUS "Removed ${MATCH_EARLY_INLINE} hip compiler flag to avoid repitition")
+    endif()
+    string(REGEX MATCH " -mllvm -amdgpu-function-calls=(false|true)" MATCH_FUNC_CALLS "${HIP_COMPILER_FLAGS}")
+    if(MATCH_FUNC_CALLS)
+        string(REGEX REPLACE "${MATCH_FUNC_CALLS}" "" HIP_COMPILER_FLAGS "${HIP_COMPILER_FLAGS}")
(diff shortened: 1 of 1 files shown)


JehandadKhan:
@pfultz2 @atamazov Can I bother you guys to take a look at this ?

CMakeLists.txt
```diff
@@ -236,7 +230,12 @@ if(MIOPEN_HIP_COMPILER MATCHES ".*clang\\+\\+$")
         PATH_SUFFIXES bin
         PATHS /opt/rocm/llvm
     )
+    # we are using hip-clang and need to remove repeated hip compiler args
+    string(REGEX REPLACE "-mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false" "" HIP_COMPILER_FLAGS "${HIP_COMPILER_FLAGS}")
```
pfultz2, line 234: Why is this is regex? This seems very specific to how the flags maybe added, which could be wrong if the way the flags are added change.

Wouldn't it be better to find the last `-mllvm -amdgpu-early-inline-all=\w+`, then remove all `-mllvm -amdgpu-early-inline-all=\w+` and then add the last item found back?

atamazov:
This is LGTM provided that Paul's comment is resolved. I already remove these options on COMgr path.

atamazov:
Notice that compiler guys always insist on NOT using `-mllvm` options but contacting compiler team when we need any of those.

CMakeLists.txt
daniellowell: @JehandadKhan Resolved?

pfultz2:
> Notice that compiler guys always insist on NOT using -mllvm options but contacting compiler team when we need any of those.

But these flags are coming from hip::device.

atamazov:
[Informative] @pfultz2 Yes. That is general notice unrelated to this PR. For example, we still have:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/13470cebfe05c865db2c6c636dd32f0abb6f5339/src/hip/hip_build_utils.cpp#L130-L134

CMakeLists.txt
JehandadKhan: The issue is not that different make files add them repeatedly, the issue is that the `hip-clang` adds these flags by default! Thus even if they exist only once here, the repetition would still happen. 

I agree with your point of these arguments being too specific and I will update the regex to handle that. 

Please note that we are only working around hip-config.cmake since they should not be added by that file in the first place and we would not have to do this at all.

daniellowell:
@JehandadKhan What is  status of this PR?

daniellowell:
All please re-review.

CMakeLists.txt
asroy: I don't follow the discussion here. Is that hip-clang adding the same flag twice, so you are removing the duplicated flags? If that is the case, I think it should be fixed by compiler team.

asroy: @JehandadKhan 
> I don't follow the discussion here. Is that hip-clang adding the same flag twice, so you are removing the duplicated flags? If that is the case, I think it should be fixed by compiler team.

asroy:
@JehandadKhan 
> I don't follow the discussion here. Is that hip-clang adding the same flag twice, so you are removing the duplicated flags? If that is the case, I think it should be fixed by compiler team.

atamazov:
Good question.

JehandadKhan:
It appears this issue is no longer relevant. We can open it again if observed.

PR #395, opened 2020-08-22
Refactor some macro guard for iGEMM code

asroy:
Fix macro issue that blocks MP_BD_winograd_xdlops. Also refactor some iGEMM solver side of code


src/kernels/composable_kernel/include/utility/config.hpp
```diff
@@ -11,6 +11,23 @@
 // device backend
 #define CK_DEVICE_BACKEND_AMD 1
 
+// hardware instruction support list
+#ifndef CK_HARDWARE_SUPPORT_AMD_V_MAC_F32
+#define CK_HARDWARE_SUPPORT_AMD_V_MAC_F32 1
```
atamazov, line 16: AFAICS these macros are always defined by the host. Does it worth to #error instead of defining the default values? Alternatively, the defaults could be set to something that would work on all GPUs (if host code fails to define some of macros).

asroy: I change these values to 0. Setting them to 1 does cause problem with compilation.

daniellowell:
@asroy status?

zjing14 (approved):

daniellowell:
@zjing14 Great that you approve, now @asroy let's get it to pass CI.

daniellowell:
@asroy PING

daniellowell:
@asroy 
```
./threadwise_gemm.hpp:132:21: error: use of undeclared identifier 'amd_assembly_outer_product_1x4'

                    amd_assembly_outer_product_1x4(p_a[aindex],

                    ^
```

@zjing14 please review.

src/kernels/composable_kernel/include/utility/config.hpp
```diff
@@ -11,6 +11,23 @@
 // device backend
 #define CK_DEVICE_BACKEND_AMD 1
 
+// hardware instruction support list
+#ifndef CK_HARDWARE_SUPPORT_AMD_V_MAC_F32
+#define CK_HARDWARE_SUPPORT_AMD_V_MAC_F32 0
```
zjing14, line 16: @asroy CK_HARDWARE_SUPPORT_AMD_V_MAC_F32 = 0? Why?

zjing14:
> @asroy
> 
> ```
> ./threadwise_gemm.hpp:132:21: error: use of undeclared identifier 'amd_assembly_outer_product_1x4'
> 
>                     amd_assembly_outer_product_1x4(p_a[aindex],
> 
>                     ^
> ```
> 
> @zjing14 please review.

It is due to CK_HARDWARE_SUPPORT_AMD_V_MAC_F32 = 0, I think.

src/kernels/composable_kernel/include/utility/config.hpp
asroy: v_mac is not supported on all hardware. So we should turn it off by default.

I should have but forgot to turn it on again in solver

atamazov:
Jenkins failures

asroy:
close becauase Kamil no longer need it

Issue #392, opened 2020-08-20
Extend FastHybrid mode -- exclude tunable solutions.


atamazov:
Extension of #299.

@daniellowell, originated from https://github.com/ROCmSoftwarePlatform/MIOpen/issues/380#issuecomment-675734858:
> We have 2 domains of MIOpen tuned known configs and untuned unknown configs. If it untuned and unknown, I don't think we should bother compiling tunable source kernels as they would naturally have inferior performance. If the user is a power user then they will know from reading documentation how to tune and how to set the find mode.

Note that not all tunable Solutions show bad performance. We have some Solvers that have advanced logic that produce good default PerformanceConfigs. IIRC legacy OCL Solutions are of this kind. So I think the design should allow us to specify which tunable Solutions considered "slow" by default and which are not.

Also we need to evaluate feasibility and complexity of the implementation.

@daniellowell Please set priority for this ticket.

atamazov:
Please note that:
- Right now the library does NOT read perf-db during GWSS calls. That was done for optimization purposes (see https://github.com/AMDComputeLibraries/MLOpen/pull/1930).
  - [Informative] This is possible because required workspace size does not depend on perf-config. This limitation is implied by Find API (let me omitt detailed explanations here).
- Therefore, during GWSS calls, we can't know if we have perf-db record in the dabase or not.
- Therefore we can't filter out "untuned" solvers during GWSS calls.
- The resulting side effect is that values returned by GWSS can be excessive in NT_HYBRID mode.

This can be worked out, if necessary, when/if this feature will have the "production" status (right now it is experimental) at the cost of some re-design.

atamazov:
The implementation is ready for testing with frameworks. I have smoke tested it and get promising results.

PR #396, opened 2020-08-23
NT_HYBRID Find mode

atamazov:
Experimental implementation of feature #392. This is is basically HYBRID mode that skips the most of untuned Solutions.

Smoke tested on Radeon VII, ROCm 3.5. Shows performance/behavior as expected.

### Evaluation of performance impact

Note that the tests were performed with BUILD_DEV=On and binary cache disabled. This is far from what happens when the real framework is running. However, this is more or less consistent with a situation where kernels are always compiled. I mean that the trends should be similar, i.e. if NT_HYBRID shows better wall time than HYBRID here, then I would expect the same in real life.

Find Mode | Wall time acceleration | GPU_perf_drop | Untuned Solutions disabled
-- | -- | -- | --
NORMAL | 1.00 | 0% | -
FAST | 26.44 | 51% | -
HYBRID | 1.00 | 0% | -
FAST_HYBRID | 1.55 | 0% | -
NT_HYBRID, commit (03), b8b6b9cd4 | 2.79 | 20% | All
NT_HYBRID, commit (04), f126b18b4 | 1.79 | 0% | All except ConvOclBwdWrW2<1>, ConvOclDirectFwd,   ConvOclDirectFwd1x1, ConvAsm3x3U, ConvAsmBwdWrW1x1, ConvBinWinogradRxSf2x3
NT_HYBRID, commit (05), d2ceb60bd | 2.43 | 5% | All except ConvOclBwdWrW2<1>, ConvAsm3x3U, ConvAsmBwdWrW1x1,   ConvBinWinogradRxSf2x3

### Smoke test - OK

How to smoke test:
- Clean user-find-db & run some config in NORMAL Find mode using the driver, collect logs at level 5.
  - The config should not be listed in the system find-db, for example: `./bin/MIOpenDriver conv -x 10 -y 5 -W 341 -H 79 -c 32 -n 3 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0` (note `-n 3`)
- Clean user-find-db, run the same config in the HYBRID mode, collect log. 
- Run the same config in the same mode (do not clean find-db), collect log. 
- Clean user-find-db, run the same config in the new NT_HYBRID mode, collect log
- Run the same config in the same mode (do not clean find-db), collect log. 
 - Compare logs ;)


atamazov:
CI tests passed.

atamazov:
Wall time is better on unknown configs (~2.8 times less than normal Find and Hybrid, ~1.8 times better than FastHybrid), which is good.

The GPU performance of this is significantly worse, however. Some untuned Solvers actually have good performance with default configs, so those shoud not be disabled. Specifically, `ConvOclBwdWrW2<1>`, `ConvBinWinogradRxSf2x3` and maybe more.

I am working on this.

atamazov:
After investigating logs, I've ended up enabling back untuned configs for
- `ConvOclBwdWrW2<1>`
- `ConvAsm3x3U`
- `ConvAsmBwdWrW1x1`
- `ConvBinWinogradRxSf2x3`

Now this mode has ~5% GPU perf drop but ~2.5X total wall time acceleration (`BUILD_DEV=on`). Topmost comment updated accordingly.
R is ready for review & merging. It is NFC provided that NT_HYBRID mode is not activated, so it should be safe to merge it.

However we need to test this with real frameworks and decide whether to keep it, fix it, or remove this now mode from the library.

/cc @daniellowell

daniellowell:
> -20%

Perf drop? Does that mean it got 20% faster, or slower?

atamazov:
> Perf drop? Does that mean it got 20% faster, or slower?

Of course, slower. Fixed.

daniellowell:
This is really good analysis. I think we need a phased approach. 5% is acceptable, but 20% is way too much, even for a fallback solution.

src/include/miopen/find_solution.hpp
```diff
@@ -97,6 +97,11 @@ auto FindSolutionImpl(rank<1>, Solver s, const Context& context, Db& db)
         }
     }
 
+    if(miopen::FindMode(context).IsNtHybrid() && !s.IsEnabledForNtHybrid())
+    {
+        // For now, let's use this for solutions skipped due to NT_HYBRID.
+        return {static_cast<miopenStatus_t>(100)};
```
pfultz2, line 103: Don't we need to call `GetPerformanceConfig` to find out if there are tunable parameters?

src/include/miopen/solver.hpp
```diff
@@ -1415,6 +1419,7 @@ struct ConvBinWinogradRxSf2x3 : SolverBase<ConvolutionContext>
     PerformanceConfigConvBinWinogradRxSf2x3 Search(const ConvolutionContext&) const;
 
     bool IsApplicable(const ConvolutionContext& params) const;
+    bool IsEnabledForNtHybrid() const { return true; }
```
pfultz2, line 1422: Why do we need to add this method to the solvers? Why not just get the tunable parameters from the performance config?

pfultz2: We can find out if the performance config has parameters by adding either a `GetParameters` or `HasParameters` method to the `Serializable` class:

```cpp
    std::vector<std::string> GetParameters() const
    {
        std::vector<std::string> result;
        auto out = static_cast<const Derived&>(*this);
        Derived::Visit(out, [&](const std::string& name, auto&&) {
            result.push_back(name);
        });
        return result;
    }

    bool HasParameters() const
    {
        bool result = false;
        auto out = static_cast<const Derived&>(*this);
        Derived::Visit(out, [&](auto&&, auto&&) {
            result = true;
        });
        return result;
    }
```

src/include/miopen/find_solution.hpp
pfultz2: Ah wait, if a solver is not tunable then it won't have a `GetPerformanceConfig` method. So this should just be:

```cpp
if(miopen::FindMode(context).IsNtHybrid()
    return {static_cast<miopenStatus_t>(100)};
```

atamazov: Yes. And `FindSolutionImpl(rank<1>,...)` is called with tunable solvers only.

src/include/miopen/solver.hpp
atamazov: > We can find out if the performance config has parameters...

Thanks. However we do not need this. All performance configs have parameters. Otherwise we solver doesn't need (and doesn't "have") a PerformanceConfig class, and, therefore is does not have the GetSolution that takes two parameters:
```C++
    /// Searchable solvers provide a GetSolution that takes a Context and PerformanceConfig
    /// ConvSolution GetSolution(const ConvolutionContext& params,
    ///                          const PerformanceConfig& config) const;
```
This is enough for now.

atamazov: > Why do we need to add this method to the solvers? Why not just get the tunable parameters from the performance config?

Well, this is something different. 

During testing of the initial implementation, I've found that some untuned Solvers actually have good performance with default configs, so those should not be disabled to avoid perf drop (see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/396#issuecomment-680135243). This method is very simple way to workaround this. It allows some tunable Solvers to participate in Find(), even if there is perf-db miss.

I do not like this -- basically this looks like a patch. However the whole PR looks like a crutch, so who cares. What I would really like to do is implement #410 and then get rid of both FAST_HYBRID and this NT_HYBRID.

atamazov:
@daniellowell (originated from https://github.com/ROCmSoftwarePlatform/MIOpen/issues/410#issuecomment-685797615)
> First get the PR merge that implements the reduced set of solvers so that we have the dynamic Find...

The design of this intermediate Dynamic Find does not match the design of NT_HYBRID, so I will prepare different PR with different DYNAMIC_HYBRID mode. This experimental PR is ready for testing, approval or rejection.

PR #139, opened 2020-04-08
MIOpenTensile Part1 : Add MIOpenTensile in GEMM path

ce1adon:
The purpose of this PR is to add MIOpenTensile path in GEMM algorithm for fp32 on hip backend. It replaces the default GEMM path through rocblas in HIP backend for fp32 type.

PR #400, opened 2020-08-24
disable ocl wrw2 solvers

ce1adon:


src/solver.cpp
```diff
@@ -231,12 +231,14 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
     RegisterWithSolver(registry, ++id, ConvBinWinogradRxS{}, miopenConvolutionAlgoWinograd);
     RegisterWithSolver(registry, ++id, ConvAsmBwdWrW3x3{}, miopenConvolutionAlgoDirect);
     RegisterWithSolver(registry, ++id, ConvAsmBwdWrW1x1{}, miopenConvolutionAlgoDirect);
+#if MIOPEN_BACKEND_OPENCL
```
atamazov, line 234: This is wrong. We shall put placeholders (`++id`) if we want remove entries from the list. See comment at the beginning of it.

atamazov (changes requested):
The simplest way to disable some Solver is making its `IsApplicable()` return `false'.

Notice that this would allow Immediate mode (and Fast Find mode) to work properly with existing fin-db, because immediate mode checks `IsApplicable()` before running the Solution. Otherwise it would be necessary to regenerate find-db.

atamazov:
Notice that some Ocl WrW2 solvers show very good performance untuned. These were originally non-tunable. What is the reason of disabling these?

atamazov:
IIRC right now we use the same system find-db for both HIP and OCL bakends (generated with HIP backend). After this PR, the system find-db would need to be generated separately for OCL backend, otherwise, with OCL backend, WrW convolutions run in Immediate, Fast Find, and Hybrid Find modes will suffer.

JehandadKhan:
I am not sure I follow the rationale behind disabling solvers only for HIP. Why is it ok to keep a solver for OpenCL and not for HIP ?

ce1adon:
> Notice that some Ocl WrW2 solvers show very good performance untuned. These were originally non-tunable. What is the reason of disabling these?

Three reasons:
1. wrw2<1> is leading logic error in this test https://github.com/ROCmSoftwarePlatform/MIOpen/pull/139#issuecomment-678445039

2. In fdb of all architectures wrw usages is less than 4% of all wrw related fdb records 

3. It has been compiled many times in some DL architecture but not finally get chosen, which means they contributed considerable amount of ineffectual exe time.


> The simplest way to disable some Solver is making its `IsApplicable()` return `false`.
> 
> Notice that this would allow Immediate mode (and Fast Find mode) to work properly with existing fin-db, because immediate mode checks `IsApplicable()` before running the Solution. Otherwise it would be necessary to regenerate find-db.

I was thinking about set isapplicable() to false, but worried these kernels may still contribute in the compilation time in the original building. Let's discuss this and nail a solution.

ce1adon:
> I am not sure I follow the rationale behind disabling solvers only for HIP. Why is it ok to keep a solver for OpenCL and not for HIP ?

I thought we still need these solvers on opencl backend, or we want to eliminate them on both backend?

JehandadKhan:
> > I am not sure I follow the rationale behind disabling solvers only for HIP. Why is it ok to keep a solver for OpenCL and not for HIP ?
> 
> I thought we still need these solvers on opencl backend, or we want to eliminate them on both backend?

Having them in one and not in the other creates imbalance, perhaps we should collect some data for the OpenCL backend as well and make a holistic decision.

ce1adon:
> Having them in one and not in the other creates imbalance, perhaps we should collect some data for the OpenCL backend as well and make a holistic decision.

@atamazov @JehandadKhan 
Another strategy in my mind is to have another macro or env to control the use of these kernels. Please evaluate and comment.

ce1adon:
@atamazov @JehandadKhan 
Before any changes of these solvers applied in dev, we will also need help from Peng's team in evaluating impact on performance. @sunway513

sunway513:
Do you think there are any particular workloads that this change can affect their performance?

atamazov:
@ce1adon 
> Another strategy in my mind is to have another macro or env to control the use of these kernels. Please evaluate and comment.

We have `MIOPEN_DEBUG_CONV_DIRECT_OCL_WRW2` env var.

atamazov:
https://github.com/AMDComputeLibraries/MLOpen/issues/2534 created.

atamazov:
See https://github.com/AMDComputeLibraries/MLOpen/issues/2534#issuecomment-699075454

PR #403, opened 2020-08-27
Refactored Xdlops gemm for vector type output

zjing14:
- Change output_c from array to vector type
- Resolve accvgpr over-allocation issue
- Resolve unnecessary accvgpr movement issue


src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops_fp16_bfp16.hpp
```diff
@@ -365,6 +357,7 @@ struct GridwiseGemmTransposedANormalBNormalCXdlopsFp16Bfp16_v1
             constexpr index_t BlkSize = OutputLayout.GetBlkSize();
             constexpr index_t NumBlks = OutputLayout.GetNumBlks();
 
+#pragma unroll
```
whchung, line 360: @zjing14 This pragma unroll is crucial to performance. Perhaps it's worthwhile to add comment to explain its importance?

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops_fp16_bfp16.hpp
```diff
@@ -724,6 +709,7 @@ struct GridwiseBatchedGemmTransposedANormalBNormalCXdlopsFp16Bfp16_v1
             constexpr index_t BlkSize = CLayout.GetBlkSize();
             constexpr index_t NumBlks = CLayout.GetNumBlks();
 
+#pragma unroll
```
whchung, line 712: same here. this unroll is extremely important to overall performance. better be documented.

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops_fp16_bfp16.hpp
```diff
@@ -1009,6 +988,7 @@ struct GridwiseBatchGemmXdlops_gkmkpack_gknkpack_gmn_v2
             constexpr index_t BlkSize = blockwise_gemm.GetBlkSize();
             constexpr index_t NumBlks = blockwise_gemm.GetNumBlks();
 
+#pragma unroll
```
whchung, line 991: same here. this unroll is extremely important to overall performance. better be documented.

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops_fp16_bfp16.hpp
zjing14: Added

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops_fp16_bfp16.hpp
zjing14: Done.

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops_fp16_bfp16.hpp
zjing14: Done.

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops.hpp
```diff
@@ -171,11 +169,7 @@ struct GridwiseGemmTransposedANormalBNormalCXdlops_v1
         __shared__ Float p_b_block_double[2 * b_block_space];
 
         // register allocation for output
-        AccFloat p_c_thread[c_k_thread_mtx_desc.GetElementSpace()];
-
-        // zero out threadwise output
-        threadwise_matrix_set_zero(c_k_thread_mtx_desc, p_c_thread);
-        blockwise_gemm.XdlopsMatrixCSetZero();
+        auto c_thread_vec = blockwise_gemm.GetOutputVec();
```
asroy, line 172: Is c_thread_vec initialized to be 0 here? If so, it's better to indicate it's 0-initialization from the function name

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops.hpp
```diff
@@ -325,7 +318,7 @@ struct GridwiseGemmTransposedANormalBNormalCXdlops_v1
                      m_thread_data_on_global % (M2 * M1) / M2,
                      m_thread_data_on_global % M2,
                      n_thread_data_on_global})
-                    .Run(p_c_thread + i * BlkSize, p_c_global);
+                    .Run(c_thread_vec.n + i * BlkSize, p_c_global);
```
asroy, line 321: Do you need a #pragma unroll above ? Line 294

not comment, just question:
Is this the only place in the code where pointer ```c_thread_vec.n``` is used to access c_thread_vec?

src/kernels/composable_kernel/include/tensor_operation/xdlops_gemm.hpp
```diff
@@ -330,14 +382,20 @@ struct mfma_info<mfma_instr::mfma_f32_32x32x2bf16>
     static constexpr index_t cycles          = 64;
     static constexpr index_t k_base          = 2;
 
-    template <index_t MPerXdlops, index_t NPerXdlops>
-    __device__ void run(const ushort* a, const ushort* b, float* reg_c) const
+    template <index_t MPerXdlops,
+              index_t NPerXdlops,
+              index_t AStride = 1,
+              index_t BStride = 1,
```
asroy, line 388: I thought C++ would allow default template argument only as the last arguments. I may be wrong
```
template<index_t MPerXdlops,
              index_t NPerXdlops,
              class FloatA,
              class FloatB,
              class FloatC,
              index_t AStride = 1,
              index_t BStride = 1>
```

src/kernels/composable_kernel/include/tensor_operation/xdlops_gemm.hpp
```diff
@@ -663,30 +754,16 @@ struct XdlopsGemm_t
...
-                                p_c_thread + (NRepeats * m_i + n_i) * GetRegSizePerXdlops());
-                    }
-                }
+            for(index_t k_i = 0; k_i < K * KRepeats; ++k_i)
+            {
+                p_c_thread = mfma_type.template run<MPerXdlops * MRepeats,
+                                                    NPerXdlops * NRepeats,
+                                                    AStride,
+                                                    BStride>(
+                    &pa[k_i * mfma_type.k_base], &pb[k_i * mfma_type.k_base], p_c_thread);
```
asroy, line 766: ```pa``` still seems to be a pointer to scalar, not vector. Am I right?

src/kernels/composable_kernel/include/utility/float_type.hpp
```diff
@@ -21,6 +21,153 @@ typedef ushort ushort2_t __attribute__((ext_vector_type(2)));
 typedef ushort ushort4_t __attribute__((ext_vector_type(4)));
 typedef ushort ushort8_t __attribute__((ext_vector_type(8)));
 
+struct c_vec32_4_t
```
asroy, line 24: Nice implementation for vector

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops.hpp
whchung: +1
`GetOutputVec` would 0 initialize `c_thread_vec`, so it's better to name it properly.

src/kernels/composable_kernel/include/tensor_operation/xdlops_gemm.hpp
whchung: That's default argument to functions, not templates.

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops.hpp
zjing14: Yes. #pragma unroll is required to eliminate stack.

Yes. c_thread_vec.n is only used here.

src/kernels/composable_kernel/include/tensor_operation/xdlops_gemm.hpp
zjing14: Removed these default arguments, which is useless.

src/kernels/composable_kernel/include/tensor_operation/gridwise_gemm_xdlops.hpp
asroy: Also, I suggest to use work ```Make``` instead of ```Get```, because it's creating a runtime variable

zjing14: The output vector is created at compiler time, right?

asroy: Yes, on ISA, it's created on compile time. From C++ side, it would looks like it's creating an variable from nowhere, and give it 0-value. So I think ```Make``` or ```Create``` is better than ```Get```

src/kernels/composable_kernel/include/tensor_operation/xdlops_gemm.hpp
whchung: Since `a/b/pa/pb` are normally small, I don't expect defining them as vectors would make much differences.

Perhaps it can be a future PR.

asroy: @whchung If compiler can optimize pointer access out entirely, maybe it's better to keep using pointer in C++?

whchung: True. So I think it's prudent to keep the PR in its current fashion. And file a separate PR should it be proven moving `a/b/pa/pb` to vector types really helps.

atamazov:
1. The compiler should be able to unroll loops automatically for the best performance.
1. Worse, each pragma unroll prevents compiler from doing some optimizations. This means that compiler regressions in these optimizations may went uncaught via all our testing.

That is why compiler team is asking us to report (open a ticket in Jira) each case where `#pragma unroll` gives some performance benefit.

___I highly recommend following this.___ Otherwise there is a danger that we will need to constantly edit the source codes of the kernels to maintain performance, -- rearranging and editing different pragmas depending on the compiler version.

Issue #406, opened 2020-08-28
Implement workaround for correctness problems in ConvHipImplicitGemmV4R4GenWrWXdlops


atamazov:
The workaround should make the solver non-applicable for the cases described at https://github.com/AMDComputeLibraries/MLOpen/issues/2532#issuecomment-682304119. This could be done quickly and would ensure correctness in case of delaying the delivery of the new bwd-wrw xdlops kernel mentioned at https://github.com/AMDComputeLibraries/MLOpen/issues/2532#issuecomment-683137628.

asroy:
@ltqin has test new bwd-wrw on these cases and they passed.

@zjing14 Please create workaround to disable the configs for ConvHipImplicitGemmV4R4GenWrWXdlops for now. The workaround doesn't need to be very accurate because ConvHipImplicitGemmV4R4GenWrWXdlops will be deprecated once new kernel is merged in.

zjing14:
@asroy Based on the investigation from @ekuznetsov139, the issue is related to BBlockCopySrcDataPerRead > 1. As the kernel will be deprecated, we can force BBlockCopySrcDataPerRead = 1, though it degrades performance.

atamazov:
This looks acceptable, but may lead to worse perf pregressions than if we make the solver non-applicable.

zjing14:
Yes. But other kernels without xdlops and atomic support won't perform well anyway.

atamazov:
> Yes. But...

If `ConvHipImplicitGemmV4R4GenWrWXdlops` may win over all other kernels even with enforced `BBlockCopySrcDataPerRead = 1`, then I was wrong, and narrowing applicability is not the best choice for the workaround ;)

zjing14:
I can do tests after enforcing BBlockCopySrcDataPerRead = 1 to see if ConvHipImplicitGemmV4R4GenWrWXdlops still outperform other kernels.

aserio:
@zjing14, can this issue be closed as your PR has been merged?

zjing14:
@aserio My PR does not fix this issue. We are waiting for this PR https://github.com/ROCmSoftwarePlatform/MIOpen/pull/413.

atamazov:
@zjing14 This blocker issue is dedicated to the workaround and now should be closed. It is the https://github.com/AMDComputeLibraries/MLOpen/issues/2532 issue that is of high priority and needs #413.

PR #409, opened 2020-08-31
Workaround for the failure of ConvHipImplicitGemmV4R4GenWrWXdlops

zjing14:
Resolves issue #406

### DIFF
--- a/src/solver/implicitgemm_util.hpp
+++ b/src/solver/implicitgemm_util.hpp
@@ -23,6 +23,8 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_BLOCK_SYNC_LDS_WITHOUT_SY
 // LLVM xdlops instrinsic will do unnecessey VGRP <--> AGPR movement, and result in
 // register spill, for bfloat16 datatype, when doing wave-wise GEMM larger than 64x64
 #define WORKAROUND_SWDEV_240356 1
+// workaround failure of ConvHipImplicitGemmV4R4GenWrWXdlops with vector load
+#define WORKAROUND_ISSUE_2532 1
 
 namespace miopen {
 
--- a/src/solver/conv_hip_implicit_gemm_v4r4_gen_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_v4r4_gen_xdlops.cpp
@@ -152,6 +152,11 @@ static inline ConvSolution GetSolutionBase(const ConvolutionContext& ctx,
         BBlockCopySrcDataPerRead_GemmN = 1;
     }
 
+#if WORKAROUND_ISSUE_2532
+    if(ctx.direction.IsBackwardWrW())
+        BBlockCopySrcDataPerRead_GemmN = 1;
+#endif
+
     std::size_t ABlockCopySrcDataPerRead_GemmK     = 1;
     std::size_t ABlockCopySrcDataPerRead_GemmKPACK = 1;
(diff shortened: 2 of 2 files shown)


zjing14:
Tested with configs from the issue. No performance drop after workaround.

src/solver/implicitgemm_util.hpp
```diff
@@ -23,6 +23,8 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_BLOCK_SYNC_LDS_WITHOUT_SY
 // LLVM xdlops instrinsic will do unnecessey VGRP <--> AGPR movement, and result in
 // register spill, for bfloat16 datatype, when doing wave-wise GEMM larger than 64x64
 #define WORKAROUND_SWDEV_240356 1
+// workaround failure of ConvHipImplicitGemmV4R4GenWrWXdlops with vector load
+#define WORKAROUND_ISSUE_2532 1
```
atamazov, line 27: [Recommendation] Used only once in a .cpp? Difine right there.

atamazov (approved):
LGTM

atamazov:
Topmost comment & priority fixed.

ekuznetsov139:
"No performance drop after workaround"

That sounds wrong to me. In my tests, I've seen performance drops as high as 50%.

atamazov:
@ekuznetsov139 Perhaps re-tuning is required. @zjing14 What do you think?

zjing14:
@ekuznetsov139 Could you post the config with regression here? @atamazov No, I did not retune.

ekuznetsov139:
MIOpenDriver convfp16 -n 256 -c 512 -H 28 -W 28 -k 128 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 1

Without the fix:
stats: bwdw-conv1x1u1, 256, 512, 28, 28, 1, 1, 128,  26306674688, 0, 0, 29434, 0, 0.893762
Backward Convolution Weights Failed: 0.241877 > 0.082

With the fix:
stats: bwdw-conv1x1u1, 256, 512, 28, 28, 1, 1, 128,  26306674688, 0, 0, 14505, 0, 1.813586
Backward Convolution Weights Verifies on CPU and GPU (4.11355e-05)

zjing14:
@ekuznetsov139 Thanks. Sorry, I did not compare the performance of failed configs, since I think it does not fair. Yes, that is huge performance degradation.

daniellowell:
@zjing14  So, this is a WIP again?

zjing14:
> @zjing14  So, this is a WIP again?

No, the PR is ready.

daniellowell:
@zjing14 What is the plan to reduce the impact of the performance regression? How widespread is this regression?

zjing14:
@daniellowell This solver will be deprecated after the new wrw solver merged in. So, the impact is temporary.

atamazov:
@zjing14 
> @ekuznetsov139 Thanks. Sorry, I did not compare the performance of failed configs, since I think it does not fair. Yes, that is huge performance degradation.

You were correct. The notion of performance is not applicable to the kernels that produce wrong outputs.

However, currently find-db contains too optimistic information about `ConvHipImplicitGemmV4R4GenWrWXdlops`, especially where find-db records say that it wins. That might lead to performance drops, **_unless_** `ConvHipImplicitGemmV4R4GenWrWXdlops` remains the winner even after workaround applied (or at least on par with the real winner). Does it?

PR #421, opened 2020-09-09
Set the CC compiler to /opt/rocm/llvm/bin/clang if available

daniellowell:
This PR will look for ROCm's clang for CC compilation if it can and if CC / CMAKE_CC_COMPILER is not set.


JehandadKhan:
@daniellowell what is the reason for this change ?

atamazov:
> @daniellowell what is the reason for this change ?

+1

PR #431, opened 2020-09-12
Use named params for buildJob

pfultz2:
This will help improve readability and also we can add more parameters easily.


atamazov:
Good change, thank you, Paul!

daniellowell (approved):

PR #435, opened 2020-09-14
Add code coverage reports

pfultz2:
This measures codecov using the gcc debug stage. We could try to add coverage from the hip backend as well in the future.

Right now, it only shows the coverage for this branch [here](https://codecov.io/gh/ROCmSoftwarePlatform/MIOpen/branch/codecov). Once its merged to develop then it will report back the diff coverage between develop and PRs.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,5 +1,4 @@
 ignore:
   - "addkernels/"
   - "driver/"
-  - "driver/"
   - "build/" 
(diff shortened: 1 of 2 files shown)


codecov.yml
```diff
@@ -0,0 +1,5 @@
+ignore:
+  - "addkernels/"
+  - "driver/"
```
JehandadKhan, line 3: why is driver repeated ?

pfultz2: Oops, its a type.

daniellowell (approved):

PR #439, opened 2020-09-16
Partially enable buffer_load/store for fp16 and bfp16

asroy:
* ROCm-3.7 (SWDEV-231101) supports fp16 and bfp16 buffer load/store instruction through "raw" buffer intrinsic. This PR partially enables these "raw" buffer intrinsics for fp16x1 and bfp16x1. This is partial enablement, because Soffset is set to 0, and only Voffset is used. Enabling Soffset requires major overhaul and is left for future. Performance should be comparable to global load/store for fp16x1 and bfp16x1 that use 64 bit SGPRs.
* fp32x1/x2/x4, fp16x2/x4/x8 and bpf16x2/x4/x8 load and store are still using the old (deprecated by compiler support) buffer intrinsic. Use of raw intrinsic for these data type will be left for future as well.
* BTW, also remove workaround SWDEV-240356 that was resolved by PR #403


src/kernels/composable_kernel/include/utility/amd_buffer_addressing.hpp
```diff
@@ -250,15 +194,11 @@ __device__ half_t amd_buffer_load<half_t, 1>(const half_t* p_src_block,
...
-#if !CK_WORKAROUND_SWDEV_231101
     index_t src_thread_addr_offset = src_thread_data_offset * sizeof(half_t);
     index_t src_const_addr_offset  = src_const_data_offset * sizeof(half_t);
 
-    return __llvm_amdgcn_buffer_load_f16(
-        src_block_config.data, 0, src_thread_addr_offset + src_const_addr_offset, false, false);
-#else
-    return p_src_block[src_thread_data_offset + src_const_data_offset];
-#endif
+    return __llvm_amdgcn_raw_buffer_load_f16(
```
whchung, line 200: @asroy perhaps it's better the add a comment here saying Soffset is currently hard-coded as 0.

src/kernels/composable_kernel/include/utility/amd_buffer_addressing.hpp
```diff
@@ -358,15 +284,11 @@ __device__ ushort amd_buffer_load<ushort, 1>(const ushort* p_src_block,
...
-#if !CK_WORKAROUND_SWDEV_231101
     index_t src_thread_addr_offset = src_thread_data_offset * sizeof(ushort);
     index_t src_const_addr_offset  = src_const_data_offset * sizeof(ushort);
 
-    return __llvm_amdgcn_buffer_load_bf16(
-        src_block_config.data, 0, src_thread_addr_offset + src_const_addr_offset, false, false);
-#else
-    return p_src_block[src_thread_data_offset + src_const_data_offset];
-#endif
+    return __llvm_amdgcn_raw_buffer_load_bf16(
```
whchung, line 290: @asroy same here. it's better to comment that we are hard-coding Soffset as 0 at this moment.

src/kernels/composable_kernel/include/utility/amd_buffer_addressing.hpp
```diff
@@ -545,19 +453,11 @@ __device__ void amd_buffer_store<half_t, 1>(const half_t* p_src,
...
-    __llvm_amdgcn_buffer_store_f16(*p_src,
-                                   dst_block_config.data,
-                                   0,
-                                   dst_thread_addr_offset + dst_const_addr_offset,
-                                   false,
-                                   false);
-#else
-    p_dst_block[dst_thread_data_offset + dst_const_data_offset] = *p_src;
-#endif
+    __llvm_amdgcn_raw_buffer_store_f16(
```
whchung, line 459: @asroy same here.

src/kernels/composable_kernel/include/utility/amd_buffer_addressing.hpp
```diff
@@ -649,19 +531,11 @@ __device__ void amd_buffer_store<ushort, 1>(const ushort* p_src,
...
-    __llvm_amdgcn_buffer_store_bf16(*p_src,
-                                    dst_block_config.data,
-                                    0,
-                                    dst_thread_addr_offset + dst_const_addr_offset,
-                                    false,
-                                    false);
-#else
-    p_dst_block[dst_thread_data_offset + dst_const_data_offset] = *p_src;
-#endif
+    __llvm_amdgcn_raw_buffer_store_bf16(
```
whchung, line 537: @asroy same here.

src/kernels/composable_kernel/include/utility/amd_buffer_addressing.hpp
asroy: fixed

src/kernels/composable_kernel/include/utility/amd_buffer_addressing.hpp
asroy: fixed

src/kernels/composable_kernel/include/utility/amd_buffer_addressing.hpp
asroy: fixed

src/kernels/composable_kernel/include/utility/amd_buffer_addressing.hpp
asroy: fixed

zjing14:
The performance of buffer_load with soffset=0 is the same as global_load?

asroy:
> The performance of buffer_load with soffset=0 is the same as global_load? I ask this since we observed that the compiler treats buffer_load and global_load differently.

Individual ISA should be comparable. But I'm not sure if the scheduling of the ISA have different, and if it would impact performance. I need to look into it more

asroy:
closing. Will add in future PR for dynamic composable kernel

PR #441, opened 2020-09-17
[dynamic-igemm][bwd-xdlops]enable asm igemm generic tensor contraction for backward_data

jane-zxy:
The goal of these solvers is to add a dynamic version of igemm, to reduce compilation time. This is implemented in asm, hence a new solver is added. This solver will have `IsDynamic()` return true, so any find mode utilize dynamic will utilize this solver
1. refactor the asm-igemm to reduce index caculation.
2. support range is the same as rocm3.7 HIP source code.
3. avg performance drop within limitation.
perf data and comparison with rocm3.7 HIP  is here: [perf_data](https://amdcloud-my.sharepoint.com/:x:/r/personal/xiaoyazh_amd_com/_layouts/15/Doc.aspx?sourcedoc=%7B27DE07A1-64CD-4E7E-BEB7-92F36E1A1240%7D&file=bwd_gtc_performance_for_models.xlsx&action=default&mobileredirect=true&cid=2c1a3ca3-0fab-42db-bf9a-09474cd6e680)
4. validated user cases are here: https://github.com/ROCmSoftwarePlatform/iGEMMgen/blob/master/script/gtc_conv_model.sh 

All kernel is in `src/kernels/dynamic_igemm/igemm_gtc_bwd/` folder

- igemm_bwd_gtc_gfx908.s 
- igemm_bwd_gtc_gfx908_004x064.inc 
- igemm_bwd_gtc_gfx908_008x064.inc 
- igemm_bwd_gtc_gfx908_016x032.inc 
- igemm_bwd_gtc_gfx908_016x064.inc 
- igemm_bwd_gtc_gfx908_016x128.inc 
- igemm_bwd_gtc_gfx908_016x256.inc 
- igemm_bwd_gtc_gfx908_032x016.inc 
- igemm_bwd_gtc_gfx908_032x032.inc 
- igemm_bwd_gtc_gfx908_032x064.inc 
- igemm_bwd_gtc_gfx908_032x128.inc 
- igemm_bwd_gtc_gfx908_032x256.inc 
- igemm_bwd_gtc_gfx908_064x004.inc 
- igemm_bwd_gtc_gfx908_064x008.inc 
- igemm_bwd_gtc_gfx908_064x016.inc 
- igemm_bwd_gtc_gfx908_064x032.inc 
- igemm_bwd_gtc_gfx908_064x064.inc 
- igemm_bwd_gtc_gfx908_064x128.inc 
- igemm_bwd_gtc_gfx908_064x256.inc 
- igemm_bwd_gtc_gfx908_128x016.inc 
- igemm_bwd_gtc_gfx908_128x032.inc 
- igemm_bwd_gtc_gfx908_128x064.inc 
- igemm_bwd_gtc_gfx908_128x128.inc 
- igemm_bwd_gtc_gfx908_128x256.inc 
- igemm_bwd_gtc_gfx908_256x016.inc 
- igemm_bwd_gtc_gfx908_256x032.inc 
- igemm_bwd_gtc_gfx908_256x064.inc 
- igemm_bwd_gtc_gfx908_256x128.inc 

perf data can be found in:
https://github.com/AMDComputeLibraries/MLOpen/issues/2535


jane-zxy:
@atamazov, 
Hi artem, would you help to review this PR, since it's scheduled to meet the 3.10 release. thks

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/*
+* Return with true if kernel is found, and its kernel_name, block_size, grid_size as well.
+* Return with false if no kernel can be executed with input ConvolutionContext
+*/
+static inline bool FindImplicitGemmGtcDynamicBwdKernel(const ConvolutionContext& ctx,
+                                                       std::string& kernel_name,
+                                                       int& block_size,
+                                                       int& grid_size)
```
shaojiewang, line 151: Here maybe use a std::tuple as return type which contains valid, kernel_name, block_size and grid_size.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,323 @@
...
+        if(pConfig->nxe == 0)
+        {
+            if((x != 1) || (y != 1) || (stride_h != 1) || (stride_w != 1) || (dilation_h != 1) ||
+               (dilation_w != 1) || (pad_h != 0) || (pad_w != 0))
+            {
+                continue;
+            }
+        }
+        if((gemm_n % pConfig->gemm_n_per_block != 0) || (gemm_m % pConfig->gemm_m_per_block != 0) ||
+           (k % pConfig->gemm_k_per_block != 0))
```
shaojiewang, line 206: I think that it could be better to put gemm_k instead of k. And gemm_k = k*y*x. In some conv layers of mask rcnn y=x=2.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,323 @@
...
+        {
+            int i_y_tilda          = gemm_id / x_tilda;
+            int i_x_tilda          = gemm_id % x_tilda;
+            int y_dot_slice        = (i_y_tilda + 1) * y_dot <= y ? y_dot : y % y_dot;
+            int x_dot_slice        = (i_x_tilda + 1) * x_dot <= x ? x_dot : x % x_dot;
+            int gemm_k             = k * y_dot_slice * x_dot_slice;
+            bool is_gemm_not_empty = gemm_k > 0;
+            if(is_gemm_not_empty)
+            {
+                if(gemm_k % pConfig->gemm_k_per_block != 0)
```
shaojiewang, line 234: If you need to check if gemm_k is divided gemm_k_per_block without remainder here, you could delete the check above.

shaojiewang (changes requested):

shaojiewang (commented):
I remember that you fixed a bug in bwd invoker when stride==2. But I have not seen any changes in invoker source code.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
fronteer: Should not call MIOPEN_THROW when no tunable is found

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: yes, you are right, I'll fix it.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: @shaojiewang @fronteer , you are all right, I'll fix.

sabreshao:
@jane-zxy Please add a kernel list which you validated for applicability.

atamazov:
@DrizztDoUrden Please review invokers.

jane-zxy:
> @jane-zxy Please add a kernel list which you validated for applicability.

ok, I've add this infor into this pr's description.

src/solver.cpp
```diff
@@ -330,6 +330,11 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
     RegisterWithSolver(
         registry, ++id, ConvAsmImplicitGemmV4R1DynamicWrw{}, miopenConvolutionAlgoImplicitGEMM);
 
+    RegisterWithSolver(registry,
```
asroy, line 333: new solver should be added to the end of the solver registry, so that the existing solver's ID will not be chagned

asroy (changes requested):
new solver should be added to the end of the solver registry. Otherwise, looks good

src/solver.cpp
carlushuang: @asroy hi, this is changed

atamazov:
I am beginning the reviewing process. At first I need to better understand what is the essence of the proposed changes.

> To reduce compilation time, re-implement the implicit gemm in assemble, and make parameters dynamic as well.

The PR doesn't reduce compilation time. It adds new solver at the end of `GetImplicitGemmSolvers()`. In normal Find mode this can't lead to better wall time. Perhaps you means some other Find mode? Let's make the description formally valid at least.

> 1. refactor the asm-igemm to reduce index caculation.

I do not see any changes in the existing solvers/kernels. Perhaps some of the existing iGemm HIP kernels was copied (which one?) and then that copy has been reworked?

> 2. support range is the same as rocm3.7 HIP source code.

Could you please elaborate? Do you mean then the new Solver has the same applicability as some other existing Solver?

> 3. avg performance drop within limitation.

What is the limitation?

> 4. validated user cases are here: https://github.com/ROCmSoftwarePlatform/iGEMMgen/blob/master/script/gtc_conv_model.sh

What is the `./out/conv_driver.exe` used in the script? What kinds of _validation_ have been performed?

> perf data can be found in:
> [AMDComputeLibraries/MLOpen#2535](https://github.com/AMDComputeLibraries/MLOpen/issues/2535)

Please look at https://github.com/AMDComputeLibraries/MLOpen/issues/2535#issuecomment-701453772, there are some important questions that must be answered. Thanks.

carlushuang:
Hi @atamazov 

> The PR doesn't reduce compilation time. It adds new solver at the end of GetImplicitGemmSolvers(). In normal Find mode this can't lead to better wall time. Perhaps you means some other Find mode? Let's make the description formally valid at least.

The goal of these solvers is to add a dynamic version of igemm. and this is implemented in asm, so we add a new solver. These solver will have `IsDynamic()` return `true`, so I think this solver will be utilized in find mode that use dynamic

> I do not see any changes in the existing solvers/kernels. Perhaps some of the existing iGemm HIP kernels was copied (which one?) and then that copy has been reworked?

I think what @jane-zxy means is refactor some index calculation logic used in original HIP igemm. the existing solver is not changed, and this is a new solver.

> Could you please elaborate? Do you mean then the new Solver has the same applicability as some other existing Solver?
Yes, this new solver have same applicability as HIP version backward data. I updated the applicability table in https://github.com/AMDComputeLibraries/MLOpen/issues/2535

> What is the limitation?
Previous we sync with Daniel, that average performance drop should within 10% compared to HIP version. So according to current result in https://github.com/AMDComputeLibraries/MLOpen/issues/2535, we have some cases better than HIP, some worse than HIP, and the average is improved.

> What is the ./out/conv_driver.exe used in the script? What kinds of validation have been performed?
Since we use https://github.com/ROCmSoftwarePlatform/iGEMMgen to generate all the asm kernels in these PRs, so in that repo we have a small host driver code (accept the same commandline args as MIOpenDriver), then we do layer wise validation & performance measue listed in [gtc_conv_model.sh](https://github.com/ROCmSoftwarePlatform/iGEMMgen/blob/master/script/gtc_conv_model.sh). And again, after integrate into MIOpen, we also run that script with MIOpenDriver, force to use the solver in these PR.

atamazov:
@carlushuang 

>> Let's make the description formally valid at least.

Thanks for the explanations. Please update PR description. It still confuses the reader.

> Previous we sync with Daniel, that average performance drop should within 10% compared to HIP version. So according to current result in [AMDComputeLibraries/MLOpen#2535](https://github.com/AMDComputeLibraries/MLOpen/issues/2535), we have some cases better than HIP, some worse than HIP, and the average is improved.

What I see looks like huge performance gain that makes HIP solvers unnecessary. However I am not sure if this is correct because some non-standard tool have been used to collect data. I am asking for additional testing round. Please see https://github.com/AMDComputeLibraries/MLOpen/issues/2535

>> What is the `./out/conv_driver.exe` used in the script? What kinds of _validation_ have been performed?
>
> ...in that repo we have a small host driver code ... then we do layer wise validation & performance measue listed in [gtc_conv_model.sh](https://github.com/ROCmSoftwarePlatform/iGEMMgen/blob/master/script/gtc_conv_model.sh). And again, after integrate into MIOpen, we also run that script with MIOpenDriver, force to use the solver in these PR.

Please use MIOpenDriver for performance and correctness testing. Please let me know if anything prevents your team from doing so before merging.

carlushuang:
Hi @atamazov I've updated the PR description. And for the tool used to measure performance, that is just a tool to speedup the debug/perf-measure process, and after that is done, we also double-checked that perf using MIOpenDriver. Because the tool in [iGEMMgen](https://github.com/ROCmSoftwarePlatform/iGEMMgen) use the same command line arg, so that make the whole process much easier and smooth. Any further information will update in [MLOpen#2535](https://github.com/AMDComputeLibraries/MLOpen/issues/2535)

asroy (dismissed):

src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -239,11 +239,13 @@ float CallImplGemmDynamicBackwardData(const miopen::Handle& handle,
         opArgs[26]             = OpKernelArg(_y_dot_slice);
         opArgs[27]             = OpKernelArg(_x_dot_slice);
         if(is_gemm_not_empty)
+        {
```
atamazov, line 242: This function performs many computations, thus wasting time. The most of those can be (and therefore must be) moved into the Invoker Factory. An Invoker instance will hold copies of the values computed _once_ by the its Factory.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+namespace miopen {
+namespace solver {
+
+static inline const std::vector<TunableImplicitGemmGTCDynamic_t>&
+GetImplicitGemmGtcDynamicBwdTunablesList()
+{
+    // clang-format off
+    static const std::vector<TunableImplicitGemmGTCDynamic_t> tunables = {
+        { "bwd", "fp32",  16,   0, 128, 128,  16,  32,  32,   1,   1,   2,   2,   {2,   1,   1,   4},   {1,   8,   1,  32},   {2,   1,   1,   4},   {1,   8,   1,  32},   0},
```
atamazov, line 41: What is the semantics of things like "bwd" and "fp32" and how these are used in the code?

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+        { "bwd", "fp32",   1,   1,   8,  64,  16,   4,  64,   1,   1,   1,   1,   {1,   1,   1,   1},   {1,  16,   1,   8},   {1,   1,   8,   1},   {1,  16,   1,   8},   0},
+        { "bwd", "fp32",   1,   0,   8,  64,  16,   4,  64,   1,   1,   1,   1,   {1,   1,   1,   1},   {1,  16,   1,   8},   {1,   1,   8,   1},   {1,  16,   1,   8},   0},
+        { "bwd", "fp32",   1,   1,   4,  64,  16,   4,  64,   1,   1,   1,   1,   {1,   1,   1,   1},   {1,  16,   1,   4},   {1,   1,  16,   1},   {1,  16,   1,   4},   0},
+        { "bwd", "fp32",   1,   0,   4,  64,  16,   4,  64,   1,   1,   1,   1,   {1,   1,   1,   1},   {1,  16,   1,   4},   {1,   1,  16,   1},   {1,  16,   1,   4},   0},
+    };
+    // clang-format on
+    return tunables;
+}
+
+/*
```
atamazov, line 144: Please use C++ comments.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+    int h_tilda_left = std::max(0, pad_h - dilation_h * (y_tilda - 1)) / stride_h;
+    int w_tilda_left = std::max(0, pad_w - dilation_w * (x_tilda - 1)) / stride_w;
+
+    int h_tilda_right = std::min(h_tilda, (pad_h + hi - 1 + stride_h - 1) / stride_h + 1);
+    int w_tilda_right = std::min(w_tilda, (pad_w + wi - 1 + stride_w - 1) / stride_w + 1);
+
+    int h_tilda_slice = h_tilda_right - h_tilda_left;
+    int w_tilda_slice = w_tilda_right - w_tilda_left;
+    int num_of_gemm   = y_tilda * x_tilda;
+    // clang-format on
```
atamazov, line 191: Please remove this line.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+* Return with false if no kernel can be executed with input ConvolutionContext
+*/
+static bool FindImplicitGemmGtcDynamicBwdKernel(const ConvolutionContext& ctx,
+                                                std::string& kernel_name,
+                                                int& block_size,
+                                                int& grid_size)
+{
+    auto tunables  = GetImplicitGemmGtcDynamicBwdTunablesList();
+    auto pConfig   = tunables.begin();
+    int hi         = ctx.out_height;
```
atamazov, line 155: `const auto&` if you just need an alias to the existing variable (everywhere).

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+    int wi         = ctx.out_width;
+    int n          = ctx.batch_sz;
+    int k          = ctx.n_inputs;
+    int c          = ctx.n_outputs;
+    int ho         = ctx.in_height;
+    int wo         = ctx.in_width;
+    int stride_h   = ctx.in_height > 1 ? ctx.kernel_stride_h : 1;
```
atamazov, line 162: `const` if you initialize a variable and do not modify it later (everywhere).

atamazov, line 162: `auto` instead of `int`

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+    int num_of_gemm   = y_tilda * x_tilda;
+    // clang-format on
+    int gemm_m = c;
+    int gemm_n = n * h_tilda_slice * w_tilda_slice;
+
+    for(; pConfig != tunables.end(); pConfig++)
+    {
+        if(pConfig->nxe == 0)
+        {
+            if((x != 1) || (y != 1) || (stride_h != 1) || (stride_w != 1) || (dilation_h != 1) ||
```
atamazov, line 199: [Coding style] Do you really need parentheses around comparisons, like `(x != 1)`?

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+               (dilation_w != 1) || (pad_h != 0) || (pad_w != 0))
+            {
```
atamazov, line 201: [Coding style] I highly recommend removing curly braces that enclose only one statement. These are not necessary.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+                continue;
+            }
+        }
+        if((gemm_n % pConfig->gemm_n_per_block != 0) || (gemm_m % pConfig->gemm_m_per_block != 0))
```
atamazov, line 205: We need programmatic guarantee that `pConfig->gemm_n_per_block` is never zero, otherwise DIV/0 could happen.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+        {
+            continue;
+        }
+        if(pConfig->gemm_n_per_block % pConfig->nxb != 0)
```
atamazov, line 209: We need programmatic guarantee that `pConfig->nxb` is never zero, otherwise DIV/0 could happen.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+                    integer_divide_ceil(gemm_n, pConfig->gemm_n_per_block);
+        return true;
+    }
+    return false;
+}
+
+bool ConvAsmImplicitGemmGTCDynamicBwdXdlops::IsApplicable(const ConvolutionContext& ctx) const
+{
+    const auto device_name = ctx.GetStream().GetDeviceName();
+    if(!(StartsWith(device_name, "gfx908")))
```
atamazov, line 258: `if(device_name != "gfx908")`

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/*
+* Return with true if kernel is found, and its kernel_name, block_size, grid_size as well.
+* Return with false if no kernel can be executed with input ConvolutionContext
+*/
+static bool FindImplicitGemmGtcDynamicBwdKernel(const ConvolutionContext& ctx,
+                                                std::string& kernel_name,
+                                                int& block_size,
+                                                int& grid_size)
```
atamazov, line 151: Use pointers instead of references and check for `nullptr` before writing.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+    if(!ctx.rmv.IsV3())
+        return false;
+
+    if(ctx.group_counts != 1)
+        return false;
+    std::string kernel_name;
+    int block_size;
+    int grid_size;
+    return FindImplicitGemmGtcDynamicBwdKernel(ctx, kernel_name, block_size, grid_size);
```
atamazov, line 278: [Clarity] pass nullptr instead of dummy variables.

atamazov (commented):
Part 1

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+    ConvSolution result;
+    KernelInfo kernel;
+    std::ostringstream options;
+
+    std::string kernel_name;
+    int block_size;
+    int grid_size;
+    bool ret = FindImplicitGemmGtcDynamicBwdKernel(ctx, kernel_name, block_size, grid_size);
+    if(!ret)
+        MIOPEN_THROW("this kernel should not run with igemm dynamic!");
```
atamazov, line 293: This check is not necessary, please remove.

atamazov (commented):
Part 2

src/CMakeLists.txt
```diff
@@ -251,6 +252,7 @@ endif()
 if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "HIP")
     file(GLOB_RECURSE COMPOSABLE_KERNEL_INCLUDE "kernels/composable_kernel/include/*/*.hpp")
     file(GLOB_RECURSE COMPOSABLE_KERNEL_SOURCE "kernels/composable_kernel/src/*/*.cpp")
+    file(GLOB_RECURSE COMPOSABLE_KERNEL_DYNAMIC_ASM_INCLUDE "kernels/dynamic_igemm/*.inc")
```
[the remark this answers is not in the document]
jane-zxy: thanks, I'll fix it.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include "miopen/solver.hpp"
```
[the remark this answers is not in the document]
jane-zxy: ok.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
```diff
@@ -0,0 +1,318 @@
...
+    kernel.l_wk.push_back(1);
+    kernel.l_wk.push_back(1);
+
+    GenerateClangDefsym(options, "ROCM_METADATA_VERSION", ctx.rmv.UseV3() ? 5 : 4);
+
+    kernel.comp_options = options.str();
+
+    MIOPEN_LOG_I2(kernel.kernel_file + ":" + kernel.kernel_name);
+
+    result.invoker_factory = conv::MakeImplGemmDynamicBackwardDataInvokerFactory(ctx);
```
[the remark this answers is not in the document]
jane-zxy: Hi artem,
MakeImplGemmDynamicBackwardDataInvokerFactory is used for two solvers: ConvAsmImplicitGemmGTCDynamicBwdXdlops and ConvAsmImplicitGemmV4R1DynamicBwd, so I'd like to keep it in 'impl_gemm_dynamic.cpp'

src/conv/invokers/impl_gemm_dynamic.cpp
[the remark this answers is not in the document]
jane-zxy: Hi artem,
no, this function has already been used for solver "ConvAsmImplicitGemmV4R1DynamicBwd" beforehand.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: Hi Artem,
'bwd' means one of the convolution directions, other two are 'fwd' and 'wrw', which you must already see in pr441 pr471. 
'fp32' means format of the data, which will be used in convolution.
both of them will be used to form the kernel name to be choosed, which will distinguish other kernels.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: ok, I'll change the function comments from /* */ to //

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: ok

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: ok

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: ok

jane-zxy: ok

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: Hi arterm,
thanks for your comments, however, I think it would be more readable if there's more than two expressions.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: pls see above comments.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: yes, you are right. I'll fix it.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: yes, you're right, I'll fix it.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: ok, accept.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: Hi Artem,
actually, I think using reference would be more safe than 'pointer', it would be guaranteed to be non-null while compiling, so programmer doesn't need check nullreference.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: pls see above comments

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: ok, I'll delete it.

src/conv/invokers/impl_gemm_dynamic.cpp
jane-zxy: Hi artem,
I don't quite understand your comments, MakeImplGemmDynamicBackwardDataInvokerFactory is called by solver when GetSolution(), and for each convolution, we need to pass specific kernArg into kernels, and for bwd, we need pre-computation to determine if we need launch several kernels or not.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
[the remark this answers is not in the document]
atamazov: I see, thanks.

src/conv/invokers/impl_gemm_dynamic.cpp
atamazov: The library keeps instances of Invokers in the cache, -- an instance per each convolution (problem config). That is why Invoker can hold parameters that are invariant for a problem config. For example, `n` (and, generally, the most of the parameters computed in this function, that depend only on problem config). This way, Invoker is not required to compute these parameters (or take these from ProblemDescription) each time when we want to run a kernel. This is one of the basic Invoker concepts.

The Factory should compute these parameters and pass to Invoker by-value (it is a lambda, so it it will hold copies). When library needs Invoker for some convolution (problem config), it will find it in the cache. Invoker will "read" invariant values from itself ans pass as parameters to the kernel. The only parameters that we need to pass to Invoker each time are, basically, input & output buffer addresses.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
atamazov: Repeating of the same values looks really bad. Please remove these from the table and replace
```
kernel_name = pConfig->GetKernelName();
```
by
```
kernel_name = pConfig->GetKernelName("bwd", "fp32");
```
in `FindImplicitGemmGtcDynamicBwdKernel()`.

atamazov: Please also move definition of `TunableImplicitGemmGTCDynamic_t` from solver.hpp. This is internal detail of implementation of a family of asm igemm gtc solvers (perhaps you'll need to create a dedicated header to store such details).

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
atamazov: Okay, but please notice that all other programmers do not use extra parentheses for readability (unless extreme cases). These will be confused by your code, just like me. And you will be confused by theirs code. Up to you.

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
atamazov: up to you

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
atamazov: Okay

src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp
jane-zxy: hi Artem,
As you know, for asm igemm, all kernels are physically generated by iGEMMgen[https://github.com/ROCmSoftwarePlatform/iGEMMgen/](url).
So actually TunableImplicitGemmGTCDynamic_t is the key shared structure, iGEMMgen use it to generate thousands of  asm kernels, and specific data member, like 'bwd', 'fwd', 'wrw','fp32' are used to generated the different kernels for different support senarior, it's like a config data base. 
For different convolution direction, there'll be different TunableImplicitGemmGTCDynamic_t list, you can refer to file: src/solver/conv_asm_implicit_gemm_gtc_bwd.cpp, src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp, src/solver/conv_asm_implicit_gemm_wrw_gtc_dynamic_xdlops.cpp.
So to make a conclusion, keep 'bwd', 'fwd', 'wrw' in the structure, will make solver more consistent with iGEMMgen, and make solver's config list more clear to specific convolution direction and data type. What do you think?

atamazov: > For different convolution direction, there'll be different TunableImplicitGemmGTCDynamic_t list

Of course I know that. And that is why repeating stuff like `"bwd", "fp32"` many times looks useless and confuses the reader. However, I am not going to insist on removing these.

BTW, is it so that the contents of "tunables" is generated from .config files (like this one: https://github.com/ROCmSoftwarePlatform/iGEMMgen/blob/master/config/igemm_bwd_gtc_gfx908.config) by iGEMMgen?

atamazov: Regardless of the above, moving the definition of `TunableImplicitGemmGTCDynamic_t` from solver.hpp to some header (common for asm igemm gtc solvers) should be done.

src/conv/invokers/impl_gemm_dynamic.cpp
jane-zxy: @carlushuang ,what do you think of this comments, since it's initially created by you.

jane-zxy: @atamazov , thanks, actually, our asm-igemm kernels need many kernels arguments, like buffer pointer, convolution parameters(like n, c, k, hi, wi, pad, stride, etc, also other config parameters as well), I'm quite don't understand how to make all these parameters avalable in 'invoker', but not get and compute from 'ConvolutionContext& ctx', is there example code existed in MIOpen now?

carlushuang: @jane-zxy  @atamazov  below is a tentative change for this invoker. This move some computation outside lambda, so within lambda now there is only 4 lines of computation, since bwd need multiple launch
```
InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ConvolutionContext& ctx)
{
    const auto& conv_problem = ctx.conv_problem;
    // clang-format off
    int hi          = conv_problem.GetOutHeight();
    int wi          = conv_problem.GetOutWidth();
    int n           = conv_problem.GetInBatchSize();
    int k           = conv_problem.GetInChannels();
    int c           = conv_problem.GetOutChannels();
    int ho          = conv_problem.GetInHeight();
    int wo          = conv_problem.GetInWidth();
    int stride_h    = conv_problem.GetInHeight() > 1 ? conv_problem.GetKernelStrideH() : 1;
    int stride_w    = conv_problem.GetInWidth() > 1 ? conv_problem.GetKernelStrideW() : 1;
    int dilation_h  = conv_problem.GetWeightsHeight() > 1? conv_problem.GetDilationH() : 1;
    int dilation_w  = conv_problem.GetWeightsWidth() > 1? conv_problem.GetDilationW() : 1;
    int pad_h       = conv_problem.GetPadH();
    int pad_w       = conv_problem.GetPadW();
    int y           = conv_problem.GetWeightsHeight();
    int x           = conv_problem.GetWeightsWidth();

    int gcd_stride_dilation_h = gcd(stride_h, dilation_h);
    int gcd_stride_dilation_w = gcd(stride_w, dilation_w);
    int y_tilda     = stride_h / gcd_stride_dilation_h;
    int x_tilda     = stride_w / gcd_stride_dilation_w;

    int y_dot = (y +  y_tilda - 1) / y_tilda;
    int x_dot = (x +  x_tilda - 1) / x_tilda;

    int h_tilda     = ho + (dilation_h * (y - 1) + stride_h - 1) / stride_h;
    int w_tilda     = wo + (dilation_w * (x - 1) + stride_w - 1) / stride_w;

    int h_tilda_left = std::max(0, pad_h - dilation_h * (y_tilda - 1)) / stride_h;
    int w_tilda_left = std::max(0, pad_w - dilation_w * (x_tilda - 1)) / stride_w;

    int h_tilda_right = std::min(h_tilda, (pad_h + hi - 1 + stride_h - 1) / stride_h + 1);
    int w_tilda_right = std::min(w_tilda, (pad_w + wi - 1 + stride_w - 1) / stride_w + 1);

    int h_tilda_slice = h_tilda_right - h_tilda_left;
    int w_tilda_slice = w_tilda_right - w_tilda_left;

    int num_of_gemms  = x_tilda * y_tilda;

    //int dtile_iy    = 0;
    //int dtile_ix    = 0;
    int dtile_dy    = dilation_h / gcd_stride_dilation_h;
    int dtile_dx    = dilation_w / gcd_stride_dilation_w;
    int dtile_y     = y_tilda;
    int dtile_x     = x_tilda;
    int dtile_h     = h_tilda;
    int dtile_w     = w_tilda;
    //int dslice_y    = 0;
    //int dslice_x    = 0;
    int dslice_h    = h_tilda_slice;
    int dslice_w    = w_tilda_slice;
    int dslice_h_left   = h_tilda_left;
    int dslice_w_left   = w_tilda_left;
    int __pack0     = 0;
    // clang-format on

    return [=](const std::vector<Kernel>& kernels) {
        const auto kernel = kernels[0];
        return [=](const Handle& handle, const AnyInvokeParams& primitive_parameters) {
            decltype(auto) data_ctx = primitive_parameters.CastTo<conv::DataInvokeParams>();
            const auto& tensors     = data_ctx.tensors;

            for(int gemm_id = 0; gemm_id < num_of_gemms; gemm_id++)
            {
                int dtile_iy_gid          = gemm_id / x_tilda;
                int dtile_ix_gid          = gemm_id % x_tilda;
                int y_dot_slice_gid       = (dtile_iy_gid + 1) * y_dot <= y ? y_dot : y % y_dot;
                int x_dot_slice_gid       = (dtile_ix_gid + 1) * x_dot <= x ? x_dot : x % x_dot;
                int gemm_k_gid            = k * y_dot_slice_gid * x_dot_slice_gid;
                bool is_gemm_not_empty = gemm_k_gid > 0;

                if(is_gemm_not_empty)
                {
                    handle.Run(kernel)(
                        tensors.out,
                        tensors.w,
                        tensors.in,
                        hi,
                        wi,
                        n,
                        k,
                        c,
                        ho,
                        wo,
                        stride_h,
                        stride_w,
                        dilation_h,
                        dilation_w,
                        pad_h,
                        pad_w,
                        y,
                        x,
                        dtile_iy_gid,
                        dtile_ix_gid,
                        dtile_dy,
                        dtile_dx,
                        dtile_y,
                        dtile_x,
                        dtile_h,
                        dtile_w,
                        y_dot_slice_gid,
                        x_dot_slice_gid,
                        dslice_h,
                        dslice_w,
                        dslice_h_left,
                        dslice_w_left,
                        __pack0);
                }
            }
        };
    };
}
```

atamazov: Yes, I think this is correct implementation of the approach. Except `__pack0` which is reserved identifier and must be renamed ;)

atamazov: Please also try to remove `//clang-format` stuff, perhaps the resulting formatting will be acceptable.

atamazov: Also I recommend moving the computational contents of the loop to the factory, and using arrays for storing results. In Factory:
```C++
//define arrays and then...
for(int gemm_id = 0; gemm_id < num_of_gemms; gemm_id++)
{
    dtile_iy_gid[gemm_id]    = gemm_id / x_tilda;
    dtile_ix_gid[gemm_id]    = gemm_id % x_tilda;
    y_dot_slice_gid[gemm_id] = (dtile_iy_gid[gemm_id] + 1) * y_dot <= y ? y_dot : y % y_dot;
    x_dot_slice_gid[gemm_id] = (dtile_ix_gid[gemm_id] + 1) * x_dot <= x ? x_dot : x % x_dot;
    const int gemm_k_gid     = k * y_dot_slice_gid[gemm_id] * x_dot_slice_gid[gemm_id];
    is_gemm_not_empty[gemm_id] = gemm_k_gid > 0;
}
```
In Invoker:
```C++
for(int gemm_id = 0; gemm_id < num_of_gemms; gemm_id++)
{
    if(!is_gemm_not_empty[gemm_id])
    	continue;
     handle.Run(kernel)(
            tensors.out,
            tensors.w,
            tensors.in,
            hi,
            wi,
            n,
            k,
            c,
            ho,
            wo,
            stride_h,
            stride_w,
            dilation_h,
            dilation_w,
            pad_h,
            pad_w,
            y,
            x,
            dtile_iy_gid[gemm_id],
            dtile_ix_gid[gemm_id],
            dtile_dy,
            dtile_dx,
            dtile_y,
            dtile_x,
            dtile_h,
            dtile_w,
            y_dot_slice_gid[gemm_id],
            x_dot_slice_gid[gemm_id],
            dslice_h,
            dslice_w,
            dslice_h_left,
            dslice_w_left,
            __pack0);
}

```

carlushuang: OK, @jane-zxy  let's follow the comment and update into https://github.com/ROCmSoftwarePlatform/MIOpen/pull/508

atamazov: [Resolved] Let's keep this in non-hidden state to allow hyperlinks to this thread.

PR #203, opened 2020-05-06
Invoker auto-tune

DrizztDoUrden:
This PR updates the generic search to use invokers instead of run and measure.
All relevant solvers are updated to support the new way of things, which has led to significant removal of duplicated code.

Things not done yet:

1. Fix fused conv 1x1u: at the moment it will have some build parameters duplicated when built for running and probably doesn't use invoker in the actual run.
2. Rigorous testing.

PR #454, opened 2020-09-22
MP_bidirect_winograd with xdlops [FP32]

shurale-nkn:
Added new MP_bidirect_winograd solver with xdlops support.
This solver is tunable.

gfx908 time improvements ___without new solver tuning___

N | C | K | H | W | x | y | p | q | u | v | directioon | time reduce | develop | xdlops |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
  |   |   |   |   |   |   |   |   |   |   |   |   | best algo | this solver |  
256 | 128 | 128 | 28 | 28 | 3 | 3 | 1 | 1 | 1 | 1 | FWD | 15% | ConvMPBidirectWinograd<6-3> | ConvMPBidirectWinograd_xdlops<6-3>
256 | 128 | 128 | 28 | 28 | 3 | 3 | 1 | 1 | 1 | 1 | BWD | 15% | ConvMPBidirectWinograd<6-3> | ConvMPBidirectWinograd_xdlops<6-3>
256 | 256 | 256 | 14 | 14 | 3 | 3 | 1 | 1 | 1 | 1 | FWD | 39% | ConvBinWinogradRxSf2x3 | ConvMPBidirectWinograd_xdlops<5-3>
256 | 256 | 256 | 14 | 14 | 3 | 3 | 1 | 1 | 1 | 1 | BWD | 38% | ConvBinWinograd3x3U | ConvMPBidirectWinograd_xdlops<5-3>
256 | 512 | 512 | 7 | 7 | 3 | 3 | 1 | 1 | 1 | 1 | FWD | 54% | ConvMPBidirectWinograd<6-3> | ConvMPBidirectWinograd_xdlops<4-3>
256 | 512 | 512 | 7 | 7 | 3 | 3 | 1 | 1 | 1 | 1 | BWD | 51% | ConvMPBidirectWinograd<6-3> | ConvMPBidirectWinograd_xdlops<4-3>


daniellowell:
@shurale-nkn Can you share target configs and performance data?

shurale-nkn:
> @shurale-nkn Can you share target configs and performance data?

If you need more data, I can create a private thread.

atamazov:
@shurale-nkn 
> If you need more data, I can create a private thread.

Better open an ticket for discussions at https://github.com/AMDComputeLibraries/MLOpen

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -31,22 +31,33 @@
...
+#include <miopen/miopen.h>
+#include <miopen/generic_search.hpp>
+#include <miopen/conv/invokers/impl_gemm.hpp>
 
 #include <boost/any.hpp>
-
-#if(MIOPEN_BACKEND_HIP && MIOPEN_USE_ROCBLAS)
 #include <miopen/conv/data_invoke_params.hpp>
+
+#if(MIOPEN_BACKEND_HIP)
```
atamazov, line 42: Please remove surrounding parentheses in all #ifs

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -31,22 +31,33 @@
...
-MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_MPASS_F3X3)
-MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_MPASS_F4X3)
-MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_MPASS_F5X3)
-MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_MPASS_F6X3)
-MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_MPASS_WORKSPACE_MAX)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_MP_BD_WINOGRAD_F2X3)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_MP_BD_WINOGRAD_F3X3)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_MP_BD_WINOGRAD_F4X3)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_MP_BD_WINOGRAD_F5X3)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_MP_BD_WINOGRAD_F6X3)
```
atamazov, line 53: Do we really need to rename envvars? If so, then we have to update the documentation. Also all the users who use these envvars would need to update theirs scripts.

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -87,6 +98,11 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_CONV_PRECISE_ROCBLAS_TIMING)
     GenerateClangDefsym((options), "fdilation_w", params.kernel_stride_w);                         \
     GenerateClangDefsym((options), "fdilation_h", params.kernel_stride_h);
 
+#define DEFINE_WORKSPACE_OFFSETS(in_size, wei_size, out_size) \
+    size_t wino_in_offset  = 0;                               \
+    size_t wino_out_offset = (in_size);                       \
+    size_t wino_wei_offset = wino_out_offset + (out_size);
```
atamazov, line 104:
```suggestion
struct WinoOffsets {
    const size_t in, out, wei;
    WinoOffsets(size_t in_size, size_t wei_size, size_t out_size)
    : in(0), out(in_size), wei(in_size + out_size)
    {}
};
```
and so on. Make it const as much as possible.

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -368,96 +348,138 @@ ConvSolution ConvMPBidirectWinograd<WinoDataH, WinoFilterH, WinoDataW, WinoFilte
...
+                    static_cast<int>(wino_wei_offset / GetTypeSize(params.in_data_type)),
+                    workSpace,
+                    static_cast<int>(wino_in_offset / GetTypeSize(params.in_data_type)),
+                    workSpace,
+                    static_cast<int>(wino_out_offset / GetTypeSize(params.in_data_type)),
+                    nullptr,
+                    false,
+                    GemmBackend_t::rocblas);
+#else
+                (void)handle;
```
atamazov, line 399: Why not MIOPEN_THROW?

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -530,5 +623,193 @@ template struct ConvMPBidirectWinograd<4, 3>;
...
+ConvMPBidirectWinograd_xdlops<WinoDataH, WinoFilterH, WinoDataW, WinoFilterW>::Search(
+    const ConvolutionContext& ctx) const
+{
+    return ConvHipImplicitGemmForwardV4R4Xdlops().Search(GetTransformedConvContext(ctx));
+}
+
+template <int WinoDataH, int WinoFilterH, int WinoDataW, int WinoFilterW>
+template <typename B, typename T>
+int ConvMPBidirectWinograd_xdlops<WinoDataH, WinoFilterH, WinoDataW, WinoFilterW>::
+    RunAndMeasureSolution(const miopen::Handle& handle,
```
atamazov, line 778: I recommend putting this on hold until #203 is merged in. `RunAndMeasureSolution` will not be needed after that.

src/include/miopen/solver.hpp
```diff
@@ -1492,6 +1492,76 @@ extern template struct ConvMPBidirectWinograd<4, 3>;
...
+    int RunAndMeasureSolution(const miopen::Handle& handle,
+                              B bot_buf,
+                              T top_buf,
+                              ConstData_t wei_buf,
+                              ConstData_t bias_buf,
+                              const ConvolutionContext& ctx,
+                              const ConvSolution& solution,
+                              float& elapsed_time) const;
+};
+extern template struct ConvMPBidirectWinograd_xdlops<2, 3>;
```
atamazov, line 1559: `<2,3>` is not needed.

atamazov (changes requested):

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -530,5 +623,193 @@ template struct ConvMPBidirectWinograd<4, 3>;
...
+            return false;
+    if(wino_data_tile == 4 && wino_filter_tile == 3)
+        if(miopen::IsDisabled(MIOPEN_DEBUG_AMD_MP_BD_XDLOPS_WINOGRAD_F4X3{}))
+            return false;
+    if(wino_data_tile == 3 && wino_filter_tile == 3)
+        if(miopen::IsDisabled(MIOPEN_DEBUG_AMD_MP_BD_XDLOPS_WINOGRAD_F3X3{}))
+            return false;
+
+    return IsApplicableTransform<WinoDataH, WinoFilterH, WinoDataW, WinoFilterW>(ctx) &&
+           ConvHipImplicitGemmForwardV4R4Xdlops().IsApplicable(GetTransformedConvContext(ctx));
```
atamazov, line 726: Depends on `ConvHipImplicitGemmForwardV4R4Xdlops`?

atamazov (changes requested):

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: Make it const or make it struct???

src/include/miopen/solver.hpp
shurale-nkn: Do you know any other winograd2x3 wtih xdlops? Or are there other reasons?

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: Yes

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: Without renaming 1 envvar will manage 3 solvers
ConvMPBidirectWinograd
ConvMPBidirectWinograd_xdlops
ConvWinograd3x3MultipassWrW

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: ``#if(MIOPEN_BACKEND_HIP && MIOPEN_USE_ROCBLAS)``
used only to avoid compiler error with OpenCl.
And ``MIOPEN_THROW`` used only for class functions to avoid errors if function must return an object.

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: Hmm, wait, it's too long... I'd better remove it .....right now.

src/include/miopen/solver.hpp
shurale-nkn: Needed! We don't have any other winograd f(2,3)  with xdlops

asroy:
@shurale-nkn Has rocblas been tuned for these configs?

src/solver/conv_MP_bidirectional_winograd.cpp
asroy: We need to write a proper solver for it, ```ConvHipImplicitGemmForwardV4R4Xdlops``` is not using the best layout. I would need to know what layout you are using for the transformation kernels.

src/solver/conv_MP_bidirectional_winograd.cpp
atamazov: Make it struct first. Then you can replace macro invocation with
```C++
const WinoOffsets winoOffset(in_size, wei_size, out_size);
...
// use:
... winoOffset.in ...
```

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -368,96 +349,143 @@ ConvSolution ConvMPBidirectWinograd<WinoDataH, WinoFilterH, WinoDataW, WinoFilte
...
+                    static_cast<int>(wino_in_offset / GetTypeSize(params.in_data_type)),
+                    workSpace,
+                    static_cast<int>(wino_out_offset / GetTypeSize(params.in_data_type)),
+                    nullptr,
+                    false,
+                    GemmBackend_t::rocblas);
+#else
+                (void)handle;
+                (void)ctx;
+                MIOPEN_THROW(miopenStatusBadParm, "ConvMPBidirectWinograd Unsupported ");
```
atamazov, line 404:
```suggestion
                MIOPEN_THROW(miopenStatusBadParm, "ConvMPBidirectWinograd is not supported");
```
And this is internal error, I think.

src/solver/conv_MP_bidirectional_winograd.cpp
atamazov: After #203, all Invokers must be able to measure elapsed time. Please double-check your Invoker and then press `[Resolve conversation]`. Thanks.

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -518,6 +546,78 @@ ConvSolution ConvMPBidirectWinograd<WinoDataH, WinoFilterH, WinoDataW, WinoFilte
             }
         };
     };
+#else
+    (void)params;
+    (void)xdlops_factory;
+    (void)isXdlops;
+    MIOPEN_THROW(miopenStatusBadParm, "ConvMPBidirectWinograd Unsupported ");
```
atamazov, line 553: Ditto.

src/include/miopen/solver.hpp
atamazov: But is wasn't instantiated. Now I see you've removed this declaration.

src/solver/conv_MP_bidirectional_winograd.cpp
atamazov: It is necessary to clarify why this is necessary. Please add respective comment.

Please separately comment cases when:
- The `ConvHipImplicitGemmForwardV4R4Xdlops` solver is disabled (e.g. via environment).
- The iGEMM algo is totally disabled.

src/solver/conv_MP_bidirectional_winograd.cpp
atamazov: Thanks for fixing.

src/solver/conv_MP_bidirectional_winograd.cpp
atamazov: Formally speaking, no. The old names would look ambiguous without renaming, but this is acceptable cost of the backward compatibility. Thanks for fixing.

atamazov (changes requested):
Minor issues

src/solver/conv_MP_bidirectional_winograd.cpp
asroy: Is this PR using Tensile or ```ConvHipImplicitGemmForwardV4R4Xdlops``` for the GEMM part?

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: They were able to do this before #203. 
But now they can make tunning without 'RunAndMeasureSolution'.

But this solver during the tuning process, calls 'ConvHipImplicitGemmForwardV4R4Xdlops.Search()' and tunes ConvHipImplicitGemmForwardV4R4Xdlops solver without using its own kernels.

src/solver/conv_MP_bidirectional_winograd.cpp
shurale-nkn: @asroy 
>Is this PR using Tensile or ConvHipImplicitGemmForwardV4R4Xdlops for the GEMM part?

ConvHipImplicitGemmForwardV4R4Xdlops 


>We need to write a proper solver for it, ConvHipImplicitGemmForwardV4R4Xdlops is not using the best layout. I would need to know what layout you are using for the transformation kernels.

The original format presented in ConvMPBidirectWinograd is - GCNHW (for data and filter).
To work with ConvHipImplicitGemmForwardV4R4Xdlops the context is converted to NGCHW but N=1, G=G, C=C, H=1, W=H*G*W (for data), and in filter buffer used GNCHW (G=G, N=K, C=C, H=1, W=1). (The filter kernel code was changed specifically for this purpose)

shurale-nkn: @atamazov 

>It is necessary to clarify why this is necessary. Please add respective comment.
>
>Please separately comment cases when:
>
>The ConvHipImplicitGemmForwardV4R4Xdlops solver is disabled (e.g. via environment).

As you can see, there is no such environment variable here
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/3ff4d7fec01888ed18fad69565f703b82afee451/src/solver/conv_hip_implicit_gemm_v4r4_gen_xdlops.cpp#L439-L451

But if conv_hip_implicit_gemm_v4r4_gen_xdlops is disabled, this solver disabled to. (concept)

>The iGEMM algo is totally disabled.

Same principle as with GEMM.  This solver will be enabled.

I don't see any non-obvious points here. The previous solution worked on the same logic.

shurale-nkn:
> @shurale-nkn Has rocblas been tuned for these configs?

@asroy 
If we speak about ConvMPBidirectWinograd algo in develop, then no, because ticket for tuning created a week ago. They wouldn't have had time to do it.

If about ConvMPBidirectWinograd_xdlops, than no, ConvHipImplicitGemmForwardV4R4Xdlops not tuned, used config from EuristicInit .

src/kernels/xform_bidirect_winograd_code.inc
```diff
@@ -91,6 +91,7 @@
 
 default pipe_config, 0
 default use_exec_for_vmem, 1 // todo test and remove
+default XDLOPS_CONV_GEMM, 0
```
Kirpich30000, line 94: I suggest to name the variable by it's meaning. What it really do is switching layout. Please name it accordingly.

Kirpich30000 (changes requested):
minor changes

Kirpich30000 (approved):
LGTM

daniellowell:
@shurale-nkn Merge conflicts and blocking review. Please update this ASAP, otherwise it will not get into ROCm 3.10.

Kirpich30000 (approved):
LGTM

src/solver/conv_MP_bidirectional_winograd.cpp
atamazov: OK

atamazov (approved):
Good job of reusing existing kernels/solvers
makeitso

atamazov:
Please do not forget to document `MIOPEN_DEBUG_AMD_MP_BD_WINOGRAD_WORKSPACE_MAX` in separate PR. Also it would be nice to get rid of `// cppcheck-suppress unreadVariable` for `WinoOffsets`.

shurale-nkn:
@daniellowell 
Update! Git doesn't display this, but the testing was completed successfully.

PR #462, opened 2020-09-24
[dynamic-igemm][fwd-xdlops]Assembly implementation of dynamic implicit-gemm generic tensor contraction for convolution-forward

fronteer:
Dynamic implicit-gemm for gtc convolution-forward path using fp32 Xdlops instruction assembly:

    1) Reduce kernel compile-time due to the coding in assembly
    2) support the same range of user cases as rocm3.7's HIP-based igemm kernels (except group conv);


src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp
```diff
@@ -0,0 +1,557 @@
...
+    int pad_h      = ctx.pad_h;
+    int pad_w      = ctx.pad_w;
+    int y          = ctx.kernel_size_h;
+    int x          = ctx.kernel_size_w;
+
+    int gemm_m = k;
+    int gemm_n = n * ho * wo;
+    int gemm_k = c * y * x; 
+
+    auto pConfig   = tunables.begin();
```
carlushuang, line 436: `pConfig` should be more suitblable in `for( auto pConfig ....`, or clang-tidy might not pass?

src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp
```diff
@@ -0,0 +1,557 @@
...
+        if(pConfig->nxe == 0)
+        {
+            if((x != 1) || (y != 1) || (stride_h != 1) || (stride_w != 1) || (dilation_h != 1) ||
+               (dilation_w != 1) || (pad_h != 0) || (pad_w != 0))
+            {
+                continue;
+            }
+        };
+        if((gemm_n % pConfig->gemm_n_per_block != 0) || (gemm_m % pConfig->gemm_m_per_block != 0) || (gemm_k % pConfig->gemm_k_per_block != 0))
+	{
```
carlushuang, line 448: please run clang-format locally. some indent seems strange

carlushuang (commented):
several format

src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp
fronteer: Changed !

src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp
fronteer: It seems my clang-format-3.8 does not solve the issue.  Fixed manually

shaojiewang: clang-format does not work because you set "//clang-format-off" at the very beginning of the first function in this file. You might reset clang-format to ON at the end of the first function to make clang-format really work.

src/solver.cpp
```diff
@@ -330,6 +330,11 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
     RegisterWithSolver(
         registry, ++id, ConvAsmImplicitGemmV4R1DynamicWrw{}, miopenConvolutionAlgoImplicitGEMM);
 
+    RegisterWithSolver(registry,
```
asroy, line 333: A new solver should always be added to the end of the list, so it doesn't change the solver-id of existing solvers

asroy (changes requested):
new solver should be append to the end of solver registry. Otherwise, looks good

asroy:
@fronteer please provide a link perf number

asroy:
@fronteer This branch ```fronteer:dyn-igemm-fwd``` seems to come from another repo? CI would not pick up this branch. You need to use a branch in this repo.

src/solver.cpp
fronteer: Yes, right!

src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp
fronteer: @shaojiewang,  oh, yes, thank you.   I will never notice this without your mentioning.

fronteer:
Refer to PR.471 for a replacing

https://github.com/ROCmSoftwarePlatform/MIOpen/pull/471

PR #466, opened 2020-09-26
Fixed EuristicInit function in BwdDataV4R1Xdlops solver

zjing14:
- Corrected the search range of EuristicInit
- Moved the BwdDataV4R1Xdlops up to expose it to CI

PR #457, opened 2020-09-22
[Workaround] Disable bwd v1r1 xdlops due to compiler issues

zjing14:
- Disabled bwd v1r1 xdlops due to the issue #456
- Related to SWDEV_251757
- Merge after #466

PR #468, opened 2020-09-28
[Draft] Fail-safe for PR  #466 and #457. May not needed

asroy:
Follow-up of #466 in case it failed.

This PR add a fail-safe in ConvHipImplicitGemmBwdDataV4R1Xdlops::IsApplicable() by returning false if ConvHipImplicitGemmBwdDataV4R1Xdlops::EuristicInit() cannot find valid performance parameter. However, if EuristicInit() doesn't iterate over the full tuning space of GenericSearch, then this PR will reduce the supporting range of this solver.

If #466 passed, this PR can be abandoned.


asleepzzz:
looks good

asroy:
no longer needed because #466 has fixed the issue

PR #470, opened 2020-09-29
remove workaround for issue #2335

ce1adon:
The issue https://github.com/AMDComputeLibraries/MLOpen/issues/2335 is gone in rocm3.7. 
Remove workaround https://github.com/AMDComputeLibraries/MLOpen/pull/2336.


ce1adon:
The same failure was observed in certain architecture. Further evaluation is needed.

atamazov:

junliume:
@ce1adon @atamazov As part of pre-holiday clean up, this PR seems to be low complexity and do we plan to proceed with it? Thanks!

atamazov:
> ...Then first merge from develop. Push the `[Ready for review]` button only after that. Thanks.

Sure. Let's merge from develop and continue ;)

ce1adon:
Up to latest develop now

junliume (approved):

atamazov (changes requested):
This is Ok provided that we do not support for ROCm < 3.7 anymore. @junliume Please confirm.

Otherwise we have to add a HIP version check around W/A removal.

atamazov:
This is how (if necessary):
```C++
#include <miopen/config.h>

// OpenCL error creating buffer: 0 Invalid Buffer Size
#define WORKAROUND_MLOPEN_ISSUE_2335 ((HIP_PACKAGE_VERSION_FLAT < 3007000000ULL) && MIOPEN_BACKEND_OPENCL)
...
#if WORKAROUND_MLOPEN_ISSUE_2335
        std::cout << "Skip test for Issue #2335: " << std::endl;
        return;
#endif
```

junliume:
> This is Ok provided that we do not support for ROCm < 3.7 anymore. @junliume Please confirm.
> 
> Otherwise we have to add a HIP version check around W/A removal.

I was under the impression that we are phasing out 3.7 in our CI because of so, but I might be wrong ... 
I will check with the program about life support schedule. Meanwhile, @ce1adon could you add a HIP version check around W/A removal as suggested by Artem? Thanks!

ce1adon:
@junliume @atamazov 
Sure, I will add a version workaround. Although I doubt that the latest MIOpen is still compatible to rocm version <3.7.

junliume (approved):

atamazov:
Forgotten header:

> This is how (if necessary):
> 
> ```c++
> #include <miopen/config.h>
> ...
> ```

But it seems acceptable because CI builds the test without problems.

atamazov:
@ce1adon 
> @junliume @atamazov
> Sure, I will add a version workaround. Although I doubt that the latest MIOpen is still compatible to rocm version <3.7.

Theoretically MIOpen is compatible with 3.5. IIRC hcc is replaced by hip-clang starting from that version.

atamazov (commented):
No objections.

ce1adon:
> Forgotten header:
> 
> > This is how (if necessary):
> > ```c++
> > #include <miopen/config.h>
> > ...
> > ```
> 
> But it seems acceptable because CI builds the test without problems.

The line "#include <miopen/config.h>" is included in test/dropout_util.hpp, which is then included in test/dropout.cpp in this PR, like matryoshka.

atamazov:
@ce1adon 
> The line "#include <miopen/config.h>" is included in test/dropout_util.hpp, which is then included in test/dropout.cpp in this PR, like matryoshka.

Of course -- in that header or somewhere else, -- otherwise build error would happen. But this is implicit dependence. The generally adopted practice is: if a file uses some entity from the header, then include that header into that file, explicitly.

PR #413, opened 2020-09-02
implicit gemm backward weight with xdlops and atomic add

ltqin:
Realize backward weight function, some optimization:
1.	use v2 version of gridwise-GEMM
2.	use buffer_atomic_fadd

PR #472, opened 2020-09-30
implicit gemm backward weight with xdlops and atomic add 2

ltqin:
this is a copy of PR #413, because PR #413 is using a branch from forked repo, and it was testing using the old testing script (every restart will restart from zero) and keeps timing out.

Realize backward weight function, some optimization:
use v2 version of gridwise-GEMM
use buffer_atomic_fadd


atamazov:
Formatting fixed.

asroy (approved):

PR #474, opened 2020-09-30
[BN] Update kernel to stabilize variance calculation.

daniellowell:
~~This comes from this issue. Should stabilize the calculation:
http://ontrack-internal.amd.com/browse/SWDEV-253606~~

Turns out not to be the issue. Still though calculations are not stable in our tests.


atamazov:
@muralinr Do we need this patch, what do you think?

muralinr (approved):
> @muralinr Do we need this patch, what do you think?

I think we need it for any corner cases..

atamazov:
Passed all CI tests 4 months ago.

Issue #483, opened 2020-10-02
WORKAROUND_SWDEV_227826 should not affect iGEMM solvers written in ASM


atamazov:
Originated from https://github.com/AMDComputeLibraries/MLOpen/issues/2535#issuecomment-702984109.

I expect that importance of this will grow over time.

/cc @daniellowell @asroy @JehandadKhan

asroy:
PR #488 would remove this workaround, so all iGEMM solver will be enabled by default.

ppanchad-amd:
@atamazov Is this still a relevant issue?

atamazov:
WORKAROUND_SWDEV_227826 is disabled in #1595.

PR #488, opened 2020-10-05
Enable all iGEMM solvers by default

asroy:
This PR ~remove a~ disables workaround, so MIOpen would enable all iGEMM solvers by default.

Also resolves issue #483


atamazov:
Let's discuss this on meeting today.

src/mlo_dir_conv.cpp
```diff
@@ -255,14 +239,7 @@ std::vector<miopen::solver::ConvSolution>
 FindImplicitGemmWrWAllSolutions(const miopen::ConvolutionContext& ctx,
                                 const miopen::AnyInvokeParams& invoke_ctx)
 {
-#if WORKAROUND_SWDEV_227826
-    if(miopen::IsEnabled(MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS{}))
```
atamazov, line 259: If we want to reverse the default behavior, then let's simply change `IsEnabled()` to `!IsDisabled()` wherever `MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS` is used. No other changes required for now, I think.

atamazov (changes requested):

src/mlo_dir_conv.cpp
asroy: fixed

atamazov: Let's minimize changes.

atamazov (approved):
LGTM

src/mlo_dir_conv.cpp
atamazov: Resolved by https://github.com/ROCmSoftwarePlatform/MIOpen/pull/488/commits/bbe1ffb1ac9cd4ad103405d4c0cd16d2e02f2596

JehandadKhan (approved):

asroy:
@daniellowell test_conv_for_implicit_gemm failed. But there is no log on which case failed. Restarted the CI job

atamazov:
@asroy @daniellowell Do we still need this (i.e. what is the value of it)?

streamhsa:
blocked by #540 which include a bug fix, otherwise CI would not pass

asroy:
@atamazov All iGEMM solvers (HIP/asm) will be used if MIOPEN_FIND_MODE=1.

This PR is blocked by #540 which has a bug fix for asm-iGEMM kernel, otherwise this PR will not pass CI

atamazov:
@asroy 
> All iGEMM solvers (HIP/asm) will be used if MIOPEN_FIND_MODE=1.

Of course. The question is how do we estimate the overall value of this change. One may think this is not good (the value is negative), because `Find()` will take longer. Another one may consider this is of small value.

daniellowell:
@asroy This PR needs more work. It consistently fails on FP32 gfx908 tests.
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/enable_all_igemm_solver/18/pipeline

Please open additional tickets for bug fixes.

atamazov:

asroy:
No longer pursuit

atamazov:
> No longer pursuit

Why? Since #615 (2 days ago) we have parallel builds of all solutions, so the library can tolerate this.

Issue #336, opened 2020-07-09
API to set the MIOPEN_FIND_MODE


daniellowell:
If we are adding more modes, we need to have a API to control this. This is a requirement for MIGraphX via @pfultz2.

The question is, where will this be tied into? Set to the handle?

JehandadKhan:
I think is a good time for the (very) long await Find 2.0

pfultz2:
We could create an opaque object that has the find settings or use a local struct, but we will need a new API function to pass this along. 

It would also be nice if we can merge `miopenConvAlgoPerf_t` and `miopenConvSolution_t`, but it looks like we screwed that one up(the field for `memory` is called `workspace_size`).

JehandadKhan:
Perhaps the same opaque object may merge immediate and non-immediate as well ?

pfultz2:
> Perhaps the same opaque object may merge immediate and non-immediate as well ?

But that will require adding v2 for immediate mode as well, which then we could switch the name.

atamazov:
Find 2.0 API and Find Modes (and other modes like this) are unrelated so far.

@pfultz2 Would you like to have a global (per library instance) Find Modes or you need capability to specify mode in each call?

atamazov:
> It would also be nice if we can merge `miopenConvAlgoPerf_t` and `miopenConvSolution_t`, but it looks like we screwed that one up(the field for `memory` is called `workspace_size`).

Sure, Replacing `miopenConvAlgoPerf_t` with `miopenConvSolution_t` in Find API 2.0 calls is one of its targets. There is a short survey at https://github.com/AMDComputeLibraries/MLOpen/issues/1938 (100).

pfultz2:
> Find 2.0 API and Find Modes (and other modes like this) are unrelated so far.

Kind of. They are somewhat unrelated but it may be necessary to have a new API function call, and if we are adding a new Find API function call we should make sure it can work with Find 2.0.

> Would you like to have a global (per library instance) Find Modes or you need capability to specify mode for each Find API call?

It would be best to specify it for each API call.

daniellowell:
> > Perhaps the same opaque object may merge immediate and non-immediate as well ?
> 
> But that will require adding v2 for immediate mode as well, which then we could switch the name.

I don't see this as a problem.

daniellowell:
I would like unify the data structures and parameters between find and immediate mode. If this requires V2 of everything that is fine. 

[...]

PR #489, opened 2020-10-05
Find Mode API

atamazov:
Implements API as per https://github.com/ROCmSoftwarePlatform/MIOpen/issues/336#issuecomment-701882762 (and below).

- `FindMode` made a member of `ConvolutionDescriptor`.
- Added & implemented hidden C API for MIGraphX
- Added & implemented an experimental C API (made non-public for now).
  - Used in `MIOpenDriver` and in `test_find_db`.
- Got rid of `miopen::debug::FindModeDisable` hackery.
- `MIOPEN_FIND_ENFORCE_SCOPE` dismissed.

## :cyclone: Local performance & correctness testing

- Radeon VII, ROCm 3.5.
- Baseline (develop at e08e23d24) vs this PR (ccd674c8a)
- GPU time, Wall time, Aux Wall time were examined, as well as correctness.
- NORMAL Find mode, Immediate mode
  - HIP backend, 93 known FP32 + 93 known FP16 configs - :heavy_check_mark: _NO REGRESSIONS_
  - OCL backend, 93 known FP32 configs  - :heavy_check_mark: _NO REGRESSIONS_
- NORMAL, FAST, HYBRID, FAST_HYBRID, DYNAMIC_HYBRID Find Modes
  - HIP backend, 93 _unknown_ FP32 configs - :heavy_check_mark: _NO REGRESSIONS_


atamazov:
All local testing done.

include/miopen/miopen.h
```diff
@@ -1824,6 +1824,69 @@ MIOPEN_EXPORT miopenStatus_t miopenConvolutionBackwardBias(miopenHandle_t handle
                                                            const miopenTensorDescriptor_t dbDesc,
                                                            void* db);
 
+/* Begin of Find Mode API */
+/*! @brief Version of the Find Mode API for convolutions.
+*
+* The Find Mode API is experimental and therefore subject to change.
```
JehandadKhan, line 1830: @atamazov We agreed that this API call would not be public since we do not wish to support this API in the future and is a stop gap arrangement for MIGraphX

JehandadKhan (changes requested):

include/miopen/miopen.h
atamazov: I will make it. However I would appreciate if you clearly explain the reasons. What should we be afraid of?

test/find_db.cpp
```diff
@@ -61,6 +61,7 @@ struct FindDbTest : test_driver
 
     FindDbTest()
     {
+        filter.findMode.Set(FindMode::Values::Normal);
```
atamazov, line 64: @pfultz2 Use of internal interface

include/miopen/miopen.h
atamazov: @JehandadKhan Done

atamazov: > we do not wish to support this API

I thought we _want_ to have means of controlling the Find Mode in our API. Of course, we do not want to support this specific public interface, because it is intermittent.

That is why I've introduced the `MIOPEN_API_VERSION_CONVOLUTION_FIND_MODE` macro. Current public interface is experimental, so we are free to change it. And if we change it, we will bump the API_VERSION macro. Or we can remove the public interface and undefine the macro. This way the application can easily find out if Find Mode API is supported in the library, and how it is supported.

Our only commitment is to keep the API_VERSION macro up to date.

driver/conv_driver.hpp
```diff
@@ -514,6 +511,7 @@ int ConvDriver<Tgpu, Tref>::GetandSetData()
                                           conv_strides.data(),
                                           conv_dilations.data(),
                                           mode);
+        miopenSetConvolutionFindMode(warmupConvDesc, miopenConvolutionFindModeNormal);
```
atamazov, line 514: Use of experimental API

JehandadKhan (approved):

src/include/miopen/miopen_internal.h
```diff
@@ -0,0 +1,114 @@
...
+#endif
+
+/* Begin of Find Mode API */
+
+/*! @brief Version of the Find Mode API for convolutions.
+*
+* The Find Mode API is experimental and therefore subject to change.
+* This macro allows applications to adapt to the future.
+*/
+#define MIOPEN_API_VERSION_CONVOLUTION_FIND_MODE 1
```
pfultz2, line 48: This should be a function, as this macro will never be accessible outside of miopen(as this header is not installed).

src/include/miopen/miopen_internal.h
```diff
@@ -0,0 +1,114 @@
...
+* times but better GPU performance.
+*/
+typedef enum {
+    miopenConvolutionFindModeNormal        = 1,                         /*!< Normal mode */
+    miopenConvolutionFindModeFast          = 2,                         /*!< Fast mode */
+    miopenConvolutionFindModeHybrid        = 3,                         /*!< Hybrid mode */
+    miopenConvolutionFindModeFastHybrid    = 4,                         /*!< Fast Hybrid mode */
+    miopenConvolutionFindModeDynamicHybrid = 5,                         /*!< Dynamic Hybrid mode */
+    miopenConvolutionFindModeDefault = miopenConvolutionFindModeHybrid, /*!< Default setting */
+} miopenConvolutionFindMode_t;
```
pfultz2, line 77: Can we define this as an `int` and use constants instead?

src/include/miopen/miopen_internal.h
atamazov: This header is intended to hold APIs that we do not want to expose to the users. However, the contents should be _as if_ it is a public API, by design. This allows us to disclose API at any moment by simple moving text from this file to the public header.

[Notice] I still do not fully agree with reasons of hiding this experimental API (see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/489#discussion_r500558737). The cost of support is zero because we are free to discontinue support of any experimental API at any given moment (provided that we do not forget to #undef the API_VERSION macro).

atamazov: @pfultz2 However please let me know if you need a function (instead of a macro) for your needs.

src/include/miopen/miopen_internal.h
atamazov: Hmm, I do not think we have constants in C. Or you mean something else?

DrizztDoUrden (dismissed):

src/include/miopen/miopen_internal.h
pfultz2: Then at least change the function to accept an `int` instead of `miopenConvolutionFindMode_t`. This way I only need to define the function in migraphx.

src/include/miopen/miopen_internal.h
pfultz2: Then we should just remove this macro as no one is going to use it(since its not available to anyone). We are probably just going to check for this function using [`check_symbol_exists`](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html) in CMake.

atamazov: I do not think so, because I would like to follow the design principle mentioned above:

> This header is intended to hold APIs that we do not want to expose to the users. However, the contents should be as if it is a public API, by design. This allows us to disclose API at any moment by simple moving text from this file to the public header.

Also I do not want to redo the work (even partially) when/if we decide to disclose this API.

src/include/miopen/miopen_internal.h
atamazov: Let me add another simple wrapper function that accepts int for you.

atamazov: BTW alternatively, I can move this stuff back t othe public API but hide it within
```C++
#ifdef MIOPEN_ENABLE_FIND_MODE_API
...
#endif
```
So only users that write:
```C++
#define MIOPEN_ENABLE_FIND_MODE_API
#include <miopen.h>
```
will be able to use this functionality.

Please let me know if you prefer this way.

src/include/miopen/miopen_internal.h
atamazov: @pfultz2 
> We are probably just going to check for this function using check_symbol_exists in CMake.

Hmm... AFAICS it checks header files, so it will fail unless we expose the function in the public header.

atamazov: So perhaps you would prefer the approach [shown here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/489#discussion_r502062038).

atamazov: @pfultz2 I am waiting for feedback from you.

atamazov:
BTW 308d389 has passed all CI checks. I do not know why this page doesn't show this.

src/include/miopen/miopen_internal.h
pfultz2: But I should still be able to use `check_cxx_source_compiles` in cmake.

pfultz2: > Also I do not want to redo the work (even partially) when/if we decide to disclose this API.

I dont think we want to expose the API ever. We should just implement find 2.0.

src/include/miopen/miopen_internal.h
pfultz2: > BTW alternatively, I can move this stuff back to the public API but hide it within

No, I dont think its good to have this in the header.

pfultz2:
Will normal find mode be enabled when the user passes `true` for exhaustiveSearch`?

src/ocl/convolutionocl.cpp
```diff
@@ -768,7 +768,6 @@ void ConvolutionDescriptor::FindConvFwdAlgorithm(Handle& handle,
 
     std::vector<PerfField> perf_db;
 
-    const miopen::FindMode fm(ctx);
```
pfultz2, line 771: This shouldn't be removed. The `findMode` should be constructed locally from the find mode in the convolution descriptor so it can be overridden when the user pass `true` for `exhaustiveSearch`.

atamazov:
> Will normal find mode be enabled when the user passes `true` for exhaustiveSearch`?

No

atamazov:
Let me try to support this. Earlier that wasn't possible.

atamazov:
Well, this is not possible unless `exhaustiveSearch` is passed to the GWSS calls.

atamazov:
@pfultz2 Right now, in order to use `exhaustiveSearch`, the user should explicitly set Find Mode to NORMAL in the environment or via the API calls provided in this PR. Maybe it is worth to declare this flag deprecated.

Then [we can implement similar API for Search Mode](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/336#issuecomment-704529225) (now MIOPEN_FIND_ENFORCE env var)m and that API will be able to enforce Find Mode properly. What do think?

src/include/miopen/miopen_internal.h
atamazov: Ok

src/include/miopen/miopen_internal.h
atamazov: > I dont think we want to expose the API ever. We should just implement find 2.0.

Yes. Unless users ask us to provide them with such a function before 2.0. So who knows?

atamazov: @pfultz2 
> But I should still be able to use `check_cxx_source_compiles` in cmake.

Okay, if this works for you. So I am going to make a C++ function.

atamazov:
[off-topic] @daniellowell @pfultz2 Right now, after committing f9d49b5, github shows "All checks have passed" but testing didn't even started yet. Looks like github WebUI is one step behind our CI.

atamazov:
[off-topic] Aha, we just have a time gap. After the commit, the github shows the previous status until the next round of testing begins. It can be up to 20 minutes long.

src/include/miopen/miopen_internal.h
atamazov: Done in f9d49b5.

atamazov:
All, CI tests passed 16 days ago (github's indication is wrong again). Let's do something if we need this.

src/include/miopen/miopen_internal.h
```diff
@@ -0,0 +1,114 @@
...
+*
+* Note that the default Find Mode is set by the MIOPEN_FINE_MODE environment variable,
+* if it is set. If unset, the default is as specified by miopenConvolutionFindModeDefault.
+*
+* @param convDesc   Convolution layer descriptor (input)
+* @param findMode   Find Mode of convDesc (input)
+* @return           miopenStatus_t
+*/
+MIOPEN_EXPORT miopenStatus_t miopenSetConvolutionFindMode(miopenConvolutionDescriptor_t convDesc,
+                                                          miopenConvolutionFindMode_t findMode);
```
pfultz2, line 93: The API still takes an enum, we need a function to take an `int` instead.

src/convolution_api.cpp
```diff
@@ -90,6 +94,41 @@ extern "C" miopenStatus_t miopenSetConvolutionGroupCount(miopenConvolutionDescri
...
+    MIOPEN_LOG_FUNCTION(convDesc, findMode);
+    return miopen::try_([&] {
+        miopen::deref(findMode) =
+            static_cast<miopenConvolutionFindMode_t>(miopen::deref(convDesc).findMode.Get());
+    });
+}
+
+// Hidden C++ functions for MIGraphX.
+namespace miopen {
+miopenStatus_t SetConvolutionFindMode(miopenConvolutionDescriptor_t convDesc, int findMode)
```
pfultz2, line 118: This should be declared with extern `C`.

src/convolution_api.cpp
```diff
@@ -90,6 +94,41 @@ extern "C" miopenStatus_t miopenSetConvolutionGroupCount(miopenConvolutionDescri
...
+{
+    return miopen::try_([&] {
+        miopen::deref(convDesc).findMode.Set(static_cast<miopen::FindMode::Values>(findMode));
+    });
+}
+miopenStatus_t GetConvolutionFindMode(miopenConvolutionDescriptor_t convDesc, int* findMode)
+{
+    return miopen::try_([&] {
+        miopen::deref(findMode) = static_cast<int>(miopen::deref(convDesc).findMode.Get());
+    });
```
atamazov, line 128: @pfultz2 These two functions are for you.

driver/conv_driver.hpp
```diff
@@ -514,6 +516,10 @@ int ConvDriver<Tgpu, Tref>::GetandSetData()
...
+        miopen::SetConvolutionFindMode(
+            warmupConvDesc,
+            static_cast<int>(miopenConvolutionFindModeNormal)); // Repeat via hidden API.
```
atamazov, line 522: @pfultz2 Example of use.

src/include/miopen/miopen_internal.h
atamazov: Don't care, these functions are not for you.

src/convolution_api.cpp
atamazov: @pfultz2 You need C functions?

atamazov: @pfultz2 
> But I should still be able to use `check_cxx_source_compiles` in cmake.

I thought you need C++, see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/489#discussion_r502653632

src/ocl/convolutionocl.cpp
atamazov: > so it can be overridden when the user pass `true` for `exhaustiveSearch`.

Not feasible without API change. See https://github.com/ROCmSoftwarePlatform/MIOpen/pull/489#issuecomment-706332733 and https://github.com/ROCmSoftwarePlatform/MIOpen/pull/489#issuecomment-706382083.

atamazov:
@pfultz2 Hidden functions changed from C++ to C. Prototype:
```C++
miopenStatus_t miopenHiddenSetConvolutionFindMode(
    miopenConvolutionDescriptor_t convDesc,
    int findMode);
```

src/convolution_api.cpp
atamazov: Done.

pfultz2 (dismissed):

JehandadKhan (approved):

atamazov:
@pfultz2 @JehandadKhan Merge conflicts resolved, no changes. Please re-approve.

JehandadKhan (approved):

atamazov:
@daniellowell All CI checks passed, so this can be merged now.

pfultz2 (approved):

atamazov:
@daniellowell Ping. Or I can merge this myself.

atamazov:
Ping

DrizztDoUrden (approved):

driver/conv_driver.hpp
```diff
@@ -514,6 +515,10 @@ int ConvDriver<Tgpu, Tref>::GetandSetData()
                                           conv_strides.data(),
                                           conv_dilations.data(),
                                           mode);
+        miopenSetConvolutionFindMode(warmupConvDesc, miopenConvolutionFindModeNormal);
+        miopenHiddenSetConvolutionFindMode(
+            warmupConvDesc,
+            static_cast<int>(miopenConvolutionFindModeNormal)); // Repeat via hidden API.
```
atamazov, line 521: Use of hidden API.

src/include/miopen/miopen_internal.h
hgaspar: In terms of a public API, I recommend that it should be more abstract, and not specific to convolutions. The abstraction will also provide flexibility moving forwards, and should have arguments that are conceptually clear. 
Example:
SetAlgoSearchLevel(  some uchar typedef with values right now: exhaustiveSearch=1,  ... , minimizeCompilation = 5 ) .  the params roughly correspond to the current find modes. I would avoid enums, and favor typedefs to uchar for this kind of thing, for forward and backward compatibility at the API level at least.

This API should be public, as different clients to miOpen have different requirements, and would want to control this programmatically based on the particular application (network+dataset).

atamazov: @hgaspar Good idea, thanks. However, that would be overdesign right now. We do not plan to add Find API for other primitives and do not foresee the need for it. Current Find Modes are only applicable for convolutions and can only be explained (documented) in convolution-specific terms.

test/find_db.cpp
atamazov: But it should be Hybrid mode in this test.

atamazov: But it should be Hybrid mode in this test.

PR #490, opened 2020-10-06
Upgrade to cppcheck 2.2

pfultz2:


src/ocl/convolutionocl.cpp
```diff
@@ -671,28 +671,28 @@ static void DirConvFindCore(Handle& handle,
...
+        {
+            ConvolutionUserBuffers bufs(workSpace, workSpaceSize);
+            bufs.SetFwd(x, w, y);
+            const auto all = conv.FindDataDirectSolutions(
+                handle, xDesc, wDesc, yDesc, exhaustiveSearch, true, bufs, invoke_ctx);
+            PrecompileSolutions(handle, all);
+            const auto algorithm_name = AlgorithmName{"miopenConvolutionFwdAlgoDirect"};
+            EvaluateInvokers(handle, all, algorithm_name, network_config, invoke_ctx, record);
+        }
+        // Implicit GEMM algo
```
atamazov, line 686: Does cppcheck 2.2 require this change? If yes, then please disable related check and roll this back.

src/ocl/convolutionocl.cpp
```diff
@@ -2172,26 +2173,28 @@ void ConvolutionDescriptor::FindConvBwdDataAlgorithm(Handle& handle,
...
+                    ConvolutionUserBuffers bufs(workSpace, workSpaceSize);
+                    bufs.SetBwd(dx, w, dy);
+                    const auto all = FindDataDirectSolutions(
+                        handle, dxDesc, wDesc, dyDesc, exhaustiveSearch, false, bufs, invoke_ctx);
+                    const auto algorithm_name = AlgorithmName{"miopenConvolutionBwdDataAlgoDirect"};
+                    PrecompileSolutions(handle, all);
+                    EvaluateInvokers(
+                        handle, all, algorithm_name, network_config, invoke_ctx, record);
+                }
+                // Implicit GEMM algo
```
atamazov, line 2186: Ditto.

src/solver/conv_ocl_dir2D_bwdWrW_2.cpp
```diff
@@ -276,23 +276,6 @@ bool PerformanceConfigConvOclBwdWrw2<N_BATCH_LOOPS>::IsValid(const ConvolutionCo
         return false;
     }
 
-    // For group config, reducing search space
```
atamazov, line 279: Dead code?

src/solver/conv_ocl_dir2Dfwdgen.cpp
```diff
@@ -153,7 +153,7 @@ ConvSolution ConvOclDirectFwdGen::GetSolution(const ConvolutionContext& params)
     n_outs = std::min(n_outs, params.n_outputs);
     n_ins  = std::min(n_ins, params.batch_sz);
 
-    n_out_stacks   = (n_outs * n_out_stacks <= params.n_outputs) ? n_out_stacks : 1;
+    n_out_stacks   = 1;
```
atamazov, line 156: [Note] This line can be removed ;)

atamazov (changes requested):

src/ocl/convolutionocl.cpp
pfultz2: This check was previously disabled due to too many false positives. Cppcheck 2.2 no longer has FPs in miopen, so I re-enabled it. This is not a FP. There is no point in checking `use_winograd_only` twice when a an unamed scope should be used.

src/solver/conv_ocl_dir2D_bwdWrW_2.cpp
pfultz2: This code doesn't do anything. The condition `(n_output_channels_per_group % 1 == 0 && n_out_channels_per_tile % 1 == 0)` is always true because `x % 1` is always 0. This makes the early return never reached. Beyond this early return, there are no other side effects, so all this can be removed.

Please double check that these assumptions are correct.

src/ocl/convolutionocl.cpp
atamazov: There is a point. I do not want extra indenting, so I combined scope with `if`. Do you insist?

src/solver/conv_ocl_dir2D_bwdWrW_2.cpp
atamazov: Kill it

src/ocl/convolutionocl.cpp
atamazov: [Recommendation]

src/ocl/convolutionocl.cpp
atamazov: [Recommendation]

atamazov (approved):
LGTM but I would appreciate if you still follow two [Recommendations]

src/ocl/convolutionocl.cpp
pfultz2: An unnamed scope makes the intention clearer. If you want to avoid indentation and code duplication then a local lambda could be used:

```cpp
    auto eval = [&](auto find, auto name) {
        ConvolutionUserBuffers bufs(workSpace, workSpaceSize);
        bufs.SetFwd(x, w, y);
        const auto all = find(
            conv, handle, xDesc, wDesc, yDesc, exhaustiveSearch, true, bufs, invoke_ctx);
        PrecompileSolutions(handle, all);
        const auto algorithm_name = AlgorithmName{name};
        EvaluateInvokers(handle, all, algorithm_name, network_config, invoke_ctx, record);
    };

    if(!use_winograd_only)
    {
        // Direct algo
        eval(std::mem_fn(&ConvolutionDescriptor::FindDataDirectSolutions), "miopenConvolutionFwdAlgoDirect");
        // Implicit GEMM algo
        eval(std::mem_fn(&ConvolutionDescriptor::FindDataImplicitGemmSolutions), "miopenConvolutionFwdAlgoImplicitGEMM");
    }
```

atamazov: Thanks for the advice, Paul.

JehandadKhan (approved):

daniellowell:
This PR still has issues.

pfultz2:
I merged from develop. When I run this locally I dont see the error.

atamazov:

junliume:
@atamazov  @JehandadKhan  @shurale-nkn 
Shall we revive this PR as a follow up for #1111 ? 
Would it be a good practice for new onboards please?

JehandadKhan:
> Shall we revive this PR as a follow up for #1111 ?

Yes
> Would it be a good practice for new onboards please?

I think its a good idea.

junliume:
@technicalgrp89 could you please take this PR further and work on it? Thanks!

junliume:
replaced with more recent changes cppcheck 2.9 and https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1807

PR #493, opened 2020-10-07
Batchnorm gfx10 changes

muralinr:
Adding Batchnorm gfx1030 changes for bring up


daniellowell:
@muralinr This actually needs the patch to enable gfx1030.

muralinr:
> @muralinr This actually needs the patch to enable gfx1030.

yes Daniel. I will make changes to avoid this requirement.

src/kernels/MIOpenBatchNormActivBwdPerAct.cl
```diff
@@ -33,9 +33,9 @@
 #pragma clang diagnostic ignored "-Wsometimes-uninitialized"
 #endif
 
-/*#ifdef __AMDGCN__
+#if(MIO_BN_GPU_ARCH == gfx1030)
```
daniellowell, line 36: Instead of this I would use:
```
#if(MIO_BN_GFX1030 == 1 && __AMDGCN__)
#undef __AMDGCN__
#endif
```

src/ocl/batchnormocl.cpp
```diff
@@ -290,7 +290,7 @@ void BatchNormForwardTraining(Handle& handle,
                         " -DMIO_BN_LDSGCN_SIZE=" + std::to_string(ldsgcn) + " -DMIO_BN_N=" +
                         std::to_string(n) + " -DMIO_BN_GRP0=" + std::to_string(xlocalsize) +
                         " -DMIO_BN_GRP1=" + std::to_string(ylocalsize) + " -DMIO_BN_GRP2=" +
-                        std::to_string(zlocalsize);
+                        std::to_string(zlocalsize) + " -DMIO_BN_GPU_ARCH=" + handle.GetDeviceName();
```
daniellowell, line 293: Here set this first:
```
if(handle.GetDeviceName() == gfx1030)
    parms += " -DMIO_BN_GFX1030=1";
```

src/kernels/MIOpenBatchNormActivBwdPerAct.cl
daniellowell: At the header put:
```
#ifndef MIO_BN_GFX1030
#define MIO_BN_GFX1030 0
#endif 
```

daniellowell (approved):

src/kernels/MIOpenBatchNormActivBwdPerAct.cl
```diff
@@ -33,9 +33,9 @@
 #pragma clang diagnostic ignored "-Wsometimes-uninitialized"
 #endif
 
-/*#ifdef __AMDGCN__
+#if(MIO_BN_GFX1030 == 1 && __AMDGCN__)
 #undef __AMDGCN__
```
atamazov, line 37: This change leads to undefined behavior. We shouldn't undef predefined macros.

PR #496, opened 2020-10-08
Added arch support for gfx10

daniellowell:


src/include/miopen/device_name.hpp
```diff
@@ -55,7 +53,7 @@ std::string inline GetDeviceNameFromMap(const std::string& name)
         {"gfx900", "gfx900"},
         {"gfx901", "gfx900"},
         {"gfx906", "gfx906"},
-        {"gfx908", "gfx908"},
+        {"10.3.0 Sienna_Cichlid 18", "gfx908"},
```
muralinr, line 56: Do we need to mention gfx1030 instead of "gfx908" in "device_name_map" here?

daniellowell: Arg, merge issues. This is not right. Let me fix this...

daniellowell:
This PR is corrupted.

PR #499, opened 2020-10-09
Passing HIP compiler version to composable kernel

asroy:
Passing HIP compiler version number to composable kernel, to fix this issue due to change intrinsic function:
http://ontrack-internal.amd.com/browse/SWDEV-255582

The hip-clang version that will have the change in intrinsic is 3.10.20405

### DIFF
--- a/src/include/miopen/hip_build_utils.hpp
+++ b/src/include/miopen/hip_build_utils.hpp
@@ -26,6 +26,7 @@
 #ifndef MIOPEN_GUARD_MLOPEN_HIP_BUILD_UTILS_HPP
 #define MIOPEN_GUARD_MLOPEN_HIP_BUILD_UTILS_HPP
 
+#include <miopen/config.h>
 #include <miopen/kernel.hpp>
 #include <miopen/tmp_dir.hpp>
 #include <miopen/write_file.hpp>
@@ -42,13 +43,14 @@
 #define HIP_PACKAGE_VERSION_PATCH 0
 #endif
 
-// 3 decimal digits for major and minor, 5 digits for patch number.
+// 3 decimal digits for major and minor, 6 digits for patch number.
+// Max number is 999,999,999999 == 0xE8,D4A5,0FFF that fits into 64-bit math.
 #if HIP_PACKAGE_VERSION_MAJOR > 999 || HIP_PACKAGE_VERSION_MAJOR > 999 || \
-    HIP_PACKAGE_VERSION_PATCH > 99999
+    HIP_PACKAGE_VERSION_PATCH > 999999
 #error "Too big HIP version number(s)"
 #endif
-#define HIP_PACKAGE_VERSION_FLAT                                               \
-    ((HIP_PACKAGE_VERSION_MAJOR * 1000 + HIP_PACKAGE_VERSION_MINOR) * 100000 + \
+#define HIP_PACKAGE_VERSION_FLAT                                                   \
+    ((HIP_PACKAGE_VERSION_MAJOR * 1000ULL + HIP_PACKAGE_VERSION_MINOR) * 1000000 + \
      HIP_PACKAGE_VERSION_PATCH)
 
 namespace miopen {
--- a/src/include/miopen/hip_build_utils.hpp
+++ b/src/include/miopen/hip_build_utils.hpp
@@ -32,6 +32,25 @@
 #include <boost/optional.hpp>
 #include <string>
 
+#ifndef HIP_PACKAGE_VERSION_MAJOR
+#define HIP_PACKAGE_VERSION_MAJOR 0
+#endif
+#ifndef HIP_PACKAGE_VERSION_MINOR
+#define HIP_PACKAGE_VERSION_MINOR 0
+#endif
+#ifndef HIP_PACKAGE_VERSION_PATCH
+#define HIP_PACKAGE_VERSION_PATCH 0
+#endif
+
+// 3 decimal digits for major and minor, 5 digits for patch number.
+#if HIP_PACKAGE_VERSION_MAJOR > 999 || HIP_PACKAGE_VERSION_MAJOR > 999 || \
+    HIP_PACKAGE_VERSION_PATCH > 99999
+#error "Too big HIP version number(s)"
+#endif
+#define HIP_PACKAGE_VERSION_FLAT                                               \
+    ((HIP_PACKAGE_VERSION_MAJOR * 1000 + HIP_PACKAGE_VERSION_MINOR) * 100000 + \
+     HIP_PACKAGE_VERSION_PATCH)
+
 namespace miopen {
 
 boost::filesystem::path HipBuild(boost::optional<miopen::TmpDir>& tmp_dir,
--- a/src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
@@ -826,11 +826,7 @@ ConvHipImplicitGemmWrwV4R4Xdlops::GetPerformanceConfig(const ConvolutionContext&
     MIOPEN_LOG_I(config.ToString());
     return config;
 }
-static inline bool is_support_amd_buffer_atomic_add(const ConvolutionContext& ctx)
-{
-    const auto device_name = ctx.GetStream().GetDeviceName();
-    return StartsWith(device_name, "gfx908");
-}
+
 ConvSolution ConvHipImplicitGemmWrwV4R4Xdlops::GetSolution(
     const ConvolutionContext& ctx, const PerformanceImplicitGemmWrwV4R4Xdlops& config, bool) const
 {
@@ -889,7 +885,6 @@ ConvSolution ConvHipImplicitGemmWrwV4R4Xdlops::GetSolution(
 
     // clang-format off
     construction_parameters.comp_options =
-        std::string(" -std=c++14 ") +
         std::string(" -DCK_PARAM_PROBLEM_G=") + std::to_string(ConvolutionContextInterpreter::GetGroupCountG(ctx)) +
         std::string(" -DCK_PARAM_PROBLEM_N=") + std::to_string(ConvolutionContextInterpreter::GetBatchN(ctx)) +
         std::string(" -DCK_PARAM_PROBLEM_K=") + std::to_string(ConvolutionContextInterpreter::GetOutputChannelK(ctx)) +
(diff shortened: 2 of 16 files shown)


daniellowell:
>The hip-clang version that will have the change in intrinsic is 3.9.20354.

Are you sure about this?

BTW, did you test in the docker?

asroy:
> > The hip-clang version that will have the change in intrinsic is 3.9.20354.
> 
> Are you sure about this?
> 
> BTW, did you test in the docker?

I tested with the docker, the hip-clang version is 3.9.20354

I'm not sure it's the formal version number that will received the compiler change. I'm confirming with compiler team in the JIRA

searlmc1:
> > > The hip-clang version that will have the change in intrinsic is 3.9.20354.
> > 
> > 
> > Are you sure about this?
> > BTW, did you test in the docker?
> 
> I tested with the docker, the hip-clang version is 3.9.20354
> 
> I'm not sure it's the formal version number that will received the compiler change. I'm confirming with compiler team in the JIRA

I do not think it is the correct version. I'm checking on what it should be. I will reply in the jira

asroy:
> I do not think it is the correct version. I'm checking on what it should be. I will reply in the jira

Updated to 3.10.20405 according to the JIRA

atamazov (changes requested):
Let's make the flat HIP package version number available everywhere, at build time.

src/solver/implicitgemm_util.hpp
```diff
@@ -768,6 +768,37 @@ int amd_lds_write_max_length()
...
+        const auto hip_version = HipCompilerVersion();
+
+        auto hip_version_major = std::to_string(hip_version.major);
+        auto hip_version_minor = std::to_string(hip_version.minor);
+        auto hip_version_patch = std::to_string(hip_version.patch);
+
+        hip_version_minor = std::string(2 - hip_version_minor.length(), '0') + hip_version_minor;
+        hip_version_patch = std::string(5 - hip_version_patch.length(), '0') + hip_version_patch;
+
+        compiler_flag += std::string(" -DCK_HIP_VERSION=") + hip_version_major + hip_version_minor +
```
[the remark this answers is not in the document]
atamazov: #include <miopen/stringutils.hpp> for stringize macro.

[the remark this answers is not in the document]
asroy: I got negative value for HIP_PACKAGE_VERSION_FLAT, it seems xxx,xxx,xxxxxx will go out of range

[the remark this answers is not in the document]
atamazov: It seems like 32-bit math is still used somewhere on the way, most likely in the clang command line.

Patch number could be very high, so I recommend using the following convention:
```C++
// 2 decimal digits for major and minor, 6 digits for patch number.
// Max 32 bit signed is 21,47,483647, so major should be limited to
// max 20 to avoid overflow. Max flat version number is
// Max number is 20,99,999999.
#if HIP_PACKAGE_VERSION_MAJOR > 20 || HIP_PACKAGE_VERSION_MAJOR > 99 || \
    HIP_PACKAGE_VERSION_PATCH > 999999
#error "Too big HIP version number(s)"
#endif
#define HIP_PACKAGE_VERSION_FLAT                                                \
    ((HIP_PACKAGE_VERSION_MAJOR * 100 + HIP_PACKAGE_VERSION_MINOR) * 1000000 + \
     HIP_PACKAGE_VERSION_PATCH)
```
This should work correctly with 32-bit signed math.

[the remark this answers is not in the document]
atamazov: oops I am wrong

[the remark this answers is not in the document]
atamazov: Overflow happens here: `((3 * 1000 + 7) * 100000 + 20315)`. How to fix:
- Replace `1000` with `1000ULL` in the HIP_PACKAGE_VERSION_FLAT macro so overflow wouldn't happen.
- Keep using `std::to_string` to pass a string literal instead of expression to the HIP compiler.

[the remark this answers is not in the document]
atamazov: No need to change from 3-3-6 digits to 2-2-6.

src/include/miopen/hip_build_utils.hpp
```diff
@@ -32,6 +32,25 @@
 #include <boost/optional.hpp>
 #include <string>
 
+#ifndef HIP_PACKAGE_VERSION_MAJOR
```
atamazov, line 35: Please include config.h

src/include/miopen/hip_build_utils.hpp
```diff
@@ -32,6 +32,25 @@
...
+#define HIP_PACKAGE_VERSION_PATCH 0
+#endif
+
+// 3 decimal digits for major and minor, 5 digits for patch number.
+#if HIP_PACKAGE_VERSION_MAJOR > 999 || HIP_PACKAGE_VERSION_MAJOR > 999 || \
+    HIP_PACKAGE_VERSION_PATCH > 99999
+#error "Too big HIP version number(s)"
+#endif
+#define HIP_PACKAGE_VERSION_FLAT                                               \
+    ((HIP_PACKAGE_VERSION_MAJOR * 1000 + HIP_PACKAGE_VERSION_MINOR) * 100000 + \
```
atamazov, line 51: See https://github.com/ROCmSoftwarePlatform/MIOpen/pull/499#discussion_r502716865

src/solver/implicitgemm_util.hpp
```diff
@@ -18,14 +18,9 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_BLOCK_SYNC_LDS_WITHOUT_SY
 #define WORKAROUND_SWDEV_229564 1
 // workaround for buffer load/store fp16/bfp16 intrinsic bug
 #define WORKAROUND_SWDEV_231101 1
-// LLVM xdlops instrinsic will do unnecessey VGRP <--> AGPR movement, and result in
-// register spill, for bfloat16 datatype, when doing wave-wise GEMM larger than 64x64
-#define WORKAROUND_SWDEV_240356 1
```
atamazov, line 23: Why this workaround removed? Even if recent ROCm have been fixed, older ROCm still have issue. So latest MIOpen installed on existing old ROCm may have issues.

src/solver/implicitgemm_util.hpp
```diff
@@ -18,14 +18,9 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_BLOCK_SYNC_LDS_WITHOUT_SY
...
 // due to compiler bug, iGEMM xdlops kernels fail verification in some cases, if using "-O3" flag,
 // (but will pass verification with "-O1" flag)
 #define WORKAROUND_SWDEV_251757 1
-// workaround failure of ConvHipImplicitGemmV4R4GenWrWXdlops with vector load
-#define WORKAROUND_ISSUE_2532 1
```
atamazov, line 28: Ditto

atamazov (changes requested):

src/solver/implicitgemm_util.hpp
```diff
@@ -688,7 +683,7 @@ static inline bool use_amd_inline_asm(const ConvolutionContext& ctx)
     return !miopen::IsDisabled(MIOPEN_DEBUG_IMPLICIT_GEMM_NON_XDLOPS_INLINE_ASM{});
 }
 
-static inline bool support_amd_buffer_atomic_add(const ConvolutionContext& ctx)
+static inline bool support_amd_buffer_atomic_fadd(const ConvolutionContext& ctx)
 {
     const auto device_name = ctx.GetStream().GetDeviceName();
     return StartsWith(device_name, "gfx908") && ctx.IsFp32();
```
ltqin, line 689: does it must call ctx.IsFp32()? sometime data is fp16, but  it can use fp32 atomic add by using workspace

src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
```diff
@@ -931,8 +926,7 @@ ConvSolution ConvHipImplicitGemmWrwV4R4Xdlops::GetSolution(
         std::string(" -DCK_USE_AMD_XDLOPS_INLINE_ASM=") + std::to_string(miopen::IsEnabled(MIOPEN_DEBUG_IMPLICIT_GEMM_XDLOPS_INLINE_ASM{}) ? 1 : 0) +
         std::string(" -DCK_USE_AMD_XDLOPS_EMULATE=") + (miopen::IsEnabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_XDLOPS_EMULATE{}) ? '1' : '0') +
         std::string(" -DCK_BLOCK_SYNC_LDS_WITHOUT_SYNC_VMEM=") + (miopen::IsDisabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_BLOCK_SYNC_LDS_WITHOUT_SYNC_VMEM{}) ? '0' : '1') +
-        std::string(" -DCK_WORKAROUND_SWDEV_231101=") + std::to_string(WORKAROUND_SWDEV_231101) +
-        std::string(" -DCK_USE_AMD_BUFFER_ATOMIC_ADD=") + (is_support_amd_buffer_atomic_add(ctx) ? '1' : '0') +
+        get_ck_common_compiler_flag(ctx) +
```
ltqin, line 929: ctx.IsFp32 will return false, so can not directly replace this to  is_support_amd_buffer_atomic_add function

src/solver/implicitgemm_util.hpp
asroy: We fixed this issue by fixing our kernel implementation.

src/solver/implicitgemm_util.hpp
asroy: ```ConvHipImplicitGemmV4R4GenWrWXdlops``` has been removed

src/include/miopen/hip_build_utils.hpp
asroy: fixed

src/solver/implicitgemm_util.hpp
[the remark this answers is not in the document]
asroy: fixed. Thanks

src/include/miopen/hip_build_utils.hpp
asroy: fixed

src/solver/implicitgemm_util.hpp
asroy: fixed. Thanks for pointing out

src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
asroy: fixed. I removed ctx.IsFp32() check.

asroy:
@daniellowell The CI passed

src/include/miopen/hip_build_utils.hpp
daniellowell: @asroy Has this been addressed?
@atamazov You review is still blocking this PR. Are there additional changes needed?

atamazov (commented):
LGTM

atamazov (approved):
LGTM

atamazov:
This affects the pre-compiled binary cache. All HIP kernels built with certain version of the compiler will be skipped (cache miss) if the user's compiler has different version.

PR #505, opened 2020-10-10
Workaround for SWDEV-255735

shurale-nkn:
Now the assembler does a stronger check than before. Therefore, AMDHSA kernel xnack directive update are required.

`.amdhsa_reserve_xnack_mask` directive value (`1` or `0`) must match passed compiler argument (`-mxnack` or`-mno-xnack`) and curent GPU xnack mode (enabled or disabled). If the value does not match, the solver must be disabled.

Decided to disable xnack by default for asm kernels, since in most devices exclude APUs xnack disabled by default.

Full support for xnack requires the ability to detect the current GPU mode. This possibly can be done by using `Code Object Target Identification` which include xnack and ecc  mode ``amdgcn-amd-amdhsa--gfx906:sramecc-:xnack-``.

igemm kernels were updated using ``find ./ -type f -exec sed -i 's/.end_amdhsa_kernel/    \/\/xnack disabled by default for asm kernels\n    .amdhsa_reserve_xnack_mask 0\n.end_amdhsa_kernel/g' {} \;``

### DIFF
--- a/src/kernels/Conv_Winograd_v13_3_12_epilogue.inc
+++ b/src/kernels/Conv_Winograd_v13_3_12_epilogue.inc
@@ -32,6 +32,11 @@ workgroup_size_x = 512
 .amdgcn.next_free_sgpr = 101
 .amdgcn.next_free_vgpr = 128
 
+//xnack disabled by default for asm kernels
+__sgpr_reserve_vcc_default = 1
+__sgpr_reserve_xnack_default = 0
+__sgpr_reserve_flatscr_default = 0
+
 .rodata
 .p2align 6
 .amdhsa_kernel miopenSp3AsmConvRxSU
@@ -45,9 +50,9 @@ workgroup_size_x = 512
     .amdhsa_system_vgpr_workitem_id 0
     .amdhsa_next_free_vgpr .amdgcn.next_free_vgpr
     .amdhsa_next_free_sgpr .amdgcn.next_free_sgpr
-    .amdhsa_reserve_vcc 1
-    .amdhsa_reserve_xnack_mask 1
-    .amdhsa_reserve_flat_scratch 0
+    .amdhsa_reserve_vcc __sgpr_reserve_vcc_default
+    .amdhsa_reserve_xnack_mask __sgpr_reserve_xnack_default
+    .amdhsa_reserve_flat_scratch __sgpr_reserve_flatscr_default
     .amdhsa_ieee_mode 0
     .amdhsa_dx10_clamp 0
 .end_amdhsa_kernel
--- a/src/kernels/dynamic_igemm/igemm_bwd_gtc_dynamic.s
+++ b/src/kernels/dynamic_igemm/igemm_bwd_gtc_dynamic.s
@@ -1227,6 +1227,8 @@ L_igemm_v4r1_bwd_dynamic_end:
     .amdhsa_next_free_sgpr 72
     .amdhsa_ieee_mode 0
     .amdhsa_dx10_clamp 0
+    //xnack disabled by default for asm kernels
+    .amdhsa_reserve_xnack_mask 0
 .end_amdhsa_kernel
 ;----------------------------------------------------------
 ; starting of kernel igemm_bwd_gtc_bt128x128x16_tt8x8_gm2x4x4_gn2x4x4_ta1x1x1x2x4_16x1x1x16x1_tb1x1x1x2x4x1x1_16x1x1x16x1x1x1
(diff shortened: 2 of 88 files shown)


src/kernels/xform_bidirect_winograd_code.inc
```diff
@@ -297,6 +297,7 @@ gid_x = 2
 .SGPR_ALLOC x_TW_stride
 .SGPR_ALLOC x_TH_stride
 .SGPR_ALLOC addr_step
+.SGPR_RESERVE_VCC
```
atamazov, line 300: oops ;)

src/ocl/gcn_asm_utils.cpp
```diff
@@ -239,7 +244,10 @@ static void AmdgcnAssembleQuiet(const std::string& source, const std::string& pa
 #ifdef __linux__
     std::stringstream clang_stdout_unused;
     const auto clang_path = GetGcnAssemblerPath();
-    const auto args       = " -x assembler -target amdgcn--amdhsa " + params + " " + source +
+    const auto args       = " -x assembler -target amdgcn--amdhsa " +
```
atamazov, line 247: Changes is in this file are not necessary. Why:
- Currently the library supports only parts that have XNACK mode OFF by default
  - See https://llvm.org/docs/AMDGPUUsage.html#processors
- The default value of the target feature is automatically set to the default XNACK setting of the target.
  - https://llvm.org/docs/AMDGPUUsage.html#target-features says: "_The target features supported by each processor, and the default value used if not specified explicitly, is listed in AMDGPU Processors._"

Moreover, this breaks compatibility with older assemblers that do not support the `-m[no-]xnack` option.

I am going to fix this right away.

atamazov (changes requested):

src/kernels/Conv_Winograd_v13_3_12_epilogue.inc
```diff
@@ -32,6 +32,11 @@ workgroup_size_x = 512
 .amdgcn.next_free_sgpr = 101
 .amdgcn.next_free_vgpr = 128
 
+//xnack disabled by default for asm kernels
```
atamazov, line 35: The changes like this are not necessary, but let's remove these later, when time permits.

atamazov: I mean that it's enough to change only one line.

src/kernels/dynamic_igemm/igemm_bwd_gtc_dynamic.s
```diff
@@ -1227,6 +1227,8 @@ L_igemm_v4r1_bwd_dynamic_end:
     .amdhsa_next_free_sgpr 72
     .amdhsa_ieee_mode 0
     .amdhsa_dx10_clamp 0
+    //xnack disabled by default for asm kernels
+    .amdhsa_reserve_xnack_mask 0
```
atamazov, line 1231: [Future] Note that documentation says that the default of the `.amdhsa_reserve_xnack_mask` is Target Feature Specific, and, most likely, the changes like this are NOT required. Let's keep this change, however, for the sake of reliability. I think we will roll back such changes after upcoming release.

atamazov:
It seems like `.amdhsa_reserve_xnack_mask 0` is not working without explicit `-mno-xnack` option. Another bug that requires a workaround. Working on it...

src/kernels/dynamic_igemm/igemm_bwd_gtc_dynamic.s
shurale-nkn: directive ``.amdhsa_reserve_xnack_mask``  added for each asm kernel at the request of management.

atamazov: "management" is an abstract notion. Who?

atamazov (approved):
LGTM

src/ocl/gcn_asm_utils.cpp
```diff
@@ -239,9 +261,9 @@ static void AmdgcnAssembleQuiet(const std::string& source, const std::string& pa
 #ifdef __linux__
     std::stringstream clang_stdout_unused;
     const auto clang_path = GetGcnAssemblerPath();
-    const auto args       = " -x assembler -target amdgcn--amdhsa " + params + " " + source +
-                      " -o /dev/null" + // We do not need output file
-                      " 2>&1";          // Keep console clean from error messages.
+    const auto args = " -x assembler -target amdgcn--amdhsa " + GetXnackOption() + params + " " +
```
atamazov, line 264: Bug is here. Suggestion: insert ` + " " ` after `GetXnackOption()`

Issue #482, opened 2020-10-02
Convert MIOpenUtilKernels4.cl to dynamic


JehandadKhan:
`MIOpenUtilKernels4.cl` is not dynamic at the moment and is compiled fairly often. It needs to be converted to a dynamic kernel.

PR #507, opened 2020-10-11
Dynamic transpose kernels

zjing14:
- Convert transpose kernels in MIOpenUtilKernels4.cl to dynamic kernels (Issue #482)

### DIFF
--- a/src/ocl/convolutionocl.cpp
+++ b/src/ocl/convolutionocl.cpp
@@ -89,21 +89,12 @@ static inline bool IsAnyBufferBF16(const TensorDescriptor& xDesc,
 
 size_t GetKernelGlobalWorkDim(const KernelInvoke& kernel, int dim)
 {
-#if(MIOPEN_BACKEND_HIP)
     return kernel.gdims[dim];
-#else
-    return kernel.global_work_dim[dim];
-#endif
 }
 
 size_t GetKernelLocalWorkDim(const KernelInvoke& kernel, int dim)
 {
-#if(MIOPEN_BACKEND_HIP)
     return kernel.ldims[dim];
-#else
-    // sometimes local_work_dim = {0,0,0} look in issue #1724
-    return kernel.local_work_dim[dim];
-#endif
 }
 
 static inline void AddKernels(const Handle& handle,
--- a/src/ocl/convolutionocl.cpp
+++ b/src/ocl/convolutionocl.cpp
@@ -87,15 +87,9 @@ static inline bool IsAnyBufferBF16(const TensorDescriptor& xDesc,
 }
 #endif
 
-size_t GetKernelGlobalWorkDim(const KernelInvoke& kernel, int dim)
-{
-    return kernel.gdims[dim];
-}
+size_t GetKernelGlobalWorkDim(const KernelInvoke& kernel, int dim) { return kernel.gdims[dim]; }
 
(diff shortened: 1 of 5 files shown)


zjing14:
@JehandadKhan @daniellowell Updates: 
1) Experiments to identify the cause of failures in test_tensor_trans. First, I just keep a single kernel in the kernel file, still got failure. So, the failure is not due to call the wrong kernel. Second, if I ran the same multiple transposes individually with the same cache kernels, no issue. Thus, I think there is something wrong with the host side or test program.
2) The slow down problem is due to fluctuated performance of rocblas gemm with dynamic transpose. The performance is stable in the develop branch. But with dynamic transpose, the rocblas gemm performance drop significantly in some runs  That is weird.

src/ocl/utilocl.cpp
```diff
@@ -699,107 +699,119 @@ float transpose_NCHW2CNHW(const Handle& handle,
...
+        }
 
+        kernel_name += "_" + READ_TYPE;
+
+        auto&& kernels = handle.GetKernels(kernel_name, network_config);
+        if(!kernels.empty())
+        {
+            auto kernel = kernels.front();
+            kernel.ldims = {vld[0], vld[1], vld[2]};
+            kernel.gdims = {vgd[0], vgd[1], vgd[2]};
```
JehandadKhan, line 747: @zjing14 Here is the fix to the issue you were seeing.

zjing14: I see. Thanks.

daniellowell:
Coding errors:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/dynamic_UtilKernels4/9/pipeline

JehandadKhan:
@zjing14 Test failure in convolution

src/ocl/convolutionocl.cpp
```diff
@@ -87,24 +87,9 @@ static inline bool IsAnyBufferBF16(const TensorDescriptor& xDesc,
...
 
-size_t GetKernelGlobalWorkDim(const KernelInvoke& kernel, int dim)
-{
-#if(MIOPEN_BACKEND_HIP)
-    return kernel.gdims[dim];
-#else
-    return kernel.global_work_dim[dim];
-#endif
-}
+size_t GetKernelGlobalWorkDim(const KernelInvoke& kernel, int dim) { return kernel.gdims[dim]; }
```
daniellowell, line 90: What is this change for?

zjing14: Changed it for updating gdims for each run, correct? @JehandadKhan

JehandadKhan: That is correct, I updated the member names to make them uniform between both the HIP and the OpenCL backend

daniellowell: Curious that they were different in the first place. Are you sure there are no side effects (usages of the old style)?

JehandadKhan:
@zjing14 can you please fix the CI issues so we can get this in ?

zjing14:
> @zjing14 can you please fix the CI issues so we can get this in ?

The CI is still running without an issue, though it has red "X".

zjing14:
CI passed.

JehandadKhan (approved):

PR #510, opened 2020-10-12
fix bug about wrw atomic add

ltqin:
fix two bug in PerformanceImplicitGemmWrwV4R4Xdlops  class:
1 data_per_thread_copy_gemmn should be limited by GemmNPerBlock(not GemmKPerBlock)  in CalculateGemmBBlockCopyPerformanceParameters function 
2.The name of GemmBThreadCopyMoreGemmKPack should be  GemmBThreadCopyMoreGemmK

### DIFF
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -1928,7 +1928,7 @@ struct PerformanceImplicitGemmWrwV4R4Xdlops : Serializable<PerformanceImplicitGe
     int GemmNPerWave;
     int GemmKPack;
     bool GemmAThreadCopyMoreGemmK;
-    bool GemmBThreadCopyMoreGemmKPack;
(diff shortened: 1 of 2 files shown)


asroy:
@ltqin Does this PR change performance?

asroy (approved):

ltqin:
@asroy some case is better than before, most of case is similar

shaojiewang (approved):

shaojiewang:
> @asroy some cases are better than before, most of cases are similar

@ltqin Why are some cases better?

ltqin:
@shaojiewang GemmNPerBlock is normally bigger than GemmKPerBlock, so can read more data in N dimension.

asleepzzz (approved):

asroy:
@daniellowell CI has passed

PR #511, opened 2020-10-12
Relocate compiler flag for hip version into HipBuild()

asroy:
Relocate compiler flag for hip version into HipBuild(), so that it's not part of the key for kernel binary cache.

Also refactors some code.

----
This is to resolve the problem mentioned at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/499#issuecomment-707391991.

### DIFF
--- a/src/hip/hip_build_utils.cpp
+++ b/src/hip/hip_build_utils.cpp
@@ -335,4 +335,9 @@ bool external_tool_version_t::operator>=(const external_tool_version_t& rhs) con
         return false;
 }
 
+bool external_tool_version_t::operator<(const external_tool_version_t& rhs) const
+{
+    return rhs >= *this;
+}
+
 } // namespace miopen
--- a/src/hip/hip_build_utils.cpp
+++ b/src/hip/hip_build_utils.cpp
@@ -154,6 +154,10 @@ boost::filesystem::path HipBuild(boost::optional<TmpDir>& tmp_dir,
     }
 #endif
 
+    // hip version
+    params +=
+        std::string(" -DHIP_PACKAGE_VERSION_FLAT=") + std::to_string(HIP_PACKAGE_VERSION_FLAT);
+
     params += " ";
     auto bin_file = tmp_dir->path / (filename + ".o");
 
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -700,7 +700,8 @@ void BuildHip(const std::string& name,
         {
             auto raw = options                                 //
                        + " " + GetDebugCompilerOptionsInsert() //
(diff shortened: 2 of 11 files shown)


atamazov:
Description updated.

atamazov (commented):
Please note that there is another method. The compiler version can be passed in HipBuild(). This would not affect the binary cache key, i.e. works as if the compiler version macro is predefined by the compiler. This approach looks more attractive to me:
- The version, once supplied, could be used for various source code adjustments in the future.
- The binary cache is not affected at all. For example, HIP kernels built with >= 3.10.20405 will be found in the cache even if user's compiler is < 3.10. Right now this PR doesn't allow this.

asroy:
@atamazov I moved the hip version flag into HipBuild()

atamazov (approved):
Thanks @asroy

daniellowell:
@asroy @atamazov 
Macro not defined in the file:
```/var/jenkins/workspace/ibs_MIOpen_remove_ck_hip_version/src/comgr.cpp:704:74: error: use of undeclared identifier 'HIP_PACKAGE_VERSION_FLAT'

                       (" -DHIP_PACKAGE_VERSION_FLAT=") + std::to_string(HIP_PACKAGE_VERSION_FLAT);

                                                                         ^

/var/jenkins/workspace/ibs_MIOpen_remove_ck_hip_version/src/comgr.cpp:716:74: error: use of undeclared identifier 'HIP_PACKAGE_VERSION_FLAT'

                       (" -DHIP_PACKAGE_VERSION_FLAT=") + std::to_string(HIP_PACKAGE_VERSION_FLAT);

                                                                         ^

2 errors generated when compiling for gfx900.
```
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/remove_ck_hip_version/2/pipeline

atamazov (approved):

daniellowell:
@asroy This PR has a merge conflict now.

atamazov:
Conflict is due to my commit. Resolved.

atamazov (approved):

asroy:
@daniellowell This PR passed CI

src/hip/hip_build_utils.cpp
```diff
@@ -335,4 +339,9 @@ bool external_tool_version_t::operator>=(const external_tool_version_t& rhs) con
         return false;
 }
 
+bool external_tool_version_t::operator<(const external_tool_version_t& rhs) const
+{
+    return rhs >= *this;
```
atamazov, line 344: :bug: Bug here -- should be `>`. I'll fix it in my next PR.

PR #516, opened 2020-10-14
ConvolutionDescription:DirConvFindCore non-static 

alexandraBara:
For Fin/Tuna we need this function non-static.
Moved to ConvDescription for ease of use.

### DIFF
--- a/src/include/miopen/convolution.hpp
+++ b/src/include/miopen/convolution.hpp
@@ -33,6 +33,7 @@
 #include <miopen/solver_id.hpp>
 #include <miopen/names.hpp>
 #include <miopen/invoke_params.hpp>
+#include <miopen/db_record.hpp>
 
 #include <boost/any.hpp>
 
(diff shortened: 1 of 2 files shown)


atamazov:
How urgent is this?

I think you can use `ConvolutionDescriptor::FindConvFwdAlgorithm()`, right?

WRT `DirConvFindCore()` -- basically, this function should be eliminated (its contents moved into `ConvolutionDescriptor::FindConvFwdAlgorithm()`)

alexandraBara:
> How urgent is this?
> 
> I think you can use `ConvolutionDescriptor::FindConvFwdAlgorithm()`, right?
--> i will look into this function
> 
> WRT `DirConvFindCore()` -- basically, this function should be eliminated (its contents moved into `ConvolutionDescriptor::FindConvFwdAlgorithm()`)
--> we need this as soon as possible so we can progress with our tool.

atamazov:
> we need this as soon as possible so we can progress with our tool.

What do you plan to do with backward convolutions? There is no functions like `DirConvFindCore()` for Bwd and WrW.

Shall we start from the beginning, perhaps. Could you please explain what functionality you need?

atamazov:
Maybe this topic deserves a separate ticket for discussion etc.

alexandraBara:
> Maybe this topic deserves a separate ticket for discussion etc.

I see, so far we were only looking at forward conv. I will discuss this with @JehandadKhan  and then start a discussion thread. We definitely want the other directions covered as well.

JehandadKhan:
@atamazov Thanks for the suggestions, lets re-evaluate what needs to be done and come up with a better solution. @alexandraBara will convert this to a draft until we flush the design completely.

atamazov:
@JehandadKhan If I knew your needs better, I might be able to try to suggest something.

Issue #387, opened 2020-08-18
Post-merge review(s) of PR #347


atamazov:
This is to be closed once all comments mentioned below are resolved.

- [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/347#pullrequestreview-469461853

/cc reviewers of #347 -- @daniellowell @asroy @zjing14 @carlushuang @TejashShah

aserio:
@atamazov, should we close this issue as #347 has been merged?

atamazov:
@aserio Just the opposite.

aserio:
@atamazov, Ah! Skimmed through to fast!

aserio:
@jane-zxy What is the status of these post-merge reviews?

aserio:
@jane-zxy What is the status of these post-merge reviews?

asroy:
After double checking the comments of, I think there are not issues with original PR

https://github.com/ROCmSoftwarePlatform/MIOpen/pull/347#issuecomment-713729502

@atamazov @jane-zxy

atamazov:
I just resolved all review comments but one, -- `NextFlag<false, false>` is not good.

PR #525, opened 2020-10-20
post review of adding vector-load tune parameter

jane-zxy:
This PR is for the post review of  adding vector-load tune parameter, previous pr is: [pr347](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/347)

the related issue link is here: [issue387](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/387)

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
@@ -99,8 +99,6 @@ bool PerformanceImplicitGemmForwardV4R4Xdlops::SetNextValue()
         }
         if(!NextFlag<false, true>(GemmBThreadCopyMoreGemmKPack))
             break;
-        if(!NextFlag<false, false>(GemmAThreadCopyMoreGemmK))
-            break;
         if(!NextTwoPower<1, 8>(GemmKPack))
             break;
         if(!NextTwoPower<4, 128>(GemmNPerWave))
--- a/src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
@@ -136,7 +136,7 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
                 {
                     // list in reverse order of importance,
                     // and favor large GEMM
-                    if(!PreviousTwoPower<1, 8>(tmp.GemmBThreadDataPerRead_GemmN))
+                    if(!PreviousTwoPower<1, 4>(tmp.GemmBThreadDataPerRead_GemmN))
                         break;
                     if(!PreviousTwoPower<1, 8>(tmp.GemmKPerBlock))
                         break;
@@ -152,8 +152,11 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
                         break;
 
                     all_visited = true;
+                    break;
                 } while(false);
 
+                if(all_visited)
+                    break;
                 if(is_valid_func(tmp, ctx))
                     break;
             } while(!all_visited);
@@ -184,8 +187,11 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
                         break;
 
                     all_visited = true;
+                    break;
                 } while(false);
 
+                if(all_visited)
+                    break;
                 if(is_valid_func(tmp, ctx))
                     break;
             } while(!all_visited);
@@ -217,8 +223,11 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
                         break;
 
                     all_visited = true;
+                    break;
                 } while(false);
 
+                if(all_visited)
+                    break;
                 if(is_valid_func(tmp, ctx))
                     break;
             } while(!all_visited);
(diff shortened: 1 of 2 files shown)


src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -184,8 +187,11 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
                         break;
 
                     all_visited = true;
+                    break;
                 } while(false);
 
+                if(all_visited)
```
asroy, line 193: same as above

asroy:
@jane-zxy @atamazov I think this PR is no longer needed
https://github.com/ROCmSoftwarePlatform/MIOpen/issues/387#issuecomment-713733501

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -152,8 +152,11 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
                         break;
 
                     all_visited = true;
+                    break;
```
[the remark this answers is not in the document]
jane-zxy: However, logically, it's possible to run here, "all_visited = true", and if that would be, it will be an endless loop, here.

jane-zxy (commented):
> @jane-zxy @atamazov I think this PR is no longer needed
> [#387 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/387#issuecomment-713733501)

let alone 'no need to add 'break' here",  there's other code change in this pr.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -217,8 +223,11 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
...
                 } while(false);
 
+                if(all_visited)
+                    break;
```
asroy, line 230: This check is doing the same thing as the ```while(!all_visited)``` below, so it's not necessary.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
asroy: This do_while loop is not straightforward, I also had hard time understand it when I modifed it. But I think because of ```while(false)``` below, this is actually not a loop, it's just a single pass, I guess the reason ```do while``` is used here is to use the ```break``` inside.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -236,7 +245,7 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
     });
 
     // second round: really valid
-    if(!tmp.IsReallyValid(ctx))
+    if(!(tmp.IsReallyValid(ctx) && tmp.IsFastToBeUsedForTuning(ctx)))
```
asroy, line 248: If tmp is ```IsReallyValid``` but not ```IsFastToBeUsedForTuning```, we wouldn't want to loop over the tuning range to find another config that is ```IsReallyValid```, we could just accept tmp.

jane-zxy: ok, I see.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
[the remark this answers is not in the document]
jane-zxy: oh, it's 'while(false)', so there will be no dead loop here.
ok, I also think no need to change it here.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -136,7 +136,7 @@ void PerformanceImplicitGemmForwardV4R4Xdlops::EuristicInit(const ConvolutionCon
                 {
                     // list in reverse order of importance,
                     // and favor large GEMM
-                    if(!PreviousTwoPower<1, 8>(tmp.GemmBThreadDataPerRead_GemmN))
+                    if(!PreviousTwoPower<1, 4>(tmp.GemmBThreadDataPerRead_GemmN))
```
jane-zxy, line 139: so maybe only this change is necessary.

asroy: Yes. This change is good. Please revert other changes in this PR.

src/solver/conv_hip_implicit_gemm_fwd_v4r4_xdlops.cpp
```diff
@@ -99,8 +99,6 @@ bool PerformanceImplicitGemmForwardV4R4Xdlops::SetNextValue()
         }
         if(!NextFlag<false, true>(GemmBThreadCopyMoreGemmKPack))
             break;
-        if(!NextFlag<false, false>(GemmAThreadCopyMoreGemmK))
```
atamazov, line 102: Please add comment in accordance to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/347#discussion_r512380380

atamazov (changes requested):

asroy:
@jane-zxy Please address the review comments

atamazov (approved):

asroy (approved):

asroy:
@daniellowell This PR has CI

PR #529, opened 2020-10-20
[COMGR] OCL builds: fix for sram-ecc feature for gfx900

atamazov:
`-msram-ecc` shouldn't be passed to compiler during OCL builds when device does not support sram-ecc target feature. Otherwise HIP runtime fails to load binary.

### DIFF
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -140,6 +140,14 @@ static auto GetOptionsNoSplit()
     return rv;
 }
 
+static bool IsEnabledFeatureSramEcc(const std::string& device)
+{
+    /// \todo Read actual feature status from runtime.
+    static const auto rv = (device == "gfx906" || device == "gfx908") &&
+                           !miopen::IsEnabled(MIOPEN_DEBUG_SRAM_EDC_DISABLED{});
+    return rv;
+}
+
 namespace gcnasm {
 
 static void RemoveOptionsUnwanted(OptionList& list)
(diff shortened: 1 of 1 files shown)


src/comgr.cpp
```diff
@@ -140,6 +140,14 @@ static auto GetOptionsNoSplit()
     return rv;
 }
 
+static bool IsEnabledFeatureSramEcc(const std::string& device)
```
atamazov, line 143: [Note] Refactoring (NFC)

src/comgr.cpp
```diff
@@ -185,7 +193,7 @@ static void AddCompilerOptions(OptionList& list)
 
     // It seems like these options are used only in codegen.
     // However it seems ok to pass these to compiler.
-    if(!miopen::IsEnabled(MIOPEN_DEBUG_SRAM_EDC_DISABLED{}))
+    if(IsEnabledFeatureSramEcc(device))
```
atamazov, line 196: This is fix.

src/comgr.cpp
```diff
@@ -261,10 +269,7 @@ static void RemoveLinkOptionsUnwanted(OptionList& list)
 /// \todo Get list of supported isa names from comgr and select.
 static std::string GetIsaName(const std::string& device)
 {
-    const char* const ecc_suffix = (!miopen::IsEnabled(MIOPEN_DEBUG_SRAM_EDC_DISABLED{}) &&
-                                    (device == "gfx906" || device == "gfx908"))
-                                       ? "+sram-ecc"
-                                       : "";
+    const char* const ecc_suffix = IsEnabledFeatureSramEcc(device) ? "+sram-ecc" : "";
```
atamazov, line 272: [Note] Refactoring (NFC)

PR #537, opened 2020-10-21
[rocm-3.9.x] Added missing checks for new iGemm solvers so they obey the ASM & HIP kernel on/off controls.

atamazov:
Without this fix, some HIP-based iGemm solutions do not obey `MIOPEN_USE_HIP_KERNELS` cmake flag and `MIOPEN_DEBUG_HIP_KERNELS` envvar. This could lead to HIP build errors on systems that lack HIP compilation capabilities, e.g. some embedded ones.

Similar fix for ASM-based iGemm solutions (less important, only affects debugging).

:warning: This is `rocm-3.9.x` specific patch. I will make separate patch for develop.

### DIFF
--- a/src/include/miopen/db_record.hpp
+++ b/src/include/miopen/db_record.hpp
@@ -225,7 +225,7 @@ class DbRecord
 
         const bool ok = values.Deserialize(s);
 
-#if(MIOPEN_SUPPRESS_DB_WARNING==1)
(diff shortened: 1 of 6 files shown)


atamazov:
Not very important because we do not need to support xdlops capable parts in rocm-3.9.x

atamazov:
@daniellowell Please close if this is not needed in rocm-3.9.x anymore.

PR #555, opened 2020-10-29
Follow up to PR#343: Fix problem description Get*Size()

jerryyin:
Follow up from @atamazov's comments [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343#discussion_r456113882). The alternative condition for `Get<In/Out/Weight>Size()` are wrong, and we are not sure what it actually calculates. Fixing it in case it cause any issues in the future.

-----
Example, say we have NCHW layout, and have N: 128, C: 3 H: 5 W:5, for input size:
 - The default condition: N x C x H x W = 128 x 3 x 5 x 5
 - The alternate condition: N x CHW x HW x W = 128 x 75 x 25 x 5

### DIFF
--- a/src/include/miopen/conv/problem_description.hpp
+++ b/src/include/miopen/conv/problem_description.hpp
@@ -188,9 +188,8 @@ struct ProblemDescription
     std::size_t GetInSize() const
     {
         // clang-format off
-        return (GetInLayout() == "NCHW")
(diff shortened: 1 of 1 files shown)


atamazov (approved):
LGTM

atamazov:
CI tests passed.

atamazov:
@jerryyin Thanks for fixing this.

PR #558, opened 2020-10-30
find db generation, reverting gemm time calculation to miopen default

cderb:


daniellowell:
This should go into rocm-3.10.x-staging instead of develop.

PR #563, opened 2020-11-02
[igemm] fix xdlop wrw padding kernel

asleepzzz:
fix
./bin/MIOpenDriver conv -F 4 -n 120 -c 510 -H 5 -W 5 -k 250 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1 -t 1
this patch is not fixing nan issue


daniellowell:
Then what does it do?

asroy:
> Then what does it do?

It fix an issue when testing http://ontrack-internal.amd.com/browse/SWDEV-258337

asroy:
#568 should be able fix the issue, so this PR is no longer needed

PR #566, opened 2020-11-04
fix nan while training by conditionally clear input gradient tensor before launch bwd kernel

carlushuang:
This fix [SWDEV-258416](http://ontrack-internal.amd.com/browse/SWDEV-258416), by clear the input gradient buffer before launch bwd kernel

### DIFF
--- a/src/conv/invokers/impl_gemm_dynamic.cpp
+++ b/src/conv/invokers/impl_gemm_dynamic.cpp
@@ -242,7 +242,7 @@ InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ConvolutionCo
         is_gemm_not_empty.emplace_back(gemm_k_gid > 0);
     }
     bool need_set_zero = false;
-    if(y < stride_h || x < stride_w)
+    if(y < stride_h || x < stride_w || dilation_h != 1 || dilation_w != 1)
         need_set_zero = true;
 
     return [=](const std::vector<Kernel>& kernels) {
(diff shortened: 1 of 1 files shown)


zjing14 (approved):
LTGM

asroy (approved):

src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -241,13 +241,24 @@ InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ConvolutionCo
         const int gemm_k_gid = k * y_dot_slice_gid[gemm_id] * x_dot_slice_gid[gemm_id];
         is_gemm_not_empty.emplace_back(gemm_k_gid > 0);
     }
+    bool need_set_zero = false;
+    if(y < stride_h || x < stride_w)
```
asroy, line 245: Dilation is also a factor. I don't have an exact logic to decide when to set-zero, but it's safe to od

```
if(!(dilation_h == 1 && dilation_w==1 && y >=  stride_h && x >= stride_w))
    need_set_zero = true;
```

asroy (changes requested):

asroy:
@carlushuang  This fix will fix both gfx906 and gfx908?

src/conv/invokers/impl_gemm_dynamic.cpp
carlushuang: Oh, yes, dilation also have same effect. Let me change it

carlushuang:
@asroy Yes. As long as I use some strategy like bwd v4r1 (split filter into dot, slice, ...), we will always have some pixels of input tensor not get updated

asroy (approved):

zjing14 (approved):

PR #569, opened 2020-11-06
Fix db index and query

JehandadKhan:


atamazov:
@JehandadKhan "ignoring instead of replacing" looks good. Does this PR change only binary cache's behavior?

JehandadKhan:
@atamazov That is correct, only sql based binary cache

daniellowell:
How was this tested?

daniellowell (approved):

PR #571, opened 2020-11-06
MP any directional winograd

shurale-nkn:
new features:
-  wrw support
-  stride 2 support for fwd,bwd, wrw


src/solver/conv_MP_anydirectional_winograd.cpp
```diff
@@ -46,66 +47,163 @@
...
+    const int d_dilation_h = (is_bwd) ? (params).kernel_stride_h : 1;    \
+    const int d_dilation_w = (is_bwd) ? (params).kernel_stride_w : 1;
+
+#define DEFINE_SHADER_CONV_MOD_ALIASES(params)                            \
+    DEFINE_SHADER_CONV_MOD_ALIASES_4((params),                            \
+                                     (params).direction.IsForward(),      \
+                                     (params).direction.IsBackwardData(), \
+                                     (params).direction.IsBackwardWrW())
+
+#define GENERATE_MAIN_OPTIONS(options)                             \
```
DrizztDoUrden, line 132: Make this a function like

```suggestion
static void GenerateMainOptions(std::ostream& options)
```

src/solver/conv_MP_anydirectional_winograd.cpp
```diff
@@ -46,66 +47,163 @@
...
+    GenerateClangDefsym((options), "ROCM_METADATA_VERSION", 5);    \
```
DrizztDoUrden, line 133: I'd suggest using this for a new code: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/31e6d8ee1f07d166a687fb47446b41480d7243d9/src/include/miopen/kernel_build_params.hpp#L56.

Like here:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/31e6d8ee1f07d166a687fb47446b41480d7243d9/src/solver/conv_asm_3x3u.cpp#L247
It produces much less boilerplate of GenerateClangDefsym's and string concatenations for OpenCL.

DrizztDoUrden (changes requested):

src/solver/conv_MP_anydirectional_winograd.cpp
DrizztDoUrden: Otherwise multiline macros should be wrapped in a `do { ... } while(false)` construct. It provides safety for ifs and stuff and makes a semicolon mandatory.

shurale-nkn:
@atamazov 
Thank you, but I missed the target. The next define was needed (WORKAROUND_SWDEV_257202).
And for WORKAROUND_SWDEV_257202 need to set 10 variables at a time. And this is a temporary solution, to verify all kernels. 
Required until convergence data is available.

atamazov:
> And for WORKAROUND_SWDEV_257202 need to set 10 variables at a time. 

No problem, set all 10. Or introduce another env var like `MIOPEN_DEBUG_WORKAROUND_SWDEV_257202` and modify the `IS_DISABLED` macro. This is not only fun, but also useful. Let's protect your precious solvers from being spoiled by some other PR in the temporary future (that always tends to become permanent).

src/solver/conv_MP_anydirectional_winograd.cpp
```diff
@@ -56,66 +58,167 @@
...
+    const int d_dilation_w = (is_bwd) ? (params).kernel_stride_w : 1;
+
+#define DEFINE_SHADER_CONV_MOD_ALIASES(params)                            \
+    DEFINE_SHADER_CONV_MOD_ALIASES_4((params),                            \
+                                     (params).direction.IsForward(),      \
+                                     (params).direction.IsBackwardData(), \
+                                     (params).direction.IsBackwardWrW())
+
+#define GENERATE_MAIN_OPTIONS(options)                   \
+    KernelBuildParameters options{};                     \
```
DrizztDoUrden, line 144: It would be better if you would move the declaration outside the macro at least. Otherwise, the macro is still unsafe in usage. Placing it in if would for sure not work as one may expect now. And then you may change it to function because it doesn't declare anything.

DrizztDoUrden: Although if you don't want to copypaste the declaration, I guess, it's kind of ok, I won't stop you from merging this.

DrizztDoUrden (changes requested):

atamazov:
@shurale-nkn The diff looks as if you removed `xform_bidirect_winograd_code.inc` and created a brand new `xform_anydirect_winograd_code.inc`. I am suspecting this is not entirely true. Can you update the branch to reflect renaming the file and changes in it? This would ease reviewing. Thanks.

shurale-nkn:
> @shurale-nkn The diff looks as if you removed `xform_bidirect_winograd_code.inc` and created a brand new `xform_anydirect_winograd_code.inc`. I am suspecting this is not entirely true. Can you update the branch to reflect renaming the file and changes in it? This would ease reviewing. Thanks.

@atamazov 
Please open commit "env vars update, solver renamed" 76afb2e. And then you will see that it has already been renamed. But for some unknown reason, github displays this differently. 
In any case, github able to display correctly changes for commit range. xform_anydirect_winograd_code.inc file code was changed the in range from e0d17cf to a7a83b1. There is no changes after a7a83b1,  other than renaming

DrizztDoUrden (approved):

atamazov:

atamazov:
@shurale-nkn Please set urgency/importance labels on this PR and resolve merge conflicts. Thanks.

JehandadKhan:
This PR is stale and has not been updated in two years

atamazov:
@Kirpich30000 @shurale-nkn Please confirm that this PR is not important for performance.

PR #572, opened 2020-11-06
Multi-Process compilation coord using Binary Cache

JehandadKhan:
This PR implements a coordination scheme between multiple processes so that each binary gets compiled only once. 

The record in the binary cache is used as a lock to indicate to other processes that the object in question is being compiled. This blocks the `miopen::LoadBinary` function until the object is compiled which is then loaded and returned to the process. This approach would remove the load of multiple compilations in a multi-GPU scenario.


src/hip/handlehip.cpp
```diff
@@ -367,23 +367,26 @@ Program Handle::LoadProgram(const std::string& program_name,
...
-    auto hsaco = miopen::LoadBinary(
-        this->GetDeviceName(), this->GetMaxComputeUnits(), program_name, params, is_kernel_str);
+    const auto device = this->GetDeviceName();
+    const auto num_cu = this->GetMaxComputeUnits();
+    params += " -mcpu=" + device;
+    auto hsaco = miopen::LoadBinary(device, num_cu, program_name, params, is_kernel_str);
     if(hsaco.empty())
     {
+#if MIOPEN_ENABLE_SQLITE_KERN_CACHE
+        miopen::MarkBinary(device, num_cu, program_name, params, is_kernel_str);
```
atamazov, line 377: Am I correct that: If process A calls `MarkBinary` before process B, does `MarkBinary` block process B until process A creates a binary and stores it in the cache?

JehandadKhan: That is correct

JehandadKhan:
Closing due to dormancy and questionable value.

PR #576, opened 2020-11-10
[rocm-3.10.x] add bwd upsampling clear check

carlushuang:
create PR same as #566 to for `rocm-3.10.x` branch

### DIFF
--- a/src/conv/invokers/impl_gemm_dynamic.cpp
+++ b/src/conv/invokers/impl_gemm_dynamic.cpp
@@ -226,6 +226,7 @@ float CallImplGemmDynamicBackwardData(const miopen::Handle& handle,
     opArgs.emplace_back(dslice_w_left);
     opArgs.emplace_back(__pack0);
 
+
     for(int gemm_id = 0; gemm_id < num_of_gemms; gemm_id++)
     {
         int _dtile_iy          = gemm_id / x_tilda;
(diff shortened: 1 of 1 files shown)


src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -304,7 +304,16 @@ InvokerFactory MakeImplGemmDynamicForward1x1InvokerFactory(const ConvolutionCont
 InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ConvolutionContext& ctx)
 {
     const auto& conv_problem = ctx.conv_problem;
-    return [conv_problem](const std::vector<Kernel>& kernels) {
+    int stride_h             = conv_problem.GetInHeight() > 1 ? conv_problem.GetKernelStrideH() : 1;
```
asroy, line 307: Is GetInHeight() input (Hi) or really just output image height (Ho)? I know ctx swap them internally.

Ho should be used here for adjusting stride_h.

I think you need some comment here to explain if Hi/Wo get swapped, and that you are adjust stride_h based on Ho, etc

asroy, line 307: Is GetInHeight() input (Hi) or really just output image height (Ho)? I know ctx swap them internally.

Ho should be used here for adjusting stride_h.

I think you need some comment here to explain if Hi/Wo get swapped, and that you are adjusting stride and dilation based on Ho, etc

asroy (changes requested):

src/conv/invokers/impl_gemm_dynamic.cpp
carlushuang: Yes. In backward-data context, the GetInHeight() is indeed the output-gradient tensor hight. This is internally swapped in ctx. Let me add some comment about this

asroy (approved):

asroy:
@daniellowell This PR is doing the same fix as #566, but for ```rocm-3.10.x```. But since you were able to cherry pick #566 into ```rocm-3.10.x-staging```, do we still need this PR?

daniellowell:
@carlushuang I cherry-picked the 566 PR into rocm-3.10.x-staging.

PR #577, opened 2020-11-10
[WORKAROUND] All MP Bidirect Winograd Solvers disabled by default.

atamazov:
W/A for SSD convergence issue mentioned in SWDEV-257202 for 3.10 release. Full blown fix to be provided later by @shurale-nkn.


shurale-nkn (approved):
LGTM

PR #582, opened 2020-11-11
RFC: Add code object v4 support

kzhuravl:
RFC: Add code object v4 support

  - Rename MIOPEN_DEBUG_OPENCL_ENFORCE_COV3 -> MIOPEN_DEBUG_OPENCL_ENFORCE_CODE_OBJECT_VERSION
  - MIOPEN_DEBUG_OPENCL_ENFORCE_CODE_OBJECT_VERSION accepts 4 values:
    - 0 (or unset) - code object version is detected automatically (default)
    - 2 - code object version is 2
    - 3 - code object version is 3
    - 4 - code object version is 4
    - other - error message to the log + fall back to the default.
  - Update automatic code object version detection to pick
    - Code object version 4 if rocm/compiler version is >= 4
    - Code object version 3 if rocm/compiler version is < 4 and >= 3
    - Code object version 2 otherwise
  - Include oflload-arch fix from @pfultz2


src/hipoc/hipoc_program.cpp
```diff
@@ -58,30 +58,30 @@ namespace miopen {
...
-    if(IsDisabled(MIOPEN_DEBUG_OPENCL_ENFORCE_COV3{}))
-        return false;
-    // Otherwise, let's assume that OpenCL kernels shall be compiled to
-    // CO v3 format by default since ROCm 3.0. The simplest way to find out
-    // this right now is checking the HIP compiler version string.
-    return (HipCompilerVersion() >= external_tool_version_t{3, 0, -1});
-}
+    static const std::string opt_enable_v2{"-mno-code-object-v3"};
+    static const std::string opt_enable_v3{"-mcode-object-v3"};
+    static const std::string opt_enable_v4{"-mllvm --amdhsa-code-object-version=4"};
```
t-tye, line 65: Should the new -mcode-object-version be used?

kzhuravl: Totally, I could not find it last night. Do you recall new option's name?

t-tye: I believe the option is called "-mcode-object-version".

On further discussion I think the idea is to restructure this code so if the ENV variable is not present then no option is added so the compiler uses its default.

If the ENV variable is present then the option used is based on the compiler version as follows.

If compiler is <4.0 then use the -mno-code-object-v3/-mcode-object-v3 for code object V2/V3 if the ENV variable is 2/3 otherwise give an error.

If compiler is >=4.0 use -mcode-object-version=N where N is the value of the ENV variable.

atamazov (dismissed):
LGTM!

atamazov:
This PR introduces a couple of quiet changes. I am going to document these in the PR description right now.

atamazov:
PR description updated.

daniellowell (dismissed):

atamazov (dismissed):

atamazov:
Failures on ROCm 3.5, which is not good. I am working on the fix.

atamazov:
No more quiet changes.

atamazov (dismissed):

src/hipoc/hipoc_program.cpp
```diff
@@ -58,31 +59,46 @@ namespace miopen {
...
+inline std::string GetCodeObjectVersionOption()
 {
-    // These options are Ok for ROCm for a long time (since 2.5 or so):
-    static const std::string opt_enable{"-Xclang -target-feature -Xclang +code-object-v3"};
-    static const std::string opt_disable{}; // CO v2 is compiler default.
-    if(enable)
-        return opt_enable;
-    else
-        return opt_disable;
+    static const auto option = GetCodeObjectVersionOptionImpl();
```
atamazov, line 98: @kzhuravl This is optimization for performance. The option is computed only once per library instance.

DrizztDoUrden (dismissed):

Jenkinsfile
```diff
@@ -231,13 +231,6 @@ pipeline {
                     }
                 }
 
-                stage('GCC Debug') {
```
atamazov, line 234: [todo later] Fix the issue with test_conv_igemm_dynamic_small (that takes > 4000 sec to finish -- reason unknown). Remove `codecov`. Restore this stage after that.

Jenkinsfile
```diff
@@ -400,35 +393,12 @@ pipeline {
                         buildJob('hcc', flags: '-DMIOPEN_TEST_HALF=On -DBUILD_DEV=On -DCMAKE_BUILD_TYPE=release', image: image+"rocm", prefixpath: '/opt/rocm', gpu_arch: "gfx906")
                     }
                 }
-
-                stage('Half GCC Debug') {
```
atamazov, line 404: [todo later] This can be just restored.

Jenkinsfile
```diff
@@ -400,35 +393,12 @@ pipeline {
...
-                }
-    
                 stage('Half GCC Release') {
                     agent{ label rocmnode("vega20") }
                     steps{
                         buildJob('g++-5', flags: '-DMIOPEN_TEST_HALF=On -DBUILD_DEV=On -DCMAKE_BUILD_TYPE=release', gpu_arch: "gfx906")
                     }
                 }
-
-                stage('Int8 Hip Release') {
```
atamazov, line 418: [todo later] This can be just restored.

Jenkinsfile
```diff
@@ -400,35 +393,12 @@ pipeline {
...
-                    agent{ label rocmnode("vega20") }
-                    steps{
-                        buildJob('hcc', flags: '-DMIOPEN_TEST_INT8=On -DBUILD_DEV=On -DCMAKE_BUILD_TYPE=release', image: image+"rocm", prefixpath: '/opt/rocm', gpu_arch: "gfx906")
-                    }
-                }
-
-                stage('Int8 GCC Debug') {
```
atamazov, line 425: [todo later] This can be just restored.

Jenkinsfile
```diff
@@ -461,30 +430,20 @@ pipeline {
...
+
+                stage('GCC codecov') {
+                    agent{ label rocmnode("vega") }
                     steps{
-                        buildJob('hcc', flags: '-DMIOPEN_TEST_INT8=On -DBUILD_DEV=On -DMIOPEN_TEST_ALL=On -DCMAKE_BUILD_TYPE=release', image: image+"rocm", prefixpath: '/opt/rocm', gpu_arch: "gfx906")
+                        buildJob('g++-5', flags: '-DBUILD_DEV=On -DCMAKE_BUILD_TYPE=debug', codecov: true, gpu_arch: "gfx900;gfx906")
                     }
                 }
-
-                stage('Bfloat16 Hip Release All') {
```
atamazov, line 471: [todo later] This `Bfloat16 Hip Release All` does not execute any tests. Fix and restore.

atamazov (approved):
- Marked some items that can be done later.
- LGTM!

Jenkinsfile
atamazov: image

Jenkinsfile
atamazov: image

src/hipoc/hipoc_program.cpp
```diff
@@ -58,31 +59,46 @@ namespace miopen {
...
+        MIOPEN_LOG_E("Bad MIOPEN_DEBUG_OPENCL_ENFORCE_CODE_OBJECT_VERSION, using default");
+        code_object_version = 0;
+    }
+
+    if(HipCompilerVersion() >= external_tool_version_t{4, 0, -1})
+    {
+        if(code_object_version == 0) // Env.var is unset.
+            return {};               // Rely on compiler's default.
+        else
+            return std::string("-mcode-object-version=") + std::to_string(code_object_version);
```
atamazov, line 77: @kzhuravl @t-tye Does this option really functional beginning from 4.0.0? CQE is having problems with 4.0.20425.

t-tye: Yes, this option is part of TargetID changes. Is CQE using the compiler that is built from the TargetID sources, or are they using some older BKC version of the compiler?

Note that internal builds may have started reporting they are 4.0 before the TargetID changes were committed. But internal testing should be using the latest mainline sources.

atamazov:
@t-tye @kzhuravl FYI the "-mcode-object-v3" option is also not working in the 4.0 BKC (4.0.20425).

Jenkinsfile
atamazov: Not necessary

Jenkinsfile
atamazov: Not necessary

Jenkinsfile
atamazov: Not necessary

PR #583, opened 2020-11-11
[gpu_reference_kernel] add gpu reference convolution

carlushuang:
[gpu_reference_kernel]
kernel place: `src/kernels/gpu_reference_kernel/`

This is to add a gpu reference kernel,
1. to be used to boost the result validation of ctest and MIOpenDriver.  (https://github.com/ROCmSoftwarePlatform/MIOpen/issues/586)
2. as a fallback kernel for convolution, before we have igemm fully support any tensor size (as mensioned in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/588) 

This reference kernel is very simple and naive, that everybody can easily read & modify it. No optimization, no template, no fancy techniques, only utilize very basic gpu parallel computing knowledge. Correct computation result (and easy maintenance) is top priority.

Currently add convolution kernel. This can support fp32, fp16, bf16, nchw, ncdhw, with/without group.
~Internal API to call such reference layer is in header `src/include/miopen/gpu_reference_kernel.hpp` , currently add 3 as blow~:
re-use immediate API to call these reference kernel

add environment variable `MIOPEN_DRIVER_USE_GPU_REFERENCE`, if set to 1, then use gpu reference kernel while run MIOpenDriver and need verification. by default is 0 . Currently if use gpu reference kernel while validation, ~~the gpu reference always use fp32 data type to run (same as cpu reference)~~, the gpu reference will use same gpu data type (currently fp16/fp32/bf16), different from cpu validation which always use fp32.

- [x] add hip kernel in `src/kernels/gpu_reference_kernel/`, support p32/fp16/bf16, nchw/ncdhw, with/without group.
- [x] ~add internal API to call reference kernel.~ use immediate api to call reference kernel.
- [x] add a method to be used in MIOpenDriver while we use cpu to do validation. (`MIOPEN_DRIVER_USE_GPU_REFERENCE` env)
- [x] add a ctest to test this naive implementation with cpu code
- [x] add this reference kernel a solver under Direct algorithm (`ConvDirectNaiveConvFwd`, `ConvDirectNaiveConvBwd`, `ConvDirectNaiveConvWrw`)


driver/conv_driver.hpp
```diff
@@ -1175,8 +1185,8 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
             for(int i = 0; i < in_sz; i++)
             {
                 if(is_fwd || is_wrw)
-                    in.data[i] =
-                        Data_scale * RAN_GEN<Tgpu>(static_cast<Tgpu>(0.0), static_cast<Tgpu>(1.0));
+                    in.data[i] = static_cast<Tgpu>(RAN_GEN<int>(static_cast<int>(-10), static_cast<Tgpu>(10)));
```
atamazov, line 1188: Why this change is necessary?

driver/conv_driver.hpp
```diff
@@ -1186,8 +1196,8 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
         {
             for(int i = 0; i < out_sz; i++)
                 if(is_bwd || is_wrw)
-                    dout.data[i] =
-                        Data_scale * RAN_GEN<Tgpu>(static_cast<Tgpu>(0.0), static_cast<Tgpu>(1.0));
+                    dout.data[i] = static_cast<Tgpu>(RAN_GEN<int>(static_cast<int>(-10), static_cast<Tgpu>(10)));
```
atamazov, line 1199: Same quection

driver/conv_driver.hpp
```diff
@@ -1275,6 +1285,27 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
             (is_int8 ? out_dev->ToGPU(q, out_int8.data()) : out_dev->ToGPU(q, out.data.data()));
     }
 
+    if(miopen::IsEnabled(MIOPEN_DRIVER_USE_GPU_REFERENCE{}))
+    {
+        if(is_fwd)
+        {
+            out_ref_dev = std::unique_ptr<GPUMem>(
+                new GPUMem(ctx, out_sz, is_int8 ? sizeof(float) : sizeof(Tref)));
+            status |= (is_int8 ? out_ref_dev->ToGPU(q, out_int8.data())
```
atamazov, line 1294: Don't allocate additional memory on GPU, just re-use existing buffers. Otherwise the new method will have limited applicability.

driver/conv_driver.hpp
```diff
@@ -2955,7 +3124,11 @@ int ConvDriver<Tgpu, Tref>::VerifyForward()
         std::cout << "Forward Convolution Failed: " << error << " > " << tolerance << std::endl;
         return EC_VerifyFwd;
     }
-    std::cout << "Forward Convolution Verifies on CPU and GPU (" << error << ')' << std::endl;
+    if(miopen::IsEnabled(MIOPEN_DRIVER_USE_GPU_REFERENCE{}))
+        std::cout << "Forward Convolution Verifies on GPU(ref) and GPU (" << error << ')'
+                  << std::endl;
+    else
+        std::cout << "Forward Convolution Verifies on CPU and GPU (" << error << ')' << std::endl;
```
atamazov, line 3131: Please revert this. You can indicate that driver uses GPU reference somewhere else in the driver's output.

driver/conv_driver.hpp
```diff
@@ -2995,16 +3173,25 @@ int ConvDriver<Tgpu, Tref>::VerifyBackward()
         }
         else
         {
-            std::cout << "Backward Convolution Data Verifies on CPU and GPU (" << error_data << ')'
-                      << std::endl;
+            if(miopen::IsEnabled(MIOPEN_DRIVER_USE_GPU_REFERENCE{}))
```
atamazov, line 3176: Ditto.

driver/conv_driver.hpp
```diff
@@ -3037,8 +3224,12 @@ int ConvDriver<Tgpu, Tref>::VerifyBackward()
         }
         else
         {
-            std::cout << "Backward Convolution Weights Verifies on CPU and GPU (" << error_weights
-                      << ')' << std::endl;
+            if(miopen::IsEnabled(MIOPEN_DRIVER_USE_GPU_REFERENCE{}))
```
atamazov, line 3227: Ditto.

driver/conv_driver.hpp
```diff
@@ -3059,8 +3250,12 @@ int ConvDriver<Tgpu, Tref>::VerifyBackward()
         }
         else
         {
-            std::cout << "Backward Convolution Bias Verifies on CPU and GPU (" << error_bias << ')'
-                      << std::endl;
+            if(miopen::IsEnabled(MIOPEN_DRIVER_USE_GPU_REFERENCE{}))
```
atamazov, line 3253: Ditto.

atamazov (changes requested):

driver/conv_driver.hpp
carlushuang: Oh sorry, this should not changed. This is changed while I'm debugging the code. I will change it back. my fault.

driver/conv_driver.hpp
carlushuang: Sorry, I'll change it back

driver/conv_driver.hpp
carlushuang: While re-use `in_dev` and `wei_dev` as the input tensor for the original gpu convolution, but still need another gpu buffer (here is `out_ref_dev` for fwd) for the reference kernel, as an output. So each direction we still need at least one gpu buffer as an output, hence we can compare the final result.

driver/conv_driver.hpp
carlushuang: OK... let me find another place to print out

asroy:
@carlushuang @atamazov @daniellowell Should we also compile these cpp kernels into ISA, and integrate the ISA into MIOpen too?

driver/conv_driver.hpp
atamazov: We have output device buffers for each convolution under test. And we have host buffers that currently hold reference data.

After convolution executed, you can copy it's output from device buffer to the host buffer. Then you can run reference kernel and obtain reference output in the device buffer. Then, compare device buffer and host buffer as usual.

driver/conv_driver.hpp
atamazov: Well, actually "Verifies on CPU and GPU" would be incorrect for the new case... Let me think about it.

atamazov:
@asroy do you mean binaries? Which compiler you would like to use for building them?

driver/conv_driver.hpp
carlushuang: Oh, indeed requested from @daniellowell that we will ultimately use another gpu kernel to calculate the nrms value of the GPU and reference kernel (to avoid cpu computation & memcpy gpu->cpu). hence indeed we need 2 buffers, so, I guess I still need these buffers

driver/conv_driver.hpp
carlushuang: Sure.

carlushuang:
Hip-clang, I think. And @asroy do you mean that, we measure the fwd/bwd/wrw gpu time of the naive_conv of this release, and then in next release we do a measurement again, and compare the gpu time to check if it became slow or fast?

driver/conv_driver.hpp
atamazov: @daniellowell @carlushuang I do not see substantial value of using GPU for calculating RMS. Current implementation is not a bottleneck. The cost is smaller coverage of GPU-based verification due to earlier exhaustion of device memory. **_I vote for computing RMS on CPU._**

If you instist, we can do the the following:
- Pass two buffers that contain input tensors to the verification kernel (as usual)
- Instead of allocating a buffer for output (for reference tensor), pass buffer that is already allocated and contains output of the convolution under test.
- The verification kernel computes reference tensor, but not writing it into device memory. It should use computed values for RMS calculation. The output of the kernel is scalar RMS value.

The drawbacks of the alternative approach:
- The verification kernel is not a convolution kernel anymore (so it can't be reused somewhere else).
- There is no way for transferring reference tensor to the host (for caching purposes).

carlushuang: @atamazov @daniellowell A side note here. If we do not want to introduce extra gpu memory, while use gpu reference kernel as validation backend, it means it will use the same gpu data type (currently fp16/bf16/fp32). This is different from cpu, which always use fp32 to do calculation. Currently I think there should be no big difference (indeed we always cast data back to fp32 inside kernel).

asroy:
> Hip-clang, I think. And @asroy do you mean that, we measure the fwd/bwd/wrw gpu time of the naive_conv of this release, and then in next release we do a measurement again, and compare the gpu time to check if it became slow or fast? Which may indicate some compiler behavior change



> @asroy do you mean binaries? Which compiler you would like to use for building them?

Any release rocm can be used to generate the ISA. We don't need to update the ISA for every rocm release, we can just keep it forever, until we need to support a new hardware.

carlushuang:
> > Hip-clang, I think. And @asroy do you mean that, we measure the fwd/bwd/wrw gpu time of the naive_conv of this release, and then in next release we do a measurement again, and compare the gpu time to check if it became slow or fast? Which may indicate some compiler behavior change
> 
> > @asroy do you mean binaries? Which compiler you would like to use for building them?
> 
> Any release rocm can be used to generate the ISA. We don't need to update the ISA for every rocm release, we can just keep it forever, until we need to support a new hardware.

OK, I added it in `src/kernels/gpu_reference_kernel/naive_conv_gcn.s`

shaojiewang:
How many times does GPU reference kernel run faster than CPU reference kernel?

carlushuang:
@shaojiewang it's hard to say, but I only can say that if flop is bigger, then gpu will much much faster than cpu. about 2~10x times I think.

asroy (approved):

atamazov:
> 
> 
> @shaojiewang it's hard to say, but I only can say that if flop is bigger, then gpu will much much faster than cpu. about 2~10x times I think.

For the following config
```
./bin/MIOpenDriver conv -F 1 -n 256 -c 128 -k 128 -H 28 -W 28 -x 3 -y 3 -p 1 -q 1
```
acceleration on my Radeon VII / Threadripper 1950X system is ~40X

atamazov:
I see some verification failures with `MIOPEN_DRIVER_USE_GPU_REFERENCE=1`. Please expect update soon.

atamazov:
I run a set of 93 popular FP32 configs with MIOPEN_DRIVER_USE_GPU_REFERENCE=1, some failures:
```
./bin/MIOpenDriver conv -x 20 -y 5 -W 700 -H 161 -c 1 -n 4 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 5.90656e-06 > 3e-06
./bin/MIOpenDriver conv -x 20 -y 5 -W 700 -H 161 -c 1 -n 8 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 1.27719e-05 > 3e-06
./bin/MIOpenDriver conv -x 20 -y 5 -W 700 -H 161 -c 1 -n 16 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 4.05545e-05 > 3e-06
./bin/MIOpenDriver conv -x 20 -y 5 -W 700 -H 161 -c 1 -n 32 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 0.000142916 > 3e-06
./bin/MIOpenDriver conv -x 10 -y 5 -W 341 -H 79 -c 32 -n 8 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 3.48422e-06 > 3e-06
./bin/MIOpenDriver conv -x 10 -y 5 -W 341 -H 79 -c 32 -n 16 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 5.48012e-06 > 3e-06
./bin/MIOpenDriver conv -x 10 -y 5 -W 341 -H 79 -c 32 -n 32 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 1.17529e-05 > 3e-06
./bin/MIOpenDriver conv -x 3 -y 3 -W 480 -H 48 -c 1 -n 16 -k 16 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 2.79963e-05 > 3e-06
./bin/MIOpenDriver conv -x 3 -y 3 -W 240 -H 24 -c 16 -n 16 -k 32 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 4.65254e-06 > 3e-06
./bin/MIOpenDriver conv -x 3 -y 3 -W 224 -H 224 -c 3 -n 16 -k 64 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 0.000124008 > 3e-06
./bin/MIOpenDriver conv -x 7 -y 7 -W 224 -H 224 -c 3 -n 16 -k 64 -p 3 -q 3 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 1.11165e-05 > 3e-06
./bin/MIOpenDriver conv -x 3 -y 3 -W 56 -H 56 -c 64 -n 16 -k 64 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 3.36763e-06 > 3e-06
./bin/MIOpenDriver conv -x 5 -y 5 -W 700 -H 161 -c 1 -n 16 -k 64 -p 1 -q 1 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 4.2319e-05 > 3e-06
./bin/MIOpenDriver conv -x 3 -y 3 -W 350 -H 80 -c 64 -n 16 -k 64 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 4.22074e-05 > 3e-05
./bin/MIOpenDriver conv -x 5 -y 5 -W 350 -H 80 -c 64 -n 16 -k 128 -p 1 -q 1 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 5.74359e-06 > 3e-06
./bin/MIOpenDriver conv -x 1 -y 1 -W 112 -H 112 -c 64 -n 8 -k 64 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 5.40505e-06 > 3e-06
./bin/MIOpenDriver conv -x 1 -y 1 -W 112 -H 112 -c 64 -n 16 -k 64 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 1.16249e-05 > 3e-06
./bin/MIOpenDriver conv -x 1 -y 1 -W 56 -H 56 -c 64 -n 16 -k 256 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 3.48879e-06 > 3e-06
./bin/MIOpenDriver conv -x 1 -y 1 -W 56 -H 56 -c 256 -n 16 -k 64 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 ...Backward Convolution Weights Failed: 3.48305e-06 > 3e-06
```
All failures are related to WrW convolutions so far. The worst case is 0.000142916 (142.916e-06) which is ~50 times bigger than the current threshold (3e-06).

atamazov:
All FP16 tests fail during WrW with:
```
$ ./bin/MIOpenDriver convfp16 -x 3 -y 3 -W 17 -H 17 -c 192 -n 128 -k 192 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6
...
Wall-clock Time Backward Weights Conv. Elapsed: 1.07413 ms, Auxiliary API calls: 4766.92 ms (GWSS: 50.3467)
MIOpen Backward Weights Conv. Algorithm: 5, Solution: 31/ConvHipImplicitGemmV4R1WrW
GPU Kernel Time Backward Weights Conv. Elapsed: 0.927936 ms (average)
stats: bwdw-conv3x3u2, 128, 192, 8, 8, 3, 3, 192,  5435817984, 0, 0, 5858, 0, 0.927936
MIOpen(HIP): Warning [SetupFloats] Unsupported data types configuration: FP16xFP16xFP32
  what():  /home/atamazov/github/MLOpen2/src/include/miopen/conv/conv_direct_naive_conv.hpp:67: unsupported data type:
```
Is it expected?

src/gpu_reference_kernel.cpp
```diff
@@ -0,0 +1,397 @@
...
+#include <miopen/handle.hpp>
+#include <miopen/logger.hpp>
+#include <miopen/tensor.hpp>
+#include <miopen/datatype.hpp>
+#include <miopen/problem_description.hpp>
+#include <miopen/conv/conv_direct_naive_conv.hpp>
+
+namespace miopen {
+
+void GPUReferenceConvolutionForward(Handle& handle,
```
atamazov, line 38: Internal API is used exclusively by the driver and should be considered as a part of the driver. Therefore it should reside in the driver source tree.

[Note] I am in doubt that we really need this API (see next comment).

atamazov, line 38: This PR introduces the full-blown Solver for reference kernels, including Invokers. Let's reuse it instead of duplicating it in different form. You can use Immediate Mode (`miopen::miopenConvolutionForwardImmediate()`) directly.

HINT: The only thing you need to know is a solver id. To get it, use `miopen::solver::Id("solverName").Value()`

src/gpu_reference_kernel.cpp
```diff
@@ -0,0 +1,397 @@
...
+                                    const ProblemDescription& conv_param,
+                                    ConstData_t input_data,
+                                    ConstData_t weight_data,
+                                    Data_t output_data)
+{
+    if(!conv_param.IsLayoutDefault())
+    {
+        MIOPEN_LOG_E("not default layout, will not run");
```
atamazov, line 46: Internal API is used exclusively by the driver and should be considered as a part of the driver. Therefore it should not use `MIOPEN_LOG_*`

atamazov (changes requested):
We can get rid of severe duplication of code.

atamazov:
Great job, but little bit too much for a single PR ;)

carlushuang:
Hi @atamazov Thanks for the careful review. Let me change the code design based on comments.

atamazov:
[Off-topic] @carlushuang Our CI is low on resources. Therefore I think it is worth keeping it from testing intermediate commits. Please avoid pushing interim commits to repo. If you need to push, then:
- Either use [skip ci] in commit message, 
- Or open CI web UI, "Scan Repository Now" and then terminate unwanted job.

atamazov:
:warning: Fwd kernels segfaults on big configs. I've seen this with 6043fb2b66eb0, but I suspect the issue still persists in the most recent commit. ROCm 3.5, Radeon VII (gfx906/56 16GiB).

Example (you need #638 to reproduce this):
```
atamazov@morpheus:~/github/MLOpen2/build/release.hip-clang$ ./bin/MIOpenDriver conv -W 256 -H 256 -c 128 -k 3 -x 3 -y 3 -n 257 -p 1 -q 1 -w 1 -t 1 -i 6 -F 1 -S ConvDirectNaiveConvFwd -V 0
MIOpenDriver conv -W 256 -H 256 -c 128 -k 3 -x 3 -y 3 -n 257 -p 1 -q 1 -w 1 -t 1 -i 6 -F 1 -S ConvDirectNaiveConvFwd -V 0
MIOpen(HIP): Info [Handle] stream: 0x33441e0, device_id: 0
MIOpen(HIP): Info [GetFindModeValueImpl] MIOPEN_FIND_MODE = NORMAL(1)
MIOpen(HIP): Info [GPUMem] hipMalloc 0x7faeb3e00000 8623489024
MIOpen(HIP): Info [GPUMem] hipMalloc 0x7fb5dc809000 13824
MIOpen(HIP): Info [GPUMem] hipMalloc 0x7fae9fc00000 202113024
MIOpen(HIP): Info [GetForwardSolutionCount]
MIOpen(HIP): Info [Measure] Db::Prefetch time: 160.554 ms
MIOpen(HIP): Info [GetForwardSolutions]
MIOpen(HIP): Info [HipCompilerVersionImpl] 3.5.20214
MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, MIOpen version 2.10.0.8312-6043fb2b6
Forward Conv solutions available: 3
- id: 11 algo: 1, time: 30.1283 ms, ws: 0, name: ConvOclDirectFwd
- id: 14 algo: 3, time: 56.2034 ms, ws: 0, name: ConvBinWinograd3x3U
- id: 33 algo: 0, time: 425.099 ms, ws: 301989888, name: gemm
Warning: Solution id (85) is not reported by the library. Trying it anyway...
MIOpen(HIP): Info [GetForwardSolutionWorkspaceSize] solver_id = ConvDirectNaiveConvFwd
MIOpen(HIP): Info [CompileForwardSolution] solver_id = ConvDirectNaiveConvFwd
MIOpen(HIP): Info [FindSolutionImpl] ConvDirectNaiveConvFwd (not searchable)
MIOpen(HIP): Info [ConvolutionForwardImmediate] solver_id = ConvDirectNaiveConvFwd, workspace = 0
Memory access fault by GPU node-2 (Agent handle: 0x32f2ec0) on address 0x7facb3e00000. Reason: Page not present or supervisor privilege.
Aborted (core dumped)
```

carlushuang:
@atamazov ok let me check this.

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -0,0 +1,1580 @@
...
+                    int cur_w   = sx * iwo - px + dx * ix;
+                    if(cur_w < 0 || cur_w >= wi)
+                        valid_w &= 0;
+
+                    if(valid_w & valid_h)
+                    {
+                        int i_idx = ic * hi * wi + cur_h * wi + cur_w;
+                        int w_idx = ic * fy * fx + iy * fx + ix;
+                        value +=
+                            static_cast<double>(p_in[i_idx]) * static_cast<double>(p_wei[w_idx]);
```
atamazov, line 131: max `i_idx` is 2GiB, so max addressable size of float array is 8 GiB. Ditto `w_idx`.

carlushuang: OK. I changed every address calculation to size_t

carlushuang: After new commit to change address calculation with size_t, the above crash is fixed

atamazov:
>  add .s file for gcn 36359d5 

@carlushuang What about implementation limits for asm kernels?

src/kernels/gpu_reference_kernel/naive_conv.cpp
atamazov: I afraid the fix is incorrect. Max `i_idx` is still 2GiB. Please let me know if you need more explanations of some C training.

carlushuang:
@atamazov  asm kernel is just dumped from the hip kernel, exactly the same source code and same logic. This is added requested by @asroy to make sure the asm remain the same across different releases.

src/include/miopen/conv/conv_direct_naive_conv.hpp
```diff
@@ -0,0 +1,99 @@
...
+    return "naive_conv.cpp";
+}
+
+std::string inline ConvDirectNaiveConvCompileOption(const miopen::ConvolutionContext& ctx)
+{
+    const auto device_name = ctx.GetStream().GetDeviceName();
+    std::string filename   = ConvDirectNaiveConvKernelFile(ctx);
+    if(miopen::EndsWith(filename, ".s") && (device_name == "gfx906" || device_name == "gfx908"))
+    {
+        if(ctx.rmv.IsV3())
```
atamazov, line 89: Only .s check is necessary here.

atamazov (changes requested):

src/include/miopen/conv/conv_direct_naive_conv.hpp
carlushuang: Ah, indeed this .s file only works for gfx906/908. Different arch need different .s file, here for simplicity I only let 906/908 to use this .s

atamazov: [Notice] Yes, and different code object version also would require updating the .s file.

daniellowell:
@carlushuang If there are blocking reviews on a PR it is not `READY TO MERGE`.

carlushuang:
@daniellowell  OK. I guess this PR I updated all the review comments, @atamazov please help check if there is anything missing on this PR?

src/include/miopen/conv/conv_direct_naive_conv.hpp
atamazov: Again, gfx arch & CO version checks are performed in `ConvDirectNaiveConvKernelFile()` and not necessary here.

src/kernels/gpu_reference_kernel/naive_conv.cpp
carlushuang: @atamazov sorry for the delay, I missed change the data type of like `i_idx`. This is fixed in latest commit

src/include/miopen/conv/conv_direct_naive_conv.hpp
carlushuang: @atamazov OK, I removed cov & arch check, only keep the `.s` file check

carlushuang:
Hi @atamazov I think all the change is done based on comments, please help check if anything else need change, thanks.

atamazov:
Will re-review today.

atamazov:
Rewiew in progress... Big diff...

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include <miopen/conv/conv_direct_naive_conv.hpp>
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_BWD)
+
+namespace miopen {
+namespace solver {
+
+bool ConvDirectNaiveConvBwd::IsApplicable(const ConvolutionContext& ctx) const
+{
+    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_BWD{}))
```
atamazov, line 41: This ENV switch should not be applicable for cases when DirectNaive is used to compute reference in the driver.

Right away, the simplest solution for this is:
- introducing a global boolean flag in the `miopen::debug` namespace, e.g. `AlwaysEnableConvDirectNaive`
- setting it to `true` in the driver on entry to `Run*GPUReference()` and resetting it prior exit
  - please use guard idiom for that; example -- search for `AutoMiopenWarmupMode` in the conv_driver.hpp.

Example of using of similar approach is `FindEnforceDisable`:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/4ae5799347d139c87553d62e6164d0cee630cb13/src/include/miopen/find_controls.hpp#L36-L43

The same is applicable to `ConvDirectNaiveConvFwd` and `ConvDirectNaiveConvWrw`

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -0,0 +1,195 @@
...
+    if(!ctx.IsLayoutDefault())
+    {
+        return false;
+    }
+
+    if(!(ctx.IsFp32() || ctx.IsFp16() || ctx.IsBfp16()))
+        return false;
+
+    if(!ctx.direction.IsBackwardData())
+    {
```
atamazov, line 53: [Recommendation] [Everywhere] It's a matter of taste, but I wouldn't use curly braces in such cases.

src/include/miopen/conv/conv_direct_naive_conv.hpp
```diff
@@ -0,0 +1,95 @@
...
+#pragma once
+
+#include <string>
+#include <ostream>
+#include <miopen/problem_description.hpp>
+#include <miopen/conv/context.hpp>
+#include <miopen/gcn_asm_utils.hpp>
+#include <miopen/stringutils.hpp>
+
+namespace miopen {
```
atamazov, line 35: Please move these functions into `miopen::solver` namespace.

src/solver/conv_direct_naive_conv_fwd.cpp
```diff
@@ -0,0 +1,195 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <miopen/solver.hpp>
+#include <miopen/conv/data_invoke_params.hpp>
+#include <miopen/conv/wrw_invoke_params.hpp>
```
atamazov, line 29: This header is not necessary.

src/solver/conv_direct_naive_conv_fwd.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include <miopen/env.hpp>
+#include <miopen/visit_float.hpp>
```
atamazov, line 31: This header is not necessary.

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -0,0 +1,195 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <miopen/solver.hpp>
+#include <miopen/conv/data_invoke_params.hpp>
+#include <miopen/conv/wrw_invoke_params.hpp>
+#include <miopen/env.hpp>
+#include <miopen/visit_float.hpp>
```
atamazov, line 31: This header is not necessary.

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -0,0 +1,195 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <miopen/solver.hpp>
+#include <miopen/conv/data_invoke_params.hpp>
+#include <miopen/conv/wrw_invoke_params.hpp>
```
atamazov, line 29: This header is not necessary.

src/solver/conv_direct_naive_conv_wrw.cpp
```diff
@@ -0,0 +1,195 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <miopen/solver.hpp>
+#include <miopen/conv/data_invoke_params.hpp>
+#include <miopen/conv/wrw_invoke_params.hpp>
+#include <miopen/env.hpp>
+#include <miopen/visit_float.hpp>
```
atamazov, line 31: This header is not necessary.

src/solver/conv_direct_naive_conv_wrw.cpp
```diff
@@ -0,0 +1,195 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <miopen/solver.hpp>
+#include <miopen/conv/data_invoke_params.hpp>
```
atamazov, line 28: This header is not necessary.

atamazov (changes requested):
Very good progress! but still needs some improvements, I guess.

src/solver/conv_direct_naive_conv_bwd.cpp
atamazov: This problem is a side effect of my previous comment, https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583#discussion_r535570814. Hopefully, this debug switch is much simpler than an internal API.

src/gpu_reference_kernel.cpp
atamazov: [Resolved]

atamazov:
@carlushuang The problem shown at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583#issuecomment-739033426 is not reproducible anymore.

Please check if these two comments are resolved as well:
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583#issuecomment-738166046
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583#issuecomment-738167963

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: Oh this is to **always** use naive kernel when used as compute reference. Yes let me add this feature

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: OK, let me remove all curly braces if there is only one statement under the `if...else...` condition

src/solver/conv_direct_naive_conv_fwd.cpp
carlushuang: OK let me remove them.

src/solver/conv_direct_naive_conv_fwd.cpp
carlushuang: OK let me remove them.

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: ditto

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: ditto

src/solver/conv_direct_naive_conv_wrw.cpp
carlushuang: ditto

src/solver/conv_direct_naive_conv_wrw.cpp
carlushuang: ditto

carlushuang:
> @carlushuang The problem shown at [#583 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583#issuecomment-739033426) is not reproducible anymore.
> 
> Please check if these two comments are resolved as well:
> 
> * [#583 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583#issuecomment-738166046)
> * [#583 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583#issuecomment-738167963)

Yes, it solves above comments as well. The root cause for above failure is in previous commit not using `double` type as accumulator. By changing this to `double` can keep the accuracy and no computation error anymore. Indeed CPU reference code currently used in MIOpen also use `double`

```
    for(int tid = threadIdx.x; tid < thread_length; tid += 256)
    {
        int iho = tid / wo;
        int iwo = tid % wo;

        double value = .0f;    // use double as accumulator

        for(int ic = 0; ic < c_per_group; ic++)
```

src/include/miopen/conv/conv_direct_naive_conv.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+std::string inline ConvDirectNaiveConvKernelName(const miopen::ConvolutionContext& ctx)
```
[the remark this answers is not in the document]
carlushuang: OK, let me move this function definitions to `src/solver/conv_direct_naive_conv.cpp` and move the header to `src/solver` as well

src/include/miopen/conv/conv_direct_naive_conv.hpp
```diff
@@ -0,0 +1,95 @@
...
+{
```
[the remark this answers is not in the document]
carlushuang: OK

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: @atamazov I put the `AlwaysEnableConvDirectNaive` declaration in `src/include/miopen/solver.hpp`, then can be used in `driver/conv_driver.hpp`. Since I move the `conv_direct_naive_conv.hpp` to `src/solver/conv_direct_naive_conv.hpp`, and if put here is not a good choice to include into `driver/conv_driver.hpp`

src/include/miopen/conv/conv_direct_naive_conv.hpp
carlushuang: OK, moved

zjing14:
@daniellowell @carlushuang Can we get the PR in the next release (i.e., ROCm 4.1)?

carlushuang:
@zjing14 I hope so. currently I'm waiting for @atamazov 's comments on if anything need to change for the code design. At least the feature introduced by this PR is done.

driver/conv_driver.hpp
```diff
@@ -125,6 +125,26 @@ struct AutoMiopenWarmupMode
...
+{
+    AutoConvDirectNaiveAlwaysEnable()
+    {
+        debug_always_enable_conv_direct_naive_prev = miopen::debug::AlwaysEnableConvDirectNaive;
+        miopen::debug::AlwaysEnableConvDirectNaive = true;
+    }
+    AutoConvDirectNaiveAlwaysEnable(const AutoConvDirectNaiveAlwaysEnable&) = delete;
+    AutoConvDirectNaiveAlwaysEnable(AutoConvDirectNaiveAlwaysEnable&&)      = delete;
+    AutoConvDirectNaiveAlwaysEnable& operator=(const AutoConvDirectNaiveAlwaysEnable&) = delete;
+    AutoConvDirectNaiveAlwaysEnable& operator=(AutoConvDirectNaiveAlwaysEnable&&) = delete;
```
atamazov, line 138: Great! But are you really sure we need this for such a simple object?

driver/conv_driver.hpp
```diff
@@ -125,6 +125,26 @@ struct AutoMiopenWarmupMode
...
+    ~AutoConvDirectNaiveAlwaysEnable()
+    {
+        miopen::debug::AlwaysEnableConvDirectNaive = debug_always_enable_conv_direct_naive_prev;
+    }
+
+    private:
+    bool debug_always_enable_conv_direct_naive_prev;
```
atamazov, line 145: [Recommendation] Use short names for small scopes. `prev` is enough. Or delete this. We simply need to swtch this debug flag ON and OFF. Let's follow the KISS principle.

driver/conv_driver.hpp
atamazov, line 138: [Recommendation] Great! But are you really sure we need this kind of protection for such a simple object? I would remove this.

driver/conv_driver.hpp
```diff
@@ -1997,6 +2017,8 @@ int ConvDriver<Tgpu, Tref>::RunForwardCPU()
 template <typename Tgpu, typename Tref>
 int ConvDriver<Tgpu, Tref>::RunForwardGPUReference()
 {
+    AutoConvDirectNaiveAlwaysEnable auto_naive_direct_naive_always_enable;
```
atamazov, line 2020: [Recommendation] Use shorter name (unless you are thinking this code is ugly and so you would like to emphasize it). Everywhere.

src/include/miopen/solver.hpp
```diff
@@ -45,6 +45,14 @@
 
 namespace miopen {
 
+namespace debug {
+
+/// If set to true, then always enable ConvDirectNaive* solver, regardless of environment value
+/// Currently used during driver using naive kernel as gpu reference.
+extern bool AlwaysEnableConvDirectNaive;
```
atamazov, line 52: [Recommendation] Put comment about MT environments.

atamazov (dismissed):
This PR is good to go, but I would appreciate if you fix "recommendation" comments. Up to you.

driver/conv_driver.hpp
carlushuang: OK, let me remove these lines.

driver/conv_driver.hpp
carlushuang: OK, changed to a shorter name `naive_conv_enable`

src/include/miopen/solver.hpp
carlushuang: comments updated about the environment value name.

atamazov: The commend should say that the use of this debug variable is multi-threaded (MT) contexts may produce unspecified results (btw MIOpenDriver does not use it in MT contexts, so we are safe).

atamazov (approved):
Restore green mark

PR #585, opened 2020-11-11
Fixed offload_arch cflags with targetID HIP

zjing14:
- To fix the issue: SWDEV-259381 that introduces new offload_arch flags for all supported devices, which causes the following error
`fatal error: error in backend: Cannot select: intrinsic %llvm.amdgcn.mfma.f32.32x32x4f16`
- Remove all offload_arch flags in HIP_COMPILER_FLAGS, and then add offload_arch flag back with the target device.

### DIFF
--- a/src/hip/hip_build_utils.cpp
+++ b/src/hip/hip_build_utils.cpp
@@ -118,7 +118,6 @@ boost::filesystem::path HipBuild(boost::optional<TmpDir>& tmp_dir,
         params += " -O3 ";
     }
 
-    //std::cerr << ">>>>>>>>>>>> HIP_COMPILER_FLAGS: " << MIOPEN_STRINGIZE(HIP_COMPILER_FLAGS) << std::endl;
     params += " -Wno-unused-command-line-argument -I. ";
     params += MIOPEN_STRINGIZE(HIP_COMPILER_FLAGS);
     params += " --" + std::string(MIOPEN_ARCH_FLAG) + "=" + dev_name;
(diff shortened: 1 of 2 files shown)


src/hip/hip_build_utils.cpp
```diff
@@ -120,6 +120,7 @@ boost::filesystem::path HipBuild(boost::optional<TmpDir>& tmp_dir,
 
     params += " -Wno-unused-command-line-argument -I. ";
     params += MIOPEN_STRINGIZE(HIP_COMPILER_FLAGS);
+    params += " --" + std::string(MIOPEN_ARCH_FLAG) + "=" + dev_name;
```
atamazov, line 123: Move it to line 115 (i.e. replace line 115 with it). And what about line 116 ("--cuda-device-only")?

atamazov (changes requested):

atamazov:
@t-tye Please let me know if COMgr path also needs to be fixed.

t-tye:
> 
> 
> @t-tye Please let me know if COMgr path also needs to be fixed. I guess not -- right now we are not passing anything like "---cuda..." to COMgr.

COMgr takes the name of the target as a target ID now. So you can askp for a "gfx908:sramecc+:xnack-". If you ask simply for "gfx908" you will get a code object that will work with any setting of sramecc or xnack. So does the code currently specify the xnack/sramecc settings? If so the old method still works (using -mxnack/-msramecc) but the new method can be used instead.

zjing14:
CI passed.

pfultz2 (approved):

daniellowell:
@zjing14 @atamazov is this needed still?

atamazov:
@daniellowell I guess so.

daniellowell:
@zjing14 is this still required?

atamazov:
Oops, SWDEV-259381 is closed. I think I or someone else has already fixed this.

PR #595, opened 2020-11-16
Remove wrong error message in HIP-iGEMM

asroy:
Error msg in EuristicInit (called by IsApplicable) should really just be warning message

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_bwd_data_v4r1_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_data_v4r1_xdlops.cpp
@@ -690,8 +690,7 @@ void PerformanceImplicitGemmBwdDataV4R1Xdlops::EuristicInit(const ConvolutionCon
     // final check
     if(!tmp.IsReallyValid(ctx))
     {
-        MIOPEN_LOG_E("All attempts failed");
-        assert(false);
+        MIOPEN_LOG_I("All attempts failed");
     }
     *this = tmp;
     MIOPEN_LOG_I(ToString());
@@ -840,7 +839,11 @@ ConvSolution ConvHipImplicitGemmBwdDataV4R1Xdlops::GetSolution(
 {
     ConvSolution result;
 
-    assert(config.IsValid(ctx));
+    if(!config.IsReallyValid(ctx))
+    {
+        MIOPEN_LOG_E("invalid performance parameter");
+        assert(false);
+    }
 
     // a series of kernels
     for(std::size_t gemm_id = 0; gemm_id < CalculateNumberOfGemm(ctx); ++gemm_id)
--- a/src/solver/conv_hip_implicit_gemm_bwd_v1r1_xdlops_nchw_kcyx_nkhw.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_v1r1_xdlops_nchw_kcyx_nkhw.cpp
@@ -224,8 +224,7 @@ void PerformanceImplicitGemmBwdV1R1Xdlops::EuristicInit(const ConvolutionContext
     // final check
     if(!tmp.IsReallyValid(ctx))
     {
-        MIOPEN_LOG_E("All attempts failed");
-        assert(false);
+        MIOPEN_LOG_I("All attempts failed");
     }
     *this = tmp;
     MIOPEN_LOG_I(ToString());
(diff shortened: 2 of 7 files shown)


atamazov (changes requested):
> Error msg in EuristicInit (called by IsApplicable) should really just be warning message

Indeed, `IsApplicable()` should not issue error messages when problem is not applicable.

However, `GetPerformanceConfig()` should continue issuing errors if it cannot return a valid `PerformanceConfig` for some problem.

The fact that they both use `EuristicConfig()` is somewhat of a coincidence. `EuristicInit()` is not a part of the Solver interface. The design that uses it was borrowed some time ago from some other solver.

As `EuristicInit()` is not a part of the Solver interface, you can easily add a boolean parameter, like `beQuiet` and pass `true` when calling it from `IsApplicable`. In other cases, pass `false` so `EuristicInit ()` will issue error message (and keep `assert()` please). I think this would be the simplest solution.

asroy:
@atamazov I've moved assert into ```GetSolution```, doesn't it have the same effect?

```
    if(!config.IsReallyValid(ctx))
    {
        MIOPEN_LOG_E("invalid performance parameter");
        assert(false);
    }
```

asroy:
@atamazov Can you re-review the PR?

src/solver/conv_hip_implicit_gemm_bwd_data_v4r1_xdlops.cpp
```diff
@@ -840,7 +839,11 @@ ConvSolution ConvHipImplicitGemmBwdDataV4R1Xdlops::GetSolution(
 {
     ConvSolution result;
 
-    assert(config.IsValid(ctx));
+    if(!config.IsReallyValid(ctx))
```
atamazov, line 842: [Recommendation] `GetSolution()` is not allowed to fail if `IsApplicable()` is true, unless there is a bug in the Solver (see doc in solver.hpp). If you see possibility of this, and would like to double-check, then:
- hide all checks under `#if !NDEBUG` so this will happen only in Debug builds and wouldn't waste time in releases.
- Use MIOPEN_THROW to terminate as this is internal error.

Of course you can keep this as is, if you think that probability of such errors is high. But I would say that the Solvers that have high probability of failure in GetSolution are not of production quality yet and should be disabled by default.

src/solver/conv_hip_implicit_gemm_bwd_data_v4r1_xdlops.cpp
```diff
@@ -690,8 +690,7 @@ void PerformanceImplicitGemmBwdDataV4R1Xdlops::EuristicInit(const ConvolutionCon
     // final check
     if(!tmp.IsReallyValid(ctx))
     {
-        MIOPEN_LOG_E("All attempts failed");
-        assert(false);
+        MIOPEN_LOG_I("All attempts failed");
```
atamazov, line 693: This may produce unnecessary spam during `IsApplicable()` calls. Make it Info2 at least. Or, better, follow https://github.com/ROCmSoftwarePlatform/MIOpen/pull/595#pullrequestreview-531881874.

atamazov (approved):
This can be merged.

atamazov:
@asroy 
> @atamazov Can you re-review the PR?

Done.

src/solver/conv_hip_implicit_gemm_bwd_data_v4r1_xdlops.cpp
asroy: #597 has been created to track the review comment

src/solver/conv_hip_implicit_gemm_bwd_data_v4r1_xdlops.cpp
asroy: Current IsApplicable() will call EuristicInit(), which will loop over all tuning range to try to find one set of parameters. If IsApplicable() will return true if EuristicInit() find THAT set of parameters. So GetSolution() should never fail, if IsApplicable() is true. So, this assert in GetSolution() is not supposed to be hit.

atamazov: Yes, thanks, of course I know that. I give recommendation, you decide.

atamazov: @asroy If you decide to follow the resomendation, please use `#if !MIOPEN_NDEBUG`, _NOT_ `#if !NDEBUG` (this is incorrect).

Issue #598, opened 2020-11-17
iGEMM tuning failure on gfx908


zjing14:
Following configs from ResNext101 failed with tuning (MIOPEN_FIND_ENFORCE=4) on gfx908. These configs can pass with default tuning parameters (i.e., parameters from EnuristInit)

```
MIOpenDriver convfp16 -V 0 -F 0 -n 128 -c 1024 -H 14 -W 14 -k 1024 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 32 -t 1
MIOpen(HIP): Warning [GenericSearch] ConvHipImplicitGemmBwdDataV4R1Xdlops: Searching the best solution among 0 (spare)...
```

```
MIOpenDriver convfp16 -V 0 -F 0 -n 128 -c 1024 -H 28 -W 28 -k 1024 -y 3 -x 3 -p 1 -q 1 -u 2 -v 2 -l 1 -j 1 -m conv -g 32 -t 1
MIOpen(HIP): Warning [GenericSearch] ConvHipImplicitGemmBwdDataV4R1Xdlops: Searching the best solution among 0 (spare)...
```

```
MIOpenDriver convfp16 -V 0 -F 0 -n 128 -c 512 -H 28 -W 28 -k 512 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 32 -t 1
MIOpen(HIP): Warning [GenericSearch] ConvHipImplicitGemmBwdDataV4R1Xdlops: Searching the best solution among 0 (spare)...
MIOpen(HIP): Warning [GenericSearch] ConvHipImplicitGemmWrwV4R4Xdlops: Searching the best solution among 0 (spare)...
```

```
MIOpenDriver convfp16 -V 0 -F 0 -n 128 -c 512 -H 56 -W 56 -k 512 -y 3 -x 3 -p 1 -q 1 -u 2 -v 2 -l 1 -j 1 -m conv -g 32 -t 1
MIOpen(HIP): Warning [GenericSearch] ConvHipImplicitGemmBwdDataV4R1Xdlops: Searching the best solution among 0 (spare)...
MIOpen(HIP): Warning [GenericSearch] ConvHipImplicitGemmWrwV4R4Xdlops: Searching the best solution among 0 (spare)...
```

```
[...]

PR #601, opened 2020-11-18
[igemm]fix tuning paramter of spare(issue 598)

asleepzzz:
fix issue #598
we remove spare set judgement in is_fast function
due to we need fully searching it when 2nd round

PR #605, opened 2020-11-19
Fix iGEMM wrw v4r4 tuning failure on gfx908

ltqin:
This PR fix iGEMM wrw v4r4 tuning failure on gfx908, origin from #598 
Change:
Add spare to wrw v4r4 solver

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
@@ -41,7 +41,7 @@ namespace solver {
 
 PerformanceImplicitGemmWrwV4R4Xdlops::PerformanceImplicitGemmWrwV4R4Xdlops()
     : PerformanceImplicitGemmWrwV4R4Xdlops::PerformanceImplicitGemmWrwV4R4Xdlops(
-          4, 4, 1, 4, 4, 1, false, false,false)
+          4, 4, 1, 4, 4, 1, false, false, false)
 {
 }
 
@@ -690,7 +690,7 @@ bool PerformanceImplicitGemmWrwV4R4Xdlops::IsReallyValid(const ConvolutionContex
 bool PerformanceImplicitGemmWrwV4R4Xdlops::IsFastToBeUsedForTuning(
     const ConvolutionContext& ctx) const
 {
-  
+
     if(use_spare_set)
         return true;
     // somehow, 128x128 wave-wise GEMM tend to spill register
--- a/src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
@@ -85,7 +85,7 @@ operator==(const PerformanceImplicitGemmWrwV4R4Xdlops& other) const
         && GemmKPack == other.GemmKPack
         && GemmAThreadCopyMoreGemmK  == other.GemmAThreadCopyMoreGemmK
         && GemmBThreadCopyMoreGemmK  == other.GemmBThreadCopyMoreGemmK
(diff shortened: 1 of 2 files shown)


daniellowell:
What does this fix that #601 doesn't fix?

asroy (approved):

src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
```diff
@@ -76,7 +84,8 @@ operator==(const PerformanceImplicitGemmWrwV4R4Xdlops& other) const
         && GemmNPerWave == other.GemmNPerWave
         && GemmKPack == other.GemmKPack
         && GemmAThreadCopyMoreGemmK  == other.GemmAThreadCopyMoreGemmK
-        && GemmBThreadCopyMoreGemmK  == other.GemmBThreadCopyMoreGemmK;
+        && GemmBThreadCopyMoreGemmK  == other.GemmBThreadCopyMoreGemmK
+	&& use_spare_set == other.use_spare_set;
```
atamazov, line 88: [formatting] Tab -> 4 spaces

src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
```diff
@@ -681,6 +690,9 @@ bool PerformanceImplicitGemmWrwV4R4Xdlops::IsReallyValid(const ConvolutionContex
 bool PerformanceImplicitGemmWrwV4R4Xdlops::IsFastToBeUsedForTuning(
     const ConvolutionContext& ctx) const
 {
+
+    if(use_spare_set)
```
atamazov, line 694: [Question] So the idea is: If there is none "Fast" perfConfigs exist, let's use all? This seems suspicious, because in this case we can't expect good performance from the Solver, even tuned one. Perhaps it is better to make it non-applicable, thus saving host time?

atamazov (changes requested):

atamazov:
@ltqin 
> This PR fix iGEMM wrw v4r4 tuning failure on gfx908, origin from #598

Could you please denote (provide URL) the specific review comment (from #598) that you would like to resolve by this PR? Thanks.

atamazov:
Good and clear PR descriptions save reviewers' time.

src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
asleepzzz: @zjing14 ,I think we still need 2 pass due to tuning time is too long
Isn't it?

asroy: "Fast" is only relative. It means this tuning parameter is relatively fast compare with other tuning parameters. It doesn't indicate if this solver using this tuning parameter is "fast" or "slow" compare to other solvers. @atamazov

asroy:
@asleepzzz When open a PR that fixes an issue, please add some description what is the issue and the solution.

src/solver/conv_hip_implicit_gemm_wrw_v4r4_xdlops.cpp
atamazov: Okay then.

atamazov: [Recommendation]
> It means this tuning parameter is relatively fast compare with other tuning parameters. 

Rename the function or insert relevant comment explaining this.

atamazov (approved):
LGTM

asleepzzz:
> @asleepzzz When open a PR that fixes an issue, please add some description what is the issue and the solution.

ok

atamazov:
All CI tests passed, ready to merge!

PR #606, opened 2020-11-19
Parallel compilation is performed whether or not perform tuning process

ZPEthanWen:
Let's just talk about the case where there is no kernel cached in memory and kernel database while performing convolution. There is two situations when MIOpen-2.8.0 finds solutions.

1. If the tuning process is not required to execute.  MIOpen will first determine whether a solver is applicable. If the solver is applicable, then, get the default solution of the solver. MIOpen repeats the above process until all default solutions are provide by each applicable solver. After that, MIOpen compiles all the solutions to kernels in parallel.
2. If the tuning process is required to execute explicitly. MIOpen will first determine whether a solver is applicable. If so, then, MIOpen will perform a loop process serially to find a optimal solution of the solver. The process compiles solutions of a solver one by one, run the kernel to determine which solution is best and then return it. MIOpen repeats the above process until all optimal solutions are provide by each applicable solver. After that, MIOpen still executes the parallel compilation but almost do nothing.

I think the two process is conflict and the tuning process is too time-consuming because of serial compilation. So, I make the two process unified. Here is the test result.
```
rm -rf ~/.config/miopen ~/.cache/miopen /tmp/*
./MIOpenDriver conv --batchsize 16 --in_channels 64 --out_channels 256 --iter 1 -V 0
```

image


atamazov:
Hello @ZPEthanWen, thanks for your initiative. Unfortunately, the development process does not include adding new features to existing releases. And AFAICS the feature (parallel compilation for tuning) is already implemented in `develop` branch:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2a182a8c6a1456f0eace0e4665cc62942c238bbe/src/include/miopen/generic_search.hpp#L348-L362
In the future, I recommend opening a feature request before beginning actual development.

ZPEthanWen:
@atamazov, I'm very happy to witness the rapid development of miopen. I think the the feature (parallel compilation for tuning) in `develop` branch is only for a solver, For all solvers, it's still process one by one.
```
miopen::each_args(
            [&](auto solver) {
                if(count >= limit)
                    return;
                if(find_only.IsValid() && find_only != Id{SolverDbId(solver)})
                { // Do nothing (and keep silence for the sake of Tuna), just skip.
                }
                else if(!solver.IsApplicable(search_params))
                    MIOPEN_LOG_I2(SolverDbId(solver) << ": Not applicable");
                else if(search_params.use_dynamic_solutions_only && !solver.IsDynamic())
                    MIOPEN_LOG_I2(SolverDbId(solver) << ": Skipped (non-dynamic)");
                else
                {
                    const Solution s = FindSolution(solver, search_params, db, invoke_ctx);
                    if(s.Succeeded())
                    {
                        ++count;
                        ss.push_back(s);
                        MIOPEN_LOG_I2(SolverDbId(solver) << ": Success.");
                    }
                    else
                    {
                        /// \todo If Solver is applicable it must provide an appropriate Solution.
                        /// This is not the case for some 20x5 convolutions (and possibly others).
                        /// Normally we should not get here and message level should be Error.
                        /// For now, let's use Info (not Warning) level to avoid
                        /// flooding the console.
                        MIOPEN_LOG_I(SolverDbId(solver)
                                     << ": [Warning] Applicable Solver not succeeded.");
                    }
                }
            },
            Solvers{}...);
```
I'm not sure if it's my version or the develop branch is performing well. I will check it out. Anyway, thank you very much for your reply.

atamazov:
@ZPEthanWen 
> I think the the feature (parallel compilation for tuning) in `develop` branch is only for a solver, For all solvers, it's still process one by one.

Correct. Parallel tuning of several solvers doesn't make much sense, because even one Solver usually has hundreds of kernels to build.

daniellowell:
This PR seems to contain a lot of changes not relevant to this PR itself. Please rebase with the latest develop.

atamazov:
@daniellowell We already [have this functionality in develop](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/606#issuecomment-730375322).

PR #611, opened 2020-11-19
Early escape for compile_only + custom err

alexandraBara:
During compile only perf_db is empty hence we would see "...cannot be executed due to incorrect params" err thrown, but we want to customize this err since its caused by compiling only and we want Tuna to recognize it.

### DIFF
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -103,6 +103,7 @@ typedef enum {
     miopenStatusNotImplemented = 6, /*!< Use of unimplemented feature. */
     miopenStatusUnknownError   = 7, /*!< Unknown error occurred. */
     miopenStatusUnsupportedOp  = 8, /*!< Unsupported operator for fusion. */
+    miopenStatusNoGPU          = 9, /*!< GPU operations skipped. */
 } miopenStatus_t;
 
 /*! @brief Get character string for an error code.
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -94,16 +94,16 @@ MIOPEN_DECLARE_OBJECT(miopenHandle);
  * Error codes that are returned by all MIOpen API calls.
 */
 typedef enum {
-    miopenStatusSuccess        = 0, /*!< No errors */
-    miopenStatusNotInitialized = 1, /*!< Data not initialized. */
-    miopenStatusInvalidValue   = 2, /*!< Incorrect variable value. */
-    miopenStatusBadParm        = 3, /*!< Incorrect parameter detected. */
-    miopenStatusAllocFailed    = 4, /*!< Memory allocation error. */
-    miopenStatusInternalError  = 5, /*!< MIOpen failure. */
-    miopenStatusNotImplemented = 6, /*!< Use of unimplemented feature. */
-    miopenStatusUnknownError   = 7, /*!< Unknown error occurred. */
-    miopenStatusUnsupportedOp  = 8, /*!< Unsupported operator for fusion. */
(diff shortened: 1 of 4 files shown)


atamazov:
@alexandraBara I think that the application that you use in `compile-only` mode should be modified. If this is MIOpenDriver, then you can add new flag (or use some existing one, e.g. `-V 2`) and update the driver so it will ignore the error returned by the library.

JehandadKhan:
@atamazov we do not wish to modify MIOpenDriver since we have [another application](https://github.com/ROCmSoftwarePlatform/fin) in development which is specifically designed for Tuna.

atamazov:
@alexandraBara So modify your application. What prevents you from doing so?

alexandraBara:
> @alexandraBara So modify your application. What prevents you from doing so?

Are you saying we should not be throwing a custom error here and just let the err propagate and let the Driver handle it?
If so, its not quite so straight forward, there are other places down the pipeline that will fail if we dont escape early.

JehandadKhan:
@atamazov  The major issue we wish to deal with is a more descriptive error message when the `MIOPEN_DEBUG_COMPILE_ONLY` env var is set. If we do not catch it here and let the execution continue other parts of the library either try to execute a kernel when there is no GPU ( or a GPU of different kind) or throw an error (`"Forward Convolution cannot be executed due to incorrect params"`) which has no correlation to the root cause. 

Throwing a specific error allows the client application (whether the driver or fin) to know for sure what exactly happened.

atamazov:
@alexandraBara 
> Are you saying we should not be throwing a custom error here and just let the err propagate and let the Driver handle it? If so, its not quite so straight forward, there are other places down the pipeline that will fail if we dont escape early.

The library will fail immediately with "...Convolution cannot be executed due to incorrect params...". The driver must be aware of the "compile-only" mode, so there is no problem for the driver to skip running and benchmarking kernels. However, I am not against your solutions, I just want to understand why we need it.

@JehandadKhan 
> @atamazov The major issue we wish to deal with is a more descriptive error message when the `MIOPEN_DEBUG_COMPILE_ONLY` env var is set. If we do not catch it here and let the execution continue other parts of the library ... throw an error (`"Forward Convolution cannot be executed due to incorrect params"`) which has no correlation to the root cause.
> 
> Throwing a specific error allows the client application (whether the driver or fin) to know for sure what exactly happened.

:+1: Sounds reasonable.

[Informative]
> If we do not catch it here and let the execution continue other parts of the library either try to execute a kernel...

No. All benchmarking already done before this new `if(IsEnabled(MIOPEN_DEBUG_COMPILE_ONLY{})) MIOPEN_THROW(...)` stuff.

include/miopen/miopen.h
```diff
@@ -103,6 +103,7 @@ typedef enum {
     miopenStatusNotImplemented = 6, /*!< Use of unimplemented feature. */
     miopenStatusUnknownError   = 7, /*!< Unknown error occurred. */
     miopenStatusUnsupportedOp  = 8, /*!< Unsupported operator for fusion. */
+    miopenStatusNoGPU          = 9, /*!< GPU operations skipped. */
```
atamazov, line 106:
```suggestion
    miopenStatusGpuOperationsSkipped          = 9, /*!< This is not an error. */
```

In fact, I doubt it. The library did exactly what was requested. It can complete the `Find()` call normally and simply return an empty list of found solutions and a success code. But let's keep dedicated code as we just discussed.

atamazov (changes requested):
Let's use more descriptive name for status code.

atamazov (dismissed):

atamazov (dismissed):

atamazov (approved):

PR #613, opened 2020-11-19
[tests] Reduce of workload for "test_sqlite_perfdb --all"

atamazov:
Currently the test has imperfections that can lead to timeouts under heavy loads. This should reduce the probability of timeouts, especially on systems with slow HDD.

On my system (fast SSD storage):

test | \<blank\> | --all
-- | -- | --
test_perfdb | 1.6 sec | 13 sec 
test_sqlite_perfdb (baseline) | 29 sec | __480 sec__
test_sqlite_perfdb (this PR) | 29 sec | ___114 sec___

### DIFF
--- a/test/sqlite_perfdb.cpp
+++ b/test/sqlite_perfdb.cpp
@@ -1236,9 +1236,9 @@ struct PerfDbDriver : test_driver
         if(full_set)
         {
             tests::full_set()                         = true;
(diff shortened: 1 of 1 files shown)


DrizztDoUrden (approved):
I love that bot thinks this changes coverage. Its metric should be strange indeed.

PR #619, opened 2020-11-20
[Legacy OCL] Parallel compile in search

### LINKED ISSUE #618: [Legacy OCL] The ConvOclDirectFwdLegacyExhaustiveSearch:Search is still serial compilation for solutions
The `ConvOclDirectFwdLegacyExhaustiveSearch:Search` is calling `SearchImpl` function to tune the parameters. In the `SearchImpl`, the number of solutions is up to five hundred. But it still performs serial compilation for all solutions by calling `MeasurePerfConfig` function. This is very inefficient, will take a long time to complete.

zpwenjh:
Resolves issue #618.

This PR solves the #618 temporarily. My code of `SearchImpl` can be simplified and universal, I think.


atamazov:
@ZPEthanWen Useful change. But needs to be thoroughly tested. Could you please add testing information (so one can reproduce the testing procedure you've did?)

atamazov:
@ZPEthanWen Another question (haven't read the code yet). The library should only use parallelization for building kernels, but not during benchmarking (time measurements). Is it so?

DrizztDoUrden:
It looks like the generic search and looks fine, but I think @alexandraBara is the developer most familiar with PrecompileKernels and parallel compilation stuff.
Updating this solver to use the generic search would be a better solution overall, but the proposed change is still an improvement.

alexandraBara:
> It looks like the generic search and looks fine, but I think @alexandraBara is the developer most familiar with PrecompileKernels and parallel compilation stuff.
> Updating this solver to use the generic search would be a better solution overall, but the proposed change is still an improvement.

I agree with this, i don't think it is sustainable to update individual solvers to use the same logic as in generic_search. If these solvers can be moved to use generic_search instead i think that would be a better approach. Even the implementation in generic_search can use some updating as it currently relies on kern_cache being turned on.

atamazov:

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
```diff
@@ -447,70 +429,99 @@ ConvOclDirectFwdLegacyExhaustiveSearch::SearchImpl(const ConvolutionContext& par
...
-                                                       << ' '
-                                                       << result);
-                        }
-                        run_counter++;
                     }
                 }
             }
         }
+// PrecompileKernels call saves to binary_cache, this needs to be escaped if KERN_CACHE is not on.
+#if MIOPEN_ENABLE_SQLITE_KERN_CACHE
```
atamazov, line 445: Please use this to decide is binary cache is enabled (rename to `IsBinaryCacheDisabled` and make it public):
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/584da20a32d26164ac1b65966eda54a1524a6801/src/binary_cache.cpp#L103

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
```diff
@@ -602,115 +613,95 @@ ConvOclDirectFwdLegacyExhaustiveSearch::SearchImpl(const ConvolutionContext& par
...
-                                                               top_ocl_ptr,
-                                                               wei_ocl_ptr,
-                                                               bias_ocl_ptr,
-                                                               default_time,
-                                                               params,
-                                                               default_config);
-        }
-        else
+// PrecompileKernels call saves to binary_cache, this needs to be escaped if KERN_CACHE is not on.
+#if MIOPEN_ENABLE_SQLITE_KERN_CACHE
```
atamazov, line 632: Ditto

atamazov (changes requested):
I think this is useful, but we need to test it thoroughly.

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
zpwenjh: @atamazov I'll do it as soon as possible.

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
zpwenjh: Whether binary  cache is enabled or disabled, parallel compilation by `PrecompileKernels`  is required here. And `LoadProgram` called in `PrecompileKernels` has the mechanism of whether binary cache is disabled or not. 
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/8fe06e67a16b7261d59d70a0ae75d83e678ad788/src/hip/handlehip.cpp#L411-L421
So I think there is no need to judge the binary cache is disabled or not here. And This also applies to `generic_search`:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/8fe06e67a16b7261d59d70a0ae75d83e678ad788/src/include/miopen/generic_search.hpp#L352-L365
According to the current situation, if the binary cache is disabled, then MIOpen will still perform serial compilation here and in `generic_search`, this is not what we want.

zpwenjh: Do I need to incorporate this "search" code into `generic_search`? I'm not sure I can do it as your wish.

DrizztDoUrden: Updates to the generic_search should be applied in a separate PR.

atamazov:
Let's agree that legacy OCL solvers needs to be totally reworked to fully comply with Solver/Solution architecture. We have some thoughts how and when this can be done. But not now.

atamazov:
The repository has been deleted (`zpwenjh:parallel_compile_in_search`)

PR #622, opened 2020-11-20
Add compile time enabled timer function

JehandadKhan:
Introduces a timer class which measures the amount of time spent in the declared scope. 

The functionality is disabled by deafult and would need to be enabled using the `-DMIOPEN_TIME_FUNCTIONS=On` on the cmake command line.

This is useful for measuring host time overhead in MIOpen. 

Any function where measurement is required would need to have the macro added. For example 

```cpp
void foo(...)
{
    MIOPEN_LOG_SCOPE_TIME;
    ...
}
```

The logging is independent of MIOpen logging to improve measurement accuracy.


atamazov:
@JehandadKhan What is the intended usage of this? I think you have something in mind already. Also it would be nice to look at example console log.

Notice that we already have `MIOPEN_ENABLE_LOGGING_ELAPSED_TIME` which is always available.

JehandadKhan:
@atamazov  The intended purpose of this macro is to measure the time spent in a specific function. For example if I wish to measure how much time does MIOpen spend while in the find function I would add the macro to the Find function in convolution_api.cpp.

According to our documentation: `MIOPEN_ENABLE_LOGGING_ELAPSED_TIME` "Adds a timestamp to each log line. Indicates the time elapsed since the previous log message, in milliseconds." which may be used to estimate the time spent in a specific function but is indirect and tedious. Furthermore, logging itself has overhead and therefore I wanted an additional mechanism which is independent of the logging overhead.

atamazov:
> Furthermore, logging itself has overhead  and therefore I wanted an additional mechanism which is independent of the logging overhead.

The overhead is one read of static const boolean variable and one conditional jump.

src/include/miopen/timer.hpp
```diff
@@ -73,6 +73,29 @@ class CompileTimer
...
+    std::cout << m_name << " : " << dur.count() << " musec\n";
+  }
+ private:
+  std::string m_name;
+  std::chrono::time_point<std::chrono::high_resolution_clock> m_beg;
+};
+
+#if MIOPEN_TIME_FUNCTIONS
+#define MIOPEN_FUNC_TIMER                               \
+        miopen::FunctionTimer miopen_timer(miopen::LoggingParseFunction(__func__, __PRETTY_FUNCTION__))
```
atamazov, line 93: AFAICS this is more logging than timing functionality. Please move it into logger.h and rename to `MIOPEN_LOG_FUNC_TIMER` (or `...TIME`? or `...SCOPE_TIME`?).

(Consequently, please rename `FunctionTimer` to something that begins with `Log`, for ex. `LogScopeTime`, ane cmake variable to something like `MIOPEN_LOG_FUNC_TIME_ENABLE`.)

src/include/miopen/timer.hpp
```diff
@@ -73,6 +73,29 @@ class CompileTimer
...
 
+class FunctionTimer{
+ public:
+  FunctionTimer(std::string name)
+      : m_name(std::move(name)),
+        m_beg(std::chrono::high_resolution_clock::now()) { }
+  ~FunctionTimer() {
+    auto end = std::chrono::high_resolution_clock::now();
+    auto dur = std::chrono::duration_cast<std::chrono::microseconds>(end - m_beg);
+    std::cout << m_name << " : " << dur.count() << " musec\n";
```
atamazov, line 84: [Question] This seems unusable in MT/MP apps (as the reader needs some means to distinguish one thread/process from another thread/process; see `MIOPEN_ENABLE_LOGGING_MPMT`). Is it Ok for you?

atamazov (changes requested):

JehandadKhan:
@atamazov all your suggestions/comments have been incorporated.

src/include/miopen/logger.hpp
```diff
@@ -342,6 +343,33 @@ std::string LoggingParseFunction(const char* func, const char* pretty_func);
...
+        ss << "MIOpenLogScopeTime: " << m_name << " : " << dur.count() << " us";
+        MIOPEN_LOG_I2(ss.str());
+    }
+
+    private:
+    std::string m_name;
+    std::chrono::time_point<std::chrono::high_resolution_clock> m_beg;
+};
+
+#if MIOPEN_LOG_FUNC_TIME_ENABLE
```
atamazov, line 367: [Recommendation] move this `#if` just above line 346 to hide `class LogScopeTime` when it is unnecessary.

atamazov (changes requested):
Almost ready.

src/include/miopen/logger.hpp
```diff
@@ -342,6 +343,33 @@ std::string LoggingParseFunction(const char* func, const char* pretty_func);
...
+        : m_name(std::move(name)), m_beg(std::chrono::high_resolution_clock::now())
+    {
+    }
+    ~LogScopeTime()
+    {
+        auto end = std::chrono::high_resolution_clock::now();
+        auto dur = std::chrono::duration_cast<std::chrono::microseconds>(end - m_beg);
+        std::ostringstream ss;
+        ss << "MIOpenLogScopeTime: " << m_name << " : " << dur.count() << " us";
+        MIOPEN_LOG_I2(ss.str());
```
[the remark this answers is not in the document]
JehandadKhan: updated

src/include/miopen/logger.hpp
JehandadKhan: updated

JehandadKhan:
@atamazov  Your comments have been addressed

atamazov (approved):
LGTM

atamazov (approved):

JehandadKhan:
@atamazov  CI Passes

PR #630, opened 2020-12-01
Restrict static checks to non-gfx908 systems

daniellowell:


atamazov:
Is it to make build faster or to avoid using gfx908 nodes for this check?

daniellowell:
In preparation to upgrade the stack of MI100. Right now Jenkins via the Dockerfile will check the underlying arch before installing HCC. I want to move away from that behavior. So, right now I want to isolate those systems from anything HCC related.

PR #424, opened 2020-09-09
Improve Immediate mode Fallback

atamazov:
Implements (30) from #410  

- Updates the core of the library.
- Opens path for integration of MLP classifier
- Adds very basic `GetWti()` for 3x3 Winograd only.

Basically, this is NFC, except:
- Better Aux Wall time in FAST_FIND mode for unseen 3x3 configs
- Performance of Immediate Mode Fallback improved (for unseen 3x3 cases)

Performance of unseen configs can be further improved by adding `GetWti()` for dynamic solvers and for GEMM.

:warning: ___For detailed understanding of this code, I recommend reviewing each commit individually.___ Each commit has clear description, so the reviewer has luxury to skip what he or she is not interested in (for ex. refactors).

## Summary of code changes

- Added `GetWti()` to `SolverBase` and to `AnySolver`.
- Implemented generic machinery that uses `GetWti()` in Immediate Mode Fallback.
- Implemented very basic ConvBinWinograd3x3U::GetWti().
- Got rid of special semantics of negative solution.time values.
- Got rid of (now unnecessary) FAST_FIND trick that avoids Immediate Mode fallback when `IsWinograd3x3SupportedAndFast()`.

## By-products

These are useful regardless of Immediate Mode Fallback improvements:

- Fixes:
  - ~Fixed GWSS for FFT to return 0 when FFT is disabled.~ (Done in #567)
  - Added missing direction checks in all existing Solvers.
- Refactors:
  - Factored out and fixed some log messages
  - Reworked `GetSolutions()` on Immediate Mode Fallback path.
  - Reworked GWSS on Immediate Mode Fallback path.
  - Reworked `GetSolutionCount()` on Immediate Mode Fallback path.
  - Refined code related to workaround 1430.
  - Factored out `SortWrapper` helper class.

## :cyclone: Local performance & correctness testing

- Radeon VII, ROCm 3.5.
- 93 known FP32 + 93 known FP16 convolution configs
- Normal Find mode + Immediate mode
- OCL backend (FP32 configs only) + HIP backend
- Develop 1db8b1ad3 vs this PR 862f57b0e
- Re-tested develop daa9232fb vs this PR 1b51fcd6f

No regressions. Detailed description of the test is available upon request.

## :cyclone: Evaluation of Initial Iteration Time impact

- ___Installable Release build___, binary cache is empty prior each test.
- Radeon VII, ROCm 3.5.
- 93 unknown FP32 convolution configs (find-db miss)
- HIP backend
[...]

PR #631, opened 2020-12-01
AI Based heuristics for the Find API

JehandadKhan:
This PR depends on #424 

Includes models for gfx906 and gfx908 for all three directions


JehandadKhan:
@atamazov This PR is based off of `imm-fallback-improv` to reduce dependence of unopened PRs. Also, there was no code in `getwti-wino2111` that would have been helpful to this PR.

src/include/miopen/conv/heur/memref.hpp
```diff
@@ -0,0 +1,97 @@
...
+    Tensor2D(int64_t _size0, int64_t _size1) : Tensor2D(_size0, _size1, 0) {}
+    float* data() { return d.data(); }
+    void print()
+    {
+        // TODO: Guard with debug
+        for(auto idx0 = 0; idx0 < size0; ++idx0)
+        {
+            for(auto idx1 = 0; idx1 < size1; ++idx1)
+            {
+                std::cout << d[idx0 * stride0 + idx1 * stride1] << '\t';
```
pfultz2, line 90: Cant this be written with a range-for loop?

```cpp
for(auto&& x:d) std::cout << x << '\t';
```

src/include/miopen/conv/heur/metadata.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <miopen/handle.hpp>
+#include <miopen/problem_description.hpp>
+#include <miopen/conv/heur/memref.hpp>
+
+#include <unordered_map>
+#include <vector>
+
+#define PPCAT_NX(A, B) A##B
+
+#define PPCAT(A, B) PPCAT_NX(A, B)
```
pfultz2, line 39: These macros seem to be never used. Also, macros should be prefixed with `MIOPEN_`.

src/conv/heur/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# save off path w/o virtualenv
+ENV ONNX_MLIR_VERBOSE=1
+ENV OLD_PATH=${PATH}
+ENV VIRTUAL_ENV=/build/myvenv
+RUN python3 -m venv ${VIRTUAL_ENV}
+ENV PATH="${VIRTUAL_ENV}/bin:$PATH"
+ADD requirements.txt /requirements.txt
+RUN pip install -r /requirements.txt
+RUN pip install torch==1.7.0+cpu torchvision==0.8.1+cpu torchaudio==0.7.0 -f https://download.pytorch.org/whl/torch_stable.html
```
pfultz2, line 33: Why arent these listed in the requirements.txt?

src/conv/heur/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM onnxmlirczar/onnx-mlir-build:x86
```
pfultz2, line 1: This should use a digest to ensure this image tag doesn't change `FROM onnxmlirczar/onnx-mlir-build:x86@sha256:c8fdeb4f771d13398516e653ecb7ecc1d439c7ac380fc41927f1a4733c2ae092`

src/conv/heur/Dockerfile
JehandadKhan: I forget the reason, I will try and figure out a way to add them there.

atamazov:
@JehandadKhan 
> @atamazov This PR is based off of `imm-fallback-improv` to reduce dependence of unopened PRs. Also, there was no code in `getwti-wino2111` that would have been helpful to this PR.

It's Ok so far, as a concept check.

However I have the following design in mind. If AI-based classifier fails to find the "fastest" solution due to some reason, then `getwti-wino2111` should be used as a "spare". I can perform integration of AI-based classifier on top of `getwti-wino2111`.

`getwti-wino2111` optimizes Immediate Mode Fallback better than #424 and fully tested.

src/include/miopen/conv/heur/memref.hpp
```diff
@@ -0,0 +1,97 @@
...
+        for(auto idx0 = 0; idx0 < size0; ++idx0)
+        {
+            for(auto idx1 = 0; idx1 < size1; ++idx1)
+            {
+                std::cout << aligned[idx0 * stride0 + idx1 * stride1] << '\t';
+            }
+            std::cout << std::endl;
+        }
+    }
+};
```
jerryyin, line 61: I was randomly browsing PRs in MIOpen and this memref catched my eye. The interaction with mlir will soon require me to also come up with a memref type so I'd hope we can together come up with one robust solution that can be used in shared scenarios.

Clarification question: Does this memref interface with external dependencies?

The one that is defined in mlir is [here](https://github.com/llvm/llvm-project/blob/master/mlir/include/mlir/ExecutionEngine/CRunnerUtils.h#L115-L155). That memref is almost similar to yours except that the first field and second field has swapped their positions: In MLIR, first field is unaligned. Although in your use cases, I always see the two fields getting similar values but I want to make sure they do meant to swap.

Is it possible to use the templatized version for extendability into different dimensions? Attaching the MLIR definition below:

```cpp
template <typename T, int N>
struct StridedMemRefType {
  T *basePtr;
  T *data;
  int64_t offset;
  int64_t sizes[N];
  int64_t strides[N];
  // This operator[] is extremely slow and only for sugaring purposes.
  StridedMemRefType<T, N - 1> operator[](int64_t idx) {
    StridedMemRefType<T, N - 1> res;
    res.basePtr = basePtr;
    res.data = data;
    res.offset = offset + idx * strides[0];
    dropFront<N>(sizes, res.sizes);
    dropFront<N>(strides, res.strides);
    return res;
  }
};

```

src/conv/heur/Dockerfile
JehandadKhan: The above pip command loads the specified packages from paths specified in the url after the -f switch. I am not aware of any way to specify the same in the requirements.txt file. I tried using similar syntax but it did not work.

src/include/miopen/conv/heur/memref.hpp
JehandadKhan: @jerryyin  Thank you for pointing this out, I was not aware of that implementation as well as the bug you pointed out. I have removed my implementation and replaced it with the one in LLVM. Of course I have copied it over since I don't wish MIOpen to depend on LLVM for a single struct. 

I would appreciate if you can take a second look.

jerryyin: The new changes that pull in MemRef2D looks good!

I don't think I understand all use case of MemRef2D in your PR so leave this for you to decide: In MLIR, Memref2d has exposed a lot of the internals of the data layout. When this needs to be exposed on a binary interface, an unranked memref which exposed fewer implementation details are typically used. If the Memref is used across binary interface, I'd suggest you to use unranked memref in that interface. If not, you are all good.

atamazov:

PR #633, opened 2020-12-02
Workaround for SWDEV-262823

atamazov:
- Trim target features from the device name, e.g. `gfx908:sramecc+:xnack-` -> `gfx908`.
  - [Note] This PR is NFC if target features are not reported by runtime.
- Optimize for speed to compensate for additional calculations in `GetDeviceName()`.

### DIFF
--- a/src/include/miopen/device_name.hpp
+++ b/src/include/miopen/device_name.hpp
@@ -59,9 +59,9 @@ std::string inline GetDeviceNameFromMap(const std::string& name)
     if(p_asciz != nullptr && strlen(p_asciz) > 0)
         return {p_asciz};
 
-    auto device_name_iterator = device_name_map.find(name);
-    if(device_name_iterator != device_name_map.end())
-        return device_name_iterator->second;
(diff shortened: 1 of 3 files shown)


src/include/miopen/device_name.hpp
```diff
@@ -49,32 +51,26 @@ std::string inline GetDeviceNameFromMap(const std::string& name)
         {"Fiji", "gfx803"},
         {"gfx800", "gfx803"},
         {"gfx802", "gfx803"},
-        {"gfx803", "gfx803"},
```
daniellowell, line 52: Why do we need to delete these?

atamazov: This one "converts" gfx803 to gfx803. Obviously, this kind of conversion is not necessary.

daniellowell: Oh I see.

atamazov:
CI tests passed!

daniellowell (approved):

PR #634, opened 2020-12-02
Adding batchnorm fix for GFX1030

muralinr:
Adding workaround fix for GFX1030 ASM kernel

### DIFF
--- a/src/ocl/batchnormocl.cpp
+++ b/src/ocl/batchnormocl.cpp
@@ -36,6 +36,7 @@
 /// \todo Get rid of this during implementation of #1938 (60)
 #include <miopen/convolution.hpp>
 #include <miopen/mlo_internal.hpp>
+#include <miopen/stringutils.hpp>
 
 #define WORKAROUND_SWDEV_253606 1
 #include <chrono>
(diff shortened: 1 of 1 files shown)


src/ocl/batchnormocl.cpp
```diff
@@ -1051,7 +1051,8 @@ void BatchNormBackward(Handle& handle,
                 std::string parms;
 
                 if((n > 64) && (n % 2 == 0) && (variant == 3) && (bfpmixparm) && (useSaved) &&
-                   ctx.use_asm_kernels && ctx.rmv.IsV2orV3())
+                   ctx.use_asm_kernels && ctx.rmv.IsV2orV3() &&
+                   (handle.GetDeviceName() != "gfx1030"))
```
atamazov, line 1055:
```suggestion
                   !StartsWith(name, "gfx1"))
```
or
```suggestion
                   (StartsWith(name, "gfx8") || StartsWith(name, "gfx9")))
```

And add `#include <miopen/stringutils.hpp>` at the beginning of file.

atamazov (changes requested):
Let's foresee other gfx10 derivatives and, possibly, other new gfx architectires, at once.

src/ocl/batchnormocl.cpp
muralinr: Ok. Done..

atamazov (approved):

daniellowell:
@muralinr was this tested on Navi21 and confirmed to solve the issue on the JIRA?

muralinr:
[AMD Official Use Only - Internal Distribution Only]

Yes Daniel.. I tested it on Navi21..

Thanks,
Murali

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Issue #639, opened 2020-12-04
potential immediate bwd crash


carlushuang:
found a potential crash in immediate mode API of bwd

```
void ConvolutionDescriptor::ConvolutionBackwardImmediate(Handle& handle,
                                                         const TensorDescriptor& dyDesc,
                                                         ConstData_t dy,
                                                         const TensorDescriptor& wDesc,
                                                         ConstData_t w,
                                                         const TensorDescriptor& dxDesc,
                                                         Data_t dx,
                                                         Data_t workSpace,
                                                         std::size_t workSpaceSize,
                                                         solver::Id solver_id) const
{
    MIOPEN_LOG_I("solver_id = " << solver_id.ToString() << ", workspace = " << workSpaceSize);
    auto tensors = ConvBwdTensors{dyDesc, dy, wDesc, w, dxDesc, dx};

    ...
    ConvBwdCheckNumerics(handle, tensors, &beta, [&]() {
        ...

        auto ctx = ConvolutionContext{dxDesc, wDesc, dyDesc, *this, conv::Direction::BackwardData};

```
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/e06e6c0c6e8c827419e8dce8de0af484f86fc715/src/ocl/convolutionocl.cpp#L2739

Right after context is created by `auto ctx=...`, where should be a `ctx.SetStream(&handle);` to set stream of the newly created context (fwd/wrw immediate already have this setStream, but not bwd). Otherwise the `stream` member of this context will always be `nullptr`, and will likely crash in later software flow.

Currently no crash found in miopenDriver with immediate mode, maybe because of every solver is within DB and will not go create a new invoker (which will require the `stream` member)
https://github.com/asleepzzz/nhwc_shuffle/blob/8d41a4482874a6f943d491d5fd4753d0a89ebb49/src/mlo_dir_conv.cpp#L81

But there is another PR: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583, which require utilize of immediate API to call reference kernel, with solver_id. If not change like descibe above, will lead to crash.

currently this change is included in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/583 and no dedicated PR to fix this (since no crash for now)

cc @asroy @atamazov

atamazov:
Thank you @carlushuang , nice catch! Could you please make dedicated PR and assign blocker urgency for it. Thanks.

carlushuang:
[...]

PR #641, opened 2020-12-05
Fix potential bwd immediate crash by add missing setStream() call to context

carlushuang:
This fixes https://github.com/ROCmSoftwarePlatform/MIOpen/issues/639

Right after context is created by `auto ctx=...`, there should be a `ctx.SetStream(&handle);` to set stream of the newly created context (fwd/wrw immediate already have this setStream, but not bwd). Otherwise the `stream` member of this context will always be nullptr, and will likely crash in later software flow.


atamazov (approved):
LGTM!

DrizztDoUrden (approved):

PR #644, opened 2020-12-07
[igemm]fix tuning paramter of spare(issue 598 and issue 636)

asleepzzz:
this PR is the same with 601,but add more protection, so drop 601 now

we add more protection to some corner cases ,and prevent invoker not calling isValidConfig
btw, increase the tuning range

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
@@ -203,7 +203,7 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmBBlockCopyPerformancePara
         // calculate threadwise copy size
         int b_data_per_thread_copy = (GemmKPerBlock * GemmNPerBlock * GemmKPACKSize) / BlockSize;
 
-        if(GemmKPACKSize > b_data_per_thread_copy) // will fix in future
+        if(GemmKPACKSize > b_data_per_thread_copy && ctx.IsFp32()) // will fix in future
             MIOPEN_THROW("invalid performance parameter");
 
         if(!(b_data_per_thread_copy > 0))
--- a/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
@@ -102,7 +102,7 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmABlockCopyPerformancePara
         SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, a_data_per_thread_copy);
         int mMax             = GemmMPerBlock / BlockSize;
         if(mMax < 1)
-            mMax = 1;
+            mMax             = 1;
         SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, mMax);
 
         // decide threadwise copy lengths
@@ -212,7 +212,7 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmBBlockCopyPerformancePara
         SrcDataPerRead_GemmN = gcd(SrcDataPerRead_GemmN, b_data_per_thread_copy);
         int nMax             = GemmNPerBlock / BlockSize;
         if(nMax < 1)
-            nMax = 1;
+            nMax             = 1;
         SrcDataPerRead_GemmN = gcd(SrcDataPerRead_GemmN, nMax);
 
         const auto data_per_thread_copy_gemmn = SrcDataPerRead_GemmN;
@@ -404,7 +404,7 @@ bool PerformanceImplicitGemmBwdDataV4R1Xdlops::IsReallyValid(const ConvolutionCo
     if(!valid)
         return false;
 
-    std::size_t lds_size      = 0;
+    std::size_t lds_size = 0;
     std::tie(lds_size, valid) = CalculateLdsNumberOfByte(ctx);
 
     return (valid and lds_size <= 64 * 1024);
--- a/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
@@ -203,8 +203,8 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmBBlockCopyPerformancePara
             MIOPEN_THROW("invalid performance parameter");
 
         // GemmBBlockCopySrcDataPerRead_GemmN also bounded by size of threadwise copy
-        SrcDataPerRead_GemmN                  = gcd(SrcDataPerRead_GemmN, b_data_per_thread_copy);
-        int nMax                              = std::max(GemmNPerBlock / BlockSize, 1);
+        SrcDataPerRead_GemmN = gcd(SrcDataPerRead_GemmN, b_data_per_thread_copy);
+        int nMax             = std::max(GemmNPerBlock / BlockSize, 1);
         SrcDataPerRead_GemmN = gcd(SrcDataPerRead_GemmN, nMax);
 
         const auto data_per_thread_copy_gemmn = SrcDataPerRead_GemmN;
--- a/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
@@ -100,7 +100,9 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmABlockCopyPerformancePara
 
         // GemmABlockCopySrcDataPerRead_GemmM also bounded by size of threadwise copy
         SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, a_data_per_thread_copy);
-        int mMax             = std::max(GemmMPerBlock / BlockSize, 1);
+        int mMax             = GemmMPerBlock / BlockSize;
+        if(mMax < 1)
+            mMax = 1;
         SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, mMax);
 
         // decide threadwise copy lengths
@@ -131,6 +133,8 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmABlockCopyPerformancePara
                 MIOPEN_THROW("invalid performance parameter");
         }
 
+        if(DstDataPerWrite_GemmKPack > data_per_thread_copy_gemmkpack)
+            DstDataPerWrite_GemmKPack = data_per_thread_copy_gemmkpack;
         DstDataPerWrite_GemmKPack = gcd(DstDataPerWrite_GemmKPack, data_per_thread_copy_gemmkpack);
 
         if(!(GemmKPerBlock % data_per_thread_copy_gemmk == 0 &&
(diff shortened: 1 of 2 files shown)


src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
```diff
@@ -100,6 +100,8 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmABlockCopyPerformancePara
 
         // GemmABlockCopySrcDataPerRead_GemmM also bounded by size of threadwise copy
         SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, a_data_per_thread_copy);
+        int mMax             = std::max(GemmMPerBlock / BlockSize, 1);
+        SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, mMax);
```
asroy, line 104: If GemmMPerBlock < BlockSize, then mMax=1, and SrcDataPerRead_GemmM will always be 1. This is not ideal.

Is there a better logic?

asroy (changes requested):

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asleepzzz: sometimes it’s 2
So use max=1 may not enough

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
```diff
@@ -852,7 +860,7 @@ ConvSolution ConvHipImplicitGemmBwdDataV4R1Xdlops::GetSolution(
 
     if(!config.IsReallyValid(ctx))
     {
-        MIOPEN_LOG_E("invalid performance parameter");
+        MIOPEN_THROW("invalid performance parameter");
```
atamazov, line 863: Please explain how invalid PerformanceConfig can reach this point.

atamazov (dismissed):

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asleepzzz: you are right, I will modify it

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asleepzzz: I need time to reproduce it.....

atamazov: Do you mean that this is impossible unless a bug in the library?

asleepzzz: I can't reproduce it now, back to log and watch it

atamazov: ?

daniellowell:
@asleepzzz Placing `on_hold` label until we understand what is the purpose of this PR and we have answers to questions posed in this PR.

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asleepzzz: I ever reach this point before,
at last time searching for config,

asleepzzz:
> @asleepzzz Placing `on_hold` label until we understand what is the purpose of this PR and we have answers to questions posed in this PR.

Without this PR, some cases failed, this PR is urgent, artem wants answers of throw, but they are different things, due to I can’t reproduce the issue which throw can fix it,I remove the throw parts and still keep other parts

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asleepzzz: The throw part is another issue and I can’t reproduce, but it’s impossible into this code section, when the issue happened, if I add throw , test pass, but since I can’t find how to reproduce it , I remove the throw part, I don’t know if this answer you ....

asleepzzz:
Without this PR , we can’t tune a good db,
after this PR into the develop we need a new db ,it must take lots of time

zjing14 (approved):

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asleepzzz: already changed

zjing14:
@daniellowell We should merge this PR as soon as possible. We need to retune after it. We may create a post-merge issue if having some comments.

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asroy: @asleepzzz This doesn't fix the issue. The behaviour will be the same as your previous commit.

I think this is what you are trying to do
```
SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, a_data_per_thread_copy); // don't change how SrcDataPerRead_GemmM is calculated

const auto a_data_per_thread_copy_gemmm_min = max(GemmMPerBlock / BlockSize, 1); // this is the minimal data a thread need to copy on GemmM dimension, it should no be smaller than 1 or GemmMPerBlock / BlockSize

const auto a_data_per_thread_copy_gemmm = lcm(a_data_per_thread_copy_gemmm_min, SrcDataPerRead_GemmM) // a_data_per_thread_copy_gemmm should be the least common multiple of a_data_per_thread_copy_gemmm_min and SrcDataPerRead_GemmM
```

If these logic matches what you are trying to do, please use them and update the logic for B-block-copy too.

asroy (changes requested):
@asleepzzz I think the logic still has the same issue, please review my comment and let me know if it's right or not ASAP.

asleepzzz:
> @asleepzzz I think the logic still has the same issue, please review my comment and let me know if it's right or not ASAP.

Hi Chao ,you may watch the legacy code
        int mMax             = GemmMPerBlock / BlockSize;
        if(mMax < 1)
            mMax             = 1;
        SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, mMax);

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asleepzzz: Hi Chao,
the logic is in below ,because copy SrcDataPerRead_GemmM to a_data_per_thread_copy_gemmm,
so the results is also correct :
        int mMax             = GemmMPerBlock / BlockSize;
        if(mMax < 1)
            mMax             = 1;
        SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, mMax);

        // decide threadwise copy lengths
        const auto a_data_per_thread_copy_gemmm = SrcDataPerRead_GemmM;

asleepzzz: fwd use lcm for gemmKpack due to it's contiguous for gemmkpack, 
I use gcd for gemmM ,I think they are the same

atamazov (commented):
Continuation of https://github.com/ROCmSoftwarePlatform/MIOpen/pull/644#discussion_r541830832

[Off-topic] Please do not misunderstand me. I usually try to be as straightforward as possible when reviewing. I did not ask to remove MIOPEN_THROW, I literally asked questions, simply to understand the intent of this PR.

Now back to reviewing. Am I correct that:
- While ago you've seen the cases when `config.IsReallyValid(ctx) == false` in `GetSolution()`
- If this happens, then it means an bug in the library.

If I'm correct above, please keep on throwing. Basically, this means internal error, so `MIOPEN_THROW(miopenStatusInternalError, ...)` should be used.

I have no objections wrt this PR.

asleepzzz:
> Continuation of [#644 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/644#discussion_r541830832)
> 
> [Off-topic] Please do not misunderstand me. I usually try to be as straightforward as possible when reviewing. I did not ask to remove MIOPEN_THROW, I literally asked questions, simply to understand the intent of this PR.
> 
> Now back to reviewing. Am I correct that:
> 
> * While ago you've seen the cases when `config.IsReallyValid(ctx) == false` in `GetSolution()`
> * If this happens, then it means an bug in the library.
> 
> If I'm correct above, please keep on throwing. Basically, this means internal error, so `MIOPEN_THROW(miopenStatusInternalError, ...)` should be used.
> 
> I have no objections wrt this PR.

Hi atam, never mind ,
Yes I ever saw config.IsReallyValid(ctx) == false in GetSolution(),
I knew it's strange ,so I want to reproduce and we may find some bug in the library
unfortunately, I can't reproduce it again ,but if add this throw in solver, we can avoid getting wrong solution
I will keep tracking the symptom

daniellowell:
@asleepzzz @asroy We need this PR in shape. It is blocking updating ROCm mainline.

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asroy: > If GemmMPerBlock < BlockSize, then mMax=1, and SrcDataPerRead_GemmM will always be 1. This is not ideal.

However, this issue would still exist, if using your current logic, isn't it?

asleepzzz:
[performance compare.xlsx](https://github.com/ROCmSoftwarePlatform/MIOpen/files/5708291/performance.compare.xlsx)
I know Chao think  lcm is better, but after testing, 
some configs better, some worse, the attachment is performance data (time ms) 
for all configs, sum of time current logic is not worse than lcm logic

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
```diff
@@ -92,15 +92,10 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmABlockCopyPerformancePara
...
-
-        if(!(a_data_per_thread_copy > 0))
-            MIOPEN_THROW("invalid performance parameter");
-
-        // GemmABlockCopySrcDataPerRead_GemmM also bounded by size of threadwise copy
-        SrcDataPerRead_GemmM = gcd(SrcDataPerRead_GemmM, a_data_per_thread_copy);
+        auto a_data_per_thread_copy =
+            std::max(1, (GemmKPerBlock * GemmMPerBlock * GemmKPACKSize) / BlockSize);
 
+        a_data_per_thread_copy = lcm(a_data_per_thread_copy, SrcDataPerRead_GemmM);
```
asroy, line 98: a_data_per_thread_copy should not be changed here, because its meaning is how many data each thread would copy, and it's decided by A block matrix size and BlockSize. But we can change a_data_per_thread_copy_gemmm (how many data each thread copy along GemmM dimension)

```
const auto a_data_per_thread_copy_gemmm_min = max(GemmMPerBlock / BlockSize, 1); // this is the minimal data a thread need to copy on GemmM dimension, it should be no smaller than 1 or GemmMPerBlock / BlockSize

const auto a_data_per_thread_copy_gemmm = lcm(a_data_per_thread_copy_gemmm_min, SrcDataPerRead_GemmM) // a_data_per_thread_copy_gemmm should be the least common multiple of a_data_per_thread_copy_gemmm_min and SrcDataPerRead_GemmM

if(a_data_per_thread_copy % a_data_per_thread_copy_gemmm != 0)
            MIOPEN_THROW("invalid performance parameter"); // sanity check
```

asroy (changes requested):

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
```diff
@@ -195,18 +192,23 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmBBlockCopyPerformancePara
...
+        int b_data_per_thread_copy =
+            std::max(1, (GemmKPerBlock * GemmMPerBlock * GemmKPACKSize) / BlockSize);
 
         if(!(b_data_per_thread_copy > 0))
             MIOPEN_THROW("invalid performance parameter");
 
-        // GemmBBlockCopySrcDataPerRead_GemmN also bounded by size of threadwise copy
-        SrcDataPerRead_GemmN                  = gcd(SrcDataPerRead_GemmN, b_data_per_thread_copy);
+        b_data_per_thread_copy = lcm(SrcDataPerRead_GemmN, b_data_per_thread_copy);
+        if(BlockSize > GemmNPerBlock && GemmKPACKSize > BlockSize / GemmNPerBlock)
```
asroy, line 202: [NOT URGENT]This logic "BlockSize > GemmNPerBlock" doesn't seem capture the exact condition why such performance parameter is not valid

Please fix in future PR

asroy (approved):
This PR should be considered a hot-fix. It's OK to merge, so mainline can be unblocked.

I will open an follow-up issue to track the remaining issues

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
asleepzzz: ok

asroy:
@daniellowell This PR is ready to merge

asleepzzz:
CI passed and already reviewed
@daniellowell

PR #626, opened 2020-11-25
Winograd RxSf2x3 solver V21_1_2 update

Slimakanzer:
Winograd V21_1_2 changes compared to #175 :
* Fixed bugs
* No performance regression compared to the [previous version](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/175)
* Supports extended cases set (solver's **IsApplicable** will be updated in the next PR)
* Tensor offset support (64 bit byte offsets)

PR #653, opened 2020-12-10
Adding Winograd v21_1_2 gfx10 implementation

Slimakanzer:
Continuation of #626. This PR provides implementation of Winograd v21_1_2 for gfx10.
⚠️This PR does not change the performance model. It will be done in a subsequent PR.

## Summary of code changes
* Added `Conv_Winograd_v21_1_2_gfx10_***` gfx10 shader implementations
* Updated `Conv_Winograd_v21_1_2_metadata.inc` to support gfx10 dispatch parameters
* Updated conv_winoRxF_f2x3's `IsApplicable` and `GetSolution` to support gfx10

## Local verification
* gfx1030, ROCm 3.9 + gfx908, ROCm 3.5.1
* OCL + HIP backend
* fp16 + fp32

W | H | c | n | k | x | y | p | q | u | v | g | dir
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- 
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | fwd
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | bwd
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | wrw
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 1 | 1 | 32 | fwd
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 1 | 1 | 32 | bwd
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 1 | 1 | 32 | wrw
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 2 | 2 | 1 | fwd
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 2 | 2 | 1 | bwd
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 2 | 2 | 32 | fwd
14 | 14 | 1024 | 32 | 1024 | 3 | 3 | 1 | 1 | 2 | 2 | 32 | bwd
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | fwd
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | bwd
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | wrw
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 32 | fwd
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 32 | bwd
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 32 | wrw
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 2 | 2 | 1 | fwd
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 2 | 2 | 1 | bwd
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 2 | 2 | 32 | fwd
56 | 56 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 2 | 2 | 32 | bwd
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | fwd
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | bwd
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 1 | wrw
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 64 | fwd
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 64 | bwd
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 1 | 1 | 64 | wrw
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 2 | 2 | 1 | fwd
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 2 | 2 | 1 | bwd
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 2 | 2 | 64 | fwd
28 | 28 | 512 | 32 | 512 | 3 | 3 | 1 | 1 | 2 | 2 | 64 | bwd


Kirpich30000:
Also it's better to avoid copy-past and reduce number of `.s` files:
1) remove `gfx*` suffix from kernel names and `*.s` files
2) inside each `*.s` file use `.if` to include propper `.inc` file

src/solver/conv_winoRxS_f2x3.cpp
```diff
@@ -354,9 +354,10 @@ bool ConvBinWinogradRxSf2x3::IsApplicable(const ConvolutionContext& params) cons
         return false;
 
     const auto name = params.GetStream().GetDeviceName();
-    if(!(StartsWith(name, "gfx9")))
+    if(!(StartsWith(name, "gfx9") || StartsWith(name, "gfx10")))
         return false;
-    if(params.IsFp16() && !(StartsWith(name, "gfx906") || StartsWith(name, "gfx908")))
+    if(params.IsFp16() &&
+       !(StartsWith(name, "gfx906") || StartsWith(name, "gfx908") || StartsWith(name, "gfx10")))
```
Kirpich30000, line 360: gfx101* doesn't support fp16 dot. It should be enabled only on gfx103*

Kirpich30000 (dismissed):

Kirpich30000:
Note that performance might be affected by #654

daniellowell:
@Kirpich30000 @Slimakanzer For planning purposes, how many more Winograd kernels are we expecting to flow into MIOpen for Navi 21?

Kirpich30000:
There are plans to add support for F2x3 single pass Winograd. ETA is unknown. Multipass is on hold for now.

Kirpich30000 (approved):
LGTM

src/kernels/Conv_Winograd_v21_1_2_metadata.inc
```diff
@@ -146,22 +153,30 @@ __sgpr_reserve_flatscr_default = 0
...
-	PROLOG_KERNEL_DESCRIPTOR miopenSp3AsmConv_v21_1_2_\kernel_name_postfix
+	PROLOG_KERNEL_DESCRIPTOR_WRAPPER %machine_version, \kernel_name_postfix
 .endm
 
 .macro KERNEL_EPILOG kernel_name_postfix
-	EPILOG_KERNEL_DESCRIPTOR miopenSp3AsmConv_v21_1_2_\kernel_name_postfix
+	EPILOG_KERNEL_DESCRIPTOR_WRAPPER %machine_version, \kernel_name_postfix
 .endm
 
+.set machine_version, .amdgcn.gfx_generation_number
```
atamazov, line 178: Do we really need `machine_version`? I think we can use `.amdgcn.gfx_generation_number` in `KERNEL_PROLOG` and `KERNEL_EPILOG` directly, right?

src/solver/conv_winoRxS_f2x3.cpp
```diff
@@ -451,9 +451,11 @@ static bool IsApplicableBase(const ConvolutionContext& params)
         return false;
 
     const auto name = params.GetStream().GetDeviceName();
-    if(!(StartsWith(name, "gfx9")))
+    if(!(StartsWith(name, "gfx9") || StartsWith(name, "gfx10")))
         return false;
-    if(params.IsFp16() && !(StartsWith(name, "gfx906") || StartsWith(name, "gfx908")))
+    if(params.IsFp16() &&
+       !(StartsWith(name, "gfx906") || StartsWith(name, "gfx908") || StartsWith(name, "gfx1011") ||
+         StartsWith(name, "gfx1012") || StartsWith(name, "gfx103")))
```
atamazov, line 458: Do these kernels work with all these gfx10 derivatives? I suspect some of these require XNACK ON.

src/solver/conv_winoRxS_f2x3.cpp
```diff
@@ -572,15 +578,25 @@ ConvBinWinogradRxSf2x3::GetSolution(const ConvolutionContext& params,
...
+        kernel.comp_options += std::string(" -mcumode -mwavefrontsize64");
+    }
 
     if(params.IsFp32())
-        kernel_postfix += "_fp32";
+        kernel_postfix = "_fp32";
     else
     {
-        kernel_postfix += "_fp16_dot2_edc";
+        kernel_postfix = "_fp16_dot2_edc";
```
atamazov, line 599: Let's be as expressive as C/C ++ allows us.
```suggestion
        const std::string kernel_postfix = params.IsFp32() ? "_fp32" : "_fp16_dot2_edc";
```

atamazov (changes requested):
Well done, but there are a couple of questions.

src/solver/conv_winoRxS_f2x3.cpp
Slimakanzer: Kernels work correctly on all these gfx10 regardless of xnack.

src/kernels/Conv_Winograd_v21_1_2_metadata.inc
Slimakanzer: `machine_version` also used in each shader declaration (_**Conv_Winograd_v21_1_2_xxx.s**_ files) to determine which kernel to use (gfx9 or gfx10).

Here is an [example](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/148bb845d8d557e7fa5a7dce3a9708d5bd134661/src/kernels/Conv_Winograd_v21_1_2_fp16_dot2_edc_dilation2.s#L30-L34)

src/solver/conv_winoRxS_f2x3.cpp
Slimakanzer: Thank you for your help, fixed in 148bb84

src/kernels/Conv_Winograd_v21_1_2_metadata.inc
atamazov: Thanks. The question is: if we always set `machine_version` to `.amdgcn.gfx_generation_number`, then why we need this symbol (`machine_version`) at all?

atamazov (dismissed):
LGTM!

Kirpich30000 (dismissed):
LGTM

src/kernels/Conv_Winograd_v21_1_2_metadata.inc
Slimakanzer: This abstraction implied the ability to add COv2 support. However, at this moment the solver only supports COv3, so you are right and this symbol is not needed here.

atamazov: Correct.

NOTE: For COv2/v3 kernels, we use `.option.machine_version_major/minor/stepping` (see `rocm_version.inc`).

atamazov (approved):

Kirpich30000 (approved):
LGTM

src/kernels/Conv_Winograd_v21_1_2_metadata.inc
Slimakanzer: Thank you for explanation. I'll keep this in mind for the future

atamazov:
CI tests passed

Issue #557, opened 2020-10-30
gfx1030 not work properly when using buffer instructions


ltqin:
Although gfx1030 supports buffer instructions ，but it not work properly  when we use  the  corresponding llvm intrinsic functions . Because those llvm intrinsic functions  are deprecated by compiler team, so I record this issue here.  ~and record this is for writing a workaround~

----
@atamazov EDIT: In short, we need to use LLVM intrinsics correctly, at least.

atamazov:
Originated from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/545#discussion_r512298129

asroy:
This is a MIOpen issue, because we need to start using the new LLVM intrinsic.
@atamazov

atamazov:
So this is workaround for MIOpen. We do not need to check compiler version, but the fix is still workaround.

atamazov:
AFAICS, the workaround will fix correctness for gfx10. The objective is a full-blown fix that should take into use new LLVM intrinsics and, thus, restore performance.

ltqin:
@asroy you mean need to test  out intrinsic  like "llvm.amdgcn.raw.buffer.load.v4f32"?
https://github.com/llvm/llvm-project/search?q=raw+buffer+load

asroy:
> @asroy do you mean we need to test out intrinsic like "llvm.amdgcn.raw.buffer.load.v4f32"?
> https://github.com/llvm/llvm-project/search?q=raw+buffer+load

Yes, it's not fully tested yet.

atamazov:
@ltqin @asroy Do we have any plans wrt this ticket?

@junliume This is not a blocker since WORKAROUND_MIOPEN_ISSUE_557 is merged in from #660. I recommend setting
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/performance
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/workaround 
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_low.

atamazov:
@ltqin @asroy @junliume Ping

atamazov:
Right now, most likely we need to extend the W/A to cover gfx11.

ppanchad-amd:
@ltqin Please check if this still an issue with the latest ROCm 6.0.2? If issue not reproducible, please closes ticket.  Thanks!

ltqin:
[...]

PR #660, opened 2020-12-22
Disable the using of llvm instrinsic for Generic-Reduction Kernels on gfx1030

fronteer:
Disable  CK_USE_AMD_BUFFER_ADDRESSING when compiling Generic-Reduction kernels on gfx1030

EDIT by @atamazov: Workaround for #557

### DIFF
--- a/src/reducetensor.cpp
+++ b/src/reducetensor.cpp
@@ -38,6 +38,11 @@
 #include <ostream>
 #include <iostream>
 
+// although gfx1030 supports buffer instructions,but it does not work properly when we using
+// corresponding llvm intrinsic functions
+// so using those llvm intrinsic functions is disabled on gfx1030
+#define WORKAROUND_MIOPEN_ISSUE_557 1
+
 namespace miopen {
 
 enum ReductionMethod_t
@@ -551,6 +556,12 @@ void ReduceTensorDescriptor::ReduceTensor(const Handle& handle,
 
     // to remove the warning from clang-tidy checking
     param += " -DMIOPEN_USE_FP32=0 -DMIOPEN_USE_FP16=0 ";
+
+#if WORKAROUND_MIOPEN_ISSUE_557
+    const auto device_name = handle.GetDeviceName();
+    if ( StartsWith(device_name, "gfx1030") )
+	 param += " -DCK_USE_AMD_BUFFER_ADDRESSING=0 ";
+#endif
 
     std::string program_name = "gridwise_generic_reduction.cpp";
     std::string algo_name    = "generic_reduce_tensor";
--- a/src/reducetensor.cpp
+++ b/src/reducetensor.cpp
@@ -30,6 +30,7 @@
 #include <miopen/reduce_common.hpp>
 #include <miopen/handle.hpp>
 #include <miopen/reducetensor.hpp>
+#include <miopen/stringutils.hpp>
 
 #include <cassert>
 #include <cstddef>
--- a/src/reducetensor.cpp
+++ b/src/reducetensor.cpp
@@ -38,9 +38,6 @@
 #include <ostream>
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


src/reducetensor.cpp
```diff
@@ -38,6 +38,11 @@
 #include <ostream>
 #include <iostream>
 
+// although gfx1030 supports buffer instructions,but it does not work properly when we using
+// corresponding llvm intrinsic functions
+// so using those llvm intrinsic functions is disabled on gfx1030
```
atamazov, line 43: Please remove the comment.

src/reducetensor.cpp
```diff
@@ -552,6 +557,12 @@ void ReduceTensorDescriptor::ReduceTensor(const Handle& handle,
     // to remove the warning from clang-tidy checking
     param += " -DMIOPEN_USE_FP32=0 -DMIOPEN_USE_FP16=0 ";
 
+#if WORKAROUND_MIOPEN_ISSUE_557
+    const auto device_name = handle.GetDeviceName();
+    if ( StartsWith(device_name, "gfx1030") )
```
atamazov, line 562:
```suggestion
    if ( StartsWith(handle.GetDeviceName(), "gfx10") )
```
Why:
- I think that old intrinsics do not work for all gfx10 targets.
- We do not need `device_name` after this. Let's use the expressiveness of the C++ language.

atamazov (changes requested):

src/reducetensor.cpp
fronteer: This comment and definition refers to the same thing in src/solver/implicitgemm_util.hpp.  Here Imy codes does not use the definition in src/solver/implicitgemm_util.hpp because I am not sure if it is suitable to include this header file.

atamazov: The comment is incorrect, so just remove it, please. Don't mind other header.

asroy:
@fronteer Please address  @atamazov 's comment, I don't have any

src/reducetensor.cpp
fronteer: Removed.

src/reducetensor.cpp
fronteer: Corrected. Thanks

atamazov (approved):

asroy:
@fronteer Formatting issue

daniellowell:
@fronteer This PR still has formatting issues. Please fix them.

fronteer:
Format issue fixed

atamazov (approved):
LGTM

atamazov:
@asroy Please review, thanks.

src/reducetensor.cpp
```diff
@@ -552,6 +554,11 @@ void ReduceTensorDescriptor::ReduceTensor(const Handle& handle,
     // to remove the warning from clang-tidy checking
     param += " -DMIOPEN_USE_FP32=0 -DMIOPEN_USE_FP16=0 ";
 
+#if WORKAROUND_MIOPEN_ISSUE_557
+    if(StartsWith(handle.GetDeviceName(), "gfx10"))
```
atamazov, line 558: Compilation fails. Please add `#include <miopen/stringutils.hpp>`

atamazov (changes requested):

src/reducetensor.cpp
fronteer: Added.   Strange!  Probably because the "develop" branch I am basing my test branch on is less latest,  and which implicitly has the stringutils.hpp included through its tree of headers

atamazov: 🤷‍♂️

atamazov (approved):
LGTM!

atamazov:
All tests passed except "Packaging" which is ongoing right now.

PR #672, opened 2021-01-07
Feature jenkinsfile

okakarpa:
1. Enabled reboot for all the stages in Jenkins-file  except static checks
2. Every stage will go through reboot triggers this Jenkins-job (http://micimaster.amd.com/view/automation/job/reboot-slaves) irrespective of the build-step getting executed as fail or pass.
3. Keeping track of all the Ansible playbooks here (https://github.com/ROCmSoftwarePlatform/rocAutomation)


atamazov (changes requested):
Please explain why this change (~1500 lines of code) is necessary

Issue #560, opened 2020-10-30
Tracking changes requested in merged PR #307


alexandraBara:
@atamazov  I will address changes requested here: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/307 in an upcoming PR as we want to unblock this PR: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/533

also to address:

Find IsCacheDisabled() in binary_cache.cpp
Rename it to IsBinaryCacheDisabled()
Make it global (remove static and add its declaration to binary_cache.hpp)
Use it here instead of MIOPEN_ENABLE_SQLITE_KERN_CACHE

TODO: src/ocl/convolutionocl.cpp perf_db.emoty() check needs to be:

perf_db.emty() and device_cu/compile_only etc not set --> throw error
Otherwise we are still seeing "..incorrect params" thrown from here bc Invokers dont populate the perf_db during compile_only

atamazov:
@alexandraBara Direct links to relevant items:
- [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/533#discussion_r513783447
- [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/533#discussion_r515365226
- [ ] leftovers of https://github.com/ROCmSoftwarePlatform/MIOpen/pull/307#pullrequestreview-518920649

atamazov:
@alexandraBara Any progress?

atamazov:
- [ ] :bug: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/533#pullrequestreview-527654975

alexandraBara:
> @alexandraBara Any progress?

@atamazov  I have been tuning and I am about to finish with it. I picked up this task today and started working on it. Should have a PR this week.

atamazov:
@alexandraBara Good, thank you!

daniellowell:
@alexandraBara @atamazov Is this `urgency_blocker` ticket ready to close?

atamazov:
@daniellowell I don't think so.

atamazov:
The TODO list at https://github.com/ROCmSoftwarePlatform/MIOpen/issues/560#issuecomment-719842997 is updated. Note that there is one bug that presumably affects auto-tuning, not fixed for long.

DrizztDoUrden:
[...]

PR #307, opened 2020-06-23
GPU'less compile 

alexandraBara:
I am posting this PR preemptively so Tuna people can see my changes. 

The idea is:

1. Set env var to compile only (for desired arch/num_cu) - through Precompile kernels and dump blobs into kern_db
2. Move kern cache to machine where the kernels will execute
3. Run through regular invokers code which will compile (if kernel not already present in kern_db) and execute. Note: the kernel will already be in kern_db from the compile step.

**The following steps need to happen:**
build with MIOPEN_SQLITE_KERN_CACHE=ON

**To compile we need to set the following env vars:**
export MIOPEN_DEBUG_CONV_GEMM=0  (GEMM code path has not been implemented for this)
export MIOPEN_CUSTOM_CACHE_DIR=test (this will be appended to ~/.cache)
export MIOPEN_FIND_ENFORCE=4
export MIOPEN_COMPILE_AND_RUN=0 (compile only, escape generic search early)
export MIOPEN_DEVICE_ARCH=gfx906 (or whatever other supported arch)
export MIOPEN_DEVICE_CU=60 (or any other num_cu)

**To run:**
export MIOPEN_DEBUG_CONV_GEMM=0 (this can also be left out and invokers will actually compile and run as well)
export MIOPEN_CUSTOM_CACHE_DIR=test (must be same location as compile so it can find compiled kernels)
export MIOPEN_FIND_ENFORCE=4
export MIOPEN_COMPILE_AND_RUN=2 (anything but 0, can also be left unset)

I have tested as such:
- compile only(for arch/num_cu different than my own machine) and check kern_db to make sure its populated with compiled kernels.
- moved ukdb to machine with arch/num_cu for which it was compiled for.
- set appropriate env vars to pick up custom cache location and run.
- executed above steps for both hip and opencl
- tested with rocm 3.5

Feel free to propose some actual utest I can add. 

Blocked by: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/203

PR #681, opened 2021-01-14
No GPU backend for Tuna 

JehandadKhan:
The PR adds another backend to MIOpen to support operations that do not require a GPU. This is accomplished by modifying the HIP backend so that the HIP API calls corresponding to those operations are not issued. 

Additionally, this PR enables the access to the handle implementation outside MIOpen so that the client application can override Handle behaviors around device properties.

This backend can be used to enable functionality that only depends on the host side code and primarily intended to support testing and Tuna related operation around collection of data and library introspection purposes.


JehandadKhan:
@atamazov  Can I bother you to take a look at this PR

CMakeLists.txt
```diff
@@ -142,7 +142,7 @@ endif()
 set( MIOPEN_BACKEND ${MIOPEN_DEFAULT_BACKEND} CACHE STRING
     "Which of MIOpens's backends to use?" )
 set_property( CACHE MIOPEN_BACKEND PROPERTY STRINGS
-    OpenCL HIP HIPOC )
+    OpenCL HIP HIPOC NOGPU)
```
pfultz2, line 145: So maybe this should be name `HIPNOGPU` since it still requires hip runtime but no gpu.

src/nogpu/handle.cpp
```diff
@@ -0,0 +1,232 @@
...
+    MIOPEN_HANDLE_LOCK
+    this->Finish();
+    return ddata;
+}
+
+void Handle::ReadTo(void* /* data */,
+                    const Allocator::ManageDataPtr& /* ddata */,
+                    std::size_t /* sz */) const
+{
+    MIOPEN_HANDLE_LOCK
```
pfultz2, line 110: This handle lock is actually for the GPU, so I dont think its needed here.

src/nogpu/handle.cpp
```diff
@@ -0,0 +1,232 @@
...
+    this->Finish();
```
pfultz2, line 111: Since `Finish` does nothing, this function call could be removed.

pfultz2:
We should probably add some tests for this on jenkins so we can make sure someone doesn't accidentally break this in another PR. I guess at first, we could just add jenkins job that just builds this.

CMakeLists.txt
```diff
@@ -192,7 +192,10 @@ add_definitions("-DHIP_COMPILER_FLAGS=${HIP_COMPILER_FLAGS}")
 
 
 # HIP
-if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC")
+if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "NOGPU")
+    if( MIOPEN_BACKEND STREQUAL "NOGPU")
+        set(MIOPEN_BACKEND_NOGPU 1)
```
atamazov, line 197: This naming gives the impression that `MIOPEN_BACKEND_NOGPU` and `MIOPEN_BACKEND_HIP` have equal "rights", but it is not so.

Please rename this to `MIOPEN_BACKEND_MODE_NOGPU` or delete (it is unused anyway).

src/CMakeLists.txt
```diff
@@ -482,7 +482,17 @@ if( MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "HIP")
         )
 endif()
 
-if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "HIP")
+if( MIOPEN_BACKEND STREQUAL "NOGPU")
+    list(APPEND MIOpen_Source
+        hip/hiperrors.cpp
+        nogpu/handle.cpp
```
atamazov, line 488: [Notice] AFAICS in this PR does not use a mechanism designed to add new backends -- I mean the HandleImpl machinery, -- in the originally intended way. So I am curious, why.

However, I do not have time to consider this issue in detail. If it doesn't break anything, and it works for you, then okay. For now.

atamazov (changes requested):
Please merge latest develop here. I am afraid that some of the latest changes (TargetID related) might break functional compatibility of this PR with develop.

This PR looks like a continuation of #307. Two related questions:
- There is #560 which is essentially leftovers of #307. Are we going to resolve it and when?
- It seems like this PR makes some changes done in #307 useless. If so, are we going to revert that changes and when?

src/nogpu/handle.cpp
JehandadKhan: Fixed

src/nogpu/handle.cpp
JehandadKhan: Fixed

CMakeLists.txt
JehandadKhan: Fixed

src/nogpu/handle.cpp
JehandadKhan: Removed

src/nogpu/handle.cpp
JehandadKhan: Removed

JehandadKhan (commented):
> We should probably add some tests for this on jenkins so we can make sure someone doesn't accidentally break this in another PR. I guess at first, we could just add jenkins job that just builds this.

Updated the Jenkinsfile

CMakeLists.txt
```diff
@@ -192,7 +192,10 @@ add_definitions("-DHIP_COMPILER_FLAGS=${HIP_COMPILER_FLAGS}")
 
 
 # HIP
-if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC")
+if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "NOGPU")
```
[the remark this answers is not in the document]
JehandadKhan: Updated the description

CMakeLists.txt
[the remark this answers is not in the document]
JehandadKhan: Fixed

CMakeLists.txt
JehandadKhan: I have renamed the define, however, it is not useless since I use it in the client application to detect if the No GPU mode was enabled or not, which provides for runtime checks. Therefore, I need to keep it.

atamazov: Thanks for the explanation.

src/CMakeLists.txt
JehandadKhan: The PR does introduce a new implementation [here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/38017007a19e34f005458ddeb505f703b65c78c7/src/include/miopen/nogpu/handle_impl.hpp#L1) which has been moved out of the cpp file so that the client program can include the header file and override the behavior when desired.

JehandadKhan:
> Please merge latest develop here. I am afraid that some of the latest changes (TargetID related) might break functional compatibility of this PR with develop.

Updated the branch with latest changes from develop.

> This PR looks like a continuation of #307. 

@atamazov I concur with your observation, there certainly is considerable overlap. In retrospect we should have implemented this PR from the get go and avoided littering the library with numerous env vars and conditionals to check for them. However, recently it became clear to me that we need a more thorough approach to the issue of Tuna ( + fin) trying to override various MIOpen behaviors which is not achievable using env vars. For example an API centric way to override the device name / cu count and in the future, behavior around the program/kernel cache. After discussing with @pfultz2 I decided to implement the first necessary step (this PR) which may be expanded upon later (for binary cache etc). 

> Two related questions:
> 
> * There is #560 which is essentially leftovers of #307. Are we going to resolve it and when?

I am not sure, perhaps its best to ask the author on the same issue ie #560 

> * It seems like this PR makes some changes done in #307 useless. If so, are we going to revert that changes and when?

That is correct, we can revert those changes as soon as the corresponding changes ( using the proposed PR) are implemented in Tuna. In that case the leftovers in #560 would be irrelevant and we can close them.

JehandadKhan:
@atamazov and @pfultz2 please re-review

atamazov:
@JehandadKhan Thanks for feedback.

atamazov (dismissed):
LGTM!

atamazov (approved):

JehandadKhan:
@daniellowell The CI passes

src/kernels/migrate_db.py
```diff
@@ -0,0 +1,52 @@
...
+cur.execute("CREATE UNIQUE INDEX `idx_perf_db` ON perf_db(solver, config);")
+orphan_cnt = cur.execute("SELECT count(*) from config where id not in ( select config from perf_db);").fetchone()[0]
+if orphan_cnt != 0:
+	cur.execute("Delete from config where id not in (select config from perf_db);")
+
+
+new_cnt = cur.execute("SELECT count(*) from perf_db").fetchone()[0]
+assert(new_cnt == total_cnt)
+new_slv_cnt = {}
+for cnt, slv in cur.execute("select count(*), solver from perf_db group by solver;"):
```
alexandraBara, line 45: @JehandadKhan  should there be a cur.close() somewhere here?

src/nogpu/handle.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+std::size_t Handle::GetLocalMemorySize() const { return this->impl->local_mem_size; }
+
+std::size_t Handle::GetGlobalMemorySize() const { return this->impl->global_mem_size; }
+
+std::size_t Handle::GetMaxComputeUnits() const { return this->impl->num_cu; }
+
+std::size_t Handle::GetImage3dMaxWidth() const { return this->impl->img3d_max_width; }
+
+std::size_t Handle::GetWavefrontWidth() const { return this->impl->warp_size; }
```
atamazov, line 177: IIRC the NOGPU backend is used for building kernels. GetWavefrontWidth() should return correct value (in accordance to the target GPU type) in order to Solvers generate proper Solutions. Build options may depend on wavesize. This would lead to cache misses when using a cache filled by kernels built with NOGPU.

PR #687, opened 2021-01-18
Adding applicable filter to consider layout to existing kernels

jerryyin:
All existing kernels are NCHW only. I'm adding this filter just so that when NHWC kernels are added, MIOpen won't get confused and return a kernel with wrong layout support.

### DIFF
--- a/src/solver/conv_MP_bidirectional_winograd.cpp
+++ b/src/solver/conv_MP_bidirectional_winograd.cpp
@@ -313,6 +313,11 @@ bool ConvMPBidirectWinograd<WinoDataH, WinoFilterH, WinoDataW, WinoFilterW>::IsA
     // HIP backend required for sending ptr (buffer + offset)
     // ROCBLAS for GEMM step
 
+    if(!params.IsLayoutDefault())
+    {
+        return false;
+    }
+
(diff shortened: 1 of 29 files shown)


atamazov (approved):
Thanks for fixing this! LGTM.

[Recommendation] Remove functionally useless curly braces.

jerryyin:
Thanks @atamazov! I will leave the braces as-is for now as I used to get some random failures on the tidy/format tests without those. :-p

atamazov:
@jerryyin 
> Thanks @atamazov! I will leave the braces as-is for now as I used to get some random failures on the tidy/format tests without those. :-p

[TIP] Run this script from `src/` after each edit to avoid formatting failures:
```bash
#!/bin/bash
find . -iname '*.h' \
    -o -iname '*.hpp' \
    -o -iname '*.cpp' \
    -o -iname '*.h.in' \
    -o -iname '*.hpp.in' \
    -o -iname '*.cpp.in' \
    -o -iname '*.cl' \
| xargs -n 1 -P $(nproc) -t clang-format-3.8 -style=file -i 2>/dev/null
```
`| grep -v -E '(build/)|(install/)'` is to avoid formatting within `src/build/` and `src/install/` (I use these for building etc.); please fix this line accorting to your needs.

PR #700, opened 2021-01-21
Check  for programs with CMAKE_INSTALL_PREFIX

daniellowell:
Fixes issue: [SWDEV-269156](http://ontrack-internal.amd.com/browse/SWDEV-269156)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,9 @@ if( MIOPEN_BACKEND STREQUAL "OpenCL")
     find_package( OpenCL REQUIRED )
     find_program(MIOPEN_HIP_COMPILER hcc
             PATH_SUFFIXES bin
-            PATHS /opt/rocm
+            PATHS 
+	        /opt/rocm
+		${CMAKE_INSTALL_PREFIX}
     )
     set(MIOPEN_USE_MIOPENGEMM ON CACHE BOOL "")
     if(MIOPEN_HIP_COMPILER)
@@ -157,7 +159,9 @@ if( MIOPEN_BACKEND STREQUAL "OpenCL")
     else()
         find_program(MIOPEN_HIP_COMPILER clang++
             PATH_SUFFIXES bin
-	        PATHS /opt/rocm/llvm
+	        PATHS 
+		    /opt/rocm/llvm
+		    ${CMAKE_INSTALL_PREFIX}/llvm
         )
         if(MIOPEN_HIP_COMPILER)
             message("hip compiler: ${MIOPEN_HIP_COMPILER}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -149,15 +149,19 @@ if( MIOPEN_BACKEND STREQUAL "OpenCL")
     find_package( OpenCL REQUIRED )
     find_program(MIOPEN_HIP_COMPILER hcc
             PATH_SUFFIXES bin
-            PATHS /opt/rocm
+            PATHS 
+	        /opt/rocm
+		${CMAKE_INSTALL_PREFIX}
```
pfultz2, line 154: CMake already searches for in the `CMAKE_INSTALL_PREFIX`, so this isn't needed here.

CMakeLists.txt
```diff
@@ -149,15 +149,19 @@ if( MIOPEN_BACKEND STREQUAL "OpenCL")
...
     set(MIOPEN_USE_MIOPENGEMM ON CACHE BOOL "")
     if(MIOPEN_HIP_COMPILER)
         message("hip compiler: ${MIOPEN_HIP_COMPILER}")
     else()
         find_program(MIOPEN_HIP_COMPILER clang++
             PATH_SUFFIXES bin
-	        PATHS /opt/rocm/llvm
+	        PATHS 
+		    /opt/rocm/llvm
+		    ${CMAKE_INSTALL_PREFIX}/llvm
```
pfultz2, line 164: So the `/opt/rocm/llvm` can be removed as its already added to the `CMAKE_PREFIX_PATH` on line 110. We should probably add `${CMAKE_INSTALL_PREFIX}/llvm` to the CMAKE_PREFIX_PATH as well and remove adding these extra paths everywhere.

CMakeLists.txt
daniellowell: But it fails to.

daniellowell: Scratch that, just for clang-offload-bundler.

atamazov: But SWDEV-269156 is about extractkernels isn't it? And this is 4.0.x, so we are not using extractkernels, so why we are looking for it?

daniellowell: This is about the symlinks not existing in build dockers. Which means that searching for clang-offload-bundler fails, then the assumption is that we are now in a legacy environment, so MIOpen looks for extractkernel, which also then fails.

atamazov: I see. Usually I ignore that and simply `ln -s /opt/rocm-... /opt/rocm` but this is inconvenient when one works with multiple installations.

Issue #26, opened 2017-12-28
Warning message due to file unreadable


tdsk108:
When I ran CaffeNet inference sample after following the installation steps in https://github.com/ROCmSoftwarePlatform/hipCaffe/blob/hip/README.ROCm.md, I observed the following warning message. I obtained the same result with the example on the hipCaffe quick start guide. Could you please let me know if this warning can be ignored?

MIOpen(HIP): Warning [ReadFile] File is unreadable.

It looks the warning is caused by failure in access to /opt/rocm/miopen/share/miopen/db/gfx803_16.cd.pdb.txt which does not exist. There are only following files under /opt/rocm/miopen/share/miopen/db/ in my environment. 
gfx803_64.cd.pdb.txt gfx900_64.cd.pdb.txt

### System configuration
GPU: RX-560
OS: Ubuntu 16.04.3
Package versions: 
 rocm-dkms/rocm-libs 1.7.60
 hip_base 1.4.17494
 hcc 1.1.17493
 miopen-hip 1.2.0-a9949e3
 miopengemm 1.1.5-3c3145b

dagamayank:
@tdsk108 Thanks for the comment. The issue here is that you are running on a GPU for which MIOpen is not optimized for; MIOpen is optimized for 36/64CUs (compute units) for the `gfx803` architecture. However, you are running a 16 CU GPU hence, the warning.

The easiest workaround is that you can copy the `gfx803_64.cd.pdb.txt` file and create a file called `gfx803_16.cd.pdb.txt` in the same location. This will enable decent performance on your GPU.

In the future, we will be tuning MIOpen for a 16 CU GPU as well.

tdsk108:
@dagamayank Thank you. I understand reason of the warning and could remove the warning by the workaround.

Ukhupacha:
Hello there, 
I encountered the same problem but instead of gfx803_**16**.cd.pdx.txt I have  gfx803_**8**.cd.pdx.txt
`MIOpen(HIP): Warning [FindRecordUnsafe] File is unreadable: /opt/rocm/miopen/share/miopen/db/gfx803_8.cd.pdb.txt`

I have a Radeon E9171 MCM  and Fedora 28, and I'm using using the docker rocm/tensorflow. 
Is there a way I can get the gfx803_**8**.cd.pdx.txt file ?

Thanks in advance,

PR #701, opened 2021-01-22
Split the binary db and remove redundant db files

JehandadKhan:


atamazov:
PerformanceConfigs and depend on number of CUs. For example, for some Solvers PerformanceConfigs may become invalid when number of CUs does not match to the number of CUs used during tuning.

Results of Find also depend on number of CUs.

atamazov:
gfx803 support should not be removed without prior discussion & updating of the following ticket: https://github.com/AMDComputeLibraries/MLOpen/issues/1934. Thanks.

JehandadKhan:
> gfx803 support should not be removed without prior discussion & updating of the following ticket: [AMDComputeLibraries/MLOpen#1934](https://github.com/AMDComputeLibraries/MLOpen/issues/1934). Thanks.

@atamazov Please note that support for gfx8 is not being removed. However, we will no longer be tuning for gfx803 based architectures. The ticket referenced above has been updated.

atamazov:

junliume:
Ping @JehandadKhan are we still preparing for the DB split?

junliume:
@JehandadKhan 

[2021-08-06T19:26:39.032Z]  5/81 Test #26: **test_immed_conv3d** ..........................***Failed  Error regular expression found in output. Regex=[FAILED] 40.73 sec

> ... ...
> 
> [2021-08-06T19:26:39.032Z] MIOpen(HIP): Info [GetPerfDbPath] Unable to find exact perf database file
> 
> [2021-08-06T19:26:39.032Z] MIOpen(HIP): Info [GetPerfDbPath] Iterating over perf db directory /data/workspace/MLLibs_MIOpen_split_db/src/kernels
> 
> [2021-08-06T19:26:39.032Z] /data/workspace/MLLibs_MIOpen_split_db/build/bin/test_immed_conv3d --float --cmode conv --pmode default --group-count 1 --enable-fdb 0 --batch_size 8 --input_channels 2 --output_channels 16 --spatial_dim_elements 16 16 16 --filter_dims 5 5 5 --pads_strides_dilations 0 0 0 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW 
> 
> [2021-08-06T19:26:39.032Z] FAILED: filesystem::recursive_directory_iterator increment error: No such file or directory

JehandadKhan:
@junliume  Thanks for the pointer, I am already aware and working on figuring out what is going on.

junliume (approved):
CI has passed, we can merge when ready

DrizztDoUrden:
Sure

src/find_db.cpp
```diff
@@ -45,12 +48,156 @@ boost::optional<std::string>& testing_find_db_path_override()
...
+            }
+
+            const auto db_id        = handle.GetTargetProperties().DbId();
+            const int real_cu_count = handle.GetMaxComputeUnits();
+            int closest_cu          = std::numeric_limits<int>::max();
+            fs::path best_path;
+            for(const auto& entry : all_files)
+            {
+                const auto fname = entry.stem().string();
+                MIOPEN_LOG_I("Checking embedded find db file: " << fname);
```
DrizztDoUrden, line 87: This should probably be I2

src/find_db.cpp
```diff
@@ -45,12 +48,156 @@ boost::optional<std::string>& testing_find_db_path_override()
...
+        {
+            MIOPEN_LOG_I2("Found exact find database file");
+            return file_path.string();
+        }
+        else
+        {
+            MIOPEN_LOG_I2("inexact find database search");
+            if(fs::exists(root_path) && fs::is_directory(root_path))
+            {
+                MIOPEN_LOG_I("Iterating over find db directory " << root_path.string());
```
DrizztDoUrden, line 139: Same as above, this should probably be I2.

DrizztDoUrden (approved):
LGTM overall. Just some log verbosity recommendations.

src/find_db.cpp
JehandadKhan: fixed, here and elsewhere

JehandadKhan:
Ping @junliume @atamazov @DrizztDoUrden

src/find_db.cpp
```diff
@@ -45,12 +48,156 @@ boost::optional<std::string>& testing_find_db_path_override()
...
 std::string FindDbRecord_t<TDb>::GetInstalledPath(Handle& handle)
 {
 #if !MIOPEN_DISABLE_SYSDB
-    return GetSystemDbPath() + "/" + handle.GetDbBasename() + "." + GetSystemFindDbSuffix() +
-           ".fdb.txt";
+#if MIOPEN_EMBED_DB
+    return GetInstalledPathEmbed(handle);
+#else
+    return GetInstalledPathFile(handle);
+#endif
```
atamazov, line 200: `GetInstalledPath()` is invoked whenever FindDbRecord instance is created. The previous version of `GetInstalledPath()` is very fast and takes ~constant time. The new version adds `boost::filesystem::exists()` at best or loop with file lookup at worst.

The simplest workaround is caching return value in a static variable.

atamazov, line 200: `GetInstalledPath()` is invoked whenever FindDbRecord instance is created. The previous version of `GetInstalledPath()` is very fast and takes ~constant time. The new version adds `boost::filesystem::exists()` at best or loop with file lookup at worst. This would lead to performance drop.

The simplest workaround is caching return value in a static variable.

src/include/miopen/execution_context.hpp
```diff
@@ -87,20 +89,158 @@ struct ExecutionContext
...
+        return result;
+    }
+#endif
+
+    std::string GetPerfDbPath() const
+    {
+#if MIOPEN_EMBED_DB
+        return GetPerfDbPathEmbed();
+#else
+        return GetPerfDbPathFile();
```
atamazov, line 242: Ditto. This function must be fast.

atamazov (commented):
The changes in the core of library must be thoroughly reviewed and tested for performance.

src/find_db.cpp
```diff
@@ -45,12 +48,156 @@ boost::optional<std::string>& testing_find_db_path_override()
     return data;
 }
 
+#if MIOPEN_EMBED_DB
+template <class TDb>
+std::string FindDbRecord_t<TDb>::GetInstalledPathEmbed(Handle& handle)
+{
+    static const auto embed_path = [&] {
```
JehandadKhan, line 55: @atamazov Please review here.

src/find_db.cpp
```diff
@@ -45,12 +48,156 @@ boost::optional<std::string>& testing_find_db_path_override()
...
+    }();
+    return embed_path;
+}
+
+#else
+
+template <class TDb>
+std::string FindDbRecord_t<TDb>::GetInstalledPathFile(Handle& handle)
+{
+    static const auto installed_path = [&] {
```
JehandadKhan, line 122: @atamazov  please review here.

src/find_db.cpp
JehandadKhan: @atamazov Both `GetInstalledPathEmbed` and `GetInstalledPathFile` functions return a static const variable and therefore the above logic should execute only once. Adding another static variable here would be redundant.

src/find_db.cpp
atamazov: Good!

atamazov:
@JehandadKhan The last question - have you seen https://github.com/AMDComputeLibraries/MLOpen/issues/1674? The approach proposed there differs from this PR:
- (1) It does not iterate over directories because CMake can provide us with the list of supported databases
- (2) The criteria of "closest CUs" is different

What do you think? Shall we try to adopt (2) at least? Or even both? This could be a good exercise for Alexey ;)

src/find_db.cpp
atamazov: Yep, I missed those lambdas ;/

JehandadKhan:
> 
> 
> @JehandadKhan The last question - have you seen [AMDComputeLibraries/MLOpen#1674](https://github.com/AMDComputeLibraries/MLOpen/issues/1674)? The approach proposed there differs from this PR:
> 
>     * (1) It does not iterate over directories because CMake can provide us with the list of supported databases
> 
>     * (2) The criteria of "closest CUs" is different
> 
> 
> What do you think? Shall we try to adopt (2) at least? Or even both? This could be a good exercise for Alexey ;)

Agreed on both accounts. I like your approach better than the one I implemented in this PR.

Issue #703, opened 2021-01-25
[OCL BE] #674 breaks miopenCreateWithStream()


atamazov:
#674 breaks `miopenCreateWithStream()` for OCL BE.

/cc @Kirpich30000

atamazov:
via #705

PR #705, opened 2021-01-26
[HOTFIX] Fix miopenCreateWithStream for OCL BE (issue 703)

atamazov:
- Resolves issue #703.
- Updates handle test to check if implementation of `miopenCreateWithStream()` works.


Kirpich30000 (approved):
LGTM

PR #707, opened 2021-01-26
Device name changes

muralinr:
Device name file changes are updated.


atamazov:
@muralinr Why this is needed? Optional: Why old code worked, what do you think?

daniellowell (changes requested):
This should use gcnArchName

PR #473, opened 2020-09-30
Fwd gemm invokers

DrizztDoUrden:
- Find-db is updated with new solver IDs right here.
- By-product: Fixes #519 by making gemm not applicable to the case.

Forward gemm is now split between the following solvers:
- GemmFwd1x1_0_2 - applicable when 2d && fp32 && kernel is 1x1 && conv pads are 0x0 && conv strides are 2x2
- GemmFwd1x1_0_1_int8 - applicable when weights are int8 && kernel is 1x1 && conv pads are 0x0 && conv strides are 1x1 && groups_count is 1
- GemmFwd1x1_0_1 - applicable when weights are <b><i>not</i></b> int8 && kernel is 1x1 && conv pads are 0x0 && conv strides are 1x1
- GemmFwdRest - applicable when none of the solvers above is.

The fix of #519 is the part where GemmFwd1x1_0_1_int8 is not applicable with `group_count != 1` and GemmFwd1x1_0_1 is not applicable with int8.

"The rest" kind of the solver logic is flawed, but at the moment, it is implemented as it worked before. Changes to that may be applied in separate PR later.

PR #709, opened 2021-01-27
[CI] Improved error logs. Eliminated some false failures.

atamazov:
- Fixed: "failed" at Info logging level leads to false CI failures (as tests assume "failed" is indication of a failure).
- Fixed: The default log level was incorrectly used instead of level 5 at stages "Hip Debug COMGR" and "Hip Debug Embedded Vega20".
- Improved readability of logs. Whenever some detailed log level are used (5 or more), `--verbose` added to the tests.

All these are important for triaging/debugging complicated PRs, like @DrizztDoUrden's #473, so high priority & value.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -314,13 +314,13 @@ pipeline {
                             rm -rf build
                             mkdir build
                             cd build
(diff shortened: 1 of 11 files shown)


DrizztDoUrden (approved):

atamazov:
This PR unblocks #473

Issue #692, opened 2021-01-20
[RNN][OCL BE] test_lstm fails due to OCL compiler warning


atamazov:
- The test fails with compiler mainline builds 6326 and 6416 at least.
- Most likely, the reason is change in the compiler - a new warning was added recently or some bug (false warning) introduced.
- Who is affected
  - The problem should NOT affect the end users (warnings do not lead to errors in non-developer builds)
  - QA should not be affected because they use non-developer builds as well.
  - All developers who work with compiler mainline.
  - Our CI, but not now -- after we upgrade CI to ROCm 4.1.
- What looks suspicious to me is that compiler clams that `n_work_groups_with_1_more == 0` which is indeed strange and my be a symptom of incorrectness in miopengemm (affects only OCL backend).

Due to all the above, let's assign the high priority.

Error log:
```
root@morpheus:/atamazov-github/MLOpen1/build/6416.release.opencl# ./bin/test_lstm
Empty batch sequence. Filling uniformly with batch size: 17
MIOpen(OpenCL): Info [Handle] Device name: gfx906:sramecc-:xnack-
MIOpen(OpenCL): Info [Handle] stream: 0x55aadb1d85e0, device_id: 0x55aadb1ab970
MIOpen(OpenCL): Error [BuildProgram] Build log: /tmp/comgr-e9c850/input/CompileSource:182:62: error: result of comparison of unsigned expression < 0 is always false [-Werror,-Wtautological-unsigned-zero-compare]
  int n_unrolls_remaining = (KV__ / G_UNROLL) +  (group_id_z < n_work_groups_with_1_more);
                                                  ~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
Error: Failed to compile opencl source (from CL or HIP source to LLVM IR).
```

How to reproduce (Radeon VII):
- Pull code from PR #690 
- Pull docker w/mainline build 6326 or 6416
- Add `-DMIOPEN_OVERRIDE_HIP_VERSION_MINOR=1` to cmake command line when building MIOpen
- Build for OpenCL backend
- Run `test_lstm`

/cc @daniellowell

ce1adon:
@atamazov Which architecture?

atamazov:
> How to reproduce (Radeon VII):...

This is gfx906.

ce1adon:
It's not replicable locally with 6416.

ce1adon:
[...]

PR #715, opened 2021-01-29
[Workaround] Skip opencl lstm test for Issue #692

ce1adon:
Workaround for issue #692

### DIFF
--- a/test/lstm_common.hpp
+++ b/test/lstm_common.hpp
@@ -51,6 +51,8 @@
 #define MIO_LSTM_TEST_DEBUG 0
 #define MIO_RNN_TIME_EVERYTHING 0
 
+#define WORKAROUND_ISSUE_692 1
+
 //****************************************************
 // FORWARD BASE
 //****************************************************
@@ -2928,9 +2930,10 @@ struct lstm_basic_driver : test_driver
     {
 
 #if(MIOPEN_BACKEND_OPENCL == 1)
+#if WORKAROUND_ISSUE_692 == 1
         std::cout << "Skip test for Issue #692: " << std::endl;
-        return;
-
+        exit(EXIT_SUCCESS);
+#endif
         if(type == miopenHalf)
             exit(EXIT_SUCCESS);
 #endif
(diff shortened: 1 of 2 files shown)


test/lstm_common.hpp
```diff
@@ -2928,6 +2928,9 @@ struct lstm_basic_driver : test_driver
     {
 
 #if(MIOPEN_BACKEND_OPENCL == 1)
+        std::cout << "Skip test for Issue #692: " << std::endl;
```
atamazov, line 2931: I think we shall 
- formalize workaround
- use exit(...)
- avoid "unreachable code" warnings

atamazov (changes requested):

atamazov:
@ce1adon Please let me know if you do not have time for resolving https://github.com/ROCmSoftwarePlatform/MIOpen/pull/715#pullrequestreview-579784264, and I will do that.

Also I do not think it is worth to re-start CI testing until the review comments have been resolved. ;)

test/lstm_common.hpp
ce1adon: 1. What is current formal workaround format?
2.  will do
3.  where does the "unreachable code" warnings happen?

test/lstm_common.hpp
```diff
@@ -2928,6 +2928,9 @@ struct lstm_basic_driver : test_driver
...
+        return;
+
         if(type == miopenHalf)
             exit(EXIT_SUCCESS);
```
atamazov, line 2935: This is unreachable code (perhaps CI will not complain)

test/lstm_common.hpp
atamazov: Thanks!

test/lstm_common.hpp
```diff
@@ -2928,6 +2930,10 @@ struct lstm_basic_driver : test_driver
     {
 
 #if(MIOPEN_BACKEND_OPENCL == 1)
+#if WORKAROUND_ISSUE_692 == 1
```
atamazov, line 2933: [Note] FYI `#if WORKAROUND_ISSUE_692` is just enough. No need to fix this in this PR.

atamazov (approved):

PR #727, opened 2021-02-04
Implementation of four additional generic tensor reduction operations (AVG, AMAX, NORM1, NORM2)

fronteer:
Four generic tensor reduction operations are added (AVG, AMAX, NORM1, NORM2).  So far, all 8 reduction operations specified by the cudnnReduceTensor() interface are implemented for MIOpen (ADD, MUL, MIN, MAX, AVG, AMAX, NORM1, NORM2). 

To test: 
>#for fp32
> `#bin/test_reduce_test --all`                              
>
>#for fp16
> `#bin/test_reduce_test --all --half`


include/miopen/miopen.h
```diff
@@ -470,14 +470,14 @@ typedef enum {
...
         3, /*!< the operation is getting the maximum value of the reduced elements */
-    // MIOPEN_REDUCE_TENSOR_AMAX =
-    //    4, /*!< the operation is getting the maximum absolute value of the reduced elements */
-    // MIOPEN_REDUCE_TENSOR_AVG =
-    //    5, /*!< the operation is getting the averaged value of the reduced elements */
-    // MIOPEN_REDUCE_TENSOR_NORM1 =
-    //    6, /*!< the operation is adding the absolute values of the reduced elements */
-    // MIOPEN_REDUCE_TENSOR_NORM2 = 7, /*!< the operation is getting the square root of the sum of
-    //                                   squares of the reduced elements */
+    MIOPEN_REDUCE_TENSOR_AMAX =
```
atamazov, line 473: Support in the MIOpenDriver is also necessary. Are you going to implement it in next PR?

atamazov, line 473: Add this before enum:
```C++
/*! @ingroup TensorReduce
* @brief Level of support provided by TensorReduce operations.
* Applications may use it to enshure backward compatibility with older
* library versions.
*
* 0 or undefined - ADD, MIN, MIN, MAX.
* 1 - Added AVG, NORM1, NORM2.
*/
#define MIOPEN_API_VERSION_REDUCE_LEVEL 1
```

atamazov (changes requested):
Partial.

include/miopen/miopen.h
fronteer: What do you mean?   I think it is already supported.   Both the driver/reduce_driver.hpp(which include smiopen_ReductionHost.hpp) and test/reduce_test.cpp share the test/cpu_reduce_util.hpp.

driver/miopen_ReductionHost.hpp
```diff
@@ -205,6 +205,8 @@ class miopenReductionHost
     RunImpl_with_indices(Tgpu alpha, const Tgpu* in_data, Tgpu beta, Tref* out_data, int* indices)
     {
         using reduce::ReduceOpFn2;
+        using reduce::PreUnaryOpFn;
```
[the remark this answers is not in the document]
atamazov: Let's rename it eventually, thanks for the comment.

include/miopen/miopen.h
atamazov: I've overlooked that.

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,194 @@
...
+{
+    (void)x;
+    static_assert(static_cast<T>(0), "float_equal_one() is not implemented for this data type");
+    return false;
+};
+
+template <>
+inline bool float_equal_one<float>(float x)
+{
+    return std::isfinite(x) and x <= 1.0f and x >= 1.0f;
```
atamazov, line 77: [Q] Well, `(x == 1.0f)` should work properly and provides better performance. Why we need this?

atamazov:
I do not see anything that prevents merging, except that https://github.com/ROCmSoftwarePlatform/MIOpen/pull/727#discussion_r576761340 is not resolved yet.

JehandadKhan:
@fronteer  Please resolve the issues as well as ensure that the CI passes so that we can merge this in.

test/cpu_reduce_util.hpp
fronteer: I don't how  x == const y  is implemented in instructions.  I used this <= and >= combination since it came across compiler warning previously (either for GPU or CPU kernel).

atamazov: [Recommendation] Add a comment explaining why the unusual code is needed to make it easier for readers to understand.

fronteer: The reason that I used the <= and >= combination is that this method removes the warning and should have better performance than using the std::nextafer by float_equal_fn defined in  src/include/miopen/float_equal.hpp

atamazov: @fronteer Is performance important?

[Hint] 
- You can use `x == 0.0` and `x==1.0` that should compile to a single instruction. I am not sure only about `half` on CPU.
- In case of insurmountable problems, bitwise comparison can be used, e.g. `(*reinterpret_cast<uint32_t*>(&x) == 0x3f800000)` for `float` etc.

fronteer: x == const y should be accurate if without compiler warning since here x is not used for calculated values, rather it is used for the constants to scaling the in/out tensor

fronteer: Thank you.   Changed.   Using == to compare floats on host side should not be an issue as x86 does have single instruction to do that and I saw half_float::half has "==" overloaded

fronteer: Just saw that  DeepCode report issue for using "==" with float/double

atamazov: This DeepCode warning is junk. It assumes that the programmer does not understand what he is doing.

Use this Ignore Suggestion from Deepcode if you care:

> Copy the following line:
> ```
> //  file deepcode ignore FloatingPointEquals: <comment the reason here>
> ```
> And paste it in cpu_reduce_util.hpp to prevent this suggestion from appearing again in this file. 

BTW the DeepCode's explanation is incorrect for zero, NaNz and Infinities:
> Floating point numbers are only considered equal if their bit representation is equal.

test/cpu_reduce_util.hpp
```diff
@@ -74,20 +74,19 @@ static inline bool float_equal_one(T x)
 template <>
 inline bool float_equal_one<float>(float x)
 {
-    return std::isfinite(x) and x <= 1.0f and x >= 1.0f;
+    return std::isfinite(x) and x == 1.0f;
```
atamazov, line 77: BTW isfinite is not necessary as well, esp. if you care about performance. Up to you.

atamazov (dismissed):

test/cpu_reduce_util.hpp
fronteer: I regressed to the previous codes and added some comments

test/cpu_reduce_util.hpp
```diff
@@ -71,22 +71,25 @@ static inline bool float_equal_one(T x)
     return false;
 };
 
+// use combination of "<=" and ">=" for float values equality checking to remove DeepCode warning
+// caused by direct using of "=="
```
atamazov, line 75: In this specific case, DeepCode warning is a misleading garbage that doesn't worth commenting (except disabling that check).

We need to reach consensus on this topic, so let's discuss.

atamazov (changes requested):

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,196 @@
...
 #include <miopen/bfloat16.hpp>
 
 namespace reduce {
 
+template <typename T>
+static inline bool IsNan(T x)
+{
+    // C++ isnan() is used for float and double
+    return (std::isnan(x));
+};
```
pfultz2, line 45: Cant we just do `using std::isnan;` and then call `isnan` unqualified? Then the half overload should be found by ADL lookup

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,196 @@
...
+    return (half_float::isfinite(x));
+};
+
+template <typename T>
+static inline bool float_equal_one(T x)
+{
+    (void)x;
+    static_assert(static_cast<T>(0), "float_equal_one() is not implemented for this data type");
+    return false;
+};
```
pfultz2, line 72: We already have a `float_equal` function [here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/float_equal.hpp#L44). You should use that and add an overload for `half`.

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,196 @@
...
+
+    case MIOPEN_REDUCE_TENSOR_ADD:
+    case MIOPEN_REDUCE_TENSOR_NORM1:
+    case MIOPEN_REDUCE_TENSOR_MUL:
+    case MIOPEN_REDUCE_TENSOR_MIN:
+    case MIOPEN_REDUCE_TENSOR_MAX:
+    case MIOPEN_REDUCE_TENSOR_AMAX: return ([&](compType& a_) { (void)a_; });
+    }
+
+    return (std::function<void(compType&)>{});
```
pfultz2, line 210: We should probably throw an error here.

test/cpu_reduce_util.hpp
fronteer: Do you mean std::isnan() can also be directed to call half's overloaded isnan() implementation ?    Can you point to me some document about which C++ interfaces can be used for non-builtin object like half_float::half, but which implements overloaded common interfaces like operator == and isnan(), isfinite()

test/cpu_reduce_util.hpp
fronteer: float_equal is more expensive.  float_equal_one/float_equal_zero are used to check the alpha/beta values for scaling the tensor. The alpha and beta are passed as float constants, which need not be checked by std::isfinite()

test/cpu_reduce_util.hpp
fronteer: Also for static_cast<>， can it be called directly to do transformation between float/double/in and half_float::half ?

test/cpu_reduce_util.hpp
fronteer: Yes, Changed

test/cpu_reduce_util.hpp
fronteer: Just go to the simplest codes and I have verified it works correctly on host side: 
1)  operator ==  works for both float/double. It is able to regard two zero values as equal when one of them is 0x00000000, the other is 0x80000000 
2) half_float::half also implemented the correct overload for operator ==

test/cpu_reduce_util.hpp
fronteer: Yes.  std::isnan() works for half_float::half.  But I kept the definition of IsNan() there in case other types are involved.

test/cpu_reduce_util.hpp
atamazov: @fronteer Let me ask again: is performance important here?

test/cpu_reduce_util.hpp
atamazov: > But I kept the definition of IsNan() there in case other types are involved.

This is overdesign.

> std::isnan() works for half_float::half.

`std::isnan()` can't find `isnan` for half, because the latter is defined in the `half_float` namespace.

> Can you point to me some document about which C++ interfaces can be used for non-builtin object like half_float::half,...

Please refer to https://en.cppreference.com/w/cpp/language/adl which is applicable to the design proposed by Paul.

Or you can import `isnan` for half into `std` like this
```C++
namespace std {
using half_float::isnan;
}
```
But I am not sure if this is good practice.

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,177 @@
...
+    return (std::isnan(x));
+};
+
+template <typename T>
+static inline bool float_equal_one(T x)
+{
+    (void)x;
+    static_assert(static_cast<T>(0), "float_equal_one() is not implemented for this data type");
+    return false;
+};
```
atamazov, line 55:
```suggestion
template <typename T>
static inline bool float_equal_one(T);
```
No definition given for a static function, so build will fail at compilation step.

test/cpu_reduce_util.hpp
pfultz2: You shouldn't call `std::isnan`. Instead just call `isnan` unqualified(without `std::`). This will lookup the function using ADL, which means the function will first check in the data types' associated namespace, otherwise it will fallback on the current namespace.

This means that calling `isnan` using `half` will lookup the function in the `half_float` namespace. However, for builtin types there is no associated namespace so you need to add `using std::isnan` to your current namespace so it will lookup `std::isnan` for builtin types. This is also what is recommended from [half's documentation](http://half.sourceforge.net/index.html#usage):

> Additionally the half_float namespace also defines half-precision versions for all mathematical functions of the C++ standard library, which can be used directly through ADL:

You should use the standard floating-point functions, which makes everything consistent. If we add a data type in the future then that data type should provide overloads for these functions that can be found with ADL as well.

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,177 @@
...
+static inline bool IsNan(T x)
+{
+    // C++ isnan() is used for float, double and half_float::half
+    return (std::isnan(x));
+};
+
+template <typename T>
+static inline bool float_equal_one(T x)
+{
+    (void)x;
```
pfultz2, line 52: Just remove the parameter name instead of using `(void)x`.

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,177 @@
...
+    static_assert(static_cast<T>(0), "float_equal_one() is not implemented for this data type");
+    return false;
+};
+
+template <>
```
pfultz2, line 57: Remove this line. Functions are found using overload resolution and not template specialization.

test/cpu_reduce_util.hpp
pfultz2: It seems like performance is not important as it calls through std::function which is pretty slow for numeric computations.

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,177 @@
...
+inline half_float::half Abs<half_float::half>(half_float::half a)
+{
+    return half_float::abs(a);
+};
+
+template <typename compType>
+static inline std::function<void(compType&)> PreUnaryOpFn(miopenReduceTensorOp_t op_, int divider)
```
pfultz2, line 144: If performance matters then this should use a visitor pattern instead of returning `std::function`:

```cpp
template <class compType, class F>
static inline void PreUnaryOpFn(miopenReduceTensorOp_t op_, int divider, F f)
{
    switch(op_)
    {
    case MIOPEN_REDUCE_TENSOR_NORM1:
        f([&, divider](compType& a_) {
            a_ = Abs<compType>(a_) / convert_type<compType>(static_cast<float>(divider));
        });
    case MIOPEN_REDUCE_TENSOR_NORM2:
        f([&, divider](compType& a_) {
            a_ = a_ * a_ / convert_type<compType>(static_cast<float>(divider));
        });
    case MIOPEN_REDUCE_TENSOR_AMAX: f([&](compType& a_) { a_ = Abs<compType>(a_); });

    case MIOPEN_REDUCE_TENSOR_AVG:
    case MIOPEN_REDUCE_TENSOR_ADD:
    case MIOPEN_REDUCE_TENSOR_MUL:
    case MIOPEN_REDUCE_TENSOR_MIN:
    case MIOPEN_REDUCE_TENSOR_MAX: f([&](compType& a_) { (void)a_; });
    }

    throw std::runtime_error(std::string(__FUNCTION__) +
                             ": using undefined Reduction operation is not permitted");
};
```

Then you would call it like this:

```cpp
PreUnaryOpFn<compType>(reduceOp, divider, [&](auto PreUnaryOp) {
    
});
```

But you probably need to combine both `PreUnaryOpFn` and `PosUnaryOpFn` so it can reuse the same switch statement.

test/cpu_reduce_util.hpp
fronteer: Yes. Done

test/cpu_reduce_util.hpp
fronteer: Yes.  Good

test/cpu_reduce_util.hpp
fronteer: Yes, Removed the default function template definition.

test/cpu_reduce_util.hpp
fronteer: I could not very much understand your point here.

test/cpu_reduce_util.hpp
fronteer: Understand.  Changed

test/cpu_reduce_util.hpp
```diff
@@ -307,6 +245,8 @@ inline half_float::half ReduceOpZeroVal<half_float::half>(miopenReduceTensorOp_t
                              ": using undefined Reduction operation is not permitted");
 };
 
+using std::isnan;
```
atamazov, line 248: Move this into `binop_with_nan_check()`

atamazov (changes requested):
Do not inject isnan from std into files that use the header.

test/cpu_reduce_util.hpp
atamazov: @fronteer I highly recommend spending some time for learning this topic, at least read https://en.wikipedia.org/wiki/Visitor_pattern

atamazov:
Almost good to go, I guess. @pfultz2 Thanks for reviewing this.

atamazov:
Clang-tidy errors, please fix.
image

test/cpu_reduce_util.hpp
fronteer: Done

fronteer:
> Clang-tidy errors, please fix.
> image

Fixed.

test/cpu_reduce_util.hpp
```diff
@@ -29,19 +29,117 @@
...
+static inline half_float::half Sqrt(half_float::half a) { return half_float::sqrt(a); };
+
+template <typename T>
+static inline T Abs(T a);
+
+static inline float Abs(float a) { return fabsf(a); };
+
+static inline double Abs(double a) { return fabs(a); };
+
+static inline half_float::half Abs(half_float::half a) { return half_float::abs(a); };
```
pfultz2, line 80: `sqrt` and `abs` can be found through ADL as well, so you should do `using std::abs` and `using std::sqrt`.

fronteer: Yes.  I had ever though about that, but it takes me some time to confirm the availability of abs/sqrt for float/double in libstdc++-dev.     Corrected!  Thanks

fronteer:
> std::accumulate(toReduceLengths.begin(), toReduceLengths.end(), std::size_t{1}, std::multiplies<>{})

Yes,  Changed!

atamazov (approved):
LGTM

atamazov:
@pfultz2 Thanks again for reviewing this. If you see any leftovers, please let us know and we'll fix those in separate PR.

atamazov:
@fronteer Please re-test locally as thorough as possible to make sure that recent review changes didn't break anything. I will try to push this into mainline ASAP.

fronteer:
> @fronteer Please re-test locally as thorough as possible to make sure that recent review changes didn't break anything. I will try to push this into mainline ASAP.

Yes, I do.

atamazov:
Most likely this will fail during our CI testing or during QA Staging tests.

ROCm 4.0, `-DBUILD_DEV=On`:
```C++
$ make test_reduce_test && ./bin/test_reduce_test
[  0%] Built target sqlite_memvfs
[  3%] Built target addkernels
[100%] Built target MIOpen
[100%] Built target test_reduce_test
MIOpen(HIP): Info [get_device_name] Raw device name: gfx906
MIOpen(HIP): Info [Handle] stream: 0xb303e0, device_id: 0
MIOpen(HIP): Info [HipCompilerVersionImpl] 4.0.20496
In file included from gridwise_generic_reduction.cpp:6:
In file included from ./gridwise_generic_reduction.hpp:31:
./reduction_operator.hpp:155:44: error: use of old-style cast [-Werror,-Wold-style-cast]
    static constexpr float scaler = 1.0f / (float)divider;
                                           ^      ~~~~~~~
./reduction_operator.hpp:169:44: error: use of old-style cast [-Werror,-Wold-style-cast]
    static constexpr float scaler = 1.0f / (float)divider;
                                           ^      ~~~~~~~
./reduction_operator.hpp:188:44: error: use of old-style cast [-Werror,-Wold-style-cast]
    static constexpr float scaler = 1.0f / (float)divider;
                                           ^      ~~~~~~~
./reduction_operator.hpp:207:44: error: use of old-style cast [-Werror,-Wold-style-cast]
    static constexpr float scaler = 1.0f / (float)divider;
                                           ^      ~~~~~~~
./reduction_operator.hpp:220:38: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
    __device__ inline constexpr void operator()(half_t& a) const
                                     ^
./reduction_operator.hpp:222:33: note: non-constexpr function '__habs' cannot be used in a constant expression
        a = static_cast<half_t>(__habs(a));
                                ^
/opt/rocm-4.0.0/hip/../include/hip/hcc_detail/hip_fp16.h:1273:13: note: declared here
            __half __habs(__half x)
                   ^
In file included from gridwise_generic_reduction.cpp:6:
In file included from ./gridwise_generic_reduction.hpp:31:
./reduction_operator.hpp:235:38: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
    __device__ inline constexpr void operator()(half_t& a) const
                                     ^
./reduction_operator.hpp:237:33: note: non-constexpr function 'hsqrt' cannot be used in a constant expression
        a = static_cast<half_t>(hsqrt(a));
                                ^
/opt/rocm-4.0.0/hip/../include/hip/hcc_detail/hip_fp16.h:1528:20: note: declared here
            __half hsqrt(__half x)
                   ^
6 errors generated when compiling for gfx906.
```

fronteer:
Artem, thank you for adding the two fixes and for telling me to use "-DBUILD_DEV=on",   so that I will be able to have all issues solved through local building/testing.

atamazov:
ROCm 4.0, gfx906: test_reduce_test fails with NORM1 and NORM2 operation types. I do not know how this was tested. I am going to disable these two op types or disable testing of them at least.

Logs of testing of commits 293bf7482 (tip of this branch) and 178a35d44 (just prior massive review related changes) are identical. Attached below.
- [test_reduce_test.release.hip-clang.178a35d44.console.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6050284/test_reduce_test.release.hip-clang.178a35d44.console.txt)
- [test_reduce_test.release.hip-clang.178a35d44.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6050287/test_reduce_test.release.hip-clang.178a35d44.txt)
- [test_reduce_test.release.hip-clang.293bf7482.console.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6050288/test_reduce_test.release.hip-clang.293bf7482.console.txt)
- [test_reduce_test.release.hip-clang.293bf7482.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6050289/test_reduce_test.release.hip-clang.293bf7482.txt)

fronteer:
Let me look.   The previous version of Reduction is well tested on gfx900 and gfx906.  But the adding is only tested on gxf900 and gfx908.

fronteer:
> ROCm 4.0, gfx906: test_reduce_test fails with NORM1 and NORM2 operation types. I do not know how this was tested. I am going to disable these two op types or disable testing of them at least.
> 
> Logs of testing of commits [293bf74](https://github.com/ROCmSoftwarePlatform/MIOpen/commit/293bf7482beaefb6e4ae7fdf9414a500e89e1e04) (tip of this branch) and [178a35d](https://github.com/ROCmSoftwarePlatform/MIOpen/commit/178a35d44e262b3be3433918f727c91790203369) (just prior massive review related changes) are identical. Attached below.
> 
>     * [test_reduce_test.release.hip-clang.178a35d44.console.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6050284/test_reduce_test.release.hip-clang.178a35d44.console.txt)
> 
>     * [test_reduce_test.release.hip-clang.178a35d44.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6050287/test_reduce_test.release.hip-clang.178a35d44.txt)
> 
>     * [test_reduce_test.release.hip-clang.293bf7482.console.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6050288/test_reduce_test.release.hip-clang.293bf7482.console.txt)
> 
>     * [test_reduce_test.release.hip-clang.293bf7482.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6050289/test_reduce_test.release.hip-clang.293bf7482.txt)

So the issue is from your manual testing, right?  
#> test_reduce_test --all          does have some issue with some cases using NORM1 and NORM2,  because NORM1 and NORM2 always have positive values added, so the final GPU result could have some error bigger than the threshold when comparing with the cpu result. 
The accumulative error is more obvious when the reduced size is bigger.  So that is why  #> test_reduce_test   does not have failure but #> test_reduce_test --all have. 
I will check whether the threshold can be adjusted according to the testing

fronteer:
Using the following patch to expand the RMS tolerance is able to remove the failure.  It seems the other modules in MIOpen also use similar method, which gives the tolerance value by multiplying the output size on 80. 
```C++
         auto outputTensor = tensor<T>{outLengths};
 
+        std::size_t outputElementSize = std::accumulate(outLengths.begin(), outLengths.end(), std::size_t{1}, std::multiplies<>{});
+
+        this->tolerance = 80 * outputElementSize;  
+
         std::fill(outputTensor.begin(), outputTensor.end(), convert_type<T>(0.0f));
```

atamazov:
> So the issue is from your manual testing, right?

I had to test manually because our CI is out of order. But I believe NORM1/2 will fail there.

atamazov:
@fronteer 
> Using the following patch to expand the RMS tolerance is able to remove the failure. It seems the other modules in MIOpen also use similar method, which give the tolerance by multiplying the output size on the default tolerance 80.

I am not quite familiar with topics related to precision for reduction ops. Let's confine ourselves to AVG and AMAX for now, -- it is enough for 4.2. NORM1/2 will be needed in 4.3, so you have time to figure out all the details and make a high quality implementation.

atamazov (approved):

test/reduce_test.cpp
```diff
@@ -747,32 +747,18 @@ struct reduce_driver : test_driver
...
         add(toReduceDims, "R", generate_data(get_toreduce_dims()));
-        add(reduceOp, "ReduceOp", generate_data({
-                MIOPEN_REDUCE_TENSOR_ADD, MIOPEN_REDUCE_TENSOR_MIN, MIOPEN_REDUCE_TENSOR_AMAX,
-                    MIOPEN_REDUCE_TENSOR_AVG
-#if MIOPEN_API_VERSION_REDUCE_TENSOR > 1
-                    ,
-                    MIOPEN_REDUCE_TENSOR_NORM1, MIOPEN_REDUCE_TENSOR_NORM2
-#endif
-            }));
+        add(reduceOp, "ReduceOp", generate_data({0, 4, 5, 6, 7}));
```
atamazov, line 759: Why symbolic names are not used?

test/reduce_test.cpp
```diff
@@ -843,6 +825,14 @@ struct reduce_driver : test_driver
                     tensor_elem_gen_checkboard_sign{}(is...));
         };
 
+        if(reduceOp == MIOPEN_REDUCE_TENSOR_NORM1 || reduceOp == MIOPEN_REDUCE_TENSOR_NORM2)
+        {
+            if(toReduceDims.size() == 4)
+                this->tolerance = 80 * 500;
```
atamazov, line 831: [Q] Are you pretty sure that implementation is correct and increasing of tolerance is justified?

test/reduce_test.cpp
atamazov, line 759: [Q] Perhaps dumb question, but -- why operations 1, 2 and 3 are not tested?

atamazov (changes requested):
Why the following https://github.com/ROCmSoftwarePlatform/MIOpen/pull/727#issuecomment-787116274 was ignored?

I tested commit bcf71f5 manually as described at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/727#issuecomment-786689932. Have you fully tested updated PR on gfx906 and MI100?

atamazov (approved):
This should suit our immediate needs. LGTM!

test/reduce_test.cpp
fronteer: - Take too long to run `#> test_reduce_test --all`
- For MUL, the values could easily reduce to "zero"  and MIOpen will report failure when both outputs (host & gpu) are all zeroes. 
- MIN & MAX are actually covered by AMAX

test/reduce_test.cpp
fronteer: Yes.  `std::numeric_limits<float>::epsilon()  * (80*500)`  is 0.00476837158203, not abnormally large

fronteer:
> Why the following [#727 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/727#issuecomment-787116274) was ignored?
> 
> I tested commit [bcf71f5](https://github.com/ROCmSoftwarePlatform/MIOpen/commit/bcf71f5583ad1ba017b2194a68abcc0ac6821ade) manually as described at [#727 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/727#issuecomment-786689932). Have you fully tested updated PR on gfx906 and MI100?

It is well tested on gfx906 and gfx900, but not MI100.   I will try to get a MI100 box for testing today.

test/reduce_test.cpp
fronteer: Also, in some case, when doing all-reduction of ADD to reduce a big tensor to one value, it is also easy to get a reduced result of one zero value.  This case is avoided by me.  Later  I will try to improve the mean of the initialized data to be a little above 0.0f so that this could not occur

test/reduce_test.cpp
atamazov: Ok

test/reduce_test.cpp
atamazov: > Take too long to run

This is not a problem. Some convolution tests take _hours_. 10-15 minutes is absolutely ok. Now it takes < 5 minutes.

atamazov: If it is possible to test all ops, then this should be done. Of course, this is a bit out of formal scope of this PR (except operation 2 which was tested but now not). What do you think?

atamazov:
> It is well tested on gfx906 and gfx900, but not MI100. I will try to get a MI100 box for testing today.

Save your time. It will be tested in our CI. Thanks.

fronteer:
> Why the following [#727 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/727#issuecomment-787116274) was ignored?
> 
> I tested commit [bcf71f5](https://github.com/ROCmSoftwarePlatform/MIOpen/commit/bcf71f5583ad1ba017b2194a68abcc0ac6821ade) manually as described at [#727 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/727#issuecomment-786689932). Have you fully tested updated PR on gfx906 and MI100?

Actually, the accumulative error with NORM1 and NORM2 cases exists on all architectures tested by me (gfx908, gfx900 and gfx906)

My updated PR is only tested on gfx900 and gfx908 and gfx906.

fronteer:
Another thing about NORM1 and NORM2 should be said that: 

- For NORM1, our calculation is as  `Sum(abs(X0), abs(X1) ... abs(Xn)) / n`
- For NORM2, our calculation is as  `sqrt( Sum(square(X0), square(X1), ... square(Xn)) / n )`

But from the document of cudnnTensorReduce() API,  I could not know whether **dividing by n** is included in their implementation.  So, should we some how shows the users how our implementation exactly is ?

test/reduce_test.cpp
fronteer: I don't know what your means here?   Had I ignored any update made by you?   I just directly used two `#>git revert`.

atamazov:
@fronteer I am opening a topic about NORM1/2 in the follow-up issue.

test/reduce_test.cpp
atamazov: @fronteer You can use symbolic names like `MIOPEN_REDUCE_TENSOR_NORM1` which is more informative.

Issue #509, opened 2020-10-12
Support for different tensor layouts


atamazov:
This is follow-up of #343, see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343#discussion_r495241947 and other comments there.

The goal is to get uniform support for different tensor layouts at the architecture level, in all library components. Presumably, at first, we need to add layout information into problem description (which solvers or algorithms will support them is another matter).

atamazov:
> we need to add layout information into problem description

The problem is already solved in the `miopen::conv::ProblemDescription` class. It aggregates the copies of tensor descriptors, that contain strides and provide the `GetLayout()` method (that implements approach as proposed at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343#discussion_r471737474).

ppanchad-amd:
@atamazov Is this ticket still relevant? Does it need to be kept open? Thanks!

PR #728, opened 2021-02-04
[NHWC] add gpu reference kernel for nhwc

carlushuang:
* support nhwc/ndhwc, fwd/bwd/wrw, fp32/fp16/bf16, group conv, for reference kernel.
* fix `MIOpenDriver` to have correct result of cpu NHWC convolution.
* refactor `LayoutToStrides()` into `test/tensor_layout.hpp`, to let driver/test code both can call this function
* dump layout info when `MIOPEN_ENABLE_LOGGING_CMD=1`, if currently is not default NCHW layout

## Testing

Basically all test is in [test/gpu_reference_kernel.cpp](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/83dc135c0a9f798597a1d43fbc6461a4cb86c73c/test/gpu_reference_kernel.cpp), this contains cases with a combination of [these conv problems](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/83dc135c0a9f798597a1d43fbc6461a4cb86c73c/test/gpu_reference_kernel.cpp#L100) with a little bit of [randomness](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/83dc135c0a9f798597a1d43fbc6461a4cb86c73c/test/gpu_reference_kernel.cpp#L141) to speedup ctest process, and contains [NCHW/NHWC, 2d/3d, fp32/fp16/bf16](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/83dc135c0a9f798597a1d43fbc6461a4cb86c73c/test/gpu_reference_kernel.cpp#L1069) cases.


driver/conv_driver.hpp
```diff
@@ -1019,9 +1019,12 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
                 break;
```
jerryyin, line 1019: The way you use layout utilities looks good. I'm not a native MIOpen developer so abstain from the rest of code changes. I have two feedbacks:
- Could you add a test stage that use ctest to test specifically NHWC layout, in particular with 1x1 and 3x3 kernels. Some configs to choose from [here](http://confluence.amd.com/pages/viewpage.action?spaceKey=MLSE&title=Convolution+Models+Performance+Optimization#ConvolutionModelsPerformanceOptimization-Resnet50_v1.5,TensorflowNVexample,FP16BS256) are probably good candidates.
- Could you make sure `MIOPEN_ENABLE_LOGGING_CMD=1` populates the correct logging command? This is such that when framework people take your PR and test, they have a good baseline to compare with.

jerryyin (changes requested):

driver/conv_driver.hpp
carlushuang: Hi @jerryyin ,
1. I added a test to test gpu_reference_kernel specifically, with NHWC layout, [in this file](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/728/files#diff-b8f24fcbf6af64fb36fd0cacded514848f06dc3046c6008edd9f293b706ad1a5). This is a ctest and will run while `make check`. But I think you are right, I need at least add a flag to parse in NHWC layout info for test_conv2d. Maybe follow the [MIOpenDriver command line flags]( https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/driver/conv_driver.hpp#L570)
2. `MIOPEN_ENABLE_LOGGING_CMD=1` sure I'll add layout info of this senario.

carlushuang: Hi @jerryyin  @atamazov for add layout print when `MIOPEN_ENABLE_LOGGING_CMD=1`, I updated the changes and now it support it. For test_conv2d to support NHWC layout, it seems need some refactor of test driver structure (it heavily utilize template and may need more time to evaluate), I think maybe left that change in another PR

src/convolution_api.cpp
```diff
@@ -343,6 +344,15 @@ static void LogCmdConvolution(const miopenTensorDescriptor_t xDesc,
                 << " -v " << miopen::deref(convDesc).GetConvStrides()[1]
                 << " -l " << miopen::deref(convDesc).GetConvDilations()[0]
                 << " -j " << miopen::deref(convDesc).GetConvDilations()[1]; // clang-format on
+            std::string x_layout = miopen::deref(xDesc).GetLayout("NCHW");
+            std::string w_layout = miopen::deref(wDesc).GetLayout("NCHW");
+            std::string y_layout = miopen::deref(yDesc).GetLayout("NCHW");
+            if(x_layout != "NCHW")
```
JehandadKhan, line 350: Lets avoid the check and print the layout regardless ?

JehandadKhan:
@asroy  Can I bother you to review the kernel ?

src/convolution_api.cpp
atamazov: If default layouts are not printed, than it will be easier for human readers

src/problem_description.cpp
```diff
@@ -42,6 +42,18 @@ bool ProblemDescription::IsLayoutDefault() const
     }
 }
 
+bool ProblemDescription::IsLayoutNHWC() const
+{
+    if(spatial_dims == 2)
+    {
+        return (in_layout == "NHWC") && (out_layout == "NHWC") && (weights_layout == "NHWC");
```
atamazov, line 49: IIRC the layouts for the input and weights are usually naturally derived from the input layout. Thus, for the NCHW input, the output and weight will always be KCHW and KCYX.

It looks terrible to me that values of `out_layout` and `weights_layout` are different from actual layouts. Do we have any  plans what to do with this?

Are we ever going to support "unnatural" layout combinations? If not, can we just get rid of `out_layout` and `weights_layout`?

driver/conv_driver.hpp
```diff
@@ -1019,9 +1019,12 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
...
             }
 
-            warmup_in  = tensor<warmup_Tgpu>(miopen::deref(warmupInputTensor).GetLengths());
-            warmup_wei = tensor<warmup_Tgpu>(miopen::deref(warmupWeightTensor).GetLengths());
-            warmup_out = tensor<warmup_Tgpu>(miopen::deref(warmupOutputTensor).GetLengths());
+            warmup_in = tensor<warmup_Tgpu>(miopen::deref(warmupInputTensor).GetLengths(),
+                                            miopen::deref(warmupInputTensor).GetStrides());
```
atamazov, line 1023: Optimization for speed?

src/problem_description.cpp
jerryyin: I don't know about if reference kernel support "unnatural" layouts. However, MLIR backend solvers does give us the capability to do "unnatural" layouts convolution. Unless there's strong opposition I hope to still keep the `out-layout` and `weights_layout`.

I'd hope to address the letters as more natural one as much as you think but I think that's not possible without addressing this [issue 509](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/509) first.

jerryyin:
I did a test run for this branch:

> MIOPEN_FIND_MODE=1 MIOPEN_ENABLE_LOGGING=1 MIOPEN_LOG_LEVEL=6 ./bin/MIOpenDriver convfp16 -n 256 -c 128 -H 28 -W 28 -k 512 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -m conv -g 1 -F 4 -t 1

The result is:

> MIOpen Backward Weights Conv. Algorithm: 0, Solution: 33/gemm
GPU Kernel Time Backward Weights Conv. Elapsed: 10.103986 ms (average)
stats: bwdw-conv1x1u1, 256, 128, 28, 28, 1, 1, 512,  26306674688, 0, 0, 2604, 0, 10.103986
Backward Convolution Weights Verifies OK on CPU reference (0.00277143)

Only when I add the `MIOPEN_DEBUG_CONV_GEMM=0` that I can see `Solution: 87/ConvDIrectNaiveConvWrw`.

My question is: why is gemm still selected when I didn't disable it explicitly. I'd assume that either cases, only naive convolution kernel being picked.

driver/conv_driver.hpp
jerryyin: > For test_conv2d to support NHWC layout, it seems need some refactor of test driver structure (it heavily utilize template and may need more time to evaluate ), I think maybe left that change in another PR.

Sounds good to me. Once you wrapped up this PR, could you prioritize on working on that? It makes generic NHWC testing possible with ctest utility. Thanks!

src/convolution_api.cpp
carlushuang: Yes... by default layout is not printout, so here I just not print out if current is indeed `NCHW`

driver/conv_driver.hpp
carlushuang: Nope... this is indeed the way we pass in layout information into `tensor` type, by giving the `stride` information when call constructor.

carlushuang:
> I did a test run for this branch:
> 
> > MIOPEN_FIND_MODE=1 MIOPEN_ENABLE_LOGGING=1 MIOPEN_LOG_LEVEL=6 ./bin/MIOpenDriver convfp16 -n 256 -c 128 -H 28 -W 28 -k 512 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -m conv -g 1 -F 4 -t 1
> 
> The result is:
> 
> > MIOpen Backward Weights Conv. Algorithm: 0, Solution: 33/gemm
> > GPU Kernel Time Backward Weights Conv. Elapsed: 10.103986 ms (average)
> > stats: name, n, c, ho, wo, x, y, k, flopCnt, bytesRead, bytesWritten, GFLOPs, GB/s, timeMs
> > stats: bwdw-conv1x1u1, 256, 128, 28, 28, 1, 1, 512,  26306674688, 0, 0, 2604, 0, 10.103986
> > Backward Convolution Weights Verifies OK on CPU reference (0.00277143)
> 
> Only when I add the `MIOPEN_DEBUG_CONV_GEMM=0` that I can see `Solution: 87/ConvDIrectNaiveConvWrw`.
> 
> My question is: why is gemm still selected when I didn't disable it explicitly. I'd assume that either cases, only naive convolution kernel being picked.

I guess gemm maybe already support NHWC? by some sort of NHWC2NCHW? Anyway, this naive implementation is not performance optimized, so speed might be very slow

src/problem_description.cpp
carlushuang: reference kernel can support different layout, but that still need write by hand. There is no optimization, nor template usage in reference kernel, everything is just to keep it as simple as possible, so in this case reference kernel do lack flexibility compared with MLIR.

So I think, maybe keep check `out_layout` 'weights_layout' here? because we indeed only support this kind of scenario, for now.

driver/conv_driver.hpp
carlushuang: @jerryyin OK

jerryyin:
> I guess gemm maybe already support NHWC? by some sort of NHWC2NCHW? 

I don't think so. As far as I remember, gemm is hard-coded to support NCHW. What this means is that MIOpen profiles the available algorithms, finding the gemm is faster (naturally), and therefore using gemm instead of naive direct convolution. In this situation, execution will give wrong results.

I think what this implies is that you will need to disable gemm in NHWC mode in order not to return wrong results. I will leave it for MIOpen developers to decide whether this needs to be addressed in this PR or file an issue to address it later (in case it got forgotten). For now looks like a short term alternative is to use the macro to disable gemm.

carlushuang:
Let's first open an issue for gemm NHWC https://github.com/ROCmSoftwarePlatform/MIOpen/issues/742

atamazov:
@jerryyin @carlushuang 
> I don't think so. As far as I remember, gemm is hard-coded to support NCHW. 

But how it passed validation then?

carlushuang:
@atamazov I think it is due to computation error. I tested a fwd case, and computation error is larger than default nrms, while @jerryyin 's case may happen to be within default nrms.
```
 ./bin/MIOpenDriver  conv -n 256 -c 128 -H 28 -W 28 -k 512 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -m conv -g 1 -F 1 -t 1
MIOpenDriver conv -n 256 -c 128 -H 28 -W 28 -k 512 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -m conv -g 1 -F 1 -t 1
MIOpen Forward Conv. Algorithm: 0, Solution: 33/gemm
GPU Kernel Time Forward Conv. Elapsed: 1.145352 ms (average)
stats: fwd-conv1x1u1, 256, 128, 28, 28, 1, 1, 512,  26306674688, 103022592, 411041792, 22968, 449, 1.145352
Forward Convolution Failed: 0.28159 > 1.5e-06
```

driver/conv_driver.hpp
jerryyin: Me && Carlus talked offline. Carlus will work on adding NHWC support to ctest starting 2/17. Marking this as resolved.

asroy:
@carlushuang Could you elaborate on what tests you have done for these reference kernels?

carlushuang:
@asroy basically all test is in [test/gpu_reference_kernel.cpp](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/83dc135c0a9f798597a1d43fbc6461a4cb86c73c/test/gpu_reference_kernel.cpp), this contains cases with a combination of [these conv problems](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/83dc135c0a9f798597a1d43fbc6461a4cb86c73c/test/gpu_reference_kernel.cpp#L100), and contains [NCHW/NHWC, 2d/3d, fp32/fp16/bf16](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/83dc135c0a9f798597a1d43fbc6461a4cb86c73c/test/gpu_reference_kernel.cpp#L1069) cases.

driver/conv_driver.hpp
atamazov: Ah, I see. 

[Recommendation] Revert changes for `warmup_*` tensors. These are always NCHW, and strides can be initialized implicitly.

src/problem_description.cpp
asroy: I think it's good to not having assumption on the output layout based on input and weight

atamazov: > Thus, for the NCHW input, the "natural" output and weight will be KCHW and KCYX.

Well, actually tensor layout becomes "KCHW" or "KCYX" only in the context of convolution. But tensors do not depend on the operations in which they are supposed to be used.

atamazov: So,
> It looks terrible to me that values of out_layout and weights_layout are different from actual layouts. Do we have any plans what to do with this?

we can't do anything with this for now.

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -102,6 +81,40 @@ ConvSolution ConvDirectNaiveConvBwd::GetSolution(const ConvolutionContext& ctx)
...
+    else
+        MIOPEN_THROW("Unsupported layout");
+
+    KernelInfo kernel;
+
+    kernel.kernel_file = ConvDirectNaiveConvKernelFile(ctx);
+    kernel.kernel_name = ConvDirectNaiveConvKernelName(ctx);
+    kernel.g_wk.clear();
+
+    kernel.g_wk.push_back(grid_size * block_size);
```
atamazov, line 106: [Notice] If there is possibility that `grid_size * block_size` exceeds INT_MAX, please consider using `size_t`. `g_wk` and `l_wk` in `KernelInfo` are vectors of `size_t`.

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -102,6 +81,40 @@ ConvSolution ConvDirectNaiveConvBwd::GetSolution(const ConvolutionContext& ctx)
...
+    kernel.g_wk.push_back(1);
+    kernel.g_wk.push_back(1);
+    kernel.l_wk.clear();
+    kernel.l_wk.push_back(block_size);
+    kernel.l_wk.push_back(1);
+    kernel.l_wk.push_back(1);
+
+    kernel.comp_options = ConvDirectNaiveConvCompileOption(ctx);
+
+    MIOPEN_LOG_I2(kernel.kernel_file + ":" + kernel.kernel_name);
```
atamazov, line 116: Please remove this. We should have general machinery that prints Solutions (including filenames and kernel names) when necessary.

src/solver/conv_direct_naive_conv_fwd.cpp
```diff
@@ -102,6 +81,40 @@ ConvSolution ConvDirectNaiveConvFwd::GetSolution(const ConvolutionContext& ctx)
...
+    kernel.g_wk.push_back(1);
+    kernel.g_wk.push_back(1);
+    kernel.l_wk.clear();
+    kernel.l_wk.push_back(block_size);
+    kernel.l_wk.push_back(1);
+    kernel.l_wk.push_back(1);
+
+    kernel.comp_options = ConvDirectNaiveConvCompileOption(ctx);
+
+    MIOPEN_LOG_I2(kernel.kernel_file + ":" + kernel.kernel_name);
```
atamazov, line 116: Please remove this.

src/solver/conv_direct_naive_conv_wrw.cpp
```diff
@@ -102,6 +81,27 @@ ConvSolution ConvDirectNaiveConvWrw::GetSolution(const ConvolutionContext& ctx)
...
+    kernel.g_wk.push_back(1);
+    kernel.g_wk.push_back(1);
+    kernel.l_wk.clear();
+    kernel.l_wk.push_back(block_size);
+    kernel.l_wk.push_back(1);
+    kernel.l_wk.push_back(1);
+
+    kernel.comp_options = ConvDirectNaiveConvCompileOption(ctx);
+
+    MIOPEN_LOG_I2(kernel.kernel_file + ":" + kernel.kernel_name);
```
atamazov, line 103: Please remove this.

atamazov (changes requested):
Some minor comments. Otherwise LGTM.

atamazov:
PR description updated with testing info from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/728#issuecomment-778260473

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: OK, let me change this

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: OK

src/solver/conv_direct_naive_conv_fwd.cpp
carlushuang: OK

src/solver/conv_direct_naive_conv_wrw.cpp
carlushuang: OK

src/solver/conv_direct_naive_conv_wrw.cpp
```diff
@@ -81,8 +81,8 @@ ConvSolution ConvDirectNaiveConvWrw::GetSolution(const ConvolutionContext& ctx)
     int c_per_group = c / group;
     int k_per_group = k / group;
 
-    int block_size = 256;
-    int grid_size  = k;
+    size_t block_size = 256;
+    size_t grid_size  = static_cast<size_t>(k);
```
atamazov, line 85: [Note] FYI static cast is not needed here. No need to fix.

atamazov (approved):
LGTM

asroy (approved):

atamazov:

atamazov:
@carlushuang Please first merge from develop, then re-run CI, otherwise your job will never pass. CI job stopped.

atamazov:
@jerryyin Is is PR ready to be merged? There is a blocking review from you.

src/solver/conv_direct_naive_conv_bwd.cpp
atamazov: ?

jerryyin (approved):

carlushuang:
@atamazov  OK let me merge develop

atamazov:
[Gagarin: Off we go! ](https://en.wikipedia.org/wiki/Yuri_Gagarin#Vostok_1) :rocket:

atamazov:
@carlushuang Please look at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532#discussion_r1114926055, thanks!

test/tensor_layout.hpp
atamazov, line 1: 🐛 This PR creates a copy of tensor_layout.hpp from ./src/include/miopen, which is wrong as this creates implicit dependence and error-prone. This immediate bug is that this header uses THE SAME guard as the original: `GUARD_TENSOR_LAYOUT_HPP`.

Issue #561, opened 2020-10-31
Lifecycle of workarounds related to external components


atamazov:
## Motivation

Usually we use workarounds when external components have issues that affect the library. As a rule, such issues begin with some version of the external component, and fixed in one of the subsequent versions.

The library, in turn, is also evolving. During this process, old versions of external components become a bottleneck, and we find ourselves forced to dismiss support of these old versions.

One typical example of such external components is the HIP compiler. A bug appears in some version, then in some other version it is fixed, then at some point in time the old version of the compiler is no longer supported by the library.

These three points should be reflected in the workaround life cycle. Ironically, the main objective of the life cycle is the end of life of the workaround, i.e. cleanup the workarounds that have become unnecessary.

## Guidelines 

- (10) Introducing the workaround
  - (10.10) Define some symbolic entity of the form `...WORKAROUND_<reference-to-the-issue>...`. Consistently use it across the source code.
    - Use `..._ISSUE_XXX` in the macro name to refer to the private repo (___preferred___).
    - Use `..._MIOPEN_ISSUE_XXX` to refer to the [public repo](https://github.com/ROCmSoftwarePlatform/MIOpen).
    - Use `..._SWDEV_XXX` to denote Jura issue.
    - _Example for C/C++_: macro definition like `#define WORKAROUND_ISSUE_2331 1`.
  - (10.20) Mixing several issues in the single symbol, e.g. `WORKAROUND_ISSUE_2174_2222_2224_2243`, is ___not allowed___. This situation means either duplicate issues or different issues that can be resolved by identical workaround.
    - Duplicates shall be resolved within bug tracking system, so only one is left open.
    - Otherwise we shall use the boolean logic, like `#if WORKAROUND_ISSUE_2174 || WORKAROUND_ISSUE_2222 || WORKAROUND_ISSUE_2224` for C/C++.
  - (10.30) Check previous version of the external component.
    - If the previous version is free of this issue, then use version detection machinery and switch the workaround off for previous and older versions.
      - This may require the development of the version detection machinery for the component, if we do not have it already.
    - Otherwise keep workaround enabled unconditionally.
- (20) Disable the workaround after the fix.
  - This needs to be done as soon as the issue in the external component is fixed, even prior the official release of it.
  - Use version detection (see 10.30 above) and disable workaround for the `fixed-version` and newer versions.
- (30) End of life - clean up.
[...]

PR #730, opened 2021-02-05
Batchnorm fix for disabling ASM code

muralinr:
This is a temporary workaround fix to disable batchnorm backward code.


daniellowell (approved):

atamazov:
@muralinr 
> This is a temporary workaround fix to disable batchnorm backward code.

Workarounds should refer to the original problem, see #561. However, in this particular case, I would say that this is not a workaround, but a bug fix -- you disable kernel for an unsupported target device.

atamazov (approved):

PR #741, opened 2021-02-10
Remove outdated (and now wrong & harmful) workaround for apple

atamazov:
The reason of the issue: https://github.com/AMDComputeLibraries/MLOpen/pull/2400/files#r572106557. Unlike `AddProgram` and `AddKernel`, `HasProgram` does not use the `ProcessParams` trick, that adds a space to the compile args. That is why `HasProgram` fails to find existing program, if it has been added by `AddProgram` or `AddKernel`. Consequently, `PrecompileSolutions` rebuilds programs that already reside in the binary cache.

However we do not need `ProcessParams` at all, because currently -cl-std is appended to the end or params ([see comment in `ProcessParams`](https://github.com/AMDComputeLibraries/MLOpen/commit/2358609cfcc3d6d84245d89636b61b5b14d8d733#r46885082) that explains why it is required).

### DIFF
--- a/src/kernel_cache.cpp
+++ b/src/kernel_cache.cpp
@@ -52,19 +52,6 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEVICE_ARCH)
 
 namespace miopen {
 
-static void ProcessParams(std::string& params)
-{
-    if(params.length() > 0)
-    {
-        // Ensure only one space after the -cl-std.
-        // >1 space can cause an Apple compiler bug. See clSPARSE issue #141.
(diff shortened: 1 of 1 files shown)


JehandadKhan (approved):
I agree with this change, once merged please let me know so I can revert the adjustment for this in Tuna

atamazov:

atamazov:
@JehandadKhan 
> I agree with this change, once merged please let me know so I can revert the adjustment for this in Tuna

Is now the appropriate time to merge this and update Tuna?

atamazov:
All tests passed.

PR #748, opened 2021-02-20
Winograd binaries to sources

Slimakanzer:
This PR replaces the winograd binaries with winograd assembly sources.
Please note that some winograd binaries are compatible with _gfx9_ and _gfx8_, but the assembly code is different. Therefore, two assembly source versions (__gfx8_, __gfx9_) are added for such kernels.

## Local verification
All replaced kernels have been verified using configuration:
* gfx1030 ROCm 3.10
* gfx906 ROCm 4.0
* gfx906 ROCm 3.5
* gfx803 ROCm 3.10

Also kernel assembly was tested with **Code Object V2** and **Code Object V3** for shaders that support both versions of metadata (using _MIOPEN_DEBUG_AMD_ROCM_METADATA_ENFORCE_ option to enforce CO version)


src/kernels/conv_3x3_wheel_alpha_v9_2_7.s
```diff
@@ -24,5 +24,9 @@
  *
  *******************************************************************************/
 .include "conv_3x3_wheel_alpha_v9_2_7_prologue.inc"
-.include "conv_3x3_wheel_alpha_v9_2_7.inc"
+.if (.amdgcn.gfx_generation_number == 8)
```
atamazov, line 27: CO V2 assember generates old predefined symbol, `.option.machine_version_major`, NOT `.amdgcn.gfx_generation_number`. So we are losing CO V2 compatibility.

I recommend including `rocm_version.inc`, that defines old symbols for CO V3, and using old symbols here.

[Note] Currently all ASM sources that support CO V2 use old `option...` symbols. Eventually we will change `rocm_version.inc` so it will define new symbols for CO V2 and use new `amdgcn...` symbols everywhere. If you wish, you can include this change right in this PR.

src/kernels/conv_3x3_wheel_alpha_v9_0_15_stride_2_dil.s
```diff
@@ -24,5 +24,9 @@
  *
  *******************************************************************************/
 .include "conv_3x3_wheel_alpha_v9_0_15_prologue.inc"
-.include "conv_3x3_wheel_alpha_v9_0_15_stride_2_dil.inc"
+.if (.amdgcn.gfx_generation_number == 8)
```
atamazov, line 27: Ditto

src/kernels/conv_3x3_wheel_alpha_v9_0_15_stride_2_dec.s
```diff
@@ -24,5 +24,9 @@
  *
  *******************************************************************************/
 .include "conv_3x3_wheel_alpha_v9_0_15_prologue.inc"
-.include "conv_3x3_wheel_alpha_v9_0_15_stride_2_dec.inc"
+.if (.amdgcn.gfx_generation_number == 8)
```
atamazov, line 27: Ditto.

src/kernels/conv_3x3_wheel_alpha_v9_0_15.s
```diff
@@ -24,5 +24,9 @@
  *
  *******************************************************************************/
 .include "conv_3x3_wheel_alpha_v9_0_15_prologue.inc"
-.include "conv_3x3_wheel_alpha_v9_0_15.inc"
+.if (.amdgcn.gfx_generation_number == 8)
```
atamazov, line 27: Ditto.

src/kernels/conv_3x3_wheel_alpha_v9_2_7_stride_2_dec.s
```diff
@@ -24,5 +24,9 @@
  *
  *******************************************************************************/
 .include "conv_3x3_wheel_alpha_v9_2_7_prologue.inc"
-.include "conv_3x3_wheel_alpha_v9_2_7_stride_2_dec.inc"
+.if (.amdgcn.gfx_generation_number == 8)
```
atamazov, line 27: Ditto.

atamazov (changes requested):
:+1: but we shall do something with CO V2.

atamazov:
Is it possible to smoke test also with ROCm 3.5?

JehandadKhan:
@Slimakanzer can we remove the redundant kernels first as discussed earlier and only convert the ones that remain?

atamazov:
@JehandadKhan 
> @Slimakanzer can we remove the redundant kernels first as discussed earlier and only convert the ones that remain?

Let's do that separately because this requires extensive performance testing.

/cc @Kirpich30000

src/kernels/conv_3x3_wheel_alpha_v9_2_7.s
Slimakanzer: Thank you for your careful reviewing, fixed in 4622fe5.
I also locally tested such cases and updated PR description.

src/kernels/Conv_Winograd_v13_3_12_fp16dot_stride1.inc
```diff
@@ -23,2975 +23,3038 @@
...
+s_mov_b32 s46, 0x80000000
+s_mov_b32 s47, 0x20000
+s_mov_b32 s50, 0x80000000
+s_mov_b32 s51, 0x20000
+v_cmp_le_u32_e32 vcc, 0x100, v0
+s_cbranch_vccnz 7
+v_xor_b32_dpp v108, v0, v0  quad_perm:[1,3,2,2] row_mask:0xf bank_mask:0xf
+v_subrev_co_u32_e32 v108, vcc, 1, v108
+v_cvt_f16_i16_e32 v108, v108
+.long 0xd38f406c
```
Kirpich30000, line 385: Such .long directives should be replaced with actual instruction or macro

Slimakanzer: Thank you! Fixed in e5b26b2

atamazov (dismissed):

Kirpich30000:
The v21 version is also affected by the .long issue. Let's fix it also and make sure the binary stays the same.

atamazov:
@Kirpich30000 Does this fix important for 4.2?

atamazov (dismissed):
LGTM

Kirpich30000 (approved):
LGTM

Slimakanzer:
@atamazov @Kirpich30000 
I apologize for the discarded reviews. Can you check new changes for gfx10?

atamazov:
@Slimakanzer 
> I apologize for the discarded reviews. Can you check new changes for gfx10?

No problem at all, the recent fixes are important! :+1:

atamazov (approved):
GOOD.

Kirpich30000 (approved):
LGTM

atamazov:

PR #750, opened 2021-02-22
[tests] Remove useless tests from low-precision stages.

atamazov:
This resolves issue described at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/139/files#r578672245. This is expected to speed up CI testing. Also some other edits for clarity.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -56,12 +56,15 @@ add_custom_target(tests)
 set(SKIP_TESTS)
 set(MIOPEN_TEST_FLOAT_ARG)
 
+# The usage is non-trivial, see function add_test_command.
+set(SKIP_ALL_EXCEPT_TESTS)
+
 if(MIOPEN_TEST_HALF)
     if(MIOPEN_BACKEND_OPENCL)
         set(SKIP_TESTS test_gru test_rnn_vanilla test_lstm test_conv_igemm_dynamic)
     endif()
     if(MIOPEN_TEST_GFX908)
-       set(SKIP_TESTS test_immed_conv3d test_conv3d test_fusion_aux test_activation test_lrn_test test_ctc test_conv2d_bias test_conv3d_bias test_cba_inference test_cbna_inference test_pooling2d test_na_train test_na_inference test_bn_aux test_conv_igemm_dynamic)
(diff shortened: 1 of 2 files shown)


test/CMakeLists.txt
```diff
@@ -218,7 +221,7 @@ function(add_custom_test NAME)
     OR (MIOPEN_TEST_BFLOAT16 AND ${PARSE_ALLOW_BFLOAT16})
     OR (MIOPEN_TEST_HALF AND ${PARSE_ALLOW_HALF})
     OR (MIOPEN_TEST_INT8 AND ${PARSE_ALLOW_INT8})
-    OR (NOT (MIOPEN_TEST_MIOTENSILE AND (NAME IN_LIST SKIP_TESTS))))
+    OR (MIOPEN_TEST_MIOTENSILE AND NOT(NAME IN_LIST SKIP_TESTS)))
```
atamazov, line 224: This is the most important fix.

carlushuang: if `MIOPEN_TEST_MIOTENSILE=Off` and under fp32 case, all test will be added due to first condition `NOT (MIOPEN_TEST_INT8 OR MIOPEN_TEST_BFLOAT16 OR MIOPEN_TEST_HALF OR MIOPEN_TEST_MIOTENSILE))`, hence even if add to `SKIP_TESTS` of vega20 case, if on gfx908 machine still will be added?

JehandadKhan (dismissed):

test/CMakeLists.txt
atamazov: If you look closely, then you'll see that SKIP_TESTS (as well as SKIP_ALL_EXCEPT_TESTS) are NOT intended to affect "custom" tests. Adding "custom" tests to SKIP_TESTS is basically meaningless (in theory, such things should be fixed, but this is out of scope of this PR).

AFAICS the only exception is MIOPEN_TEST_MIOTENSILE testing mode. In this mode, if name of a "custom" test is added to SKIP_TESTS, then that "custom" test will be skipped (unless low-precision testing mode & respective ALLOW flag is specified). This hack was added in #139.

`test_conv_igemm_dynamic` is a "custom" test so it will be used on FP32 tests on all targets, including Vega20 and gfx908. If you want to exclude it for some targets, I recommend enclosing relevant `add_custom_test` call into `if`, like this:
```
if(NOT MIOPEN_TEST_GFX908)
    add_custom_test(test_conv_igemm_dynamic_small
    ...
    )
endif()
```

carlushuang: OK, I see. Let me comment in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/649

shurale-nkn (dismissed):
LGTM

shurale-nkn (approved):
LGTM

PR #738, opened 2021-02-09
Remove HCC

daniellowell:
depends on PR #147 

Contains tidy issues which need resolving.

PR #759, opened 2021-02-28
[CI] Resolve issues after upgrading to ROCm 3.7. Switch to hip-clang in all tests that use GPU.

atamazov:
Supersedes #738. Static checks will be updated separately.


dev-requirements.txt
```diff
@@ -1,5 +1,5 @@
 pfultz2/rocm-recipes
-ROCm-Developer-Tools/HIP@roc-3.0.0
+# ROCm-Developer-Tools/HIP@rocm-3.7.0
```
pfultz2, line 2: This version of hip is only for hcc. When we switch to hip-clang, we can remove these dependencies.

atamazov: @pfultz2 As you can see, this PR in fact switches all tests (but static tests) to hip-clang. Unfortunately, docker build fails. I've tried some fixes, but no success yet. ___I will appreciate if you look at build logs at Jenkins and give me some advice.___

/cc @shurale-nkn

hip-clang.docker
```diff
@@ -49,21 +53,41 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
...
+
+# Make sure /opt/rcom is in the paths
+ENV PATH="/opt/rocm:${PATH}"
+
 # Build using hip-clang
-RUN cget -p $PREFIX init --cxx /opt/rocm/llvm/bin/clang++ --std=c++14 -DAMDGPU_TARGETS=${GPU_ARCH}
+RUN cget -p $PREFIX init --cxx g++ --std=c++14 -DAMDGPU_TARGETS=${GPU_ARCH}
 
 # Install dependencies
-RUN cget -p $PREFIX install pfultz2/rocm-recipes
```
shurale-nkn, line 62: Bring it back, it is required for min-requirements downloading

hip-clang.docker
```diff
@@ -49,21 +53,41 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-
...
+ADD dev-requirements.txt /dev-requirements.txt
+ADD requirements.txt /requirements.txt
 ADD min-requirements.txt /min-requirements.txt
-RUN CXXFLAGS='-isystem $PREFIX/include' cget -p $PREFIX install -f /min-requirements.txt
-RUN export HIPCC_LINK_FLAGS_APPEND='-O3 -parallel-jobs=4'
-RUN export HIPCC_COMPILE_FLAGS_APPEND='-O3 -Wno-format-nonliteral -parallel-jobs=4'
+RUN locale
+RUN export HIPCC_LINK_FLAGS_APPEND='-O3 -parallel-jobs=4' && \
+    export HIPCC_COMPILE_FLAGS_APPEND='-O3 -Wno-format-nonliteral -parallel-jobs=4' && \
+    CXXFLAGS='-isystem $PREFIX/include' cget -p $PREFIX install -f /dev-requirements.txt
```
shurale-nkn, line 85: You don't need to install dev-requirements.txt, only min-requirements.txt

shurale-nkn (changes requested):

hip-clang.docker
shurale-nkn: result in branch:
https://github.com/ROCmSoftwarePlatform/MIOpen/tree/help_with_ci

dev-requirements.txt
atamazov: Change reverted. This is used only in hip3.0 based dockers that will be used for static checks for a while. @shurale-nkn Is working on a patch that should switch static checks to hip-clang.

atamazov:
It seems that this PR works fine now. Under testing in our CI right now.

shurale-nkn (approved):
LGTM

atamazov:
Static checks & all "fast" tests passed.

PR #761, opened 2021-03-01
Tidy update

shurale-nkn:
Static checks update for modern rocm hip-clang.

Stages with HCC usage and hcc dockerfile was removed.
Docker image building now does not prevent the task from termination.


src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -94,7 +94,7 @@ float CallImplGemmDynamicForward1x1(const miopen::Handle& handle,
     int dilation_w  = conv_problem.GetDilationW();
     int pad_h       = conv_problem.GetPadH();
     int pad_w       = conv_problem.GetPadW();
-    int __pack0     = 0;
+    int __pack0     = 0; //NOLINT (bugprone-reserved-identifier)
```
pfultz2, line 97: This variable should be renamed.

src/include/miopen/mlo_utils.hpp
```diff
@@ -84,7 +84,7 @@ typedef unsigned int uint;
 #include <mach/mach.h>
 #include <mach/mach_time.h>
 #endif
-using __int64 = long long;
+using __int64 = long long; // NOLINT (bugprone-reserved-identifier)
```
pfultz2, line 87: This type should be renamed as it can conflict with identifiers in the runtime.

test/reduce_test.cpp
```diff
@@ -777,8 +779,8 @@ struct reduce_driver : test_driver
             assert(toReduceDims[i] < inLengths.size());
 
         // set the lengths of the dimensions to be reduced to 1 to represent the output Tensor
-        for(int i                       = 0; i < toReduceDims.size(); i++)
-            outLengths[toReduceDims[i]] = static_cast<std::size_t>(1);
+        for(int& toReduceDim : toReduceDims)
```
pfultz2, line 782: This should be `const int&` or just `int`.

CMakeLists.txt
```diff
@@ -431,22 +431,26 @@ include(ClangTidy)
...
+        -cert-dcl37-c
+        -cert-dcl51-cpp
         -clang-analyzer-alpha.core.CastToStruct
         -clang-analyzer-optin.performance.Padding
         -clang-diagnostic-deprecated-declarations
         -clang-diagnostic-extern-c-compat
         -clang-diagnostic-unused-command-line-argument
         -cppcoreguidelines-avoid-c-arrays
         -cppcoreguidelines-avoid-magic-numbers
+        -cppcoreguidelines-avoid-non-const-global-variables
```
pfultz2, line 453: This seems like an important check that we should not disable. Why is this necessary?

shurale-nkn: Because I thought we were using a lot of global variables like " extern bool"
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f8f7e325cd2a9c2e0a7e3944f226c8b800b87968/src/include/miopen/logger.hpp#L202
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f8f7e325cd2a9c2e0a7e3944f226c8b800b87968/src/include/miopen/solver.hpp#L53
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f8f7e325cd2a9c2e0a7e3944f226c8b800b87968/src/include/miopen/find_controls.hpp#L41
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f8f7e325cd2a9c2e0a7e3944f226c8b800b87968/src/include/miopen/find_db.hpp#L64

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f8f7e325cd2a9c2e0a7e3944f226c8b800b87968/src/solver/conv_direct_naive_conv.cpp#L38

but there were only 5 of them

shurale-nkn: I'll just add them to the exceptions

pfultz2: Can we and `NOLINT` to those lines instead?

Also, `AlwaysEnableConvDirectNaive` can be removed as its not used.

shurale-nkn: Yes, by exception I meant NOLINT

atamazov:

test/perfdb.cpp
```diff
@@ -51,21 +51,25 @@ namespace tests {
...
+    // NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
     static std::mutex mutex;
     std::lock_guard<std::mutex> lock(mutex);
 
+    // NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
     static boost::optional<std::string> path(boost::none);
     return path;
 }
 
 static bool& full_set()
```
shurale-nkn, line 70: Does anyone know if this feature is necessary? Or can it be removed?

shurale-nkn:
static checks passed

test/perfdb.cpp
atamazov: @shurale-nkn please ask @DrizztDoUrden

atamazov:
@shurale-nkn could you please merge latest develop? It contains massive change (Fwd GEMM invokers) and may hide some problems that will appear with this tidy update.

atamazov:
@shurale-nkn BTW we will need to ask everyone to update their PRs as soon as this merged.

src/solver/conv_asm_implicit_gemm_wrw_gtc_dynamic_xdlops.cpp
```diff
@@ -44,6 +44,7 @@ GetImplicitGemmWrwGTCDynamicXdlopsKernelList()
 {
     // retrieve dynamic igemm wrw pass's possible kernel name
     // clang-format off
+    // NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
```
atamazov, line 47: And here (and wherever similar cases appear)

test/pooling_common.hpp
```diff
@@ -48,11 +48,15 @@
 #include "cpu_conv.hpp"
 
 #define TEST_PADDING_MODE 0
-
-static int num_uint16_case        = 0;
-static int num_uint32_case        = 0;
+// NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
```
atamazov, line 51: [Noice] Tests are cans of worms

driver/conv_driver.hpp
```diff
@@ -125,19 +125,6 @@ struct AutoMiopenWarmupMode
     bool debug_find_enforce_disable_prev;
 };
 
-struct AutoConvDirectNaiveAlwaysEnable
```
atamazov, line 128: Can you bring this back?

CMakeLists.txt
atamazov: > Also, AlwaysEnableConvDirectNaive can be removed as its not used.

It is used.

test/perfdb.cpp
atamazov: @DrizztDoUrden ping

atamazov (changes requested):
One blocking comment

test/perfdb.cpp
DrizztDoUrden: I am not sure if anyone commonly uses this now, but in ancient times I have used it to debug PlainTextDb in a more stressful test. Is there any specific reason to remove it? It helps to triage db MP/MT failures, but one can just change values manually for the same result.

atamazov: @DrizztDoUrden @shurale-nkn It is used here, not Ok to remove:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/eae8583eb49db4dcc874d2eadee9051e70f9eae4/test/perfdb.cpp#L1290

shurale-nkn: if true than true?)

atamazov: Read carefully: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/eae8583eb49db4dcc874d2eadee9051e70f9eae4/test/perfdb.cpp#L1288-L1290

atamazov (approved):

test/perfdb.cpp
shurale-nkn: I'm not sure is it a good decision. We are using static global tests::full_set() only in DbMultiProcessReadTest().Run(); and DbMultiProcessTest().Run(); which are called by the 6 lines below in same function, which means we can pass it as argument.

shurale-nkn: The problem is that these ... ambiguous solutions, become the rule, not the exception, and they begin to be thoughtlessly copied.

atamazov: This decision is not good, but addressing it is out of the scope of this PR. /cc @DrizztDoUrden

DrizztDoUrden (approved):
LGTM

test/perfdb.cpp
DrizztDoUrden: If you are talking about static variable specifically rather than --all mode for this test in general, I am 100% ok with changing it to a method parameter.

PR #765, opened 2021-03-03
[NHWC] support nhwc in ctest

carlushuang:
This PR depend on https://github.com/ROCmSoftwarePlatform/MIOpen/pull/728
support parse argument `--in_layout`, `--fil_layout`, `--out_layout` to conv_2d/conv_3d driver to support different tensor layout

```
export MIOPEN_DEBUG_CONV_GEMM=0

./bin/test_conv2d --verbose --input 8 4 32 32 --weights 8 4 3 3 --pads_strides_dilations 1 1 1 1 1 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
./bin/test_conv2d --verbose --input 16 128 36 36 --weights 32 128 1 1 --pads_strides_dilations 0 0 1 1 1 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
./bin/test_conv2d --verbose --input 128 768 17 17 --weights 128 768 1 1 --pads_strides_dilations 0 0 1 1 1 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
./bin/test_conv2d --verbose --input 128 128 35 35 --weights 128 128 3 3 --pads_strides_dilations 1 1 1 1 1 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
./bin/test_conv2d --verbose --input 64 3 224 224 --weights 64 3 7 7 --pads_strides_dilations 3 3 2 2 1 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
./bin/test_conv2d --verbose --input 8 512 4 4 --weights 512 16 1 1 --pads_strides_dilations 0 0 1 1 1 1 --cmode trans --pmode valid --group-count 32 --in_layout NHWC --fil_layout NHWC --out_layout NHWC

./bin/test_conv3d --verbose --conv_dim_type conv3d --input 16 32 4 9 9 --weights 64 32 3 3 3 --pads_strides_dilations 0 0 0 2 2 2 1 1 1 --in_layout NDHWC --fil_layout NDHWC --out_layout NDHWC
./bin/test_conv3d --verbose --conv_dim_type conv3d --input 16 128 4 56 56 --weights 256 4 3 3 3 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --group-count 32 --in_layout NDHWC --fil_layout NDHWC --out_layout NDHWC
./bin/test_conv3d --verbose --conv_dim_type conv3d --input 16 64 3 4 4 --weights 64 32 1 3 3 --pads_strides_dilations 0 0 0 1 2 3 1 2 3 --trans_output_pads 0 0 0 --group-count 4 --cmode trans --pmode default --in_layout NDHWC --fil_layout NDHWC --out_layout NDHWC
```

but below cases can not pass (no suitable solution)
```
./bin/test_conv2d --verbose --input 8 3 108 108 --weights 63 1 3 3 --pads_strides_dilations 1 1 1 1 1 1 --group-count 3  --in_layout NHWC --fil_layout NHWC --out_layout NHWC
./bin/test_conv3d --verbose --conv_dim_type conv3d --input 4 4 4 161 700 --weights 32 1 3 5 20 --pads_strides_dilations 1 1 1 2 2 2 1 1 1 --group-count 4 --in_layout NDHWC --fil_layout NDHWC --out_layout NDHWC
```

The reason is not because of this PR, but because a bug in [TensorDescriptor::GetLayout()](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/tensor.hpp#L188).
Issue is created for this problem: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/767


jerryyin:
I worked out a heuristic such that it: 
 - Do a depth first search to generate all possible layout
 - Examine every possible layout by checking if dimension generated strides can match the source stride. This way we can eliminate the "wrong" ambiguous results. Refer to the example above, only 2 out of 6 are valid cases.
 - Pick "NCHW" if it exists in possible results. Otherwise, pick "NHWC". Or else, pick the first possible layout. In the above example, NHWC will be picked because it has higher priority over the 2 valid cases.

The minimized example is [here](https://onlinegdb.com/r1F99pIQd). 

-----

Please note, this is **untested** in miopen scope yet.

carlushuang:
OK to run conv_2d/3d with `C=1` test (in above description) and verified in miopen log that it peek NHWC/NDHWC naive kernel successfully.

src/include/miopen/tensor_layout.hpp
```diff
@@ -79,6 +89,123 @@ inline std::string tensor_layout_get_default(int size)
...
+        // std::cout << tmp << std::endl;
+        all_layouts.push_back(tmp);
+        return;
+    }
+
+    for(int i = 0; i < labels.size(); ++i)
+    {
+        // Skip current label if it already exists
+        if(find(tmp.begin(), tmp.end(), labels[i]) != tmp.end())
+            continue;
```
jerryyin, line 117: TODO: use an additional `is_used` vector instead of find to reduce time complexity.

src/include/miopen/tensor_layout.hpp
```diff
@@ -27,14 +27,24 @@
 #define GUARD_TENSOR_LAYOUT_HPP
 
 #include <map>
+#include <unordered_map>
 #include <algorithm>
 #include <vector>
+#include <numeric>
 #include <string>
 #include <iterator>
 #include <miopen/errors.hpp>
```
jerryyin, line 37: TODO: wrap this header into a proper class, not a collection of static functions.

src/include/miopen/tensor.hpp
```diff
@@ -175,37 +176,44 @@ struct TensorDescriptor : miopenTensorDescriptor
 
     std::string ToString() const;
 
-    template <class Vector, class Op>
-    static inline std::vector<int64_t> sort_permutation(const Vector& data, Op op)
+    std::string GetLayout(std::string labels) const
```
jerryyin, line 179: Seek for feedback on this:

- If we do store layout, then get rid of labels because layout will be based on a default set of labels.
- If we don't store layout, then I can bring the `setLayout()` logic to here, and instead of setting it to a class member, return it.

pfultz2: If you want to cache the computation you could cache the permutation, and the layout can apply the permutation to the string layout.

src/include/miopen/tensor.hpp
```diff
@@ -175,37 +176,50 @@ struct TensorDescriptor : miopenTensorDescriptor
...
-    std::string GetLayout(std::string labels) const
+    std::string ComputeLayout(const std::string& labels)
     {
-        if(labels.size() != strides.size())
+        //std::cout << "lens: ";
+        //printVector(lens);
+        //std::cout << "Strides: ";
+        //printVector(strides);
+
+        auto valid_layouts = compute_valid_layouts(strides, lens, labels);
```
pfultz2, line 192: I dont think we need to compute all layouts. We can compute the permutation correctly by minimizing the adjacent difference. This can be done by grouping the duplicate strides and then sort them based on the difference from the previous index:

```cpp
template <class Iterator, class Output, class Predicate>
void group_unique(Iterator start, Iterator last, Output out, Predicate pred)
{
    while(start != last)
    {
        auto it = std::find_if(start, last, [&](auto&& x) { return not pred(*start, x); });
        out(start, it);
        start = it;
    }
}

std::vector<int64_t> find_permutation(const std::vector<std::size_t>& strides)
{
    std::vector<std::int64_t> result;
    std::multimap<std::size_t, int64_t, std::greater<>> m;
    int i = 0;
    for(auto&& x : strides)
        m.emplace(x, i++);
    auto pred     = by(std::equal_to<>{}, [](auto&& p) { return p.first; });
    int64_t base  = 0;
    int64_t delta = 0;
    auto each     = [&](auto start, auto last) {
        auto n = std::distance(start, last);
        std::transform(start, last, std::back_inserter(result), [](auto&& p) { return p.second; });
        auto mag = delta > 0 ? -1 : 1;
        std::stable_sort(result.end() - n, result.end(), by(std::less<>{}, [&](auto x) {
                             return (x - base) * mag;
                         }));
        delta = result.back() - base;
        base  = result.back();
    };
    group_unique(m.begin(), m.end(), each, pred);
    return result;
}
```

src/include/miopen/tensor.hpp
pfultz2: The disadvantage of storing the permutation or layout string in the tensor is that you can't use the other tensors to help infer the correct layout.

So, alternatively, we could compute the layout in the ProblemDescription. If we write a `IsLayout("NCHW", "NHWC")` function to check if the tensor has the layout. Then you can just loop over the two layouts:

```cpp
std::set<std::string> inputLayouts;
std::set<std::string> outputLayouts;

for(const std::string& layout:{"NCHW", "NHWC"})
{
    if (in.IsLayout("NCHW", layout))
        inputLayouts.insert(layout);
    if (out.IsLayout("NCHW", layout))
        outputLayouts.insert(layout);
}
```

Then you know if the layout is ambiguous because the set may be larger than one element. When its ambiguous, its likely that both input and output are not ambiguous so you can do a set_intersection to prune out the ambiguous layouts. A similar approach can be applied to the filter.

src/include/miopen/tensor.hpp
pfultz2: Although this will still be ambiguous for cases like `64, 1, 1, 1`.

src/include/miopen/tensor.hpp
jerryyin: > If we write a IsLayout("NCHW", "NHWC") function to check if the tensor has the layout. 

I like this for its elegance. However, I think the problem of this approach is:
- Reduced the supported layout from arbitrary(by igemm kernels) to 2(NCHW and NHWC)
- Other components inside MIOpen can reach at different conclusion than the decision made in `ProblemDescription`. @carlushuang's rest of PR has a number of use cases where he will invoke `GetLayout()` on an individual tensor. At those use cases, `ProblemDescription` is not available in the context. The conclusions made with individual `GetLayout()` can be inconsistent with conclusion made in `ProblemDescription`.

We can adopt this if we agree:
- Do not support other non-trivial layout
- @carlushuang Need to figure out how to get coherent views of layout in different component of MIOpen.

src/include/miopen/tensor.hpp
jerryyin: > auto pred     = by(std::equal_to<>{}, [](auto&& p) { return p.first; });

@pfultz2 Just for understanding what does this `by` keyword do?

> then sort them based on the difference from the previous index

The original index is based `NCHW`, why is `NCHW` the golden standard to judge all other layouts?

src/include/miopen/tensor.hpp
jerryyin: > If you want to cache the computation you could cache the permutation, and the layout can apply the permutation to the string layout.

Thank you. I like this better than the current approach. Indeed caching the permutation gives more flexibility for us to assign different labels to the layout.

jerryyin:
@carlushuang This PR is still quite important to unblock rest of the changes needed for June Demo. Just want to make sure we are on the same page:
 - Do you have other changes to make in this PR or this is almost ready?
 - Do you find any bug in existing implementation of the heuristic, or have any thoughts between me/ @pfultz2 comments?

src/include/miopen/tensor.hpp
pfultz2: > Just for understanding what does this by keyword do?

It applies projection to an operator. It is defined [here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/functional.hpp#L46). So `by(f, p)(x, y)` is equivalent to `f(p(x), p(y))`.

> The original index is based NCHW, why is NCHW the golden standard to judge all other layouts?

Because that is the order of dimensions regardless of layout in memory.

src/include/miopen/tensor.hpp
pfultz2: > Reduced the supported layout from arbitrary(by igemm kernels) to 2(NCHW and NHWC)

Well when we assign a layout in the ProblemDescription, it can be an enum of NCHW, NHWC, or Unknown. So if we have a solver that supports arbitrary layouts this enum can be ignored. 

> Other components inside MIOpen can reach at different conclusion than the decision made in ProblemDescription

Yes but to be able to do that in the solvers the ProblemDescription needs to store the tensor descriptors.

> rest of PR has a number of use cases where he will invoke GetLayout() on an individual tensor. 

First, in this PR `GetLayout` is only used to compute the output tensor which is only used for testing(and the driver). In MIGraphX, we have a [`with_lens`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/data-layout/src/shape.cpp#L225) function that will try to propagate the layout, you should add a similar method. 

Second, for testing we should probably explicitly call the output layout we want rather than trying to infer it from the strides of the input. Also, because we will want to test having different input and output layouts in the future possibly.

Thirdly, if we cache a single interpretation of the permutation in the tensor descriptor then there is no room to interpret it as a different layout in ambiguous cases.

carlushuang: Yes currently in this PR if all layout is explicitly specified by `--in_layout`, `--fil_layout`, `--out_layout`, will use this layout specification.

carlushuang:
@jerryyin this PR is ready and no further changes. I'm relaunching https://github.com/ROCmSoftwarePlatform/MIOpen/pull/728 CI and hopefully if passed & merged this PR can remove draft label. Current the layout implementation is fine for this PR and other comments please see above conversation

src/include/miopen/tensor.hpp
carlushuang: > At those use cases, ProblemDescription is not available in the context. The conclusions made with individual GetLayout() can be inconsistent with conclusion made in ProblemDescription

Yes, and in ctest we may have not yet created `ProblemDescription` but do need layout information just from `TensorDescriptor`.

src/include/miopen/tensor.hpp
jerryyin: I still don't have a good enough understanding of your algorithm.

> auto it = std::find_if(start, last, [&](auto&& x) { return not pred(*start, x); });

`pred()` returns true if strides are equal, but with a not in front you will find those cases whose strides are different. I'd thought you would do otherwise and only calculate the adjacent difference for strides equal cases because those are ambiguous indexes.

>    int64_t base  = 0;
>    int64_t delta = 0;
>    auto each     = [&](auto start, auto last) { ... }

I got completely lost in what it is trying to do in this lambda. I was trying to come up with a minimal example but can't get it to compile, [here](http://cpp.sh/83qpz). Could you help with making it to compile, and click on "GetUrl", just so that I can step through for better understanding.

I do realize that it doesn't try to check the calculated layout's validity like:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/5203a870b3f27d91750e17d10112565545138b60/src/include/miopen/tensor_layout.hpp#L179-L188

I think this is still essential to include to confirm that derived strides are consistent with lens.

src/include/miopen/tensor.hpp
jerryyin: > So if we have a solver that supports arbitrary layouts this enum can be ignored.

We don't have this enum right now. Are you suggesting to start with only `NCHW` and `NHWC` and come up with the enum in a later API change, then update the heuristic again? 
 - Follow up: Let me rephrase: when it is a `UNKNOWN` enum and and it is ambiguous, we skip the `IsLayout()` checks and directly use the layout getter from `tensor.hpp`.

If this is the case, I see a two step process here, and we don't necessarily address all concerns in this PR alone.
- step 1: `tensor.hpp::IsLayout()` + Find Heuristic in ProblemDescription, only enable `NCHW` + `NHWC` + enforce ctest to know all layout deterministically from user input
- step 2: `tensor.hpp::find_permutation` + API function to set enum layout, enable `nonNCHW/NHWC`

jerryyin: I coded step 1 in commit 31e693f.

This commit is independent from 2b80369. It implements the heuristic init as @pfultz2 suggested. We can revert 2b80369 until we reach at a consensus, then implement step 2.

src/include/miopen/tensor.hpp
jerryyin: Either way, I don't think this thread necessarily block the PR. Could you take a look at 31e693f? If we agree this is the way to go, I can revert the previous two commits and have carlus unblocked.

atamazov:
@carlushuang Please merge from recent develop, assign urgency (and maybe other) labels, and assign reviewers. Thanks.

jerryyin:
@atamazov Technically this PR is in my plate right now because @carlushuang has done all he has to do. I'm taking over to address all layout related problems.

The conflict is less concerning to me. Once I reached on the same page with the right way to implement layout heuristic, I will do a clean up, merge from develop. Then give @carlushuang back control.

atamazov:
___Please merge from recent develop.___ Each push to this branch leads to waste of CI resources -- it builds obsolete dockers.

src/include/miopen/tensor.hpp
pfultz2: > pred() returns true if strides are equal, but with a not in front you will find those cases whose strides are different.

This is part of the grouping, its trying to find the next element that is not equal, so it can find the end to the subrange. This grouping is done since there is not a way to traverse just the keys. So instead it iterates over the elements `(x, a), (y, b), (y, c), (z, d)` and tries to group them into 3 subranges: `{(x, a)}, {(y, b), (y, c)}, {(z, d)}`. It passes the iterators to the `each` lambda for each subrange, where `last` is one past the end(just like calling `.end()` on STL containers). 

Of course this only works when duplicate keys appear together(which it will with `std::multimap`). That is why it is called `group_unique`(instead of `group_by`) as `std::unique` relies on the same precondition. To make it work for unordered containers then `std::partition` can be used instead(but `std::partition` wont work for associative containers). You can see how that is implemented in migraphx's [`group_by`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/algorithm.hpp#L11).

So once it has everything grouped together it tries to sort the values based on the delta of the previous in order to minimize the adjacent difference and direction in the permutation. So if `strides[1] == strides[3]` then we could have two valid permutations: `0,2,1,3` and `0,2,3,1`. So in this case, the delta of 2 with 3 or 1 is the same, but the direction is different. The permutation is increasing in value so `0,2,3,1` should be chosen instead.

> I think it doesn't try to check the calculated layout's validity like below:

No it doesn't but I think the sort comparator could be updated to take in account the lens so it can sort the group better, rather than just minimize adjacent difference. This is even simpler as we don't need to put it in a multimap or grouping:

```cpp
std::vector<int64_t> find_permutation(const std::vector<std::size_t>& lens, const std::vector<std::size_t>& strides)
{
    std::vector<std::int64_t> result(lens.size());
    std::iota(result.begin(), result.end(), 0);
    std::stable_sort(result.begin(), result.end(), by(std::greater<>{}, [&](auto x) {
                         return std::make_tuple(strides[x], lens[x]);
                     }));
    return result;
}
```

My initial tests show that this works, but it may need more testing or review to understand if this always produces a "valid" layout.

> I believe this is still essential to include to confirm that derived strides are consistent with lens.

But that wont work with non-packed tensors(which the tensor ops support), it should probably be an assert only when the tensor is packed.

pfultz2: So thinking about this, I believe sorting with lens is correct, because the 1 lens should be moved to the end. In general, lens that share the same strides should all be ones except for one of them. The stride with a non-1 dimension should come first.

I dont think there would be scenario where you would have more than one dimension be non-1 that have the same strides. If that is the case we could change the comparator for sort to do `return std::make_tuple(strides[x], std::max<std::size_t>(2, lens[x]));`, which makes the dimension either 1 or 2 when sorting, which will ensure that `std::stable_sort` will use the order of dimensions in the tensor for the non-1 dimensions. However, this may not even be a possible scenario.

jerryyin: Thanks a lot for the thorough explanation. I've got to say you are one step ahead of me in deriving this simple and elegant solution.

#### Adjacent difference solution
Counter example: 
lens = [16, 1, 3, 3]
strides = [1, 1, 48, 16]
The only valid layout is `HWNC`, however the adjacent difference solution will give a `HWCN`

#### Sorting lens && strides together

> The stride with a non-1 dimension should come first.

Excellent. Once this requirement get clarified, you solution makes total sense. The part that I like this solution in that even for the cases where we got multiple 1s, we are guaranteed to give back the valid solution that always start with the non-1 dimension. `[1, 1, 1, 1]` strides actually give back the perfect `NCHW` result.

jerryyin:
@carlushuang With latest commits I believe my job related with ambiguous layout is done in this PR. Sorry for flooding the PR with not the most relevant information to you. Could you take over the ownership of this PR from this point on (including formatting changes or reviewers' request on my commits)? If you find any further problem related with layouts please let me know. Thank you!

carlushuang:
@jerryyin OK, thanks for your work in this PR, let me take over the left part and make this PR ready

carlushuang:
Merged https://github.com/ROCmSoftwarePlatform/MIOpen/pull/806 https://github.com/ROCmSoftwarePlatform/MIOpen/pull/808

carlushuang:
@jerryyin @atamazov @pfultz2 please help review this PR and if anything further need changed.

carlushuang:
merged https://github.com/ROCmSoftwarePlatform/MIOpen/pull/814

src/conv/problem_description.cpp
```diff
@@ -59,6 +60,41 @@ std::ostream& operator<<(std::ostream& stream, std::function<void(std::ostream&)
     return stream;
 }
 
+void ProblemDescription::HeuristicUpdateLayouts()
+{
+    std::string labels = tensor_layout_get_default(in_layout.size());
```
atamazov, line 65: `const`, `default_labels`?

src/conv/problem_description.cpp
```diff
@@ -59,6 +60,41 @@ std::ostream& operator<<(std::ostream& stream, std::function<void(std::ostream&)
...
+            input_layouts.insert(layout);
+        if(out.IsLayout(labels, layout))
+            output_layouts.insert(layout);
+        if(weights.IsLayout(labels, layout))
+            weight_layouts.insert(layout);
+
+        // If we have found consistent layout among 3 tensors, assign them
+        if((input_layouts.count(layout) == 1) && (weight_layouts.count(layout) == 1) &&
+           (output_layouts.count(layout) == 1))
+        {
```
atamazov, line 88: Looks strange. How to you define "consistency" between three layouts?

src/conv/problem_description.cpp
```diff
@@ -59,6 +60,41 @@ std::ostream& operator<<(std::ostream& stream, std::function<void(std::ostream&)
...
+    std::set<std::string> weight_layouts;
+
+    std::vector<std::string> supported_layouts = {"NCHW", "NHWC", "NCDHW"};
+    for(const std::string& layout : supported_layouts)
+    {
+        // Skip layouts that doesn't match dimension sizes
+        if(layout.size() != labels.size())
+            continue;
+
+        if(in.IsLayout(labels, layout))
```
atamazov, line 78: `IsLayout` -> `IsLayoutPossible`  or `IsPossibleLayout`?

src/convolution.cpp
```diff
@@ -255,10 +258,25 @@ TensorDescriptor ConvolutionDescriptor::GetForwardOutputTensor(const TensorDescr
     out_lens[0] = in_n;
     out_lens[1] = out_c;
 
+    std::string default_layout = tensor_layout_get_default(xDesc.GetSize());
```
atamazov, line 261: const

src/convolution.cpp
```diff
@@ -140,9 +141,11 @@ const std::vector<int>& ConvolutionDescriptor::GetTransposeConvPads() const
 
 int ConvolutionDescriptor::GetGroupCount() const { return group_count; }
 
-TensorDescriptor ConvolutionDescriptor::GetForwardOutputTensor(const TensorDescriptor& xDesc,
-                                                               const TensorDescriptor& wDesc,
-                                                               miopenDataType_t yType) const
+TensorDescriptor
+ConvolutionDescriptor::GetForwardOutputTensorWithLayout(const TensorDescriptor& xDesc,
+                                                        const TensorDescriptor& wDesc,
+                                                        std::string yLayout,
```
atamazov, line 147: `const std::string&`

src/convolution.cpp
```diff
@@ -255,10 +258,25 @@ TensorDescriptor ConvolutionDescriptor::GetForwardOutputTensor(const TensorDescr
...
+                            out_strides);
+}
+
+TensorDescriptor ConvolutionDescriptor::GetForwardOutputTensor(const TensorDescriptor& xDesc,
+                                                               const TensorDescriptor& wDesc,
+                                                               miopenDataType_t yType) const
+{
+    // output layout same as input
+    std::string default_layout = tensor_layout_get_default(xDesc.GetSize());
+    std::string in_layout      = xDesc.GetLayout(default_layout);
```
atamazov, line 278: `const` for both

src/include/miopen/tensor_layout.hpp
```diff
@@ -0,0 +1,84 @@
...
+ *******************************************************************************/
+#ifndef GUARD_TENSOR_LAYOUT_HPP
+#define GUARD_TENSOR_LAYOUT_HPP
+
+#include <map>
+#include <algorithm>
+#include <vector>
+#include <string>
+#include <iterator>
+#include <miopen/errors.hpp>
```
atamazov, line 34: [style] Please move internal header before standard ones.

src/pooling.cpp
```diff
@@ -216,7 +217,13 @@ TensorDescriptor PoolingDescriptor::GetForwardOutputTensor(const TensorDescripto
 {
     std::vector<int> out_dim(xDesc.GetSize());
     GetForwardOutputDimNd(xDesc, xDesc.GetSize(), out_dim.data());
-    return TensorDescriptor(xDesc.GetType(), out_dim);
+
+    std::string default_layout = tensor_layout_get_default(xDesc.GetSize());
+    std::string in_layout      = xDesc.GetLayout(default_layout);
```
atamazov, line 222: both are `const`

test/conv_common.hpp
```diff
@@ -178,11 +179,18 @@ struct conv_stats
 template <class T, class Tout = T>
 tensor<Tout> get_output_tensor(const miopen::ConvolutionDescriptor& filter,
                                const tensor<T>& input,
-                               const tensor<T>& weights)
+                               const tensor<T>& weights,
+                               std::string out_layout = "")
```
atamazov, line 183: As far as I see, the default is not used.
```suggestion
                               const tensor<T>& weights,
                               const std::string& out_layout)
```

test/conv_common.hpp
```diff
@@ -1646,6 +1660,36 @@ struct conv_driver : test_driver
 
     void run()
     {
+        if(input.desc.GetSize() != in_layout.size() ||
+           weights.desc.GetSize() != fil_layout.size() || input.desc.GetSize() != out_layout.size())
+        {
+            MIOPEN_LOG_E("layout not match dimension size!");
```
atamazov, line 1666: `MIOPEN_LOG*` macros should not be used in tests.

atamazov, line 1666: This looks like a test error, shall we `std::cerr << "... FAILED ... " ... ` and return?

test/conv_common.hpp
```diff
@@ -1646,6 +1660,36 @@ struct conv_driver : test_driver
...
+        }
+
+        // replace desc to set layout information here
+        if(in_layout != "NCHW" || in_layout != "NCDHW")
```
atamazov, line 1670: Please explain why this is required.

test/conv_common.hpp
```diff
@@ -1646,6 +1660,36 @@ struct conv_driver : test_driver
...
+        {
+            std::vector<std::size_t> dim_lens = input.desc.GetLengths();
```
atamazov, line 1672: const

test/conv_common.hpp
```diff
@@ -1646,6 +1660,36 @@ struct conv_driver : test_driver
...
+            miopen::tensor_layout_to_strides(
+                dim_lens,
+                miopen::tensor_layout_get_default(input.desc.GetSize()),
+                in_layout,
+                dim_strides);
+            input.desc = miopen::TensorDescriptor(miopen_type<T>{}, dim_lens, dim_strides);
+        }
+        if(fil_layout != "NCHW" || fil_layout != "NCDHW")
+        {
+            std::vector<std::size_t> dim_lens = weights.desc.GetLengths();
```
atamazov, line 1683: const

atamazov (changes requested):

driver/tensor_driver.hpp
```diff
@@ -125,14 +125,14 @@ int SetTensorNd(miopenTensorDescriptor_t t,
     }
 
     // Dimension lengths vector 'len' comes with a default layout.
-    std::string len_layout = tensor_layout_get_default(layout.size());
+    std::string len_layout = miopen::tensor_layout_get_default(layout.size());
```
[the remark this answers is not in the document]
carlushuang: I guess if change to enum it will go back to the starting point of previous discussion of support multiple layout like here https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343#discussion_r474778988. I mean, currently all implementation is based on stride, and then permute/shuffle the layout string then return the final string interpretation.

src/conv/problem_description.cpp
```diff
@@ -59,6 +60,41 @@ std::ostream& operator<<(std::ostream& stream, std::function<void(std::ostream&)
...
 
+void ProblemDescription::HeuristicUpdateLayouts()
+{
+    std::string labels = tensor_layout_get_default(in_layout.size());
+
+    std::set<std::string> input_layouts;
+    std::set<std::string> output_layouts;
+    std::set<std::string> weight_layouts;
+
+    std::vector<std::string> supported_layouts = {"NCHW", "NHWC", "NCDHW"};
```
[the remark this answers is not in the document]
carlushuang: OK

src/conv/problem_description.cpp
carlushuang: OK, add a `const`

src/conv/problem_description.cpp
carlushuang: `input_layouts`/`weight_layouts`/`output_layouts` is `std::set`, and we check the layout is in that set by `count()` method, so if the `layout` is in all of them, the count will return `1`

src/conv/problem_description.cpp
carlushuang: [`IsLayout` ](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/765/files#diff-2a2c874ce9a2ebeea7603f4faa717453773c0ff152197455aa4ef0806df794ddR179) try to verify a exact match of the `layout` with `labels`, so I think the name here may be suitable

src/convolution.cpp
carlushuang: OK

src/convolution.cpp
carlushuang: OK

src/convolution.cpp
carlushuang: OK

src/include/miopen/tensor_layout.hpp
carlushuang: OK

src/pooling.cpp
carlushuang: OK

test/conv_common.hpp
carlushuang: Indeed this function is used in multiple files like in [cba_inference.cpp](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/test/cba_inference.cpp#L86). since fusion kernel currently have no kernel support different layout, which lead to no suitable test or validation, so I just leave it cba/cbna unchanged and give this function a default value of `out_layout`

test/conv_common.hpp
carlushuang: OK

carlushuang: OK, change to std::cerr...

test/conv_common.hpp
carlushuang: the code under `if` is to re-construct `input.desc`, `input.desc` is first constructed by its constructor [here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/test/conv2d.cpp#L33), and the `tensor` constructor seems very hard to support different layout only from command line.

test/conv_common.hpp
carlushuang: OK

test/conv_common.hpp
carlushuang: OK

driver/tensor_driver.hpp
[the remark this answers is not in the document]
jerryyin: Agreed with @carlushuang switching to enum everywhere will make us lose the ability to execute arbitrary layout (Unless we want to specify 4 * 3 * 2 *1 = 24 layouts for 2d tensors and even more for 3d tensors.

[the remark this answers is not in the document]
atamazov: What about format like this:
```C+
enum layout : uint32_t {
    NCHW  = 0x00010203ULL,
    NHWC  = 0x00020301ULL,
    NCDHW = 0x0001020304ULL,
};
```
and so on. It is functionally equivalent to strings. You can shuffle bytes like symbols in chars. Supports up to 8 dims. You can define only required enum members. And comparison is fast.

src/conv/problem_description.cpp
atamazov: Can a tensor have more than one layout?

src/conv/problem_description.cpp
```diff
@@ -59,6 +60,41 @@ std::ostream& operator<<(std::ostream& stream, std::function<void(std::ostream&)
...
+        if(in.IsLayout(labels, layout))
+            input_layouts.insert(layout);
+        if(out.IsLayout(labels, layout))
+            output_layouts.insert(layout);
+        if(weights.IsLayout(labels, layout))
+            weight_layouts.insert(layout);
+
+        // If we have found consistent layout among 3 tensors, assign them
+        if((input_layouts.count(layout) == 1) && (weight_layouts.count(layout) == 1) &&
+           (output_layouts.count(layout) == 1))
```
atamazov, line 87:
```suggestion
        if(in.IsLayout(labels, layout) && out.IsLayout(labels, layout) && weights.IsLayout(labels, layout))
```
and remove 
```
    std::set<std::string> input_layouts;
    std::set<std::string> output_layouts;
    std::set<std::string> weight_layouts;
```

atamazov (changes requested):

test/conv_common.hpp
atamazov: No, that files use definition from `fusionHost.hpp`.

driver/tensor_driver.hpp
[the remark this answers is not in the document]
pfultz2: I dont think the strings here are substantially slower because they are fairly small(no heap allocations, and single word comparisons). What might be slower is doing `layout == "NCHW"` since it will convert `NCHW` to `std::string` and will call `strlen` on the string. This could be mitigated by making a string const(ie `const std::string NCHW = "NCHW"`). and doing `layout == NCHW` instead.

Alternatively, we could look into packing these into a 4 byte integer(with 4 bits for each permutation axis), but I don't think that would be faster(and would be a more complicated implementation). We tried to do similar optimizations in the past with network_config string by replacing it with integers. The integer implementation surprisingly wasn't faster than string concat+integer-to-string conversions. So I dont see this as a worthwhile optimization to pursue.

src/conv/problem_description.cpp
carlushuang: @jerryyin what do you think about this change?

src/conv/problem_description.cpp
carlushuang: No, but here is just a simple check and I think the name `IsLayout` do meet its purpose.

driver/tensor_driver.hpp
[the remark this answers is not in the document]
atamazov: I am in doubt too. I am not going to insist unless you think it's worth trying this.

Btw have you noted that enum members hold permutation indices in individual bytes? This makes possible additional optimization, right?

> This could be mitigated by making a string const(ie `const std::string NCHW = "NCHW"`). and doing `layout == NCHW` instead.

Not fully, I guess, because string comparison needs to compare string lengths anyway (or to compare up to the ending `'\0'`, which means reading the 2nd word from memory).

> Alternatively, we could look into packing these into a 4 byte integer(with 4 bits for each permutation axis), but I don't think that would be faster

Yes, we need shifts to access nibbles ;(

src/conv/problem_description.cpp
jerryyin: @carlushuang Thanks for ping me. Yep, I think @atamazov 's suggestion is good. I don't think storing all possible layout in the set is necessary. It sets the first consistent layout according to `supported_layouts` and be done.

driver/tensor_driver.hpp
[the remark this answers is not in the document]
pfultz2: > What about format like this:

NCDHW overflows the 32-bit integer in your snippet.

> Not fully, I guess, because string comparison needs to compare string lengths anyway (or to compare up to the ending '\0', which means reading the 2nd word from memory).

The lengths are already stored in `std::string` so a call to `strlen` is not necessary(only when constructing from a string literal). Since the sizes are known, the comparison then can be done with a `memcmp` which is pretty fast(and its equivalent to comparing 32-bit integers for 4-byte strings).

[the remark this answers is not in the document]
atamazov: > NCDHW overflows the 32-bit integer in your snippet.

That is why `uint64_t` should be used (and ULL suffix). Typo fixed))

src/conv/problem_description.cpp
atamazov: No? I thought that a tensor may match a number of layouts here. Am I wrong?

driver/tensor_driver.hpp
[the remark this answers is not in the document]
atamazov: Ok, let's postpone this idea for now.

src/conv/problem_description.cpp
carlushuang: OK, I'll change based on this review

src/conv/problem_description.cpp
carlushuang: Oh I think I got your idea. you mean here is just to check that this `TensorDescriptor` can be interpreted as the `labels` right? And this function's use scenario is for such a check, and maybe used multiple times.
Hi @jerryyin let me double confirm with you this function is designed for this purpose right?

jerryyin: > Can a tensor have more than one layout?

@atamazov Correct, one tensor, when ambiguous, can have more than one layout. Here, it is checking for if a possible layout is coherent with its strides.

@carlushuang That's right, `IsPossibleLayout` is a better name, thanks.

carlushuang: @jerryyin  OK, I'll change to this name

carlushuang:
Hi @atamazov please help check the review fix again. I've changed all the code according to review.

test/conv_common.hpp
atamazov: This is not resolved.

test/conv_common.hpp
atamazov: Now I see. Could you please add this reasoning in C++ comments.

carlushuang: OK, done in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/765/commits/ee0302f0bd4ffe149ab55fbb80437773ab7cf148

test/conv_common.hpp
carlushuang: OK, fixed in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/765/commits/ee0302f0bd4ffe149ab55fbb80437773ab7cf148

atamazov (approved):
LGTM

atamazov:
@DrizztDoUrden This looks good, and I am going to merge it now. But please have a look. Perhaps you'll have some ideas how to improve of refactor this.

DrizztDoUrden (approved):

src/include/miopen/tensor_layout.hpp
```diff
@@ -0,0 +1,84 @@
...
+                   [&layout, &dim_to_len](char cur_layout_char) {
+                       auto pos = layout.find(cur_layout_char);
+                       if(pos == std::string::npos)
+                       {
+                           MIOPEN_THROW(std::string("mismatched layout string, unexpect char: ")
+                                            .append(1, cur_layout_char));
+                       }
+                       return std::accumulate(layout.begin() + pos + 1,
+                                              layout.end(),
+                                              1,
```
atamazov, line 66: 🐛 1 is of type int and therefore std::accumulate uses int as an accumulator, even if T is size_t. I am going to fix this soon.

PR #766, opened 2021-03-03
[CI] Select image names and docker build args to get rid of useless image rebuilds. [tests][FP16][MI100] Remove useless tests.

atamazov:
- [Optimization] Even if stage is intended to run only on Vega10 node, build docker image as if it is intended to run on any Vega GPU.
  - Currently things like `gpu_arch: gfx900` are used with "standard" image names. This leads to many useless image rebuilds.
- [Optimization] Same for Vega20 stages
- [Optimization] [Fix] Use `miopen-hip-clang` as a basename for hip-clang based docker image.
  -  Note: Currently Jenkinsfile uses `miopen` which is actually intended for hcc-based dockers. This issue was introduced in #759
- [Optimization] Use the same image name for `HipNoGPU` -- it actually uses the same image
- [Optimization] Use different images names for MIOpenTensile "latest" docker image
  - Currently we use "standard" name and that leads to useless image rebuilds per each job.
- [Optimization] Use distinct image names for "Packages" as docker build options are different.
- Some refactors.

All of the above optimizations can be considered as fixes for issues that reduce docker cache efficiency.

The expected side effect is ~2X better throughoutput of our CI.

Also skips useless FP16 tests from MI100 targets. Why: If some test is not run on MI100 with FP32 (float), then all the more it should not be run with FP16 (half). But currently this is not so. For example, `test_dropout` takes about of 2 hours to complete on MI100 with FP16 datatype, thus wasting ~2 hours of total CI job time.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -211,7 +211,7 @@ pipeline {
         parallelsAlwaysFailFast()
     }
     environment{
-        image = "miopen"
+        image = "miopen-hip-clang"
     }
     stages{
         // Run all static analysis tests
@@ -225,8 +225,8 @@ pipeline {
                     steps{
                         script{
                             try{
-                                buildJob('hcc', flags: '-DCMAKE_BUILD_TYPE=release', cmd: cmd, gpu_arch: "all")
-                            } 
+                                buildJob('hcc', flags: '-DCMAKE_BUILD_TYPE=release', cmd: cmd, gpu_arch: "all", image: 'miopen')
+                            }
                             catch(e){
                                 echo "throwing error exception for the stage"
                                 echo 'Exception occurred: ' + e.toString()
@@ -252,8 +252,8 @@ pipeline {
                     steps{
                         script{
                             try{
-                                buildJob('hcc', flags: '-DCMAKE_BUILD_TYPE=release', cmd: cmd, gpu_arch: "all")
-                            } 
+                                buildJob('hcc', flags: '-DCMAKE_BUILD_TYPE=release', cmd: cmd, gpu_arch: "all", image: 'miopen')
+                            }
                             catch(e){
                                 echo "throwing error exception for the stage"
                                 echo 'Exception occurred: ' + e.toString()
@@ -271,8 +271,8 @@ pipeline {
                     steps{
                         script{
                             try{
-                                buildJob('hcc', flags: '-DCMAKE_BUILD_TYPE=release', cmd: cmd, gpu_arch: "all")
-                            } 
+                                buildJob('hcc', flags: '-DCMAKE_BUILD_TYPE=release', cmd: cmd, gpu_arch: "all", image: 'miopen')
+                            }
                             catch(e){
                                 echo "throwing error exception for the stage"
                                 echo 'Exception occurred: ' + e.toString()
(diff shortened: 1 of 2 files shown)


shurale-nkn (dismissed):
LGTM

Jenkinsfile
```diff
@@ -59,7 +59,7 @@ def buildJob(Map conf, compiler){
         def env4make = conf.get("env4make", "")
         def image = conf.get("image", "miopen")
         def cmd = conf.get("cmd", "")
-        def gpu_arch = conf.get("gpu_arch", "all")
+        def gpu_arch = conf.get("gpu_arch", "gfx900;gfx906")
```
pfultz2, line 62: I believe this flag is for MIOpenTensile, which will still treat this as `all`. This is because tensile only allows setting one arch or all archs. So when setting the archs to `gfx900;gfx906` internally it will set the tensile arch to "all":

https://github.com/ROCmSoftwarePlatform/MIOpenTensile/blob/develop/CMakeLists.txt#L37

atamazov: Thanks.

atamazov: I think we can tolerate that provided that dockers images will not be rebuilt too often. However it would be nice to save some time during build of Tensile, in the near future...

atamazov: @pfultz2 btw in `hip-clang.docker` we are passing `-DAMDGPU_TARGETS=${GPU_ARCH}` to Tensile build, so `Tensile_ARCHITECTURE`  [here](https://github.com/ROCmSoftwarePlatform/MIOpenTensile/blob/develop/CMakeLists.txt#L37) should take its value, right?

atamazov: [Side note] I think that `if(AMDGPU_TARGETS MATCHES ";")` takes true path only if `AMDGPU_TARGETS` equal to `";"`. When it contains `gfx900;gfx906`, `'if(...)` takes false path, right? -- https://cmake.org/cmake/help/latest/command/if.html#matches

pfultz2: > I think we can tolerate that provided that dockers images will not be rebuilt too often. However it would be nice to save some time during build of Tensile, in the near future...

Yea, I think we need to pull down prebuilt binaries. Or we could have a jenkins job that builds the binaries if they haven't been built already, and then we can archive it on jenkins. 

> btw in hip-clang.docker we are passing -DAMDGPU_TARGETS=${GPU_ARCH} to Tensile build, so Tensile_ARCHITECTURE here should take its value, right?

Yes, but if it is `gfx900;gfx906` then it will still build for `all`.

>  I think that if(AMDGPU_TARGETS MATCHES ";") takes true path only if AMDGPU_TARGETS equal to ";". if("gfx900;gfx906" MATCHES ";") takes false path, isn't it?

No, this is a regex match. If you want it to be true for only `;` then you would do `if(AMDGPU_TARGETS STREQUAL ";")`.

pfultz2 (approved):

Jenkinsfile
pfultz2: Of course, I dont think this change is a blocker, so I approved it already.

atamazov: > No, this is a regex match. If you want it to be true for only `;` then you would do `if(AMDGPU_TARGETS STREQUAL ";")`.

Aha, it works like `echo "a;b" | grep ";"`. Thanks for reviewing!

shurale-nkn (approved):
LGTM

PR #768, opened 2021-03-03
[tests][FP16][MI100] Remove useless tests

atamazov:
Is some test is not run on MI100 with FP32 (float), then all the more it should not be run with FP16 (half). For example, `test_dropout` -- it takes about of 2 hours to complete on MI100 with FP16 datatype, thus wasting ~2 hours of total CI job time.


atamazov:
Moved to #766. 

Why: This accelerates CI testing process because dockerfile images already built and available on CI nodes.

PR #769, opened 2021-03-03
[UTILITY] Enable CI testing of develop branch

atamazov:
This is utility PR that enables CI testing for the `develop` branch. Current Jenkins config discover _only branches that filed as PRs_, so `develop` is not being tested.


atamazov:
Jenkins switched to discovering all branches.

Issue #755, opened 2021-02-25
Hanging fused Winograd kernels on gfx1030


zjing14:
Got GPU hangs on gfx1030 with fused Winograd kernels

```
./bin/test_cba_inference --float --input 1, 64, 14, 14 --weights 1, 64, 3, 3 --pads_strides_dilations 0 0 1 1 1 1 --alpha 0.5 --beta 0.5 --gamma 0.5 --bmode 1 --pmode default --test_activ 1 --amode 3
MIOpen(HIP): Info2 [GetKernels] 1 kernels for key: miopenConvolutionWinogradBiasActiv "1, 64, 14, 14FP321, 1, 12, 12FP3264x14x14x3x3x1x12x12x1xNCHWxFP32x0x0x1x1x1x1x1xFbiasOnActivFwd3"
MIOpen(HIP): Info [Execute] miopenConvolutionWinogradBiasActiv,1, 64, 14, 14FP321, 1, 12, 12FP3264x14x14x3x3x1x12x12x1xNCHWxFP32x0x0x1x1x1x1x1xFbiasOnActivFwd3
```

Kirpich30000:
@zjing14 could you try #740? Does it resolve the issue?

zjing14:
Yes. I tested with #740.

atamazov:
@Kirpich30000 how it could be so? I thought #740 is NFC.

Kirpich30000:
The first guess is that it might be dispatched with a wrong ndrange or arguments. Does the fusion reuse existing solver or it has it's own machinery for dispatching the kernel?

atamazov:
It uses own machinery.

atamazov:
#740 merged to develop.

zjing14:
@atamazov Do you mean we should drop the issue currently?

Kirpich30000:
@zjing14 am I correct, that the issue is still observed even with #740 merged in? We might get confused with `Yes` answering `Does it resolve the issue?` and not `could you try #740?`
If so, does any changes are needed to enable fused Winograd on gfx10?

zjing14:
@Kirpich30000 Sorry for the confusion. The hanging issue still exists with #740 merged. 

To enable fused Winograd on gfx10, we need to add gfx1030 into supported arch lists.

[...]

Issue #763, opened 2021-03-02
Test_mdgraph fails on gfx1030 due to unsupported fused kernels


zjing14:
- test_mdgraph fails with unsupported fused kernels for gfx1030 - miopenConvolutionWinogradBiasActiv and miopenConvolutionDirectBiasActivAsm

```
./bin/test_mdgraph --all
FAILED: krn_name == "miopenSp3AsmConvRxSU_CBA": /dockerx/MIOpen/test/mdgraph.cpp: 118
Aborted (core dumped)
```

atamazov:
@zjing14 Well, I would generalize the description of the issue. `test_mdgraph` is a unit test that should be aware of the implementation. Specifically, it must "know" which fused algorithms _and_ kernel names are supported on which GPUs. And that knowledge should be used in the test to avoid testing of unsupported algorithms/kernels. Do you agree?

zjing14:
@atamazov But the logic of test_mdgraph is to make sure fused algorithms supported on all GPUs.

zjing14:
@atamazov Let us refactor the test_mdgraph as you suggested in a separate PR.

atamazov:
Test refactored in #813

PR #772, opened 2021-03-04
Adding Winograd v21 fused

Slimakanzer:
This PR adds support for winograd v21_1_2 kernels into fusion. Resolves issue #755.

Also fixes two related bugs:
1. `supported_arch` were not checked at **fusion** compilation. Because of what fusion could launch unsupported kernels.
2. `supported_arch` for _winograd v9_2_7_ relu operations was not specified in the _MDGraph_. For that reason kernel could run on a device that is not supported:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/b7630f76dad1926e224a23d5e470e44987d84fc3/src/md_graph.cpp#L579-L584

## Summary of code changes
* Check device compatibility and kernel when **fusion** Compile
* Added `ConvBinWinogradRxSf2x3g1Fused` solver 
* Added tensor/op parameters required for the _v21_1_2_ winograd version
* Added _winograd v21_1_2_ kernel to the Fusion MDGraph

## Local verification
* gfx1030, rocm 3.10
* gfx906, rocm 4.0
* gfx803, rocm 3.10
* OCL + HIP backend


atamazov:
This does not resolve #763 because that issue is about a bug in the unit test. I will update PR description.

PS Please avoid using automatic closing of issues.

src/fusion.cpp
```diff
@@ -263,6 +270,22 @@ bool ConvForwardOpDescriptor::GetOpAttr(const std::string& sym, int& val) const
     {
         val = base_desc.GetGroupCount();
     }
+    else if(sym == "f_byte_stride_nk")
+    {
+        val = f_strides[0];
```
Kirpich30000, line 275: What about bwd/wrw? Will it work correctly?

Slimakanzer: Fusion API introduces only forward operations like:
* Convolution Forward
* Activation Forward
* Bias Forward

This is explicitly defined in Winograd constraints:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/8c172d466d25146656ae0a53f03e82ab3cb76d30/src/md_graph.cpp#L521-L524

atamazov:
Please merge from develop (see mail)

atamazov:
@Kirpich30000 Is it ready for merge?

atamazov:
@Slimakanzer Do we still need ConvBinWinogradRxSFused Solver? Related .s files?

src/md_graph.cpp
```diff
@@ -476,215 +476,255 @@ static std::vector<DefaultKernelArg> WinogradNodeArgs()
...
+                FusionMDGraph_Edge_Map map_wino_conv;
+                map_wino_conv["constraints"] = {stride_constr};
+                add_meta_wino(map_wino_conv, 5);
+                g.AddEdge(nullptr, vc, map_wino_conv);
+
+                // add 3x3 with higher priority since its the fastest case
+                FusionMDGraph_Edge_Map map_wino_conv_xe3;
+                map_wino_conv_xe3["constraints"] = {stride_constr, "(y == 3) & (x == 3)"};
+                add_meta_wino(map_wino_conv_xe3, 100);
+                g.AddEdge(nullptr, vc, map_wino_conv_xe3);
```
Kirpich30000, line 719: FYI winograd might not be the best for some configurations even if the filter size is 3x3

src/ocl/fusionopbiasbnactivocl.cpp
```diff
@@ -8,7 +8,8 @@ namespace fusion {
 bool IsWinograd(const std::vector<solver::AnySolver>& ss)
 {
     assert(ss.size() == 1);
-    return ss[0].GetSolverDbId() == "ConvBinWinogradRxSFused";
+    auto solverId = ss[0].GetSolverDbId();
+    return (solverId == "ConvBinWinogradRxSFused" || solverId == "ConvBinWinogradRxSf2x3g1Fused");
```
Kirpich30000, line 12: Relying on kernel name looks too hacky. @atamazov what is your opinion?

src/solver/conv_winoRxS_f2x3.cpp
```diff
@@ -867,5 +867,42 @@ ConvSolution ConvBinWinogradRxSf2x3g1::GetSolution(const ConvolutionContext& par
     return tunable.GetSolution(params, tunable.GetPerformanceConfig(params), false);
 }
 
+bool ConvBinWinogradRxSf2x3g1Fused::IsApplicable(const ConvolutionContext&) const
+{
+    return true; // Actual checks moved to FusionMDGraph.
```
Kirpich30000, line 872: This one looks suspicious. Might it break some usecases?

src/ocl/fusionopbiasbnactivocl.cpp
atamazov: Inevitable for now.

src/md_graph.cpp
atamazov: You mean v21 Winograd could be slower than older fused Winograd?

src/solver/conv_winoRxS_f2x3.cpp
atamazov: If implementation in MDGraph is correct, then no.

Slimakanzer:
> @Slimakanzer Do we still need ConvBinWinogradRxSFused Solver? Related .s files?

@atamazov `ConvBinWinogradRxSFused` provides _gfx8_ compatible kernels, however the previous _gfx9_ winograd kernels are no longer needed (removed in da35c78). Unfortunately, the _v21_ and _v9_2_7_ metadata cannot be combined because they have a different set of arguments.

Kirpich30000 (approved):
LGTM

atamazov (approved):
Ok

PR #777, opened 2021-03-06
[WORKAROUND][HOTFIX][ROCm3.7] Fix memory access fault in MIOpenGEMM

atamazov:
Fix for the following issue:
GPU memory access fault with Vega20, ROCm 3.7, OpenCL backend. Failing config:
```
./bin/test_conv2d --float --cmode conv --pmode default --group-count 4 \
--input 4, 4, 161, 700 --weights 32, 1, 5, 20 --pads_strides_dilations 0 0 2 2 1 1 --trans_output_pads 0 0 \
--verbose --disable-verification-cache \
--disable-backward-data \
--disable-backward-weights
```
Occurs when MIOpenGEMM kernel is run. Failure may happen with any N. The suspected root reason is a bug in the OpenCL compiler. Not reproducible with ROCm 4.0 (same MIOpenGEMM).

The workaround applies to 3.7 only and does not affect MIOpen running with other ROCm versions.

The bug blocks our CI, stage "OpenCL Release All". Fix tested locally and on CI (8 times)


shurale-nkn (approved):
LGTM

atamazov:
Thank you, Kamil!

atamazov:
@DrizztDoUrden We need to apply this patch to the Bwd Gemm Solvers.

PR #779, opened 2021-03-07
[CI][HOTFIX][Jenkinsfile] Fix "Packages" stage for hip-clang docker

atamazov:
When Tensile is being built in a hip-clang docker, `GPU_ARCH=all` is not accepted.

711f081 is only for testing and will be reverted.


shurale-nkn (approved):
/\

atamazov:
:pray:

PR #790, opened 2021-03-09
Change lock location for ufd file update

cderb:
Allows for dockers launched by tuna to properly communicate file usage.


atamazov:
@cderb Please also remove branch (if you do not need it anymore)

PR #138, opened 2020-04-07
Remove Co-located lock files

JehandadKhan:
This PR removes co-located files which are causing issue with permission, and restores the old behavior.

The rationale for this change was to share the directory with multiple dockers running MIOpen. However, we see many lock file creation failures in logs, furthermore if the user config directory is not present the lock file creation fails. 

Since, we are getting rid of this mechanism for SQLite pdb in a subsequent PR, it doesn't make sense to keep this and create issues.

PR #791, opened 2021-03-09
move lock file location

cderb:
Allows for dockers launched by tuna to properly communicate file usage.

### DIFF
--- a/src/db.cpp
+++ b/src/db.cpp
@@ -66,7 +66,8 @@ PlainTextDb::PlainTextDb(const std::string& filename_,
 
 PlainTextDb::PlainTextDb(const std::string& filename_, bool is_system)
     : filename(filename_),
-      lock_file(LockFile::Get(LockFilePath(filename_).c_str())),
+      lock_file(LockFile::Get((is_system) ? LockFilePath(filename_).c_str()
+                                          : std::string(filename_ + ".lock").c_str())),
       warn_if_unreadable(is_system)
 {
     if(!is_system)


atamazov:
@cderb 
> Allows for dockers launched by tuna to properly communicate file usage.

Could you please explain what is going wrong in that case? Thanks.

atamazov (dismissed):
Unanswered Q at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/791#issuecomment-794353191

cderb:
We had an issue where the previous lock file location was outside of the bounds of the docker file system that we were using when launching MIOpen through Tuna. Tuna runs multiple processes which all try to update this file, and since they were all in dockers with separate file trees, they were not syncing correctly. 
This moves the lock to be within the user home directory which we map to the docker, solving the synchronization issue.

atamazov:
As far as I remember lockfiles reside in the same directory where the database that it protects. Moving lockfile will hide more general sync issue and database could be silently corrupt.

@DrizztDoUrden what do you think?

cderb:
@JehandadKhan This just changes the user database lock location?

JehandadKhan:
> As far as I remember lockfiles reside in the same directory where the database that it protects. Moving lockfile will hide more general sync issue and database could be silently corrupt.

@atamazov Currently, the lock files are in the temp directory and NOT with the protected file. This PR actually achieves what you describe ( for user Dbs). Please let me know if my understanding is faulty.

atamazov:
Yes, I was wrong, currently we create lockfiles in [$TMPDIR](https://www.boost.org/doc/libs/1_69_0/libs/filesystem/doc/reference.html#temp_directory_path)`/miopen-lockfiles`. Initially I wanted to put lockfiles near databases, but later we decided to use temp directory, and I do not remember rationale. I need to discuss this with @DrizztDoUrden.

atamazov:
Btw this story is quite old:
- https://github.com/AMDComputeLibraries/MLOpen/pull/2468
- https://github.com/AMDComputeLibraries/MLOpen/issues/2477

DrizztDoUrden (approved):
I am OK with moving user db locks to the home. If the user doesn't have access to the home, being unable to lock would be the least of theirs problems.

src/db.cpp
```diff
@@ -66,7 +66,8 @@ PlainTextDb::PlainTextDb(const std::string& filename_,
 
 PlainTextDb::PlainTextDb(const std::string& filename_, bool is_system)
     : filename(filename_),
-      lock_file(LockFile::Get(LockFilePath(filename_).c_str())),
+      lock_file(LockFile::Get((is_system) ? LockFilePath(filename_).c_str()
+                                          : std::string(filename_ + ".lock").c_str())),
```
DrizztDoUrden, line 70: Although I would advise changing LockFilePath so it would receive is_system and make the actual path.

atamazov:
@cderb Please let us know if you are going to make changes as recommended at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/791#discussion_r594365529.

atamazov:
This doesn't work as expected: 
image

atamazov:
It seems like we already had this feature, but it was reverted by #138.

atamazov (changes requested):
CI failures, https://github.com/ROCmSoftwarePlatform/MIOpen/pull/791#issuecomment-805260046

PR #793, opened 2021-03-10
[CI] Optimize out unnecessary image rebuilds that affect Static checks. Add explicit notion of "default image name".

atamazov:
This ~should~ could save another couple of hours for each CI job.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,6 +15,10 @@ def rocmnode(name) {
     return node_name
 }
 
+def default_image_name() {
+    return 'miopen-hip-clang'
+}
+
 def show_node_info() {
     sh """
         echo "NODE_NAME = \$NODE_NAME"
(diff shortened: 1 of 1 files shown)


shurale-nkn:
This change will not save time, because it will not affect current cache state.

Each active  command line in dockerfile is a new commit, and docker saves each created commit. For caches docker uses command line output and if it is changes docker saves new result with new cache.
 So cache is independent from image name or Dockerfile name, it depends only on previous commits and current command output.

shurale-nkn:
> Are you saying the docker image name is not important? What I see is that if docker image name is the same for different docker.build() commands, it always rebuilds!

Yes, image name is not important. Build process starts in any case, but if we have cached results it will be faster.

pfultz2:
> You can build one image with name1 and than build image with name2 and 2-nd build will use cached results.
>
> Also afther you can change some lines in docker file build image name1 again with new information and then revert Dockerfile changes. And the last build will use previously cached commits from building original Dockerfile.

This is correct, see my comment [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/676#discussion_r591714443) for a longer explanation. First, docker layers are not deleted when tagging an image name. Secondly, unused docker layers are only deleted when explicitly calling [`docker prune`](https://docs.docker.com/config/pruning/):

> Docker takes a conservative approach to cleaning up unused objects (often referred to as “garbage collection”), such as images, containers, volumes, and networks: these objects are generally not removed unless you explicitly ask Docker to do so.

shurale-nkn (approved):
LGTM, performance is not affected.

atamazov:
@shurale-nkn @pfultz2 Thanks for not blocking this.

atamazov:
@pfultz2 @shurale-nkn @ce1adon I did a quick test and can confirm that docker image names in Jenkinsfile are not important for CI performance. Eventually I will create a PR that removes all that stuff.

PR #795, opened 2021-03-10
Additions to the HIPNOGPU backend 

JehandadKhan:
This PR adds the capability for the NOGPU backend to compile kernels and create programs which can be extracted using Tuna.


atamazov:
Quite a lot of copying code from HIP backend, and we need to support and maintain this copy. Can we use different approach? I mean, using existing HIP backend code but exclude all HIP-runtime related calls by
```C++
if !MIOPEN_MODE_NOGPU
...
endif
```
Conditional compilation is not good, but it is better than two partially overlapped copies.

src/nogpu/handle.cpp
```diff
@@ -99,63 +97,131 @@ void Handle::ReadTo(void* /* data */,
...
-                            std::string /* params */,
-                            bool /* is_kernel_str */,
-                            const std::string& /* kernel_src */) const
+Program Handle::LoadProgram(const std::string& program_name,
+                            std::string params,
+                            bool is_kernel_str,
+                            const std::string& kernel_src) const
 {
-    return {};
+    params += " -mcpu=" + this->GetTargetProperties().Name();
```
atamazov, line 169:
```C+++
    if(!miopen::EndsWith(program_name, ".mlir-cpp"))
    {
        params += " -mcpu=" + this->GetTargetProperties().Name();
    }
```
Otherwise the binary cache would miss MLIR kernels percompiled with HIPNOGPU backend.

atamazov (changes requested):

src/nogpu/handle.cpp
atamazov: Caught accidentally.

JehandadKhan:
> Quite a lot of copying code from HIP backend, and we need to support and maintain this copy. Can we use different approach? I mean, using existing HIP backend code but exclude all HIP-runtime related calls by
> 
> ```c++
> if !MIOPEN_MODE_NOGPU
> ...
> endif
> ```
> 
> Conditional compilation is not good, but it is better than two partially overlapped copies.

I am not happy about it either, but was not sure if wrapping each API call in a macro is better. Let me create another branch and test it out.

atamazov:
@JehandadKhan 
> I am not happy about it either, but was not sure if wrapping each API call in a macro is better. Let me create another branch and test it out.

Just to note that even if such a design may look clumsy, it is mush better protected from accidental errors like [this](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/795#discussion_r602467076).

atamazov (approved):
Let it go for solving urgent problems. However I hope to see this undone soon, together with adding a new design that would exclude code duplication.

atamazov:
@JehandadKhan Unfortunately, more problems are coming up (not here yet), please look at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/841#discussion_r615301929

PR #800, opened 2021-03-15
Backward data GEMM solvers and invokers

DrizztDoUrden:
Backward GEMM is now split between the following solvers:
- GemmBwd1x1_0_2 - applicable when 2d && kernel is 1x1 && conv pads are 0x0 && conv strides are 2x2
- GemmBwd1x1_0_1 - applicable when kernel is 1x1 && conv pads are 0x0 && conv strides are 1x1
- GemmBwdRest - applicable when none of the solvers above is.

"The rest" kind of the solver logic is flawed, but at the moment, it is implemented as it worked before. Changes to that may be applied in separate PR later.

I have updated solver ids in the find-db.


DrizztDoUrden:
To be honest I am not sure who to add as reviewers.

atamazov:
Please update from develop. Each PR needs to be checked with the most recent clang-tidy checks.

atamazov:
@DrizztDoUrden HIP build failed. Ditto Hip Clang Tidy check.

atamazov:
"Bf16 Hip Release gfx908" failed twice in a row. Please look at CI logs. In the past, it failed also at run 7, so the issue is not from the recent changes.

I will run my tests tonight (but this is gfx906).

DrizztDoUrden:
Sure, I'll try to fix it today, not much to do this weekend xD

atamazov:
@DrizztDoUrden gfx906, ROCm 4.0, memory access fault with BF16 when `-i` > 1:
```bash
$ ./bin/MIOpenDriver convbfp16 -x 1 -y 1 -W 14 -H 14 -c 1024 -n 8 -k 512 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 2 -V 0 -i 2
...
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] miopenConvolutionBwdDataAlgoGEMM   0.194634        1204224
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] miopenConvolutionBwdDataAlgoDirect 0.26096 0
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] BWD Chosen Algorithm: GemmBwd1x1_stride2 , 1204224, 0.194634
MIOpen(HIP): Info [ConvolutionBackwardData] algo = 0, workspace = 1204224
MIOpen(HIP): Info [ConvolutionBackwardData] algo = 0, workspace = 1204224
Memory access fault by GPU node-2 (Agent handle: 0x172c720) on address 0x7f627bd61000. Reason: Page not present or supervisor privilege.
Aborted (core dumped)
```
With `-i 1` it does not fail with GPU memory access fault, but **_validation_** fails:
```bash
$ ./bin/MIOpenDriver convbfp16 -x 1 -y 1 -W 14 -H 14 -c 1024 -n 8 -k 512 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 2 -V 0 -i 1 -V 1 
...
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] miopenConvolutionBwdDataAlgoGEMM   0.195769        1204224
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] miopenConvolutionBwdDataAlgoDirect 0.2624  0
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] BWD Chosen Algorithm: GemmBwd1x1_stride2 , 1204224, 0.195769
MIOpen(HIP): Info [ConvolutionBackwardData] algo = 0, workspace = 1204224
Read data from input file /spb-hlc-fs2/1/users/atamazov/miopen-verify2/conv_bwd_dat_0_2_0_1_8x1024x14x14_512x1024x1x1_0x0_2x2_1x1_0x0_0_0_GPUbfloat16_REFfloat
Backward Convolution Data Failed: 0.114971 > 0.0656
```
GEMM in baseline 7bd59992a works correctly.

DrizztDoUrden:
Fixed

atamazov:
@DrizztDoUrden FP32, FP16, BF16, both OpenCL and HIP backends. Incorrect time measurements during Backward Convolution phase (after Find, MODE=NORMAL) for the `GemmBwdRest` solver.

Whenever `GemmBwdRest` solver is selected (in the log: `BWD Chosen Algorithm: GemmBwdRest`), the measured time is ~2X bigger than it should be (and actually measured during Find). It is even bigger than Wall-clock time:
```
Wall-clock Time Backward Data Conv. Elapsed: 7.74295 ms...
...
GPU Kernel Time Backward Data Conv. Elapsed: 13.993815 ms (average)
```
Time measurements in Immediate Mode are also affected.

This is minor issue, but we need to fix it because it affects performance testing.

[8e7a0fb8e.ddu_bwd-gemm-solvers.logs.zip](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6192908/8e7a0fb8e.ddu_bwd-gemm-solvers.logs.zip)

DrizztDoUrden:
No idea why it seems stuck on GH. It looks finished on Jenkins...

DrizztDoUrden:
Btw, fixed the last mentioned issue.

atamazov (approved):
LGTM!

src/solver/gemm_bwd.cpp
```diff
@@ -0,0 +1,761 @@
...
+
+    const auto group_count = conv.group_count;
+
+    const auto spatial_dim  = conv.GetSpatialDimension();
+    const auto in_spatial_  = boost::adaptors::slice(dxDesc.GetLengths(), 2, 2 + spatial_dim);
+    const auto wei_spatial_ = boost::adaptors::slice(wDesc.GetLengths(), 2, 2 + spatial_dim);
+    const auto out_spatial_ = boost::adaptors::slice(dyDesc.GetLengths(), 2, 2 + spatial_dim);
+    const auto in_spatial   = std::vector<std::size_t>(in_spatial_.begin(), in_spatial_.end());
+    const auto wei_spatial  = std::vector<std::size_t>(wei_spatial_.begin(), wei_spatial_.end());
+    const auto out_spatial  = std::vector<std::size_t>(out_spatial_.begin(), out_spatial_.end());
```
atamazov, line 595: @asroy @fronteer @shurale-nkn Notice this. We pass everything to InvokerFactory (and to Invoker) by value. The copies of values are captured in Factory (and Invoker) lambda instances, so later we can call Factory (or Invoker) from any context.

However `slice` returns _reference_ to the original object. Even if you pass it to an InvokerFactory (or to Invoker) by-value, only reference will be captured . Subsequent invocation of Factory (or Invoker) will cause referencing of non-existing object.

That is why @DrizztDoUrden creates copies passes that copies to the Factory.

DrizztDoUrden: The alternative is calculating slices inside of invoker from invoke ctx tensor descriptors.

DrizztDoUrden: Essentially, you would want to capture anything tiny or time-consuming to calculate, leaving only the rest to be calculated in the invoker.

PR #805, opened 2021-03-17
Add latest rocblas api flag (SWDEV-275854)

ce1adon:
To jira ticket http://ontrack-internal.amd.com/browse/SWDEV-275854

### DIFF
--- a/src/gemm_v2.cpp
+++ b/src/gemm_v2.cpp
@@ -429,7 +429,7 @@ miopenStatus_t CallGemm(const Handle& handle,
                 rocblas_datatype::rocblas_datatype_i32_r,
                 rocblas_gemm_algo::rocblas_gemm_algo_standard,
                 0,
-                0);
+                rocblas_gemm_flags_pack_int8x4);
         }
         break;
         case miopenInt32: break;
@@ -731,7 +731,7 @@ miopenStatus_t CallGemmStridedBatched(const Handle& handle,
                 rocblas_datatype::rocblas_datatype_i32_r,
                 rocblas_gemm_algo::rocblas_gemm_algo_standard,
                 0,
-                0);
+                rocblas_gemm_flags_pack_int8x4);
         }
         break;
         case miopenInt32: break;
@@ -971,7 +971,7 @@ miopenStatus_t CallGemmStridedBatchedSequential(const Handle& handle,
                     rocblas_datatype::rocblas_datatype_i32_r,
                     rocblas_gemm_algo::rocblas_gemm_algo_standard,
                     0,
-                    0);
+                    rocblas_gemm_flags_pack_int8x4);
             }
         }
         break;
--- a/src/gemm_v2.cpp
+++ b/src/gemm_v2.cpp
@@ -58,6 +58,9 @@
 /// "disabled expansion of recursive macro" injected by rocblas headers.
 #define AVOID_ROCBLAS_WRAPPERS_204 (MIOPEN_ROCBLAS_VERSION_DECIMAL >= 204)
 
+/// Maintain API compatibility with various rocBLAS version
+#define USE_GEMM_FLAGS_PACK_INT8X4 (MIOPEN_ROCBLAS_VERSION_DECIMAL >= 238)
+
 template <class... Ts>
 auto miopen_rocblas_gemm_ex(Ts... xs)
 {
@@ -429,7 +432,12 @@ miopenStatus_t CallGemm(const Handle& handle,
                 rocblas_datatype::rocblas_datatype_i32_r,
                 rocblas_gemm_algo::rocblas_gemm_algo_standard,
                 0,
-                rocblas_gemm_flags_pack_int8x4);
+#if USE_GEMM_FLAGS_PACK_INT8X4
+                rocblas_gemm_flags_pack_int8x4
+#else
+                0
+#endif
+                );
         }
         break;
         case miopenInt32: break;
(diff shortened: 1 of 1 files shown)


src/gemm_v2.cpp
```diff
@@ -429,7 +429,7 @@ miopenStatus_t CallGemm(const Handle& handle,
                 rocblas_datatype::rocblas_datatype_i32_r,
                 rocblas_gemm_algo::rocblas_gemm_algo_standard,
                 0,
-                0);
+                rocblas_gemm_flags_pack_int8x4);
```
atamazov, line 432: Add something like
```
#define USE_GEMM_FLAGS_PACK_INT8X4 (MIOPEN_ROCBLAS_VERSION_DECIMAL >= 238)
```
and use it to maintain backward compatibility with previous rocBLAS. Their changelog says this begins from 2.38, https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/72a59738b291e642f5fecb0b2d5ab2123a2e7793/CHANGELOG.md.

Example:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/bdd81a25f0dd6a913c2c2636ecd2c3edd1baf733/src/gemm_v2.cpp#L57-L59

atamazov (changes requested):

src/gemm_v2.cpp
```diff
@@ -971,7 +971,7 @@ miopenStatus_t CallGemmStridedBatchedSequential(const Handle& handle,
                     rocblas_datatype::rocblas_datatype_i32_r,
                     rocblas_gemm_algo::rocblas_gemm_algo_standard,
                     0,
-                    0);
+                    rocblas_gemm_flags_pack_int8x4);
```
scxiao, line 974: Not sure whether this flag also applies to int32 data type, see https://rocblas.readthedocs.io/en/latest/functions.html#rocblas-gemm-ex-batched-strided-batched

src/gemm_v2.cpp
scxiao: For the data type miopenInt8, I think the last flag should still be 0. Or this is just a work around for the bug?

ce1adon:
> For the data type miopenInt8, I think the last flag should still be 0. Or this is just a work around for the bug?
In fact the entire int8 path is based on int8x4 from the every beginning. This should be OK. Let's see if the CI complains.

src/gemm_v2.cpp
ce1adon: We don't support int32 yet.

src/gemm_v2.cpp
ce1adon: In fact the entire int8 path is based on int8x4 from the every beginning. This should be OK. Let's see if the CI complains. @scxiao

scxiao: I guess the reason for everything is int8x4 is because rocblas only support int8x4 (root cause is our hardware only supports int8x4). With rocblas supporting int8 (no need to be int8x4), we will have the non-int8x4 input, maybe that can be fixed later.

atamazov (approved):
LGTM

scxiao:
@ce1adon I just tried the PR, the unit test failed passed with this change.

ce1adon:
> @ce1adon I just tried the PR, the unit test failed passed with this change.

Can you provide docker, SW version and test name?

scxiao:
> > @ce1adon I just tried the PR, the unit test failed passed with this change.
> 
> Can you provide docker, SW version and test name?

we have a docker image mevermeulen/rocm-migraphx:osdb-6629 that is based on rocm4.2, you can use the following steps to verify you changes:

```
docker pull mevermeulen/rocm-migraphx:osdb-6629
docker run -it --device=/dev/kfd --device=/dev/dri --group-add video --cap-add SYS_PTRACE mevermeulen/rocm-migraphx:osdb-6629
# install packages required to run unit tests
pip3 install onnx==1.7.0 pytest
# go to the home directory
cd
git clone https://github.com/ROCmSoftwarePlatform/AMDMIGraphX.git
cd AMDMIGraphX
mkdir build && cd build
CXX=/opt/rocm/llvm/bin/clang++ cmake ..
make -j64 check
```

At this time, you will see the quant_conv unit test failed.
If you build MIOpen with your changes and install it to override whatever in the container, all unit tests should pass.

scxiao (approved):

ce1adon:
@scxiao The docker you provided doesn't have python venv properly installed. Could that be part of the reason of test failure?

ce1adon:
@scxiao 

```
CMake Error at /usr/share/cmake-3.10/Modules/CMakeFindDependencyMacro.cmake:48 (find_package):
  By not providing "FindAMDDeviceLibs.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "AMDDeviceLibs", but CMake did not find one.

  Could not find a package configuration file provided by "AMDDeviceLibs"
  with any of the following names:

    AMDDeviceLibsConfig.cmake
    amddevicelibs-config.cmake

  Add the installation prefix of "AMDDeviceLibs" to CMAKE_PREFIX_PATH or set
  "AMDDeviceLibs_DIR" to a directory containing one of the above files.  If
  "AMDDeviceLibs" provides a separate development package or SDK, be sure it
  has been installed.
  /opt/rocm/hip/lib/cmake/hip/hip-config.cmake:128 (find_dependency)
  /tmp/cget-22-50-33-bp67N-1/build/virtualenv/lib/python3.6/site-packages/Tensile/Source/CMakeLists.txt:40 (find_package)
```

scxiao:
> Could not find a package configuration file provided by "AMDDeviceLibs"

The container works well for me. After some search, can you try the following command

```
apt install comgr --reinstall
```

atamazov:
@ce1adon Thanks for quick response!

ce1adon:
@scxiao Changes merged in develop now.

Issue #801, opened 2021-03-16
[Jenkinsfile][NFC] Remove docker image name handling -- it is useless and just clutters the file.


atamazov:
> Are you saying the docker image name is not important? What I see is that if docker image name is the same for different docker.build() commands, it always rebuilds!

Yes, image name is not important. Build process starts in any case, but if we have cached results it will be faster.

_Originally posted by @shurale-nkn in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/793#issuecomment-795806197_

PR #798, opened 2021-03-14
[rocm-4.1.x][HOTFIX][WORKAROUND][OCL] Disable MIOpenGEMM for failing configs. Backup GEMM with Naive Direct Solvers for Immediate Fallback.

atamazov:
This PR is intended for 4.1 release only. Related ticket: http://ontrack-internal.amd.com/browse/SWDEV-276757

W/A for MIOpenGEMM issues with ROCm 4.1. The exact reason is not known yet, the most likely reason is a bug in MIOpenGEMM. Note that MIOpenGEMM is used only for OCL BE and is deprecated. It is also possible that the reason is an issue in the OpenCL compiler, but I do not see any problems with other OpenCL kernels.

Some failing cases:
```
test_immed_conv2d --float --cmode conv --pmode default --group-count 1 --input 1, 3, 224, 224 --weights 1, 3, 11, 11 --pads_strides_dilations 1 1 1 1 1 1 --trans_output_pads 0 0
test_immed_conv2d --float --cmode conv --pmode default --group-count 1 --input 1, 3, 224, 224 --weights 1, 3, 7, 7 --pads_strides_dilations 3 3 2 2 1 1 --trans_output_pads 0 0
test_immed_conv3d --float --cmode conv --pmode default --group-count 1 --input 1, 4, 4, 161, 700 --weights 1, 4, 3, 11, 11 --pads_strides_dilations 3 3 3 2 2 2 4 4 4 --trans_output_pads 0 0 0
```

W/A is in effect only when MIOpenGEMM is used (OCL BE) and includes:
- Disabling GEMM for failing configs.
- Adding Naive Solvers (already cherrypicked to `rocm-4.1.x-staging` from #583). Naive solvers are inteded for use as backup for Immediate Mode Fallback when GEMM is disabled.
- Note: When MIOpenGEMM is not in use, Naive Solvers are disabled. This minimizes impact of the W/A to the HIP backend.

## Some info about the issue:
- Memory access faults of verification errors during Forward convolutions.
- The issue does NOT happen with DEV builds.
- Seen on gfx906 and gfx908, with both ROCm 4.1 RC3 and RC2.
- When you run a single config after clean installation, the issue does not happen.
- The issue does not happen with even if you run `test_conv2d --all`, even if several instances are launched in parallel (I tried with 8).
- What works for me right now is `CTEST_PARALLEL_LEVEL=4 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0 make check -j`
[...]

PR #806, opened 2021-03-19
[Jenkinsfile] Improvements and simplifications

atamazov:
- [coverage] Use installable builds in 4 Full stages out of 9 (BF16/FP16/FP32, OCL/HIP, MI100/Vega)
- [improvement] Get rid of `MIOPEN_GPU_SYNC`
- [code quality] Remove useless cargo cult docker image name handling (resolves issue #801)
- [UX] Always show datatype in stage names. Remove useless comments and empty lines.
- [UX] Remove 'Release' from stage names. Rename 'Fast' and 'Aux' stages to 'Smoke'.
- [code quality] Use only nodes marked with 'nogpu' for tests that require no GPU.
- [code quality] Require 'rocmtest' and 'miopen' labels on all nodes, including gfx908.

[Side note] All relevant CI nodes updated with 'nogpu', 'miopen and 'rocmtest' labels:
- 'nogpu' is set on nodes that are allowed to run tests that do not require a GPU. Currently, these are all nodes except MI100. 
- 'rocmtest' means that the node belongs to the "ROCmSoftwarePlatform" github organization (which now includes MIOpen, AMDMIGraphX, MIOpenGEMM, MIOpenTensile).
- 'miopen' means that the node is allowed to run MIOpen tests.


atamazov:
@pfultz2 This Codecov works like a random number generator ;)

atamazov:
@ce1adon @pfultz2 @shurale-nkn @JehandadKhan All CI tests passed. None substantial performance gain or drop observed. Please approve.

JehandadKhan (approved):

Jenkinsfile
```diff
@@ -1,22 +1,5 @@
...
-        node_name = 'gfx908';
-    } else {
-        node_name = name
-    }
-    return node_name
-}
-
-def default_image_name() {
-    return 'miopen-v2'
+    return 'rocmtest && miopen && ' + name
```
pfultz2, line 2: This should be `(rocmtest || miopen)` so you can allocate nodes just for miopen in the future.

Jenkinsfile
```diff
@@ -43,6 +26,12 @@ def cmake_build(compiler, flags, env4make, extradebugflags, prefixpath){
     if (prefixpath != "/usr/local") {
         configargs = "-DCMAKE_PREFIX_PATH=${prefixpath}"
     }
+    
+    if(flags.contains('-DBUILD_DEV=Off'))
```
pfultz2, line 30: BUILD_DEV is off by default, so the absence of this flag doesn't mean that BUILD_DEV is enabled.

Jenkinsfile
```diff
@@ -43,6 +26,12 @@ def cmake_build(compiler, flags, env4make, extradebugflags, prefixpath){
...
+    {
+    	config_targets = 'install ' + config_targets
+    	flags = '-DCMAKE_INSTALL_PREFIX=../install ' + flags
```
pfultz2, line 33: Why aren't we using the default install prefix? The packages we build should be able to be used by others.

Jenkinsfile
```diff
@@ -212,7 +196,7 @@ pipeline {
                     }
                 }
                 stage('Clang Format') {
-                    agent{ label rocmnode("rocmtest") }
+                    agent{ label rocmnode("nogpu") }
```
pfultz2, line 199: What happens if only have nodes with gpus? Ideally, these stages should be moved to github hosted runners.

pfultz2:
> This Codecov works like a random number generator ;)

Its not random. This is comparing against 7bd59992a5bcd7c0b615bab10526de262369b324 which is not the latest on develop. The code coverage changes are from 1c720997d853a73d00f5495becd4a2353417f80d.

It seems like Codecov is not finding the base commit correctly, I am not sure why as I dont see this problem with migraphx.

Jenkinsfile
atamazov: The idea is to allocate nodes exclusively for miopen by removing other labels (like `migraphx`) from the nodes. From the PR description:

> [Side note] All relevant CI nodes updated with 'nogpu', 'miopen and 'rocmtest' labels:
> - 'nogpu' is set on nodes that are allowed to run tests that do not require a GPU. Currently, these are all nodes except MI100. 
> - 'rocmtest' means that the node belongs to the "ROCmSoftwarePlatform" github organization (which now includes MIOpen, AMDMIGraphX, MIOpenGEMM, MIOpenTensile).
> - 'miopen' means that the node is allowed to run MIOpen tests.

Jenkinsfile
atamazov: I would like to be explicit in this file.

atamazov: And yes, you are correct, but let's follow the convention that `BUILD_DEV` must always be specified in Jenkinsfile. Then the text will be easier to understand. And there will be no need to insert an additional rule in the "if".

Jenkinsfile
atamazov: I tried once -- it didn't work and I do not know why. I can try once more , if you wish.

Jenkinsfile
atamazov: Any node that has `nogpu` is allowed to run such tests. Currently, these are all Vega nodes and none of MI100s.

atamazov: I can rename it to `allow_nogpu` but this is shorter...

Jenkinsfile
atamazov: This is noted in the the PR description.

Jenkinsfile
atamazov: This is noted in the the PR description.

atamazov:
@pfultz2 pls let me know if this can be merged (or we can discuss more..)

Jenkinsfile
pfultz2: MIGraphX uses `rocmtest || migraphx` so you would need to remove both migraphx label and the rocmtest label.

Jenkinsfile
pfultz2: >  And there will be no need to insert an additional rule in the "if".

There is no additional rule needed. You can write it as `if(!flags.contains('-DBUILD_DEV=On'))` and it will work whether its explicit or not.

Jenkinsfile
pfultz2: Ah yea it may not work because it may need to be sudo. Why are we adding an install step here? When we call `make package` it will run install into the build directory, but the package will use the correct install prefix.

pfultz2:
> pls let me know if this can be merged (or we can discuss more..)

This shouldn't have been merged yet until the broken packages was fixed.

Jenkinsfile
atamazov: Thanks! Currently I do not plan to be greedy and allocate nodes for MIOpen exclusively.

However -- I would appreciate if you find the hierarchic approach in this file more logical and borrow it for MIGraphX.

Jenkinsfile
atamazov: Thanks, I'll do that on next Jenksinfile update.

Jenkinsfile
atamazov: I would like to test "installed" MIOpen. Some aspects of its behavior differ for DEV builds, e.g. binary cache is enabled. This may accelerate the tests. And recently I fiddled around with an issue which is not reproducible with DEV builds (#798).

Or you mean that install step is not necessary to test non-DEV build? IIRC at least installation of system databases is required.

atamazov:
@pfultz2 
> This shouldn't have been merged yet until the broken packages was fixed.

Sorry, but I do not understand. Could you please explain a bit more?

Jenkinsfile
atamazov: resolved via #814

PR #680, opened 2021-01-14
integration of fp16 fwd dynamic asm kernel for gfx908

shaojiewang:
## Add dynamic asm igemm kernel for conv-fwd with fp16 data type on gfx908.
Implementation for multiple dynamic igemm asm kernels with fp16 data type for fwd pass on gfx908. Test is done under **MIOpenDriver**. Test script and perf num data can be found here.
Kernels can run for any `nchw kcyx` size.
### Test script
[run-test-fwd.sh](https://amdcloud-my.sharepoint.com/:u:/g/personal/shaowang_amd_com/EWc7BUVd6w9Ohoidz7yYRpYB6JcJkP4qzsxBRdN8HTF2WQ?e=5C0mdd)

### Perf num
 [fwd_perf_link.xlsx](https://amdcloud-my.sharepoint.com/:x:/g/personal/shaowang_amd_com/ESRsCNgwN2dGr2iMW1kDs2MBFmKbJpPxZtPzObRmsqOcTw?e=7dldYg)

PR #808, opened 2021-03-20
[CI] Remove huge generated kernel headers from tidy checks. Accelerate parallel library builds.

atamazov:
- Removes huge generated kernel headers from tidy checks. On current develop, `make tidy` progressed from ~40 minutes to 15 mins. Without this, #680 takes >5 hours to complete.
- Moves builds of generated kernel sources to the beginning of `MIOpen_source`. This saves ~1 minute during library build.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -474,6 +474,9 @@ if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN
         ocl/fusionopconvocl.cpp
         ocl/fusionopbiasbnactivocl.cpp
         ${PROJECT_BINARY_DIR}/db_path.cpp
+        )
+
+    list(INSERT MIOpen_Source 0
         ${PROJECT_BINARY_DIR}/kernel.cpp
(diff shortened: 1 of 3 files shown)


Kirpich30000 (approved):
LGTM

shurale-nkn:
Better not to exclude kernels data code "${INIT_KERNELS}" from tidy checks, but to reduce the size of this hude file.

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ef4907c43c18358361a908e5271c0d4046a20e09/src/CMakeLists.txt#L53
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ef4907c43c18358361a908e5271c0d4046a20e09/src/CMakeLists.txt#L70

The 'add_kernels' and 'add_kernel_includes' functions are responsible for data generation, and better to reduce list of used kernels for tidy checks, to save test coverage and and minimize test time.

atamazov:
@shurale-nkn "${INIT_KERNELS}" ar so trivial that any problem with it will lead to build error (or will be uncaught by both library build and tidy checks and appear during tests). But you are welcome to implement this feature ;)

I need to merge this PR asap to unblock #680. Any objections?

PR #810, opened 2021-03-21
[HIP build] Do not write unnecessary files to disk.

atamazov:
Each time when HIP kernel is being compiled, the library writes to disk all include files. The total size of this with #680 is ~81MB. ~80 MB of those are assembly `.inc` files, which are irrelevant. This affects performance. HDD write speed on my machine:
```bash
$ sudo dd if=/dev/zero of=/mnt/hdd/dd-test bs=330k count=25000
25000+0 records in
25000+0 records out
8448000000 bytes (8.4 GB, 7.9 GiB) copied, 58.5778 s, 144 MB/s
```
Which means that 80MB of useless files adds ~0.55 sec at least.

This PR writes to disk only HIP headers thus reducing to total size down to ~1MB.


DrizztDoUrden (approved):
LGTM

Issue #803, opened 2021-03-16
Proposal: Override the launch parameter from kernels built by mlir


jerryyin:
## Background: MLIR Build

We added MLIR to be a new backend for kernel compilations. When a solver populates a kernel file that end with `mlir-cpp`, hipoc_program will choose the MLIR backend to build the kernel. Then the hipoc_program will be used similarly as built with any other backend. So far, this has been proven working well with MLIR built kernels.

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/3917634a62dd06a86868303b48b15e853efffc38/src/hipoc/hipoc_program.cpp#L253-L258

## Issue Statement

There is a mismatch between launching parameters of an MLIR kernel versus a regular MIOpen kernel. Let's suppose that we disable perfdb for the below argument. 

When a regular MIOpen solver does not find the problem config in perfdb, it went through `EuristicInit()` to regress on an array of performance config, until finding the coherent set of parameters. In finding the coherent set of performance config, launching parameters are decided as well. The launching parameters decided by solver are later used to launch the kernel.

However, the `EuristicInit()` process is not necessary for MLIR kernels. Instead, MLIR backend is capable of making its own decisions in compiling its kernels. In its process, it also decide on a coherent set of parameters including the launch parameters.

Refer to the following for how MIOpen side of MLIR solver decide launch parameters:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/3917634a62dd06a86868303b48b15e853efffc38/src/solver/conv_hip_implicit_gemm_mlir_cpp_fwd.cpp#L70-L76

Since both the MLIR solver and the MLIR backed try to compute the launching parameters at the same time, there is a likelihood such that the decision made by MIOpen side of MLIR solver, and the MLIR backend are different. It becomes a problem when the launching parameters does become different, ranging from wrong values or GPU VM fault.

In the short term, we prevent this from happening by synchronizing the source code of `EuristicInit` between MLIR and MIOpen, such that both components make the same decisions. However, this source level dependency should be deprecated because of code coupling in the two independent software components:

- Changing one end of logic becomes impossible because the other end is dependent on it
- We do unnecessary computation in MIOpen end to make sure it derive the same launching parameters. The most of computation are wasted on calculation of checking if the performance config are coherent, rather than computing launch parameters.

## Solution

[...]

PR #811, opened 2021-03-22
Making MLIR solver obtain launch params from API

jerryyin:
This PR address https://github.com/ROCmSoftwarePlatform/MIOpen/issues/803

* Finished fwd/bwd/wrw solvers
* Decoupled the MLIR fwd solver from base class
* Refactor from Mlir -> Miir, used latest Miir API

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -95,7 +95,7 @@ RUN pip install -r /doc-requirements.txt
 RUN if [ "$MIOTENSILE_VER" = "latest" ] ; then cget -p $PREFIX install ROCmSoftwarePlatform/MIOpenTensile@7568654c938d42e9a91c6b18fb382f5b978d12fd; else cget -p $PREFIX install ROCmSoftwarePlatform/MIOpenTensile@5fe0bf4a8dc59f3ab62df929297280915372ce16; fi
 
 RUN cd ~ && \
-    export MLIR_COMMIT=06a6d3fc1af1fdfbe706dbfcd588fb3744526124 && \
+    export MLIR_COMMIT=31d92f4c64ae6fa6b7c5d543f68b69300b4513ce && \
     wget https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/archive/$MLIR_COMMIT.tar.gz && \
     tar -xvzf $MLIR_COMMIT.tar.gz && \
     rm -rf $MLIR_COMMIT.tar.gz && \
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -788,7 +788,6 @@ struct ConvHipImplicitGemmV4R4Fwd : SolverBase<ConvolutionContext>
                              bool disableConfigOverrideFromEnv = false) const;
 };
 
-// Mlir based igemm kernels do not support tuning, for now
 struct ConvHipImplicitGemmMlirCppFwd : SolverBase<ConvolutionContext>
 {
     static std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx);
@@ -855,7 +854,6 @@ struct ConvHipImplicitGemmV4R4WrW : SolverBase<ConvolutionContext>
                              bool disableConfigOverrideFromEnv = false) const;
 };
 
-// Mlir based igemm kernels do not support tuning, for now
 struct ConvHipImplicitGemmMlirCppWrW : SolverBase<ConvolutionContext>
 {
     static std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx);
@@ -1254,7 +1252,6 @@ struct ConvHipImplicitGemmBwdDataV1R1 : SolverBase<ConvolutionContext>
     size_t GetWorkspaceSize(const ConvolutionContext& ctx) const;
 };
 
-// Mlir based igemm kernels do not support tuning, for now
 struct ConvHipImplicitGemmMlirCppBwd : SolverBase<ConvolutionContext>
 {
     static std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx);
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -786,24 +786,14 @@ struct ConvHipImplicitGemmV4R4Fwd : SolverBase<ConvolutionContext>
     ConvSolution GetSolution(const ConvolutionContext& ctx,
                              const PerformanceImplicitGemmV4R4Fwd& config,
                              bool disableConfigOverrideFromEnv = false) const;
-
-    protected:
-    bool IsApplicableMlirCommon(const ConvolutionContext& ctx) const;
 };
 
-class ConvHipImplicitGemmMlirCppFwd : ConvHipImplicitGemmV4R4Fwd
+// Mlir based igemm kernels do not support tuning, for now
+struct ConvHipImplicitGemmMlirCppFwd : SolverBase<ConvolutionContext>
 {
-    using Base = ConvHipImplicitGemmV4R4Fwd;
-
-    public:
+    static std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx);
     bool IsApplicable(const ConvolutionContext& ctx) const;
     ConvSolution GetSolution(const ConvolutionContext& ctx) const;
-    bool IsDynamic() const { return Base::IsDynamic(); }
-    float GetWti(const ConvolutionContext& ctx) const { return Base::GetWti(ctx); }
-    size_t GetWorkspaceSize(const ConvolutionContext& ctx) const
-    {
-        return Base::GetWorkspaceSize(ctx);
-    };
 };
 
 struct PerformanceImplicitGemmV4R4GenXdlopsFwdFp32
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -853,24 +853,14 @@ struct ConvHipImplicitGemmV4R4WrW : SolverBase<ConvolutionContext>
     ConvSolution GetSolution(const ConvolutionContext& ctx,
                              const PerformanceImplicitGemmV4R4WrW& config,
                              bool disableConfigOverrideFromEnv = false) const;
-
-    protected:
-    bool IsApplicableMlirCommon(const ConvolutionContext& ctx) const;
 };
 
-class ConvHipImplicitGemmMlirCppWrW : ConvHipImplicitGemmV4R4WrW
+// Mlir based igemm kernels do not support tuning, for now
+struct ConvHipImplicitGemmMlirCppWrW : SolverBase<ConvolutionContext>
 {
-    using Base = ConvHipImplicitGemmV4R4WrW;
-
-    public:
+    static std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx);
     bool IsApplicable(const ConvolutionContext& ctx) const;
     ConvSolution GetSolution(const ConvolutionContext& ctx) const;
-    bool IsDynamic() const { return Base::IsDynamic(); }
-    float GetWti(const ConvolutionContext& ctx) const { return Base::GetWti(ctx); }
-    size_t GetWorkspaceSize(const ConvolutionContext& ctx) const
-    {
-        return Base::GetWorkspaceSize(ctx);
-    };
 };
 
 struct PerformanceImplicitGemmXdlops : Serializable<PerformanceImplicitGemmXdlops>
@@ -1262,24 +1252,14 @@ struct ConvHipImplicitGemmBwdDataV1R1 : SolverBase<ConvolutionContext>
                              const PerformanceImplicitGemmBwdDataV1R1& config,
                              bool disableConfigOverrideFromEnv = false) const;
     size_t GetWorkspaceSize(const ConvolutionContext& ctx) const;
-
-    protected:
-    bool IsApplicableMlirCommon(const ConvolutionContext& ctx) const;
 };
 
-class ConvHipImplicitGemmMlirCppBwd : ConvHipImplicitGemmBwdDataV1R1
+// Mlir based igemm kernels do not support tuning, for now
+struct ConvHipImplicitGemmMlirCppBwd : SolverBase<ConvolutionContext>
 {
-    using Base = ConvHipImplicitGemmBwdDataV1R1;
-
-    public:
+    static std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx);
     bool IsApplicable(const ConvolutionContext& ctx) const;
     ConvSolution GetSolution(const ConvolutionContext& ctx) const;
-    bool IsDynamic() const { return Base::IsDynamic(); }
-    float GetWti(const ConvolutionContext& ctx) const { return Base::GetWti(ctx); }
-    size_t GetWorkspaceSize(const ConvolutionContext& ctx) const
-    {
-        return Base::GetWorkspaceSize(ctx);
-    };
 };
 
 struct ConvHipImplicitGemmBwdDataV4R1 : SolverBase<ConvolutionContext>
--- a/src/mlir_build.cpp
+++ b/src/mlir_build.cpp
@@ -29,23 +29,37 @@
 #include <miopen/logger.hpp>
 #include <miopen/mlir_build.hpp>
 
-#include <mlir-miopen-lib.hpp>
+#include <Miir.h>
 
 #include <fstream>
 
 namespace miopen {
-
+// Anonymous namespace
+namespace {
 /// Destroys handle in case of exception.
-class AutoMlirHandle
+class AutoMiirHandle
 {
-    MlirHandle handle;
+    MiirHandle handle;
 
     public:
-    AutoMlirHandle(const std::string& options) : handle(CreateMlirHandle(options.c_str())) {}
-    ~AutoMlirHandle() { DestroyMlirHandle(handle); }
-    MlirHandle operator()() { return handle; }
+    AutoMiirHandle(const std::string& options) : handle(miirCreateHandle(options.c_str())) {}
+    ~AutoMiirHandle() { miirDestroyHandle(handle); }
+    MiirHandle operator()() { return handle; }
 };
 
+void check_miir_error(MiirStatus status, const std::string& name)
+{
+    if(status == MIIR_SUCCESS)
+        return;
+    if(status == MIIR_INVALID_PARAM)
+        MIOPEN_THROW(name + " failed: invalid parameters!");
+    if(status == MIIR_INVALID_MODULE)
+        MIOPEN_THROW(name + " failed: invalid module!");
+    if(status == MIIR_BUILD_FAILURE)
+        MIOPEN_THROW(name + " failed: build failed!");
+    MIOPEN_THROW(name + " failed: unkown error!");
+}
+} // namespace
 /// Generates HIP source, header and options for HIP compiler.
 /// Writes HIP source and header into output directory.
 ///
@@ -54,7 +68,7 @@ class AutoMlirHandle
 /// @param[in]  params   Options for MLIR generator.
 /// @param[out] cflags   Build options for HIP compiler.
 /// @param[out] cpp_filename  Name of output .cpp file (with path).
-static void MlirGenerateSourcesForHipBuild(const boost::optional<TmpDir>& tmp_dir,
+static void MiirGenerateSourcesForHipBuild(const boost::optional<TmpDir>& tmp_dir,
                                            const std::string& filename,
                                            const std::string& params,
                                            std::string& cflags,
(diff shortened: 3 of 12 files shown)


atamazov:
@jerryyin 
> I decide to not overwhelm the PR with changes from 3 solvers. Instead, this draft changed only the fwd direction solver. If everything looks good...

:+1:

src/mlir_build.cpp
```diff
@@ -29,23 +29,37 @@
...
+    ~AutoMiirHandle() { miirDestroyHandle(handle); }
+    MiirHandle operator()() { return handle; }
 };
 
+void check_miir_error(MiirStatus status, const std::string& name)
+{
+    if(status == MIIR_SUCCESS)
+        return;
+    if(status == MIIR_INVALID_PARAM)
+        MIOPEN_THROW(name + " failed: invalid parameters!");
```
atamazov, line 55:
```suggestion
        MIOPEN_THROW(miir_fn_name + ": MIIR_INVALID_PARAM");
```
Ditto other cases.

src/solver/conv_hip_implicit_gemm_mlir_cpp_fwd.cpp
```diff
@@ -48,9 +70,26 @@ bool ConvHipImplicitGemmMlirCppFwd::IsApplicable(const ConvolutionContext& ctx)
     // Future: MLIR will support non-default layouts.
     if(!ctx.IsLayoutDefault())
         return false;
+    // Below: Generic checks between this solver and ConvHipImplicitGemmV4R4Fwd
```
atamazov, line 73: This comment should be moved two lines below.

src/mlir_build.cpp
```diff
@@ -122,4 +136,14 @@ boost::filesystem::path MlirBuildViaHip(boost::optional<TmpDir>& tmp_dir,
     return HipBuild(tmp_dir, hip_filename, src, hip_options, target, true);
 }
 
+void MiirGenLaunchParams(const std::string& params, size_t& local_size, size_t& global_size)
+{
+    AutoMiirHandle handle(params.c_str());
+    miirLowerInit();
```
atamazov, line 142: [Q] I guess you need to pass handle to it?

Dockerfile
```diff
@@ -95,7 +95,7 @@ RUN pip install -r /doc-requirements.txt
 RUN if [ "$MIOTENSILE_VER" = "latest" ] ; then cget -p $PREFIX install ROCmSoftwarePlatform/MIOpenTensile@7568654c938d42e9a91c6b18fb382f5b978d12fd; else cget -p $PREFIX install ROCmSoftwarePlatform/MIOpenTensile@5fe0bf4a8dc59f3ab62df929297280915372ce16; fi
 
 RUN cd ~ && \
-    export MLIR_COMMIT=06a6d3fc1af1fdfbe706dbfcd588fb3744526124 && \
+    export MLIR_COMMIT=31d92f4c64ae6fa6b7c5d543f68b69300b4513ce && \
```
atamazov, line 98: I think this would break all other solvers. You need to disable those solvers and respective tests, otherwise CI tests won't pass.

atamazov (changes requested):

atamazov:
> ...If everything looks good, I will continue to refactor the bwd and wrw solvers.

I recommend doing that in a separate PR.

src/include/miopen/solver.hpp
```diff
@@ -786,24 +786,14 @@ struct ConvHipImplicitGemmV4R4Fwd : SolverBase<ConvolutionContext>
...
 {
-    using Base = ConvHipImplicitGemmV4R4Fwd;
-
-    public:
+    static std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx);
     bool IsApplicable(const ConvolutionContext& ctx) const;
     ConvSolution GetSolution(const ConvolutionContext& ctx) const;
-    bool IsDynamic() const { return Base::IsDynamic(); }
-    float GetWti(const ConvolutionContext& ctx) const { return Base::GetWti(ctx); }
-    size_t GetWorkspaceSize(const ConvolutionContext& ctx) const
```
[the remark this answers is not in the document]
jerryyin: Are you talking about the `GetWorkspaceSize()`? `SolverBase` has a implementation that always return 0, it should give the child class a free implementation. Also I don't see this defined in existing solvers like `ConvHipImplicitGemmV4R4Fwd`, for example.

src/mlir_build.cpp
```diff
@@ -29,23 +29,37 @@
...
     public:
-    AutoMlirHandle(const std::string& options) : handle(CreateMlirHandle(options.c_str())) {}
-    ~AutoMlirHandle() { DestroyMlirHandle(handle); }
-    MlirHandle operator()() { return handle; }
+    AutoMiirHandle(const std::string& options) : handle(miirCreateHandle(options.c_str())) {}
+    ~AutoMiirHandle() { miirDestroyHandle(handle); }
+    MiirHandle operator()() { return handle; }
 };
 
+void check_miir_error(MiirStatus status, const std::string& name)
```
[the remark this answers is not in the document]
jerryyin: I wrapped the whole thing (together with the `AutoMiirHandle` class) in an anonymous workspace so it is guaranteed to be used only inside this translation unit. I think `static` achieve the same thing here. It's a matter of preference of using one over the other.

src/mlir_build.cpp
jerryyin: Sure

src/mlir_build.cpp
jerryyin: So far I don't see a good reason to make users of `mlir_build` understand the implementation detail like`AutoMiirHandle` (Therefore I hide the type in an anonymous namespace). In each use case of `MiirGenLaunchParams` or `MiirBuildViaHip`, there's no overhead I see so far.

Dockerfile
jerryyin: Could you clarify what this means?

I think I should update the Jenkinsfile's default docker image name because I updated the dockerfile. Will do that in next commit. But I'm not sure why bumping the commit id will break all other solvers or CI?

src/mlir_build.cpp
```diff
@@ -29,23 +29,37 @@
...
+{
+    if(status == MIIR_SUCCESS)
```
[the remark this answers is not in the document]
jerryyin: Will do

src/solver/conv_hip_implicit_gemm_mlir_cpp_fwd.cpp
jerryyin: That's right.

Dockerfile
jerryyin: Should I bump this to `miopen-v1`?

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/8e2f418d41e456f760ef7deb61d4244635ce0818/Jenkinsfile#L67

src/include/miopen/solver.hpp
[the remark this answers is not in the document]
atamazov: Ok

src/mlir_build.cpp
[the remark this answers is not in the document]
atamazov: Ok, missed it.

src/mlir_build.cpp
atamazov: `miirLowerInit(handel());` ?

Dockerfile
atamazov: > Should I bump this to `miopen-v1`?

This is not needed anymore, see #801

atamazov: > Could you clarify what this means?

Unmodified MLIR solvers for Bwd/WrW directions do not pass num CUs and device name to miirBuildViaHip.

src/mlir_build.cpp
jerryyin: Gotcha, I see what you mean now. No this initialization initialize the lowering pipeline and is handle independent.

Dockerfile
jerryyin: @atamazov Thanks for link to #801.

> Unmodified MLIR solvers for Bwd/WrW directions do not pass num CUs and device name to miirBuildViaHip.

You reminded me, thank you! Locally I have already changed both bwd and wrw solvers that's why I didn't see any issue coming up. Now, it just look like it doesn't worth it to hold back those changes because it will make the code base dependent on different MLIR commit. I will add the other two changes to the PR two, but in different commit, strictly similar to changes I made in fwd so hopefully this doesn't make your review to be a pain.

src/mlir_build.cpp
atamazov: How it could be so? Where the internal state of the lowering pipeline resides?

Dockerfile
atamazov: Gotcha!

But okay. Just use different commits. Sometimes keeping this kind of discipline is painful, but it is worth it.

atamazov: [Note] Each commit pushed to repo will start a new CI job. So make all your commits locally and then push once.

src/mlir_build.cpp
```diff
@@ -47,17 +47,16 @@ class AutoMiirHandle
...
-    if(status == MIIR_BUILD_FAILURE)
-        MIOPEN_THROW(name + " failed: build failed!");
-    MIOPEN_THROW(name + " failed: unkown error!");
+    switch(status)
+    {
+    case MIIR_SUCCESS: return;
+    case MIIR_INVALID_PARAM: MIOPEN_THROW(miir_fn_name + " MIIR_INVALID_PARAM"); break;
+    case MIIR_INVALID_MODULE: MIOPEN_THROW(miir_fn_name + " MIIR_INVALID_MODULE"); break;
+    case MIIR_BUILD_FAILURE: MIOPEN_THROW(miir_fn_name + " MIIR_BUILD_FAILURE"); break;
+    default: MIOPEN_THROW(miir_fn_name + " failed: unkown error!");
```
atamazov, line 58: Typo. Recommendation:
```
MIOPEN_THROW(miir_fn_name + " <UNKNOWN ERROR>");
```

Dockerfile
jerryyin: @atamazov 

> But okay. Just use different commits. Sometimes keeping this kind of discipline is painful, but it is worth it.

Next time I should adjust my development model to this (Not just myself ahead of the PR) as well. This time reverting the changes + disabling the solvers does not worth it considering I have the changes ready.

> [Note] Each commit pushed to repo will start a new CI job. So make all your commits locally and then push once.

I noticed that, but thanks for reminding.

src/solver/conv_hip_implicit_gemm_mlir_cpp_fwd.cpp
```diff
@@ -143,7 +142,9 @@ ConvSolution ConvHipImplicitGemmMlirCppFwd::GetSolution(const ConvolutionContext
 
     size_t local_size  = 0;
     size_t global_size = 0;
+#if MIOPEN_USE_MLIR
```
atamazov, line 145: I recommend wrapping the whole function body in this #if.

[Reminder] `std:ignore = <unused param>` under #else to avoid tidy warnings.

atamazov (changes requested):

src/mlir_build.cpp
atamazov: This is important Q.

src/include/miopen/solver.hpp
```diff
@@ -853,24 +853,14 @@ struct ConvHipImplicitGemmV4R4WrW : SolverBase<ConvolutionContext>
     ConvSolution GetSolution(const ConvolutionContext& ctx,
                              const PerformanceImplicitGemmV4R4WrW& config,
                              bool disableConfigOverrideFromEnv = false) const;
-
-    protected:
-    bool IsApplicableMlirCommon(const ConvolutionContext& ctx) const;
 };
 
-class ConvHipImplicitGemmMlirCppWrW : ConvHipImplicitGemmV4R4WrW
+// Mlir based igemm kernels do not support tuning, for now
```
atamazov, line 858: Please remove the comment -- it is trivial ;)

src/include/miopen/solver.hpp
```diff
@@ -1262,24 +1252,14 @@ struct ConvHipImplicitGemmBwdDataV1R1 : SolverBase<ConvolutionContext>
                              const PerformanceImplicitGemmBwdDataV1R1& config,
                              bool disableConfigOverrideFromEnv = false) const;
     size_t GetWorkspaceSize(const ConvolutionContext& ctx) const;
-
-    protected:
-    bool IsApplicableMlirCommon(const ConvolutionContext& ctx) const;
 };
 
-class ConvHipImplicitGemmMlirCppBwd : ConvHipImplicitGemmBwdDataV1R1
+// Mlir based igemm kernels do not support tuning, for now
```
atamazov, line 1257: Ditto

src/mlir_build.cpp
jerryyin: The lowering pipelines needs to be registered as LLVM passes before being invoked in actual lowering step, [here](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp#L265-L274). Previously the `miirLowerBin` itself will do the initialization as well when it is called. But now since we added `miirLowerTuningParams` I'd rather maintain one orthogonal function that does the common initialization logic instead of duplicated code.

src/solver/conv_hip_implicit_gemm_mlir_cpp_bwd.cpp
```diff
@@ -49,9 +70,31 @@ bool ConvHipImplicitGemmMlirCppBwd::IsApplicable(const ConvolutionContext& ctx)
...
     if(!ctx.IsLayoutDefault())
         return false;
-    if(ctx.Is3d())
+    // Future: MLIR will support 3d convolution
+    if(!ctx.Is2d())
+        return false;
+    // Future: MLIR will support multiple data types
+    if(!ctx.IsFp32())
+        return false;
+    // Below: Generic checks between this solver and ConvHipImplicitGemmBwdDataV1R1
```
atamazov, line 79: [Recommendation] If/when `ConvHipImplicitGemmBwdDataV1R1` change, the comment will become invalid. I would recommend removing trivial comments.

src/solver/conv_hip_implicit_gemm_mlir_cpp_bwd.cpp
```diff
@@ -111,6 +146,20 @@ ConvSolution ConvHipImplicitGemmMlirCppBwd::GetSolution(const ConvolutionContext
         std::string(" --kernel_name ") + construction_parameters.kernel_name;
     // clang-format on
 
+    size_t local_size  = 0;
+    size_t global_size = 0;
+#if MIOPEN_USE_MLIR
```
atamazov, line 151: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/811#discussion_r599888889

src/solver/conv_hip_implicit_gemm_mlir_cpp_wrw.cpp
```diff
@@ -111,6 +140,20 @@ ConvSolution ConvHipImplicitGemmMlirCppWrW::GetSolution(const ConvolutionContext
         std::string(" --kernel_name ") + construction_parameters.kernel_name;
     // clang-format on
 
+    size_t local_size  = 0;
+    size_t global_size = 0;
+#if MIOPEN_USE_MLIR
```
atamazov, line 145: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/811#discussion_r599888889

atamazov (changes requested):

src/mlir_build.cpp
```diff
@@ -76,25 +89,25 @@ static void MlirGenerateSourcesForHipBuild(const boost::optional<TmpDir>& tmp_di
 
     MIOPEN_LOG_I2(input_file.string() << ", options: '" << params << "'");
 
-    AutoMlirHandle handle(params);
-    MlirLowerCpp(handle());
+    AutoMiirHandle handle(params);
+    miirLowerCpp(handle());
```
atamazov, line 93: `miirLowerInit()` is missing here.

atamazov (changes requested):

src/solver/conv_hip_implicit_gemm_mlir_cpp_fwd.cpp
jerryyin: Both `local_size` and `global_size` are actually used a couple of lines below. They are pushed to the `l_wk`, `g_wk` vectors. 

The thing with commenting out the body is that I still have to keep the `result` local variable and return statement such that it doesn't complain. I can't tell which is cleaner.

src/mlir_build.cpp
jerryyin: My bad

src/mlir_build.cpp
atamazov: Are you sure that https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp#L265-L274 is thread-safe?

src/mlir_build.cpp
jerryyin: `miirLowerCpp()` is the only lowering pipeline that does not need any llvm passes. We hacked out the pass such that it populate cpp, rather than other sane products such as llvm ir. The more I speak of it only make me to hate it more. Sorry about the confusions for it caused. We should just leave the cpp backend as-is for now.

src/include/miopen/solver.hpp
jerryyin: Will get rid of them

src/mlir_build.cpp
atamazov: NP

src/solver/conv_hip_implicit_gemm_mlir_cpp_bwd.cpp
jerryyin: That's true, will get rid of them.

src/solver/conv_hip_implicit_gemm_mlir_cpp_fwd.cpp
atamazov: `return {};`.

This is cleaner -- the reader would not need to think why the half-cooked result is returned.

src/mlir_build.cpp
atamazov: NP! This is experimental.

src/mlir_build.cpp
jerryyin: It is: user is allowed to initialize as many times as they want. It won't do anything concrete if pipeline is already intialized.

Say we have thread `a` and thread `b`, as long as thread a/b do initialize before their own lowering, `LLVM` will be happy. It doesn't matter if the global sequence happen to be that the initialization happens in middle of the lowering.

jerryyin: I will try to poke on it a bit more and see if I can skip those, say, if client is already initialized. Even if the whole thing can work  doesn't mean that it should be repeatedly initialized.

atamazov: What if thread A is doing the lowering but thread B is initializing? That is the case I am worrying of. Solution:
```C++
// C++
bool lazy_init()
{
    static const bool once = []()
        {
            ... // initialization
            return true;
        }();
    return once;
}
```
This can be done directly in the library.

src/solver/conv_hip_implicit_gemm_mlir_cpp_bwd.cpp
```diff
@@ -162,6 +161,9 @@ ConvSolution ConvHipImplicitGemmMlirCppBwd::GetSolution(const ConvolutionContext
     result.invoker_factory = conv::MakeImplGemmDataInvokerFactory(ctx);
     result.construction_params.push_back(construction_parameters);
     return result;
+#else
+    return {};
```
atamazov, line 165: Tidy will complain on unused `ctx`

src/mlir_build.cpp
jerryyin: I checked around and find the documents [here](https://llvm.org/doxygen/classllvm_1_1PassRegistry.html#details). Looks like it confirmed that: Each thread has to do its own registration. This means that thread A and B should both do their own intialization before hand.

Very nice code snippet. I think I see this implementation in singleton pattern but never thought about applying this elsewhere. Will adopt it in library implementation.

jerryyin: What this means is that I shouldn't expose the initialization as a client function, but as a convenient function inside the library. I shouldn't expect the client to be used in one way rather than another.

atamazov: > Will adopt it in library implementation.

Maybe. But if applied here it will perform initialization in one thread only.

> I checked around and find the documents [here](https://llvm.org/doxygen/classllvm_1_1PassRegistry.html#details). 

Well, it says "you will need to use a _separate_ PassRegistry on each thread" but this is not the same as _initialization of the same instance of the PassRegistry_ in each thread.

MIOpen spawns threads during build. In order to stay in the thread-safe realm, we need to create separate instance of PassRegistry per each thread. The simplest way to reach that goal is to aggregate PassRegistry in the MLIR handle (by value or by pointer/reference). Is it possible?

jerryyin: > MIOpen spawns threads during build. In order to stay in the thread-safe realm, we need to create separate instance of PassRegistry per each thread. 

I think I got myself confused in the `PassRegistry` reading. `PassRegistry` and initialization are two independent topics and yes, we do have separate instance of `PassRegistry` per thread(call).

Refer to [here](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp#L284).

atamazov: Ok, let's get back to the original question, https://github.com/ROCmSoftwarePlatform/MIOpen/pull/811#discussion_r599902994. Is it Ok to initialize by calling `miirLowerInit()` once? Would it work for all threads, all handles? If so, then let's implement lazy initialization, see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/811#discussion_r599941733.

This is important because I believe that [the sequence of initialization calls](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/mlir/tools/mlir-miopen-lib/mlir-miopen-lib.cpp#L265-L274) in thread A should not happen in parallel with some other thread B that performs lowering.

jerryyin: MLIR [issue](https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/115) filed for this

atamazov: This is not a problem until we do not have > 1 MLIR solver in the library. Resolved.

atamazov: ...and we still may have issues with multi-threaded user's applications, but not now because MLIR Solvers are OFF by default.

atamazov (commented):
LGTM

atamazov (approved):
LGTM

atamazov:
I need to resolve issues in develop before merging.

PR #814, opened 2021-03-24
[CI] Fix & refactor build/install directory handling.

atamazov:
We should always clean `./install` directory. Leftowers after failed/terminated tests are not just unwanted but also lead to clang-format failures.

Also resolves: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/806#discussion_r598148074


shurale-nkn (approved):
LGTM

PR #815, opened 2021-03-25
add param to control CI stages

ce1adon:
Replace with #837


ce1adon:
Looks like CI doesn't support "when" condition.
Will switch to parameter at step level.

PR #832, opened 2021-03-28
[HOTFIX][COMGR] TargetId support (SWDEV-279172)

atamazov:
- [COMGR] Add support fir TargetId. Read actually reported features & use them in COMGR.
  -  DKMS driver older than 5.9 may report incorrect state of SRAMECC feature. Therefore we compute default SRAMECC and rely on it for now, wherever possible.
- [HIP build] Decide return type of llvm.amdgcn.buffer.atomic.fadd.f32 by HIP version, disable auto-detection.
  - Auto-detection can't be used with COMGR
  - Auto-detection is unreliable with ROCm 3.10 at least
  - Keep auto-detection code in the library; it may be needed in the future.

This resolves issue with PyTorch ROCm 4.1 wheel, http://ontrack-internal.amd.com/browse/SWDEV-279172


atamazov:
Please review ASAP.

src/comgr.cpp
```diff
@@ -305,10 +303,15 @@ static void RemoveLinkOptionsUnwanted(OptionList& list)
 } // namespace hip
 
 /// \todo Get list of supported isa names from comgr and select.
-static std::string GetIsaName(const std::string& device)
+static std::string GetIsaName(const miopen::TargetProperties& target)
 {
-    const char* const ecc_suffix = IsEnabledFeatureSramEcc(device) ? "+sram-ecc" : "";
-    return {"amdgcn-amd-amdhsa--" + device + ecc_suffix};
+#if ROCM_FEATURE_TARGETID_OFF
+    const char* const ecc_suffix = target.Sramecc() ? "+sram-ecc" : "";
```
jeffdaily, line 309: Shouldn't it be this instead?
```suggestion
    const char* const ecc_suffix = target.Sramecc() ? ":sramecc+" : "";
```

jeffdaily (changes requested):
Only one change requested.

Will this be backported to the rocm-4.1.x branch, or some sort of rocm-4.1.x-hotfix branch?  This PR is targeting develop, but we would need the fixes backported to 4.1 for PyTorch wheels.

src/comgr.cpp
jeffdaily: On further thought, I see this is in the "ROCM_FEATURE_TARGETID_OFF" section of the code, so I think my change request is incorrect.  The code here is pre-targetid, and it was working there.  Right?

atamazov:
@jeffdaily Thanks for reviewing.

> Will this be backported to the rocm-4.1.x branch, or some sort of rocm-4.1.x-hotfix branch?

ROCm 4.1 release is already complete. According to the process, the `4.1.x` branch will be updated only if we will release 4.1 patch update (e.g. include updated MIOpen in ROCm 4.1.1). I am not aware of such plans for now.

However, nothing prevents me from updating the `4.1.x-staging` branch. Would it work for you?

I would appreciate if you take the development branch (`comgr-targetid-fixes`) right away and perform some smoke testing.

> This PR is targeting develop, but we would need the fixes backported to 4.1 for PyTorch wheels.

This is done according to the process. Almost all patches are coming to `develop` first. When/if we need to include some patch to the _existing_ release branch (but _before_ release in complete), we cherry-pick the patch to the `rocm-x.y-staging` branch, test it there and then merge release staging branch to the release branch.

This way we are going to update the rocm-4.2.x-staging branch and give to CQE for testing ASAP.

src/comgr.cpp
atamazov: Yes!

atamazov:
Testing info added to the PR description.

jeffdaily:
@atamazov built and tested PyTorch wheels using ROCm 4.1 + this PR.  Not a full test, but a smoke test - was able to run resnet50 microbench on MI50 and MI100.

atamazov:
@jeffdaily Thanks! I am assuming that smoke test fails without this PR, right?

jeffdaily:
> @jeffdaily Thanks! I am assuming that smoke test fails without this PR, right?

Correct, without this PR I get the dreaded `Intrinsic has incorrect return type!` when building pytorch wheels + ROCm 4.1 + miopen rocm-4.1.x + comgr.

jeffdaily:
> However, nothing prevents me from updating the `4.1.x-staging` branch. Would it work for you?

This works for me.  If you update that branch, we can switch to it for our wheel builds, no problem.

atamazov:
@jeffdaily I will update the `4.1.x-staging` branch` as soon as it will pass CI tests (or sooner). Stay tuned.

atamazov:
@jeffdaily Please approve this (technically, I need this to merge). Thanks.

atamazov:
Someone please approve this ASAP.

jeffdaily (approved):
LGTM

atamazov:
Thanks!

asroy (approved):

Kirpich30000 (approved):
LGTM

DrizztDoUrden (approved):
LGTM

atamazov:
@jeffdaily 
> I will update the `4.1.x-staging` branch` as soon as it will pass CI tests (or sooner). Stay tuned.

Done!

PR #556, opened 2020-10-29
MIOpenTensile Part2 : Support low precision types

ce1adon:
The purpose of this PR is to add MIOpenTensile path in GEMM algorithm for low precision types on hip backend.
This PR depends on #139. ~~Set "on hold" till #139 merged.~~
Update: The targetID env changes are in PR #674

PR #837, opened 2021-03-30
add param to control CI stages

ce1adon:
This PR is prerequisite of PR #556 

I've investigated @pfultz2 's suggestion about using keyword in commits to control stage on/off. It requires setting changes on node side. I don't think we want to make any changes of the nodes' properties since this whole solution is adhoc for miopentensile. This strategy provides param control from stage level rather than step level. Open to requests of major revises.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -166,7 +166,7 @@ pipeline {
     }
     stages{
         stage("Static checks"){
-            when { expression { params.BUILD_CURRENT_STAGE } }
+            when { expression { !params.BUILD_CURRENT_STAGE } }
             parallel{
                 stage('Hip Tidy') {
                     agent{  label rocmnode("nogpu") }
@@ -233,7 +233,7 @@ pipeline {
             }
         }
         stage("Smoke Fp32"){
-            when { expression { params.BUILD_CURRENT_STAGE } }
+            when { expression { !params.BUILD_CURRENT_STAGE } }
             parallel{
                stage('Fp32 OpenCL Debug') {
                     agent{ label rocmnode("vega") }
@@ -337,7 +337,7 @@ pipeline {
             }
         }
         stage("Smoke Aux 1"){
-            when { expression { params.BUILD_CURRENT_STAGE } }
+            when { expression { !params.BUILD_CURRENT_STAGE } }
             parallel{
                 stage('Fp32 HipNoGPU Debug') {
                     agent{  label rocmnode("nogpu") }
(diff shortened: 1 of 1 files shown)


ce1adon:
Here is a sample of the skip process.
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/skip-stages/2/pipeline
which is triggered by:
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/837/commits/de6d97f80a22b16ed237dd5a6aa23baf501e53f1

ce1adon:
Complete CI and ready for review

ce1adon:
@pfultz2 @atamazov @JehandadKhan 
Any thoughts?

atamazov:
I do not see any harm in this PR, basically it is NFC. But how you going to change value of `BUILD_CURRENT_STAGE`?

I am wondering if it would be easier to use `when` with `changelog` and `changeset`, combined by means of `anyOf` so MIOpenTensile stages will be triggered automatically when certain file changed (including dockerfile) or enforced when something like `[miopentensile]` reside in the commit message.

atamazov (approved):

JehandadKhan (changes requested):
@ce1adon Shouldn't the default value be false? So that if someone wishes to explicitly run these tests they can do so, while in the default run, these stages are skipped ? Am I missing something here ?

ce1adon:
> I do not see any harm in this PR, basically it is NFC. But how you going to change value of `BUILD_CURRENT_STAGE`?
> 
> I am wondering if it would be easier to use [`when` with `changelog` and `changeset`](https://www.jenkins.io/doc/book/pipeline/syntax/#when), combined by means of `anyOf` so MIOpenTensile stages will be triggered automatically when certain file changed (including dockerfile) or enforced when something like `[miopentensile]` reside in the commit message.

Thank you Artem! Using changelog and changeset is definitely a solution. The difficulty is that we cannot predict in future which branch(es) we are going to skip the tests, neither can we predict future branch name(s). With a simple bool value it will be easier to operate.

ce1adon:
> @ce1adon Shouldn't the default value be false? So that if someone wishes to explicitly run these tests they can do so, while in the default run, these stages are skipped ? Am I missing something here ?

Thank you Jehandad! The default operation is to build current stage so it is set as true. By setting the value as false it will skip current stage.

ce1adon:
@atamazov @JehandadKhan @pfultz2 
Any more comments?

JehandadKhan (approved):

PR #841, opened 2021-04-02
Implement mlir binary backend - step 1: fwd path

jerryyin:
This PR enables the back-to-back execution of mlir binary backend fwd kernels.

I did the following in this PR:
 - Bumped the MLIR commit to latest
 - Implemented the MLIR binary backend invoker
 - Implemented the MLIR binary backend solver
   - Added `.mlir` extension for binary backend compilation
   - Amended the call chain such that both `.mlir-cpp` and `.mlir` extension are handled
 - Pulled in `memref.hpp` header from a subset of llvm source, refer to [here](https://mlir.llvm.org/doxygen/CRunnerUtils_8h_source.html)
 - Added unit test for mlir binary backend

Please note: CI in full pass with initial commit 5ea65fb.

Also this is a follow-up from @atamazov 's review in [this commit](https://github.com/ROCmSoftwarePlatform/MIOpen/commit/7988bf3accc7a9a9f9115cc47360ffff5841ecaa).

-----

Update: MLIR repository has a breaking change in group convolution support. Because of that, I wouldn't be able to address any MLIR end concerns without adding the new feature. Therefore I will work in the following sequence:
 1. Get the first round of fwd/bwd/wrw solvers implemented and submitted to MIOpen
 2. Submit https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/136
 3. Bring MLIR to the latest, support group convolution, drop `StridedMemRef`


src/mlir_build.cpp
```diff
@@ -138,11 +138,23 @@ boost::filesystem::path MiirBuildViaHip(boost::optional<TmpDir>& tmp_dir,
...
+{
+    AutoMiirHandle handle(params);
+    miirLowerBin(handle());
+
+    size_t size = 0;
+    auto status = miirBufferGet(handle(), nullptr, &size);
+    check_miir_error(status, "miirLowerTuningParams");
+    buffer.resize(size);
+    status = miirBufferGet(handle(), buffer.data(), &size);
+    check_miir_error(status, "miirLowerTuningParams");
```
jerryyin, line 157: TODO: update to  `check_miir_error(status, "miirBufferGet");`

atamazov:
Reviewers, please review ;)

test/CMakeLists.txt
```diff
@@ -306,9 +307,11 @@ if(MIOPEN_TEST_MLIR)
...
-        COMMAND ${IMPLICITGEMM_MLIR_ENV_F} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 64 128 58 58  --weights  128  128 3 3 --pads_strides_dilations 0 0 1 1 1 1
+        COMMAND ${IMPLICITGEMM_MLIR_ENV_F_CPP} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 64 1024 14 14 --weights 1024 1024 1 1 --pads_strides_dilations 0 0 1 1 1 1
+        COMMAND ${IMPLICITGEMM_MLIR_ENV_F_CPP} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 64 256 56 56  --weights  256  256 1 1 --pads_strides_dilations 0 0 1 1 1 1
+        COMMAND ${IMPLICITGEMM_MLIR_ENV_F_CPP} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 64 128 58 58  --weights  128  128 3 3 --pads_strides_dilations 0 0 1 1 1 1
+        COMMAND ${IMPLICITGEMM_MLIR_ENV_F_BIN} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 64 128 58 58  --weights  128  128 3 3 --pads_strides_dilations 0 0 1 1 1 1
+        COMMAND ${IMPLICITGEMM_MLIR_ENV_F_BIN} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 128 8 32 32  --weights  128  8 3 3 --pads_strides_dilations 0 0 1 1 1 1
```
shurale-nkn, line 314: I didn't find any limitations for pads, so please add a test with non-zero padding.

shurale-nkn (changes requested):

src/include/miopen/memref.hpp
```diff
@@ -0,0 +1,82 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#pragma once
+
+namespace miopen {
+
+// clang-format off
```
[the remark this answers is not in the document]
jerryyin: This is copy-pasted from somewhere else. I don't want MIOpen specific clang-format to mess with it and keep the source formatting. This make it easier for us to observe if there are source changes in LLVM end.

test/CMakeLists.txt
jerryyin: I'd prefer not add additional test for now because:
- MLIR is an experimental option now, it does not impact MIOpen general correctness.
- The work in MIOpen is barely pass-through to MLIR. This PR aims at just test that pass-through instead of correctness. 
-  Right now MLIR is working through a fair amount of kernel issues, it is most likely to expose issues if I'm to add more test at this specific commit
   - Even if we find a kernel crashing because of MLIR, there's no way from me to address it from MIOpen end. However, I don't think this should block the pass-through logic going into MIOpen
   - Once MLIR work through all its kernel generation issues, I can just bump the commit id for MIOpen to adopt the fix.

jerryyin: But I agree that if we ever decide to drop `-MIOPEN_USE_MLIR` and make it the default compilation method, we should add more test coverage to make sure it can handle all situations.

shurale-nkn: Oh, ok, I forget that this is experimental.

But the problem is that it still can block MIOpen CI.
Example: 
As i understand MLIR is working only for convolution, and for now it is applicable for a wide range of convolutions.

MIOpen CI tests MLIR in stage('Fp32 Hip MLIR') which includes several convolution tests without MIOPEN_DEBUG_FIND_ONLY_SOLVER, and this creates conditions for MLIR execution in one of these tests.

shurale-nkn: So we can reduce number of test to test for stage('Fp32 Hip MLIR'), to test  only MLIR and only for required cases.
Or to change IsApplicable range.

shurale-nkn: For me, the first is better.

shurale-nkn (changes requested):

test/CMakeLists.txt
shurale-nkn: I can't use github suggestion for this, but:
At line:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/6128704ed8a325474868a42d8853139f0defa971/test/CMakeLists.txt#L91

write:
```
if(MIOPEN_TEST_MLIR)
    set(SKIP_ALL_EXCEPT_TESTS test_conv_igemm_mlir)
endif()
```
.
And line:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/6128704ed8a325474868a42d8853139f0defa971/test/CMakeLists.txt#L235
Change to 
```
        if( 
            ( NOT PARSE_SKIP_UNLESS_ALL OR MIOPEN_TEST_ALL)
            AND
            (NOT MIOPEN_TEST_MLIR OR (NAME IN_LIST SKIP_ALL_EXCEPT_TESTS))
        )
```

jerryyin: @shurale-nkn I see what you mean. I think a way to address this is to update the Jenkins file such that instead of `make check` replace it with `make test_conv_igemm_mlir`?

From a different perspective though. I feel like knowing that MLIR build not only MLIR path work can be beneficial from the test coverage perspective? After all, once I shrink the test coverage to MLIR build only, MIOpen lose track of whether everything else work with MLIR build.

On other thought I have, regarding about handling the possibility of flakiness in the `Fp32 Hip MLIR` stage, is to handle it from jenkinsfile:

```jenkinsfile
stage('Fp32 Hip MLIR') {
   try {
     buildHipClangJob(...)
   } catch(error) {
     echo "MLIR build flakiness observed, retrying"
     retry(2) {
        buildHipClangJob(...)
     }
   } finally {
      reboot()
   }
}
```

shurale-nkn: @jerryyin make test_conv_igemm_mlir is also posible way.

`"From a different perspective though. I feel like knowing that MLIR build not only MLIR path work can be beneficial from the test coverage perspective?"`

In that case we came to MIOpen CI blocker. The problem is, we don't know when it might show up. Better to use only tested solutions. Otherwise, someone will have to solve it later.
Convolution is pretty much isolated from another parts of MIOpen.

`"On other thought I have, regarding about handling the possibility of flakiness in the Fp32 Hip MLIR stage, is to handle it from jenkinsfile:"`
If retry (2) contains the same tests, we will have problems with correctness, which appear with some probability and we will have to restart the stage sometimes. In one form or another, we have already been through this, and it does not lead to anything good. Who know how many restarts it take.
If Test marked as failed and CI should ignore this, what's the point of this test?

Better to do full testing when MLIR will be ready, with true full applicable.
Because this is MIOpen CI :) to test something no need to merge it to develop :)

jerryyin: @shurale-nkn Make sense, better safe than sorry is a better approach here. Also, agreed that jenkins retry is not a very elegant solution.

I appreciate your very detailed instructions in updating the test cmakefiles. I decided not to do that just so that I myself can do `make check` in my dev process to run through the whole set of tests and make sure everything is as expected.

shurale-nkn (approved):
LGTM

test/CMakeLists.txt
shurale-nkn: If MLIR needs some kind of periodic testing in the future, we can create a separate work on Jenkins.

atamazov:
[Off-topic] @jerryyin Please try to not forget to terminate outdated CI job after pushing to the repo. Thanks!

test/CMakeLists.txt
jerryyin: @shurale-nkn Thanks for reminding. MLIR already has a number of integration/nightly testing. We are in that "bug fixing" stage for outstanding issues. Hopefully once those issues are caught, I can add more tests in MIOpen end as well.

jerryyin:
@atamazov Sounds good, will keep a closer eye on delinquent CI jobs.

atamazov (changes requested):
It seems like we should not borrow `StridedMemRefType<>` from MLIR. Conceptually, it represents the layout of MLIR kernel parameters. IIUC this layout may change in the future, so the relevant abstraction should be a part of the Miir API. Otherwise we will have a situation when some part of Miir implementation is penetrated into MIOpen.

Luckily we do not need much. The only thing we need is a struct type (POD type equivalent to the data members of `StridedMemRefType<void, 4>`). You can define it in the Miir.h, thus isolating MIOpen from the specific layout of MLIR kernel parameters.

When/if the implementation of `StridedMemRefType<>` in MLIR change, you would need to update that struct type in Miir.h.

Sounds reasonable?

src/include/miopen/memref.hpp
```diff
@@ -0,0 +1,82 @@
...
+// Codegen-compatible structures for StridedMemRef type.
+//===----------------------------------------------------------------------===//
+/// StridedMemRef descriptor type with static rank.
+template <typename T, int N>
+struct StridedMemRefType {
+  T* basePtr;
+  T* data;
+  int64_t offset;
+  int64_t sizes[N];
+  int64_t strides[N];
```
jerryyin, line 49: > When/if the implementation of StridedMemRefType<> in MLIR change, you would need to update that struct type in Miir.h.

Would you be pulling in this type somewhere else into MIOpen, like an existing PR [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/631)?

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,134 @@
...
+    std::make_tuple(out_dims, out_strides) = permuteDimsStrides(out.GetLengths(), out.GetStrides());
+}
+
+MlirConvArgs MakeMlirConvArgs(ConstData_t in,
+                              const std::vector<size_t>& in_dims,
+                              const std::vector<size_t>& in_strides,
+                              ConstData_t w,
+                              const std::vector<size_t>& weights_dims,
+                              const std::vector<size_t>& weights_strides,
+                              ConstData_t out,
```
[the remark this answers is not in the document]
jerryyin: I think this (including the const_cast) is going to be conceptually more correct but will cost me a lot of typing :-p

If I'm to make the filter and input as const, the only way I am able to initialize it is through brace-enclosed list. That means I would not be able to use the existing lambda to do the heavy lifting for me but manually type cast and construct several arrays in place.

Will correct in the next commit.

src/include/miopen/memref.hpp
jerryyin: > you would need to update that struct type in `Miir.h`.

I agree about exporting that in the library side. I guess when you later decided that MIOpen also need to pull it in, it can be done in a different namespace.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+    auto cpyToMemRef = [](const void* ptr,
+                          const std::vector<size_t>& dims,
+                          const std::vector<size_t>& strides,
+                          MemRef4DGeneric& target) {
+        target.basePtr = const_cast<void*>(ptr); // NOLINT (cppcoreguidelines-pro-type-const-cast)
+        target.data    = const_cast<void*>(ptr); // NOLINT (cppcoreguidelines-pro-type-const-cast)
+        target.offset  = 0;
+        std::copy(dims.cbegin(), dims.cend(), &target.sizes[0]);
+        std::copy(strides.cbegin(), strides.cend(), &target.strides[0]);
```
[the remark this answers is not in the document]
jerryyin: This can be addressed, but with some more inconvenience here. It will be some more code typing in constructing those const data strictures. (Me personally prefer the less typing, existing approach). This should become more clear in the next commit.
 - From previous comments, I need to define input and filter into const types.
 - From this one, I need to be able to update the pointer fields, so that means:
   - Outside the factory, I construct temporary non-const input and filter
   - Inside the factory, I construct const input and filter, manually

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,134 @@
...
+                              ConstData_t out,
+                              const std::vector<size_t>& out_dims,
+                              const std::vector<size_t>& out_strides)
+{
+    auto cpyToMemRef = [](const void* ptr,
+                          const std::vector<size_t>& dims,
+                          const std::vector<size_t>& strides,
+                          MemRef4DGeneric& target) {
+        target.basePtr = const_cast<void*>(ptr); // NOLINT (cppcoreguidelines-pro-type-const-cast)
+        target.data    = const_cast<void*>(ptr); // NOLINT (cppcoreguidelines-pro-type-const-cast)
```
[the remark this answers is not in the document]
jerryyin: I can understand `const_cast` is bad in general, though at this specific case it is harmless.

Will address this in next commit.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -104,20 +103,30 @@ InvokerFactory MakeMlirFwdInvokerFactory(const ConvolutionContext& ctx)
...
+            args.filter.basePtr =
+                const_cast<void*>(tensors.w); // NOLINT (cppcoreguidelines-pro-type-const-cast)
+            args.filter.data =
+                const_cast<void*>(tensors.w); // NOLINT (cppcoreguidelines-pro-type-const-cast)
+
+            // Input
+            args.input.basePtr =
+                const_cast<void*>(tensors.in); // NOLINT (cppcoreguidelines-pro-type-const-cast)
+            args.input.data =
+                const_cast<void*>(tensors.in); // NOLINT (cppcoreguidelines-pro-type-const-cast)
```
jerryyin, line 124: @atamazov I'm afraid const_cast is unavoidable because:
 - The definition of `StridedMemRef` mark T not as a const pointer
   - I don't see a benefit to expand the `StridedMemRef` with a const model. 
 - MIOpen's DataInvokeParams mark input and filter as `ConstData_t`

src/include/miopen/memref.hpp
jerryyin: @atamazov One drawback is that if we are to export this in the C header, I will have to rewrite it in C format since this is template based. But on the other hand, I really want to avoid touching this because it make any changes in this type easy to observe in MLIR source.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: It help me to drop two `const_cast` for output tensor's pointer assignment.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
pfultz2: I agree with @atamazov, `const_cast` is a bad idea.

src/include/miopen/memref.hpp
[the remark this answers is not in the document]
pfultz2: If this is copied from MLIR repo, then you should copy the unit tests for this as well.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: To me, it is just not cost effective to add a const to duplicate the data structure that the device interface doesn't even care about.

Refer to [here](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/136/files). 

For every `MemRef4D`, I will have to duplicate the exported type for a const and regular pointer type just to get rid of this `const_cast` here.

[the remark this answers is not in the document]
jerryyin: BTW @pfultz2 In the latest commit 29d5475, I have to make the `const_cast` used in combination with `reinterpret_cast`. Both are generally bad ideas in host code, but device function just only need a pointer in that offset.

pfultz2:
> It seems like we should not borrow StridedMemRefType<> from MLIR. Conceptually, it represents the layout of MLIR kernel parameters. IIUC this layout may change in the future, so the relevant abstraction should be a part of the Miir API. Otherwise we will have a situation when some part of Miir implementation is penetrated into MIOpen.

I agree, this creates a lot of unnecessary coupling between MIOpen and MLIR.

> Luckily we do not need much. The only thing we need is a struct type (POD type equivalent to the data members of StridedMemRefType<void, 4>). You can define it in the Miir.h, thus isolating MIOpen from the specific layout of MLIR kernel parameters.

That's right. It doesn't even need to be templated as it can be just passed as `void*` pointers. 

Also, if this will change in the future, it would be even better to use an opaque pointer rather than a `struct`.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -104,24 +104,24 @@ InvokerFactory MakeMlirFwdInvokerFactory(const ConvolutionContext& ctx)
...
             const auto& data_ctx = primitive_parameters.CastTo<conv::DataInvokeParams>();
             const auto& tensors  = data_ctx.tensors;
 
             // Filter
-            args.filter.basePtr =
-                const_cast<void*>(tensors.w); // NOLINT (cppcoreguidelines-pro-type-const-cast)
-            args.filter.data =
-                const_cast<void*>(tensors.w); // NOLINT (cppcoreguidelines-pro-type-const-cast)
+            // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
+            args.filter.basePtr = const_cast<void*>(reinterpret_cast<const void*>(tensors.w));
```
pfultz2, line 116: You can't reinterpret_cast a `cl_mem` object to a `void*` to get a pointer. You need to use [`clGetMemObjectInfo`](https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clGetMemObjectInfo.html).

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
pfultz2: > For every MemRef4D, I will have to duplicate the exported type for a const and regular pointer type just to get rid of this const_cast here.

You should just have two different function to set the pointer as either `const` or not, and then the `const_cast` can happen in llvm if it necessary. This really should not be pushed on to clients.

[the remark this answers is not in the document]
jerryyin: @pfultz2 LLVM only populate a device code object. `const` is a C/C++ semantic but the same concept does not exist in device or LLVM level code. In other words, if the MLIR wise populated device code doesn't care about const, and doesn't even have this abstraction, it isn't fair for the library to force this `const` as well.

jerryyin:
> That's right. It doesn't even need to be templated as it can be just passed as void* pointers.

@pfultz2 I just add you as the reviewer [here](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/136).

> Also, if this will change in the future, it would be even better to use an opaque pointer rather than a struct.

This will make the interface cleaner but after all the client still depends on the struct to interpret the data member in it.

src/conv/invokers/mlir_impl_gemm.cpp
jerryyin: Thanks for figuring this out. Will do the below to address this. Let me know if I understand this right.

```cpp
#if MIOPEN_BACKEND_OPENCL
  void* p;
  clGetMemObjectInfo(tensors.w, CL_MEM_HOST_PTR, sizeof(p), &p, NULL);
  args.filter.basePtr = p;
#else
  args.filter.basePtr = const_cast<void*>(tensors.w);
#endif
```

src/mlir_build.cpp
```diff
@@ -138,11 +138,23 @@ boost::filesystem::path MiirBuildViaHip(boost::optional<TmpDir>& tmp_dir,
...
-    miirLowerInit();
     auto status = miirLowerTuningParams(handle());
     check_miir_error(status, "miirLowerTuningParams");
     miirGetExecutionDims(handle(), &global_size, &local_size);
     check_miir_error(status, "miirGetExecutionDims");
 }
 
+void MiirGenBin(const std::string& params, std::vector<char>& buffer)
+{
+    AutoMiirHandle handle(params);
```
pfultz2, line 149: Not related to this PR, but `AutoMiirHandle` can be copied which will lead to double frees. This really should just use `MIOPEN_MANAGE_PTR`, which will create a `std::unique_ptr`(which is movable and not copyable) to manage the opaque object. Perhaps we can should open issue to address this.

src/conv/invokers/mlir_impl_gemm.cpp
atamazov: Just in case: MIOpen uses ConstData_t for input tensors (buffers) and Data_t for output buffer. This is not important when you build MIOpen with HIP backend, but important for OpenCL backend.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
atamazov: > To me, it is just not cost effective to add a const to duplicate the data structure that the device interface doesn't even care about.

This is important for OpenCL backend.

> const is a C/C++ semantic but the same concept does not exist in device or LLVM level code. 

[off-topic] Of course this semantic is important for device code, and in llvm too. 

> it isn't fair for the library to force this const as well.

[off-topic] But we didn’t promise _fair play_ :smile:

src/conv/invokers/mlir_impl_gemm.cpp
pfultz2: Yes that looks correct.

src/mlir_build.cpp
jerryyin: @pfultz2 Thanks for observing this! I agree that making it like a `unique_ptr` to turn copy into compile time error is a more robust approach. Let me open the issue and leave another comment here.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: Will address according to @pfultz2 [comment in this PR](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/136). It looks like a much more robust approach compared to exposing the structure in the interface. With that, I can create two interface function that help to address MIOpen's concern here.

Notice that both function return the similar `MiirStridedMemRef` because const-ness does not impact data layout when passing arguments to device.

```cpp
MiirStridedMemRef MiirCreateStridedMemRef(void* data,
                                          int64_t offset,
                                          int64_t * sizes,
                                          int64_t * strides,
                                          int64_t ndims); 
{
    return new StridedMemRefType<void*, ndims>(data, data, offset, sizes, strides);
}

MiirStridedMemRef MiirCreateConstStridedMemRef(const void* data,
                                               int64_t offset,
                                               int64_t * sizes,
                                               int64_t * strides,
                                               int64_t ndims); 
{
    return new StridedMemRefType<const void*, ndims>(data, data, offset, sizes, strides);
}
```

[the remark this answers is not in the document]
atamazov: @jerryyin These functions create `new` objects. Who will `delete` them?

src/mlir_build.cpp
atamazov: The programmer who copies guard objects should be killed.

test/CMakeLists.txt
atamazov: @shurale-nkn @jerryyin 
> In that case we came to MIOpen CI blocker.

If this happens, we will disable MLIR testing stage in the Jenkinsfile until the problem is fixed. Let's KISS.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: @atamazov Paul already give a solution in that comments:

`void MiirDestroyStridedMemRef(MiirStridedMemRef x);`

We can create a shared ptr (with this custom deleter) to do this. Is that a infrastrcture already exist in MIOpen, like `MIOPEN_MANAGE_PTR`, but for `shared_ptr`?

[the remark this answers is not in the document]
atamazov: See https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/136#issuecomment-820661976

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
atamazov: @jerryyin Whatever design you are going to select, it should be able to satisfy [this](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/841#discussion_r611974707) and [this](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/841#discussion_r611971489). This is the simplest test of design validity.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,153 @@
...
+    MemRef4DGeneric filter;
+    MemRef4DGeneric input;
+    MemRef4DGeneric output;
+};
+
+// Rearrange strides correctly
+// In MLIR: the layout, sizes and strides are coherent. The layout information is not
+// embedded into the permutation of strides.
+// - For NCHW, sizes = {N, C, H, W}; strides = {C*H*W, H*W, W, 1}
+// - For NHWC, sizes = {N, H, W, C}; strides = {C*H*W, W*C, C, 1}
```
pfultz2, line 28: If MLIR doesn't use the strides to describe the layout of the tensor, then should we even pass the strides at all to MLIR?

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
pfultz2: > Is that a infrastrcture already exist in MIOpen, like MIOPEN_MANAGE_PTR, but for shared_ptr?

You can construct it first as a unique_ptr using `MIOPEN_MANAGE_PTR` then you can pass it to a shared_ptr. A shared_ptr uses a type erased deleter so the type of a shared_ptr doesn't change based on the deleter. I suggest you write a simple function to construct it as a shared_ptr, so this can all be done in one place.

src/conv/invokers/mlir_impl_gemm.cpp
pfultz2: >  This is not important when you build MIOpen with HIP backend, but important for OpenCL backend.

The const only applies to HIP. In OpenCL, `void*` is a `cl_mem` object so the extra `const` only applies to the pointer and not the data.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: After I figured that MLIR side cannot change the device ABI, I decide to not pursue further with this opaque pointer idea. Sorry for the confusions I may have caused on this thread.

With the definition of the structure on the interface, I'd still prefer to keep the `const_cast`on MIOpen end.

src/conv/invokers/mlir_impl_gemm.cpp
jerryyin: I have experimented to test with strides of all zeros and confirm that it can run well, so yes, that confirms strides are not used for now.

I do this mostly for robustness:
 - I can't guarantee future development of MLIR will not use this field
 - All MLIR local testing have the complete memref testing. In other words, incomplete type is not track in CI.

src/include/miopen/memref.hpp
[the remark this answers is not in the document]
jerryyin: @pfultz2 I have updated the PR description. Will drop `memref.hpp` after 3 PRs.

src/conv/invokers/mlir_impl_gemm.cpp
atamazov: Oops, correct, I've messed up things :facepalm:

src/mlir_build.cpp
atamazov: [Recommendation] If in doubt, just disable copying by
```C++
  AutoMiirHandle(const AutoMiirHandle&) = delete;
  void operator=(const AutoMiirHandle&) = delete;
```
in https://github.com/ROCmSoftwarePlatform/MIOpen/blob/8fe06e67a16b7261d59d70a0ae75d83e678ad788/src/mlir_build.cpp#L40-L48

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -106,7 +103,7 @@ InvokerFactory MakeMlirFwdInvokerFactory(const ConvolutionContext& ctx)
     MlirConvArgs args =
         makeMlirConvArgs(in_dims, in_strides, weights_dims, weights_strides, out_dims, out_strides);
 
-    return [=](const std::vector<Kernel>& kernels) {
+    return [=](const std::vector<Kernel>& kernels) mutable {
```
atamazov, line 106: Could you please clarify why g++ wants `mutable`. If this is a g++ specific quirk, let's make it so:
```C++
/// Workaround for a g++ specific quirk. When the issue is fixed in g++, we can insert here something like
/// "&& __GNUC__ >7 && __GNUC_MINOR__ < 6" (just an example).
#if defined(__GNUC__)
#define GNUC_MUTABLE mutable
#else
#define GNUC_MUTABLE
#endif
...
        return [=](const std::vector<Kernel>& kernels) GNUC_MUTABLE {
```

atamazov (changes requested):

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
atamazov: Let's postpone this. #862 created.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
atamazov: Let's postpone this. #862 created.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
```diff
@@ -0,0 +1,160 @@
...
+std::tuple<int, int, int>
+ConvHipImplicitGemmMlirBinFwd::CalculateGemmSize(const ConvolutionContext& ctx)
+{
+    const auto n  = ConvolutionContextInterpreter::GetBatchN(ctx);
+    const auto k  = ConvolutionContextInterpreter::GetOutputChannelK(ctx);
+    const auto c  = ConvolutionContextInterpreter::GetInputChannelC(ctx);
+    const auto ho = ConvolutionContextInterpreter::GetOutputHeightHo(ctx);
+    const auto wo = ConvolutionContextInterpreter::GetOutputWidthWo(ctx);
+    const auto y  = ConvolutionContextInterpreter::GetFilterHeightY(ctx);
+    const auto x  = ConvolutionContextInterpreter::GetFilterWidthX(ctx);
```
atamazov, line 46: The functions return `int` IIRC. Make all of them `size_t` to avoid possibility of overflow.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <miopen/solver/implicitgemm_util.hpp>
+#include <miopen/handle.hpp>
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_HIP_IMPLICIT_GEMM_MLIR_BIN_FWD)
+
+namespace miopen {
+namespace solver {
+
+std::tuple<int, int, int>
+ConvHipImplicitGemmMlirBinFwd::CalculateGemmSize(const ConvolutionContext& ctx)
```
atamazov, line 38: This function should be local in module, not a class member.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
```diff
@@ -0,0 +1,160 @@
...
+    ConvSolution result;
+    KernelInfo construction_parameters;
+
+    std::string version   = "_v4r4";
+    std::string direction = "_fwd";
+    std::string operation = "conv2d";
+
+    construction_parameters.kernel_file = "mlir_gen_igemm_conv2d" + version + direction + ".mlir";
+
+    construction_parameters.kernel_name = "mlir_gen_igemm_conv2d" + version + direction;
```
atamazov, line 105: [Resommendation]
```suggestion
    construction_parameters.kernel_name = "mlir_gen_igemm_conv2d" + version + direction;
    construction_parameters.kernel_file = construction_parameters.kernel_name + ".mlir";
```

atamazov (changes requested):
Lower-level review

Jenkinsfile
```diff
@@ -516,7 +516,7 @@ pipeline {
                             ulimit -c unlimited
                             cd build
                             CXX=/opt/rocm/llvm/bin/clang++ cmake -DMIOPEN_USE_MLIR=On -DBUILD_DEV=On -DCMAKE_BUILD_TYPE=release -DMIOPEN_TEST_FLAGS='--verbose --disable-verification-cache' ..
-                            CTEST_PARALLEL_LEVEL=4 MIOPEN_LOG_LEVEL=5 make -j\$(nproc) check
+                            CTEST_PARALLEL_LEVEL=4 MIOPEN_LOG_LEVEL=5 make -j\$(nproc) test_conv_igemm_mlir
```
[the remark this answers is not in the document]
atamazov: WRT "what if this experimental stuff will break our CI?" -- if MLIR gets broken, then CI must fail, even if MLIR support is yet experimental. "Experimental" simply means that we do not guarantee the users anything, but in all other aspects this is identical to the non-experimental stuff. 

Obviously, MLIR-related PRs should not break MLIR. Likewise, all other PR should not break MLIR, otherwise their PRs should fail on CI.

/cc @shurale-nkn

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,150 @@
...
+    MemRef4DGeneric input{nullptr, nullptr, 0, {0, 0, 0, 0}, {0, 0, 0, 0}};
+    initDimStrides(in_dims, in_strides, input);
+    MemRef4DGeneric output{nullptr, nullptr, 0, {0, 0, 0, 0}, {0, 0, 0, 0}};
+    initDimStrides(out_dims, out_strides, output);
+
+    return {filter, input, output};
+}
+} // Anonymous namespace
+
+InvokerFactory MakeMlirFwdInvokerFactory(const ConvolutionContext& ctx)
```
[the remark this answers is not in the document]
jerryyin: My plan for now is to use it in the MLIR bin fwd solver only. I am mostly following `impl_gemm.hpp/cpp` here. Would you rather this to be in the solver? I find encapsulating invoker related stuff in its own component cleaner.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,150 @@
...
+            clGetMemObjectInfo(tensors.in, CL_MEM_HOST_PTR, sizeof(input), &input, nullptr);
+            clGetMemObjectInfo(tensors.out, CL_MEM_HOST_PTR, sizeof(output), &output, nullptr);
+#elif MIOPEN_BACKEND_HIP
+            // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
+            filter = const_cast<void*>(tensors.w);
+            // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
+            input = const_cast<void*>(tensors.in);
+            // NOLINTNEXTLINE (cppcoreguidelines-pro-type-const-cast)
+            output = const_cast<void*>(tensors.out);
+#endif
```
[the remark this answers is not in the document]
jerryyin: Correct. I had a separate ticket to track this from MLIR end as well. My plan is to work through them in the following sequence:

Before June:
 - Get all solver/invokers into MIOpen
 - Move the `StridedMemRef` in MLIR exported header

After June:
 - Figure out the best encapsulation for kernel args
 - Implement the most sensible kernel signature

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,150 @@
...
+            args.output.data    = output;
+
+            handle.Run(kernels[0])(args);
+            if(handle.IsProfilingEnabled())
+            {
+                float elapsed = 0;
+                elapsed += handle.GetKernelTime();
+                handle.ResetKernelTime();
+                handle.AccumKernelTime(elapsed);
+            }
```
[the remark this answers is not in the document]
jerryyin: Will remove. Thanks for reminding. Just for understanding: By having this code snippet, am I double counting the kernel time here?

src/hip/handlehip.cpp
```diff
@@ -391,7 +391,7 @@ Program Handle::LoadProgram(const std::string& program_name,
 {
     this->impl->set_ctx();
 
-    if(!miopen::EndsWith(program_name, ".mlir-cpp"))
+    if((!miopen::EndsWith(program_name, ".mlir-cpp")) && (!miopen::EndsWith(program_name, ".mlir")))
```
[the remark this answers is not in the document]
jerryyin: Is there anything I need to address here?

src/mlir_build.cpp
```diff
@@ -138,11 +138,23 @@ boost::filesystem::path MiirBuildViaHip(boost::optional<TmpDir>& tmp_dir,
...
+    miirLowerBin(handle());
+
+    size_t size = 0;
+    auto status = miirBufferGet(handle(), nullptr, &size);
+    check_miir_error(status, "miirLowerTuningParams");
```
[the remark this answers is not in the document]
jerryyin: Oooops

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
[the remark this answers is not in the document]
jerryyin: Will remove HIP in naming in a separate PR.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
```diff
@@ -0,0 +1,160 @@
...
+    return std::make_tuple(gemm_m, gemm_n, gemm_k);
+}
+
+bool ConvHipImplicitGemmMlirBinFwd::IsApplicable(const ConvolutionContext& ctx) const
+{
+#if MIOPEN_USE_MLIR
+    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_HIP_IMPLICIT_GEMM_MLIR_BIN_FWD{}))
+        return false;
+    // Future: MLIR-binary solutions do not take long time to build
+    if(ctx.skip_solutions_that_take_long_time_to_build_and_have_narrow_coverage)
```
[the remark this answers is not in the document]
jerryyin: It is fast, correct. I might have a misunderstanding in this flag. I thought this was used to tell the tuning infrastructure that we disallow tuning for this solver. If it is not the case, I will remove this conditional.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
jerryyin: Almost every solver has the `CalculateGemmSize()` function as a member, why an exception in igemm mlir solver?

atamazov:
@JehandadKhan #795 merged in. Please make necessary changes right in this PR to resolve https://github.com/ROCmSoftwarePlatform/MIOpen/pull/795#issuecomment-821889442. If you do not have time, please ask me and I will do that.

@jerryyin Plase let us know if any objections.

atamazov:
@jerryyin Merge conflicts. Please let me know if you need help.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
atamazov: No, this is not  bug, just useless code.

src/hip/handlehip.cpp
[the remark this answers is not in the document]
JehandadKhan: @jerryyin  No, this was for me. I have updated the code in the appropriate file and updated your branch.

JehandadKhan:
> @JehandadKhan #795 merged in. Please make necessary changes right in this PR to resolve [#795 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/795#issuecomment-821889442). If you do not have time, please ask me and I will do that.

@atamazov I have updated the branch.

src/conv/invokers/mlir_impl_gemm.cpp
jerryyin: @atamazov As far as I can tell this is not a quirk and clang should complain too.

A few lines below I have:

```cpp
args.filter.basePtr = filter;
args.filter.data = filter;
...
```

lambda pass-by value are implicitly `const`. In this case, the captured `args` should be a read-only variable. I need to add this mutable just so that it is allowed to update the captured `args`.

src/hip/handlehip.cpp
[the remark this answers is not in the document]
jerryyin: Thanks guys. I'm marking this as resolved now.

Jenkinsfile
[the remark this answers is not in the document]
jerryyin: Addressed in 8fc3457

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,150 @@
+#include <miopen/conv/invokers/mlir_impl_gemm.hpp>
```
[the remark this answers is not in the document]
jerryyin: Addressed in 8fc3457

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -0,0 +1,150 @@
...
+                            weights_strides,
+                            out_dims,
+                            out_strides);
+
+    MlirConvArgs args =
+        makeMlirConvArgs(in_dims, in_strides, weights_dims, weights_strides, out_dims, out_strides);
+
+    return [=](const std::vector<Kernel>& kernels) mutable {
+        return [=](const Handle& handle, const AnyInvokeParams& primitive_parameters) mutable {
+            const auto& data_ctx = primitive_parameters.CastTo<conv::DataInvokeParams>();
```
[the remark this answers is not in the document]
jerryyin: Addressed in 8fc3457

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: Addressed in 8fc3457

src/mlir_build.cpp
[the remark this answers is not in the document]
jerryyin: Addressed in 8fc3457

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
jerryyin: Addressed in 8fc3457

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
jerryyin: Addressed in 8fc3457

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: Addressed in 8fc3457

src/mlir_build.cpp
jerryyin: Addressed in 8fc3457

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
atamazov: This is discrimination! :smile: 

Seriously, many existing igemm solvers have this kind of a design flaw, #867 created. Let's avoid replicating it.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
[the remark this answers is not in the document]
atamazov: Please remove. This is related to the [FAST_HYBRID](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/find_and_immediate.md#find-modes) find mode which I am going to get rid of anyway.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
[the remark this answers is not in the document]
atamazov: Ok

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
atamazov: [Recommendation]
> Would you rather this to be in the solver? 

Yes, this is correct. 

> I find encapsulating invoker related stuff in its own component cleaner.

Well, it can be considered as a _sub-component_ of this solver. And yes, at first glance, it looks like this makes the code cleaner. But this complicates the code at the same time: When you see a global object you begin asking yourself where it is used. If you see only one use case, you begin asking yourself, why it is used only once. And you decide that there is intent to re-use it somewhere else, later, but this is not true ;)

So I recommend moving this code into the Solver.

[Side note] Defining lambdas in the `GetSolution()` (instead of a function like `MakeInvokerFactory()`) can make the code simpler, because values of all the objects that reside in `GetSolution()` are available for immediate use in the Factory and in the Invoker. Perhaps this is not important for here, just in case.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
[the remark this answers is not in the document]
atamazov: [Side note] Note that renaming of Solvers affects the `perf-db` and `find-db`. Right now this is not important, but please keep in mind for the future, thanks.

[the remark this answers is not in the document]
atamazov: #869 opened

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: Thanks for the explanation. I'd still prefer invoker to be in its own components because of code re-use possibilities:

I don't plan on this global function to be used in any other solvers. However, between different MLIR solvers, there's a decent chance of code reuse in different direction of invokers: All fwd/bwd/wrw invokers will be using the exact similar format of arguments, regardless of the direction. Besides, the process of deriving the strides and dimensions to pass to MLIR are the same as well in all direction of invokers. 

If I don't put them all in the same compilation unit and instead put them in independent solvers. I will have to copy-paste all above shared components in different solvers. It goes against our original wish to scope code usage, and makes future development and update error-prone as well.

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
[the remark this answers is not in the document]
jerryyin: Done in 113fab9

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: > Are you going to use this function in several MLIR Solvers?

Now thinking further on it, my answer would be yes. I will have to come up with new xdlops solver for running on gfx908 and gfx90a machines. This function will be re-used by xdlops fwd solver.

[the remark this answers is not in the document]
atamazov: Ok

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
atamazov: No, please fix here. `CalculateGemmSize()` should not be a part of the public Solver's interface.

atamazov:
One step to the approval left: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/841#discussion_r615304392

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
jerryyin: Done in 8821782

src/conv/invokers/mlir_impl_gemm.cpp
atamazov: True, thanks for explanation.

atamazov (dismissed):
LGTM!

src/solver/conv_hip_implicit_gemm_mlir_bin_fwd.cpp
```diff
@@ -29,6 +29,11 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/handle.hpp>
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-function"
+#endif
```
atamazov, line 35: Why you need to suppress this? Bug in clang-tidy?

atamazov: Perhaps `static` instead of unnamed namespace can be used as a workaround?

jerryyin: I don't blame clang-tidy for this. It's basically because `CalculateGemmSize()` is only used in maro-ed section of the code in `IsApplicable()`. Therefore, tidy target cannot see its call-site, and complained that there's a function being defined but never used.

jerryyin: In other words, we don't build `clang-tidy` or `opencl-tidy` target with `-MLIR_USE_MIOPEN` on caused it. But this is all intended.

atamazov: Aha. Just enclose the function into #if... #endif

atamazov: And please remove suppression.

jerryyin: Done in 57a7189.

atamazov (approved):
LGTM

PR #842, opened 2021-04-02
[Nv21][HOTFIX][TargetId] Support features disabled on target (SWDEV-278900)

atamazov:
This should fix http://ontrack-internal.amd.com/browse/SWDEV-278900 and unblock mainline. The problem is caused by recent changes in the compiler. Now the compiler does not allow you to disable features that do not exist on the target device.

For example, if a target does not have the XNACK feature (like Navi 21), then you can't explicitly tell the compiler to not generate extra code for XNACK. The compiler is smarter than you, it decides itself and beat your a** if you try to suggest it to do the same.


atamazov:
This is urgent PR required to unblock ROC-Master.

atamazov:
Reviewers, I need at least one approval ASAP.

asroy (approved):

Issue #835, opened 2021-03-30
In practical use, we should first ensure the completion of the task


zpwenjh:
![image](https://user-images.githubusercontent.com/63989327/112943095-9dca4200-9163-11eb-979c-6bbd46774737.png)
This task with search on has been running for a week, and then, there is some errors occured. According to the message and source code, we can know the last error happened in `hip_build_utils.cpp:HipBuildImpl` function while compiling kernel with default solution. Before this error, there is some kernels available and are able to complete the task, but miopen doesn't catch the exception. It's annoying to see that after weeks running. I'd rather MIOpen output error logs to help me determine if I need to run again, rather than stop directly.
I think we should first ensure the completion of the task, it's user-friendly. I checked the source code, there is no effective means to ensure that.

atamazov:
@zpwenjh 
> This task with search on has been running for a week...

Please describe the process, and I'll try to advise how to accelerate this. Am I right that you've tried `MIOPEN_FIND_ENFORCE=4`?

> ...there is some errors occurred. 

Due to too many files open, right?

atamazov:
@zpwenjh Ping --
> Please describe the process, and I'll try to advise how to accelerate this.

zpwenjh:
> @zpwenjh
> 
> > This task with search on has been running for a week...
> 
> Please describe the process, and I'll try to advise how to accelerate this. Am I right that you've tried `MIOPEN_FIND_ENFORCE=4`?
> 
> > ...there is some errors occurred.
> 
> Due to too many files open, right?

@atamazov Yes, I am using `MIOPEN_FIND_ENFORCE=4`, but the error messege is not same as "too many open files". At first, I thought there was not enough hard disk space, but it's not this situation.

atamazov:
> > @zpwenjh
> > > This task with search on has been running for a week...

I suspect that the reason is that `MIOPEN_FIND_ENFORCE=4` ___always___ performs tuning. Even if some config has been tuned 5 minutes ago. Therefore I recommend using `MIOPEN_FIND_ENFORCE=3`. In this mode, the tuning will be performed only once. 

However, as you can see [in the doc](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/perfdatabase.md#miopen_find_enforce),
> _If optimized values already reside in PerfDb, then auto-tune will not be performed._

[...]

PR #850, opened 2021-04-07
[Jenkinsfile] Don't disable inline asm for non-XDLOPs targets. Disable Codecov.

atamazov:
- Removes definition of `MIOPEN_DEBUG_IMPLICIT_GEMM_NON_XDLOPS_INLINE_ASM` that is originated from https://github.com/AMDComputeLibraries/MLOpen/pull/2491. Not needed anymore (and harmful -- the library should decide about this automatically).
- W/A for #853.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -296,7 +296,7 @@ pipeline {
                             ulimit -c unlimited
                             cd build
                             CXX=/opt/rocm/llvm/bin/clang++ cmake -DBUILD_DEV=On -DCMAKE_BUILD_TYPE=debug -DMIOPEN_TEST_FLAGS=--disable-verification-cache ..
-                            CTEST_PARALLEL_LEVEL=4 MIOPEN_DEBUG_IMPLICIT_GEMM_NON_XDLOPS_INLINE_ASM=0 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0 make -j\$(nproc) check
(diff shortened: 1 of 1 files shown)


atamazov:
Let's fix #835 and then switch Codecov back. Right now it is irritating.

DrizztDoUrden (dismissed):

shurale-nkn (approved):

Jenkinsfile
```diff
@@ -653,12 +653,12 @@ pipeline {
...
             when { expression { params.BUILD_CURRENT_STAGE } }
             parallel{
-                stage('Fp32 OpenCL Debug + Codecov') {
+                stage('Fp32 OpenCL Debug (Codecov OFF)') {
                     agent{ label rocmnode("vega") }
                     steps{
                         script{
                             try{
-                                buildHipClangJob('g++', flags: '-DBUILD_DEV=On -DCMAKE_BUILD_TYPE=debug', codecov: true)
+                                buildHipClangJob('g++', flags: '-DBUILD_DEV=On -DCMAKE_BUILD_TYPE=debug', codecov: false)
```
shurale-nkn, line 661: We have same stage in "Smoke Fp32", so disable it.

shurale-nkn (changes requested):

Jenkinsfile
atamazov: done

shurale-nkn (approved):

PR #852, opened 2021-04-09
[NFC] WrW gemm solvers

DrizztDoUrden:
WrW GEMM is now split between the following solvers:

- GemmWrw1x1_stride1 - applicable when kernel is 1x1 && conv pads are 0x0 && conv strides are 1x1
- GemmWrwUniversal - applicable when none of the solvers above is.

Solver ids update and testing are to do.


DrizztDoUrden:
Again no idea who to request a review from.

JehandadKhan:
@DrizztDoUrden for the rest of the GEMM solvers, we adopted `GEMMRest` naming convention. Can we adhere to the same  and not use a new name (`GemmWrwUniversal`) ?

src/include/miopen/solver_id.hpp
```diff
@@ -62,6 +62,7 @@ struct Id
     }
     bool operator!=(const Id& other) const { return !(*this == other); }
 
+    // For comparing with legacy find db items
```
JehandadKhan, line 65: Why not remove it ? In this PR and following it, a value of `gemm` for solver ID should be invalid.

src/ocl/convolutionocl.cpp
```diff
@@ -1012,16 +863,11 @@ void GetSolutions(Handle& handle,
         }
 
         // To stop crashes with old ufdb file
-        if(solver_id == solver::Id::gemm() && !problem.direction.IsBackwardWrW())
+        if(solver_id == solver::Id::gemm())
```
JehandadKhan, line 866: we can deal with this more efficiently at the parse time and not pollute the solver namespace.

JehandadKhan (changes requested):

src/include/miopen/solver_id.hpp
DrizztDoUrden: It is sometimes present in old database files, so we have to ignore specific cases rather than crash on them.

src/convolution.cpp
```diff
@@ -296,65 +292,32 @@ bool ConvolutionDescriptor::IsWinograd3x3SupportedAndFast(miopen::ConvolutionCon
...
-    std::ignore = wDesc;
-    std::ignore = dxDesc;
-    return 0;
-#endif
-}
-
 std::size_t
 ConvolutionDescriptor::WrwGetValidWorkSpaceSizeGemm(const TensorDescriptor& dyDesc,
-                                                    const TensorDescriptor& /*xDesc*/,
+                                                    const TensorDescriptor& xDesc,
```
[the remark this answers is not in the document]
DrizztDoUrden: The methods are kind of used that way below in the same file, every algorithm has its own method

src/ocl/convolutionocl.cpp
DrizztDoUrden: > Vasilii Filippov for the rest of the GEMM solvers, we adopted `GEMMRest` naming convention. Can we adhere to the same and not use a new name (`GemmWrwUniversal`) ?

Basically, I regret the decision to name the rest, because as far as I understand they are not, and not using them when anything is available is a kind of performance heuristics.

But I can rename this to rest as well if you wish.

DrizztDoUrden:
> Vasilii Filippov for the rest of the GEMM solvers, we adopted `GEMMRest` naming convention. Can we adhere to the same and not use a new name (`GemmWrwUniversal`) ?

Basically, I regret the decision to name the rest, because as far as I understand they are not, and not using them when anything is available is a kind of performance heuristics.

But I can rename this to rest as well if you wish.

src/ocl/convolutionocl.cpp
DrizztDoUrden: I'd rather stay with a simple solution and remove this completely, when we have system data base ready to remove it.

src/include/miopen/solver.hpp
```diff
@@ -2305,6 +2295,58 @@ struct GemmBwdRest : GemmBwdBase
     ConvSolution GetSolution(const ExecutionContext&, const conv::ProblemDescription&) const;
 };
 
+struct GemmWrwBase : SolverBase<ConvolutionContext>
+{
+    bool IsApplicable(const ExecutionContext&, const conv::ProblemDescription&) const;
+    bool IsDynamic() const { return true; }
+    float GetWti(const ConvolutionContext& ctx) const { return GetWti(ctx, ctx.conv_problem); }
+    float GetWti(const ExecutionContext& context, const conv::ProblemDescription& problem) const;
```
atamazov, line 2303: Make it non-public.

src/include/miopen/solver.hpp
```diff
@@ -2305,6 +2295,58 @@ struct GemmBwdRest : GemmBwdBase
...
+    }
+
+    ConvSolution GetSolution(const ConvolutionContext& ctx) const
+    {
+        return GetSolution(ctx, ctx.conv_problem);
+    }
+
+    size_t GetWorkspaceSize(const ExecutionContext&, const conv::ProblemDescription&) const;
+    bool IsApplicable(const ExecutionContext&, const conv::ProblemDescription&) const;
+    ConvSolution GetSolution(const ExecutionContext&, const conv::ProblemDescription&) const;
```
atamazov, line 2325: Non public methods, please

src/include/miopen/solver_id.hpp
atamazov: I agree with Jehandad. System databases shall be updated. The old user databases include VERSION_TWEAK (git hash) so these will not be used. I think that we do not need "gemm" solver anymore.

src/solver.cpp
```diff
@@ -413,6 +413,10 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
         registry, ++id, ConvHipImplicitGemmMlirBinBwd{}, miopenConvolutionAlgoImplicitGEMM);
     RegisterWithSolver(
         registry, ++id, ConvHipImplicitGemmMlirBinWrW{}, miopenConvolutionAlgoImplicitGEMM);
+
+    RegisterWithSolver(registry, ++id, GemmWrw1x1_stride1{}, miopenConvolutionAlgoGEMM);
```
atamazov, line 417: I think now we can kill solver 33/gemm.

atamazov (changes requested):

src/include/miopen/solver.hpp
DrizztDoUrden: Why? We are going to remove old ones and use only these methods after we update all the solvers to the new problem config.

atamazov: Okay

atamazov:
@DrizztDoUrden Please resolve merge conflicts and remaining review comments

src/solver.cpp
DrizztDoUrden: The solver was already removed. Only the string constant remained.

atamazov:
Performance testing failed ;(  Benchmarking of WrW Gemm Solvers during Find has issues (it is too optimistic). The result is huge perf drop for WrW.

Example for the following config (morpheus):
```
./bin/MIOpenDriver conv -x 20 -y 5 -W 700 -H 161 -c 1 -n 8 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6 -C /spb-hlc-fs2/1/users/atamazov/miopen-verify2
```
PR:
```
MIOpen(HIP): Info [EvaluateInvokers] GemmWrwUniversal: : 0.21504 < 3.40282e+38
...
MIOpen(HIP): Info [FindConvBwdWeightsAlgorithm] BWrW Chosen Algorithm: GemmWrwUniversal , 10775600, 0.21504
...
GPU Kernel Time Backward Weights Conv. Elapsed: 14.861814 ms (average)
```
Baseline results for the same configs:
```
MIOpen(HIP): Info [FindConvBwdWeightsAlgorithm] miopenConvolutionBwdWeightsAlgoGEMM	15.0152	10775600
...
MIOpen(HIP): Info [FindConvBwdWeightsAlgorithm] BWrW Chosen Algorithm: ConvOclBwdWrW2<1> , 102400, 0.36896
...
GPU Kernel Time Backward Weights Conv. Elapsed: 0.308544 ms (average)
```
Logs and CSVs attached: [smoke-052dff352.ddu_wrw-gemm-solvers.zip](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6438048/smoke-052dff352.ddu_wrw-gemm-solvers.zip)

DrizztDoUrden:
Wow, looks like I have missed a multiplication somewhere in the solver at some point. Will check in the morning.

atamazov:
Performance tests passed!

atamazov (approved):
LGTM!

JehandadKhan:
@atamazov You can go ahead and merge it, I will update Tuna

atamazov:
@JehandadKhan Merged!

PR #857, opened 2021-04-12
Update README to reflect libsqlite3-dev dependency

JehandadKhan:


junliume (dismissed):
LGTM, doc change only, low risk to pending CI 

It has fixed the explicit issue perfectly, while I am not sure if it has fixed the original problem. :) Just out of curiosity, I see min-requirements.txt has sqlite3 added about a year ago, why still "many users are finding issues"? Is `cmake -P install_deps.cmake --minimum` not part of the standard procedure or it does not work well for the users?

README.md
```diff
@@ -20,6 +20,7 @@ MIOpen supports two programming models -
 * [Half](http://half.sourceforge.net/) - IEEE 754-based half-precision floating point library
 * [Boost](http://www.boost.org/) at version 1.72 (other versions are not supported)
   * MIOpen uses `boost-system` and `boost-filesystem` packages to enable persistent [kernel cache](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/cache.html)
+* [libsqlite3-dev](https://sqlite.org/index.html) package which can be automatically installed with `cmake -P install_deps.cmake --minimum `
```
pfultz2, line 23: This really should only have a description of the dependency: `for reading and writing performance database tunings`. It also should be name `Sqlite3`.

We already document automatically installing the dependencies below nor we do repeat this for other dependencies such as boost.

JehandadKhan: thanks for the quick review @pfultz2  I will update the text.

pfultz2 (approved):

PR #858, opened 2021-04-14
[HOTFIX] Made kernel sources inlined as multiple files (full fix for SWDEV-281476)

DrizztDoUrden:
- Contains a full-blown fix for the SLES & CentOS build issues (SWDEV-281476)
- Improves the library build speed (previous attempt was #808): from ~10 to 5 minutes.

## Design notes

Kernel source files are now inlined as multiple files, batched by a factor of 50. It limits the maximum length of addkernels calls and speeds up compilation significantly. It should prevent crashes on systems with limited console call length.
50 was selected kind of arbitrarily, I have tested 1, 10, 30, 50, 100, and, obviously, unlimited (which is what happens in the develop). No values in between were tested. 30 and 50 we almost tied, 50 was 1% faster, for which reason I have decided further investigation unreasonable, compilation time curve would be close to parabolic.
As a side effect, kernel sources are declared extern, as they are built as separate object files. Variable names are prefixed with `MIOPEN_KERNEL_` to avoid accidental future conflicts.
Kernel include files are not affected, for now. The work would be easy to replicate for them.

A further build time improvement may be achieved by moving kernel source files to a separate static library, linked in MIOpen later. That would allow building hand-written source files of the library in parallel to inlining kernel sources. I don't think it would provide a significant difference right now, no measurements have been done by me.

Edit: Reverted C files to C++, because JenkinsFile and stuff do not currently support C source files in the library (when `CXX=/opt/rocm/llvm/bin/clang++`). C does not provide any noticeable speedup for batch factors significantly more than 1.


pfultz2:
> A further build time improvement may be achieved by moving kernel source files to a separate static library, linked in MIOpen later. That would allow building hand-written source files of the library in parallel to inlining kernel sources. I don't think it would provide a significant difference right now, no measurements have been done by me.

It might be even better to just use [`add_embed_library`](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/cmake/embed.cmake#L122) to inline kernels as this creates in parallel the .o files directly bypassing byte arrays in C++ source.

pfultz2 (dismissed):

DrizztDoUrden:
That's actually a great idea. But I don't think `add_embed_library` would work as-is. We would have to inline includes for the most of kernels or implement some other workaround for that.

shaojiewang:
Hi, there is a CI error: `clang-11: error: no such file or directory: '/var/jenkins/workspace/MIOpen_ddu_split_inlined_kernels/build/inlined_kernels/batch_0.cpp'`

DrizztDoUrden:
Yeah, I am working on that now.

DrizztDoUrden:
I officially hate make now. This works perfectly with ninja, but make fails pathetically by messing generated .cpp file dependencies.

atamazov (approved):
Good! Thank you!

junliume:
Thank you! Tomorrow's meeting can be a short conclusion now :)

shaojiewang:
Thank you!

PR #878, opened 2021-04-21
Handle missing file when embedded build of MIOpen (SWDEV-282133)

JehandadKhan:
* handle missing embedded files
  * Why: When an embedded file is not found in the internal map, instead of throwing an exception, MIOpen marks the database as invalid effectively converting accesses to that database as No-Op.
* remove file IO when embedded
  * See also [this comment](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/878#issuecomment-825089011).

This PR addresses [SWDEV-282133](http://ontrack-internal.amd.com/browse/SWDEV-282133)

### DIFF
--- a/src/sqlite_db.cpp
+++ b/src/sqlite_db.cpp
@@ -93,10 +93,8 @@ class SQLite::impl
             const auto& it_p = miopen_data().find(filepath.filename().string() + ".o");
             if(it_p == miopen_data().end())
             {
-                // Future: Try to load it from the disk system
-                // Also make a config point if Disk I/O is disabled
-                MIOPEN_THROW(miopenStatusInternalError,
-                             "Unknown database: " + filepath.string() + " in internal file cache");
+                MIOPEN_LOG_I("Unknown database: " + filepath.string() + " in internal file cache");
+                return SQLITE_ERROR;
             }
             const auto& p    = it_p->second;
             ptrdiff_t ptr_sz = p.second - p.first;
@@ -171,8 +169,9 @@ class SQLite::impl
 #else
         rc = CreateFileDb(filepath, is_system);
 #endif
-        sqlite3_busy_timeout(ptrDb.get(), MIOPEN_SQL_BUSY_TIMEOUT_MS);
         isValid = (rc == 0);
+        if(isValid)
+            sqlite3_busy_timeout(ptrDb.get(), MIOPEN_SQL_BUSY_TIMEOUT_MS);
     }
 
     sqlite3_ptr ptrDb = nullptr;
--- a/src/binary_cache.cpp
+++ b/src/binary_cache.cpp
@@ -123,8 +123,10 @@ KDb GetDb(const TargetProperties& target, size_t num_cu)
     boost::filesystem::path sys_path = sys_dir / (Handle::GetDbBasename(target, num_cu) + ".kdb");
     if(user_dir.empty())
         user_path = user_dir;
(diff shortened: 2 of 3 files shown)


junliume:
> * handle missing embedded files
> * remove file io when embedded
> 
> This PR addresses [SWDEV-282133](http://ontrack-internal.amd.com/browse/SWDEV-282133)

Could you elaborate how does this resolve MIOpen library from PyTorch nightly conflicts with container environment? Mostly changed files look like sys_path fix and maybe resolving sqlite conflict. Thanks!

JehandadKhan:
> > * handle missing embedded files
> > * remove file io when embedded
> > 
> > This PR addresses [SWDEV-282133](http://ontrack-internal.amd.com/browse/SWDEV-282133)
> 
> Could you elaborate how does this resolve MIOpen library from PyTorch nightly conflicts with container environment? Mostly changed files look like sys_path fix and maybe resolving sqlite conflict. Thanks!

Done!

atamazov:
@JehandadKhan Is it blocker for 4.3?

JehandadKhan:
> @JehandadKhan Is it blocker for 4.3?

Not a blocker per se but would be useful if we can get it in.

atamazov:
@JehandadKhan 
>     * remove file IO when embedded

Do you mean SQLite-related file IO? But after all, COMGR uses file IO during builds anyway, and the User Binary Cache is still useful, right?

JehandadKhan:
> @JehandadKhan
> 
> > ```
> > * remove file IO when embedded
> > ```
> 
> Do you mean SQLite-related file IO? But after all, COMGR uses file IO during builds anyway, and the User Binary Cache is still useful, right?

@atamazov Correct, only a check for sys db files has been restricted.

src/sqlite_db.cpp
```diff
@@ -93,10 +93,8 @@ class SQLite::impl
             const auto& it_p = miopen_data().find(filepath.filename().string() + ".o");
             if(it_p == miopen_data().end())
             {
-                // Future: Try to load it from the disk system
-                // Also make a config point if Disk I/O is disabled
-                MIOPEN_THROW(miopenStatusInternalError,
-                             "Unknown database: " + filepath.string() + " in internal file cache");
+                MIOPEN_LOG_I("Unknown database: " + filepath.string() + " in internal file cache");
+                return SQLITE_ERROR;
```
atamazov, line 97: [Q] Maybe it is worth keeping the old code under `#if !MIOPEN_EMBED_DB`?
```suggestion
#if !MIOPEN_EMBED_DB
                // Future: Try to load it from the disk system
                // Also make a config point if Disk I/O is disabled
                MIOPEN_THROW(miopenStatusInternalError,
                                             "Unknown database: " + filepath.string() + " in internal file cache");
#else
                MIOPEN_LOG_I("Unknown database: " + filepath.string() + " in internal file cache");
                return SQLITE_ERROR;
#endif
```

JehandadKhan: This function is only included when `MIOPEN_EMBED_DB` is `true` so the `!MIOPEN_EMBED_DB` branch will never be compiled.

JehandadKhan:
@atamazov CI passes

src/sqlite_db.cpp
atamazov: Thanks.

atamazov (approved):
LGTM

PR #882, opened 2021-04-22
Fixed reg allocation for k-reduction in xdlops gemm

zjing14:
- Fixed the reg allocation issue for K-reduction in xdlops gemm, which allocates more registers than it needs
- Refactored code to improve the readability of code


asroy (approved):

atamazov (approved):
LGTM

Issue #874, opened 2021-04-20
[tests] Failures of test_conv_embed_db


atamazov:
develop 5ee21784e: `test_conv_embed_db` fails with errors like
```
...MIOpen(HIP): Info [FindSolutionImpl] Perf Db: record not found for: ConvHipImplicitGemmV4R1WrW
```
Most likely the reason is either missing records in performance database or incorrect test (i.e. incorrect assumption that the config under test must be tuned). Many test cases in `test_conv_embed_db` are disabled due to this problem.

The issue blocks our CI. I will provide a workaround shortly.

[Informative] How to quickly check locally: build `test_conv2d` with HIP backend and then execute:
```
export MIOPEN_FIND_MODE=normal
./bin/test_conv2d --input 128 1024 14 14 --weights 2048 1024 1 1 --pads_strides_dilations 0 0 2 2 1 1 --disable-validation --verbose 2>&1 | grep -E -i "(fail)|(record not found)"
./bin/test_conv2d --input 128 1024 14 14 --weights 256 1024 1 1 --pads_strides_dilations 0 0 1 1 1 1 --disable-validation --verbose 2>&1 | grep -E -i "(fail)|(record not found)"
./bin/test_conv2d --input 128 1024 14 14 --weights 512 1024 1 1 --pads_strides_dilations 0 0 2 2 1 1 --disable-validation --verbose 2>&1 | grep -E -i "(fail)|(record not found)"
./bin/test_conv2d --input 128 128 28 28 --weights 128 128 3 3 --pads_strides_dilations 1 1 1 1 1 1 --disable-validation --verbose 2>&1 | grep -E -i "(fail)|(record not found)"
./bin/test_conv2d --input 128 1024 14 14 --weights 512 1024 1 1 --pads_strides_dilations 0 0 2 2 1 1 --disable-validation --verbose 2>&1 | grep -E -i "(fail)|(record not found)"
./bin/test_conv2d --input 128 128 28 28 --weights 512 128 1 1 --pads_strides_dilations 0 0 1 1 1 1 --disable-validation --verbose 2>&1 | grep -E -i "(fail)|(record not found)"
./bin/test_conv2d --input 128 2048 7 7 --weights 512 2048 1 1 --pads_strides_dilations 0 0 1 1 1 1 --disable-validation --verbose 2>&1 | grep -E -i "(fail)|(record not found)"
[...]

PR #885, opened 2021-04-22
[CI] Remove user db leftovers. [Jenkinsfile] Remove useless code.

### LINKED ISSUE #876: [CI] Clean up user-find-db files
When we test DEV builds, the user-find-db files remain in the ./src/kernels directory. These should be removed before launching the next round of testing. 

[Informative] See fix for similar issue: #814.

atamazov:
- Resolves #876. This improves reproducibility of test results.
  - For example, #874 will be stably reproducible (or non-reproducible).
- Removes harmless merge artifacts from PR #556


atamazov:
@ce1adon @shurale-nkn Please review, thanks!

ce1adon:
Are we sure user db is the only cache that we need to clear? Can we overwrite all files in ./build?
I'm not expert here. Please add @pfultz2 as the reviewer.

atamazov:
@ce1adon Please review 42ff550, do not worry about the rest.

Jenkinsfile
```diff
@@ -106,6 +106,7 @@ def buildHipClangJob(Map conf, compiler){
                         mkdir build
                         rm -rf install
                         mkdir install
+                        rm -rf src/kernels/*.ufdb.txt
```
shurale-nkn, line 109: pleas add src/kernels/miopen_1.0.0.udb

shurale-nkn (changes requested):

Jenkinsfile
atamazov: Do you really see the user-perf-db leftovers???

shurale-nkn: Locally yes

atamazov: Correct.

shurale-nkn (approved):

PR #888, opened 2021-04-23
Direct Asm 1x1 stride2 fixes for gfx90a

Slimakanzer:
## Local verification
* gfx906, rocm 4.0
* gfx90a, only kernel compilation


atamazov (approved):
No objections.

muralinr (approved):
Few changes are specific to gfx90A . LGTM.

PR #889, opened 2021-04-23
Implement mlir binary backend - step 2: bwd path

jerryyin:
Refer to #841 for the previous step in the series.

I did the following in this PR:
 - Implemented the MLIR binary backend invoker, bwd path
 - Implemented the MLIR binary backend solver, bwd path

I will work through the remaining work incrementally: This help to reduce the complexity for both myself as developer and the reviewers.

I did not add unit test for mlir binary backend because the corresponding commit does not guarantee correctness for bwd path. This means that I will rework this solver/invoker once I am able to bump the MLIR commit who can guarantee it.

### DIFF
--- a/src/conv/invokers/mlir_impl_gemm.cpp
+++ b/src/conv/invokers/mlir_impl_gemm.cpp
@@ -33,6 +33,7 @@
 #include <miopen/tensor_ops.hpp>
 
 #include <boost/any.hpp>
+#include <boost/range/adaptors.hpp>
 
 namespace miopen {
 namespace conv {
@@ -187,20 +188,16 @@ InvokerFactory MakeMlirBwdInvokerFactory(const ConvolutionContext& ctx)
     MlirConvArgs args =
         makeMlirConvArgs(in_dims, in_strides, weights_dims, weights_strides, out_dims, out_strides);
 
-    const auto& conv = ctx.conv_problem.GetConv();
+    const auto& conv  = ctx.conv_problem.GetConv();
     const auto& wDesc = ctx.conv_problem.GetWeights();
-    bool is_1x1_s1    = false;
-    if(miopen::all_of(conv.GetConvPads(), [](auto v) { return v == 0; }) &&
-       miopen::all_of(conv.GetConvStrides(), [](auto v) { return v == 1; }))
-    {
-        if(wDesc.GetLengths()[2] == 1 && wDesc.GetLengths()[3] == 1)
-        { // filter = 1
-            if(wDesc.GetSize() == 4 || (wDesc.GetSize() == 5 && wDesc.GetLengths()[4] == 1))
-            {
-                is_1x1_s1 = true;
-            }
-        }
-    }
+
+    const bool isFilter1Padding0Stride1 = [=]() {
+        const auto wei_spatial =
+            boost::adaptors::slice(wDesc.GetLengths(), 2, 2 + conv.GetSpatialDimension());
+        return miopen::all_of(wei_spatial, [](auto v) { return v == 1; }) &&
+               miopen::all_of(conv.GetConvPads(), [](auto v) { return v == 0; }) &&
+               miopen::all_of(conv.GetConvStrides(), [](auto v) { return v == 1; });
+    }();
 
     return [=](const std::vector<Kernel>& kernels) mutable {
         return [=](const Handle& handle, const AnyInvokeParams& primitive_parameters) mutable {
@@ -208,7 +205,7 @@ InvokerFactory MakeMlirBwdInvokerFactory(const ConvolutionContext& ctx)
             const auto& data_ctx = primitive_parameters.CastTo<conv::DataInvokeParams>();
             const auto& tensors  = data_ctx.tensors;
 
-            if(!is_1x1_s1)
+            if(!isFilter1Padding0Stride1)
             {
                 float zero = 0.f;
                 SetTensor(handle, tensors.outDesc, tensors.out, &zero);
--- a/src/conv/invokers/mlir_impl_gemm.cpp
+++ b/src/conv/invokers/mlir_impl_gemm.cpp
@@ -184,30 +184,29 @@ InvokerFactory MakeMlirBwdInvokerFactory(const ConvolutionContext& ctx)
                             weights_strides,
                             out_dims,
                             out_strides);
-
     MlirConvArgs args =
         makeMlirConvArgs(in_dims, in_strides, weights_dims, weights_strides, out_dims, out_strides);
+
     const auto& conv = ctx.conv_problem.GetConv();
+    const auto& wDesc = ctx.conv_problem.GetWeights();
+    bool is_1x1_s1    = false;
+    if(miopen::all_of(conv.GetConvPads(), [](auto v) { return v == 0; }) &&
+       miopen::all_of(conv.GetConvStrides(), [](auto v) { return v == 1; }))
+    {
+        if(wDesc.GetLengths()[2] == 1 && wDesc.GetLengths()[3] == 1)
+        { // filter = 1
+            if(wDesc.GetSize() == 4 || (wDesc.GetSize() == 5 && wDesc.GetLengths()[4] == 1))
+            {
+                is_1x1_s1 = true;
+            }
+        }
+    }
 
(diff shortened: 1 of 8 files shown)


atamazov:
> I did not add unit test for mlir binary backend because the corresponding commit does not guarantee correctness for bwd path. This means that I will rework this solver/invoker once I am able to bump the MLIR commit who can guarantee it.

In this case please disable this solver by default.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -135,33 +164,68 @@ InvokerFactory MakeMlirFwdInvokerFactory(const ConvolutionContext& ctx)
...
+            {
+                if(tensors.wDesc.GetLengths()[2] == 1 && tensors.wDesc.GetLengths()[3] == 1)
+                { // filter = 1
+                    if(tensors.wDesc.GetSize() == 4 ||
+                       (tensors.wDesc.GetSize() == 5 && tensors.wDesc.GetLengths()[4] == 1))
+                    {
+                        is_1x1_s1 = true;
+                    }
+                }
+            }
```
atamazov, line 210: Please move this out of the Factory.

src/solver/conv_hip_implicit_gemm_mlir_bin_bwd.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    return std::make_tuple(gemm_m, gemm_n, gemm_k);
+}
+#endif
+} // Anonymous namespace
+
+bool ConvHipImplicitGemmMlirBinBwd::IsApplicable(const ConvolutionContext& ctx) const
+{
+#if MIOPEN_USE_MLIR
+    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_HIP_IMPLICIT_GEMM_MLIR_BIN_BWD{}))
```
atamazov, line 63: `if(!miopen::IsEnabled(...))`

atamazov (changes requested):

src/conv/invokers/mlir_impl_gemm.cpp
jerryyin: Sorry about making this mistake once more. Addressed in 0f0c35e

src/solver/conv_hip_implicit_gemm_mlir_bin_bwd.cpp
jerryyin: Done, in 0f0c35e

src/conv/invokers/mlir_impl_gemm.cpp
atamazov: Nada problemas

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -135,33 +164,67 @@ InvokerFactory MakeMlirFwdInvokerFactory(const ConvolutionContext& ctx)
...
+       miopen::all_of(conv.GetConvStrides(), [](auto v) { return v == 1; }))
+    {
+        if(wDesc.GetLengths()[2] == 1 && wDesc.GetLengths()[3] == 1)
+        { // filter = 1
+            if(wDesc.GetSize() == 4 || (wDesc.GetSize() == 5 && wDesc.GetLengths()[4] == 1))
+            {
+                is_1x1_s1 = true;
+            }
+        }
+    }
```
atamazov, line 203: [Recommendation]
```C++
const auto isFilter1Padding0Stride1 = [](){
    const auto wei_spatial = boost::adaptors::slice(wDesc.GetLengths(), 2, 2 + conv.GetSpatialDimension());
    return  miopen::all_of(wei_spatial, [](auto v) { return v == 1; }) &&
       miopen::all_of(conv.GetConvPads(), [](auto v) { return v == 0; }) &&
       miopen::all_of(conv.GetConvStrides(), [](auto v) { return v == 1; });
}();
```

src/solver/conv_hip_implicit_gemm_mlir_bin_bwd.cpp
```diff
@@ -0,0 +1,164 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include <miopen/mlir_build.hpp>
+#include <miopen/conv/invokers/mlir_impl_gemm.hpp>
+#include <miopen/solver.hpp>
+#include <miopen/solver/implicitgemm_util.hpp>
+#include <miopen/handle.hpp>
```
atamazov, line 30: - handle.hpp is not needed.
- add env.hpp for env vars.
- add config.h for MIOPEN_USE_MLIR.

atamazov (changes requested):
Almost done

atamazov:
Please fix formatting.

src/conv/invokers/mlir_impl_gemm.cpp
jerryyin: Nice suggestion! Does make it much cleaner than nested if conditions. Addressed in 30aca46

atamazov: And now we have `const` object!

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -135,33 +165,63 @@ InvokerFactory MakeMlirFwdInvokerFactory(const ConvolutionContext& ctx)
...
+                            weights_strides,
+                            out_dims,
+                            out_strides);
+    MlirConvArgs args =
+        makeMlirConvArgs(in_dims, in_strides, weights_dims, weights_strides, out_dims, out_strides);
+
+    const auto& conv  = ctx.conv_problem.GetConv();
+    const auto& wDesc = ctx.conv_problem.GetWeights();
+
+    const bool isFilter1Padding0Stride1 = [=]() {
```
atamazov, line 194: Hmm... I do not think that you need capturing by-copy here.

atamazov (changes requested):

src/conv/invokers/mlir_impl_gemm.cpp
jerryyin: I don't think it matters in this case. I need to capture `conv` and `wDesc`, both are already const references. I suppose a copy of reference isn't that bad? I may need to make an experiment of some sort to see if the copied reference will invoke copy constructor, that'd be too expensive.

atamazov: > I don't think it matters in this case.

I think it almost always matters. Why: capturing by-reference is almost always cheaper than capturing by-copy. Therefore, there should be reasons for capturing by-copy.

If you do not need copies, then simply use `[&]`. When I see `[=]`, I begin asking myself "why the programmer wants to capture copies?"

Also I am suspecting that both `conv` and `wDesc` objects will be captured by value (but I am not sure).

> copied reference will invoke copy constructor,

Approximately, this is the same as copying the pointer.

jerryyin: > I begin asking myself "why the programmer wants to capture copies?

:grinning: Haha, it might be that part of driver development background in me. My intuition has been that capture by copies are safer than reference when you know for sure you don't want the object changed. Unfortunately lambda does not allow const reference so I do copies to be defensive. This discussion is getting to a bit philosophical but I agree at this case by reference is almost for sure cheaper. I'd still curious to find out what actually happens in the original case when capture by copy + reference combination comes together. Will take a look later and update here with the result.

atamazov: > Unfortunately lambda does not allow const reference so I do copies to be defensive.

If you capture a `const` object by-reference, then you have const reference in the lambda.

If the object is not `const`, then consider this: "_A capture with an initializer... makes it possible to capture by const reference, with_ `&cr = std::as_const(x)` _or similar_", see https://en.cppreference.com/w/cpp/language/lambda.

atamazov: [Resolved]

atamazov (approved):
LGTM

PR #879, opened 2021-04-21
Deprecate MLIR CPP solver, step 1

jerryyin:
According to https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/141

Will completely remove the MIOpen cpp backend after june demo.

Issue #869, opened 2021-04-19
[Recommendation] `ConvHipImplicitGemmMlirBinFwd` -> `ConvMlirImplicitGemmBinFwd` and ditto all other MLIR solvers.


atamazov:
[Recommendation] `ConvHipImplicitGemmMlirBinFwd` -> `ConvMlirImplicitGemmBinFwd` and ditto `MIOPEN_DEBUG_CONV_HIP_IMPLICIT_GEMM_MLIR_BIN_FWD`. Why: "Hip" here means that the kernels are (mostly, at least) written in HIP language. This one uses generated MLIR as a source. That fact that MLIR is based on the same llvm compiler as HIP (which is actually not so yet ;) does not matter in this specific case.

If you decide to follow this, I recommend making a separate PR.

_Originally posted by @atamazov in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/841#discussion_r615302923_

jerryyin:
@atamazov Could we adopt the following simplify the name such that:
 - we drop `Bin` because it is going to be the only backend in MIOpen
 - Make `ImplicitGemm` into `Igemm` to shorten the spelling

The results naming below for non-xdlops solver:
 - ConvMlirIgemmFwd
 - ConvMlirIgemmBwd
 - ConvMlirIgemmWrW

For xdlops solver:
 - ConvMlirIgemmFwdXdlops
 - ConvMlirIgemmBwdXdlops
 - ConvMlirIgemmWrWXdlops

atamazov:
No objections.

jerryyin:
For documentation purpose, the updated filename will be:

non-xdlops solvers:
  - solver/conv_mlir_igemm_fwd.cpp
  - solver/conv_mlir_igemm_bwd.cpp
  - solver/conv_mlir_igemm_wrw.cpp

xdlops solvers:
  - solver/conv_mlir_igemm_fwd_xdlops.cpp
  - solver/conv_mlir_igemm_bwd_xdlops.cpp
  - solver/conv_mlir_igemm_wrw_xdlops.cpp

atamazov:
via #896

PR #896, opened 2021-04-28
Misc update to igemm mlir solvers and invokers

jerryyin:
* Updated filename according to discussion in #869
* Updated class name according to discussion in #869
* Make local scope function conform to MIOpen naming convention

~Note: The only functional changes in ctest is to get rid of cpp solver tests. This is to avoid conflict with on-going PR #879~


atamazov:
> * Updated filename according to discussion in #869

I am not against this change, but it seems like filenames were not discussed there. If I am right, then please update the PR description.

> * Make local scope function conform to MIOpen naming convention

Well, IIRC the naming convention for functions is `CamelCase()`, Although usually nobody object from some local inconsistencies as long as they seem reasonable. Please update the description of this PR.

atamazov (approved):
No objections.

But if you undo the renaming of the local functions, there will be no objection either.

jerryyin:
> IIRC the naming convention for functions is CamelCase(), Although usually nobody object from some local inconsistencies as long as they seem reasonable.

@atamazov Somehow I see `CamelCase()` most commonly in member functions but `snake_case()` for local scoped functions. It'd good to know if all functions should follow `CamelCase()` convention. I will update my PR accordingly. Thanks.

atamazov (approved):

Issue #870, opened 2021-04-19
[Debug][test_conv2d][gfx908] Segmentation fault 


shurale-nkn:
test_conv2d failed on config from test_miopen_conv. Same error in driver.

GPU: gfx908
ROCM 3.7
develop 1f751ba8c
___cmake ... -DCMAKE_BUILD_TYPE=debug ...___

```
MIOPEN_ENABLE_LOGGING=1 MIOPEN_ENABLE_LOGGING_CMD=1  MIOPEN_LOG_LEVEL=6 ./bin/test_conv2d --verbose --input 1 64 512 1024 --weights 1 64 3 3 --pads_strides_dilations 2 2 1 1 1 1
```

<details><summary>CONSOLE LOGS</summary>

```

MIOpen(HIP): Info [GetFindModeValueImpl] MIOPEN_FIND_MODE = HYBRID(3)
MIOpen(HIP): Info [get_device_name] Raw device name: gfx908
MIOpen(HIP): Info [Handle] stream: 0x219d010, device_id: 3
MIOpen(HIP): Info2 [HipCompilerVersionImpl] Read version information from HIP package...
MIOpen(HIP): Info [HipCompilerVersionImpl] 3.7.20315
MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, MIOpen version 2.12.0.8413-1f751ba8c
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] Running: '/opt/rocm/llvm/bin/clang --version'
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] clang version 11.0.0 (/src/external/llvm-project/clang ee4e4ebbadcc8ea14ce99e34ed31ab31e94827ac)
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] Target: x86_64-unknown-linux-gnu
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] Thread model: posix
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] InstalledDir: /opt/rocm/llvm/bin
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] 
MIOpen(HIP): Info [BackwardDataGetWorkSpaceSize] 
MIOpen(HIP): Info [GetBackwardSolutions] 
MIOpen(HIP): Info [Measure] Db::Prefetch time: 134.498 ms
[...]

PR #900, opened 2021-05-03
Change gcd() implementation to avoid too deep recursion

asroy:
to fix issue #870


atamazov:
This fix is good. However we have another gcd() (from numeric.hpp) and it is still used in iGemm Invokers (https://github.com/ROCmSoftwarePlatform/MIOpen/issues/870#issuecomment-831241668).

carlushuang (approved):
LGTM, good implementation

atamazov (approved):
LGTM

PR #903, opened 2021-05-04
remove duplicated gcd() and remove src/include/miopen/numeric.hpp

carlushuang:
remove the duplicated `gcd` function defined in `src/include/miopen/numeric.hpp`, and remove this header file completely, as mentioned in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/870


atamazov (approved):
I would prefer opposite approach -- moving numeric functions to numeric.hpp (into the `miopen::numeric` namespace, btw), but that can be done later, when necessary.

Issue #908, opened 2021-05-07
miir.h header file missing, blocking build from source


carlushuang:
introduced by https://github.com/ROCmSoftwarePlatform/MIOpen/commit/f3921e7d5ca71d3f57333cc9de2bd5aa37e7fbba, now build from source can't success, due to no [`miir.h`](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/conv/invokers/mlir_impl_gemm.cpp#L35) in current source.
```
[ 48%] Building CXX object src/CMakeFiles/MIOpen.dir/solver/conv_MP_bidirectional_winograd.cpp.o
/root/MIOpen/src/conv/invokers/mlir_impl_gemm.cpp:35:10: fatal error: 'Miir.h' file not found
#include <Miir.h>
         ^~~~~~~~
[ 48%] Building CXX object src/CMakeFiles/MIOpen.dir/solver/conv_bin_wino3x3U.cpp.o
1 error generated when compiling for gfx900.
src/CMakeFiles/MIOpen.dir/build.make:11461: recipe for target 'src/CMakeFiles/MIOpen.dir/conv/invokers/mlir_impl_gemm.cpp.o' failed
make[2]: *** [src/CMakeFiles/MIOpen.dir/conv/invokers/mlir_impl_gemm.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

If this commit introduce dependency to mlir, need update document or method where to find this header file. because we might not build the docker file every time.

cc @atamazov @jerryyin

jerryyin:
Synced with @carlushuang offline. It becomes inconvenient that `Miir.h` is required when a developer come to debug a non-MIOpen docker image.

I plan on to push a quick PR that wrap the `Miir.h` include into a macro so non-MLIR build does not depend on a MLIR header.

atamazov:
This would break QA Staging testing.

atamazov:
MIOpen promotion process should wait until the fix is merged in `develop`.

atamazov:
via #909

PR #909, opened 2021-05-07
[HOTFIX] Fix build error due to missing `Miir.h` when MIOPEN_USE_MLIR=0 and MLIR library is not available

jerryyin:
From #908 

I compared this case and the `mlir_build.h` case, I honestly don't see a reason why compiler want to compile this cpp file (therefore the need to find the dependent `Miir.h`). When `MIOPEN_USE_MLIR` is not turned on, the compiler should not have visibility into any of the `mlir_impl_gemm.hpp` header functions.

However, it is what it is. Adding a macro to make sure when compiler does compile this implementation file, it does not require the `Miir.h` header.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -195,7 +195,6 @@ set( MIOpen_Source
     conv/invokers/gen_x_w_y_pad.cpp
     conv/invokers/ocl_wrw_rdc.cpp
     conv/invokers/impl_gemm.cpp
-    conv/invokers/mlir_impl_gemm.cpp
     conv/invokers/impl_gemm_dynamic.cpp
     invoker_cache.cpp
     tensor.cpp
@@ -585,6 +584,7 @@ endif()
 
 if(MIOPEN_USE_MLIR)
     list(APPEND MIOpen_Source mlir_build.cpp)
+    list(APPEND MIOpen_Source mlir_impl_gemm.cpp)
 endif()
 
 # build library
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -583,8 +583,7 @@ if(MIOPEN_USE_COMGR)
 endif()
 
 if(MIOPEN_USE_MLIR)
-    list(APPEND MIOpen_Source mlir_build.cpp)
-    list(APPEND MIOpen_Source mlir_impl_gemm.cpp)
+    list(APPEND MIOpen_Source mlir_build.cpp mlir_impl_gemm.cpp)
 endif()
 
 # build library
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -583,7 +583,7 @@ if(MIOPEN_USE_COMGR)
 endif()
 
 if(MIOPEN_USE_MLIR)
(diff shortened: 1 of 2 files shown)


atamazov:
@jerryyin 
> ...I honestly don't see a reason why compiler want to compile this cpp file (therefore the need to find the dependent `Miir.h`). When `MIOPEN_USE_MLIR` is not turned on, the compiler should not have visibility into any of the `mlir_impl_gemm.hpp` header functions.

The compiler wants to compile `mlir_impl_gemm.cpp` because is it unconditionally included into the list of the MIOpen source files:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f3921e7d5ca71d3f57333cc9de2bd5aa37e7fbba/src/CMakeLists.txt#L198

If you do want to exclude this file from the build, then move it here:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f3921e7d5ca71d3f57333cc9de2bd5aa37e7fbba/src/CMakeLists.txt#L586-L588

jerryyin:
> If you do want to exclude this file from the build, then move it here

Appreciate the pointer. Let me do it real quick in the next commit.

atamazov (approved):
LGTM but it is better to follow https://github.com/ROCmSoftwarePlatform/MIOpen/pull/909#issuecomment-834732366. Up to you.

src/CMakeLists.txt
```diff
@@ -585,6 +584,7 @@ endif()
 
 if(MIOPEN_USE_MLIR)
     list(APPEND MIOpen_Source mlir_build.cpp)
+    list(APPEND MIOpen_Source mlir_impl_gemm.cpp)
```
atamazov, line 587:
```suggestion
    list(APPEND MIOpen_Source mlir_build.cpp mlir_impl_gemm.cpp)
```

atamazov (changes requested):
Please also remove this
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f3921e7d5ca71d3f57333cc9de2bd5aa37e7fbba/src/conv/invokers/mlir_impl_gemm.cpp#L44

src/CMakeLists.txt
jerryyin: Nice catch. Addressed in 13bcd1a.

atamazov (approved):

atamazov:
Formatting issue, I guess.

src/CMakeLists.txt
```diff
@@ -584,7 +583,7 @@ if(MIOPEN_USE_COMGR)
 endif()
 
 if(MIOPEN_USE_MLIR)
-    list(APPEND MIOpen_Source mlir_build.cpp)
+    list(APPEND MIOpen_Source mlir_build.cpp mlir_impl_gemm.cpp)
```
atamazov, line 586:
```suggestion
    list(APPEND MIOpen_Source mlir_build.cpp conv/invokers/mlir_impl_gemm.cpp)
```

atamazov (changes requested):

atamazov (approved):

PR #910, opened 2021-05-07
[NFC] First activation solver

DrizztDoUrden:
This implements the first solver for activation. For now, I have taken the smallest possible PR approach, so the activation forward is split to the part with the solver and the rest, which would be translated to another solver and removed in the next PR.
I have kind of no idea how to name solvers here and would gladly accept the suggestions.
This PR also contains some essentials like activation problem description, and activation invoke parameters, which may change in the future.

Performance has been compared on a synthetic test of 105 activation forward configs with the following results: +0.91% avg GPU performance, +0.46% min GPU performance, +8.45% wall clock performance. I would like to note that this is probably not meaningful. For some reason, time is extremely volatile. Even when I run develop a couple of times in a row it varies significantly for small configs. I would experiment with larger configs and would attempt to come up with better test set.


DrizztDoUrden:
Status update: this has some performance regressions according to my tests. I am currently debugging the PR and the test.

atamazov:
@DrizztDoUrden 
> Performance has been compared...

The GPU performance should remain the same provided that this PR does not have bugs.

What is interesting is how this change affects the host-only overhead (shown by the driver for convolutions as "Aux Wall Time")

DrizztDoUrden:
I have 2.25% wall clock, 0.71% - gpu min and 0.73% - gpu avg time difference when running develop twice in a row.

atamazov:
I am going to merge this as soon as CI tests passed.

src/include/miopen/find_solution.hpp
```diff
@@ -182,6 +182,55 @@ struct SolverContainer
...
+                    }
+                    else
+                    {
+                        /// \todo If Solver is applicable it must provide an appropriate Solution.
+                        /// This is not the case for some 20x5 convolutions (and possibly others).
+                        /// Normally we should not get here and message level should be Error.
+                        /// For now, let's use Info (not Warning) level to avoid
+                        /// flooding the console.
+                        MIOPEN_LOG_I(SolverDbId(solver)
+                                     << ": [Warning] Applicable Solver not succeeded.");
```
atamazov, line 226: Do it in your next PR:
```suggestion
                        MIOPEN_LOG_E(SolverDbId(solver)
                                     << ": Applicable Solver not succeeded.");
```

src/include/miopen/find_solution.hpp
```diff
@@ -182,6 +182,55 @@ struct SolverContainer
...
+            [&](auto solver) {
+                if(count >= limit)
+                    return;
+                if(find_only.IsValid() && find_only != Id{SolverDbId(solver)})
+                { // Do nothing (and keep silence for the sake of Tuna), just skip.
+                }
+                // For better performance, check IsDynamic() first, because
+                // it is much faster than IsApplicable().
+                // else if(problem.use_dynamic_solutions_only && !solver.IsDynamic())
+                //    MIOPEN_LOG_I2(SolverDbId(solver) << ": Skipped (non-dynamic)");
```
atamazov, line 205: [future] I think we need to keep the notion of "dynamic" solutions for all primitives.

src/solver.cpp
```diff
@@ -417,6 +425,8 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
     RegisterWithSolver(registry, ++id, ConvMlirIgemmFwdXdlops{}, miopenConvolutionAlgoImplicitGEMM);
     RegisterWithSolver(registry, ++id, ConvMlirIgemmBwdXdlops{}, miopenConvolutionAlgoImplicitGEMM);
     RegisterWithSolver(registry, ++id, ConvMlirIgemmWrWXdlops{}, miopenConvolutionAlgoImplicitGEMM);
+
+    Register(registry, ++id, SolverDbId(activ::ActivFwdSolver0{}));
```
atamazov, line 429: If we have single registry for all primitives, won't this affect performance?

src/solver/activ/fwd_0.cpp
```diff
@@ -0,0 +1,217 @@
...
+                                                                   ? x_strides[1]
+                                                                   : (x_lens_.size() == 4)
+                                                                         ? x_strides[2]
+                                                                         : x_strides[3]);
+                    const auto y_stride2D = static_cast<unsigned int>(
+                        (y_lens_.size() == 2) ? y_strides[0] : (y_lens_.size() == 3)
+                                                                   ? y_strides[1]
+                                                                   : (y_lens_.size() == 4)
+                                                                         ? y_strides[2]
+                                                                         : y_strides[3]);
```
atamazov, line 194: It seems like many computations are invariant here and can be moved out from the Invoker and even out from the Factory. Shall we open a ticket for this optimization?

atamazov (approved):
Good to go!

src/solver.cpp
DrizztDoUrden: One would need hundreds of thousands of solvers to nothing something even on synthetic tests.

src/solver/activ/fwd_0.cpp
DrizztDoUrden: Oh, but they are not invariant. You are used to convolution, I have been too ;) Tensor descriptors are not part of network config in activation.

atamazov: Wow.

DrizztDoUrden: Took me a while to even consider that possibility.

src/include/miopen/find_solution.hpp
atamazov: @DrizztDoUrden agreed?

src/include/miopen/find_solution.hpp
atamazov: @DrizztDoUrden ok?

DrizztDoUrden: Yeah, probably

src/include/miopen/find_solution.hpp
DrizztDoUrden: ok

PR #921, opened 2021-05-11
Fix miminum requirements boost package download

### LINKED ISSUE #856: [Testing][Bug]Testing is stuck, boost can't be download
Rocm recipes is outdated, link on boost library is broken, access forbidden
https://dl.bintray.com/boostorg/release/1.72.0/source/boost_1_72_0.tar.bz2

Need to fix dockerfile. Change the way to install boost or to change boost source location.

jerryyin:
From Paul's change [here](https://github.com/pfultz2/rocm-recipes/blob/master/recipes/boost/1.72/package.txt).

This fix #856 when client try to `cmake -P install_deps.cmake --minimum`


pfultz2 (changes requested):
This should use the recipe not the URL(as the recipe will also install the zlib and bzip2 dependencies). Dependencies should be installed with `rbuild prepare` instead of install_deps.cmake. 

I dont know why `install_deps.cmake` still use the wrong URL as the URL has been updated for the recipe(which install_deps.cmake should use). So a fix needs to be applied to [cmake-get](https://github.com/pfultz2/cmake-get) repo but I dont really want to waste time trying to debug cmake code which is difficult to debug(although you could look into fixing it if you want, and I'll merge in such PR). It would be better to use `rbuild` or `cget` which is written in python.

pfultz2:
Alternatively, we could update install_deps.cmake to not use cmake-get, and instead install cget in a virtual env and invoke the cget commands.

pfultz2:
So PR#922 will fix the install_deps.cmake script.

PR #926, opened 2021-05-13
[MIOpenDriver] 3d conv set default layout as NCDHW

carlushuang:
After https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343, if user try MIOpenDriver with 3d convolution, it must specify the `in_layout`, `fil_layout`, `out_layout`, explicitly, otherwise it will use the default layout as `NCHW` (2d layout) while[ initialize cmd line arg](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/driver/conv_driver.hpp#L570), and result in fail.

More over, if enable `MIOPEN_ENABLE_LOGGING_CMD` and dump a 3d conv, it will not log the `in_layout`, `fil_layout`, `out_layout` if current layout is default (NCDHW). So to unify the behavior with 2d conv, this PR fix this problem by providing a default value(NCDHW) for 3d convolution in MIOpenDriver.

e.g:
A 3d conv config as below:
`./bin/MIOpenDriver conv  -n 2 -c 2 --in_d 5 -H 4 -W 3 -k 3 --fil_d 5 -y 4 -x 3 --pad_d 0 -p 0 -q 0 --conv_stride_d 1 -u 1 -v 1 --dilation_d 1 -l 1 -j 1  --spatial_dim 3  -m conv -g 1 -F 2 -t 1`
This will fail to execute, have following fail log:
```
  what():  /dockerx/repo/MIOpen/driver/tensor_driver.hpp:124: unmatched layout and dimension size
Aborted (core dumped)
```

This PR can fix this problem.

### DIFF
--- a/driver/InputFlags.hpp
+++ b/driver/InputFlags.hpp
@@ -57,6 +57,7 @@ class InputFlags
     int GetValueInt(const std::string& _long_name) const;
     uint64_t GetValueUint64(const std::string& _long_name) const;
     double GetValueDouble(const std::string& _long_name) const;
+    void SetValue(const std::string& _long_name, const std::string& _new_value);
 
     virtual ~InputFlags() {}
 };
--- a/driver/InputFlags.hpp
+++ b/driver/InputFlags.hpp
@@ -57,7 +57,7 @@ class InputFlags
     int GetValueInt(const std::string& _long_name) const;
     uint64_t GetValueUint64(const std::string& _long_name) const;
     double GetValueDouble(const std::string& _long_name) const;
-    void SetValue(const std::string& _long_name, const std::string& _new_value);
+    void SetValue(const std::string& long_name, const std::string& new_value);
 
     virtual ~InputFlags() {}
 };
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -439,6 +439,17 @@ int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
 {
     inflags.Parse(argc, argv);
 
+    // try to set a default layout value for 3d conv if not specified from cmd line
+    int spatial_dim = inflags.GetValueInt("spatial_dim");
+    if(spatial_dim == 3 &&
+       (inflags.GetValueStr("in_layout") == "NCHW" && inflags.GetValueStr("fil_layout") == "NCHW" &&
+        inflags.GetValueStr("out_layout") == "NCHW"))
+    {
+        inflags.SetValue("in_layout", "NCDHW");
+        inflags.SetValue("fil_layout", "NCDHW");
+        inflags.SetValue("out_layout", "NCDHW");
+    }
+
     num_iterations = inflags.GetValueInt("iter");
     if(num_iterations < 1)
     {
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -440,14 +440,26 @@ int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
     inflags.Parse(argc, argv);
 
     // try to set a default layout value for 3d conv if not specified from cmd line
-    int spatial_dim = inflags.GetValueInt("spatial_dim");
-    if(spatial_dim == 3 &&
-       (inflags.GetValueStr("in_layout") == "NCHW" && inflags.GetValueStr("fil_layout") == "NCHW" &&
-        inflags.GetValueStr("out_layout") == "NCHW"))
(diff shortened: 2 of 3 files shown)


driver/conv_driver.hpp
```diff
@@ -439,6 +439,17 @@ int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
 {
     inflags.Parse(argc, argv);
 
+    // try to set a default layout value for 3d conv if not specified from cmd line
+    int spatial_dim = inflags.GetValueInt("spatial_dim");
+    if(spatial_dim == 3 &&
+       (inflags.GetValueStr("in_layout") == "NCHW" && inflags.GetValueStr("fil_layout") == "NCHW" &&
+        inflags.GetValueStr("out_layout") == "NCHW"))
+    {
```
jerryyin, line 447: This feels a bit awkward that we use 2d layout try to initialize 3d layout. Since we want either 2d or 3d layout to be initialized with different values, I'd prefer to leave the `in/fil/out_layout` empty when parsing the command line args:

```cpp
bool layout_not_initialized = inflags.GetValueStr("in_layout").empty() &&  inflags.GetValueStr("fil_layout").empty() && inflags.GetValueStr("out_layout").empty() 

if (layout_not_initialized && spatial_dim == 2)
{
    inflags.SetValue("in_layout", "NCHW");
    inflags.SetValue("fil_layout", "NCHW");
    inflags.SetValue("out_layout", "NCHW");
} else if (layout_not_initialized && spatial_dim == 3)
{
    inflags.SetValue("in_layout", "NCDHW");
    inflags.SetValue("fil_layout", "NCDHW");
    inflags.SetValue("out_layout", "NCDHW");
}
```

jerryyin (changes requested):
Thanks for taking the initiative to fix this!

driver/conv_driver.hpp
carlushuang: Hi @jerryyin thanks for bring out this comment.
However `inflags.GetValueStr("in_layout").empty()` may not return true since `in/fil/out_layout` will have a default value,  initialized [here]( https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/driver/conv_driver.hpp#L570).
So if we need to set the default value depends on `spatial_dim`, we may need delete the initialization logic in above link. What do you think?

jerryyin: @carlushuang Yes that's what I mean. Make all layouts initialized with an empty string, and only set them when we have more information.

carlushuang: @jerryyin OK I see. Let me refactor the code based on above discussion.

driver/InputFlags.hpp
```diff
@@ -57,6 +57,7 @@ class InputFlags
     int GetValueInt(const std::string& _long_name) const;
     uint64_t GetValueUint64(const std::string& _long_name) const;
     double GetValueDouble(const std::string& _long_name) const;
+    void SetValue(const std::string& _long_name, const std::string& _new_value);
```
atamazov, line 60: No leading underscores, please. We will definitely face problems, it's only a matter of time.

driver/conv_driver.hpp
```diff
@@ -439,6 +439,29 @@ int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
 {
     inflags.Parse(argc, argv);
 
+    // try to set a default layout value for 3d conv if not specified from cmd line
+    int spatial_dim             = inflags.GetValueInt("spatial_dim");
+    bool layout_not_initialized = inflags.GetValueStr("in_layout").empty() &&
+                                  inflags.GetValueStr("fil_layout").empty() &&
+                                  inflags.GetValueStr("out_layout").empty();
```
atamazov, line 446: If only one layout is initialized, then the rest two will be left empty. Is it Ok? What about this:
```C++
const std::string default_layout = (spatial_dim == 2) ? "NCHW" : "NCDHW";
if(inflags.GetValueStr("in_layout").empty())
    inflags.SetValue("in_layout", default_layout);
/// the same for other two
```

atamazov (changes requested):

jerryyin (approved):
I have nothing further than @atamazov 's review comments.

driver/InputFlags.hpp
carlushuang: OK

driver/conv_driver.hpp
carlushuang: OK, I will change to this

atamazov (approved):

PR #955, opened 2021-05-27
[MLIR] Remove redundant solver local applicable checks

jerryyin:
I had a discussion with @zjing14 and @asroy and get the suggestion we set the single point of truth in judging applicability in MLIR instead of the existing hybrid (MLIR + MIOpen approach). The reasoning is that the applicability check in MIOpen end may grow out-of-sync with the actual MLIR solver capabilities. (One example of this is that EPack feature has not been implemented in MLIR but yet we have this redundant check `GetEPackLength()` that undermine the MLIR config coverage.)

In the PR, I'm removing all non-trivial applicability checks for MLIR backend solvers. Instead, the check will be performed by the pre-existing `MiirIsConfigApplicable()` check: This check will:
 - Invoke a MLIR API that does the applicability check
 - Lower through the first 3 step of passes: MIOpenOps, AffineTransform and AffixTuningParams
 - Config is applicable if and only if the lowering succeed: meaning the algorithm is able to find a compatible set of tuning parameters and proceed.

For this PR, I want all reviewers to approve before merging.

### DIFF
--- a/src/solver/conv_mlir_igemm_fwd.cpp
+++ b/src/solver/conv_mlir_igemm_fwd.cpp
@@ -38,6 +38,27 @@ namespace solver {
 
 namespace {
 #if MIOPEN_USE_MLIR
+std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx)
+{
+    const size_t g  = ConvolutionContextInterpreter::GetGroupCountG(ctx);
+    const size_t n  = ConvolutionContextInterpreter::GetBatchN(ctx);
+    const size_t k  = ConvolutionContextInterpreter::GetOutputChannelK(ctx);
+    const size_t c  = ConvolutionContextInterpreter::GetInputChannelC(ctx);
+    const size_t ho = ConvolutionContextInterpreter::GetOutputHeightHo(ctx);
+    const size_t wo = ConvolutionContextInterpreter::GetOutputWidthWo(ctx);
+    const size_t y  = ConvolutionContextInterpreter::GetFilterHeightY(ctx);
+    const size_t x  = ConvolutionContextInterpreter::GetFilterWidthX(ctx);
+
+    const auto k_per_group = k / g;
+    const auto c_per_group = c / g;
+
+    const auto gemm_m       = k_per_group;
+    const auto gemm_n       = n * ho * wo;
+    const auto gemm_k_total = c_per_group * y * x;
+
+    return std::make_tuple(gemm_m, gemm_n, gemm_k_total);
+}
+
 std::string GetKernelName()
 {
     std::string version   = "_v4r4";
@@ -64,6 +85,14 @@ bool ConvMlirIgemmFwd::IsApplicable(const ConvolutionContext& ctx) const
     if(!ctx.direction.IsForward())
         return false;
     if(!ctx.IsFp32() && !ctx.IsFp16())
+        return false;
+
+    int gemm_m = 0;
+    int gemm_n = 0;
+    int gemm_k = 0;
+
+    std::tie(gemm_m, gemm_n, gemm_k) = CalculateGemmSize(ctx);
+    if(!(gemm_m % 32 == 0 && gemm_n % 32 == 0 && gemm_k % 4 == 0))
         return false;
 
     return MiirIsConfigApplicable(
--- a/src/solver/conv_mlir_igemm_fwd.cpp
+++ b/src/solver/conv_mlir_igemm_fwd.cpp
@@ -38,27 +38,6 @@ namespace solver {
 
 namespace {
 #if MIOPEN_USE_MLIR
-std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx)
-{
-    const size_t g  = ConvolutionContextInterpreter::GetGroupCountG(ctx);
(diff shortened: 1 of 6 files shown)


atamazov:
> Instead, the check will be performed by the pre-existing MiirIsConfigApplicable() check: This check will...

Is this done already?

jerryyin:
> Is this done already?

Yes, this is already done. Therefore removing the redundant applicable checks are safe at this point.

atamazov (dismissed):
LGTM!

atamazov:
`on_hold` because
> For this PR, I want ___all reviewers to approve___ before merging.

src/solver/conv_mlir_igemm_fwd.cpp
```diff
@@ -87,14 +66,6 @@ bool ConvMlirIgemmFwd::IsApplicable(const ConvolutionContext& ctx) const
     if(!ctx.IsFp32() && !ctx.IsFp16())
```
asroy, line 66: This should also be moved into MLIR

src/solver/conv_mlir_igemm_wrw.cpp
```diff
@@ -90,14 +69,6 @@ bool ConvMlirIgemmWrW::IsApplicable(const ConvolutionContext& ctx) const
     if(!ctx.IsFp32() && !ctx.IsFp16())
```
asroy, line 69: Same as above, and below (if there is any)

asroy (dismissed):

whchung (changes requested):
Per discussion, MLIR kernel generator still needs validations, especially when it comes to cases where Gemm M/N/K dimensions have to be extended.

for example:

- C=3 config in resnet50
- some configs in resnext101 , especially in corner cases where the extent we enlarge along Gemm M / N dimensions are different.

Before that I don't think it's a good idea to merge this PR.

jerryyin:
Converted to draft. Feedbacks from @whchung and @asroy make me realize that we are probably missing this feature in MLIR now. Until we can correctly reject non-supported configs from MLIR, the checks have to be in place in MIOpen.

src/solver/conv_mlir_igemm_fwd.cpp
jerryyin: Conceptually all applicability check can be moved to let mlir handle it. However, that means every time MIOpen check for applicability of any config (like bf16), will ask MLIR's opinion. And the performance overhead of that check would be lowering through the first 3 steps of the pipeline.

My suggestion is that we try to hit at a balance between effectiveness and performance. Right now let's leave those extremely cheap checks in MIOpen untouched. Once we make the query overhead in checking with MLIR much lower than now, we can consider to remove those.

jerryyin:
Quick update to all: 

For the short term we will do the following:
  @asleepzzz  will take over the branch and add necessary checks for padding kernel.

For the mid/long term, we will shoot for removing applicable checks entirely.

whchung (approved):

asleepzzz:
it seems that fwd and wrw walk through different path, 
even remove limitations of mlir solver

bin/MIOpenDriver convfp16 -F 4 -n 256 -c 3 -H 224 -W 224 -k 64 -y 7 -x 7 -p 3 -q 3 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
blocked by other condition

asleepzzz:
resnet50 configs passed


./bin/MIOpenDriver convfp16 -F 4 -n 256 -c 3 -H 224 -W 224 -k 64 -y 7 -x 7 -p 3 -q 3 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
MIOpenDriver convfp16 -F 4 -n 256 -c 3 -H 224 -W 224 -k 64 -y 7 -x 7 -p 3 -q 3 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
MIOpen Backward Weights Conv. Algorithm: 5, Solution: 105/ConvMlirIgemmWrWXdlops
GPU Kernel Time Backward Weights Conv. Elapsed: 2199.225830 ms (average)
stats: bwdw-conv7x7u2, 256, 3, 112, 112, 7, 7, 64,  60423143424, 0, 0, 27, 0, 2199.225830
Backward Convolution Weights Verifies OK on CPU reference (0.000225131)


./bin/MIOpenDriver convfp16 -F 1 -n 256 -c 3 -H 230 -W 230 -k 64 -y 7 -x 7 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
MIOpenDriver convfp16 -F 1 -n 256 -c 3 -H 230 -W 230 -k 64 -y 7 -x 7 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
MIOpen Forward Conv. Algorithm: 5, Solution: 103/ConvMlirIgemmFwdXdlops
GPU Kernel Time Forward Conv. Elapsed: 4066.726074 ms (average)
stats: fwd-conv7x7u2, 256, 3, 112, 112, 7, 7, 64,  60423143424, 81273216, 411041792, 15, 0, 4066.726074
Forward Convolution Verifies OK on CPU reference (9.17173e-05)

Issue #894, opened 2021-04-26
[test_conv2d][half] ConvOclDirectFwd1x1 fails verification


shurale-nkn:
Using develop and #891
test_conv2d with fp16 convolution failed.

Runtime:OpenCL
GPU: gfx906

```
$ ./bin/test_conv2d --half --cmode conv --pmode default --group-count 1  --input 1, 16, 7, 7 --weights 16, 16, 1, 1 --pads_strides_dilations 0 0 1 1 1 1
./bin/test_conv2d --half --cmode conv --pmode default --group-count 1 --input 1, 16, 7, 7 --weights 16, 16, 1, 1 --pads_strides_dilations 0 0 1 1 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW 
FAILED: 0.284186
Iteration: 0
Forward convolution: ConvOclDirectFwd1x1
Input tensor: 1, 16, 7, 7
Weights tensor: 16, 16, 1, 1
Output tensor: 1, 16, 7, 7
Filter: conv2d, miopenConvolution, miopenPaddingDefault, {0, 0}, {1, 1}, {1, 1}, 
Max diff: 52
Mismatch at 0: -8 != 0
```

_Originally posted by @shurale-nkn in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/890#issuecomment-825994215_

shurale-nkn:
MIOPEN_LOG_LEVEL=5 
<details><summary>CONSOLE LOGS</summary>

```
./bin/test_conv2d --half --cmode conv --pmode default --group-count 1  --input
 1, 16, 7, 7 --weights 16, 16, 1, 1 --pads_strides_dilations 0 0 1 1 1 1 -v
MIOpen(OpenCL): Info [GetFindModeValueImpl] MIOPEN_FIND_MODE = HYBRID(3)
MIOpen(OpenCL): Info [get_device_name] Raw device name: gfx906
MIOpen(OpenCL): Info [Handle] stream: 0x555915adb180, device_id: 0x555915a6db30
MIOpen(OpenCL): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, MIOpen version 2.12.0.8426-856dc8555
MIOpen(OpenCL): Info [FindSolutionImpl] ConvAsm1x1U (db access disabled)
[...]

PR #956, opened 2021-05-28
Fix fp16 precision issue in ConvOclDirectFwd1x1

ce1adon:
For issue #894 
Please pull and check.


junliume:
@ce1adon understand that ocl runtime might be less critical, while can we still add a testing case for this issue?

ce1adon:
> @ce1adon understand that ocl runtime might be less critical, while can we still add a testing case for this issue?

@shurale-nkn Is the config reported in #894 an existing test case in CI? If no, is there any concern if we add it through this PR?

shurale-nkn:
This is not part of CI at the moment.
Better to add it like a custom_test for ocl and this solver

atamazov:
@junliume 
> understand that ocl runtime might be less critical,

The bug is in the OpenCL solver (the one that uses OpenCL kernels). MIOpen supports OpenCL kernels for both OpenCL and HIP runtimes. In other words, both runtimes are affected.

atamazov:
@ce1adon 
> If no, is there any concern if we add it through this PR?

Yes, please. _Actually this is the preferred practice_ -- I mean adding regression tests along with fixes.

Please add a small test with `MIOPEN_FIND_MODE=normal` and `MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvOclDirectFwd1x1`. I think that the test should be engaged only when `-DMIOPEN_TEST_ALL=On` and `MIOPEN_TEST_HALF=On`. Please do not hesitate to ask for assistance if necessary.

ce1adon:
@atamazov @shurale-nkn Please add an urgency label for management purposes.

ce1adon:
> Please add a small test with `MIOPEN_FIND_MODE=normal` and `MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvOclDirectFwd1x1`. I think that the test should be engaged only when `-DMIOPEN_TEST_ALL=On` and `MIOPEN_TEST_HALF=On`. Please do not hesitate to ask for assistance if necessary.

Please check if https://github.com/ROCmSoftwarePlatform/MIOpen/pull/956/commits/9432495f5dc9a2e10f0d7c88cb50ba93e64764d5 meets your requirement. Meanwhile, I notice that we DO NOT have half test for opencl backend at this moment, so even we added this test, it will not be run in CI.

test/CMakeLists.txt
```diff
@@ -906,3 +906,12 @@ if(MIOPEN_TEST_CONV)
     COMMAND	$<TARGET_FILE:test_conv2d>	--verbose	--input	1	48	7	7	--weights	1	48	5	5	--pads_strides_dilations	0	0	4	4	1	1
 )
 endif()
+
+# REGRESSION TEST for issue #894
+if(MIOPEN_BACKEND_OPENCL AND MIOPEN_TEST_ALL AND MIOPEN_TEST_HALF)
+    set(CONV_OCL_REGRESSION_TEST_ENV MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvOclDirectFwd1x1)
```
shurale-nkn, line 912: Please make it applicable for smoke, not only for MIOPEN_TEST_ALL. This test is small and fast enough for smoke.

shurale-nkn (changes requested):

atamazov:
@ce1adon 
> Meanwhile, I notice that we DO NOT have half test for opencl backend at this moment,... 

Thanks for noticing this. IIRC, the reason is that MIOpen with OpenCL backend is not intended to be used with FP16, because MIOpenGEMM does not support FP16. Hopefully your MIopenTensile/OCL work will change this for the better, and then we'll add tests. 

> ...so even we added this test, it will not be run in CI.

Well, no, it **_will_** be run. We often get confused with OpenCL solvers/kernels and OpenCL backend. MIOpen is able to use OpenCL solvers/kernels with HIP backend. And uses them, indeed.

Ditto almost all HIP solvers/kernels. MIOpen can use HIP kernels with OCL backend.

/cc @junliume

test/CMakeLists.txt
ce1adon: FYI, I've enabled it in smoke test.

test/CMakeLists.txt
```diff
@@ -906,3 +906,12 @@ if(MIOPEN_TEST_CONV)
...
 endif()
+
+# REGRESSION TEST for issue #894
+if(MIOPEN_BACKEND_OPENCL AND MIOPEN_TEST_HALF)
+    set(CONV_OCL_REGRESSION_TEST_ENV MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvOclDirectFwd1x1)
+
+    add_custom_test(test_miopen_conv_ocl_half_regression SKIP_UNLESS_ALL ALLOW_HALF
+    COMMAND	${CONV_OCL_REGRESSION_TEST_ENV} $<TARGET_FILE:test_conv2d>  --verbose   --half  --cmode conv    --pmode default --group-count   1 --input	1   16	7	7	--weights	16	16	1	1	--pads_strides_dilations	0	0	1	1	1	1
+)
+endif()
```
atamazov, line 917: Please merge recent develop and then:
```suggestion
    add_custom_test(test_regression_half HALF_ENABLED FLOAT_DISABLED GFX908_ENABLED
    COMMAND MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvOclDirectFwd1x1 ${CONV_OCL_REGRESSION_TEST_ENV} $<TARGET_FILE:test_conv2d>  --verbose   --half  --cmode conv    --pmode default --group-count   1 --input	1   16	7	7	--weights	16	16	1	1	--pads_strides_dilations	0	0	1	1	1	1
)
endif()
```
Later, we'll add here more FP16-specific regression tests (hopefully, not many ;)))

atamazov (changes requested):

test/CMakeLists.txt
atamazov: @ce1adon Thanks! Yes, I see that in another commit, but also I still see `SKIP_UNLESS_ALL` there ;)

atamazov:
@ce1adon Please resolve review comments.

test/CMakeLists.txt
ce1adon: Merged develop with ocl only constraint removed

atamazov: - We need to use `HALF_ENABLED` to enable the test for FP16 (`ALLOW_HALF` is not working anymore)
- The test is not enabled for MI100 by default; please add GFX908_ENABLED for that (I think we want this check, right?)
- This `--half` test will be run by both FP16 and FP32 stages; I recommend adding `FLOAT_DISABLED` to disable it for FP32.
- We do not need the enclosing `if(MIOPEN_TEST_HALF)...endif()` construct anymore.

Some internal documentation about new "attributes" of `add_custom_test`: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/81c0e4f98a548710626b75ef324774a174a80f0e/test/CMakeLists.txt#L295-L319

ce1adon: see update

atamazov: Thanks.

atamazov (dismissed):
LGTM!

test/CMakeLists.txt
atamazov: @shurale-nkn Let's add regression tests to "Full" stages for now. They will grow in the future.

test/CMakeLists.txt
```diff
@@ -1056,3 +1056,8 @@ endif()
 if(MIOPEN_TEST_FLOAT)
     add_custom_test(test_reduce_double SKIP_UNLESS_ALL GFX908_ENABLED COMMAND  $<TARGET_FILE:test_reduce_test> --double --all --verbose)
 endif()
+
+# REGRESSION TEST for issue #894
+add_custom_test(test_miopen_conv_ocl_half_regression SKIP_UNLESS_ALL FLOAT_DISABLED HALF_ENABLED GFX908_ENABLED 
+COMMAND	MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvOclDirectFwd1x1 $<TARGET_FILE:test_conv2d>  --verbose   --half  --cmode conv    --pmode default --group-count   1 --input	1   16	7	7	--weights	16	16	1	1	--pads_strides_dilations	0	0	1	1	1	1
```
atamazov, line 1062: Please add `--disable-backward-data --disable-backward-weights` or this will always fail. I guess it's also worth adding `--disable-verification-cache`.

atamazov (approved):
LGTM!

shurale-nkn:
I repeat again, regression test should be in smoke stage! We don't have full stage for fp16 and OpenCL.
At the moment, test has never been executed in our CI.


image

atamazov:
> I repeat again, regression test should be in smoke stage! We don't have full stage for fp16 and OpenCL.
Yes

> At the moment, test has never been executed in our CI.
No. "_Fp16 Hip Install All gfx908_" just failed on CI on this test and I am looking into it ;)

Note that the issue with `ConvOclDirectFwd1x1` is not inherent to OCL backend only, it must appear with HIP BE as well.

However, I 's starting to agree with you about adding to "Smoke" stages.

@ce1adon Can you allow me to make necessary changes in this PR?

ce1adon:
> @ce1adon Can you allow me to make necessary changes in this PR?

Go ahead, please.

atamazov:
> No. "_Fp16 Hip Install All gfx908_" just failed on CI on this test and I am looking into it ;)

Fails due to W/A for https://github.com/AMDComputeLibraries/MLOpen/issues/2298 that disables ConvOclDirectFwd1x1 for MI100. I'll make necessary changes.

test/CMakeLists.txt
atamazov: resolved

atamazov (approved):

PR #959, opened 2021-05-28
Find db update for rocm 4.3

cderb:
generated using docker with:
compute-rocm-dkms-no-npi-hipclang-7240/
MIOpen branch 9d1380b44f656887d52ed400ea9c01b4e0ce3eb5


atamazov:
Some comments from the mail thread:

> @atamazov 
>> MIOpen develop 9d1380b44 has defects, whose fixes affect applicability of kernels. **Unfortunately, that means that find-db produced with it is sub-optimal.** I don't know how much worse it is than the optimum, but there is a risk.

@junliume 
> This “backup” db is generated based on the urgent request of releasing team, with the expectation that it might take a while. I’d recommend that we generate a new one (optimal db) based on more up to date develop next Tuesday, since it would take about 10 hours which we will ensure is not an abused feature :-)

I am not against merging this provided that the new PR generated from the most recent `release/rocm-rel-4.3-staging` will be created soon.

For now, I need to test this.

/cc @JehandadKhan

atamazov (changes requested):
FP32 performance testing shows some progressions but many regressions. Average performance drop is ~10% (HYBRID mode). Some of the regressions can be explained by missing `export MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS=1` setting during tuning. :stop_sign:

atamazov:
> MIOpen develop 9d1380b44 has defects, whose fixes affect applicability of kernels. **Unfortunately, that means that find-db produced with it is sub-optimal.** I don't know how much worse it is than the optimum, but there is a risk.

I do not think that performance regressions are related to the above.

/cc @JehandadKhan @junliume

atamazov:
Logs and .csv files: [smoke-fp32-93-imm.sh.hip-clang.f558cc236.cderb_rocm4.3_fdb.zip](https://github.com/ROCmSoftwarePlatform/MIOpen/files/6580601/smoke-fp32-93-imm.sh.hip-clang.f558cc236.cderb_rocm4.3_fdb.zip)

PR #963, opened 2021-06-02
[HOTFIX] Made kernel includes inlined as multiple files (fix for SWDEV-284637)

atamazov:
This is a previously planned library improvement, continuation of #858. 

- Implemented the same "batched" machinery for inlining kernel includes as currently used for source files (#858).
  - This improves build time and removes restrictions on the number of include files.
  - As a side effect, resolves https://ontrack-internal.amd.com/browse/SWDEV-284637, whose root reason is unknown. Most likely this is x86 compiler/library issue related to `std::map` and https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables.
- Changed: Normal filename (with extension) in the map as a key for all kernels files instead of uppercased NAME_WE.
- Fixed the names of asm files that the library uses for auto-detection of some assembler features.


junliume:
@atamazov Root cause of SWDEV-284637 is still unknown, while this "fix" has been tested so that symptoms in SWDEV-284637 are no longer reproducible?

atamazov:
> 
> 
> @atamazov Root cause of SWDEV-284637 is still unknown, while this "fix" has been tested so that symptoms in SWDEV-284637 are no longer reproducible?

Yes

junliume (approved):

PR #967, opened 2021-06-04
[NFC] Second activation fwd solver

DrizztDoUrden:
It implements the second forward activation solver, complete with an invoker.
An option has been added to the activation driver to make tensors not packed so that the solver would be invoked.

### DIFF
--- a/driver/tensor_driver.hpp
+++ b/driver/tensor_driver.hpp
@@ -139,6 +139,7 @@ int SetTensorNd(miopenTensorDescriptor_t t,
 
 size_t GetTensorSize(miopenTensorDescriptor_t& tensor)
 {
+    assert(miopen::deref(tensor).IsPacked() && "GetTensorSize should not be used on an unpacked tensor.");
     const auto len = GetTensorLengths(tensor);
     size_t sz      = std::accumulate(len.begin(), len.end(), size_t{1}, std::multiplies<size_t>());
 
--- a/driver/tensor_driver.hpp
+++ b/driver/tensor_driver.hpp
@@ -143,7 +143,8 @@ int SetTensorNd(miopenTensorDescriptor_t t,
 // The implementation is a copy-paste from miopen::TensorDescriptor.
 size_t GetTensorSize(miopenTensorDescriptor_t& tensor)
 {
-    assert(miopen::deref(tensor).IsPacked() && "GetTensorSize should not be used on an unpacked tensor.");
+    assert(miopen::deref(tensor).IsPacked() &&
+           "GetTensorSize should not be used on an unpacked tensor.");
     const auto len = GetTensorLengths(tensor);
     size_t sz      = std::accumulate(len.begin(), len.end(), size_t{1}, std::multiplies<size_t>());
 
--- a/driver/tensor_driver.hpp
+++ b/driver/tensor_driver.hpp
@@ -147,16 +147,6 @@ size_t GetTensorSize(miopenTensorDescriptor_t& tensor)
 
 size_t GetTensorSpace(miopenTensorDescriptor_t& tensor)
 {
-    const auto lens    = GetTensorLengths(tensor);
-    const auto strides = GetTensorStrides(tensor);
-    std::vector<std::size_t> maxIndices(lens.size());
-    std::transform(lens.begin(),
-                   lens.end(),
-                   std::vector<std::size_t>(lens.size(), 1).begin(),
-                   maxIndices.begin(),
-                   std::minus<std::size_t>());
-    return std::inner_product(
-               maxIndices.begin(), maxIndices.end(), strides.begin(), std::size_t{0}) +
-           1;
+    return miopen::deref(tensor).GetElementSpace();
 }
 #endif // GUARD_MIOPEN_TENSOR_DRIVER_HPP
--- a/driver/tensor_driver.hpp
+++ b/driver/tensor_driver.hpp
@@ -137,6 +137,10 @@ int SetTensorNd(miopenTensorDescriptor_t t,
     return miopenSetTensorDescriptor(t, data_type, len.size(), len.data(), strides.data());
 }
 
+// This function ignores tensor strides completely and its result should not be interpreted as
+// memory required for an "unpacked" tensor. In such cases GetTensorSpace should be used instead.
+// For "packed" tensors result may be interpreted as an amount of memory required for the tensor.
+// The implementation is a copy-paste from miopen::TensorDescriptor.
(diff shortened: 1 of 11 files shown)


src/include/miopen/find_solution.hpp
```diff
@@ -170,11 +170,7 @@ struct SolverContainer
                     {
                         /// \todo If Solver is applicable it must provide an appropriate Solution.
                         /// This is not the case for some 20x5 convolutions (and possibly others).
-                        /// Normally we should not get here and message level should be Error.
-                        /// For now, let's use Info (not Warning) level to avoid
-                        /// flooding the console.
-                        MIOPEN_LOG_I(SolverDbId(solver)
-                                     << ": [Warning] Applicable Solver not succeeded.");
+                        MIOPEN_LOG_E(SolverDbId(solver) << ": Applicable Solver not succeeded.");
```
atamazov, line 173: Please revert this. https://github.com/ROCmSoftwarePlatform/MIOpen/pull/910/files#r657108163 is not related to non-convolution primitives only.

src/include/miopen/find_solution.hpp
```diff
@@ -219,11 +215,7 @@ struct SolverContainer
...
```
atamazov, line 218: Please also remove two lines of comments above.

src/ocl/activ_ocl.cpp
```diff
@@ -77,191 +77,21 @@ miopenStatus_t ActivationDescriptor::Forward(Handle& handle,
...
-                                           static_cast<long long>(yOffset));
-    });
-    return (status);
+    const auto ctx = ExecutionContext{&handle};
+    const auto solvers =
+        solver::SolverContainer<solver::activ::ActivFwdSolver0, solver::activ::ActivFwdSolver1>{};
+    const auto slns = solvers.SearchForSolutions(ctx, problem, 1);
+
+    if(slns.empty())
+        MIOPEN_THROW("No solver found for activation forward.");
```
atamazov, line 86:
```suggestion
        MIOPEN_THROW(miopenStatusNotImplemented, "No solver found for activation forward.");
```

src/solver/activ/fwd_0.cpp
```diff
@@ -131,6 +118,13 @@ ConvSolution ActivFwdSolver0::GetSolution(const ExecutionContext&,
         build_params.Define("MIOPEN_USE_FP16", 1);
         build_params.Define("MIOPEN_USE_FP32", 0);
     }
+    else
+    {
+        MIOPEN_LOG_W("Unsupported data types configuration: "
```
atamazov, line 123: `MIOPEN_LOG_E` and then `return {miopenStatusInternalError}`, I think.

src/solver/activ/fwd_1.cpp
```diff
@@ -0,0 +1,274 @@
...
+            std::tie(cOut, hOut, wOut)                   = tien<3>(yDesc.GetLengths());
+            std::tie(cOutStride, hOutStride, wOutStride) = tien<3>(yDesc.GetStrides());
+            nOutStride = cOut * cOutStride;
+            break;
+        default: assert(false);
+        }
+    }
+    else
+    {
+        MIOPEN_THROW("activation does not support tensor size larger than 4 or smaller than 1");
```
atamazov, line 104: `...miopenStatusNotImplemented,...`

atamazov: ...Or this is just a move of some existing code?

src/solver/activ/fwd_1.cpp
```diff
@@ -0,0 +1,274 @@
...
+    }
+    else if(problem.GetXDesc().GetType() == miopenHalf &&
+            problem.GetYDesc().GetType() == miopenHalf)
+    {
+        compiler_options.Define("MIOPEN_USE_FP32", 0);
+        compiler_options.Define("MIOPEN_USE_FP16", 1);
+    }
+    else
+    {
+        MIOPEN_LOG_W("Unsupported data types configuration: "
```
atamazov, line 227: LOG_E and then `return {miopenStatusInternalError}`?

atamazov, line 227: LOG_E and then `return {miopenStatusInternalError}`?

atamazov (changes requested):

src/solver/activ/fwd_1.cpp
DrizztDoUrden: Existing, but I'll change anyway.

atamazov: Thanks!

atamazov (approved):
LGTM!

atamazov:
@DrizztDoUrden Performance tests are not applicable because baseline is unable to call the second solver?

DrizztDoUrden:
I can make some perf tests with modified develop if needed, but what you said is kind of true.

atamazov:
@DrizztDoUrden Yes, it would be good to test it just in case to make sure there is no perf drop.

DrizztDoUrden:
Testing in process, found some bug, in process of fixing

DrizztDoUrden:
Fixed the bug, compared the performance. As with fwd-0, no significant difference. Minor (~10%) improvements on avg wall time with 251 test cases, compared to develop. But as before, I won't call it significant due to high variation. Although I don't want to spend time comparing develop to develop again to prove that point.

driver/tensor_driver.hpp
```diff
@@ -144,4 +144,19 @@ size_t GetTensorSize(miopenTensorDescriptor_t& tensor)
 
     return sz;
 }
+
+size_t GetTensorSpace(miopenTensorDescriptor_t& tensor)
```
atamazov, line 148: Can you please add a comment that describes the difference between `GetTensorSize()` and `GetTensorSpace()`? In other words, when `GetTensorSpace()` must be used?

PS Re-runing CI job is not necessary except formatting check, so please can terminate the job after adding the comment.

atamazov (dismissed):

driver/tensor_driver.hpp
DrizztDoUrden: I think I should also add an assert about the incorrect use of 'GetTensorSize()'. That is extremely easy to step in, and it may not be immediately obvious due to how memory is allocated on GPU. With an 'assert', there would be at least some additional chance future development time would be saved due to immediate failure in debug on any incorrect use, rather than only with a large enough difference between requested and required memory.

atamazov: Very good, thanks!

atamazov (approved):
LGTM!

PR #968, opened 2021-06-07
Fix: Explicitly remove temp dirs (avoid leftovers when the user's app exits abnormally)

zpwenjh:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/01ff3da68b68b2ffbc31cb681184a8082ac7ce26/src/hipoc/hipoc_program.cpp#L230
The temp directory `dir` saved in `HIPOCProgramImpl` will not be deleted until the end of the process. If MIOpen does not exit normally, the temp dir will never be deleted.
This PR is to solve the above problem.


atamazov:
@DrizztDoUrden Please review.

src/include/miopen/hipoc_program_impl.hpp
```diff
@@ -53,7 +53,7 @@ struct HIPOCProgramImpl
     TargetProperties target;
     boost::filesystem::path hsaco_file;
     hipModulePtr module;
-    boost::optional<TmpDir> dir;
+    mutable boost::optional<TmpDir> dir;
```
JehandadKhan, line 56: Can you please help me understand why this was required ?

zpwenjh: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/81c0e4f98a548710626b75ef324774a174a80f0e/src/include/miopen/hipoc_program.hpp#L55
In `HIPOCProgram`, `HIPOCProgramImpl` is const object. I can't change `dir` to `boost::none` without `mutable` keyword.

JehandadKhan (dismissed):

src/include/miopen/hipoc_program_impl.hpp
atamazov: The rule of thumb is that a member can be made mutable when is does not change the externally visible state of the object. This is not the case, because change of `dir` affects the validity of the value returned by `GetCodeObjectPathname()`. In other words, after deletion of temporary directory, the pathname of the CO file is no longer valid.

atamazov, line 56: Please revert this and make `HIPOCProgramImpl` non-const.

src/include/miopen/hipoc_program.hpp
```diff
@@ -61,6 +61,7 @@ struct HIPOCProgram
     /// \return True if CO blob resides in-memory.
     /// False if CO resides on filesystem.
     bool IsCodeObjectInMemory() const;
+    void DelTemp() const;
```
atamazov, line 64: Please rename this to `ReleaseCodeObjectFileStorage()`.

src/hipoc/hipoc_program.cpp
```diff
@@ -369,6 +369,8 @@ std::string HIPOCProgram::GetCodeObjectBlob() const
     return {impl->binary.data(), impl->binary.size()};
 }
 
+void HIPOCProgram::DelTemp() const { impl->dir = boost::none; }
```
atamazov, line 372: Please also invalidate `impl->hsaco_file` (just clear it).

atamazov: When `impl->hsaco_file` is invalid, `HIPOCProgram::GetCodeObjectPathname()` must `MIOPEN_THROW(miopenStatusInternalError...)`.

atamazov (changes requested):
The _original_ design of `HIPOCProgram` is cumbersome, but let's try to keep it under control.

src/hipoc/hipoc_program.cpp
atamazov: Please add the following to the documentation of `GetCodeObjectPathname():
```C++
/// This function should not be called after FreeCodeObjectFileStorage().
```

src/hipoc/hipoc_program.cpp
```diff
@@ -362,13 +362,29 @@ HIPOCProgram::HIPOCProgram(const std::string& program_name, const std::string& h
 
 hipModule_t HIPOCProgram::GetModule() const { return impl->module.get(); }
 
-boost::filesystem::path HIPOCProgram::GetCodeObjectPathname() const { return impl->hsaco_file; }
+boost::filesystem::path HIPOCProgram::GetCodeObjectPathname() const
+{
+    if(!impl->hsaco_file.empty())
+    {
+        impl->hsaco_file;
```
atamazov, line 369:
```suggestion
        return impl->hsaco_file;
```

atamazov (changes requested):

atamazov:
@DrizztDoUrden Please have a look.

DrizztDoUrden (approved):

atamazov (approved):

atamazov:
@zpwenjh Thanks for another contribution!

PR #969, opened 2021-06-07
[HOTFIX][HIP iGemm][NFC] Rename type.hpp -> ck_utils_type.hpp (another W/A for SWDEV-284637)

atamazov:
This change is "No Functional Changes", but nevertheless it resolves [SWDEV-284637](https://ontrack-internal.amd.com/browse/SWDEV-284637) (as well as #963 does). This confirms the idea that SWDEV-284637 is caused by some problem outside of the MIOpen.

I am going to cherry-pick this to 4.3 release in order to kill the unknown enemy twice.

PR #971, opened 2021-06-07
Fix for tidy failures when branch name contains 'hpp'

shurale-nkn:
Header filter works wrong if path contains 'hpp' substring.

1 jenkins run shows this error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -528,7 +528,7 @@ enable_clang_tidy(
         ${MIOPEN_TIDY_CHECKS}
     ${MIOPEN_TIDY_ERRORS}
     HEADER_FILTER
-        ".*hpp"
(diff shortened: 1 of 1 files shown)


junliume:
Together with #969 ?

shurale-nkn:
This should fix the issue with tidy that occurred in PR#969.

atamazov:
I spend ~3 days trying to "fix" my PR

## :exploding_head:

atamazov (approved):

atamazov:
> Together with #969 ?

Not necessary, I've renamed the branch to avoid this issue.

atamazov:
Okay, let's lower priority a bit. But I believe nobody wants to repeat my _experience_ with this ~sh~ stuff.

PR #640, opened 2020-12-04
Immediate Mode Fallback: Implement & use GetWti for Winograd v21_1_1 and GEMM

atamazov:
Continuation of #424. Implements (30.10) and (20) for WinogradRxS F(2,3) (n_groups=1) from #410.

- Updates the core of the library.
- Expected to allow simpler integration of MLP classifier
- Adds `GetWti()` for 3x3 WinogradRxS F(2,3), all directions.
- Adds very approximate `GetWti()` for GEMM, all directions.

Provides:
- Performance improvements for non-group convolutions, all directions:
  - For Immediate Mode Fallback & FAST Find mode (unseen configs)
  - Significant FP16 improvements because now we utilize Winograd

:warning: ___For detailed understanding of this code, I recommend reviewing each commit individually.___ Each commit has clear description, so the reviewer has luxury to skip what he or she is not interested in (for ex. refactors).

## Summary of code changes

- Updated machinery that uses `GetWti()` in Immediate Mode Fallback.
  - Estimation of Gemm performance is better generalized.
- Implemented shader performance model for ConvBinWinogradRxSf2x3, non-grouped (shader v.21_1_1).
- Added `GetWti()` for `ConvBinWinogradRxSf2x3g1` and for `ConvBinWinogradRxSf2x3`.
  - The latter always returns "unknown" due to WTI model limitations.
- Implemented simple shader performance model for GEMM.
- Removed rudimentary `ConvBinWinograd3x3U::GetWti()`. We have better Winograd shader with good WTI model.

## By-products

- Removed useless applicability limitation for `ConvBinWinogradRxSf2x3`, WrW direction.
- Removed some dead code in `ConvBinWinogradRxSf2x3`.

## :cyclone: Local performance & correctness testing

- Radeon VII, ROCm 3.5.
- 93 known FP32 + 93 known FP16 convolution configs
- Normal Find mode + Immediate mode
- OCL backend (FP32 configs only) + HIP backend

No regressions. Detailed description of the test is available upon request.

## :cyclone: Evaluation of Initial Iteration Time impact

- ___Installable Release build___, binary cache is empty prior each test.
- Radeon VII, ROCm 3.5.
- 93 _unknown_ FP32 convolution configs (find-db miss)
- 93 _unknown_ FP16 convolution configs (find-db miss)
- HIP backend

&nbsp; | Total Wall time, ms | &nbsp; | &nbsp; | GPU perf vs. Normal Find | &nbsp; | &nbsp;
[...]

PR #986, opened 2021-06-16
[gfx10] Winograd f3x2 v21_1_2 update

Slimakanzer:
This PR adds _Winograd v21_1_2 f3x2_ shaders and upgrades Winograd f3x2 solver.

## Shader changes compared with _v16_5_0_
- Supports extended cases set
- Supports WrW direction
- Supports grouped convolution
- Tensor offset support (64 bit byte offsets)

⚠️ fp16 kernels will be added in a subsequent PR


src/solver/conv_winoRxS_f3x2.cpp
```diff
@@ -32,154 +32,42 @@
...
-                                         const int N,
-                                         const miopen::ConvolutionContext& params,
-                                         const bool fp16,
-                                         const unsigned filter_tile_size)
+namespace {
+/// \todo Consider re-using code from RxS_f2x3.
+inline bool IsShaderContraintsMet(const int R,
+                                  const int S,
+                                  const int,
+                                  const int,
```
atamazov, line 51: Shall we remove unused params?

src/solver/conv_winoRxS_f3x2.cpp
```diff
@@ -32,154 +32,42 @@
...
-        assert(padded_R % TILE == 0 && padded_S % (is_dilated_stride_2 ? TILE_X2 : TILE) == 0);
-        const int k = padded_R / TILE;
-        const int l = padded_S / (is_dilated_stride_2 ? TILE_X2 : TILE);
-        if(k * l * C < 18)
-            return false;
-    }
     // Padding for bwd data shall not be negative.
     /// \todo Either remove WrW related code or re-use function from RxS
-    if(params.direction.IsBackwardData() || params.direction.IsBackwardWrW())
+    if(params.direction.IsBackwardData())
```
atamazov, line 63: Bugfix?

src/solver/conv_winoRxS_f3x2.cpp
```diff
@@ -218,19 +105,13 @@ bool ConvBinWinogradRxSf3x2::IsApplicable(const ConvolutionContext& params) cons
...
         return false;
-    if(!(params.direction.IsForward() || params.direction.IsBackwardData()))
-        return false;
     if(!params.use_asm_kernels)
         return false;
-    if(!params.rmv.IsV2orV3())
-        return false;
-    if(!params.IsLayoutDefault())
-    {
+    if(!params.rmv.IsV3())
```
atamazov, line 110: Do we support non-NCHW layout now?

src/solver/conv_winoRxS_f3x2.cpp
```diff
@@ -239,141 +120,339 @@ bool ConvBinWinogradRxSf3x2::IsApplicable(const ConvolutionContext& params) cons
...
+    if(!IsWarned)
+    {
+        if(params.GetStream().GetMaxHardwareComputeUnits() > MAX_CU_LIMIT)
+            MIOPEN_LOG_WE(SolverDbId(*this) << ": GPU has "
+                                            << params.GetStream().GetMaxHardwareComputeUnits()
+                                            << "CUs, but this solver supports max "
+                                            << MAX_CU_LIMIT
+                                            << "and thus may show sub-optimal performance.");
+        IsWarned = true;
+    }
```
atamazov, line 176: - Do you want to issue this message only once?
- Why would you like to issue an error message for DEV builds?
  - If this is an error, then maybe we shall make it non-applicable?

atamazov:
@Slimakanzer Do you have any performance data? I would like to compare the performance of the new shader vs. the old one.

src/solver/conv_winoRxS_f3x2.cpp
Slimakanzer: Thank you! Fixed in 1e8b8a8

src/solver/conv_winoRxS_f3x2.cpp
Slimakanzer: This is an incorrect limitation that originally comes from `conv_winoRxS` solver.
Now wino_f3x2 has the same constraints as `conv_winoRxS_f2x3` solver, where this limitation was  [removed](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/640/commits/c20d531b086b179c5625ec1a642d1505991a8232).

src/solver/conv_winoRxS_f3x2.cpp
Slimakanzer: No, we dont. Constraint checked in the `IsShaderContraintsMet`: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/bcf87b840cb07f728582bf745b9f76bf9565a220/src/solver/conv_winoRxS_f3x2.cpp#L71-L74

However, this check is not obvious, so it was fixed in e105526

src/solver/conv_winoRxS_f3x2.cpp
Slimakanzer: Shader still will work, but there might be severe performance penalties. It's better to set as non-applicable in this case.
Thank you, fixed in 9db3bc6

src/solver/conv_winoRxS_f3x2.cpp
atamazov: I see, thanks.

src/solver/conv_winoRxS_f3x2.cpp
atamazov: Good!

src/kernels/Conv_Winograd_v21_1_2_f3x2_fp32_stride1_group.s
```diff
@@ -0,0 +1,40 @@
...
+ *******************************************************************************/
+.include "Conv_Winograd_v21_1_2_metadata.inc"
+
+KERNEL_PROLOG f3x2_fp32_stride1_group
+
+.if (.amdgcn.gfx_generation_number == 9)
+    .if (.amdgcn.gfx_generation_stepping == 10)
+        .error "gfx90a is not supported yet"
+    .else
+        .include "Conv_Winograd_v21_1_2_f3x2_gfx9_fp32_stride1_group.inc"
```
Kirpich30000, line 34: Please make file naming consistent with kernel names. Code in conv_winoRxS_f3x2.cpp might also be simplified if the kernel name and file name would be consistent.

atamazov (dismissed):
LGTM!

src/kernels/Conv_Winograd_v21_1_2_f3x2_fp32_stride1_group.s
atamazov: @Kirpich30000 You review does not block merging. Is it wnat you want, i.e. is it Ok to merge this without resolving your review comments first?

Kirpich30000: I preffer to resolve the issues.

Kirpich30000 (changes requested):
See comments

src/kernels/Conv_Winograd_v21_1_2_metadata.inc
```diff
@@ -210,20 +210,28 @@ METADATA_WRAPPER total_sgpr_count,.amdgcn.next_free_vgpr,workgroup_size_x, <\ker
 
 .endm
 
-.macro PROLOG_KERNEL_DESCRIPTOR_WRAPPER machine_version, kernel_name_postfix
-	PROLOG_KERNEL_DESCRIPTOR miopenSp3AsmConv_v21_1_2_gfx\machine_version\()_\kernel_name_postfix
+.macro PROLOG_KERNEL_DESCRIPTOR_WRAPPER machine_version, kernel_name_postfix, kernel_name_prefix
```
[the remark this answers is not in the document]
Slimakanzer: Thank you, fixed in 8af72ae

Slimakanzer:
@atamazov
I apologize for the discarded reviews. Can you check new changes?

atamazov:
@Slimakanzer No problem. I'll check but review comments are from @Kirpich30000 )))

Kirpich30000 (approved):
LGTM

PR #987, opened 2021-06-19
[HOTFIX] Fix bug ConvHipImplicitGemmBwdDataV4R1Xdlops (SWDEV-291202)

asroy:
Fix a typo

@asleepzzz

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
@@ -198,7 +198,7 @@ PerformanceImplicitGemmBwdDataV4R1Xdlops::CalculateGemmBBlockCopyPerformancePara
 
         // calculate threadwise copy size
         int b_data_per_thread_copy =
-            std::max(1, (GemmKPerBlock * GemmMPerBlock * GemmKPACKSize) / BlockSize);
+            std::max(1, (GemmKPerBlock * GemmNPerBlock * GemmKPACKSize) / BlockSize);
 
         if(!(b_data_per_thread_copy > 0))
             MIOPEN_THROW("invalid performance parameter");


asleepzzz (approved):
LGTM

junliume:
@asroy Can we add a test case to the CI?

atamazov:
> @asroy Can we add a test case to the CI?

+1. I am not going to delay merging this PR and cherry-picking it, however. But if this PR will not contain a regression test, then I'll open a high priority ticket for this.

atamazov:
:warning: [IMPORTANT QUESTION] @asroy AFAICS this changes PerformanceConfigs for ConvHipImplicitGemmBwdDataV4R1Xdlops. This should make some existing PerformanceConfigs in the System perf-db invalid.

### Was `ConvHipImplicitGemmBwdDataV4R1Xdlops::IsValidPerformanceConfig()` updated automatically, so that invalid PerformanceConfigs will be automatically rejected by the library core?

It should be so, but it may worth double-checking.

atamazov:
@JehandadKhan If the answer to the question above is positive, then this PR may lead to the new "Warning [FindSolutionImpl] Invalid config loaded from Perf Db..." messages. A quick W/A would be purging invalid entries for `ConvHipImplicitGemmBwdDataV4R1Xdlops` from the System find-db.

Otherwise we'll need to retune `ConvHipImplicitGemmBwdDataV4R1Xdlops`.

asroy:
> Was `ConvHipImplicitGemmBwdDataV4R1Xdlops::IsValidPerformanceConfig()` updated automatically, so that invalid PerformanceConfigs will be automatically rejected by the library core?

Yes, IsValidPerformanceConfig is updated automatically by this PR

asroy:
ctest added 
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/997

JehandadKhan:
> 
> 
> @JehandadKhan If the answer to the question above is positive, then this PR may lead to the new "Warning [FindSolutionImpl] Invalid config loaded from Perf Db..." messages. A quick W/A would be purging invalid entries for `ConvHipImplicitGemmBwdDataV4R1Xdlops` from the System find-db.
> 
> Otherwise we'll need to retune `ConvHipImplicitGemmBwdDataV4R1Xdlops`.

Is there a way to identify invalid entries for this solver without known which problem config they are applicable to ? 
If the answer to above is yes ( which is typically the case if we increase or reduce the number of parameters ) then yes we can simply purge those entries. If the answer is no then we would need to re-tune the entire solver.

cc @asroy

atamazov:
@JehandadKhan Some hints:
- When `MIOPEN_FIND_ENFORCE=3`, MIOpen re-tunes only invalid (or missing) entries. 
- `MIOPEN_FIND_MODE=normal` + `MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvHipImplicitGemmBwdDataV4R1Xdlops` makes sure that only ConvHipImplicitGemmBwdDataV4R1Xdlops will be used, thus saving time.
- Maybe it is possible to use Tuna database in the look? Only configs that have ConvHipImplicitGemmBwdDataV4R1Xdlops records in the Tuna database needs to be tried, thus saving time again. 

Hope this helps (but I suspect it doesn't)

Issue #540, opened 2020-10-22
ConvAsmImplicitGemmGTCDynamicWrwXdlops memory access fault


asroy:
found on CI for PR #488:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/enable_all_igemm_solver/8/pipeline

```
 test_conv2d --float --cmode conv --pmode default --group-count 1 --input 4, 32, 79, 341 --weights 4, 32, 5, 10 --pads_strides_dilations 0 0 2 2 1 1 --trans_output_pads 0 0 --disable-forward --disable-backward-data
```
```
MIOpen(OpenCL): Info [Handle] Device name: gfx908
MIOpen(OpenCL): Info [Handle] stream: 0x279ffa0, device_id: 0x284df20
MIOpen(OpenCL): Info [BackwardDataGetWorkSpaceSize]
MIOpen(OpenCL): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, MIOpen version 2.10.0.8243-3a090530d
MIOpen(OpenCL): Info2 [ValidateGcnAssemblerImpl] Running: '/opt/rocm/llvm/bin/clang --version'
MIOpen(OpenCL): Info2 [ValidateGcnAssemblerImpl] clang version 11.0.0 (/src/external/llvm-project/clang ee4e4ebbadcc8ea14ce99e34ed31ab31e94827ac)
MIOpen(OpenCL): Info2 [ValidateGcnAssemblerImpl] Target: x86_64-unknown-linux-gnu
MIOpen(OpenCL): Info2 [ValidateGcnAssemblerImpl] Thread model: posix
MIOpen(OpenCL): Info2 [ValidateGcnAssemblerImpl] InstalledDir: /opt/rocm/llvm/bin
MIOpen(OpenCL): Info2 [ValidateGcnAssemblerImpl]
MIOpen(OpenCL): Info [GetFindModeValueImpl] MIOPEN_FIND_MODE = NORMAL(1)
MIOpen(OpenCL): Info2 [SQLiteBase] Initializing system database file /root/workspace/MIOpen/MIOpen-pr488/src/kernels/miopen.db
MIOpen(OpenCL): Info2 [SQLiteBase] Initializing user database file /root/workspace/MIOpen/MIOpen-pr488/src/kernels/miopen_1.0.0.udb
[...]

Issue #980, opened 2021-06-11
[MI100][FP16] Fatal error in ConvAsmImplicitGemmGTCDynamicWrwXdlops


atamazov:
```
MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlops \
./bin/MIOpenDriver convfp16 -n 1 -c 128 -H 56 -W 56 -y 5 -x 5 -k 1 -p 0 -q 0 -u 2 -v 2 -i 1 -V 0 -F 4
```
Output:
```
MIOpenDriver convfp16 -n 1 -c 128 -H 56 -W 56 -y 5 -x 5 -k 1 -p 0 -q 0 -u 2 -v 2 -i 1 -V 0 -F 4
MIOpen(HIP): Info [get_device_name] Raw device name: gfx908:sramecc+:xnack-
MIOpen(HIP): Info [Handle] stream: 0x14b12f0, device_id: 0
MIOpen(HIP): Info [GetFindModeValueImpl] MIOPEN_FIND_MODE = NORMAL(1)
MIOpen(HIP): Info [BackwardWeightsGetWorkSpaceSize]
MIOpen(HIP): Info2 [HipCompilerVersionImpl] Read version information from HIP package...
MIOpen(HIP): Info [HipCompilerVersionImpl] 4.2.21155
MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, MIOpen version 2.12.0.c8532c566
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] Running: '/opt/rocm/llvm/bin/clang --version'
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] clang version 12.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-4.2.0 21155 fd12620eb54b192daf0231b6950421a85cda1feb)
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] Target: x86_64-unknown-linux-gnu
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] Thread model: posix
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl] InstalledDir: /opt/rocm/llvm/bin
MIOpen(HIP): Info2 [ValidateGcnAssemblerImpl]
[...]

PR #1002, opened 2021-06-22
Add regression test for issue 540

shaojiewang:
add regression test for #540


atamazov:
@shaojiewang This PR adds test for #540, but for #980 we already have it in place (currently disabled due to some W/A), right?

shaojiewang:
> @shaojiewang This PR adds test for #540, but for #980 we already have it in place (currently disabled due to some W/A), right?

right.

atamazov (approved):
LGTM

PR #998, opened 2021-06-22
[WORKAROUND] Generic-Reduction: Partially resolve SWDEV-291479. W/A for SWDEV-274384.

qianfengz:
1)  The first commit is supposed to prevent the failure of lost index updates reported in [SWDEV-291479](https://ontrack-internal.amd.com/browse/SWDEV-291479) and [SWDEV-274384](https://ontrack-internal.amd.com/browse/SWDEV-274384)
2) The second commit is supposed to prevent the  failure of all-zero accumulated output (both cpu and gpu) for some reduce_test cases
3) The third commit is for fixing one potential issue. 

- The P.R has passed the following test 

```bash
bin/test_reduce_test --all  --double
bin/test_reduce_test --all
bin/test_reduce_test --all --half
```

- About the Performance 

No performance loss was seen when running `bin/test_reduce_test --all`  on cached binary kernels produced with or without this P.R

PR #1011, opened 2021-06-25
[Navi21] Fix generic reduction kernels by removing __llvm_amdgcn_s_barrier (SWDEV-274384)

asroy:
remove __llvm_amdgcn_s_barrier(), which is not used but trigger compilation error because of function naming

https://ontrack-internal.amd.com/browse/SWDEV-274384

### DIFF
--- a/src/kernels/composable_kernel/include/utility/synchronization.hpp
+++ b/src/kernels/composable_kernel/include/utility/synchronization.hpp
@@ -5,8 +5,6 @@
 
 namespace ck {
 
-__device__ void __llvm_amdgcn_s_barrier() __asm("llvm.amdgcn.s.barrier");
-
 __device__ void block_sync_lds()
 {
 #if CK_BLOCK_SYNC_LDS_WITHOUT_SYNC_VMEM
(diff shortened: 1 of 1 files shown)


atamazov:
@asroy How urgent is this? Should be delivered to 4.6 latest?

qianfengz (approved):

atamazov:
@asroy @qianfengz Shall we keep the "volatile" W/A introduced in #998?

/cc @junliume

qianfengz:
Yes,  I think  "volatile" should be kept since  PR.1011 and PR.998 are for solving different issues.   PR.1011 is for addressing a building error occurred with the specific version of compiler (4261-STG2),  this building error does not occur with 7235-STG2 and it also does not occur with original version 3681 where SWDEV-274384 is reported.   In a summary,
1) Issues reported by SWDEV-291479 and  original SWDEV-274384 are the same issue (lost indices)
2) Issue occurred with 4261-STG2 is just a building error,  the reason 4261-STG2 is tested is just because the compiler team want me have a try with it for checking the "lost indices" issue

atamazov:
@qianfengz Thanks for the explanation.

Issue #1012, opened 2021-06-25
[OCL] ConvAsmImplicitGemmGTCDynamicWrwXdlops: "FAILED ... oclkernel.hpp:109: Error setting argument #19 to kernel"


atamazov:
Found during triaging of https://ontrack-internal.amd.com/browse/SWDEV-292684.

- Current develop, Staging, Mainline
- OpenCL backend only (HIP seems Ok)
- MI100 only
- Environment
   - Default Find mode (dynamic_hybrid) AND removed User-find-db
   - Or, `MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS=1 MIOPEN_FIND_MODE=normal`

Symptom: exception is thrown
```
FAILED: /dockerx/github/miopenx01/src/include/miopen/oclkernel.hpp:109: Error setting argument #19 to kernel (size = 4):  Invalid Arg Index
```

Some known failing configs
```
./bin/test_conv2d --float --cmode conv --pmode default --group-count 1 --input 128, 832, 7, 7 --weights 32, 832, 1, 1 --pads_strides_dilations 0 0 1 1 1 1
./bin/test_conv2d --float --cmode conv --pmode default --group-count 1 --input 64, 192, 28, 28 --weights 64, 192, 1, 1 --pads_strides_dilations 0 0 1 1 1 1
./bin/test_conv2d --float --cmode conv --pmode default --group-count 1 --input 64, 256, 28, 28 --weights 128, 256, 1, 1 --pads_strides_dilations 0 0 1 1 1 1
./bin/test_conv2d --float --cmode conv --pmode default --group-count 1 --input 64, 480, 14, 14 --weights 64, 480, 1, 1 --pads_strides_dilations 0 0 1 1 1 1
./bin/test_conv2d --float --cmode conv --pmode default --group-count 1 --input 64, 512, 14, 14 --weights 128, 512, 1, 1 --pads_strides_dilations 0 0 1 1 1 1
./bin/test_conv2d --float --cmode conv --pmode default --group-count 1 --input 64, 512, 28, 28 --weights 128, 512, 1, 1 --pads_strides_dilations 0 0 1 1 1 1
./bin/test_conv2d --float --cmode conv --pmode default --group-count 1 --input 64, 64, 56, 56 --weights 256, 64, 1, 1 --pads_strides_dilations 0 0 1 1 1 1
```

atamazov:
[...]

PR #1013, opened 2021-06-26
[HOTFIX][MI100][OCL] Remove unsed kernel arg for wrw fp32 NCHW

carlushuang:
This resolve issue https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1012, when under fp32 wrw, it seems feed in more kernel arg than needed. This seems OK under HIP, but in OCL it will report issue. But anyway we need keep consistent kernel arg with metadata declaration.

@shaojiewang please double check this fp32 wrw


shaojiewang (dismissed):

atamazov (changes requested):
Thanks for quick response. Constness of buffers also needs to be fixed, see https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1012#issuecomment-868895217.

What about regression tests?

atamazov:
> This seems OK under HIP.

I think this is not ok, https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1012#issuecomment-868896151

carlushuang:
> Constness of buffers also needs to be fixed, see #1012 (comment).
OK, let me add that part into this PR

> What about regression tests?
I guess this fix does not introduce regression? If can't work (e.g. after this PR fail on HIP), the CI will just fail

atamazov:
@carlushuang 
> > What about regression tests?
> 
> I guess this fix does not introduce regression? If can't work (e.g. after this PR fail on HIP), the CI will just fail

Or course, not. But this PR should add some regression tests that should prevent this bug in the future (see #1012). If you need assistance, please let me know and I'll add them to this branch.

carlushuang:
@atamazov ok sure, I think need your help on adding some tests in this branch, in case I'm not quite clear of the regression of this PR.
Thanks :)

atamazov:
@carlushuang 
> @atamazov ok sure, I think need your help on adding some tests in this branch, in case I'm not quite clear of the regression of this PR.  Thanks :)

Done in b6a3ff7. I would like to make this commit absolutely clear for you _before_ merging this. Please review this commit and compare it with the description of #1012. Ask if something is unclear.

carlushuang:
@atamazov the flag `--disable-validation` is added, because we don't care about the result is correct or not, but this test it self will fail (because of extra kernel arg) right?

atamazov:
We didn't have a MI100 test for OpenCL backend, so the test weren't run. dca7901 resolves this.

atamazov (dismissed):

atamazov (approved):

test/CMakeLists.txt
```diff
@@ -1169,3 +1169,24 @@ add_custom_test(test_regression_half_vega FLOAT_DISABLED HALF_ENABLED GFX908_DIS
...
+
+set(ENVS_REGRESSION_ISSUE_1012
+    MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS=1
+    MIOPEN_FIND_MODE=normal)
+
+set(ARGS_REGRESSION_ISSUE_1012
+    --verbose
+    --disable-forward
+    --disable-backward-data
+    --disable-validation)
```
atamazov, line 1181: :anchor:

PR #1016, opened 2021-06-28
[MLIR] Remove gemm size checks from ConvMlirIgemmBwd and ConvMlirIgemmBwdXdlops

yiqian1:
These checks have been moved to MLIR tuning parameter pass.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -118,7 +118,7 @@ RUN if [ "$USE_TARGETID" = "OFF" ] ; then echo "MIOpenTensile is not installed."
 
 RUN if [ "$USE_MLIR" = "ON" ]; \
     then cd ~ && \
-    export MLIR_COMMIT=950823986052e4750468e4e3a9641d0ce7be74a4 && \
+    export MLIR_COMMIT=96d97c20fdf46a2d2f38ed0d80e440513c0eb0cd && \
     wget https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/archive/$MLIR_COMMIT.tar.gz && \
     tar -xvzf $MLIR_COMMIT.tar.gz && \
     rm -rf $MLIR_COMMIT.tar.gz && \
--- a/src/solver/conv_mlir_igemm_bwd.cpp
+++ b/src/solver/conv_mlir_igemm_bwd.cpp
@@ -80,19 +80,6 @@ bool ConvMlirIgemmBwd::IsApplicable(const ConvolutionContext& ctx) const
     if(!IsComposableKernelSupportedHardware(ctx))
         return false;
 
-    const auto k = ConvolutionContextInterpreter::GetOutputChannelK(ctx);
-    if(k % GetEPackLength(ctx, false) != 0)
-        return false;
-
-    int gemm_m = 0;
-    int gemm_n = 0;
-    int gemm_k = 0;
-
-    std::tie(gemm_m, gemm_n, gemm_k) = CalculateGemmSize(ctx);
-
-    if(!(gemm_m % 32 == 0 && gemm_n % 32 == 0 && gemm_k % 4 == 0))
-        return false;
-
     return MiirIsConfigApplicable(
         mlir::ConstructBuildOptions(ctx, GetOperation(), GetKernelName(), false));
 #else
--- a/src/solver/conv_mlir_igemm_bwd.cpp
+++ b/src/solver/conv_mlir_igemm_bwd.cpp
@@ -38,27 +38,6 @@ namespace solver {
 
 namespace {
 #if MIOPEN_USE_MLIR
-std::tuple<int, int, int> CalculateGemmSize(const ConvolutionContext& ctx)
-{
-    const size_t g  = ConvolutionContextInterpreter::GetGroupCountG(ctx);
-    const size_t n  = ConvolutionContextInterpreter::GetBatchN(ctx);
-    const size_t k  = ConvolutionContextInterpreter::GetOutputChannelK(ctx);
-    const size_t c  = ConvolutionContextInterpreter::GetInputChannelC(ctx);
-    const size_t ho = ConvolutionContextInterpreter::GetOutputHeightHo(ctx);
-    const size_t wo = ConvolutionContextInterpreter::GetOutputWidthWo(ctx);
-    const size_t y  = ConvolutionContextInterpreter::GetFilterHeightY(ctx);
-    const size_t x  = ConvolutionContextInterpreter::GetFilterWidthX(ctx);
-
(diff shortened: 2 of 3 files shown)


jerryyin (changes requested):
Missing another solver: conv_mlir_igemm_bwd_xdlops.cpp

You also need to bump the MLIR commit after your MLIR PR is merged in Dockerfile.

yiqian1:
conv_mlir_igemm_bwd_xdlops.cpp doesn't have the same checks. It does call IsValidXdlopsGemm(). Do you mean IsValidXdlopsGemm() should be removed?

jerryyin:
@yiqian1 Correct, any non-trivial checks in any solver needs to be removed and make its way into MLIR. Right now there is all except the two bwd solvers that have those checks.

yiqian1:
Got it. I'll work on it.

atamazov:
Converted to draft.

jerryyin (approved):
LGTM, thanks!

yiqian1:
jenkins retest this please

yiqian1:
jenkins: retest this please

atamazov:
How this affects performance of `IsApplicable()`? It must be pretty fast, otherwise the performance of the library will suffer. Measurements? I can recommend collecting logs with `export MIOPEN_ENABLE_LOGGING_ELAPSED_TIME=1` and comparing them. 

Or you can temporarily instrument code with with `MIOPEN_LOG_SCOPE_TIME`, see #622.

atamazov:
@yiqian1 

>  yiqian1 force-pushed the cleanupMLIRBwdSolver branch from 3625018 to 37fef14 23 minutes ago

This is NOT recommended due to several reasons. Force-pushing can be more or less safely used only if the branch has been published for review yet.

atamazov:
@yiqian1 

> jenkins: retest this please

Hmm. Jenkins automatically begins a new job after branch is updated and pushed to the repo.

In order to restart the job after accidental failure (without branch update), you need to visit micimaster.amd.com, login and click here: 
image

yiqian1:
>How this affects performance of IsApplicable()? It must be pretty fast, otherwise the performance of the library will suffer. Measurements? I can recommend collecting logs with export MIOPEN_ENABLE_LOGGING_ELAPSED_TIME=1 and comparing them.



The changes are removing code from IsApplicable(), so they would make IsApplicable() faster, though the real impact of the change might be too trivial to have any meaningful impact on the performance of kernels. For example, I tested one config:

The original code gave:
MIOpen Backward Data Conv. Algorithm: 5, Solution: 99/ConvMlirIgemmBwd
GPU Kernel Time Backward Data Conv. Elapsed: 2.264169 ms (average)
stats: bwdd-conv1x1u2, 256, 128, 1, 1, 2048, 7, 7,  6576668672, 269484032, 102760448, 2905, 164, 2.264169
Backward Convolution Data Verifies OK on GPU reference (9.19717e-08)

The new code gave:
MIOpen Backward Data Conv. Algorithm: 5, Solution: 99/ConvMlirIgemmBwd
GPU Kernel Time Backward Data Conv. Elapsed: 2.260250 ms (average)
stats: bwdd-conv1x1u2, 256, 128, 1, 1, 2048, 7, 7,  6576668672, 269484032, 102760448, 2910, 165, 2.260250
Backward Convolution Data Verifies OK on GPU reference (9.19717e-08)

I don't think it's necessary to collect and compare the performance for all test cases, as it's obvious that there won't be negative impact on performance. However if it is the standard way for merging, I will definitely follow it. Please let me know.

atamazov:
@yiqian1 

> The changes are removing code from IsApplicable(), so they would make IsApplicable() faster, though the real impact of the change might be too trivial to have any meaningful impact on the performance of kernels. 

This is not so.
- Before this change, the IsApplicable was able to return `false` without need to call `MiirIsConfigApplicable()`. How MIOpen does not have this shortcut.
- This change includes update of `MLIR_COMMIT`. If I understand correctly, the updated MLIR performs some additional job in `MiirIsConfigApplicable()`.

atamazov:
> For example, I tested one config:...

Please note that the logs do not include information I've asked for. I need to see how the performance of `IsApplicable()` changed, taken in isolation.

atamazov (changes requested):
Unfortunately, I still need an answer for https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1016#issuecomment-884536638

yiqian1:
The checks that were removed from IsApplicable() were added in MiirIsConfigApplicable(). So instead of returning false from IsApplicable(), now MiirIsConfigApplicable() would return false. These changes are conceptually same as PR#1026. Let me try to get the logs of the isolated performance of IsApplicable().

yiqian1:
Here is what I got. 

These changes would degrade the performance of IsApplicable() by 20x for an invalid configuration. For example, I tested an invalid configuration which would fail the checks that were removed by this PR. The original IsApplicable() takes 12us due to the short-cut check. With the changes, IsApplicable() takes 2397us, as it needs to call the library. 

For applicable configurations (I used the configurations in test_conv_igemm_mlir, and each config ran twice) on gfx900,  there was no significant performance change of IsApplicable(). Without the changes, the average time of IsApplicable() is 941us. With the changes, the average time is 845us.

jerryyin:
@yiqian1 To add in a few background information: Very soon when the padding kernel becomes available, we'd no longer need to call into the API of `miirLowerTuningParams()`, but instead only leave the handle creation in the applicability checks. I believe 90% of time in that 2397 us is because of the tuning parameter lowering, which is known to be more expensive. Would you be able to deduct the time of the tuning paramter lowering API call and see the fair comparison in this context?

yiqian1:
> I believe 90% of time in that 2397 us is because of the tuning parameter lowering, which is known to be more expensive. Would you be able to deduct the time of the tuning paramter lowering API call and see the fair comparison in this context?

Yes. My experiment shows miirLowerTuningParams() takes 60% execution time of IsApplicable().

yiqian1:
@atamazov Please let me know if you think the performance of IsApplicable() is acceptable. Jerry and I had a discussion and believe we can improve the performance of IsApplicable(). This needs some investigation and will be in another PR.

atamazov:
> @atamazov Please let me know if you think the performance of IsApplicable() is acceptable.

Unfortunately, not. IsApplicable is used in many contexts. It must be really really fast.

> Jerry and I had a discussion and believe we can improve the performance of IsApplicable(). This needs some investigation and will be in another PR.

We can merge this as is, provided that a performance regression ticket is created in this repo. If you want to go that way, then please go ahead and create a ticket, and let me know.

atamazov (approved):
LGTM!

PR #970, opened 2021-06-07
Jenkins ci rocm 4.2

shurale-nkn:
- Use ROCm 4.2 for all but COMGR and MIOpenTensile
- MIOpenTensile path stays at 4.1.1
- MLIR builds remain at 3.7 due to missing TargetId support in MLIR
- Disable some new tidy checks
- Hack away concurrency-mt-unsafe warnings
- WORKAROUND_ISSUE_898
- WORKAROUND_ISSUE_936

PR #1019, opened 2021-06-29
[TESTS] Resolve PR #970 review comments

shurale-nkn:
resolve https://github.com/ROCmSoftwarePlatform/MIOpen/pull/970#pullrequestreview-695472994


atamazov (approved):

atamazov:
@shurale-nkn Smoke testing is enough for this PR as it may affect only COMGR testing. Please ping me when smoke tests passed and I'll merge this.

shurale-nkn:
> @shurale-nkn Smoke testing is enough for this PR as it may affect only COMGR testing. Please ping me when smoke tests passed and I'll merge this.

@atamazov 
Ready

test/CMakeLists.txt
```diff
@@ -160,7 +160,7 @@ function(add_test_command NAME EXE)
         endif()
     endif()
     
-    if(WORKAROUND_ISSUE_898 AND MIOPEN_BACKEND_USE_COMGR)
+    if(WORKAROUND_ISSUE_898 AND MIOPEN_USE_COMGR)
         set_property(TEST ${TEST_NAME} PROPERTY ENVIRONMENT MIOPEN_DEBUG_COMGR_HIP_PCH_ENFORCE=0)
```
atamazov, line 164: @shurale-nkn It should be `set_property(TEST ${NAME}...` isn't it? I am wondering how this works...

shurale-nkn: @atamazov 
Yes, in this function better to use ${NAME}.

But it works because function can access to scope defined before the moment then function was called.
So add_test_command was called from add_test_executable where TEST_NAME is defined as argument, and because add_test_executable is not finished the scope is not deleted. And the TEST_NAME variable accessible from add_test_command.

atamazov: Oh great CMake syntax!

PR #1023, opened 2021-07-07
[NFC] Activation backward solver 0

DrizztDoUrden:
It implements the first backward activation solver, complete with an invoker.
Performance testing is complete and doesn't show any difference to develop.

### DIFF
--- a/src/activ/problem_description.cpp
+++ b/src/activ/problem_description.cpp
@@ -60,8 +60,12 @@ NetworkConfig ProblemDescription::MakeNetworkConfig() const
     std::ostringstream ss;
 
     ss << "activ-";
+
     if(direction == Direction::Backward)
         ss << "bwd-";
+    else
+        ss << "fwd-";
+
     ss << ((packed) ? "11" : "10"); // + lite bit
     ss << xDesc.GetType();
     ss << activDesc.GetMode();
--- a/src/activ/problem_description.cpp
+++ b/src/activ/problem_description.cpp
@@ -60,6 +60,8 @@ NetworkConfig ProblemDescription::MakeNetworkConfig() const
     std::ostringstream ss;
 
     ss << "activ-";
+    if(direction == Direction::Backward)
+        ss << "bwd-";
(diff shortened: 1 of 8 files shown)


atamazov:
@DrizztDoUrden test_gru has failed.

src/activ/problem_description.cpp
```diff
@@ -60,6 +60,8 @@ NetworkConfig ProblemDescription::MakeNetworkConfig() const
     std::ostringstream ss;
 
     ss << "activ-";
+    if(direction == Direction::Backward)
```
JehandadKhan, line 63: Can we add the else branch too? That would make parsing these easier

DrizztDoUrden: Sure!

src/include/miopen/activ/problem_description.hpp
```diff
@@ -45,18 +45,35 @@ enum class Direction
...
+          dxDesc(dxDesc_),
+          dyDesc(dyDesc_)
     {
     }
 
     Direction GetDirection() const { return direction; }
     const ActivationDescriptor& GetActivDesc() const { return activDesc; }
     const TensorDescriptor& GetXDesc() const { return xDesc; }
     const TensorDescriptor& GetYDesc() const { return yDesc; }
+    const TensorDescriptor& GetDXDesc() const { return xDesc; }
```
JehandadKhan, line 75: Should this check for the direction ?

DrizztDoUrden: Can do that, but for bwd it's inited with {}.

DrizztDoUrden: I think better way if we want to change it would be to make it boost::optional<TensorDescriptor>.

atamazov: @DrizztDoUrden Let's move forward. Please make _separate_ PR with these changes.

atamazov (approved):
LGTM!

Issue #985, opened 2021-06-16
[TESTS] Implement detection of GPU type


atamazov:
This could be implemented in CMake. According to that, the default values of CMake variables like `MIOPEN_TEST_GFX908` should be set.

Right now we at MIOpen use node labeling for that. However, the users can't simply run "make check", they need to check the GPU type and set a CMake variable accordingly.

This is not good, especially for our CQE, who use the same script for all GPU types. For now, I am recommending the following W/A in the testing script:
```bash
tempcmd="/opt/rocm/opencl/bin/clinfo | grep gfx908"; eval "$tempcmd"; if [[ $? -ne 0 ]] ; then export MIOPEN_TEST_GFX908_VALUE=Off; else export MIOPEN_TEST_GFX908_VALUE=On; fi
...
cmake ... -DMIOPEN_TEST_GFX908=${MIOPEN_TEST_GFX908_VALUE} ...
```

Originated from https://ontrack-internal.amd.com/browse/SWDEV-290980?focusedCommentId=7474697&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-7474697

JehandadKhan:
How about checking the GPU type from the handle in the test and return success with a message which says that the test is skipped since there is no GPU for it.

atamazov:
This wouldn't work for "custom" tests, which are basically sets of executables with options...

atamazov:
@pfultz2 Can you please advice how can we detect GPU type from CMake?

pfultz2:
> Can you please advice how can we detect GPU type from CMake?

You can run `execute_process` to run a command that shows what gpus are installed via rocminfo. Of course, I dont know how easy it is to parse that.

Alternatively, we could update the test driver to only run on certain archs. We could add a flag like `--only-arch` that will skip the tests if the arch doesn't match. Such a flag can be added to the custom tests.

atamazov:
@pfultz2 
> You can run `execute_process` to run a command that shows what gpus are installed via rocminfo. Of course, I dont know how easy it is to parse that.

Thanks for the recommendation. Theoretically we can grep output and check the status code, but I am now aware of CMake-specific implementation details.

> Alternatively, we could update the test driver to only run on certain archs. We could add a flag like `--only-arch` that will skip the tests if the arch doesn't match. Such a flag can be added to the custom tests.

[...]

PR #1024, opened 2021-07-08
[TESTS] Detection of GPU type (SWDEV-290980)

atamazov:
Implements #985 

* [TESTS] Added detection of GPU type for testing.
  * Explicit setting of `MIOPEN_TEST_GFX*` overrides it (for backward compatibility).
* [Jenkinsfile] Removed explicit setting of the MIOPEN_TEST_GFX908 flag because it is not required anymore.
* [Dockerfile] Installed `kmod` as it is required by `rocminfo` to work.

## :warning: QUIET CHANGES
- The default GPU for testing was gfx900/906 (Vega) before this PR. Now the default GPU for testing is identified by grepping the output of `rocminfo`.
- If `rocminfo` is not available or fails, then `make check` fails immediately. In this case the user need to explicitly specify `MIOPEN_TEST_GFX908=On` or `MIOPEN_TEST_VEGA=On` etc. _Before this PR_, Vega testing was enabled by default in this case.
- Ditto if detected GPU is not known to the test/CMakeLists.txt.
    :red_circle: Both Navi21 and gfx90a are NOT KNOWN to the tests yet!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,9 +150,10 @@ if( MIOPEN_BACKEND STREQUAL "OpenCL")
     set(MIOPEN_BACKEND_OPENCL 1)
     find_package( OpenCL REQUIRED )
     find_program(MIOPEN_HIP_COMPILER hcc
-            PATH_SUFFIXES bin
-            PATHS 
-	        /opt/rocm
+        PATH_SUFFIXES bin
+        PATHS
+            ${ROCM_ROOT}
+            /opt/rocm
     )
     set(MIOPEN_USE_MIOPENGEMM ON CACHE BOOL "")
     if(MIOPEN_HIP_COMPILER)
@@ -160,9 +161,10 @@ if( MIOPEN_BACKEND STREQUAL "OpenCL")
     else()
         find_program(MIOPEN_HIP_COMPILER clang++
             PATH_SUFFIXES bin
-	        PATHS 
-		    /opt/rocm/llvm
-		    ${CMAKE_INSTALL_PREFIX}/llvm
+	        PATHS
+                ${ROCM_ROOT}/llvm
+    		    /opt/rocm/llvm
+	    	    ${CMAKE_INSTALL_PREFIX}/llvm
         )
         if(MIOPEN_HIP_COMPILER)
             message("hip compiler: ${MIOPEN_HIP_COMPILER}")
@@ -187,7 +189,7 @@ endif()
 
 
 # HIP is always required
-find_package(hip REQUIRED PATHS /opt/rocm)
+find_package(hip REQUIRED PATHS ${ROCM_ROOT} /opt/rocm)
 message(STATUS "Build with HIP ${hip_VERSION}")
 target_flags(HIP_COMPILER_FLAGS hip::device)
 # Remove cuda arch flags
@@ -233,10 +235,10 @@ if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_B
 
     find_program(HIP_OC_COMPILER clang-ocl
         PATH_SUFFIXES bin
-        PATHS 
-	    /opt/rocm
-	    ${CMAKE_INSTALL_PREFIX}
-
+        PATHS
+            ${ROCM_ROOT}
+    	    /opt/rocm
+    	    ${CMAKE_INSTALL_PREFIX}
     )
     if(HIP_OC_COMPILER)
         message(STATUS "hip compiler: ${HIP_OC_COMPILER}")
@@ -272,7 +274,7 @@ if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_B
     # rocblas
     set(MIOPEN_USE_ROCBLAS ON CACHE BOOL "")
     if(MIOPEN_USE_ROCBLAS)
-        find_package(rocblas REQUIRED PATHS /opt/rocm)
+        find_package(rocblas REQUIRED PATHS ${ROCM_ROOT} /opt/rocm)
         message(STATUS "Build with rocblas")
     else()
         message(STATUS "Build without rocblas")
@@ -284,9 +286,10 @@ message( STATUS "${MIOPEN_BACKEND} backend selected." )
 if(MIOPEN_HIP_COMPILER MATCHES ".*clang\\+\\+$")
     find_program(MIOPEN_OFFLOADBUNDLER_BIN clang-offload-bundler
         PATH_SUFFIXES bin
-        PATHS 
-	    /opt/rocm/llvm
-	    ${CMAKE_INSTALL_PREFIX}/llvm
+        PATHS
+            ${ROCM_ROOT}/llvm
+	        /opt/rocm/llvm
+	        ${CMAKE_INSTALL_PREFIX}/llvm
     )
 endif()
 if(MIOPEN_OFFLOADBUNDLER_BIN)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -299,13 +302,15 @@ else()
     find_program(EXTRACTKERNEL_BIN extractkernel
         PATH_SUFFIXES bin
         PATHS
+            ${ROCM_ROOT}/hip
+            ${ROCM_ROOT}/hcc
+            ${ROCM_ROOT}
```
pfultz2, line 307: Remove ROCM_ROOT variable, the CMAKE_PREFIX_PATH(or CMAKE_INSTALL_PREFIX) should be used when rocm is not installed into /opt/rocm or standard directories(ie /usr or /usr/local).

test/CMakeLists.txt
```diff
@@ -44,6 +44,45 @@ option( MIOPEN_TEST_MLIR "Add tests for MLIR -- EXPERIMENTAL" ${MIOPEN_USE_MLIR}
 option( WORKAROUND_ISSUE_898 "" ON)
 option( WORKAROUND_ISSUE_936 "" ON)
 
+# Detect GPU type for testing
+if(NOT (MIOPEN_TEST_VEGA OR MIOPEN_TEST_GFX908))
+    find_program(ROCMINFO
+        NAMES rocminfo
+        PATHS
+            ${ROCM_ROOT}
+            /opt/rocm
```
pfultz2, line 53: These additional paths(even for /opt/rocm) is not needed because we already append these paths to the `CMAKE_PREFIX_PATH`).

test/CMakeLists.txt
```diff
@@ -44,6 +44,45 @@ option( MIOPEN_TEST_MLIR "Add tests for MLIR -- EXPERIMENTAL" ${MIOPEN_USE_MLIR}
...
+        PATH_SUFFIXES
+            /bin
+        NO_DEFAULT_PATH
```
pfultz2, line 56: Remove the `NO_DEFAULT_PATH` as this prevents the user specifying dependencies with `CMAKE_PREFIX_PATH`.

test/CMakeLists.txt
atamazov: It doesn't work on my machine. Does it work for you?

pfultz2: I think that is because you are using NO_DEFAULT_PATH.

CMakeLists.txt
atamazov: Done.

test/CMakeLists.txt
atamazov: Yeah, thanks. Made similar to ./CMakeLists.txt

test/CMakeLists.txt
atamazov: Done.

atamazov:
@shurale-nkn Thanks for e459cdd!

atamazov:
@pfultz2 Ready for another review round.

CMakeLists.txt
```diff
@@ -342,7 +340,6 @@ find_program(MIOPEN_AMDGCN_ASSEMBLER
         /opencl/bin/x86_64
         /opencl/bin
         /bin
-    NO_DEFAULT_PATH
```
atamazov, line 345: @pfultz2 It is good change?

test/CMakeLists.txt
```diff
@@ -44,6 +44,44 @@ option( MIOPEN_TEST_MLIR "Add tests for MLIR -- EXPERIMENTAL" ${MIOPEN_USE_MLIR}
...
+            /opt/rocm
+            ${CMAKE_INSTALL_PREFIX}
+        PATH_SUFFIXES
+            /bin
+    )
+    message(STATUS "rocminfo utility: ${ROCMINFO}")
+    if(ROCMINFO)
+        execute_process (
+            COMMAND ${ROCMINFO}
+            COMMAND grep -i -E "(gfx900)|(gfx906)"
```
pfultz2, line 61: Why is the grep necessary? Why not just do `if(MATCHES)` on the output?

test/CMakeLists.txt
```diff
@@ -44,6 +44,44 @@ option( MIOPEN_TEST_MLIR "Add tests for MLIR -- EXPERIMENTAL" ${MIOPEN_USE_MLIR}
...
+            RESULT_VARIABLE ROCMINFO_VEGA
+            OUTPUT_QUIET
+        )
+        execute_process (
+            COMMAND ${ROCMINFO}
```
pfultz2, line 66: No need to run this twice. Just run it once, and save the output to cmake variables. Then use two `if(MATCHES)` to check if the variable matches the regex.

test/CMakeLists.txt
atamazov: Who knows? Maybe I simply like grep more than all other tools))

test/CMakeLists.txt
atamazov: let me try

test/CMakeLists.txt
atamazov: ...but capturing once + if(MATCHES) provide more compact code. Resolved.

test/CMakeLists.txt
atamazov: It works.

atamazov:
Ready for another review round.

shurale-nkn (dismissed):
LGTM

test/CMakeLists.txt
```diff
@@ -64,6 +64,42 @@ if(MIOPEN_BACKEND_HIP AND NOT MIOPEN_TEST_HIP_NOGPU)
...
+        endif()
+        if(ROCMINFO_OUTPUT MATCHES "gfx900|gfx906")
+            set(MIOPEN_TEST_VEGA ON)
+        elseif(ROCMINFO_OUTPUT MATCHES "gfx908")
+            set(MIOPEN_TEST_GFX908 ON)
+        else()
+            message(WARNING "TESTING IS NOT SUPPORTED FOR THE DETECTED GPU")
+        endif()
+    else()
+        message(FATAL_ERROR "ROCMINFO NOT FOUND, GPU TYPE UNKNOWN. Manually set respective MIOPEN_TEST_GFX* CMake variable to specify target GPU for testing.")
```
pfultz2, line 97: We probably dont want to make this a fatal error because this will fail when the user just wants to build MIOpen and not run the tests(perhaps the user builds on a machine with no gpus and then copies the binaries to the machines with GPUs).

If you want an error then you can add a test which prints out the message and then fails:

```cmake
add_test(NAME rocminfo_check COMMAND echo "ROCMINFO FAILED: ...")
set_tests_properties(rocminfo_check PROPERTIES FAIL_REGULAR_EXPRESSION "FAILED")
```

Or you could add a custom target that `tests` depends on and prints out a failed message as well(this will show an error before running the tests):

```cmake
add_custom_target(rocminfo_check COMMAND echo "ROCMINFO FAILED ..." COMMAND exit 1)
add_dependencies(tests rocminfo_check)
```

atamazov: @pfultz2 Good idea, let me try that...

atamazov: @pfultz2 Yep, that's better. Thanks for the idea.

atamazov:
@shurale-nkn @pfultz2 Ready for the next review round.

test/CMakeLists.txt
```diff
@@ -64,6 +64,45 @@ if(MIOPEN_BACKEND_HIP AND NOT MIOPEN_TEST_HIP_NOGPU)
...
+    message(STATUS "rocminfo utility: ${ROCMINFO}")
+    if(ROCMINFO)
+        execute_process (
+            COMMAND ${ROCMINFO}
+            OUTPUT_VARIABLE ROCMINFO_OUTPUT
+            RESULT_VARIABLE ROCMINFO_EXIT_STATUS
+        )
+        if(NOT ROCMINFO_EXIT_STATUS EQUAL 0)
+            message(WARNING "ROCMINFO FAILED, GPU TYPE UNKNOWN. Manually set respective MIOPEN_TEST_GFX* CMake variable to specify target GPU for testing.")
+            set(MIOPEN_TEST_GPU_DETECTION_FAILED TRUE)
```
pfultz2, line 89: Maybe this could be done using a function rather than setting a cmake variable?

atamazov: Yes, this can be done, but the function can be called only after `add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -C ${CMAKE_CFG_INTDIR})`. The `MIOPEN_TEST_GPU_DETECTION_FAILED` variable which holds the status of detection provides more flexibility. Let me keep it as is for now.

atamazov:
CI tests passed.

test/CMakeLists.txt
pfultz2: Sure

pfultz2 (approved):

atamazov:
@pfultz2 Thanks for reviewing!

PR #1026, opened 2021-07-08
[MLIR] Remove validations that have been moved into MLIR.

krzysz00:
Some of the checks that the MLIR solvers perform can be moved into
MLIR, allowing MLIR to enable support for more types of convolution
without needing to coordinate as many changes with MIOpen. This also
reduces redundancy in the solver applicability checks.

### DIFF
--- a/src/solver/mlir_common.cpp
+++ b/src/solver/mlir_common.cpp
@@ -24,9 +24,11 @@
  *
  *******************************************************************************/
 
+#include "miopen/miopen.h"
 #include <miopen/errors.hpp>
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/solver/mlir_common.hpp>
+#include <string>
 
 namespace miopen {
 namespace solver {
@@ -41,6 +43,30 @@ std::string InsertGToLayout(const std::string& layout, char dim)
     return layout_with_g.insert(index, 1, 'G');
 }
 
+const char* DTypeName(miopenDataType_t ty) {
+    switch (ty) {
+    case miopenHalf:
+        return "fp16";
+    case miopenFloat:
+        return "fp32";
+    case miopenDouble:
+        return "fp64";
+    case miopenBFloat16:
+        return "bf16";
+    case miopenInt32:
+        return "i32";
+    case miopenInt8:
+        return "i8";
+    case miopenInt8x4:
+        return "i8x4";
+    }
+    assert(false); // All cases should be covered before here
+}
+
+/* Construct the options string passed to MLIR to cause it
+to generate a given convolution.
+
+Returns an empty string on unsupported convolutions */
 std::string ConstructBuildOptions(const ConvolutionContext& ctx,
                                   const std::string& operation,
                                   const std::string& kernel_name,
@@ -56,10 +82,13 @@ std::string ConstructBuildOptions(const ConvolutionContext& ctx,
     std::string out_layout = InsertGToLayout(CI::GetOutputLayout(ctx), 'C');
 
     std::string mlir_handle;
-    if (is_xdlops)
-        mlir_handle += std::string(" --x2 ") + "1";
+    if (!ctx.Is2d()) {
+        // Future: Remove this once MLIr supports 3D convolutions
+        return mlir_handle;
+    }
 
-    std::string data_type = ctx.IsFp32() ? "fp32" : "fp16";
+    if (is_xdlops)
+        mlir_handle += std::string(" --x2 1");
 
     mlir_handle +=
         std::string(" --operation ") + operation +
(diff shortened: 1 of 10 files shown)


krzysz00:
This PR depends on [MLIR#298](https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/298), so this is here for review and coordination.

atamazov:
@krzysz00 Is it ready for merging? If not, please convert to draft.

src/solver/mlir_common.cpp
```diff
@@ -41,6 +43,30 @@ std::string InsertGToLayout(const std::string& layout, char dim)
     return layout_with_g.insert(index, 1, 'G');
 }
 
+const char* DTypeName(miopenDataType_t ty) {
```
jerryyin, line 46: I know in all scenarios you are returning a string literal with static storage duration. However I'd prefer this to be returning a string so client of this function never worry about the implementation details of this one.

src/solver/mlir_common.cpp
```diff
@@ -41,6 +43,30 @@ std::string InsertGToLayout(const std::string& layout, char dim)
...
+    case miopenDouble:
+        return "fp64";
+    case miopenBFloat16:
+        return "bf16";
+    case miopenInt32:
+        return "i32";
+    case miopenInt8:
+        return "i8";
+    case miopenInt8x4:
+        return "i8x4";
```
jerryyin, line 61: @whchung FYI, we are creating this 1:1 mapping string for data type representations. Right now, all except fp16, fp32 should be rejected by MLIR. To enable a new data type wouldn't require any change to MIOpen in the future.

src/solver/mlir_common.cpp
```diff
@@ -56,10 +82,13 @@ std::string ConstructBuildOptions(const ConvolutionContext& ctx,
     std::string out_layout = InsertGToLayout(CI::GetOutputLayout(ctx), 'C');
 
     std::string mlir_handle;
-    if (is_xdlops)
-        mlir_handle += std::string(" --x2 ") + "1";
+    if (!ctx.Is2d()) {
+        // Future: Remove this once MLIr supports 3D convolutions
+        return mlir_handle;
+    }
```
jerryyin, line 88: Can we move this to MLIR?

jerryyin (changes requested):
Some minor nits. Overall good job. One missing thing is that you'd need to bump the MLIR commit in MIOpen's Dockerfile once your MLIR PR is merged.

src/solver/mlir_common.cpp
krzysz00: Well, if we wanted to add a series of `--*_depth` options to MLIR and then reject on depth != 0, then we could. Is that a preferable refactor.

src/solver/mlir_common.cpp
krzysz00: On the other hand, if we're ever using anything but string literals, we can refactor, and this has the advantage of making it clear that it's meant to be a static string literal.

src/solver/mlir_common.cpp
jerryyin: Can we do an additional check on layout to filter off those that whose length are not 5?

src/solver/mlir_common.cpp
jerryyin: Fine by me.

src/solver/mlir_common.cpp
jerryyin: On a second look you already have it in the MLIR PR. Please drop this check.

krzysz00: ... yeah, we do have a rather strict layout filter, and so could filter out 3D that way. Sure.

jerryyin:
Considering that e4946eb has a full passing record whereas regression appears on f0457a2. It is likely a regression has already been introduced in MLIR end in the `miopen-dialect` branch. Could you prioritize a fix or revert on the MLIR side?

jerryyin (approved):
No further opinions provided that all MLIR CI continue to pass.

src/solver/mlir_common.cpp
```diff
@@ -68,11 +86,11 @@ std::string ConstructBuildOptions(const ConvolutionContext& ctx,
         std::string(" --arch ") + ctx.GetStream().GetDeviceName() +
         std::string(" --groupsize ") + std::to_string(CI::GetGroupCountG(ctx)) +
         std::string(" --fil_layout ") + fil_layout +
-        std::string(" --fil_type ") + data_type +
+        std::string(" --fil_type ") + DTypeName(ctx.weights_data_type) +
```
atamazov, line 89: [future] This is cumbersome. Use ostringstream, them return `.str()`.

atamazov (approved):
This can be merged. Please note that CI performs only basic testing, so I highly recommend supplementing CI testing with rigorous manual validation.

Please let me know if you would like me to merge this as is.

src/solver/mlir_common.cpp
krzysz00: Agreed and fixed

src/solver/mlir_common.cpp
```diff
@@ -41,6 +43,23 @@ std::string InsertGToLayout(const std::string& layout, char dim)
...
+    {
+    case miopenHalf: return "fp16";
+    case miopenFloat: return "fp32";
+    case miopenDouble: return "fp64";
+    case miopenBFloat16: return "bf16";
+    case miopenInt32: return "i32";
+    case miopenInt8: return "i8";
+    case miopenInt8x4: return "i8x4";
+    }
+    assert(false); // All cases should be covered before here
```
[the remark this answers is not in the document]
krzysz00: Thanks for the FYI - fixed

krzysz00:
@atamazov If you have no other comments, please feel free to merge

src/solver/mlir_common.cpp
```diff
@@ -41,56 +45,74 @@ std::string InsertGToLayout(const std::string& layout, char dim)
...
+    {
+    case miopenHalf: return "fp16";
+    case miopenFloat: return "fp32";
+    case miopenDouble: return "fp64";
+    case miopenBFloat16: return "bf16";
+    case miopenInt32: return "i32";
+    case miopenInt8: return "i8";
+    case miopenInt8x4: return "i8x4";
+    }
+    MIOPEN_THROW("Internal error: value outside of datatype enum");
```
atamazov, line 60: [Recommendation] Here's what I meant:
```suggestion
    MIOPEN_THROW(miopenStatusInternalError, "Value outside of datatype enum");
```

src/solver/mlir_common.cpp
```diff
@@ -24,10 +24,14 @@
  *
  *******************************************************************************/
 
+#include "miopen/miopen.h"
```
atamazov, line 27:
```suggestion
#include <miopen/miopen.h>
```

atamazov (changes requested):

src/solver/mlir_common.cpp
atamazov: This needs to be fixed, sorry.

atamazov (approved):
LGTM!

PR #1040, opened 2021-07-19
[ASM iGEMM][NHWC] Fix for potential Memory Access Fault or NaN result (added v_flag missed during merging)

carlushuang:
fix a ~potential~ bug in merge kernels, that there is a missing v_flag while loading. If not add this v_flag, a out-of-bound read will happen and may lead to a GPU crash (Memory access fault) OR computation error (NaN -> model explosion).


atamazov:
@carlushuang Is it only NHWC related?

carlushuang:
> @carlushuang Is it only NHWC related?

Yes, and only a part of NHWC kernel related.

shaojiewang:
Maybe some test cases could be added for this PR.

carlushuang:
@shaojiewang added in 2b439bd069dcef76235446423793cda23ff0309e

shaojiewang (approved):
LGTM

atamazov:
@carlushuang Just to double-check: Does the added test fail without the fix?

carlushuang:
> @carlushuang Just to double-check: Does the added test fail without the fix?

Yes, but it will fail starting from rocm-4.1.  Before this it will fail randomly.

atamazov:
> Yes, but it will fail starting from rocm-4.1. Before this it will fail randomly. By adding the fix in this PR, this test case will not fail.

It's Ok because our CI already uses 4.2.

atamazov (approved):
I do not see anything bad.

atamazov:
@carlushuang 
> out-of-bound read will happen and lead to potential GPU crash.

Does it also lead to inevitable computation error?

atamazov:
I need to know that before merging.

carlushuang:
> Does it also lead to inevitable computation error?

@atamazov theoretically it will have computation error, in case of Nan or Inf multiply by zero. Other case it will not.
This is because out-of-bound read error is in weight tensor, but in input tensor we will set the value to zero properly if it is out-of-bound. If gpu does not crash, and weight tensor get an unwanted oob value, if it is not Nan then multiply by zero we will get zero, so everything is just fine. If it is Nan or Inf, multiply by zero we will get Nan.
In short, the computation error depends on what value we get by oob read (if gpu not crash)

atamazov:
@carlushuang Thanks. Description updated.

PR #1043, opened 2021-07-22
[TESTS] Set MIOPEN_TEST_LIMIT to 2 by default (SWDEV-295448)

shurale-nkn:
https://ontrack-internal.amd.com/browse/SWDEV-295448

### DIFF
--- a/test/driver.hpp
+++ b/test/driver.hpp
@@ -175,7 +175,7 @@ struct test_driver
     std::string cache_path = compute_cache_path();
     miopenDataType_t type  = miopenFloat;
     bool full_set          = false;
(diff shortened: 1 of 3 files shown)


test/driver.hpp
```diff
@@ -175,7 +175,7 @@ struct test_driver
     std::string cache_path = compute_cache_path();
     miopenDataType_t type  = miopenFloat;
     bool full_set          = false;
-    int limit_set          = 0;
+    int limit_set          = 2;
```
atamazov, line 178: The purpose of changes in this file is to keep backward compatibility?

atamazov (approved):

PR #982, opened 2021-06-14
[MI100]fix issue 980-fix kernel selection logic for k-padding cases

shaojiewang:
Fix issue #980

Issue #995, opened 2021-06-21
[TESTS][Fp16][MI100] Incorrect test case in test_conv_igemm_dynamic_xdlops_bwd and wrw


atamazov:
BWD:
```
$ MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicBwdXdlops \
./bin/test_conv2d --half --cmode conv --pmode default --group-count 1 --disable-forward --disable-backward-weights \
--input 4, 512, 128, 128 --weights 12, 512, 1, 1 --pads_strides_dilations 0 0 1 1 1 1 --trans_output_pads 0 0
...
FAILED: /dockerx/github/miopenx01/src/ocl/convolutionocl.cpp:1216: Backward Data Convolution cannot be executed due to incorrect params
```

WRW:
<details><summary><b><i>CONSOLE LOGS</i></b></summary>

```
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [GetFindModeValueImpl] MIOPEN_FIND_MODE = NORMAL(1)
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [get_device_name] Raw device name: gfx908
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [Handle] stream: 0x28210d0, device_id: 1
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [HipCompilerVersionImpl] 3.7.20315
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, MIOpen version 2.13.0.8492-9ebbe9fde
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [BackwardDataGetWorkSpaceSize] 
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [GetEnvFindOnlySolverImpl] 72
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [ForwardGetWorkSpaceSize] 
[2021-06-21T16:19:29.911Z] MIOpen(HIP): Info [BackwardWeightsGetWorkSpaceSize] 
[...]

PR #1048, opened 2021-07-23
[HOTFIX][WORKAROUND] Revert #982 -- partial W/A for FP16 SSD issue (memory access fault in SWDEV-295434)

atamazov:
Reverts #982. Partial W/A for FP32 SSD issue (memory access fault in https://ontrack-internal.amd.com/browse/SWDEV-295434).

:warning: Merging this will cause #980 to appear again!

### DIFF
--- a/src/solver/conv_asm_implicit_gemm_wrw_gtc_dynamic_xdlops.cpp
+++ b/src/solver/conv_asm_implicit_gemm_wrw_gtc_dynamic_xdlops.cpp
@@ -732,17 +732,15 @@ static inline std::tuple<bool, // is valid
             if(cfg.tensor_b_thread_lengths[2] * cfg.tensor_b_cluster_lengths[2] > 1)
             {
 
-                if(c % gemm_n_per_block != 0 || gemm_m % gemm_m_per_block != 0)
+                if(c % gemm_n_per_block != 0)
                 {
                     continue;
                 }
             }
-            else
-            {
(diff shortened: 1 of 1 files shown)


shaojiewang:
> Reverts #982. Partial W/A for FP32 SSD issue (memory access fault in https://ontrack-internal.amd.com/browse/SWDEV-295434).
> 
> ⚠️ Merging this will cause #980 to appear again!

Yes. it will cause #980 again. So just close this solver?

shaojiewang (commented):
maybe close the solver? or remove the relevant ctest case?

junliume (approved):

atamazov:
> maybe close the solver? 

This may affect performance... If CQE will report the issue, AND full-blown fix is not ready, then we'll consider this possibility.

> or remove the relevant ctest case?

It should be already removed in this PR when reverting #982, but, well, this is not actually so, because the test case was already disabled as a workaround for issue #995. That is why you do not see changes in tests/CMakeLists.txt in this PR.

Please note that removing test cases is generally not a way to go, because camouflaging the issue may make more harm than the issue itself.

junliume:
> maybe close the solver? or remove the relevant ctest case?

@shaojiewang here is the proposed work flow: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1049

shaojiewang (approved):

Issue #979, opened 2021-06-10
[MLIR][gfx906][test_conv_igemm_mlir] Failed creating module


shurale-nkn:
MLIR build failed on test_conv_igemm_mlir on rocm 4.2.

Posibly because of gfx906:sramecc+:xnack-

ROCM 4.2
gfx906

http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/jenkins-ci-rocm-4.2/7/pipeline/129

```
11/69 Test #58: test_conv_igemm_mlir .............***Failed    8.41 sec
[  0%] Built target sqlite_memvfs
[  2%] Built target addkernels
[100%] Built target MIOpen
[100%] Built target test_conv2d
Scanning dependencies of target test_conv_igemm_mlir
MIOpen(HIP): Info [GetFindModeValueImpl] MIOPEN_FIND_MODE = NORMAL(1)
MIOpen(HIP): Info [get_device_name] Raw device name: gfx906:sramecc+:xnack-
MIOpen(HIP): Info [Handle] stream: 0x22f5bc0, device_id: 5
MIOpen(HIP): Info [BackwardDataGetWorkSpaceSize] 
MIOpen(HIP): Info [HipCompilerVersionImpl] 4.2.21155
MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, MIOpen version 2.12.0.8dd3b8aa9
MIOpen(HIP): Info [GetEnvFindOnlySolverImpl] 98
MIOpen(HIP): Info [ForwardGetWorkSpaceSize] 
MIOpen(HIP): Info [BackwardWeightsGetWorkSpaceSize] 
/var/jenkins/workspace/LLibs_MIOpen_jenkins-ci-rocm-4.2/build/bin/test_conv2d --float --cmode conv --pmode default --group-count 1 --disable-backward-data --disable-backward-weights --input 256, 1024, 14, 14 --weights 2048, 1024, 1, 1 --pads_strides_dilations 0 0 2 2 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW 
MIOpen(HIP): Info [ForwardGetWorkSpaceSize] 
MIOpen(HIP): Info [FindConvFwdAlgorithm] requestAlgoCount = 1, workspace = 0
MIOpen(HIP): Info [TryLoad] Find-db regenerating.
[...]

PR #1052, opened 2021-07-23
[MLIR] Added support for TargetId. [TESTS] Upgrade from ROCm 3.7 to 4.2 for MLIR testing.

jerryyin:
In the pull request, MIOpen will pass the IsaName to MLIR in the follow format: \<triple>:\<chip>:\<features>
  - Eg: amdgcn-amd-amdhsa:gfx908:sramecc+:xnack-

With this, the MLIR library will create kernel according to the client set information.

This should address https://github.com/ROCmSoftwarePlatform/MIOpen/issues/979.

Test performed:
 - MIOpen is able to functional correctly on ROCm 4.2 on gfx908 with features set


Dockerfile
```diff
@@ -13,8 +12,6 @@ RUN dpkg --add-architecture i386
 
 RUN if [ "$USE_TARGETID" = "ON" ] ; \
         then export ROCM_APT_VER=.apt_4.1.1;\
-    elif [ "$USE_MLIR" = "ON" ] ; \
```
atamazov, line 16: Maybe it's worth keeping capability of using different version of ROCm for MLIR builds. Are you sure that MLIR-enabled MIOpen will always use the same ROCm as non-MLIR one? If we upgrade our CI from 4.2 to 4.3, would MLIR work normally? If not, that would delay upgrading our CI until the MLIR issue is resolved. Of course we can simply switch off smoke MLIR testing by default in such a case... What do you think?

src/solver/mlir_common.cpp
```diff
@@ -60,6 +60,17 @@ const char* DTypeName(miopenDataType_t ty)
     MIOPEN_THROW(miopenStatusInternalError, "Value outside of datatype enum");
 }
 
+static std::string GetIsaName(const miopen::TargetProperties& target)
+{
+#if ROCM_FEATURE_TARGETID_OFF
+    const char* const ecc_suffix = (target.Sramecc() && *target.Sramecc()) ? ":sramecc+" : "";
+    return {"amdgcn-amd-amdhsa:" + target.Name() + ecc_suffix};
+#else
+    const LcOptionTargetStrings lots(target);
```
atamazov, line 69: Please add `#include <miopen/hip_build_utils.hpp> // for LcOptionTargetStrings`

src/solver/mlir_common.cpp
```diff
@@ -60,6 +60,17 @@ const char* DTypeName(miopenDataType_t ty)
     MIOPEN_THROW(miopenStatusInternalError, "Value outside of datatype enum");
 }
 
+static std::string GetIsaName(const miopen::TargetProperties& target)
+{
+#if ROCM_FEATURE_TARGETID_OFF
```
atamazov, line 65: Adding `#include <miopen/rocm_features.hpp>` is necessary.

atamazov (changes requested):

Dockerfile
jerryyin: Fair enough. I don't think I can give guarantee that MLIR implementation will always be forward and backward compatible. This PR itself is an example of how ROCm version bump can potentially break MLIR. So I agree with you that it might be easier to keep the branching logic here. I will make rocm version to be 4.2 to be consistent with the rest of the CI stages and leave it to be flexible in case future breakage comes in.

jerryyin: Done in e8de8d3

src/solver/mlir_common.cpp
jerryyin: Done in e8de8d3

src/solver/mlir_common.cpp
jerryyin: Done in e8de8d3

atamazov (approved):
LGTM!

atamazov:
Merge conflicts in dockerfile resolved. @jerryyin Please double check when you have time.

jerryyin:
@atamazov Merge result looks good, thank you!

PR #1059, opened 2021-07-27
Dynamic Composable Kernel Initial Integration

asroy:
Add new dynamic composable kernel under:
```src/composable_kernel/```, which is pulled in as a "git subtree" from composable kernel repo
https://github.com/ROCmSoftwarePlatform/composable_kernel

Move old static composable kernel to:
```src/kernels/static_composable_kernel/```

Add a new iGEMM solver for Forward NCHW FP32/FP16/Int8 (int8 is not supported by ConvolutionContext so the solver support for Int8 cannot be used yet):
```src/solver/conv_ck_igemm_fwd_v6r1_dlops_nchw.cpp::ConvCkIgemmFwdV6r1DlopsNchw``` , which is calls into composable kernel solver:
```src/composable_kernel/host/driver_online/include/conv_igemm_fwd_v6r1_dlops_nchw_kcyx_nkhw.hpp::ConvIgemmFwdV6r1DlopsNchwKcyxNkhw```


atamazov:
@asroy > Move old static composable kernel to: src/kernels/static_composable_kernel/

Why, if we can keep old static ones as is, and add new kernels to .../dynamic_composable_kernel?

atamazov:
@asroy 

> src/composable_kernel/host/online_compilation/addkernels/*

Does it differ from `addkernels` that we already have in the tree?

atamazov:
@asroy 

> Add a new iGEMM solver for Forward NCHW FP32/FP16/Int8 (int8 is not supported by ConvolutionContext so the solver support for Int8 cannot be used yet)

This should not be so. We already have Int8 convolution Solvers. What is wrong?

atamazov:
@asroy 

> Add new dynamic composable kernel under: src/composable_kernel/, which is pulled in as a "git subtree" from composable kernel repo

I am sorry but I do not remember the motivation for subtree... Where can I refresh my memory (mail thread or ticket in the repo...)?

atamazov:
@asroy Have you looked at DeepCode's analysis?

asroy:
> @asroy > Move old static composable kernel to: src/kernels/static_composable_kernel/
> 
> Why, if we can keep old static ones as is, and add new kernels to .../dynamic_composable_kernel?

I plan to remove the old static kernel source code eventually

asroy:
> @asroy
> 
> > src/composable_kernel/host/online_compilation/addkernels/*
> 
> Does it differ from `addkernels` that we already have in the tree?

It's used when building composable kernel alone (and outside MIOpen). It is not used in MIOpen.

The whole composable kernel is pulled in as subtree

asroy:
> @asroy
> 
> > Add new dynamic composable kernel under: src/composable_kernel/, which is pulled in as a "git subtree" from composable kernel repo
> 
> I am sorry but I do not remember the motivation for subtree... Where can I refresh my memory (mail thread or ticket in the repo...)?

The motivation is to be able to sync composable kernel repo and MIOpen automatically.

We had the meeting in 07/06 "MIOpen and composable kernel", between you, @JehandadKhan @pfultz2, me and @zjing14. Subtree approach was suggested there

asroy:
> @asroy Have you looked at DeepCode's analysis?

Yes. You mean unused code? Those code were there because the whole repo is pulled in as "subtree"

asroy:
> @asroy
> 
> > Add a new iGEMM solver for Forward NCHW FP32/FP16/Int8 (int8 is not supported by ConvolutionContext so the solver support for Int8 cannot be used yet)
> 
> This should not be so. We already have Int8 convolution Solvers. What is wrong?

There is not "IsInt8()" in ConvolutionContext. I tried to add it once, but compiler complain something like no "InputTensorDataType()". So I gave up

atamazov:
@asroy 
> There is not "IsInt8()" in ConvolutionContext. I tried to add it once, but compiler complain something like no "InputTensorDataType()". So I gave up

If this is a showstopper, then please ask @DrizztDoUrden or me for assistance or open a ticket.

atamazov:
> > @asroy
> > > src/composable_kernel/host/online_compilation/addkernels/*
> > 
> > Does it differ from `addkernels` that we already have in the tree?
> 
> It's used when building composable kernel alone (and outside MIOpen). It is not used in MIOpen.

[Notice]
This does not answer my question. Let me assume that this addkernels is a _copy_ of the one that resides in MIOpen.

If the efforts to make the composable kernels _independent_ has resulted in the ___addition of another implicit dependency___ (a copy of addkernels), then I'm afraid that something is wrong.

But well, I signed off from reviewing this, so let's actual reviewers decide.

asroy:
The commit history in this PR is messed up, because I did subtree pulls from multiple remote branches from different composable_kernel repo forks.
 
I realized it's important to have a single remote composble_kernel branch to pull from, in order to keep the commit historyu correct.

I need to close this PR and open a new PR.

@atamazov I will refer to all your comments here in the new PR

atamazov:
@asroy
> The motivation is to be able to sync composable kernel repo and MIOpen automatically.

[Notice] Thanks for the explanation. Maybe I do not understand the meaning of "automatically" in this context. But would it allow something like a one-liner PR in MIOpen to completely change the implementation of the Solvers that employ composable kernels? If yes, then this looks dangerous to me.

atamazov:
@asroy 
> I will refer to all your comments here in the new PR

I see. Thank you.

asroy:
> > > @asroy
> > > > src/composable_kernel/host/online_compilation/addkernels/*
> > > 
> > > 
> > > Does it differ from `addkernels` that we already have in the tree?
> > 
> > 
> > It's used when building composable kernel alone (and outside MIOpen). It is not used in MIOpen.
> 
> [Notice]
> This does not answer my question. Let me assume that this addkernels is a _copy_ of the one that resides in MIOpen.
> 
> If the efforts to make the composable kernels _independent_ has resulted in the _**addition of another implicit dependency**_ (a copy of addkernels), then I'm afraid that something is wrong.
> 
> But well, I signed off from reviewing this, so let's actual reviewers decide.

I don't have a good way to avoid this issue (some host codes are not used in MIOpen) if using subtree feature. I actually brought up this issue during our 07/06 meeting as a side effect of using subtree (or submodules).

The only way I can think of it's to make composable_kernel as a separately compiled library and provide a API that return a string of kernel source code to MIOpen. This was an alternative brought up by @JehandadKhan. But the in the meeting, it was decided subtree is the way to go.

I'm fine with either approach

asroy:
> The commit history in this PR is messed up, because I did subtree pulls from different remote branches and from different composable_kernel repo forks.
> 
> I realized it's important to have a single composble_kernel repo, and a single composble_kernel branch to pull from, in order to keep the commit historyu correct.
> 
> I need to close this PR and open a new PR.
> 
> @atamazov I will refer to all your comments here in the new PR

Closing this PR.

Open new one #1071

PR #1037, opened 2021-07-17
[BN][Navi21] Get rid of "#undef __AMDGCN__"

muralinr:
Addressed compile time requests from this issue.

https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1036

PR #1065, opened 2021-07-29
[Navi21][HOTFIX] Fix kernel build issues introduced in PR #1037

atamazov:
Also fixes potential bug because `MIOPEN_USE_AMDGCN` should not be defined for gfx1030.

### DIFF
--- a/src/kernels/batchnorm_functions.h
+++ b/src/kernels/batchnorm_functions.h
@@ -129,9 +129,16 @@
 #define MIO_BN_MAXN 65
 #endif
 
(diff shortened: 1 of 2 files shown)


atamazov:
The reason of the issue is that #1037 wasn't manually tested on Navi21 with BUILD_DEV=On.

muralinr (approved):
Changes are fine.

junliume (approved):

shurale-nkn (approved):

PR #1066, opened 2021-07-29
[gfx908][TESTS] Adding a number of tests

shurale-nkn:
1) Most of test disabled on gfx908 are now enabled.
2) gfx908 is enabled by default for custom tests.


atamazov (approved):
Very good, thank you!

atamazov (commented):
@shurale-nkn Tests are now enabled by default for gfx908. Documentation at https://github.com/ROCmSoftwarePlatform/MIOpen/blob/c701af3ba83f8fb99d7efab30ee0b22de57f2f42/test/CMakeLists.txt#L408 must be updated.

PR #1081, opened 2021-08-04
[NOGPU] Fix #1080 enable MIOpen builds on GPU-less

### LINKED ISSUE #1080: [NoGPU] Build MIOpen fails when GPU device does not exist
After enablement of #1024 MIOpen cannot build successfully on GPU-less platforms, such as WSL environment

[Task]
Differentiate "ROCMINFO FAILED, GPU TYPE UNKNOWN". and "ROCk module is NOT loaded, possibly no GPU devices".
Community can play with MIOpen build with WSL
MIOpen can be built on GPU-less platforms or servers

junliume:
After enablement of #1024 MIOpen cannot build successfully on GPU-less platforms, such as WSL environment

This PR fixes #1080

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -72,6 +72,7 @@ endif()
 # For HIP_NOGPU backend, GPU detection is not required and should be disabled.
 # Also we do not detect GPU when target GPU for testing is specified explicitly.
 set(MIOPEN_TEST_GPU_DETECTION_FAILED FALSE)
+set(MIOPEN_NO_GPU FALSE)
 if(NOT (MIOPEN_TEST_VEGA OR MIOPEN_TEST_GFX908 OR MIOPEN_TEST_GFX1030 OR MIOPEN_TEST_HIP_NOGPU))
     find_program(ROCMINFO
         NAMES rocminfo
@@ -88,7 +89,10 @@ if(NOT (MIOPEN_TEST_VEGA OR MIOPEN_TEST_GFX908 OR MIOPEN_TEST_GFX1030 OR MIOPEN_
             OUTPUT_VARIABLE ROCMINFO_OUTPUT
             RESULT_VARIABLE ROCMINFO_EXIT_STATUS
         )
-        if(NOT ROCMINFO_EXIT_STATUS EQUAL 0)
+        if(ROCMINFO_OUTPUT MATCHES "no GPU devices")
+            message(WARNING "ROCk module is NOT loaded, possibly no GPU devices")
+            set(MIOPEN_NO_GPU ON)         
+        elseif (NOT ROCMINFO_EXIT_STATUS EQUAL 0)
             message(WARNING "ROCMINFO FAILED, GPU TYPE UNKNOWN. Manually set respective MIOPEN_TEST_GFX* CMake variable to specify target GPU for testing.")
             set(MIOPEN_TEST_GPU_DETECTION_FAILED TRUE)
         elseif(ROCMINFO_OUTPUT MATCHES "gfx1030")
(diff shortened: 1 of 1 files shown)


JehandadKhan (approved):

atamazov:
> After enablement of #1024 MIOpen cannot build successfully on GPU-less platforms

[Clarification] Build does not fail in this case. Testing, i.e. `make check` fails. That was done intentionally.

atamazov:
**_This PR makes CI testing unreliable._** If/when a node has problems and clinfo fails, then testing must fail on Jenkins. With this PR, it will PASS.

junliume:
> > After enablement of #1024 MIOpen cannot build successfully on GPU-less platforms
> 
> [Clarification] Build does not fail in this case. Testing, i.e. `make check` fails. That was done intentionally.

@atamazov  please refer to [SWDEV-298600](https://ontrack-internal.amd.com/browse/SWDEV-298600?focusedCommentId=7719696&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-7719696)
Our CQE is testing in a two stage manner: one stage is to build docker on a GPU-less system, and the second stage is to pull the docker and run test. PR #1024 will stop Stage 1 for them.
The workaround is to build regardless if GPU presents, and when testing, run rim -rf test first.

atamazov:
> Our CQE is testing in a two stage manner: one stage is to build docker on a GPU-less system

I guess that they should NOT use `make check` to build MIOpen. `make tests` should do the job.

junliume:
> > Our CQE is testing in a two stage manner: one stage is to build docker on a GPU-less system
> 
> I guess that they should NOT use `make check` to build MIOpen. `make tests` should do the job.

And another user case ... might be rare but maybe not so rare for community users, I wish to be able to build on Windows WSL.
"ROCk module is NOT loaded, possibly no GPU devices" is the information I got on Windows WSL :)

junliume:
> **_This PR makes CI testing unreliable._** If/when a node has problems and clinfo fails, then testing must fail on Jenkins. With this PR, it will PASS.

This is indeed problematic ... need to discuss how to deal with it. Let us align on it separately.

atamazov:
> "ROCk module is NOT loaded, possibly no GPU devices" is the information I got on Windows WSL :)

This should be just a warning that should not prevent `make -j` to build the library. Also I think we must provide explicit support for Windows WSL, _as a feature_. Currently we do not support this platform (AFAIK) so all the defects related to it are invalid, right?

junliume:
> > "ROCk module is NOT loaded, possibly no GPU devices" is the information I got on Windows WSL :)
> 
> This should be just a warning that should not prevent `make -j` to build the library. Also I think we must provide explicit support for Windows WSL, _as a feature_. Currently we do not support this platform (AFAIK) so all the defects related to it are invalid, right?

Yes, you are right, we need to be explicit in support matrix (what is supported and how long it is going to be). I am all in with you to move toward that goal.
While hopefully we can still leave some leverage for community testers and curiousers. I recall (need double check) it will fail abruptly with 
> add_custom_target(gpu_detection_check COMMAND echo "*** FATAL: GPU DETECTION FAILED DURING CMAKE PHASE, CHECK CMAKE WARNINGS ***" COMMAND exit 1)

I will focus on #1102 for now.

test/CMakeLists.txt
```diff
@@ -161,18 +165,24 @@ else()
...
                 test_immed_conv3d test_gru test_rnn_vanilla test_lstm test_gru_extra
-                test_rnn_extra test_lstm_extra )
+                test_rnn_extra test_lstm_extra)
 	endif()
 endif()
 
+if (MIOPEN_NO_GPU)
+    set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params test_lstm test_lstm_dropout 
+            test_test_errors test_type_name test_tensor_test test_sqlite_perfdb test_sequences
+            test_pooling3d test_perfdb)
```
atamazov, line 183: @junliume I think that `test_pooling3d` is not necessary here, right? I am going to remove it.

PR #1082, opened 2021-08-05
[Test][CI][gfx90a] Basic support for gfx90a: Jenkinsfile, dockerfile, unit tests

shaojiewang:

Issue #1094, opened 2021-08-18
[Test][CI][gfx90a] MIOpenTensile Stage Support on gfx90a


junliume:
[Target]
Confirm whether MIOpenTensile Stage needs to be supported on gfx90a

[Cause]
MIOpenTensile Stage currently disabled for gfx90a:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/jenkins_and_docker_file_for_90a/47/pipeline/128 

> [2021-08-17T19:39:36.704Z] ******************************************************************************* 
[2021-08-17T19:39:36.704Z] *----------------------------------- ERROR -----------------------------------* 
[2021-08-17T19:39:36.704Z] * The variable 'CMAKE_C_FLAGS' should only be set by the cmake toolchain, 
[2021-08-17T19:39:36.704Z] * either by calling 'cmake -DCMAKE_C_FLAGS=" -fopenmp=libomp"' or 
[2021-08-17T19:39:36.704Z] * set in a toolchain file and added with 
[2021-08-17T19:39:36.704Z] * 'cmake -DCMAKE_TOOLCHAIN_FILE=<toolchain-file>'. 
[2021-08-17T19:39:36.704Z] *-----------------------------------------------------------------------------* 
[2021-08-17T19:39:36.704Z] ******************************************************************************* 
[2021-08-17T19:39:36.704Z] 
[2021-08-17T19:39:36.704Z] CMake Warning at /opt/rocm/share/rocm/cmake/ROCMChecks.cmake:41 (message): 
[2021-08-17T19:39:36.704Z] The toolchain variable 'CMAKE_C_FLAGS' is modified in the CMakeLists.txt. 
[2021-08-17T19:39:36.704Z] Call Stack (most recent call first): 
[2021-08-17T19:39:36.704Z] /usr/local/cget/build/tmp-7a5e3af0aa2441dca6278c73a73a65bc/build/virtualenv/lib/python3.6/site-packages/Tensile/Source/CMakeLists.txt:9999 (rocm_check_toolchain_var) 
[2021-08-17T19:39:36.704Z] /usr/local/cget/build/tmp-7a5e3af0aa2441dca6278c73a73a65bc/build/virtualenv/lib/python3.6/site

junliume:
Some references:
https://gitlab.kitware.com/cmake/cmake/-/issues/21787
and lines from:
[...]

Issue #285, opened 2020-06-12
[Test][CI][gfx1030][test_soft_max] Investigate & fix soft_max failures and instabilities


atamazov:
PR #240 has a workaround, commit b4d6805 [test_soft_max] Workaround for cases that often fail on Jenkins and ___always___ fail with ROCm3.3/Radeon VII.

Nevertheless, `test_soft_max` continues to often fail on jenkins, for example, see:
- http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/develop/17/pipeline/
- http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/sqlite-binary-cache-reenable/3/pipeline/952

aserio:
@daniellowell what is the status of this ticket?

aserio:
@daniellowell following up with my request.

atamazov:
This still happens from time to time.

junliume:
This still happens from time to time, more often on gfx1030 servers:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/composable_kernel_init_integration_v3/35/pipeline 
CC: @asroy @aserio @atamazov

PR #1098, opened 2021-08-19
Copy of "[Test][CI][gfx90a] Basic support for gfx90a: Jenkinsfile, dockerfile, unit tests (#1082)"

junliume:
copy of # 1082


asroy (approved):

Jenkinsfile
```diff
@@ -273,6 +275,7 @@ pipeline {
         Full_test       = " -DMIOPEN_TEST_ALL=On"
         Tensile_build_env = "MIOPEN_DEBUG_HIP_KERNELS=0 "
         Tensile_setup = " -DMIOPEN_TEST_MIOTENSILE=ON -DMIOPEN_USE_MIOPENTENSILE=ON -DMIOPEN_USE_ROCBLAS=OFF"
+        gfx90a_flags = "-DMIOPEN_TEST_GFX90a=ON"
```
shurale-nkn, line 278: Remove this flag. Cmake should use auto detection

Jenkinsfile
```diff
@@ -548,6 +578,20 @@ pipeline {
                         buildHipClangJobAndReboot( setup_flags: Tensile_setup + extra_log_env + Bf16_flags, build_env: Tensile_build_env + extra_log_env, test_flags: ' --verbose ', gpu_arch: "gfx908:xnack-", miotensile_version: Tensile_version, target_id: "ON")
                     }
                 }
+                // stage('Fp32 Hip Tensile-Latest All gfx90a') {
```
shurale-nkn, line 581: What about these stages?

Jenkinsfile
```diff
@@ -600,6 +656,10 @@ pipeline {
                 }
                 stage('Fp32 OpenCL All gfx1030') {
                     agent{ label rocmnode("navi21") }
+                    options {
+                        timeout(time: 90, unit: 'MINUTES')
```
shurale-nkn, line 660: why is this required here?

Jenkinsfile
```diff
@@ -610,6 +670,12 @@ pipeline {
                         buildHipClangJobAndReboot(setup_flags: Full_test + Fp16_flags, build_env: WORKAROUND_iGemm_936, build_install: "true", gpu_arch: "gfx908")
                     }
                 }
+                stage('Fp16 Hip All Install gfx90a') {
+                    agent{ label rocmnode("gfx90a") }
+                    steps{
+                        buildHipClangJobAndReboot(setup_flags: Full_test + Fp16_flags + gfx90a_flags, build_env: WORKAROUND_iGemm_936, build_install: "true", gpu_arch: "gfx90a:xnack-")
```
shurale-nkn, line 676: remove gfx90a_flags

Jenkinsfile
```diff
@@ -373,6 +382,16 @@ pipeline {
...
                     }
                 }
+                stage('Fp32 Hip Debug gfx90a /opt/rocm') {
+                    agent{ label rocmnode("gfx90a") }
+                    environment{
+                        gpu_arch = "gfx90a:xnack-"
+                        prefixpath = "/opt/rocm"
+                    }
+                    steps{
+                        buildHipClangJobAndReboot(setup_flags: gfx90a_flags, prefixpath: prefixpath, build_type: 'debug', config_targets: Smoke_targets, gpu_arch: gpu_arch)
```
shurale-nkn, line 392: remove gfx90a_flags

Jenkinsfile
```diff
@@ -667,6 +733,30 @@ pipeline {
                         buildHipClangJobAndReboot( setup_flags: Tensile_setup + Full_test + extra_log_env + Int8_flags, build_env: Tensile_build_env + extra_log_env, test_flags: ' --verbose ', gpu_arch: "gfx908:xnack-", miotensile_version: Tensile_version, target_id: "ON")
                     }
                 }
+                // stage('Fp32 Hip Tensile All gfx90a') {
+                //     agent{ label rocmnode("gfx90a") }
```
shurale-nkn, line 737: What about these stages? *same

test/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ option( MIOPEN_TEST_HALF "Test in half mode" OFF )
 option( MIOPEN_TEST_INT8 "Test in int8 mode" OFF )
 option( MIOPEN_TEST_BFLOAT16 "Test in bfloat16 mode" OFF )
 option( MIOPEN_TEST_GFX908 "Test on MI100 (gfx908)" OFF )
+option( MIOPEN_TEST_GFX90a "Test on gfx90a" ${MIOPEN_TEST_GFX90a} )
```
shurale-nkn, line 37: auto detection

shurale-nkn (changes requested):

Jenkinsfile
junliume: Autodetection is only possible when #1099 is fully checked in.
Currently all CI nodes are using ROCm4.2, where rocminfo cannot recognize gfx90a properly thus causing GPU Detection to fail.
The "workaround" in this PR is to identify GPU type from arch ID, and pull ROCm4.3 only if gfx90a is set.
Sort of chicken and egg paradox: we need ROCm4.3 to detect gfx90a properly and then pull ROCm4.3 to build the docker :)
#1099 is going to fix the problem. However, since upgrading the ROCm version for all CI might be a big change, I hope to check in this one to unlock a lot of gfx90a related PRs. 
Thanks!

Jenkinsfile
junliume: They are tracked by this issue:
https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1094

Jenkinsfile
junliume: Fp32 OpenCL All gfx1030 recently has some stability issues: ref: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/285

Since it is OCL only and likely not a bug caused by MIOpen changes:
1.  Set a retry(2) for this stage ONLY. because if it failed before, the only way is to restart the whole set of parallel stages, which will take hours and many resources.
2. timeout is for an observation that when this stage is "unstable", usually it will *stuck* after the failure for hours (last observation is that it has ctest failure but stuck for 3 hours). We want it to fail fast and with minimal manual intervention.

Jenkinsfile
junliume: Same as here: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1098#discussion_r692316081

Jenkinsfile
junliume: Same as here: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1098#discussion_r692316081

test/CMakeLists.txt
junliume: Same as here :): https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1098#discussion_r692316081

Jenkinsfile
junliume: Same as here :) : https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1098#discussion_r692316833

Jenkinsfile
shurale-nkn: but wait. This flag used only in 4 stages, but gfx90a was tested in 7 stages. So, the 3 stages work with autodetection and work fine.

shurale-nkn: For this PR ROCm4.3 required only on gfx90a, docker rocm selection based on value in 'gpu_arch'. 
So no need to use (gfx90a_flags = "-DMIOPEN_TEST_GFX90a=ON") 
As I can see from the logs, the GPU was identified correctly.

junliume: Yes, you are right :) gpu_arch and gfx90a_flags are redundant checks.
Let me make the changes and remove this flag. Thanks! @shurale-nkn

test/CMakeLists.txt
shurale-nkn: same

shurale-nkn (approved):
LGTM

JehandadKhan (approved):

shurale-nkn (approved):

PR #1103, opened 2021-08-20
IsTunable function for Solver Id

cderb:
Add method for reporting tunable solvers.
Includes Fin update to pass status to Tuna


src/solver.cpp
```diff
@@ -113,6 +113,7 @@ struct IdRegistryEntry
     Primitive primitive            = Primitive::Convolution;
     miopenConvAlgorithm_t convAlgo = miopenConvolutionAlgoDirect;
     AnySolver solver               = {};
+    bool is_tunable                = false;
```
atamazov, line 116: I think this should be a Solver's attribute, not SolverID's one. We can easily implement a generic function that takes a Solver instance and returns if it is tunable or not, and a wrapper in `AnySolver`.

@DrizztDoUrden Am I correct?

cderb: I think it makes sense as an AnySolver attribute. @JehandadKhan suggested doing it this way would be a smaller diff, and wouldn't require editing all of the solvers.

DrizztDoUrden: I think it should be a solver attribute. Basically, it describes solver implementation details.

JehandadKhan: @cderb Since @DrizztDoUrden and @atamazov both agree that this should be a solver attribute, can you please update your PR?

cderb: I'm testing a solution.

fin/src/include/conv_fin.hpp
```diff
@@ -774,10 +778,19 @@ template <typename Tgpu, typename Tref>
...
         miopen::solver::GetSolversByPrimitive(miopen::solver::Primitive::Convolution))
-        solvers.push_back(std::make_pair(id.Value(), id.ToString()));
+    {
+        std::unordered_map<std::string, std::string> solver;
+        solver["id"]      = std::to_string(id.Value());
+        solver["name"]    = id.ToString();
+        solver["tunable"] = "0";
+        if(id.GetSolver().IsTunable())
+            solver["tunable"] = "1";
+        solvers.push_back(solver);
```
JehandadKhan, line 791: Can we add an `IsDynamic` attribute too ?

cderb: done

JehandadKhan (dismissed):

junliume:
Ping @JehandadKhan  @atamazov @DrizztDoUrden for review.
CI has failed due to HW issue "Cannot contact prj47-rack-19", but passed with the same commit at http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/cerb_solver_istunable/12/pipeline

PR #1106, opened 2021-08-21
[Composable kernel] Improve PR1097 v2

asroy:
* use ```cast_pointer_to_generic_address_space()``` in v6r1 kernel wrapper
* update ```make_dynamic_buffer```
* add buffer load/store for fp64


asroy:
It's already in #1108

Issue #1093, opened 2021-08-17
[Test][CI][gfx90a] Temporarily disabled unit tests


junliume:
[Target]
re-enable the list of tests which have been disabled temporarily on gfx90a platform:

- [x] test_conv2d 
- [x] test_find_db 
- [x] test_main 
- [x] test_immed_conv2d

[Cause]

1.  The first three is caused by *Assembly error*
2.  test_immed_conv2d is caused by:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/jenkins_and_docker_file_for_90a/52/pipeline/ 
> [2021-08-17T22:53:54.837Z] MIOpen(HIP): Warning [ParseAndLoadDb] File is unreadable: /var/jenkins/workspace/_jenkins_and_docker_file_for_90a/src/kernels/gfx90a_xnack68.HIP.fdb.txt
> [2021-08-17T22:53:54.837Z] <instantiation>:65:5: error: .amdhsa_reserve_xnack_mask does not match target id
> [2021-08-17T22:53:54.837Z]     .amdhsa_reserve_xnack_mask               __sgpr_reserve_xnack_default

shaojiewang:
As we need to open test_conv2d on gfx90a, this error is observed: 
For gfx90a arch, some asm kernels use .amdhsa_reserve_xnack_mask, it seems like to cause a compile error: ```.amdhsa_reserve_xnack_mask does not match target id```
CI is blocked: http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/asm_igemm_wrw_nhwc_gfx90a/9/pipeline/213/

carlushuang:
`EPILOG_KERNEL_DESCRIPTOR miopenSp3AsmConv_v21_1_2_gfx9_fp32_stride1`
from the log of the CI, winograd kernel happens to be compiled, where the compiler complain about an unrecognized syntax. Maybe winograd need to be disabled on gfx90a test environment
cc @DrizztDoUrden

JehandadKhan:
@Kirpich30000 can you please take a look at the Winograd kernel and suggest future steps for enablement on gfx90a.

atamazov:
As far as I see, the root reason of all the issues is assembly errors. In order to fix these errors we need access to gfx90a or logs of faulty cases taken with level 6.

Slimakanzer:
[...]

PR #1110, opened 2021-08-24
[CI][WINO][gfx90a]Fix gfx90a test conv2d error in wino solver

shaojiewang:
add gfx90a to blacklist of conv_winoRxS_f2x3

### DIFF
--- a/src/solver/conv_winoRxS_f2x3.cpp
+++ b/src/solver/conv_winoRxS_f2x3.cpp
@@ -452,7 +452,7 @@ static bool IsApplicableBase(const ConvolutionContext& params)
         return false;
 
     const auto name = params.GetStream().GetDeviceName();
-    if(!(StartsWith(name, "gfx9") || StartsWith(name, "gfx10")))
+    if(!(StartsWith(name, "gfx9") || StartsWith(name, "gfx10")) || name == "gfx90a")
         return false;
     if(params.IsFp16() &&
        !(StartsWith(name, "gfx906") || StartsWith(name, "gfx908") || StartsWith(name, "gfx1011") ||
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -199,7 +199,7 @@ endif()
 
 if (MIOPEN_TEST_GFX90a)
     if(WORKAROUND_ISSUE_1093)
-        list(APPEND SKIP_TESTS test_conv2d test_find_db test_main test_immed_conv2d)
+        list(APPEND SKIP_TESTS test_find_db test_main test_immed_conv2d)
     endif()
     if(WORKAROUND_ISSUE_1095)
         list(APPEND SKIP_TESTS test_dropout)


test/CMakeLists.txt
```diff
@@ -199,7 +199,7 @@ endif()
 
 if (MIOPEN_TEST_GFX90a)
     if(WORKAROUND_ISSUE_1093)
-        list(APPEND SKIP_TESTS test_find_db test_main test_immed_conv2d)
+        list(APPEND SKIP_TESTS test_conv2d test_find_db test_main test_immed_conv2d)
```
junliume: Hi @shaojiewang , by reviewing "all changed", this test_conv2d is still not removed from the SKIP_TESTS. Is that an error? Thanks!

junliume (commented):
Hi @shaojiewang  not sure where my previous comment went on GitHub. For this PR, would test/CMakeLists.txt need to be changed for Line 202 to remove test_conv2d from the SKIP_TESTS list? Thanks!

src/solver/conv_winoRxS_f2x3.cpp
```diff
@@ -452,7 +452,7 @@ static bool IsApplicableBase(const ConvolutionContext& params)
         return false;
 
     const auto name = params.GetStream().GetDeviceName();
-    if(!(StartsWith(name, "gfx9") || StartsWith(name, "gfx10")))
+    if(!(StartsWith(name, "gfx9") || StartsWith(name, "gfx10")) || name == "gfx90a")
```
atamazov, line 455: This should be legalized as `WORKAROUND_ISSUE_1093`. Please see #561.

atamazov (changes requested):

test/CMakeLists.txt
atamazov: Yes. I guess that at least test_conv2d should be removed from W/A for #1093 in `test/CMakeLists.txt`.

Even more, I suspect that when all the gfx90a assembly errors fixed (or avoided via W/A), the whole `WORKAROUND_ISSUE_1093` in `test/CMakeLists.txt` will not be necessary anymore. This can be easily checked by passing `-DWORKAROUND_ISSUE_1093=Off` to CMake.

@shaojiewang Please make neccesary changes of `test/CMakeLists.txt` right in this PR, thanks.

src/solver/conv_winoRxS_f2x3.cpp
atamazov: [Note] In fact, this PR moves W/A from tests to the library.

shaojiewang: @atamazov While I make gfx90a test disable as default in ctest, it still raised error from test_conv_2d compilation. So I wrote this line in library.

test/CMakeLists.txt
shaojiewang: changed

junliume:
@atamazov should we create a post merge issue with this PR, so that hopefully we can resolve those potential performance issues related to [SWDEV-302268](https://ontrack-internal.amd.com/browse/SWDEV-302268)? Yes, it is (almost) pinpointed to this commit.

junliume:
> @atamazov should we create a post merge issue with this PR, so that hopefully we can resolve those potential performance issues related to [SWDEV-302268](https://ontrack-internal.amd.com/browse/SWDEV-302268)? Yes, it is (almost) pinpointed to this commit.

@atamazov  is the performance anticipated to be recovered by #1142?

atamazov:
@junliume 
> @atamazov is the performance anticipated to be recovered by #1142?

Yes. @Slimakanzer is going to test if it is really so (thank you, Gleb!)

PR #1125, opened 2021-08-27
[CI] Upgrade to ROCm 4.3.1 [WORKAROUND] Disable dynamic reduction by default for HIP backend && ROCm 4.3

asroy:
Disable dynamic reduction by default, due to issue #1123 

----
@atamazov: Also resolves the following review comments:
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1108#discussion_r697392056
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1108#discussion_r697400230
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1108#discussion_r697402712
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1108#discussion_r697412276

PR #1127, opened 2021-08-30
[WORKAROUND] Fix W/A in PR #1125 that disables dynamic reduction by default for HIP backend && ROCm 4.3

junliume:
Partial rework of #1125

Based on comments here: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1125/files#r698730133


atamazov (approved):
Thanks!

Issue #1128, opened 2021-08-31
Reduce dropout test time


ce1adon:
- [x] step 1. Reduce number of test case  #1129 
- [ ] step 2. separate rng and dropout test procedure

atamazov:
#1129 is qualified as a workaround yet, so the original defect shouldn't be closed.

junliume:
> #1129 is qualified as a workaround yet, so the original defect shouldn't be closed.

After original discussion with @ce1adon, this PR (reduce of number of tests) is more of a permanent approach to fix the issue.  
@ce1adon : what do you think?

atamazov:
Let's discuss this topic [there](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1129#issuecomment-918421792).

atamazov:
Let's close this when step 2 is done.

atamazov:
@junliume This should be reopened. The only config which is tested right now (even with `--all`!) is 
```
--input-dim 16 4 8 1 4 --dropout 0.5 --seed 0 --use-mask 0 --rng-mode 0 
```
which opens path for bugs. In fact, dropout is only smoke tested.

Recommended labels are with https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_blocker and https://github.com/ROCmSoftwarePlatform/MIOpen/labels/workaround

I have no idea who should be assigned to this (but please feel free to assign me)

/cc @ce1adon

atamazov:
[Notice] Separating rng mode seems more or less trivial.

Issue #17, opened 2017-10-07
4gb of memory not enough to run AlexNet?


masahi:
Hi,

I am trying to run patflick's [miopen-benchmark](https://github.com/patflick/miopen-benchmark) on my R9 Nano.
When I run his AlexNet benchmark, I get 'hipErrorOutOfMemory' from MIOpen, even though I believe 4gb of RAM should be enough to run AlexNet. 
Strangely, even if I change the batch size from 128 to 1, I still get the same 'hipErrorOutOfMemory' error (but from different layer).

Below is the output when the batch size is 128. I am using a debug build for MIOpen. 
I really appreciate if you could help me figure out what is going on.

$ ./alexnet 
[INFO]  Number of HIP devices found: 1
[INFO]  Device 0: Fiji [Radeon R9 FURY / NANO Series]
[INFO]  	Arch:	803
[INFO]  	GMem:	4096 MiB
[INFO]  	warps:	64
[INFO]  	CUs:	64
[INFO]  	MaxClk:	1000000
[INFO]  	MemClk:	500000
[INFO]  	drm:	/sys/class/drm/card0
[INFO]  	hwmon:	/sys/class/drm/card0/device/hwmon/hwmon1
[DEBUG] Allocating Float Tensor (64,3,11,11), total size: 90 kB
[DEBUG] Allocating Float Tensor (64,3,11,11), total size: 90 kB
[DEBUG] Allocating Float Tensor (128,64,55,55), total size: 96800 kB
[DEBUG] Allocating Float Tensor (128,64,55,55), total size: 96800 kB
[DEBUG] Allocating Float Tensor (128,64,27,27), total size: 23328 kB
[DEBUG] Allocating Float Tensor (192,64,5,5), total size: 1200 kB
[DEBUG] Allocating Float Tensor (192,64,5,5), total size: 1200 kB
[DEBUG] Allocating Float Tensor (128,192,27,27), total size: 69984 kB
[DEBUG] Allocating Float Tensor (128,192,27,27), total size: 69984 kB
[DEBUG] Allocating Float Tensor (128,192,13,13), total size: 16224 kB
[DEBUG] Allocating Float Tensor (384,192,3,3), total size: 2592 kB
[DEBUG] Allocating Float Tensor (384,192,3,3), total size: 2592 kB
[DEBUG] Allocating Float Tensor (128,384,13,13), total size: 32448 kB
[...]

PR #1129, opened 2021-08-31
Reduce number of dropout test case

ce1adon:
To save CI resource
Issue #1128

### DIFF
--- a/test/dropout.cpp
+++ b/test/dropout.cpp
@@ -39,6 +39,8 @@
 #define WORKAROUND_MLOPEN_ISSUE_2335 \
     ((HIP_PACKAGE_VERSION_FLAT < 3007000000ULL) && MIOPEN_BACKEND_OPENCL)
 
+static int num_test_case = 0;
+
 template <class T>
 struct verify_forward_dropout
 {
@@ -269,6 +271,12 @@ struct dropout_driver : test_driver
         return;
 #endif
 
+        // Workaround for issue #1128
+        if(mask || num_test_case > 0)
+        {
+            return;
+        }
+
         miopen::DropoutDescriptor DropoutDesc;
         unsigned long max_value  = miopen_type<T>{} == miopenHalf ? 5 : 17;
         auto&& handle            = get_handle();
@@ -348,6 +356,8 @@ struct dropout_driver : test_driver
             verify(
                 verify_backward_dropout<T>{DropoutDesc, din, dout, reserveSpace, 0, 0, 0, false});
         }
+
+        num_test_case++;
     }
 };
 
--- a/test/dropout.cpp
+++ b/test/dropout.cpp
@@ -33,7 +33,6 @@
 #include "random.hpp"
 
 #define DROPOUT_DEBUG_CTEST 0
-#define DROPOUT_LARGE_CTEST 0
 // Workaround for issue #1128
 #define DROPOUT_SINGLE_CTEST 1
 
@@ -247,6 +246,7 @@ struct dropout_driver : test_driver
         add(mask, "use-mask", generate_data({false}));
         add(rng_mode_cmd, "rng-mode", generate_data({0}));
 #else
+#define DROPOUT_LARGE_CTEST 0
 #if DROPOUT_LARGE_CTEST
         input_dims.insert(input_dims.end(), get_inputs_set.begin(), get_inputs_set.end());
         input_dims.insert(input_dims.end(),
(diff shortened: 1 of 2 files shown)


test/dropout.cpp
```diff
@@ -39,6 +39,8 @@
 #define WORKAROUND_MLOPEN_ISSUE_2335 \
     ((HIP_PACKAGE_VERSION_FLAT < 3007000000ULL) && MIOPEN_BACKEND_OPENCL)
 
+static int num_test_case = 0;
```
junliume, line 42: @ce1adon Global non-const variables are evil :)

ce1adon: Hmm I can't figure out another way to count the number here. Any suggestion?

junliume: Hi @atamazov  could you chip in to help please? Actually there are two options:
-  Run this dropout test exactly once. I would recommend moving ```static int num_test_case = 0;``` into scope of ```struct dropout_driver : test_driver```, I think we'll have a single instance of static variable among all translation units.
-  I suspect the following random vector filling has been called many times
   ```
   srand(0);
        for(size_t i = 0; i < in.desc.GetElementSize(); i++)
             reserveSpace[i] =
                 static_cast<unsigned char>(float(GET_RAND()) / float(RAND_MAX) > dropout_rate);
   ```
   I have not done the profiling myself, I would guess would this be a perf bottleneck **especially with multi-threading**? Could we fill it once only, and only shuffle in other instances? Or better yet, use other random generator?

ce1adon: Solved with macro

junliume:
@ce1adon  Thank you!

Fp16 Hip Install All Vega20
> [2021-09-01T18:49:30.258Z]  1/81 Test #17: test_dropout ...............................   Passed    5.03 sec

Fp32 Hip All Vega20
> [2021-09-01T18:45:12.918Z]  1/82 Test #17: test_dropout ...............................   Passed    4.86 sec

CC: @atamazov  @JehandadKhan

junliume (approved):

test/dropout.cpp
```diff
@@ -33,7 +33,8 @@
 #include "random.hpp"
 
 #define DROPOUT_DEBUG_CTEST 0
-#define DROPOUT_LARGE_CTEST 0
+// Workaround for issue #1128
+#define DROPOUT_SINGLE_CTEST 1
```
atamazov, line 37: [Recommendation] It should be like this:
```C++
#define WORKADOUND_ISSUE_1128 1
...
#if WORKADOUND_ISSUE_1128
#define DROPOUT_SINGLE_CTEST 1
#endif
...
```

atamazov:
@ce1adon @junliume Does it affect the quality of testing? In other words, it this really a workaround? If yes, then what it the proposed full-blown fix?

junliume:
> @ce1adon @junliume Does it affect the quality of testing? In other words, it this really a workaround? If yes, then what it the proposed full-blown fix?

The rational is that bottleneck is filling input_dims numerous times with random numbers instead of actually test dropout functionality. Aside from other drop out tests (e.g. test_gru_dropout test_lstm_dropout test_rnn_vanilla_dropout), it should be sufficient to check dropout once here.
CC: @ce1adon to double check the above statement.

ce1adon:
@atamazov Hey Artem, maybe we should name it reduce number of tests instead of workaround. Please notice that what we did is to reduce test number to the limit. A real "workaround" to the test would be a complete change of the test rule to test RNG path and dropout path separately. Hope this help your evaluation.

atamazov:
@ce1adon 
> @atamazov Hey Artem, maybe we should name it reduce number of tests instead of workaround. 

This question is: Does this (reducing the number of tests) affects the quality of testing?

> A real "workaround" to the test would be a complete change of the test rule to test RNG path and dropout path separately.

Can you please explain a little bit what is the purpose of the separation of RNG and dropout? In other words, why we need it? Please put this explanation in #1128. Thanks.

ce1adon:
> This question is: Does this (reducing the number of tests) affects the quality of testing?

I would say no but I won't be surprised that different people have different opinions.

> Can you please explain a little bit what is the purpose of the separation of RNG and dropout? In other words, why we need it? Please put this explanation in #1128. Thanks.

We can do that later. I explained it many times and got approval but there are always something with higher priority blocking this change.

atamazov:
@ce1adon 
> I would say no but I won't be surprised that different people have different opinions.

Your opinion is the most important for me. So, this PR is really a fix (not a workaround). Good!

> I explained it many times and got approval...

Okay. It seems that I somehow missed the explanations. I would appreciate if you provide me with a link to the explanation (if it is somewhere on github). If not, then no problem.

PR #1114, opened 2021-08-25
[NHWC][asm igemm] add support for several NHWC bwd ssd config, when k=4x, 2x

carlushuang:
```
convfp16 -n 120 -c 256 -H 1 -W 1 -k 340 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
convfp16 -n 120 -c 256 -H 3 -W 3 -k 340 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
convfp16 -n 120 -c 256 -H 5 -W 5 -k 510 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
convfp16 -n 120 -c 512 -H 10 -W 10 -k 510 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
convfp16 -n 120 -c 512 -H 19 -W 19 -k 510 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
convfp16 -n 120 -c 256 -H 38 -W 38 -k 340 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
```

SSD configs of above are currently not supported in MIOpen. This PR add support for these configs by adding several new kernels which support padded gemm k.

~~This PR has only 1 `.cpp` file and 1 `.txt` file changed, other is asm files. Need feedback of #1113 first then consider if continue integrate asm file as plain text or in other form like .tar.bz2~~ keep current process as discussed in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1113#issuecomment-907304877

[...]

PR #1035, opened 2021-07-15
[gfx90a][conv][nhwc]Asm igemm on gfx90a-fwd/bwd/wrw fp32/fp16

shaojiewang:
Asm igemm implementation-nhwc layout:
||fwd|bwd|wrw|
| ------ | ------ | ------ | ------ |
|fp32| OK | OK | OK |
|fp16| OK | OK | OK |

test script: [test_nhwc](https://amdcloud-my.sharepoint.com/:u:/g/personal/shaowang_amd_com/EZRO3OCUiU1Ks7Y8zT9znJ4BQxSQ2QJHy12mTHc5-fVdSQ?e=8bzcA1)
perf_data: [perf_data](https://amdcloud-my.sharepoint.com/:x:/g/personal/shaowang_amd_com/Ee9OVUVQ4ixJu9s3IX3gjwIBr4nrc44yidtkL4snos4TkA?e=CGDeVV)

PR #1136, opened 2021-09-02
[NHWC][asm igemm][GFX90a] add support for several NHWC bwd ssd config, when k=4x, 2x

carlushuang:
this is the same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1114, but for gfx90a kernels

SSD test result is [gfx90a-ssd-log.txt](https://amdcloud-my.sharepoint.com/:t:/g/personal/carhuang_amd_com/ESKK-MlHRmZFtCMKN-g51RkBfS1lWuit3X_3p-4c4XsY2g?e=ZuVzLS)


atamazov (approved):
I do not see anything bad.

atamazov (changes requested):
Oops, we need tests.

carlushuang:
> Oops, we need tests.

Hi @atamazov , test is also added in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1114, that would also apply for gfx90a

atamazov:
@carlushuang 
> Hi @atamazov , test is also added in #1114, that would also apply for gfx90a

Hmm... But if #1114 includes tests for features implemented in  **_this PR_**, then #1114 will never pass CI, right?

carlushuang:
@atamazov #1114 didn't enable gfx90a, so that PR it self will be OK. Again, gfx90a kernel and the related ctest is enabled in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1035, and that PR it self will be OK. So I may think the tip develop will fail on this test cases, after above PR is merged but this PR is not merged.

junliume (approved):

junliume:
> Oops, we need tests.

@atamazov  could you re-review with the responses? Thanks!

PR #1137, opened 2021-09-02
[NHWC][asm igemm][GFX908][GFX90a] fix fwd kernel that have computation error in small cyx case

carlushuang:
fwd asm igemm will have computation error, when c*y*x is smaller than gemm_k size, e.g:
```
convfp16 -n 16 -c 3 -H 16 -W 16 -k 64 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 -t 1 -F 1 --in_layout NHWC --fil_layout NHWC   --out_layout NHWC
```

This PR fix this issue for both gfx90a/gfx908, and for fp32/fp16 fwd asm NHWC igemm.


atamazov:
@carlushuang Please answer the following questions in the PR description:
- Does it affect performance?
- Does it require re-tuning? Which Solver(s)?

carlushuang:
Hi @atamazov it does not affect performance. Previous when cyx is small (like C=3, Y=X=1), will have computation error. This PR just fix this case. And till now we don't need re-tuning, because currently no model will use such small configuration. (sorry I seems edited previous comment by accident. reverted )

junliume (approved):

PR #809, opened 2021-03-21
Winograd Direct asm workaround for gfx90a

Slimakanzer:
This PR adds winograd and direct asm workaround for `gfx90a`.

## Summary of code changes
* winograd workaround kernels (only `fp32`)
* restrict multipass-winograd for _gfx90a_
* direct asm workaround for vgpr _64bit_ alignment

## Local verification
* gfx906, rocm 4.0
* gfx90a, only kernel compilation

PR #1142, opened 2021-09-07
Explicitly disable asm solvers if XNACK is enabled

Slimakanzer:
By @atamazov: From now on we need to officially support XNACK enabled GPUs. This PR adds very simple support in Solvers that use ASM kernels: so far, none ASM kernels were are ready for XNACK enabled GPUs, so we need to disable them.

Also reverts some incorrect fixes introduced in #809 and #1110 (details [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1142#issuecomment-914681719))

<details><summary><b><i>Previous version of description</i></b></summary><hr>

## Description
As I [mentioned](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1093#issuecomment-914586825) in the #1093, some asm kernels may not support **xnack**, but this is not checked anywhere.
This pull request temporarily disables all asm solvers if **xnack** is enabled.

Next, asm kernels must be checked for support **xnack**.

## Side effect
winogradRxSf2x3 now works correctly on the gfx90a with disabled **xnack**
<hr></details>

### DIFF
--- a/src/solver/conv_asm_dir_BwdWrW3x3.cpp
+++ b/src/solver/conv_asm_dir_BwdWrW3x3.cpp
@@ -351,6 +351,11 @@ bool ConvAsmBwdWrW3x3::IsApplicable(const ConvolutionContext& params) const
         return false;
     if(!params.rmv.IsV2orV3())
         return false;
+
+    const auto target = params.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
+
     const std::string name = params.GetStream().GetDeviceName();
     if(!(StartsWith(name, "gfx8") || StartsWith(name, "gfx9")) || name == "gfx90a")
         return false;
--- a/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
@@ -668,7 +668,8 @@ bool ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::IsApplicable(const ConvolutionC
 
     const auto target = ctx.GetStream().GetTargetProperties();
     if(target.Xnack() && *target.Xnack())
-        return false;
+        return false; // NOLINT (readability-simplify-boolean-expr)
+
     return true;
 }
 ConvSolution ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::GetSolution(
--- a/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
@@ -665,6 +665,10 @@ bool ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::IsApplicable(const ConvolutionC
 
     if(!ctx.IsLayoutNHWC())
         return false;
+
+    const auto target = ctx.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
     return true;
 }
 ConvSolution ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::GetSolution(
--- a/src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
@@ -583,6 +583,10 @@ bool ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC::IsApplicable(const ConvolutionC
 
     if(!ctx.IsLayoutNHWC())
         return false;
+
+    const auto target = ctx.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
     return true;
 }
 ConvSolution ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC::GetSolution(
--- a/src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
@@ -586,7 +586,8 @@ bool ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC::IsApplicable(const ConvolutionC
 
     const auto target = ctx.GetStream().GetTargetProperties();
     if(target.Xnack() && *target.Xnack())
-        return false;
+        return false; // NOLINT (readability-simplify-boolean-expr)
+
     return true;
 }
 ConvSolution ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC::GetSolution(
--- a/src/solver/conv_asm_implicit_gemm_gtc_wrw_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_wrw_nhwc.cpp
@@ -616,6 +616,10 @@ bool ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC::IsApplicable(const ConvolutionC
 
     if(!ctx.IsLayoutNHWC())
         return false;
+
+    const auto target = ctx.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
     return true;
 }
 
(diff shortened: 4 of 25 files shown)


src/solver/conv_winoRxS_f2x3.cpp
```diff
@@ -44,8 +44,6 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3_PERF_VALS)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3_G1)
 
-#define WORKAROUND_ISSUE_1093 1
```
junliume, line 47: @Slimakanzer If we are removing this WORKAROUND (which is great, BTW), could we please also update:
> test/CMakeLists.txt

to reflect the changes? e.g. delete line 201-205:
>    if(WORKAROUND_ISSUE_1093)
>           list(APPEND SKIP_TESTS test_find_db test_main test_immed_conv2d)
>    endif()

atamazov: +1

atamazov:
@Slimakanzer Please revise all cases when some assembly Solver is disabled specifically for gfx90a but enabled for other gfx9 parts (see #809 for example). Actually gfx90a should be able to execute any gfx9 assembly shaders, so such changes should be reverted. The XNACK check should do the necessary job instead. Thanks!

atamazov:
Description updated. Converted to draft.

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
```diff
@@ -583,6 +583,10 @@ bool ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC::IsApplicable(const ConvolutionC
 
     if(!ctx.IsLayoutNHWC())
         return false;
+
+    const auto target = ctx.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
```
junliume, line 589: I got a feeling (unproven) that if you add an empty line after line 589, clang tidy will stop complaining [readability-simplify-boolean-expr]

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -665,6 +665,10 @@ bool ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::IsApplicable(const ConvolutionC
 
     if(!ctx.IsLayoutNHWC())
         return false;
+
+    const auto target = ctx.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
```
junliume, line 671: Same as this one: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1142/files#r703911724

src/solver/conv_asm_implicit_gemm_gtc_wrw_nhwc.cpp
```diff
@@ -616,6 +616,10 @@ bool ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC::IsApplicable(const ConvolutionC
 
     if(!ctx.IsLayoutNHWC())
         return false;
+
+    const auto target = ctx.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
```
junliume, line 622: Same as this one: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1142/files#r703911724

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
atamazov: It wants `return !(target.Xnack() && *target.Xnack())` but this is :shit:

atamazov: I recommend shutting it up.

src/solver/conv_asm_dir_BwdWrW3x3.cpp
```diff
@@ -351,6 +351,11 @@ bool ConvAsmBwdWrW3x3::IsApplicable(const ConvolutionContext& params) const
         return false;
     if(!params.rmv.IsV2orV3())
         return false;
+
+    const auto target = params.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
+
     const std::string name = params.GetStream().GetDeviceName();
     if(!(StartsWith(name, "gfx8") || StartsWith(name, "gfx9")) || name == "gfx90a")
```
atamazov, line 360:
```suggestion
    if(!(StartsWith(name, "gfx8") || StartsWith(name, "gfx9")))
```

atamazov (changes requested):
(this is required wherever you check XNACK)

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
junliume: It did not complain other files, so I guess its static code check got confused with the 
>    if(target.Xnack() && *target.Xnack())
        return false;
return true;

maybe just change it to
>    if(target.Xnack() && *target.Xnack())
        return false;
<--- empty line here ---> (non proven WA)
>return true;

src/solver/conv_asm_dir_BwdWrW3x3.cpp
atamazov: Okay. But we need to create tickets for individual Solvers then (or one umbrella ticket with a checklist of Solvers). May I ask you to do this while working on this PR?

Let me explain an approach in general.

Each case when an asm shader is enabled for gfx9 but specifically disabled for gfx90a should be considered as a workaround. Each workaround requires resolving. In this specific case, resolving could be one of:
- Checking & fixing the kernel (checking VGPR tuple alignment etc. as you explained above) and enabling it.
- Or, finally disabling the Solver for gfx90a with clear explanation of the reasons (in C++ comments), like
  - "We have a xDLOPs counterpart Solver for gfx90a"
  - "This Solver shows poor performance for gfx90a and we have another, faster Solvers for it, so it does not worth fixing it for gfx90a"
  - "This Solver uses some gfx9 instructions (...) that not exist on gfx90a and cannot be fixed for it" etc.

/cc @Kirpich30000

src/solver/conv_winoRxS_f2x3.cpp
Slimakanzer: Thank you! Fixed in cd83762

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
Slimakanzer: Thank you! Fixed in 0da6aa2

junliume:
@Slimakanzer could you please take a look:
> 54 - test_tensor_vec (Failed) for Int8 HIP All Vega20
ref: http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/asm-xnack-workaround/2/pipeline

atamazov (approved):
LGTM!

src/solver/conv_winoRxS_f2x3.cpp
```diff
@@ -453,12 +451,12 @@ static bool IsApplicableBase(const ConvolutionContext& params)
...
+
     const auto name = params.GetStream().GetDeviceName();
-#if WORKAROUND_ISSUE_1093
-    if(!(StartsWith(name, "gfx9") || StartsWith(name, "gfx10")) || name == "gfx90a")
-#else
     if(!(StartsWith(name, "gfx9") || StartsWith(name, "gfx10")))
-#endif
         return false;
     if(params.IsFp16() &&
        !(StartsWith(name, "gfx906") || StartsWith(name, "gfx908") || StartsWith(name, "gfx1011") ||
```
atamazov, line 462: @Slimakanzer Would this work on gfx90a?

atamazov:
@junliume This also enables `ConvBinWinogradRxSf2x3` for `gfx90a:xnack-` and removes W/A for #1093. Unfortunately we forgotten to update the PR description and (more importantly) _commit message_.

src/solver/conv_winoRxS_f2x3.cpp
Slimakanzer: @atamazov No, currently _ConvBinWinogradRxSf2x3_ does not support fp16 on gfx90a due to verification errors. I'm working on it.

Issue #1146, opened 2021-09-09
[gfx90a] check asm solver applicability


Slimakanzer:
At the moment, we cannot simply add _gfx90a_ to the list of supported architectures in asm solvers, as assembly errors may occur. _gfx90a_ requires new `.amdhsa_accum_offset` directive and VGPR tuple 64bit alignment. We have already encountered these problem in #888, #809 and SWDEV-270970.

As mentioned in [this](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1142#discussion_r704625595) comment, we need to decide what to do with each asm solver that supports _gfx9_:
- Checking & fixing the kernel and enabling it.
- Or, finally disabling the Solver for gfx90a with clear explanation of the reasons (in C++ comments), like
    - "We have a xDLOPs counterpart Solver for gfx90a"
    - "This Solver shows poor performance for gfx90a and we have another, faster Solvers for it, so it does not worth fixing it for gfx90a"
    - "This Solver uses some gfx9 instructions that not exist on gfx90a and cannot be fixed for it" etc. 

-----

- [ ] ConvBinWinograd3x3U
- [ ] ConvBinWinogradRxS
- [ ] ConvMPBidirectWinograd
- [ ] ConvWinograd3x3MultipassWrW
- [ ] ConvBinWinogradRxSf3x2
- [ ] ConvAsm3x3U
- [ ] ConvAsm5x10u2v2f1
- [ ] ConvAsm5x10u2v2b1
- [ ] ConvAsm7x7c3h224w224k64u2v2p3q3f1
- [ ] ConvAsmBwdWrW3x3
- [ ] ConvAsmImplicitGemmV4R1DynamicFwd
- [ ] ConvAsmImplicitGemmV4R1DynamicBwd
- [ ] ConvAsmImplicitGemmV4R1DynamicWrw
- [ ] miopenGcnAsmBNBwdTrainSpatial (batchnormocl)

carlushuang:
instead we never check if the previous solver for 908 can run on 90a... let me try this first

atamazov:
> instead we never check if the previous solver for 908 can run on 90a... let me try this first

They may randomly fail due to this:

> gfx90a requires new .amdhsa_accum_offset directive and VGPR tuple 64bit alignment.

So even if they _seem_ to be working on gfx90a, we still need to check the source code.

shaojiewang:
[...]

PR #1150, opened 2021-09-10
[gfx90a] explicity disable asm solvers

Slimakanzer:
This PR disables asm solvers for gfx90a and marks as _WORKAROUND_ for #1146.

### DIFF
--- a/src/solver/conv_asm_implicit_gemm_bwd_v4r1_dynamic.cpp
+++ b/src/solver/conv_asm_implicit_gemm_bwd_v4r1_dynamic.cpp
@@ -31,8 +31,6 @@
 #include <algorithm>
 #include <miopen/solver/implicitgemm_util.hpp>
 
-#define WORKAROUND_ISSUE_1146 1 // check asm solver applicability for gfx90a
-
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_BWD_V4R1)
 
 namespace miopen {
@@ -135,10 +133,6 @@ bool ConvAsmImplicitGemmV4R1DynamicBwd::IsApplicable(const ConvolutionContext& c
         return false;
 
     const auto device_name = ctx.GetStream().GetDeviceName();
-#if WORKAROUND_ISSUE_1146
-    if(device_name == "gfx90a")
-        return false;
-#endif
     if(!(StartsWith(device_name, "gfx900") || StartsWith(device_name, "gfx906")))
         return false;
 
--- a/src/solver/conv_asm_implicit_gemm_bwd_v4r1_dynamic.cpp
+++ b/src/solver/conv_asm_implicit_gemm_bwd_v4r1_dynamic.cpp
@@ -31,6 +31,8 @@
 #include <algorithm>
 #include <miopen/solver/implicitgemm_util.hpp>
 
+#define WORKAROUND_ISSUE_1146 1 // check asm solver applicability for gfx90a
+
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_BWD_V4R1)
 
 namespace miopen {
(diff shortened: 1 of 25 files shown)


src/solver/conv_asm_implicit_gemm_bwd_v4r1_dynamic.cpp
```diff
@@ -133,6 +135,10 @@ bool ConvAsmImplicitGemmV4R1DynamicBwd::IsApplicable(const ConvolutionContext& c
         return false;
 
     const auto device_name = ctx.GetStream().GetDeviceName();
+#if WORKAROUND_ISSUE_1146
+    if(device_name == "gfx90a")
```
shaojiewang, line 139: In fact due to line 142, if device name is gfx90a, this solver will return false. So the W/A might not be necessary.

junliume: +1, it really depends on how explicitly you want it to be (and for later reference to get back to this issue). While technically I agree that it is redundant.

src/solver/conv_asm_implicit_gemm_v4r1_dynamic.cpp
```diff
@@ -278,6 +280,10 @@ bool ConvAsmImplicitGemmV4R1DynamicFwd::IsApplicable(const ConvolutionContext& c
         return false;
 
     const auto device_name = ctx.GetStream().GetDeviceName();
+#if WORKAROUND_ISSUE_1146
```
junliume, line 283: Similarly, redundant conditions due to line 281/287?

src/solver/conv_asm_implicit_gemm_wrw_v4r1_dynamic.cpp
```diff
@@ -300,6 +302,10 @@ bool ConvAsmImplicitGemmV4R1DynamicWrw::IsApplicable(const ConvolutionContext& c
         return false;
 
     const auto device_name = ctx.GetStream().GetDeviceName();
+#if WORKAROUND_ISSUE_1146
```
junliume, line 305: Similar to previous comment.

src/solver/conv_bin_winoRxS.cpp
```diff
@@ -246,6 +248,11 @@ bool ConvBinWinogradRxS::IsApplicable(const ConvolutionContext& params) const
 
     const auto name = params.GetStream().GetDeviceName();
     const bool fp16 = params.IsFp16();
+
+#if WORKAROUND_ISSUE_1146
```
junliume, line 252: Similar to previous comments

src/solver/conv_asm_implicit_gemm_bwd_v4r1_dynamic.cpp
atamazov: Yes, this should be reverted as we have other Solvers for devices that have MAI (and that is why this Solver is not enabled for MI100). #1146 was updated recently.

src/solver/conv_asm_implicit_gemm_v4r1_dynamic.cpp
atamazov: +1

src/solver/conv_asm_implicit_gemm_wrw_v4r1_dynamic.cpp
atamazov: +1

src/solver/conv_bin_winoRxS.cpp
atamazov: @junliume This Solver is enabled on gfx908. We are going to to enable it on gfx90a as well. Even if technically is not necessary, I would better keep it here as a reminder.

@Slimakanzer [Recommendation] Rework this code in order to make the W/A really necessary ;)

atamazov (changes requested):
Please resolve:
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1150#pullrequestreview-752153033
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1150#pullrequestreview-752308274
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1150#pullrequestreview-752308409

Please also look at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1150#pullrequestreview-752308704 and decide yourself what to do.

atamazov:
@shaojiewang Thanks for reviewing!

src/solver/conv_bin_winoRxS.cpp
Slimakanzer: @atamazov 
Thank you! Reverted this unnecessary restriction (see [explanation](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1146#issuecomment-919021192))

atamazov (approved):
LGTM!

Issue #50, opened 2018-08-24
RCP tool fails to generate output for ROCm DeepBench 


srikantAmd:
Trying to collect performance counters for DeepBench apps of the ROCm repo here:
https://github.com/ROCmSoftwarePlatform/DeepBench

The applications run correctly with the following command line. Example:
`rcprof -t ./conv_bench`
But fail to generate any output:
`Failed to generate profile result path/to/out.atp`

Further, if we try to run it with the HSA mode of RCP, the application totally fails to run. Example:
```
$ rcprof -A ./conv_bench
Radeon Compute Profiler V5.3.0 is enabled
Failed to generate profile result path/to/out.atp
```

dagamayank:
Closing this issue as it is not related to MIOpen.

PR #1162, opened 2021-09-19
[CI] HOTFIX: Support Ubuntu 20.04. Fix: evaluate TARGET param in 'when' before entering agent. Syntax fixes/workarounds.

atamazov:
- [CI] HOTFIX: Create `render` group in the docker image. Add jenkins user to `render` group for Ubuntu 20.04. 
- [CI] Fix: evaluate TARGET param in 'when' before entering agent.
- [CI] Fix strange syntax errors

Resolves the following issues:
- :red_circle: CI fails on nodes with Ubuntu 20.04 (I see this on gfx90a nodes at least)
- :yellow_circle: CI tests takes too long time if subset of TARGETS is selected

### CI fails on nodes with Ubuntu 20.04

See https://rocmdocs.amd.com/en/latest/Installation_Guide/Installation-Guide.html#setting-permissions-for-groups.

### CI tests takes too long time if subset of TARGETS is selected

> @junliume ...observed here: jenkins / MLLibs/MIOpen / wip-rocmtest-trial-vega10 / #50 (amd.com).  It seem that several parallel stages are queued where they should not...
>
> Run `#50` timed out after 5 hours. Yes, the three “green” ones are needed and correct behavior. However, after these three are finished, it seems that the Smoke FP32 stage is “waiting” for the following stages to begin:
> - Fp32 Hip Debug gfx908 /opt/rocm
> - Fp32 OpenCL Debug gfx1030
> - Fp32 OpenCL gfx908
>
> These three stages are MI100 and should not run with Vega10 trial branch, right? I thought the correct behavior would be moving on to the next parallel stage, but the #50 pipeline stuck at Smoke FP32 till 5 hours time out.

For normal development/testing process, the issue only happens when you manually de-select one or more TARGET parameters for testing. ~However this seriously affects the `wip-rocmtest-trial-*` branches and therefore may hinder DevOps (which use the trial branches often).~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -124,3 +124,4 @@ RUN if [ "$USE_MLIR" = "ON" ]; \
     $PREFIX/bin/cmake --install . --component libMLIRMIOpen --prefix /opt/rocm && \
     cd ~ && rm -rf llvm-project-mlir-$MLIR_COMMIT; fi
     
+RUN groupadd -f render
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -115,7 +115,7 @@ def buildHipClangJob(Map conf=[:]){
         def target_id = conf.get("target_id", "OFF")
         def mlir_build = conf.get("mlir_build", "OFF")
         def build_fin = conf.get("build_fin", "OFF")
-        def dockerOpts="--device=/dev/kfd --device=/dev/dri --group-add video --cap-add=SYS_PTRACE --security-opt seccomp=unconfined"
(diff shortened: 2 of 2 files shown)


atamazov:
@okakarpa The fix is cherry-picked to all the trial branches. Lowering urgency.

atamazov:
New problem: The docker user must be added to the `render` group. Otherwise testing will fail on Ubuntu 20.04. More info at https://rocmdocs.amd.com/en/latest/Installation_Guide/Installation-Guide.html#setting-permissions-for-groups. PR description updated.

junliume (approved):

PR #1165, opened 2021-09-21
Merge develop into CK_upstream (Please don't squash when merging)

asroy:
```CK_upstream``` is the branch for sync between MIOpen and CK.

Please don't squash when merging ```develop``` into ```CK_upstream```, to keep the commit history simple


junliume (approved):
@asroy this PR is more a procedure to keep consistency and history. Please be careful not to select the "default" merge option, select from the drop down and first one would be okay?

atamazov:
@asroy The `CK_upstream` branch is not protected, so you can merge this PR (but approval is still required).

BTW I do not see any special need to create PRs for updates of `CK_upstream`, but you know better.

atamazov (approved):

junliume (approved):

junliume:
@asroy new reviews will be dismissed whenever develop is updated. Since at this current commit (2d7b2dd76464ac0fe052118718c4ab217341d25b) we will do a staging anyway, thus merging it back to CK

Issue #1167, opened 2021-09-22
Properly fix W/A from #240, commit 661308bd7 - [test_conv_for_implicit_gemm] Disable config that fails often.


atamazov:
Commit that introduces the W/A: https://github.com/ROCmSoftwarePlatform/MIOpen/commit/661308bd74d96c7ce330c94394b0a43a2bd6fa21

Where it resides now: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2d7b2dd76464ac0fe052118718c4ab217341d25b/test/CMakeLists.txt#L723

Let's revert the W/A, check if problem still persists and provide a proper fix.

junliume:
Let's see what might or might not pass with this WA reverted:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/fix_%231167/1/pipeline/

ppanchad-amd:
@atamazov, Is this ticket still relevant?

atamazov:
@ppanchad-amd Yes. The W/A was silently removed in #2513, however, in test/gtest/conv_for_implicit_gemm.cpp, we still need to re-enable the config mentioned in the description of this issue.

/cc @ozturkosu @junliume

JonathanLichtnerAMD:
Closing since MIOpen is being migrated to [ROCm/rocm-libraries](https://github.com/ROCm/rocm-libraries).  Please re-open there if you believe this is still an issue.

PR #240, opened 2020-05-25
Fix Jenkins issue

atamazov:
Thix PR contains several changes. ___PLEASE DO NOT SQUASH COMMITS WHEN MERGING.___ Also I recommend reviewing each commit individually.

- ___Fixes:___
  - adf4fdb Fixes file-based binary cache, https://github.com/AMDComputeLibraries/MLOpen/issues/2450
  - 1ff1a09 Tidy fix that is required after merging of #167 but missing in #212.
  - 167a6da1 Permanent workaround for ConvOclBwdWrW53 (NaN on gfx908)
- ___Workarounds:___
  - 0335a64 Switches to using the file-based binary cache by default.
    - This should resolve this https://github.com/ROCmSoftwarePlatform/MIOpen/issues/226#issuecomment-633659969
  - a019ef7 [test_sqlite_perfdb] Reduce amount of work.
  - 8ea95121 Disabled `ConvHipImplicitGemmV4R4GenXdlopsWrWFp32`. Reworked special case in `ConvHipImplicitGemmV4R4GenWrWXdlops::IsApplicable()`. 
  - 661308bd [test_conv_for_implicit_gemm] Disable config that fails often.
  - b4d68057 [test_soft_max] Workaround for cases that
    - often fail on Jenkins
    - always fail with ROCm3.3/Radeon VII
  - ea710fe4 [test_rnn_vanilla] [test_rnn_extra] The "Forward Inference RNN vanilla" sub-test is disabled.
  - 9899bdd0 [test_ctc] Disabled some tests that fail on Jenkins from time to time.
  - c71697a2 [test_lstm] The "Forward Inference LSTM" sub-test is disabled.
  - a983655c [test_conv_group] Disabled WrW configs that fail during "Full long tests"
- ___Improvements:___
  - 441bb75 + d15e895c [Jenkins] Added possibility to customize env settings for "make check".
  - a9a1a1e Tests reordered to better utilize parallelizm of `ctest`.
      https://github.com/ROCmSoftwarePlatform/MIOpen/blob/1ff1a09352d7d14f0ca26ecab31820cd4e36797e/test/CMakeLists.txt#L153-L163
  - 0ae054c4 Disabled `test_conv_for_implicit_gemm` for INT8. Disabled many custom tests that were previously run with implied `--float` during non-FP32 Jenkins jobs. These tests are just wasting time when run during INT8 or HALF stages.
[...]

PR #1170, opened 2021-09-23
[CI][Tests] remove workaround for test_conv_for_implicit_gemm from #240 (#1167)

junliume:
ref: #1167 
re-enabling it seems to be stable in the tip of develop

This PR is a housekeeping mini-maintenance suggested by Artem


junliume:
@atamazov please review. a very small step towards your goal of resolving the W/As :)

atamazov:
@junliume Thanks! unfortunately there is `WORKAROUND_ISSUE_936` which seems terribly excessive and may mask possibly uncovered problems:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/030a3695d5f899afbd484338558fbb70a4b204d1/test/CMakeLists.txt#L693

Let's rework that workaround first (I am working on it with @shurale-nkn)

junliume:
> @junliume Thanks! unfortunately there is `WORKAROUND_ISSUE_936` which seems terribly excessive and may mask Fwd/Bwd FP16 issues (possibly uncovered by this PR):
> https://github.com/ROCmSoftwarePlatform/MIOpen/blob/030a3695d5f899afbd484338558fbb70a4b204d1/test/CMakeLists.txt#L693
> 
> Let's rework that workaround first (I am working on it with @shurale-nkn)

Thanks! Let's keep this PR **On Hold** then.

junliume:
> I recommend holding this off until my work with `wa-936-cleanup` (no PR yet) is done. Unfortunately I am totally busy with hiprtc right now...

Sure! That's why it is still in draft phase, I am reverting a few of these W/As and use quiet time of CI systems to observe if there are any issues. :)

PR #1171, opened 2021-09-23
accept list of solvers in MIOPEN_DEBUG_FIND_ONLY_SOLVER

JehandadKhan:
This PR allows MIOPEN_DEBUG_FIND_ONLY_SOLVER to comprise of multiple solvers separated by semicolon. For example, when debugging issues in frameworks, one needs multiple solvers (different directions). This PR enables such a functionality.

### DIFF
--- a/src/find_controls.cpp
+++ b/src/find_controls.cpp
@@ -31,6 +31,9 @@
 #include <miopen/logger.hpp>
 #include <miopen/env.hpp>
 #include <miopen/solver_id.hpp>
+#include <miopen/stringutils.hpp>
+
+#include <boost/optional.hpp>
 
 #include <ostream>
 #include <cstdlib>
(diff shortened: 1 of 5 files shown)


src/find_controls.cpp
```diff
@@ -97,29 +100,45 @@ FindEnforceAction GetFindEnforceAction()
     return val;
 }
 
-solver::Id GetEnvFindOnlySolverImpl()
+boost::optional<std::vector<solver::Id>> GetEnvFindOnlySolverImpl()
```
junliume, line 103: @JehandadKhan  just curious, std::optional is available in C++17 which should be supported by our compiler. Any particular reasons we are using C++14? I see that Composable Kernel is already using C++17.

JehandadKhan: The reason for not using std version of optional is C++14 on the host side. I am all for advancing the version c++17! @pfultz2 Can you comment on the historical reasons for sticking with C++14 ?

junliume:
> This PR allows MIOPEN_DEBUG_FIND_ONLY_SOLVER to comprise of multiple solvers separated by semicolon. For example, when debugging issues in frameworks, one needs multiple solvers (different directions). This PR enables such a functionality.

@JehandadKhan could you add the use case and option in
https://rocmsoftwareplatform.github.io/MIOpen/doc/html/DebugAndLogging.html
The documentation has not been updated for a while, this new feature could be very useful. Thanks!

src/find_controls.cpp
pfultz2: We were using C++14 because that is highest we could go with gcc 5 in ubuntu 16.04. The kernels can always use a higher version of C++ since it uses our clang compiler and doesn't rely on C++ runtime. Using features like `std::optional` depends on the C++ runtime more than language features of the compiler, and gcc 5 does provide experimental support for `optional`(in migraphx we have [this header](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/optional.hpp#L42) to use `optional` on gcc 5/C++14).

Our host code should always be able to compile with gcc(we currently dont do this for HIP backend due to historical reasons of using `hipExtModuleLaunchKernel`, but we can use an [extern declaration](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/targets/gpu/kernel.cpp#L8) to workaround this issue). However, if we drop opencl, then we should drop needing `CXX=/opt/rocm/llvm/bin/clang++` for HIP.

Finally, I think its reasonable to switch to C++17 as we no longer support ubuntu 16.04. Although, I think it would be better to have a seperate PR to switch to `std::optional` and `std::filesystem` with C++17.

junliume (approved):

atamazov (commented):
This change replaces trivial `==` with something more complex, like `(std::find(find_only->begin(), find_only->end(), Id{SolverDbId(solver)}) == find_only->end())` so I would recommend encapsulating this functionality into a function. Possibly, somewhere nearby definition of `GetEnvFindOnlySolver()`.

Otherwise looks good.

PR #1174, opened 2021-09-23
Backward spatial single-kernel solver

DrizztDoUrden:
Both my and @muralinr benchs have passed. No perf or build parameters difference.


muralinr (dismissed):
Changes look good.  I see that you mentioned there is no performance impact and functional testing issues for my benchmark network configs and hence approving this PR.

muralinr (approved):
Approved.  Looks good after tidy fix.

junliume (approved):

PR #1175, opened 2021-09-24
Bump up the MIOpen version to 2.15 (after cutting the rocm-rel-4.5 branch)

atamazov:


junliume (approved):
@JehandadKhan  we may need the release notes to be updated too

PR #1179, opened 2021-09-24
[MLIR] Bump ROCm version to 4.3.1

jerryyin:
I get reminded by @JehandadKhan that MLIR used a different ROCm version then rest of stages. Bumping them to parallel version now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN dpkg --add-architecture i386
 # unless MLIR library is incompatible with current ROCm.
 
 RUN if [ "$USE_MLIR" = "ON" ] ; \
-        then export ROCM_APT_VER=.apt_4.2;\
+        then export ROCM_APT_VER=.apt_4.3.1;\
     else export ROCM_APT_VER=.apt_4.3.1;  \
     fi && \
 echo $ROCM_APT_VER &&\


Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN dpkg --add-architecture i386
 # unless MLIR library is incompatible with current ROCm.
 
 RUN if [ "$USE_MLIR" = "ON" ] ; \
-        then export ROCM_APT_VER=.apt_4.2;\
+        then export ROCM_APT_VER=.apt_4.3.1;\
```
JehandadKhan, line 18: @jerryyin  why not just remove the condition alltogether ?

JehandadKhan (changes requested):

Dockerfile
atamazov: @JehandadKhan I think this allows @jerryyin & Co to promote ROCm version for MLIR independently.

atamazov: So when we upgrade to 4.4, MLIR will continue using 4.3.1 until 4.4 is explicitly tested, approved and taken into use by MLIR team.

atamazov (approved):

Dockerfile
junliume: Thanks! Since update of ROCm env is not synced between MIOpen and MLIR yet, I believe it is better to keep that flexibility in place too.

junliume (approved):

Dockerfile
jerryyin: Correct, since the two components does not bump ROCm version at the same time it is better to leave it separate. That's also why we have comments above in the dockerfile:

> Note: The ROCm version with $USE_MLIR should keep in sync with default ROCm version
> unless MLIR library is incompatible with current ROCm.

It gives MIOpen developer the flexibility to make MLIR use a different version of ROCm, if upgrade would break MLIR somehow (It most recently happened in ROCm 4.1). If this kind of breakage becomes less often, I will consider to just get rid of the condition.

PR #1192, opened 2021-09-28
Enable fused asm kernels for gfx90a

Slimakanzer:
This PR adds gfx90a to the list of supported archs.
This allows asm kernels to be used with Fusion API.

## Local verification
- gfx90a, rocm 4.3.1

### DIFF
--- a/src/md_graph.cpp
+++ b/src/md_graph.cpp
@@ -721,8 +721,8 @@ void FusionMDGraph::InitConv(FusionMDGraph& g)
                 add_relu(kernel_file, kernel_name, vc, WinogradV21NodeArgs, supported_arch);
             };
 
-            add_v21_wino("gfx9", {"gfx900", "gfx906", "gfx908"}, 1);
-            add_v21_wino("gfx9", {"gfx900", "gfx906", "gfx908"}, 2);
+            add_v21_wino("gfx9", {"gfx900", "gfx906", "gfx908", "gfx90a"}, 1);
+            add_v21_wino("gfx9", {"gfx900", "gfx906", "gfx908", "gfx90a"}, 2);
             add_v21_wino("gfx10", {"gfx1011", "gfx1012", "gfx1030"}, 1);
             add_v21_wino("gfx10", {"gfx1011", "gfx1012", "gfx1030"}, 2);
         }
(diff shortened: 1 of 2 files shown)


atamazov (changes requested):
This looks good, but not enough. We need to disable some (or all, I do not remember) asm kernels for `xnack+`.

atamazov:
@Slimakanzer What do you think about tests? Does this PR add performance only or functionality as well (in other words, does it extend coverage of fusions on gfx90a)?

Slimakanzer:
@atamazov Yes, this PR also changes Fusion API functionality for _gfx90a_. 
Asm kernels cover some cases that are **not covered by OCL** kernels.

For example, Fused ocl kernel (`Conv-Bias`, `Conv-Bias-Activation`, `Conv-Activation` fusion mode) 
has the following restrictions:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/1c03b54ee3f21d6ef9375472e32287ac84896d85/src/md_graph.cpp#L877-L900

Slimakanzer:
@atamazov Thank you! 
Fused asm kernels are disabled for `xnack+`. Fixed in 6c0bd3e

atamazov:
@Slimakanzer 
> Asm kernels cover some cases that are not covered by OCL kernels.

Can we add some tests (or test cases) to cover these fused configs?

This work would involve extending GPU detection in test/CMakeLists.txt to detect "xnack+" (and allow for manual override as usual).

It is still questionable to me how to implement the functional test.
- Implement special option in `test_mdgraph` + add the option its command line in test/CMakeLists.txt when `xnack+` is NOT detected.
- Or, create a custom test in test/CMakeLists.txt and enable that test if `xnack+` is NOT detected.

What is your opinion?

Slimakanzer:
@atamazov 
The first option looks better, let me try it.

test/CMakeLists.txt
```diff
@@ -204,6 +210,11 @@ endif()
 message(STATUS "SKIP_TESTS: ${SKIP_TESTS}")
 message(STATUS "SKIP_ALL_EXCEPT_TESTS: ${SKIP_ALL_EXCEPT_TESTS}")
 
+# List of tests that uses xnack mode.
+# Parameter that specifies xnack mode: --xnack [0|1]
```
atamazov, line 214:
```suggestion
# List of tests that depend on the XNACK mode.
# Options convention: Tests that depend on the XNACK mode should support the "--xnack" option.
# If "--xnack 0" is specified (this is the default), then such tests should run in XNACK OFF mode.
# If XNACK is enabled on the GPU, then the testing mode must be changed by the --xnack 1 option.
```

PR #1190, opened 2021-09-27
[CI][gfx90a] Establish testing for XNACK enabled

atamazov:
This partially implement https://github.com/ROCmSoftwarePlatform/MIOpen-internal/issues/1.

PR #1194, opened 2021-09-28
[CI] Fix: Add datatype for the new 'Fp32 Hip All gfx90a Xnack+' stage

atamazov:
Leftover after #1190 and #1191

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -849,7 +849,7 @@ pipeline {
                 stage('Fp32 Hip All gfx90a Xnack+') {
                     when {
                         beforeAgent true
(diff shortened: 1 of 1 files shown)


junliume:
> Leftover after #1190 and #1191

@atamazov  technically we only need to run that specific stage instead of a whole set of CI, right? :)

junliume (approved):

atamazov:
@junliume 
> @atamazov technically we only need to run that specific stage instead of a whole set of CI, right? :)

Formally speaking, no. Who knows how many bugs this PR brings into our CI? :/

Thanks for reviewing & merging!

PR #1199, opened 2021-09-30
[MLIR] Deprecate MLIR CPP backend - step 2 of 2

jerryyin:
Completely removed all code related with MLIR cpp backend. This is the second and the final PR in the series.

Previous PR please see 
 - #879

CC: @whchung


src/hipoc/hipoc_program.cpp
```diff
@@ -273,7 +269,7 @@ void HIPOCProgramImpl::BuildCodeObjectInMemory(const std::string& params,
             comgr::BuildHip(filename, src, params, target, binary);
         else if(miopen::EndsWith(filename, ".s"))
             comgr::BuildAsm(filename, src, params, target, binary);
-        else if(miopen::EndsWith(filename, ".mlir-cpp"))
+        else if(miopen::EndsWith(filename, ".mlir"))
             MIOPEN_THROW(miopenStatusNotImplemented, "MLIR builds are not supported with COMgr");
```
atamazov, line 273: @jerryyin We are going to switch to COMGR by default soon due to https://ontrack-internal.amd.com/browse/SWDEV-287843

atamazov (approved):
One of the best PRs I've ever seen.
makeitso

src/hipoc/hipoc_program.cpp
atamazov: @jerryyin 
> If my understanding is right then I need to start working on COMGR + MLIR support when MIOpen is built as a shared library.

Yes, please go ahead. I think you would like to use MLIR as a static lib. But please also discuss with Jeff Daily if this may affect MIOpen build for PyTorch wheels. They use COMGR, but I am not sure if they use MIOpen as a static or dynamic lib. Please make sure that CMake fails when MIOpen is built as a static lib and MLIR is used as a static lib.

jerryyin: @atamazov I have double checked that:

1. In PyTorch shipped wheel, they use dynamic libs, @jeffdaily shared me their script to build MIOpen as [here](https://github.com/pytorch/builder/blob/47e89b9f4a96d016776dd56dc134135a6bca9711/common/install_miopen.sh#L85-L118). Yay! Looks like framework does get the opportunity to use MLIR.
2. On a double look since MLIR is already doing in memory build it really looks like a one line change for COMGR + MLIR to work.

junliume (approved):

src/hipoc/hipoc_program.cpp
atamazov: @jerryyin [off-topic]
> it really looks like a one line change for COMGR + MLIR to work.

I am glad to hear this. This is a goal of proper design ;)

PR #1200, opened 2021-09-30
[MLIR] Enable OpenCL backend

jerryyin:
This PR enabled OpenCL backend for MLIR solver:
 - Added a testing stage for OpenCL test for MLIR as a sanity test
    - Since rest of MLIR kernels are already covered by hip, I avoided add duplicated tests 
 - Amended `mlir_impl_gemm.cpp` to include OpenCL required sections
    - OpenCL do not implicitly linearize all kernel arguments as hip: I make it such that argument are manually expanded
    - In OpenCL there's no easy way to get the kernel address: I placed the ocl memory object such that a `clSetKernelArg` will be invoked on each kernel args
  - Updated `clhelper.cpp` to include mlir build

After this PR, state of OpenCL backend will be an orthogonal topic with MLIR. The release of MIOpen with MLIR will not be limited by its support for OpenCL backend.

Tested locally and make sure MIOpen behave the same between hip/ocl backend.


Jenkinsfile
```diff
@@ -588,6 +588,14 @@ pipeline {
                         buildHipClangJobAndReboot(setup_flags: MLIR_flags, build_env: extra_log_env, test_flags: ' --verbose ', gpu_arch: "gfx908", mlir_build: "ON")
                     }
                 }
+                stage('Fp32 OpenCL MLIR gfx908') {
+                    when { expression { params.TARGET_GFX908 } }
```
junliume, line 592:
```suggestion
                    when {
                        beforeAgent true
                        expression { params.TARGET_GFX908 && params.DATATYPE_FP32 }
                    }
```

jerryyin: Commited as-is. I'm not very familiar with the beforeAgent clause. Doing a bit of the read up it seems to make sure the expression get evaluated before node is selected.  However, the node is also filtered against `gfx908` label, so why does the evaluation sequence matter?

junliume: @jerryyin without it in Jenkins parallel stages it will wait till a node (Agent) is available and then evaluate whether it fits in the expression, causing delays and long wait times when the pipelines are congested.
CC: @atamazov

jerryyin: Make sense, thanks

jerryyin:
Gentle ping @junliume @atamazov, all CI tests have passed please let me know if you have any feedback.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -48,6 +48,15 @@ struct MlirConvArgs
     StridedMemRef5D output;
 };
 
+// Note: Below macros are required for opencl backend only because
+// opencl backend requires to invoke clSetKernelArg on every kernel
+// arguments whereas hip can implicitly linearize the struct to
+// kernel arguments
+#if MIOPEN_BACKEND_OPENCL
+#define EXPAND_ARRAY_5(x) x[0], x[1], x[2], x[3], x[4]
```
atamazov, line 56:
```suggestion
#define EXPAND_ARRAY_5(x) ((x)[0]), ((x)[1]), ((x)[2]), ((x)[3]), ((x)[4])
```
macro args must be enclosed into parentheses

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -48,6 +48,15 @@ struct MlirConvArgs
...
+#define EXPAND_MLIR_CONV_ARGS(x) x.offset, EXPAND_ARRAY_5(x.sizes), EXPAND_ARRAY_5(x.strides)
```
atamazov, line 57: Ditto here

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -212,8 +221,20 @@ InvokerFactory MakeMlirFwdInvokerFactory(const ConvolutionContext& ctx)
                 primitive_parameters.CastTo<conv::DataInvokeParams>();
             const auto& tensors = forward_invoke_params.tensors;
 
+#if MIOPEN_BACKEND_OPENCL
+            handle.Run(kernels[0])(tensors.w,
+                                   tensors.w,
+                                   EXPAND_MLIR_CONV_ARGS(args.filter),
+                                   tensors.in,
+                                   tensors.in,
```
atamazov, line 229: Can you please remind why it is necessary to pass `tensors.in` twice?

src/ocl/clhelper.cpp
```diff
@@ -158,7 +159,9 @@ ClProgramPtr LoadProgram(cl_context ctx,
     else
     {
         program_name = program;
-        if(kernel_src.empty())
+        if(miopen::EndsWith(program_name, ".mlir"))
+            source = "";
```
atamazov, line 163: Why it is necessary to clean source?

src/ocl/clhelper.cpp
```diff
@@ -189,6 +192,15 @@ ClProgramPtr LoadProgram(cl_context ctx,
         bin_file_to_str(hsaco_file, buf);
         return LoadBinaryProgram(ctx, device, buf);
     }
+#if MIOPEN_USE_MLIR
+    else if(miopen::EndsWith(program_name, ".mlir"))
+    {
+        std::vector<char> buffer;
+        MiirGenBin(params, buffer);
+        std::string buf(buffer.begin(), buffer.end());
+        return LoadBinaryProgram(ctx, device, buf);
```
atamazov, line 201: Please create `LoadBinaryProgram()` that accepts `const std::vector<char>&` and eliminate redundant copying. Eventually we must switch to using `std::vector<char>` for all blobs, but this is not so yet.

Jenkinsfile
```diff
@@ -588,6 +588,17 @@ pipeline {
                         buildHipClangJobAndReboot(setup_flags: MLIR_flags, build_env: extra_log_env, test_flags: ' --verbose ', gpu_arch: "gfx908", mlir_build: "ON")
                     }
                 }
+                stage('Fp32 OpenCL MLIR gfx908') {
+                    when {
+                        beforeAgent true
+                        expression { params.TARGET_GFX908 && params.DATATYPE_FP32 }
```
atamazov, line 594: This change is valid, but looks inconsistent. Please also merge from the recent develop.

src/conv/invokers/mlir_impl_gemm.cpp
jerryyin: Done in 70f5570

Jenkinsfile
atamazov: > Please merge from the recent develop.

I will do that for you.

atamazov (approved):
LGTM

junliume (approved):

PR #1208, opened 2021-10-04
[HOTFIX][WORKAROUND] #1207 ported to develop: W/A for issue 1206. Disable ConvHipImplicitGemmBwdDataV4R1Xdlops by default

atamazov:
This is W/A for #1206: #1207 ported for develop to unblock MIOpen promotion pipeline.

PR #1209, opened 2021-10-05
[Tuning] Remove Solver ConvHipImplicitGemmBwdDataV4R1Xdlops FP32

cderb:
Removes FP32 ConvHipImplicitGemmBwdDataV4R1Xdlops from find_db replacing with second best Backward Data Implicit GEMM solver.

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
@@ -30,10 +30,6 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <cstddef>
 
-/// Disable ConvHipImplicitGemmBwdDataV4R1Xdlops by default.
-/// \ref https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1206.
-#define WORKAROUND_ISSUE_1206 1
-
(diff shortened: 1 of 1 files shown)


junliume:
@cderb Thanks! Let me try to target 'Fp16 Hip All Install gfx908" stage and see if it passed there.

junliume:
> @cderb Thanks! Let me try to target 'Fp16 Hip All Install gfx908" stage and see if it passed there.

Platform-specific tests all passed

src/solver/conv_hip_implicit_gemm_bwd_v4r1_xdlops.cpp
```diff
@@ -30,10 +30,6 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <cstddef>
 
-/// Disable ConvHipImplicitGemmBwdDataV4R1Xdlops by default.
```
atamazov, line 33: This W/A must be not removed but narrowed to FP32 in this PR. Otherwise the next tuning round will bring the problem back.

And I am not sure that ConvHipImplicitGemmBwdDataV4R1Xdlops is good for FP16.

`test_regression_half_mi100` has failed because the failing test case is intended for ConvHipImplicitGemmBwdDataV4R1Xdlops specifically. This test should have been disabled in #1207 (and in #1208) under the `WORKAROUND_ISSUE_1206`

atamazov (commented):
Let's avoid rush ;)

Issue #1187, opened 2021-09-27
[TESTS][MI200] `test_conv_for_implicit_gemm` (and possibly more) to be enabled


atamazov:
xDLOPs solvers are enabled for gfx90a, so these must be tested. `test_conv_for_implicit_gemm` is one of the tests that validate xDLOPs kernels.

atamazov:
> could you take a look at why test_conv_for_implicit_gemm is not enabled on gfx90a please?

GPU-specific regression tests (like `test_regression_half_mi100`) should not be enabled by definition.

atamazov:
#1242 enables many tests for gfx90a but not test_conv_for_implicit_gemm yet, due to FP32 precision issues on gfx90a. WORKAROUND_ISSUE_1187 is introduced. What needs to be done soon:
- Check if the issue still persists
- Identify the reason of the failures,
- Open a new, more focused, ticket.
- Replace this W/A with a new, narrower one.
- Close this.

atamazov:
@junliume Can you please suggest which PR removes WORKAROUND_ISSUE_1187 (I do not see it)? If W/A is not removed, then test_conv_for_implicit_gemm is not enabled for MI200 yet. I highly recommend re-opening the ticket and following the plan shown at https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1187#issuecomment-953830894.

junliume:
Let's reopen it for now, now with more careful look we should keep it :)

atamazov:
@junliume @JehandadKhan The first step is checking if `test_conv_for_implicit_gemm` still fails from time to time on MI200. Unfortunately I do not have access to MI200 and therefore unable to follow the proposed plan. Can we assign some engineer? This issue us about FP32 correctness on MI200 so I guess this is important.

ppanchad-amd:
@atamazov Is this resolved in latest ROCm 6.0.2 (HIP 6.0.32831)? Thanks!

JonathanLichtnerAMD:
Closing since MIOpen is being migrated to [ROCm/rocm-libraries](https://github.com/ROCm/rocm-libraries).  Please re-open there if you believe this is still an issue.

PR #1225, opened 2021-10-15
[CI][Tests] enable test_conv_for_implicit_gemm for gfx90a

junliume:
Part of #1187

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -716,7 +716,7 @@ if(WORKAROUND_ISSUE_936 AND MIOPEN_TEST_HALF)
     LIST(APPEND IMPLICITGEMM_ARGS --disable-forward --disable-backward-data)
     #Afther fix need to remove '| grep -v "cannot be executed due to incorrect params"'
 endif()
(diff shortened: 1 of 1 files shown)


atamazov:
> @atamazov @shurale-nkn need some help here: even with the change the CI shows gfx90a does not run the enabled test. What might be missing? Thanks!

See https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1110/files#r730308137

test/CMakeLists.txt
```diff
@@ -716,7 +716,7 @@ if(WORKAROUND_ISSUE_936 AND MIOPEN_TEST_HALF)
     LIST(APPEND IMPLICITGEMM_ARGS --disable-forward --disable-backward-data)
     #Afther fix need to remove '| grep -v "cannot be executed due to incorrect params"'
 endif()
-add_custom_test(test_conv_for_implicit_gemm SKIP_UNLESS_ALL BF16_ENABLED HALF_ENABLED GFX1030_ENABLED
+add_custom_test(test_conv_for_implicit_gemm SKIP_UNLESS_ALL BF16_ENABLED HALF_ENABLED GFX90A_ENABLED GFX1030_ENABLED
```
atamazov, line 719: AFAIU this won't pass anyway, because non-xdlops solvers are not enabled for gfx90a. That is why this test is not enabled for MI100.

PR #1228, opened 2021-10-18
[MLIR] Making MIOpen build and test MLIR by default

jerryyin:
This PR implements https://github.com/ROCmSoftwarePlatform/llvm-project-private/issues/57. I made the following changes:
 - In Jenkinsfile
   - Toggled `mlir_build` to be default on to pull in mlir dependencies
   - Toggled `-DMIOPEN_USE_MLIR` to be default on to build with MLIR support
 - In test/CmakeLists.txt
   - Toggled `MIOPEN_USE_MLIR` to be default on if not specified
 - Bumped mlir commit to the latest

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,13 +323,13 @@ endif()
 set_var_to_condition(MIOPEN_USE_COMGR_DEFAULT MIOPEN_EMBED_BUILD)
 option(MIOPEN_USE_COMGR "Use comgr to build kernels instead of offline tools" ${MIOPEN_USE_COMGR_DEFAULT})
 
-set_var_to_condition(MIOPEN_USE_MLIR_DEFAULT (NOT ${MIOPEN_EMBED_BUILD}))
+set_var_to_condition(MIOPEN_USE_MLIR_DEFAULT NOT (NOT ${BUILD_SHARED_LIBS} AND ${MIOPEN_USE_COMGR}))
 option(MIOPEN_USE_MLIR "Use MLIR compilation backend" ${MIOPEN_USE_MLIR_DEFAULT})
 
 if(MIOPEN_USE_MLIR)
     # REQUIRED is not supported before cmake 3.18
     find_library(LIBMLIRMIOPEN MLIRMIOpen REQUIRED)
-    if(MIOPEN_EMBED_BUILD)
+    if(NOT ${BUILD_SHARED_LIBS} AND ${MIOPEN_USE_COMGR})
         message(FATAL_ERROR "Potential symbol conflict between mlir and comgr in static build")
     endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ else()
 endif()
 set(MIOPEN_BINCACHE_PATH "" CACHE STRING "URL or path containing binary cache files to embed")
 option(MIOPEN_EMBED_BUILD "Build with the set of embed flags." Off)
-option(MIOPEN_USE_MLIR "Use MLIR -- EXPERIMENTAL" Off)
+option(MIOPEN_USE_MLIR "Use MLIR compilation backend" NOT ${MIOPEN_EMBED_BUILD})
 option(MIOPEN_DISABLE_USERDB "Disable user database access" ${MIOPEN_EMBED_BUILD})
 
 
@@ -328,7 +328,7 @@ if(MIOPEN_USE_MLIR)
     # REQUIRED is not supported before cmake 3.18
     find_library(LIBMLIRMIOPEN MLIRMIOpen REQUIRED)
     if(NOT BUILD_SHARED_LIBS AND MIOPEN_USE_COMGR)
-        message(FATAL_ERROR "Potential symbol conflict between mlir and comgr in static build")
+        message(FATAL_ERROR "Potential symbol conflict between mlir and comgr in static build, please configure with -DMIOPEN_USE_MLIR=Off")
     endif()
 
     if(NOT LIBMLIRMIOPEN)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,6 @@ else()
 endif()
 set(MIOPEN_BINCACHE_PATH "" CACHE STRING "URL or path containing binary cache files to embed")
 option(MIOPEN_EMBED_BUILD "Build with the set of embed flags." Off)
-option(MIOPEN_USE_MLIR "Use MLIR compilation backend" NOT ${MIOPEN_EMBED_BUILD})
 option(MIOPEN_DISABLE_USERDB "Disable user database access" ${MIOPEN_EMBED_BUILD})
 
 
@@ -324,11 +323,14 @@ endif()
 set_var_to_condition(MIOPEN_USE_COMGR_DEFAULT MIOPEN_EMBED_BUILD)
 option(MIOPEN_USE_COMGR "Use comgr to build kernels instead of offline tools" ${MIOPEN_USE_COMGR_DEFAULT})
 
+set_var_to_condition(MIOPEN_USE_MLIR_DEFAULT (NOT ${MIOPEN_EMBED_DB}))
+option(MIOPEN_USE_MLIR "Use MLIR compilation backend" ${MIOPEN_USE_MLIR_DEFAULT})
+
 if(MIOPEN_USE_MLIR)
     # REQUIRED is not supported before cmake 3.18
     find_library(LIBMLIRMIOPEN MLIRMIOpen REQUIRED)
     if(NOT BUILD_SHARED_LIBS AND MIOPEN_USE_COMGR)
-        message(FATAL_ERROR "Potential symbol conflict between mlir and comgr in static build, please configure with -DMIOPEN_USE_MLIR=Off")
+        message(FATAL_ERROR "Potential symbol conflict between mlir and comgr in static build")
     endif()
 
     if(NOT LIBMLIRMIOPEN)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,13 +323,13 @@ endif()
 set_var_to_condition(MIOPEN_USE_COMGR_DEFAULT MIOPEN_EMBED_BUILD)
 option(MIOPEN_USE_COMGR "Use comgr to build kernels instead of offline tools" ${MIOPEN_USE_COMGR_DEFAULT})
 
-set_var_to_condition(MIOPEN_USE_MLIR_DEFAULT (NOT ${MIOPEN_EMBED_DB}))
+set_var_to_condition(MIOPEN_USE_MLIR_DEFAULT (NOT ${MIOPEN_EMBED_BUILD}))
 option(MIOPEN_USE_MLIR "Use MLIR compilation backend" ${MIOPEN_USE_MLIR_DEFAULT})
 
 if(MIOPEN_USE_MLIR)
     # REQUIRED is not supported before cmake 3.18
     find_library(LIBMLIRMIOPEN MLIRMIOpen REQUIRED)
-    if(NOT BUILD_SHARED_LIBS AND MIOPEN_USE_COMGR)
+    if(MIOPEN_EMBED_BUILD)
         message(FATAL_ERROR "Potential symbol conflict between mlir and comgr in static build")
     endif()
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -589,16 +589,6 @@ pipeline {
                         buildHipClangJobAndReboot()
                     }
                 }
-                stage('Fp16 Hip gfx908') {
-                    when {
-                        beforeAgent true
-                        expression { params.TARGET_GFX908 && params.DATATYPE_FP16 }
-                    }
-                    agent{ label rocmnode("gfx908") }
-                    steps{
-                        buildHipClangJobAndReboot(setup_flags: Fp16_flags, build_env: extra_log_env, config_targets: Smoke_targets, test_flags: ' --verbose ', gpu_arch: "gfx908")
-                    }
-                }
             }
         }
         stage("Smoke Fp16/Bf16/Int8") {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -255,10 +255,6 @@ pipeline {
             defaultValue: true,
             description: "")
         booleanParam(
-            name: "BUILD_SMOKE_MLIR",
-            defaultValue: true,
-            description: "")
-        booleanParam(
             name: "BUILD_SMOKE_MIOPENTENSILE_LATEST",
             defaultValue: true,
             description: "")
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,8 @@ def cmake_build(Map conf=[:]){
     def debug_flags = "-g -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=undefined " + conf.get("extradebugflags", "")
     def build_envs = "CTEST_PARALLEL_LEVEL=4 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0 " + conf.get("build_env","")
     def prefixpath = conf.get("prefixpath","/usr/local")
-    def setup_args = " -DMIOPEN_GPU_SYNC=Off " + conf.get("setup_flags","")
+    def mlir_args = " -DMIOPEN_USE_MLIR=" + conf.get("mlir_build", "ON")
+    def setup_args = mlir_args + " -DMIOPEN_GPU_SYNC=Off " + conf.get("setup_flags","")
 
     if (prefixpath != "/usr/local"){
         setup_args = setup_args + " -DCMAKE_PREFIX_PATH=${prefixpath} "
@@ -113,7 +114,7 @@ def buildHipClangJob(Map conf=[:]){
 
         def miotensile_version = conf.get("miotensile_version", "default")
         def target_id = conf.get("target_id", "OFF")
-        def mlir_build = conf.get("mlir_build", "OFF")
+        def mlir_build = conf.get("mlir_build", "ON")
         def build_fin = conf.get("build_fin", "OFF")
         def dockerOpts="--device=/dev/kfd --device=/dev/dri --group-add video --group-add render --cap-add=SYS_PTRACE --security-opt seccomp=unconfined"
         if (conf.get("enforce_xnack_on", false)) {
@@ -332,7 +333,6 @@ pipeline {
         Tensile_build_env = "MIOPEN_DEBUG_HIP_KERNELS=0 "
         Tensile_setup = " -DMIOPEN_TEST_MIOTENSILE=ON -DMIOPEN_USE_MIOPENTENSILE=ON -DMIOPEN_USE_ROCBLAS=OFF"
         Smoke_targets = "check doc MIOpenDriver"
-        MLIR_flags    = " -DMIOPEN_USE_MLIR=On"
         COMGR_flags   = " -DMIOPEN_USE_COMGR=On"
     }
     stages{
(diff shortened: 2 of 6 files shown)


jerryyin:
To capture some of the offline discussion:
 - This PR shall be on hold until ROCm 5.0 development complete is done

The proper way to implement (instead of the existing approach), according to the discussion, is:
- Cmake: MIOpen_USE_MLIR turned on by default
- Jenkinsfile: MLIR_flags turned on by default,  mlir_build turned on by default

Jenkinsfile
```diff
@@ -603,7 +603,7 @@ pipeline {
                     }
```
jerryyin, line 603: Needs suggestion: It looks to me that once we make MLIR the default build, the `test_conv_igemm_mlir_*_small` will be exercised by all smoke stages therefore we should get rid of the entire "Smoke MLIR" stage.

atamazov: Correct.

jerryyin: Will get rid of the entire "Smoke MLIR"

min-requirements.txt
```diff
@@ -1,3 +1,4 @@
 sqlite3,https://sqlite.org/2017/sqlite-autoconf-3170000.tar.gz -H sha256:a4e485ad3a16e054765baf6371826b5000beed07e626510896069c0bf013874c -X autotools -DCMAKE_POSITION_INDEPENDENT_CODE=On
 boost@1.72 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build
 half,https://github.com/pfultz2/half/archive/1.12.0.tar.gz -X header -H sha256:0a08660b68abb176ebc2a0cdf8de46e3182a7f46c66443bb80dbfaaec98cf969 --build
+ROCmSoftwarePlatform/llvm-project-mlir@a24618b5694a4e7e1e60cec8a41dffa3e9993ba7 -DBUILD_FAT_LIBMLIRMIOPEN=1
```
pfultz2, line 4: So due to PR #1233 these dependencies have been moved to requirements.txt, so this will need to be moved there as well.

jerryyin: Done, thanks for reminding.

CMakeLists.txt
```diff
@@ -132,7 +132,7 @@ else()
 endif()
 set(MIOPEN_BINCACHE_PATH "" CACHE STRING "URL or path containing binary cache files to embed")
 option(MIOPEN_EMBED_BUILD "Build with the set of embed flags." Off)
-option(MIOPEN_USE_MLIR "Use MLIR -- EXPERIMENTAL" Off)
+option(MIOPEN_USE_MLIR "Use MLIR compilation backend" NOT ${MIOPEN_EMBED_BUILD})
```
atamazov, line 135: Does CMake support usage of expressions for initialization of options?

CMakeLists.txt
```diff
@@ -328,7 +328,7 @@ if(MIOPEN_USE_MLIR)
     # REQUIRED is not supported before cmake 3.18
     find_library(LIBMLIRMIOPEN MLIRMIOpen REQUIRED)
     if(NOT BUILD_SHARED_LIBS AND MIOPEN_USE_COMGR)
-        message(FATAL_ERROR "Potential symbol conflict between mlir and comgr in static build")
+        message(FATAL_ERROR "Potential symbol conflict between mlir and comgr in static build, please configure with -DMIOPEN_USE_MLIR=Off")
```
atamazov, line 331: I would revert this. In some cases the correct recommendation might be switching off COMGR (not MLIR).

Jenkinsfile
```diff
@@ -592,65 +592,14 @@ pipeline {
...
-                        beforeAgent true
-                        expression { (params.TARGET_VEGA20 || params.TARGET_VEGA10) && params.DATATYPE_FP16 }
-                    }
-                    agent{ label rocmnode("vega") }
-                    steps{
-                        buildHipClangJobAndReboot(setup_flags: MLIR_flags + Fp16_flags, build_env: extra_log_env, config_targets: Smoke_targets, test_flags: ' --verbose ', mlir_build: "ON")
-                    }
-                }
-                stage('Fp16 Hip MLIR gfx908') {
+                stage('Fp16 Hip gfx908') {
```
atamazov, line 595: It looks like this PR adds 'Fp16 Hip gfx908' to `stage("Smoke Aux 1")`

test/CMakeLists.txt
```diff
@@ -652,12 +654,11 @@ set(IMPLICITGEMM_MLIR_ARGS_B ${IMPLICITGEMM_ARGS} --verbose --disable-forward --
...
-    COMMAND ${IMPLICITGEMM_MLIR_ENV_B} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_B} --input 256 256  56 56 --weights 256  64   1 1 --pads_strides_dilations 0 0 1 1 1 1 --group-count 4
-    COMMAND ${IMPLICITGEMM_MLIR_ENV_W} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_W} --input 128 64   56 56 --weights 64   64   1 1 --pads_strides_dilations 0 0 1 1 1 1
-    )
+    COMMAND ${IMPLICITGEMM_MLIR_ENV_F} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 64 128 14 14 --weights 128 128 1 1 --pads_strides_dilations 0 0 2 2 1 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
+    COMMAND ${IMPLICITGEMM_MLIR_ENV_B} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_B} --input 64 256 28 28 --weights 64  64  1 1 --pads_strides_dilations 0 0 1 1 1 1 --group-count 4
+    COMMAND ${IMPLICITGEMM_MLIR_ENV_W} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_W} --input 64 64  28 28 --weights 64  64  1 1 --pads_strides_dilations 0 0 1 1 1 1
```
atamazov, line 659: Is this for testing speed?

test/CMakeLists.txt
```diff
@@ -692,12 +693,11 @@ set(IMPLICITGEMM_MLIR_ENV_B_XDLOPS ${IMPLICITGEMM_MLIR_ENV_BASE} MIOPEN_DEBUG_FI
...
-    COMMAND ${IMPLICITGEMM_MLIR_ENV_B_XDLOPS} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_B} --input 256 256  56 56 --weights 256  64   1 1 --pads_strides_dilations 0 0 1 1 1 1 --group-count 4
-    COMMAND ${IMPLICITGEMM_MLIR_ENV_W_XDLOPS} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_W} --input 128 64   56 56 --weights 64   64   1 1 --pads_strides_dilations 0 0 1 1 1 1
-    )
+    COMMAND ${IMPLICITGEMM_MLIR_ENV_F_XDLOPS} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 64 128 14 14 --weights 128 128 1 1 --pads_strides_dilations 0 0 2 2 1 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
+    COMMAND ${IMPLICITGEMM_MLIR_ENV_B_XDLOPS} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_B} --input 64 256 28 28 --weights 64  64  1 1 --pads_strides_dilations 0 0 1 1 1 1 --group-count 4
+    COMMAND ${IMPLICITGEMM_MLIR_ENV_W_XDLOPS} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_W} --input 64 64  28 28 --weights 64  64  1 1 --pads_strides_dilations 0 0 1 1 1 1
```
atamazov, line 698: Same Q

atamazov (changes requested):

Jenkinsfile
jerryyin: Before you raise this question, I was just looking at the `Smoke Fp32` and `Smoke Aux1` stages before the original MLIR stage. Now I just realized that there are more specific fp16 stages in the `Smoke Fp16/Bf16/Int8`.

Based on this observation I will remove 'Fp16 Hip gfx908`

test/CMakeLists.txt
jerryyin: Yes, the original config alone would cost at least 20 minutes per config in CPU. This is especially bad for a debug build and timeout after 5 hours. Therefore just shrinking the problem size of speed.

test/CMakeLists.txt
jerryyin: Correct

CMakeLists.txt
jerryyin: Sounds good

CMakeLists.txt
jerryyin: Honestly speaking I don't know. I was assuming that since the line below works, that this line should also work. But agree on a second look this isn't supported, as from [here](https://cmake.org/cmake/help/latest/command/option.html).

Instead, which do you think is better:
 - Make MIOPEN_USE_MLIR default to be ON
 - Initialize MIOPEN_USE_MLIR not here, but in between line 140 - 148 and turn it on conditionally based on whether it is embed build

atamazov: Please look how I compute and use the default for MIOPEN_USE_COMGR: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1253/files#r745142441

jerryyin: That's indeed a better way to handle it, thanks for informing!

Jenkinsfile
```diff
@@ -22,7 +22,8 @@ def cmake_build(Map conf=[:]){
     def debug_flags = "-g -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=undefined " + conf.get("extradebugflags", "")
     def build_envs = "CTEST_PARALLEL_LEVEL=4 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0 " + conf.get("build_env","")
     def prefixpath = conf.get("prefixpath","/usr/local")
-    def setup_args = " -DMIOPEN_GPU_SYNC=Off " + conf.get("setup_flags","")
+    def mlir_args = " -DMIOPEN_USE_MLIR=" + conf.get("mlir_build", "ON")
```
atamazov, line 25: Hmm... Is it really necessary?

Jenkinsfile
```diff
@@ -115,7 +116,7 @@ def buildHipClangJob(Map conf=[:]){
 
         def miotensile_version = conf.get("miotensile_version", "default")
         def target_id = conf.get("target_id", "OFF")
-        def mlir_build = conf.get("mlir_build", "OFF")
+        def mlir_build = conf.get("mlir_build", "ON")
```
atamazov, line 119: We keep this in order to be able to switch off MLIR builds if/when necessary?

test/CMakeLists.txt
```diff
@@ -692,12 +693,11 @@ set(IMPLICITGEMM_MLIR_ENV_B_XDLOPS ${IMPLICITGEMM_MLIR_ENV_BASE} MIOPEN_DEBUG_FI
...
+)
 
-# Note: This test is unused until MLIR is switched ON by default
```
atamazov, line 700: Good.

atamazov (changes requested):
Please remove `BUILD_SMOKE_MLIR` from Jenkinsfile

Jenkinsfile
jerryyin: Since `MIOPEN_USE_MLIR` becomes the default therefore no, it isn't necessary. But at the same time the `mlir_build` option in general gives a individual stage the flexibility to turn of MLIR build if necessary. This might be useful if someone wants to use the CI to regress the non-MLIR scenario. 

Talking about which, do you think it is necessary to add a couple of smoke test that build without MLIR?

Either way I'm neutral to this... And can remove/keep based on your suggestion.

atamazov: Let's keep this.

atamazov: > do you think it is necessary to add a couple of smoke test that build without MLIR?

No. Thanks.

atamazov (approved):
LGTM!

atamazov:
@jerryyin Do you know why test_conv_igemm_cmlir_small takes ~1700 (!) seconds to finish in the "Fp32 OpenCL Debug + Codecov" stage? It is much faster in all other stages. Maybe we should disable it for this specific stage.

jerryyin:
> @jerryyin Do you know why `test_conv_igemm_mlir_small` takes ~1700 (!) seconds to finish in the "Fp32 OpenCL Debug + Codecov" stage? It is much faster in all other stages. 

I don't know, thanks for reminding. 

I believe `test_conv_igemm_mlir_small` and `test_conv_igemm_mlir_small_xdlops` are literally testing the same thing. Since kernel generation and execution are very fast, the last factor seem to point at test/CPU validation.

My theory is that the node used to test `Fp32 OpenCL Debug` has a particular weak CPU (Compared with `Fp32 Hip Debug` or `Fp32 Hip Debug gfx908`, therefore causing the time difference. Is there a quick way to prove this theory?

> Maybe we should disable it for this specific stage.

Definitely, I can turn off mlir build specifically for this stage.

atamazov:
> Definitely, I can turn off mlir build specifically for this stage.

I meant only disabling this test for "Debug + Codecov" builds in CMakeLists.txt. Thanks.

atamazov:
@jerryyin 
> force-pushed the mlir-default-env branch from 701988e to 7de26bc 14 hours ago 

This breaks URLs of existing commits and may also break URLs to related comments. I highly recommend avoiding this.

test/CMakeLists.txt
```diff
@@ -653,13 +653,16 @@ set(IMPLICITGEMM_MLIR_ARGS_F ${IMPLICITGEMM_ARGS} --verbose --disable-backward-d
 set(IMPLICITGEMM_MLIR_ARGS_B ${IMPLICITGEMM_ARGS} --verbose --disable-forward --disable-backward-weights)
 set(IMPLICITGEMM_MLIR_ARGS_W ${IMPLICITGEMM_ARGS} --verbose --disable-forward --disable-backward-data)
 
-add_custom_test(test_conv_igemm_mlir_small HALF_ENABLED MLIR_ENABLED GFX908_DISABLED GFX90A_DISABLED
-    COMMAND ${IMPLICITGEMM_MLIR_ENV_F} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_F} --input 64 128 14 14 --weights 128 128 1 1 --pads_strides_dilations 0 0 2 2 1 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC
-    COMMAND ${IMPLICITGEMM_MLIR_ENV_B} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_B} --input 64 256 28 28 --weights 64  64  1 1 --pads_strides_dilations 0 0 1 1 1 1 --group-count 4
-    COMMAND ${IMPLICITGEMM_MLIR_ENV_W} $<TARGET_FILE:test_conv2d> ${IMPLICITGEMM_MLIR_ARGS_W} --input 64 64  28 28 --weights 64  64  1 1 --pads_strides_dilations 0 0 1 1 1 1
-)
+# Note: OpenCL Debug + Codecov test stage taking longer time than a smoke test should therefore disabling that scenario
+if ((NOT CMAKE_BUILD_TYPE MATCHES Debug) OR (NOT ${CODECOV_TEST}))
```
atamazov, line 657: This doesn't work (please see CI logs). You may need to add `"` around `Debug`. And/or case-insensitive test is required here (this is nontrivial ;))

atamazov (changes requested):

atamazov:
@jerryyin Why do we have both xDLOPs and non-xDLOPs solvers enabled for gfx908? I thought that xDLOPs is always faster than its non-xDLOPs counterpart.

test/CMakeLists.txt
jerryyin: Ooops you are right. I was testing locally with `-DCMAKE_BUILD_TYPE=Debug` and didn't realize that CI go with the lower cased `debug`.

jerryyin:
> I thought that xDLOPs is always faster than its non-xDLOPs counterpart.

That's 99.99...% right but not guaranteed. The xdlops algorithm use our special instruction whereas the non-xdlops does the generic mfa implementation.

> Why do we have both xDLOPs and non-xDLOPs solvers enabled for gfx908? 

This is for test coverage: I think we still allow non-xdlops solver to be selected for gfx908 therefore it should get regressed in CI as well.

On the other hand, it might be a fair optimization to disallow fma based algorithm from running in gfx908. If we decide to do that, then I can revert this change.

test/CMakeLists.txt
```diff
@@ -654,7 +654,7 @@ set(IMPLICITGEMM_MLIR_ARGS_B ${IMPLICITGEMM_ARGS} --verbose --disable-forward --
 set(IMPLICITGEMM_MLIR_ARGS_W ${IMPLICITGEMM_ARGS} --verbose --disable-forward --disable-backward-data)
 
 # Note: OpenCL Debug + Codecov test stage taking longer time than a smoke test should therefore disabling that scenario
-if ((NOT CMAKE_BUILD_TYPE MATCHES Debug) OR (NOT ${CODECOV_TEST}))
+if ((NOT uppercase CMAKE_BUILD_TYPE MATCHES "DEBUG") OR (NOT ${CODECOV_TEST}))
```
atamazov, line 657: I am unable to find this syntax in CMake docs. You can try `string(TOUPPER ${VARNAME} VARNAME)` before if.

atamazov:
@jerryyin 
> On the other hand, it might be a fair optimization to disallow fma based algorithm from running in gfx908. If we decide to do that, then I can revert this change.

Yes, I recommend following the design of current HIP iGemm solvers. Otherwise we'll waste compilation time (in FIND_MODE=normal, including tuning process that uses it) or the non-xDLOPs Solver will not be used anyway (because of WORKAROUND_SWDEV_227826).

But let's be consistent - either disable non-xDLOPs Solvers for fdx908/90a and disable the non-xdlops test right in this PR, at once, or do that in some separate PR.

test/CMakeLists.txt
jerryyin: Done in f3dd6f4

jerryyin:
But let's be consistent - either disable non-xDLOPs Solvers for fdx908/90a and disable the non-xdlops test right in this PR, at once, or do that in some separate PR.

Sounds good, let me get that done in another PR.

CMakeLists.txt
```diff
@@ -324,10 +323,13 @@ endif()
...
 option(MIOPEN_USE_COMGR "Use comgr to build kernels instead of offline tools" ${MIOPEN_USE_COMGR_DEFAULT})
 
+set_var_to_condition(MIOPEN_USE_MLIR_DEFAULT (NOT ${MIOPEN_EMBED_BUILD}))
+option(MIOPEN_USE_MLIR "Use MLIR compilation backend" ${MIOPEN_USE_MLIR_DEFAULT})
+
 if(MIOPEN_USE_MLIR)
     # REQUIRED is not supported before cmake 3.18
     find_library(LIBMLIRMIOPEN MLIRMIOpen REQUIRED)
-    if(NOT BUILD_SHARED_LIBS AND MIOPEN_USE_COMGR)
+    if(MIOPEN_EMBED_BUILD)
```
atamazov, line 332: Please revert this. Some of MIOpen users set BUILD_SHARED_LIBS=off and MIOPEN_USE_COMGR=on without explicit use of MIOPEN_EMBED_BUILD.

CMakeLists.txt
```diff
@@ -324,10 +323,13 @@ endif()
 set_var_to_condition(MIOPEN_USE_COMGR_DEFAULT MIOPEN_EMBED_BUILD)
 option(MIOPEN_USE_COMGR "Use comgr to build kernels instead of offline tools" ${MIOPEN_USE_COMGR_DEFAULT})
 
+set_var_to_condition(MIOPEN_USE_MLIR_DEFAULT (NOT ${MIOPEN_EMBED_BUILD}))
```
atamazov, line 326:
```suggestion
set_var_to_condition(MIOPEN_USE_MLIR_DEFAULT NOT  (NOT ${BUILD_SHARED_LIBS} AND ${MIOPEN_USE_COMGR}))
```

atamazov (changes requested):
Almost good!

CMakeLists.txt
jerryyin: I will revert this but just want to give you a heads up that it will block more scenarios than we actually should. When MIOpen is built as a static library but use COMGR as a dynamic library, things still work for MLIR backend. I believe `Smoke Aux 1::Fp32 Hip Static` stage is tracking this scenario.

Strictly speaking MLIR/COMGR combination will be an issue when and only when MIOpen is built as a static library to include all its dependencies in one binary. It looks to me that the client really should be use a specific cmake configuration (like `MIOPEN_EMBED_BUILD`) for this scenario.

So one day when MIOpen turn on COMGR by default, that Fp32 Hip Static stage will populate an error because it hit on this condition (whereas it can actually run through)

jerryyin: Done in 6cad119

CMakeLists.txt
jerryyin: Done in 6cad119

CMakeLists.txt
atamazov: Let's ask @JehandadKhan -- I am not sure which build configurations we are supporting right now etc.

atamazov:
@JehandadKhan please share with us your opinion about 6cad119.

JehandadKhan (approved):

CMakeLists.txt
JehandadKhan: I agree with the current implementation, where we only block MLIR enablement when MIOpen is built fully statically. Embeded builds are ok if they link against MIOpen dynamically and therefore dynamically link against MLIR as well.

jerryyin:
@atamazov @junliume Could you help to proceed with the PR now that all review comments are addressed and CI passed? The deadline for Nov demo is tomorrow and I want to make sure I'm on time with this last feature :-)

atamazov (approved):
LGTM!

PR #1229, opened 2021-10-18
[Tuning] NHWC update

cderb:


cderb:
> CI has passed. @cderb : would performance test be required?

It might be helpful. 
In theory all of the configs are new here. In observation we are seeing over half result in overwritten fdb entries. Though regressions seem very rare. There are a significant number of new fdb entries as well. Curiously enough, all entries in fdb and sqlite are still labled as NCHW layout. Driver commands used are from attached files at https://ontrack-internal.amd.com/browse/SWDEV-301397 . Not sure if MIOpen labels these differently yet or if there is some missing argument.

junliume:
> all entries in fdb and sqlite are still labled as NCHW layout.

@atamazov  and @JehandadKhan
Would the following observation match our expectation?
**all entries in fdb and sqlite are still labled as NCHW layout.**

PR #1233, opened 2021-10-20
[Enhancement] Simplify dependencies to build

pfultz2:
This removes the old hcc dependencies in our requirements.txt files. Now min-requirements.txt and requirements.txt are the same(we could hopefully remove this file in the future).

I also moved the cmake installation to mlir-requirements.txt, since its purpose is for mlir. I am not sure why it installs to /opt/rocm instead of `$PREFIX` but I didnt change this just in case mlir has a broken cmake installation(we should be able to install mlir to any `$PREFIX` using cget and it should work without needing private env variables(ie ROCM_PATH, HIP_PATH, etc) but I havent tested this).

This also adds an rbuild.ini file which has different sessions to build dependencies and settings. It will set the clang compiler by default(for rocm releases), so users can build by just doing `rbuild build -d deps`.

There is also a session for `fin` so it can install the additional dependencies needed for fin. You can use this session with `rbuild prepare -s fin -d deps`.

Finally, it adds a `develop` session which installs the dev-requirements.txt dependencies which include requirements.txt, cppcheck, and rocm-cmake. So now you can setup MIOpen for development with `rbuild develop -d deps -B build`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,7 +106,7 @@ RUN pip3 install -r /doc-requirements.txt
 
 # Use parallel job to accelerate tensile build
 # Workaround for Tensile with TargetID feature
-RUN if [ "$USE_TARGETID" = "ON" ] ; then export HIPCC_LINK_FLAGS_APPEND='-O3 -parallel-jobs=4' && export HIPCC_COMPILE_FLAGS_APPEND='-O3 -Wno-format-nonliteral -parallel-jobs=4' && rm /usr/bin/hipcc; fi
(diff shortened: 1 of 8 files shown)


pfultz2:
> we are having some problems with the changes in CI

Let me see. It works building locally, so I am not sure the difference.

atamazov (approved):
LGTM

atamazov:
Reviewers, please go ahead.

junliume:
> Reviewers, please go ahead.

Will ping reviewers when CI has passed :)

jerryyin (approved):

atamazov:
@junliume For this PR, Full tests are not required.

CI job `#7` can be terminated after `Smoke MLIR` stage passed. CI job `#8` and `#9` will cover remaining smoke tests: `#8` is testing `Smoke MIOpenTensile` + `Packaging`, CI job `#9` is testing `Smoke FP16/BF16/Int8`. These three jobs are running in parallel.

atamazov:
CI jobs `#7` and `#9` passed as expected.

junliume (approved):
As commented, CI `#8` has passed too

PR #1234, opened 2021-10-21
[NFC] Bnorm inference solver

DrizztDoUrden:
This is the last batch norm solver.

Both my and @muralinr's benches have passed.


muralinr (commented):
Is CI testing passed?

junliume (approved):

PR #1237, opened 2021-10-22
Enable HIPRTC support as default from ROCm 5.0

atamazov:
- HIPRTC support is added and enabled by default starting from ROCm 5.0
  - This will automatically enable HIPRTC testing after CI upgrade
  - :warning: HIP version must be fixed in Mainline in order to enable HIPRTC testing by QA
- Added `MIOPEN_DEBUG_USE_HIPRTC` env var, which can be used to fall back to COMGR.
- Workarounds:
  - Added W/A for SWDEV-308073 (WORKAROUND_ISSUE_HIPRTC_TRUE_TYPE)
  - Added W/A for SWDEV-307838 (WORKAROUND_ISSUE_HIPRTC_HIPRTC_HEADER_H)


atamazov:
### TODO: Fix the test_test_gpu_nchw_nhwc_transpose issues with (gfx906, Mainline 9470 build). 

Example failure:
```
diff at:0, gpu:0, cpu:65525
[nchw2nhwc, b16 ] n:1, c:29, h:1, w:21, valid:0
CMake Error at test_test_gpu_nchw_nhwc_transpose.cmake:7 (message):
```
The failure happens randomly, approx 1 time out of each 3 runs. The issue happens with HIPRTC only; no problems with offline compilation, no problems with COMGR builds.

Log of 100 runs: [test_gpu_nchw_nhwc_transpose_FAILURES_HIPRTC.log.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/8004312/test_gpu_nchw_nhwc_transpose_FAILURES_HIPRTC.log.txt)

atamazov:
> TODO: Fix the test_test_gpu_nchw_nhwc_transpose issues with (gfx906, Mainline 9470 build).

The root reason is outdated base driver (from 4.0). Runtime reports that device is sramecc- (but it is actually sramecc+). We have W/A for this in the library, but HIPRTC does device detection in its own. Therefore HIPRTC builds ISA as if device has SRAMECC OFF. As a result, I see bugs in FP16/BF16 code.

Works fine with ROCm 5.0.

junliume:
> > TODO: Fix the test_test_gpu_nchw_nhwc_transpose issues with (gfx906, Mainline 9470 build).
> 
> The root reason is outdated base driver (from 4.0). Runtime reports that device is sramecc- (but it is actually sramecc+). We have W/A for this in the library, but HIPRTC does device detection in its own. Therefore HIPRTC builds ISA as if device has SRAMECC OFF. As a result, I see bugs in FP16/BF16 code.
> 
> Works fine with ROCm 5.0.

I have a gfx1030 installed ROCM 5.0 from scratch, and if you need some repeated stress tests on it please let me know.

src/composable_kernel/composable_kernel/include/utility/type.hpp
```diff
@@ -4,6 +4,84 @@
...
 
+#ifdef __HIPCC_RTC__
+#ifdef WORKAROUND_ISSUE_HIPRTC_TRUE_TYPE
+/// We need <type_traits> for std::remove_reference and std::remove_cv.
+/// But <type_traits> also defines std::true_type, per Standard.
+/// However the latter definition conflicts with
+/// /opt/rocm/include/hip/amd_detail/amd_hip_vector_types.h,
+/// which defines std::true_type as well (which is wrong).
+
+namespace std {
```
JehandadKhan, line 15: Can we move these to a common file and include that file everywhere instead ?

JehandadKhan (changes requested):

src/composable_kernel/composable_kernel/include/utility/type.hpp
junliume: Would this WA potentially causing the numerical changes in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1237#issuecomment-1040692300 ?

qianfengz:
> @atamazov @JehandadKhan @asroy @zjing14 @qianfengz could you take a look at CK related changes? The composable kernel-related changes in this PR might have affected accuracy:
> 
> ```
> # ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0
> ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0 
> FAILED: 0.255031
> Iteration: 0
> verify_reduce_no_indices failed
> Input Tensor 64, 3, 280, 81
> Max diff: 2.27564
> Mismatch at 0: 1.00436 != 2.00873
> ```

Very strange!  I could not reproduce the issue on MI100 and MI50 using hiprtc branch.   From your test, the reduced result value get on Host is just half that of the GPU.

junliume:
> > @atamazov @JehandadKhan @asroy @zjing14 @qianfengz could you take a look at CK related changes? The composable kernel-related changes in this PR might have affected accuracy:
> > ```
> > # ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0
> > ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0 
> > FAILED: 0.255031
> > Iteration: 0
> > verify_reduce_no_indices failed
> > Input Tensor 64, 3, 280, 81
> > Max diff: 2.27564
> > Mismatch at 0: 1.00436 != 2.00873
> > ```
> 
> Very strange! I could not reproduce the issue on MI100 and MI50 using hiprtc branch. From your test, the reduced result value get on Host is just half that of the GPU.

This might not be hiprtc related but instead ROCm 5.0 related. Removing the blocker for this PR for now.

src/composable_kernel/composable_kernel/include/utility/type.hpp
junliume: Let's fix this issue in follow up PRs

junliume (approved):

qianfengz:
> > > @atamazov @JehandadKhan @asroy @zjing14 @qianfengz could you take a look at CK related changes? The composable kernel-related changes in this PR might have affected accuracy:
> > > ```
> > > # ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0
> > > ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0 
> > > FAILED: 0.255031
> > > Iteration: 0
> > > verify_reduce_no_indices failed
> > > Input Tensor 64, 3, 280, 81
> > > Max diff: 2.27564
> > > Mismatch at 0: 1.00436 != 2.00873
> > > ```
> > 
> > 
> > Very strange! I could not reproduce the issue on MI100 and MI50 using hiprtc branch. From your test, the reduced result value get on Host is just half that of the GPU.
> 
> This might not be hiprtc related but instead ROCm 5.0 related. Removing the blocker for this PR for now.

This could be compiler issue.  I found that the `warpSize` used by my `DirectWarpSize` reduction kernels is 64, while the value of `handle.GetWavefrontWidth()` is 32.   My understand is that `warpSize` is provided as a constant by the compiler, while `handle.GetWavefrontWidth()` gets the warp size from the HIP Runtime device properties.  So the constant value maintained by the compiler under ROCM 5.0 is problematic on Navi.   The issue can be worked-around by passing the value of `handle.GetWavefrontWidth()` to the kernel and let the `DirectWarpWise` kernel to use it instead of the `warpSize`

src/comgr.cpp
```diff
@@ -989,4 +998,290 @@ void BuildAsm(const std::string& name,
...
+              const miopen::TargetProperties& target,
+              std::vector<char>& binary)
+{
+    PrintVersion();
+    try
+    {
+        auto opts =
+            miopen::SplitSpaceSeparated(options, miopen::comgr::compiler::lc::GetOptionsNoSplit());
+        compiler::lc::RemoveOptionsUnwanted(opts);
+        opts.push_back("-DWORKAROUND_ISSUE_HIPRTC_TRUE_TYPE"); // Workaround for SWDEV-308073
```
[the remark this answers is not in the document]
atamazov: @JehandadKhan 
> so that we something needs to be fixed, it only needs to be fixed in one place.

I do not understand. Can you please clarify the use case?

atamazov:
We should stop the chaos from spreading and start adding comments to the relevant tickets.

junliume:
> We should stop the chaos from spreading and start adding comments to the relevant tickets.

Yes, currently two priorities for this week: (1) what might have caused workspace diffs in last tuning updates; (2) warpSize inconsistent between different HIP kernel building methods, e.g. hip-Clang and hipRTC.

Each is tracked by an issue in blocking urgency. #1429 and #1431 The first one is actively been resolved. The second has a workaround for now (not sure if there will be other issues though).

junliume:
@atamazov resnet is getting gradient overflow with this PR enabling hipRTC as default. Is it safe to revert it? Thanks!

atamazov:
@junliume Just change MIOPEN_USE_HIPRTC_DEFAULT (line 226 in ./CMakeLists.txt) to something like `...GREATER 900000000))` as a workaround.

atamazov:
Then you'll be able to use `-DMIOPEN_USE_HIPRTC=On/Off` for experiments.

junliume:
> @junliume Just change MIOPEN_USE_HIPRTC_DEFAULT (line 226 in ./CMakeLists.txt) to something like `...GREATER 900000000))` as a workaround.

Thank you! @atamazov

Issue #866, opened 2021-04-19
Solver framework


atamazov:
## 1. Intent

Problem: Variety of convolution kernels 
- Different formats (OCL, asm, binary)
- Capable to do different (but often intersecting) subsets of convolution configs
- Some may benefit from auto-tuning, some not.
- Each kernel may require its own set of #defines for build
- There are cases when two (or more kernels) needs to be run to do the job
- Kernels need different parameters during invocation

Experience shows that straightforward attempts to support such a set of kernels result in host code which is large, fragile, difficult to develop and maintain. You may see leftovers of this in convolutionocl.cpp.

Provide abstractions which able to represent in the single place all the information required to
- Select kernel(s) which can do the required job
- Build these kernel(s)
- Run kernels (not yet implemented) 
- Auto-tune kernel(s)

Such abstractions allow working with all convolutions in unified manner. Currently, there are:
- ProblemDescriptions
- Solvers
  - Non-searchable
  - Searchable, which implement auto-tune internally (legacy OpenCL)
- Searchable, which employ Generic Search
  - Solutions
  - PerformanceConfigs (paired with searchable Solvers)

atamazov:
## 2. Problem Description
This is an input for the Solver.
- This is an object which represents a set of parameters that shall unambiguously describe the specific library primitive. Example for convolutions:
```C++
struct ProblemDescription
{
    int n_inputs         = 0;
    int in_height        = 0;
    int in_width         = 0;
    int kernel_size1     = 0;
    int kernel_size0     = 0;
    int n_outputs        = 0;
...
    struct Direction direction; // has members like IsBackwardData() etc
...
}
```
- If perf-db is used, the object must provide also "void Serialize(std::ostream& stream) const;". This is used to generate perf-db keys.
  - Each unique problem shall have unique db-key.
- May contain other helper members in order to ease implementation or Solvers, for example:
```C++
struct ProblemDescription
{...
    int GetBackwardPad0() const { return kernel_size0 - pad0 - 1; }
    int GetBackwardPad1() const { return kernel_size1 - pad1 - 1; }
...}
```

atamazov:
## 3. What is Solver
Solver is an object which encapsulates the implementation of specific primitive. 
- It "knows" all the information required to properly use the kernel(s) it hides.
- Searchable (Tunable) Solvers work together with paired PerformanceConfig objects. In such a case the information about kernel(s) is distributed between Solver and its PerformanceConfig.

[...]

PR #1247, opened 2021-10-28
[Enhancement] xdlops NCHW support by transpose

carlushuang:
- [x] transpose hip kernel `src/kernels/gpu_batched_transpose_kernel/batched_transpose.cpp`
- [x] internal API to call transpose
- [x] ctest for batched transpose `test/gpu_nchw_nhwc_transpose.cpp`
- [x] add NHWC solver support NCHW by using transpose
- [x] test [ssd](https://amdcloud-my.sharepoint.com/:x:/r/personal/carhuang_amd_com/Documents/miopen-perf-ssd.xlsx?d=w4a0fec7770424fea8be4f276e1200a16&csf=1&web=1&e=oPmF1z)

@atamazov @junliume this PR, especially the host part may need careful check. Maybe my implementation is not good and need big code refactoring, so please help review this.

tested on gfx908/gfx90a, in above link


src/include/miopen/magic_div.hpp
```diff
@@ -0,0 +1,95 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#ifndef GUARD_MLOPEN_MAGIC_DIV_HPP_
+#define GUARD_MLOPEN_MAGIC_DIV_HPP_
+
+#include <cstdint>
+#include <assert.h>
```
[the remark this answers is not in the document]
atamazov: tidy error due to assert.h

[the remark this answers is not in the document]
atamazov: CI job restarted without static checks to see of there is other problems.

src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -505,10 +509,87 @@ InvokerFactory MakeImplGemmDynamicForwardXdlopsNHWCInvokerFactory(
...
+            trans_input_idx = idx++;
+        if(!trans_weight_skippable)
+            trans_weight_idx = idx++;
+        if(!trans_output_skippable)
+            trans_output_idx = idx++;
+    }
+
+    const size_t cast_offset = isNCHW ? (trans_output_offset + trans_output_size) : 0;
+    const size_t cast_size =
+        need_cast ? static_cast<size_t>(n) * k * ho * wo * miopen::GetTypeSize(miopenFloat) : 0;
```
atamazov, line 584:
```suggestion
    const size_t cast_size = need_cast ? miopen::GetTypeSize(miopenFloat) * n * k * ho * wo  : 0;
```

src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -505,10 +509,87 @@ InvokerFactory MakeImplGemmDynamicForwardXdlopsNHWCInvokerFactory(
...
+
+    const int kID_trans_start = isGfx90aFp16altSupport ? 2 : 1;
```
atamazov, line 586: This is wrong. Kernel index shall be computed within Invoker. Otherwise it will remain the same regardless of the ALT attribute.

src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -517,42 +598,71 @@ InvokerFactory MakeImplGemmDynamicForwardXdlopsNHWCInvokerFactory(
...
-            }
-            else
-            {
-                opArgs[0] = OpKernelArg(tensors.in);
-                opArgs[1] = OpKernelArg(tensors.w);
-                opArgs[2] = OpKernelArg(tensors.out);
-            }
+            auto trans_input_buf = trans_input_size == 0
+                                       ? null_buf
+                                       : const_cast<miopen::Handle*>(&handle)->CreateSubBuffer(
```
atamazov, line 604: Make `CreateSubBuffer()` a const method and get rid of `const_cast`.

[Note] The rule of thumb: dangerous stuff (like `const_cast`) should not be used unless absolutely necessary and without explanatory comments.

atamazov (changes requested):

src/include/miopen/magic_div.hpp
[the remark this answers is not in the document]
atamazov: CI job terminated because changes are necessary anyway.

src/conv/invokers/impl_gemm_dynamic.cpp
carlushuang: OK, let me do this

src/conv/invokers/impl_gemm_dynamic.cpp
carlushuang: OK. Let me use the same method in the invoker to pick up the kernel index

src/conv/invokers/impl_gemm_dynamic.cpp
carlushuang: OK

atamazov:
It seems like I was wrong, not you. I need to recheck. (( Please hold back
code changes due to this comment.

src/conv/invokers/impl_gemm_dynamic.cpp
atamazov: Let me recheck. It seems like **_I_** was wrong.

atamazov: I withdraw my original comment. The code is correct.

src/include/miopen/magic_div.hpp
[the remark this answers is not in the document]
carlushuang: fixed

src/conv/invokers/impl_gemm_dynamic.cpp
carlushuang: fixed

src/conv/invokers/impl_gemm_dynamic.cpp
carlushuang: fixed

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -946,16 +972,57 @@ ConvSolution ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::GetSolution(
...
     result.construction_params[0].comp_options = opts_0.str();
+    std::ostringstream msg;
 
     if(isGfx90aFp16altSupport)
     {
         result.construction_params.push_back(kernel);
         std::ostringstream opts_1(options.str(), std::ios_base::ate);
         GenerateClangDefsym(opts_1, "igemm_bwd_fp16_alt_impl", 1);
         result.construction_params[1].comp_options = opts_1.str();
+        msg << ", fp16_alt:" << ctx.conv_problem.GetConv().attribute.gfx90aFp16alt.GetBwd();
```
atamazov, line 983: The computations related to `msg` should happen only if `msg` is going to be printed.

atamazov (changes requested):

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -946,16 +972,57 @@ ConvSolution ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::GetSolution(
...
+        }
+        if(!trans_output.IsSkippable())
+        {
+            result.construction_params.push_back(trans_output.GetKernel());
+            msg << ", out trans:" << trans_output.GetKernelName();
+        }
     }
 
-    MIOPEN_LOG_I2("ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC: " + config.ToString());
+    MIOPEN_LOG_I2("ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC: " + config.ToString() + msg.str());
```
atamazov, line 1025: You can use ostringstream syntax in `MIOPEN_LOG*` macros.
```suggestion
    MIOPEN_LOG_I2("ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC: " << config.ToString() << msg.str());
```

atamazov: Logging output begins with `[GetSolution]`, which creates an impression that logging happens during `GetSolution()` call. This is not so, however. The logging happens during invocation of kernels. Let's clearly indicate this:
```suggestion
    MIOPEN_LOG_I2("[INVOKER] ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC: " + config.ToString() + msg.str());
```

atamazov: Please always use SolverDbId():
```suggestion
    MIOPEN_LOG_I2(SolverDbId(*this) << ": " ...);
```

atamazov (changes requested):

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
carlushuang: > Logging output begins with `[GetSolution]`, which creates an impression that logging happens during `GetSolution()` call. This is not so, however. The logging happens during invocation of kernels. Let's clearly indicate this:

This IS in GetSolution right? not actually invoker the kernels

src/include/miopen/magic_div.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+namespace miopen {
+
+struct magic_div_u32
+{
+    uint32_t magic;
+    uint8_t shift;
+};
+
+using magic_div_u32_t = magic_div_u32;
```
atamazov, line 41: :red_circle: Something similar resides in src/include/miopen/conv/invokers/impl_gemm_dynamic.hpp. We do NOT create copies of existing code.

src/include/miopen/util_sol.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#ifndef MIOPEN_UTIL_SOL_HPP_
+#define MIOPEN_UTIL_SOL_HPP_
+
+#include <miopen/batched_transpose_sol.hpp>
```
atamazov, line 29: What is the "sol" abbreviation stands for? It would be better if all Solutions in MIOpen comply the Solver/Solution/Invoker arch (#866).

src/include/miopen/util_sol.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <miopen/miopen.h>
+#include <miopen/kernel_info.hpp>
+#include <miopen/op_kernel_args.hpp>
+#include <miopen/execution_context.hpp>
+#include <vector>
+
+namespace miopen {
+
+struct TransposeSolution_NCHW2NHWC : public BatchedTransposeSolution
```
atamazov, line 38: Why we need these wrappers?

atamazov: CamelCase
```suggestion
struct TransposeSolutionNchw2Nhwc : public BatchedTransposeSolution
```

src/include/miopen/util_sol.hpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    TransposeSolution_NCHW2NHWC(const ExecutionContext& ctx_,
```
atamazov, line 40: BTW It is hard to visually distinguish `Nchw2Nhwc` and `Nhwc2Nchw`, isn't it? Maybe better to use `Default2Nhwc` and `Nhwc2Default`?

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -30,6 +30,7 @@
 #include <miopen/gcn_asm_utils.hpp>
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/conv/asm_implicit_gemm.hpp>
+#include <miopen/util_sol.hpp>
```
atamazov, line 33: This header name looks strange here. If it provides interface to _transpose_ functionality, then it should be expressed in the header name.

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -893,16 +891,42 @@ bool ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::IsApplicable(const ConvolutionC
...
+    const auto& k         = ctx.n_inputs;
+    const auto& c         = ctx.n_outputs;
+    const auto& ho        = ctx.in_height;
+    const auto& wo        = ctx.in_width;
+    const auto& y         = ctx.kernel_size_h;
+    const auto& x         = ctx.kernel_size_w;
+    const auto& group     = ctx.group_counts;
+    const auto isNCHW     = ctx.IsLayoutDefault();
+    size_t workspace_size = 0;
+    if(isNCHW)
```
atamazov, line 906: CamelCase please

atamazov (changes requested):
Partial review

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
atamazov: Invoker is a lambda produced (instantiated) by another lambda (InvokerFactory). Both can be defined directly in GetSolution (like in this case) or somewhere else (only to avoid duplication of code).

InvokerFactory instance resides in the Solution object. The library runs it only when necessary, i.e. sometime later, after GetSolution(). The resulting Invoker object is instantiated and stored in the Invoker Cache.

Sometime more later, when the library wants to start the kernels, it will run Invoker.

src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -652,7 +772,7 @@ InvokerFactory MakeImplGemmDynamicBackwardDataXdlopsNHWCInvokerFactory(
     if(y < stride_h || x < stride_w || dilation_h != 1 || dilation_w != 1)
         need_set_zero = true;
     need_set_zero |= config.gemm_k_global_split > 0;
-    bool use_global_split = config.gemm_k_global_split > 0;
+    // bool use_global_split = config.gemm_k_global_split > 0;
```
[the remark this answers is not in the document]
carlushuang: OK

src/include/miopen/datatype.hpp
```diff
@@ -66,7 +66,34 @@ inline std::string GetDataType(miopenDataType_t type)
     return type_str;
 }
 
-inline std::size_t get_data_size(miopenDataType_t) { MIOPEN_THROW("not implemented"); }
+inline std::size_t get_data_size(miopenDataType_t type)
```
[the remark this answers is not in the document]
carlushuang: OK

src/hip/batched_transpose_sol.cpp
```diff
@@ -0,0 +1,334 @@
...
+#define BATCHED_TRANSPOSE_BLOCK_SIZE 256
+#define BATCHED_TRANSPOSE_PERSISTENT 0
+
+#if BATCHED_TRANSPOSE_PERSISTENT
+#define BATCHED_TRANSPOSE_OCCUPANCY 4
+#endif
+
+namespace miopen {
+
+std::string transpose_kernel_get_name_trait(std::size_t type_size)
```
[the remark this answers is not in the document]
carlushuang: OK, let me change it

src/hip/batched_transpose_sol.cpp
```diff
@@ -0,0 +1,334 @@
...
+    else
+        radio = static_cast<float>(kparam->tile_y) / kparam->tile_x;
+
+    // e.g. for cases like width=1000, height=10
+    // allow at least 32x64, 64x64... 16x64 not allowed
+    return radio >= 0.4;
+}
+
+template <typename T>
+static inline float get_normalized_radio(T x, T y)
```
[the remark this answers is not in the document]
carlushuang: it is a radio of for example, width/height, or height/width. I use this radio in side kernel selection logic for some special case (although not optimal, which is hard to achieve)

src/hip/batched_transpose_sol.cpp
```diff
@@ -0,0 +1,334 @@
...
+        return "byte";
+    if(type_size == 2)
+        return "half";
+    if(type_size == 4)
+        return "dword";
+    MIOPEN_THROW("data type not supported");
+}
+
+static inline const std::vector<BatchedTransposeParam>&
+get_transpose_kernel_list(std::size_t data_size)
```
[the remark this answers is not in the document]
carlushuang: Good idea

src/kernels/gpu_batched_transpose_kernel/batched_transpose.cpp
```diff
@@ -0,0 +1,2066 @@
...
+}
+
+inline __device__ void v_pack_b32_f16_00(float& c, const float& a, const float& b)
+{
+#if 0
+    asm volatile("v_pack_b32_f16 %0, %1, %2\n"
+                 : "=v"(c)
+                 : "v"(a), "v"(b));
+#else
+    const uint32_t x  = *reinterpret_cast<const uint32_t*>(&a);
```
junliume, line 48: @carlushuang @atamazov  HIP tidy is complaining about lots of casts in this file. If indeed they are necessary, let's silence the warning on `[invalidPointerCast]` for this file?

ref: [HIP tidy log](http://micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLibs/pipelines/MIOpen/branches/nchw_xdlops_support_by_transpose/runs/8/nodes/13/steps/63/log/?start=0#:~:text=kernels%20batch%20%2366-,%5B2021-11-03T10%3A44%3A22.913Z%5D%20/home/jenkins/workspace/nchw_xdlops_support_by_transpose/src,due%20to%20different%20binary%20data%20representations%20on%20different%20platforms.%20%5BinvalidPointerCast%5D,-%5B2021-11-03T10)

src/hip/batched_transpose_sol.cpp
```diff
@@ -0,0 +1,334 @@
...
+    kernel_name << type_trait;
+    return kernel_name.str();
+}
+
+static inline std::size_t get_extra_padding_size(uint32_t /* batch */,
+                                                 uint32_t height,
+                                                 uint32_t width,
+                                                 const BatchedTransposeParam* kparam)
+{
+    // for simplicity and speed, we ignore batch, only compute h*w
```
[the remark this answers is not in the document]
carlushuang: Sure

src/hip/batched_transpose_sol.cpp
```diff
@@ -0,0 +1,334 @@
...
+bool BatchedTransposeSolution::IsSkippable() const
+{
+    // if height or width is 1, actually no need to do transpose
+    // but nonthing prevent you from DO transpose...
+    return height == 1 || width == 1;
+}
+
+size_t BatchedTransposeSolution::GetSize() const
+{
+    return batch * height * width * miopen::GetTypeSize(data_type);
```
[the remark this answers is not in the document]
carlushuang: OK

src/include/miopen/util_sol.hpp
carlushuang: Because NCHW->NHWC, and NHWC->NCHW is exactly the same problem of transpose, only by switching the height/width.
By generalizing this problem in to a batched-transpose problem, we can simplify a lot of (maybe) duplicated code, and this transpose can further be extended.
This wrappers are just to make the transpose easier

src/include/miopen/magic_div.hpp
carlushuang: I understand this, the reason I don't want to use that code in `src/composable_kernel/composable_kernel/include/utility/magic_division.hpp` is that code reside in `ck` namespace. But here I use magic div and seems not related to the ck namespace. I mean, this would be code in common. If this really is a issue, maybe fix this in other PR? It may need to think how to reorganize this code

src/include/miopen/util_sol.hpp
carlushuang: sol is for solution. And I use this name because this is different from that in `util.hpp`, where there exist some method that use a single function call to launch kernel. At lease this transpose implementation, is not a single function call, but return a kernel object and let the caller to launch

src/include/miopen/util_sol.hpp
carlushuang: I'll change to `Nhwc2default`, `default2Nhwc`

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
carlushuang: I'll change to `is_nchw`

src/kernels/gpu_batched_transpose_kernel/batched_transpose.cpp
carlushuang: @junliume try workaround it by adding `// cppcheck-suppress invalidPointerCast`

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
carlushuang: fixed

junliume:
[CI] restarted CI due to a strange failure on Fp16 OpenCL Vega20 stage. Will follow up with progress.

junliume (approved):
CI has passed, review suggestions are applied.

src/include/miopen/magic_div.hpp
atamazov: Or course code duplication is an issue. It can kill any project, slowly but inevitably.

atamazov:
@junliume That was _partial_ review.

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
atamazov: "ask sol" :rofl:

junliume:
> @junliume That was _partial_ review.

Let's use post merge issues :) Sorry for pressing hard on these blocker PRs.

PR #1250, opened 2021-10-29
Non templated SolverBase

averinevg:
Changes in SolverBase:
- template -> abstract class
- all code using this class has been updated


junliume:
@atamazov @JehandadKhan 
Let's discuss next Monday and include this effort into Host/API roadmap :)

atamazov:
@junliume 
> @atamazov @JehandadKhan Let's discuss next Monday and include this effort into Host/API roadmap :)

Invite @DrizztDoUrden for in-depth tech expertiыe.

pfultz2:
Whats the purpose of this change?

We want to avoid an abstract base class(and inheritance-based polymorphism) as that forces indirection through pointers which has problems including:

* Changes the semantics of copy, assignment, and equality
* Leads to incidental data structures
* Inefficient due to heap allocation and aliasing
* Can be good as a global variable

Also the templated base class can allow us to auto register the solvers instead maintaining separate lists for registering them, like here:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/solver.cpp#L251

DrizztDoUrden:
> Changes the semantics of copy, assignment, and equality

We do neither of those for solvers.
>Leads to incidental data structures

I don't see how someone can incidentally create a data structure in solver.
>Inefficient due to heap allocation and aliasing

AnySolvers has heap allocations, aliasing and virtual calls anyway.
>Can be good as a global variable

Anything can be good as a global variable.

> Also the templated base class can allow us to auto register the solvers instead maintaining separate lists for registering them, like here
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/solver.cpp#L251

Honestly, I don't see how this part becomes any different. Just change `AnySolver` to `std::unique_ptr<SolverBase>` or something like that and it still works.

DrizztDoUrden:
> But dont you want to put these into a vector(which does all these)?

I don't see a problem with a vector of pointers.

> If a solver stored a pointer to another solver or had methods to return the pointer to another solver(perhaps to reuse methods from another solver), this can lead to an incidental data structure.

It can be used as it is now, with no pointers whatsoever. Like this: `SomeOtherSolver{}::Method(...)`.

> You dont need to use heap allocations. You can use SBO(small buffer optimization) or COW(copy on write) to reduce heap allocations. It seems to make sense for AnySolver to use SBO since most classes have no member variables. Even more so, you could enforce this requirement of no member variables with AnySolver but not with inheritance.
>
> In MIGraphX, we have a python script to generate type erasure boilerplate. It generates code using COW. Maybe we could extend this to support SBO, and MIOpen could reuse this script as well.

To be honest, I don't see any real benefit in the case of solvers for MIOpen.

> I am not saying this code wont work with a non-templated base. Rather, we can get rid of all this code when using templated base as the templated base class can auto register the solvers:
> ...

Can be done by moving the registration to a separate templated type and inheriting that as well. Even when designing the whole system, I have suggested a similar approach, but as everyone agreed earlier, that is not what we would want to implement anyway.

> Great! Then why are we refactoring it to use a pointer everywhere?

The point of the refactoring is not in the pointer itself.

First of all, virtual inheritance provides an explicit rather than implicit strong bond between the types. This leads to several benefits like readable compilation errors, meaningful autocomplete when implementing a new solver, a clear non-templated way to implement functions accepting solvers of different primitives, an easy way to encapsulate functionality in the solver type tree, lower compilation times.
We no longer need to offload a lot of related code to different places like find_solution.hpp.
AnySolver just vanishes as well, being completely redundant and duplicate. No more need to maintain two separate, but not identical copies of API.

Simply speaking the approach allows simplifying the source significantly and makes it easier for people to hop in and really understand what happens under the hood.

src/include/miopen/solver.hpp
```diff
@@ -118,24 +120,32 @@ struct SolverBase
...
     /// * Value 1.0 corresponds to the 100% utilization of HW capabilities as
     ///   if Direct computational algorithm is used.
     /// * [Notice] WTI may exceed 1.0 for highly optimized algorithms like Winograd.
     /// * @see https://github.com/ROCmSoftwarePlatform/MIOpen/issues/410
-    float GetWti(const Context&) const { return -2.0; }
+    virtual float GetWti(const boost::any& ctx_) const
+    {
+        std::ignore = ctx_;
+        return -2.0;
+    }
```
DrizztDoUrden, line 141:
```suggestion
    virtual float GetWti(const boost::any&) const { return -2.0; }
```

src/include/miopen/solver.hpp
```diff
@@ -118,24 +120,32 @@ struct SolverBase
...
 
     // Returns the workspace size required by the solver for a given ConvolutionContext
-    size_t GetWorkspaceSize(const Context&) const { return 0; };
+    virtual size_t GetWorkspaceSize(const boost::any& ctx_) const
+    {
+        std::ignore = ctx_;
+        return 0;
+    }
```
DrizztDoUrden, line 148:
```suggestion
    virtual size_t GetWorkspaceSize(const boost::any&) const { return 0; }
```

DrizztDoUrden (changes requested):
Minor cosmetic suggestions.

pfultz2:
> To be honest, I don't see any real benefit in the case of solvers for MIOpen.

The benefits for MIOpen are Avoiding heap allocations and better checking of type constraints such as there is no member variables in the classes or `GetSolution` and `GetPerformanceConfig` are using the same type. 

> Can be done by moving the registration to a separate templated type and inheriting that as well.

Are you saying the solvers need to inherit two classes? It seems simpler to inherit one templated base class that would inherit the non-template base class, and solvers wont need to worry about inheriting two classes.

> AnySolver just vanishes as well, being completely redundant and duplicate. No more need to maintain two separate, but not identical copies of API.

It would be better to change `AnySolver` to just do indirection to the base class this way you can get the advantages of using a virtual base class, the extra checking of type constraints(ie no member variables), avoid the need to do `any_cast`(which is just a step above `void*`), avoid heap allocation, avoid boilerplate of type erasure code, and auto registration of the solvers.

So you would have non-templated base class and then a templated mixin to do registration and encapsulate the `any_cast`:

```cpp
// Non-template solver base that defines the virtual interface
struct SolverBase {
    // Pure virtual methods
};

template<class Derived, class Context>
struct SolverMixin : SolverBase {
    virtual bool IsApplicable(const Context& ctx) const = 0;
    virtual bool IsApplicable(const boost::any& ctx) const override {
        return IsApplicable(boost::any_cast<const Context&>(ctx));
    }
};

template<class Derived>
using ConvSolver = SolverMixin<Derived, ConvolutionContext>;
```

And then `AnySolver` can use SBO and indirection to access methods(instead of using the type erasure boilerplate), something like this:

```cpp
struct AnySolver {

    AnySolver() : buffer(), copy(nullptr), p(nullptr)
    {}

    template<class Solver>
    AnySolver(const Solver& s)
    {
        static_assert(sizeof(Solver) == sizeof(SolverBase), "Solver must be stateless");
        static_assert(std::is_base_of<SolverBase, Solver>{}, "Not derived class of SolverBase");
        copy = +[](const Storage& src, Storage& dst, SolverBase** s) {
            new (std::addressof(dst)) Solver(cast<const Solver>(src));
            s = static_cast<SolverBase*>(cast<Solver>(dst));
        };

        new (std::addressof(buffer)) Solver(s);
        p = static_cast<SolverBase*>(cast<Solver>(buffer));
    }

    AnySolver(const AnySolver& rhs) : buffer(), copy(rhs.copy), p(nullptr)
    {
        copy(rhs.buffer, buffer, &p);
    }

    AnySolver& operator=(const AnySolver& rhs) {
        if (&rhs != this) {
            if (p)
                p->~SolverBase();
            copy(rhs.buffer, buffer, &p);
        }
        return *this;
    }

    const SolverBase* get() const noexcept {
        return p;
    }

    const SolverBase& operator*() const noexcept {
        return *get();
    }

    const SolverBase* operator->() const noexcept {
        return get();
    }

    ~AnySolver() noexcept {
        if (p)
            p->~SolverBase();
    }


private:
    using Storage = std::aligned_storage_t<sizeof(SolverBase), alignof(SolverBase)>;
    using Cloner = void(*)(const Storage&, Storage&, SolverBase**);

    template<class T, class S>
    static T& cast(S&& s) {
        return reinterpret_cast<T&>(std::addressof(buffer));
    }

    Storage buffer;
    Cloner copy;
    SolverBase* p;
};
```

I think this gets the best of both worlds. `AnySolver` works like `SolverBase*` except it is a regular type, avoids heap allocation, enforces all methods to be const, and enforces that the derived classes have no data members.

pfultz2:
> lower compilation times.

Using auto registration can improve compilation times because then you can move all solvers to a individual cpp files only(with no headers).

src/solver/conv_winoRxS_f2x3.cpp
```diff
@@ -855,15 +858,18 @@ ConvBinWinogradRxSf2x3::GetSolution(const ConvolutionContext& params,
...
+
     if(miopen::IsDisabled(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3_G1{}))
         return false;
     return IsApplicableBase(params) && params.group_counts == 1;
 }
 
-float ConvBinWinogradRxSf2x3g1::GetWti(const ConvolutionContext& params) const
+float ConvBinWinogradRxSf2x3g1::GetWti(const boost::any& ctx_) const
 {
+    auto params = boost::any_cast<const ConvolutionContext&>(ctx_);
```
pfultz2, line 872: A mixin should really be used to this casting instead of requiring every solver to do this.

src/include/miopen/find_solution.hpp
```diff
@@ -121,8 +121,6 @@ template <class Solver, class Context, class Db>
 ConvSolution
 FindSolution(Solver s, const Context& context, Db& db, const AnyInvokeParams& invoke_ctx)
 {
-    static_assert(std::is_empty<Solver>{} && std::is_trivially_constructible<Solver>{},
-                  "Solver must be stateless");
```
pfultz2, line 125: You should be able to at least check that `sizeof(Solver) == sizeof(SolverBase)`

pfultz2 (changes requested):

DrizztDoUrden:
I like the approach you have proposed with AnySolver + virtual base combination.

> Using auto registration can improve compilation times because then you can move all solvers to a individual cpp files only(with no headers).

I am with you on auto-registration, but at some point, we have decided that we want to preserve and manually control numeric solver id values.

pfultz2:
> I am with you on auto-registration, but at some point, we have decided that we want to preserve and manually control numeric solver id values.

Yea, I think it would be better to move the ID to a method of the solver. This would ensure the solvers always use the same ID regardless of what solvers are added or removed. The list seems a little fragile if someone adds the solver at the top or they remove a line and dont replace it with `id++`.

atamazov:
@pfultz2 
> The list seems a little fragile if someone adds the solver at the top or they remove a line and dont replace it with id++.

Yes, and that is why I want to replace `id++` stuff with literal constants (but do not have time yet((

pfultz2:
> Yes, and that is why I want to replace id++ stuff with literal constants

But why not put them into a `GetID` method?

src/include/miopen/find_solution.hpp
averinevg: Code was changed as you proposed, thanks

src/include/miopen/solver.hpp
averinevg: This code was changed

src/include/miopen/solver.hpp
averinevg: This code was changed

src/solver/conv_winoRxS_f2x3.cpp
averinevg: Code was changed as you proposed

averinevg:
Please review

DrizztDoUrden (commented):
LGTM

DrizztDoUrden (dismissed):

JehandadKhan (approved):

atamazov:
I think we do not want this to be included into 5.0, so let's put this on hold until release branch is cut.

atamazov (approved):
LGTM!

atamazov:
@pfultz2 Please let us know if you have any other comments.

src/include/miopen/solver.hpp
```diff
@@ -118,24 +120,24 @@ struct SolverBase
...
     /// * Value 1.0 corresponds to the 100% utilization of HW capabilities as
     ///   if Direct computational algorithm is used.
     /// * [Notice] WTI may exceed 1.0 for highly optimized algorithms like Winograd.
     /// * @see https://github.com/ROCmSoftwarePlatform/MIOpen/issues/410
-    float GetWti(const Context&) const { return -2.0; }
+    virtual float GetWti(const boost::any& ctx) const = 0;
 
     // Returns the workspace size required by the solver for a given ConvolutionContext
-    size_t GetWorkspaceSize(const Context&) const { return 0; };
+    virtual size_t GetWorkspaceSize(const boost::any& ctx) const = 0;
```
pfultz2, line 140: Dont you need a virutal method for `SolverDbId`?

src/include/miopen/solver.hpp
```diff
@@ -146,6 +148,33 @@ struct SolverBase
     ///                          const PerformanceConfig& config) const;
 };
 
+template<class Context>
+struct SolverMixin : SolverBase
```
pfultz2, line 152: The mixin should take the Derived class as a template parameter, so it can implement `SolverDbId` and possibly auto registration in the future:

```cpp
template<class Derived, class Context>
struct SolverMixin : SolverBase
```

atamazov:
@averinevg Let's provide feedback to Paul's comments, if you have time.

src/include/miopen/solver.hpp
averinevg: I suggest doing it later in separated PR

src/include/miopen/solver.hpp
averinevg: This is a good suggestion in my opinion. Let's add it after accepting this PR

averinevg:
@pfultz2 I took into account all your comments. I propose to implement them in separate PRs.

pfultz2 (approved):
Sounds good!

atamazov:
@averinevg Can you please fix merge conflicts, thanks!

atamazov (approved):
LGTM. I am going to merge this in 24hrs if no objections.

PR #1261, opened 2021-11-05
 [gfx90a][FP16][rocBLAS] Enable Gemm solver for Alt Impl

JehandadKhan:
- Follows the design of #1226
  - Therefore, FP16 ALT kernels in GEMM obey the existing `MIOPEN_DEBUG_CONVOLUTION_ATTRIB_FP16_ALT_IMPL`.
  - _[Informative]_ Also there is dedicated env control in rocBLAS, `ROCBLAS_INTERNAL_FP16_ALT_IMPL`.
- Short explanation:
  - GEMM Invokers read gfx90aFp16alt attribute from the instance of AnyInvokeParams passed to it.
  - This attribute is then forwarded down to calls of `rocblas_gemm_ex` and `rocblas_gemm_strided_batched_ex`.
  - rocBLAS will make the determination at runtime, and ignore the attribute when not executing on gfx90a
  - :warning: We expect that rocBLAS ignore the attribute for non-FP16 kernels.
- This PR have no effect for GEMM backends other than rocBLAS.

<details><summary><b><i>Previous description</i></b></summary><hr>

 **_Should be merged after #1258_** 
This PR:

* Adds the alt impl attribute to GEMM problem descriptor
* Translates the above attribute to appropriate rocblas flag
* Adds a new backward compatability macro

Does NOT:
* Change existing solvers which use GEMM (Multipass winograd) since they are inapplicable for Alt Impl scenarios
* Change existing logic in RNNs which uses rocblas since that needs additional API work to expose

Future Work:

Perhaps we can merge many of the factory functions for the GemmDescriptor.
<hr></details>


junliume:
@atamazov  and @DrizztDoUrden could you review this PR? I think it is tested as in one of the Teams chat channel.

atamazov:
@junliume According to our today's talk with @JehandadKhan, this PR needs some updates.

src/gemm_v2.cpp
```diff
@@ -61,6 +61,9 @@
 /// Maintain API compatibility with various rocBLAS version
 #define USE_GEMM_FLAGS_PACK_INT8X4 (MIOPEN_ROCBLAS_VERSION_DECIMAL >= 238)
 
+/// Maintain API compatibility for versions not supporting FP16 alternate implementations
+#define USE_GEMM_FLAGS_FP16_ALT_IMPL (MIOPEN_ROCBLAS_VERSION_DECIMAL >= 242)
```
atamazov, line 65: [Recommendation]
```suggestion
#define FP16_ALT_KERNELS_AVAILABLE (MIOPEN_ROCBLAS_VERSION_DECIMAL >= 242)
```

src/gemm_v2.cpp
```diff
@@ -484,7 +487,12 @@ miopenStatus_t CallGemm(const Handle& handle,
                 rocblas_datatype::rocblas_datatype_f32_r,
                 rocblas_gemm_algo::rocblas_gemm_algo_standard,
                 0,
-                0);
+#if USE_GEMM_FLAGS_FP16_ALT_IMPL
+                gemm_desc.gfx90aaltimpl ? rocblas_gemm_flags_fp16_alt_impl : 0
```
atamazov, line 491: Proposed approach: GEMM Invokers read `gfx90aFp16alt` attribute (bool) from the instance of `AnyInvokeParams` passed to it. This bool should be then s forward it down to this function. And here let's use it instead of `gemm_desc.gfx90aaltimpl`.

AFAIU, unlike [this code](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1226/files#r744003897), we do not need to check either for gfx90a target (and maybe FP16 datatype check is not required as well).

src/gemm_v2.cpp
```diff
@@ -1267,15 +1286,16 @@ GemmDescriptor CreateGemmDescriptorConvFwd(const TensorDescriptor& wDesc,
...
+    int lda                  = k;
+    int ldb                  = wDesc.GetType() == miopenInt8 ? k : n;
+    int ldc                  = n;
+    int batch_count          = 1;
+    long long int strideA    = 0;
+    long long int strideB    = 0;
+    long long int strideC    = 0;
+    float alpha              = 1.;
+    float beta               = 0.;
+    const auto gfx90aaltimpl = problem.IsGfx90aFp16altRequired();
```
atamazov, line 1298: [Note] `IsGfx90aFp16altRequired()` in the problem description is intended for use only in Solver that do not support ALT kernels, for disabling them.

junliume:
> @junliume According to our today's talk with @JehandadKhan, this PR needs some updates.

@JehandadKhan  when you are done, merge develop and it should resolve the CI blockers now.

src/gemm_v2.cpp
JehandadKhan: Done!

atamazov (approved):
LGTM!

atamazov:
@JehandadKhan @sunway513 PR description updated, please have a look.

atamazov (approved):
LGTM!

src/solver/gemm.cpp
```diff
@@ -368,8 +368,17 @@ ConvSolution GemmFwd1x1_0_2::GetSolution(const ExecutionContext& context,
...
+                                                         gemm_desc,
+                                                         w,
+                                                         0,
+                                                         workSpace,
+                                                         0,
+                                                         workSpace,
+                                                         x_t_size,
+                                                         nullptr,
+                                                         GemmBackend_t::miopentensile,
+                                                         conv_params.gfx90aFp16alt);
```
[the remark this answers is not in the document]
JehandadKhan: Thanks for verifying, I dont like this pattern either. Perhaps we can discuss how to fix it so that the code does what it reads it does.

PR #1264, opened 2021-11-07
Fix HIP builds for the upcoming HIP compiler (SWDEV-310166)

atamazov:
This is urgent PR that is expected to unblock hip promotion (https://ontrack-internal.amd.com/browse/SWDEV-310166).

- [HIP offline build] Remove x86 builtin lib from linking.
- [COMGR] Remove x86 builtin lib from linking more reliably.
- By-products
  - [Future ROCm] Disable `-Wreserved-identifier` for declarations of llvm intrinsics in HIP kernels. Otherwise DEV builds fail at runtime, during HIP compilation (due to -Werror).

### DIFF
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -266,19 +266,25 @@ static bool IsLinkerOption(const std::string& option)
 
 static void RemoveCommonOptionsUnwanted(OptionList& list)
 {
-    list.erase(remove_if(list.begin(),
-                         list.end(),
-                         [&](const auto& option) { // clang-format off
+    list.erase(
+        remove_if(
+            list.begin(),
+            list.end(),
+            [&](const auto& option) { // clang-format off
                              return miopen::StartsWith(option, "-mcpu=")
                                 || (option == "-hc")
                                 || (option == "-x hip") || (option == "-xhip")
                                 || (option == "--hip-link")
-                                || (option.find("builtins-x86_64") != std::string::npos)
+                                // The following matches current "-lclang_rt.builtins-x86_64" (4.5) as weel as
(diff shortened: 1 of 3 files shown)


src/comgr.cpp
```diff
@@ -273,7 +273,7 @@ static void RemoveCommonOptionsUnwanted(OptionList& list)
                                 || (option == "-hc")
                                 || (option == "-x hip") || (option == "-xhip")
                                 || (option == "--hip-link")
-                                || (option == "-lclang_rt.builtins-x86_64")
+                                || (option.find("builtins-x86_64") != std::string::npos)
```
cjatin, line 276: This check might still cause some issues in future.
clang_rt.builtins lib's name itself is under change in LLVM.
currently its clang_rt.builtins-x86_64, they are planning to move it under a folder called x86_64 and calling the lib just 'clang_rt.builtins'
We should check for 'clang_rt.builtins' instead of 'builtins-x86_64'

atamazov: @cjatin Thanks for reviewing. This sournds reasonable. But is it possible that in the future, some `clang_rt.builtins` (that resides in some other folder) will contain _device_ code?

cjatin: I am not sure, I think @yxsamliu can answer it.

atamazov: @cjatin 
> This check might still cause some issues in future.

I'm afraid the immediate cause of the issues in the future will be changes in the compiler :smile:

But seriously speaking, it would be nice to fetch something like `HIP_COMPILER_FLAGS_DEVICE_ONLY` from hip::device.

atamazov: Resolved.

alvano (approved):

PR #1267, opened 2021-11-08
[BugFix] asm igemm heuristic init, for fwd/bwd/wrw

carlushuang:
fix a bug in when heuristic init asm igemm fwd/bwd/wrw. Previously config like:

`./bin/test_conv2d --half --verbose --input 120 512 19 19 --weights 510 512 3 3  --pads_strides_dilations 1 1 1 1 1 1 --disable-backward-data --disable-backward-weights  --in_layout NHWC --fil_layout NHWC --out_layout NHWC `

will fail. This is due to a bug in heuristic init. If we run with `FIND_ENFORCE=4`, or tune the db, then everything is OK.

[informative] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1267/commits/27c8024374393e40d8ce9f904925406b00bf9bea message is not correct, actually wrw also have bug when heuristic init. bug is fixed and double checked SSD can converge

### DIFF
--- a/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
@@ -623,9 +623,7 @@ void PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC::HeuristicInit(const Convo
     MIOPEN_LOG_I("m_per_block:" << m_per_block << ", n_per_block:" << n_per_block
                                 << ", k_per_block:" << k_per_block);
 
-    if((m_per_block == 0 && n_per_block == 0 && k_per_block == 0) || not_support_vector_store)
-    {
-        // not found, let's try  gemm_k pad now.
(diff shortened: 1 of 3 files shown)


JehandadKhan:
This PR only touches HeuristicsInit and should be fine from my point of view, I have tagged @asroy for review.

atamazov (dismissed):
LGTM

shaojiewang (dismissed):
Could you please add some test cases?

junliume:
> Could you please add some test cases?

Would the sample listed in the description be added as a unit test?
> ./bin/test_conv2d --half --verbose --input 120 512 19 19 --weights 510 512 3 3 --pads_strides_dilations 1 1 1 1 1 1 --disable-backward-data --disable-backward-weights --in_layout NHWC --fil_layout NHWC --out_layout NHWC

junliume (approved):

atamazov (approved):
LGTM
This does not seem worse then it was before ;)

PR #1276, opened 2021-11-10
[NFC] Cleaned-up batchnorm and activation

DrizztDoUrden:
Primitive calls contain a lot of similar code. I have moved it to a function to reduce duplications. The same function would be used in other primitives.

### DIFF
--- a/src/include/miopen/find_solution.hpp
+++ b/src/include/miopen/find_solution.hpp
@@ -29,7 +29,9 @@
 
 #include <miopen/env.hpp>
 #include <miopen/conv_solution.hpp>
+#include <miopen/execution_context.hpp>
 #include <miopen/find_controls.hpp>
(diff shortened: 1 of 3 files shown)


atamazov:
@DrizztDoUrden Is it trivial?

atamazov (approved):
`+46 -122` is always good to merge ))

junliume (approved):

src/include/miopen/find_solution.hpp
```diff
@@ -296,6 +298,37 @@ struct SolverContainer
 
         return found;
     }
+
+    template <class Problem>
+    void ExecutePrimitive(Handle& handle,
```
atamazov, line 303: @DrizztDoUrden Is it possible to make handle const?

PR #1271, opened 2021-11-09
[Tuning][90a] Perf and Find Update

cderb:
using: rocm 4.5
perf db using miopen: '19defc31e' (develop 11/04)
find db using miopen: '0b3ea328a' (develop 11/08)

Issue #10, opened 2017-07-13
A little bug when running softmax with debug version


koverlu:
file: src/kernel_cache.cpp
line: 91
```c++
    std::cout << "runcl " << params << " src/Kernels/" << program_name << " -k " << kernel_name
              << " -dumpilisa -r 10"
              << " if#" << isize * 4 << ": if#" << msize * 4 << ": if#" << osize * 4 << ": iv#0 "
              << vgd[0] << "," << vgd[1] << "," << vgd[2] << "/" << vld[0] << "," << vld[1] << ","
              << vld[2] << std::endl;
```
When running command `MIOpenDriver softmax -n 64 -W 1 -H 1 -c 1000 -V 0 -i 1 -t 1`, with a debug building (NDBUG is undefined)
`vgd` will be a const vector which size is 1. (in src/ocl/softmaxocl.cpp#81)

dagamayank:
@koverlu Thanks for reporting. We will fix it.

dagamayank:
Fixed in 173fe4d350e40a669e86e10f1cb72d354e62f553

PR #1278, opened 2021-11-10
[HOTFIX] Fix issue: Custom tests that do not have MLIR_ENABLED are disabled by default

atamazov:
The bug is originated from my PR #1246 (which is unsuccessful attempt to fix another problem with MLIR testing). Currently many custom tests (including new test_regression_half_mi200) are not working.

## :warning: In order to prevent leaking regressions into develop, it is highly recommended to pause merging other PRs until this one is merged (and other PRs merge from `develop` after that and pass all tests).

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -48,7 +48,6 @@ option( WORKAROUND_ISSUE_898 "" ON)
 option( WORKAROUND_ISSUE_936 "" ON)
 option( WORKAROUND_ISSUE_1053 "" ON)
 option( WORKAROUND_ISSUE_1148 "" ON)
-option( WORKAROUND_ISSUE_1281 "" ON)
 
 # Run the test suite to a depth limit
 #limit greater than 2 leads to prolonged testing more than 5hrs per stage.
@@ -561,14 +560,6 @@ function(add_custom_test NAME)
           OR ${NAME} MATCHES "test_conv_extra"
           OR ${NAME} MATCHES "test_conv_for_implicit_gemm" ))
         set_tests_properties(${NAME} PROPERTIES RUN_SERIAL On)
-    endif()
-
-    if(WORKAROUND_ISSUE_1281
-        AND MIOPEN_TEST_GFX90A
-        AND MIOPEN_TEST_FLOAT
-        AND (NOT MIOPEN_TEST_GPU_XNACK_ENABLED)
-        AND (${NAME} MATCHES "test_conv_group"))
-        set_tests_properties(${NAME} PROPERTIES DISABLED On)
     endif()
 
     if(  (is_vega_check OR is_gfx908_check OR is_gfx1030_check OR is_gfx90a_check)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -48,6 +48,7 @@ option( WORKAROUND_ISSUE_898 "" ON)
 option( WORKAROUND_ISSUE_936 "" ON)
 option( WORKAROUND_ISSUE_1053 "" ON)
 option( WORKAROUND_ISSUE_1148 "" ON)
+option( WORKAROUND_ISSUE_1281 "" ON)
 
 # Run the test suite to a depth limit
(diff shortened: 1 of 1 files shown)


junliume:
> The bug is originated from my PR #1246 (which is unsuccessful attempt to fix another problem with MLIR testing). Currently many custom tests (including new test_regression_half_mi200) are not working.
> 
> ## ⚠️ In order to prevent leaking regressions into develop, it is highly recommended to pause merging other PRs until this one is merged (and other PRs merge from `develop` after that and pass all tests).

I can pause some other PRs to yield CI resources for #1271 and #1278 
Anything we can do to process the CI selectively? The whole run might take much longer. Thanks!

test/CMakeLists.txt
```diff
@@ -651,14 +648,14 @@ set(IMPLICITGEMM_MLIR_ARGS_F ${IMPLICITGEMM_ARGS} --verbose --disable-backward-d
 set(IMPLICITGEMM_MLIR_ARGS_B ${IMPLICITGEMM_ARGS} --verbose --disable-forward --disable-backward-weights)
 set(IMPLICITGEMM_MLIR_ARGS_W ${IMPLICITGEMM_ARGS} --verbose --disable-forward --disable-backward-data)
 
-add_custom_test(test_conv_igemm_mlir_small HALF_ENABLED MLIR_ENABLED GFX908_DISABLED GFX90A_DISABLED
+add_custom_test(test_conv_igemm_mlir_small HALF_ENABLED SKIP_UNLESS_MLIR GFX908_DISABLED GFX90A_DISABLED
```
jerryyin, line 651: Why is this necessary? What was the problem in the implementation of `MLIR_ENABLED`?

This seems as if we'd still need to keep the MLIR smoke tests after #1228, is that right?

atamazov: > Why is this necessary? What was the problem in the implementation of MLIR_ENABLED?

Due to a logic error in computation of `is_mlir_check`, currently `is_mlir_check` is always false, unless both MLIR_ENABLED attribute and MIOPEN_TEST_MLIR are set. This leads to that all custom tests are disabled by default.

atamazov: > This seems as if we'd still need to keep the MLIR smoke tests after #1228, is that right?

This is out of scope of this PR, but yes, because MLIR tests are intended to test MLIR solvers. If, due to some reason, we will disable MLIR, then these tests should be skipped. This applies to all MLIR tests, not only smoke ones.

atamazov:
@junliume 
> Anything we can do to process the CI selectively? The whole run might take much longer. 

No, we should run all tests to make sure that there is no new failures (that may appear due to re-enabled custom tests).

atamazov:
:warning: New failure: Newly enabled `test_conv_group` is failing for Fp32 Hip gfx908a. I am working on this.

test/CMakeLists.txt
jerryyin: Sounds good, I will amend the jenkinsfile once your PR is merged.

junliume:
@atamazov CI pipeline `#3` and `#4` are ongoing simultaneously for `be086ff`. I think we only need to test **Full Tests II** and after, right?

atamazov:
#3 can be terminated, but in this case, YES, we need to run at least Full Tests II for all targets. #4 is for quick verification on gfx90a.

junliume:
> #3 can be terminated, but in this case, YES, we need to run at least Full Tests II for all targets. #4 is for quick verification on gfx90a.

Done, and a minor issue is that we cannot split Full Tests I and Full Tests II (easily) via the build parameters?
Monitoring run `#4` and `#5`.

atamazov:
@junliume 
> Done, and a minor issue is that we cannot split Full Tests I and Full Tests II (easily) via the build parameters?

In this case it is possible to filter out all data types except FP32.

atamazov:
Full job `#9` (except static & package) launched the double check that everything is fine.

junliume (approved):
We have ensured that test_conv_group is passing.

Run `#10` is running for safety check but we can leave it as post-merge.

PR #1279, opened 2021-11-11
[gfx90a]Fix an issue in the Asm IGEMM dynamic solvers Perf Config

JehandadKhan:
This PR updates the Performance config for the ASM Implicit GEMM solvers where the deserialize protocol does not correctly read back the precision.

These code changes were drafted by @atamazov I have validated the fix on the target architecture.

### DIFF
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -2557,7 +2557,7 @@ struct PerformanceConfigAsmImplicitGemmGTC : Serializable<PerformanceConfigAsmIm
 {
     std::string direction;
     std::string tensor_layout;
-    miopenDataType_t precision;
+    std::string precision;
     int nxb;
     int nxe;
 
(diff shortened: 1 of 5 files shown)


src/include/miopen/solver.hpp
```diff
@@ -2672,12 +2697,9 @@ struct PerformanceConfigAsmImplicitGemmGTC : Serializable<PerformanceConfigAsmIm
     template <class Self, class F>
     static void Visit(Self&& self, F f)
     {
-        std::string prec_string = self.precision == miopenFloat
-                                      ? "fp32"
-                                      : (self.precision == miopenHalf ? "fp16" : "bf16");
         f(self.direction, "dir");
         f(self.tensor_layout, "lyt");
-        f(prec_string, "pre");
+        f(self.precision, "pre");
```
atamazov, line 2702: This the most important fix. Adds deserialization for precision, which was missing and that was the root of the problems.

src/solver/conv_asm_implicit_gemm_gtc_perf_config.cpp
```diff
@@ -88,13 +88,70 @@ PerformanceConfigAsmImplicitGemmGTC::PerformanceConfigAsmImplicitGemmGTC(
 {
 }
 
+PerformanceConfigAsmImplicitGemmGTC::PerformanceConfigAsmImplicitGemmGTC(
+    std::string dir,
+    std::string layout,
+    miopenDataType_t prec,
```
atamazov, line 94: We need to maintain ctor where precision is miopenDataType_t only to initialize vectors ("lists") in Solvers.

atamazov (approved):
Hope this helps.

junliume (approved):

carlushuang (approved):
LGTM

atamazov:
@junliume develop wasn't merged into this branch, so let's track if develop passes all tests now.

junliume:
> @junliume develop wasn't merged into this branch, so let's track if develop passes all tests now.

I checked before merging and it was on tip of develop (without #1278) ... I will monitor the develop CI too. Thanks!

Issue #1257, opened 2021-11-02
[gfx90a][COMGR] Verification errors of ImplicitGemm kernels (no errors with hip-clang)


atamazov:
This prevents from switching to COMGR builds by default.

[DRAFT]
Some tests PASS with hip-clang, FAIL with COMGR.
Both ROCm 4.3.1 and 4.5 RC4.
gfx906a target only ?

<details><summary><b><i>Log from 4.3.1</i></b></summary>

```
 9/87 Test #11: test_conv2d ...........................................***Failed  Error regular expression found in output. Regex=[FAILED]349.20 sec
...
./bin/test_conv2d --half --cmode conv --pmode same --group-count 1 --batch_size 1 --input_channels 16 --output_channels 32 --spatial_dim_elements 28 28 --filter_dims 5 5 --pads_strides_dilations 0 0 1 1 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW 
FAILED: 0.729438
Iteration: 284
Forward convolution: ConvHipImplicitGemmForwardV4R4Xdlops
Input tensor: 1, 16, 28, 28
Weights tensor: 32, 16, 5, 5
Output tensor: 1, 32, 28, 28
Filter: conv2d, miopenConvolution, miopenPaddingSame, {2, 2}, {1, 1}, {1, 1}, 
Max diff: 1137
Mismatch at 0: -115 != -419
./bin/test_conv2d --half --cmode conv --pmode same --group-count 1 --batch_size 1 --input_channels 16 --output_channels 32 --spatial_dim_elements 28 28 --filter_dims 5 5 --pads_strides_dilations 1 1 1 1 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW 
FAILED: 0.729438
Iteration: 286
Forward convolution: ConvHipImplicitGemmForwardV4R4Xdlops
Input tensor: 1, 16, 28, 28
Weights tensor: 32, 16, 5, 5
Output tensor: 1, 32, 28, 28
Filter: conv2d, miopenConvolution, miopenPaddingSame, {2, 2}, {1, 1}, {1, 1}, 
Max diff: 1137
Mismatch at 0: -115 != -419
...
34/87 Test #25: test_immed_conv2d .....................................***Failed  Error regular expression found in output. Regex=[FAILED]103.55 sec
...
[...]

PR #1288, opened 2021-11-13
[HIP][COMGR] Use COMGR by default for HIP backend.

atamazov:


junliume:
@atamazov  should we upgrade CI to ROCm 4.5 before or after this PR? The label is urgent. If we are talking about absolute path issue, we are getting it deferred anyway.

atamazov:
As far as I understand, only base installation should be upgraded. No changes in the Dockerfile required.

atamazov:
Blocks 5.1

junliume:
@atamazov is this PR still been blocked by #1257 ?

atamazov:
Yes.

atamazov:
#1257 is not a blocker anymore.

atamazov:
Hmm, it seems like we need both (2) and (3) from https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1257#issuecomment-972820638 to get around the issue.

atamazov:
> #1257 is not a blocker anymore.

Maybe I am wrong. Why: Some CI nodes were upgraded to 5.11.32, and therefore #1257 is hardly reproducible.

I've disabled hpe-rack-15 (with 5.11.32) for a while. Please note that after that we have only one functional node in CI; other MI200 nodes are disabled due to issues.

atamazov:
All the issues resolved, CI tests passed. Please review.

atamazov:
If we do not want this to be included into 5.0, then let's keep this on hold until the release branch is cut.

atamazov:
@junliume The PR includes:

> [COMGR] Fix: Don't remove "-mllvm ..." fetched from hip::device properties. Split before passing to COMGR.

I am going to create a separate PR with this fix in order to include it into 5.0. This is not very important because there is no known issues related to this, but let's better be proactive.

CMakeLists.txt
```diff
@@ -290,7 +290,7 @@ else()
     endif()
 endif()
 
-set_var_to_condition(MIOPEN_USE_COMGR_DEFAULT MIOPEN_EMBED_BUILD)
+set_var_to_condition(MIOPEN_USE_COMGR_DEFAULT (NOT DEFINED MIOPEN_BACKEND_OPENCL) AND (NOT DEFINED MIOPEN_MODE_NOGPU))
```
shurale-nkn, line 293: I would prefer to solve this problem in OpenCL and NOGPU handlers
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/72f06aa8cb53c30ac2f87d34378925ff9401f4ed/CMakeLists.txt#L155-L158
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/72f06aa8cb53c30ac2f87d34378925ff9401f4ed/CMakeLists.txt#L223-L226

shurale-nkn (approved):
LGTM

CMakeLists.txt
atamazov: If I do so, then I will need to first set MIOPEN_USE_COMGR_DEFAULT to TRUE and then change it to FALSE in some cases. This makes it harder to track the value of MIOPEN_USE_COMGR_DEFAULT. Do you agree?

shurale-nkn: I suppose you can set MIOPEN_USE_COMGR to FALSE in these two cases, instead of changing MIOPEN_USE_COMGR_DEFAULT

atamazov: MIOPEN_USE_COMGR is an Cmake option, I need separate variable to initialize its default.

shurale-nkn: ohh, yes, right.
Then first case with MIOPEN_USE_COMGR_DEFAULT=true
``If I do so, then I will need to first set MIOPEN_USE_COMGR_DEFAULT to TRUE and then change it to FALSE in some cases. This makes it harder to track the value of MIOPEN_USE_COMGR_DEFAULT. Do you agree?``

shurale-nkn: I want this, just not to lose the reason for these exceptions in the code

atamazov: But the reason is the same - it doesn't work with OpenCL BE and we do not need to for NOGPU ;)

atamazov: I mean that the reason is programmatically expressed in the code. Either in some local place or scattered around. I would prefer locality.

Let's look at it differently. If this is written in C++, then single initialization of MIOPEN_USE_COMGR_DEFAULT allows defining it  as `const`. But with IFs this is not possible.

shurale-nkn: For me local place - in backend handlers, afther parametrs check and error message.
But these is only my preferences. I have already approved this PR.

atamazov:
@JehandadKhan I need your blessing))

JehandadKhan (approved):

junliume (approved):

PR #1290, opened 2021-11-15
Serialize performance config to binaries

pfultz2:
This will serialize the configs to binary with metadata for version and solver ID(and it will check for mismatches). The binary config data can be added to the `Context`, which FindSolution will use when available.

The data is stored using msgpack binary data so that parsing is fast(vs text processing), compact, and secure(since the libraries are fuzz tested). I used nlohmann/json to write msgpack instead of the other msgpack library since it provides a simpler API.

Unit tests are added for this. There is no library API implemented yet, as there may still be a discussion about how the API is supposed to work. For now, this just implements the backend functionality.


DrizztDoUrden (approved):
LGTM

junliume:
Under my impression from the discussion last week (Wk46) this performance config issue is still under discussion. We'd both agree and hope this feature request to pass KD (Key Decision) process with full reviews before proceeding with implementation. Thanks!

pfultz2:
> Under my impression from the discussion last week (Wk46) this performance config issue is still under discussion.

This is why I held off on adding the library API. Those design discussion is mainly about how to interface to MIGraphX not about how MIOpen will implement it.

JehandadKhan:
Closing this since the PR has gone stale.

PR #1298, opened 2021-11-19
[MLIR] Removing MLIR dependency requirement until ROCm 5.1

jerryyin:
Clarified with @JehandadKhan and @junliume then realized that the existing develop is still in ROCm 5.0 timeline. Since ROCm release requires linear history, I come up with this PR shooting for 
 - merge pre-ROCm 5.0 branch as the last PR, such that ROCm 5.0 doesn't have MLIR dependency
 - revert once ROCm 5.0 forked, such that MLIR will be default dependency in ROCm 5.1

Or, this can be cherry-picked in ROCm 5.0 release branch so no change is necessary in the develop branch.

Call for feedbacks in the best way to deal with it.


junliume:
[comments]
Building the MLIR dependency is really slow, now `cmake -P install_deps.cmake ` takes more than 40 mins on my local machine (build dependencies from scratch).

atamazov (approved):
This matches the "(4) Disable MLIR via CMake in amd-develop only" recently discussed in mail, except that this should be merged into `amd-develop` instead of `develop`

atamazov:
Smoke tests passed, the rest of CI is in progress...

atamazov:
Rebased as per #1306

atamazov:
This should not be merged because 95b58f72f _"Removing MLIR dependency requirement until ROCm 5.1"_ was merged into `amd-develop` directly.

atamazov:
Side effect: Dependency on MLIR library will persist in 5.0 in spite of that it is not really needed.

PR #1299, opened 2021-11-19
[MLIR] Adding misc optimizations

jerryyin:
This is to capture a few optimizations I talked with @atamazov offline:
 - Disable non-xdlops kernels from picking up on gfx908 (or gfx90a) to reduce compilation overhead
 - Add cmake error string to remind user to reinstall dependencies


atamazov (approved):
LGTM

Issue #1297, opened 2021-11-19
[iGemm] Some ConvHipImplicitGemm* solvers have very slow IsApplicable(): 5000...10000 us


atamazov:
Some of the Solvers have exceptionally high execution time of `IsApplicable()`. This happens because `IsApplicable()` call `HeuristicInit()`, which, in turn, is time consuming.

The list of solvers:
- ConvHipImplicitGemmForwardV4R4Xdlops
- ConvHipImplicitGemmForwardV4R4Xdlops_Padded_Gemm
- ConvHipImplicitGemmForwardV4R5Xdlops
- ConvHipImplicitGemmWrwV4R4Xdlops
- ConvHipImplicitGemmWrwV4R4Xdlops_Padded_Gemm

## Log of GWSS and interpretation

Notice that the issue affects not only GWSS but much more functionality of the library. GWSS is simply convenient for time measurements of IsApplicable() by means of standard library's means.

<details><summary><b><i>Full log that shows the case during GetWorkspaceSize</i></b></summary>

```
# MIOPEN_FIND_MODE=normal \
G_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS=1 \
MIOPEN_LOG_LEVEL=6 \
MIOPEN_ENABLE_LOGGING_ELAPSED_> MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS=1 \
> MIOPEN_LOG_LEVEL=6 \
> MIOPEN_ENABLE_LOGGING_ELAPSED_TIME=1 \
> ./bin/MIOpenDriver conv -x 3 -y 3 -W 17 -H 17 -c 192 -n 128 -k 192 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -s 0 \
> -w 1 -t 1 -F 1 -i 1 -V 0
MIOpenDriver conv -x 3 -y 3 -W 17 -H 17 -c 192 -n 128 -k 192 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -s 0 -w 1 -t 1 -F 1 -i 1 -V 0
MIOpen(HIP)   0.001: Info [get_device_name] Raw device name: gfx90a:sramecc+:xnack-
MIOpen(HIP)   0.097: Info [Handle] stream: 0xb9f650, device_id: 0
[...]

PR #1300, opened 2021-11-19
[conv] Optimize performance of GWSS

atamazov:
- Optimizes GWSS for DYNAMIC_HYBRID mode
- Optimize out `IsApplicable()` calls when workspace is not necessary for a Solver
  - `MayNeedWorkspace()` is introduced in the [Solver API](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/866#issuecomment-822585406) for that.
- By-products
  - Refactor GetWorkspaceSize -> GetWorkspaceSizes for clarity

## Average acceleration of GWSS calls

- Immediate mode API (`miopenConvolutionForwardGetSolutionWorkspaceSize` etc): 15...115% 
- Find API (`miopenConvolutionForwardGetWorkSpaceSize` etc): 2...17%

However the overall effect on the total Aux Wall Time (that includes GWSS time) is relatively small and does not exceed 2%.

## Side effects

The total GWSS time for the use case described at #1297 is progressed from ~24 to ~7 ms.

:warning: This may camouflage the visibility of #1297 during GWSS calls. So engineers who work on #1297 may wish to revert this PR in their development trees.

### DIFF
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -135,7 +135,10 @@ struct SolverBase
     float GetWti(const Context&) const { return -2.0; }
 
     // Returns the workspace size required by the solver for a given ConvolutionContext
-    size_t GetWorkspaceSize(const Context&) const { return 0; };
+    size_t GetWorkspaceSize(const Context&) const { return 0; }
+
+    // Must return true if a Solver has its own implementation of GetWorkspaceSize().
+    bool MayNeedWorkspace() const { return false; }
 
     /// Takes problem config, optimization parameters and other info
     /// and computes information required to build and run the kernel(s).
@@ -248,6 +251,7 @@ struct ConvAsm1x1U : SolverBase<ConvolutionContext>
                                         const AnyInvokeParams& invoke_ctx) const;
     bool IsApplicable(const ConvolutionContext& params) const;
     size_t GetWorkspaceSize(const ConvolutionContext& params) const;
+    bool MayNeedWorkspace() const { return true; }
     ConvSolution GetSolution(const ConvolutionContext& params,
                              const PerformanceConfigConvAsm1x1U& config,
                              bool disableConfigOverrideFromEnv = false) const;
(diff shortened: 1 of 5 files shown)


atamazov:
Please review.

src/include/miopen/solver.hpp
```diff
@@ -2345,6 +2364,7 @@ struct GemmFwd1x1_0_2 : GemmFwdBase
     {
         return GetWorkspaceSize(ctx, ctx.conv_problem);
     }
+    bool MayNeedWorkspace() const { return true; }
```
averinevg, line 2367: Wouldn't it be better to put MayNeedWorkspace() in the base class (GemmFwdBase)?

src/include/miopen/solver.hpp
```diff
@@ -2463,6 +2487,7 @@ struct GemmBwd1x1_stride1 : GemmBwdBase
...
```
averinevg, line 2490: Wouldn't it be better to put MayNeedWorkspace() in the base class (GemmBwdBase)?

src/include/miopen/solver.hpp
```diff
@@ -2515,6 +2541,7 @@ struct GemmWrw1x1_stride1 : GemmWrwBase
...
```
averinevg, line 2544: Wouldn't it be better to put MayNeedWorkspace() in the base class (GemmWrwBase)?

DrizztDoUrden (commented):
LGTM

src/include/miopen/solver.hpp
atamazov: That would create implicit dependence. `GemmFwdBase` does not provide a "base" implementation of  GetWorkspaceSize() and does not know if its successors need workspace.

atamazov:
> LGTM

Hey, I need also a green mark ;)

DrizztDoUrden (approved):

DrizztDoUrden:
My bad, misclicked.

atamazov:
Results of testing added. @junliume Does the positive effect worth the slightly increased complexity of the code, what do you think?

junliume (approved):
As discussed, every perf optimization is good :)

PR #1312, opened 2021-12-01
[Batchnorm][ASM] Disabling ASM kernel for unsupported target (gfx940)

muralinr:
Disabling Batchnorm ASM kernel for unsupported target (gfx940).


atamazov (approved):
LGTM

junliume (approved):

PR #1296, opened 2021-11-19
[API change] Pooling backward solvers (except GetWorkspaceSize)

DrizztDoUrden:
This PR contains both backward pooling solvers.
It now includes bwd transposed solvers.

PR #1313, opened 2021-12-01
Pooling driver command-line update

DrizztDoUrden:
This PR implements a universal tensor descriptor command-line argument parser. The change allows deep customization of each tensor involved, including not supported yet configurations, striving to make the driver closer to the API of the library.
This is also required to test future transposed NHWC pooling.
This neither blocks nor requires #1296.

An example with the four pooling tensors:
```
--dinput  -D Dinput tensor descriptor.
             Format: NxC[xD]xHxW[,LayoutOrStrides]
             Default: input tensor descriptor
--doutput -H Doutput tensor descriptor.
             Format: NxC[xD]xHxW[,LayoutOrStrides]
             Default: output tensor descriptor
--output  -O Output tensor descriptor.
             Format: NxC[xD]xHxW[,LayoutOrStrides]
             Default: generated from input tensor descriptor
--input   -W Input tensor descriptor.
             Format: NxC[xD]xHxW[,LayoutOrStrides]
             Default: 100x3x32x32,NCHW
```

Good single character arguments have been taken already, so here we go.
Call examples:
```
MIOpenDriver pool --input 200x16x100x100
MIOpenDriver pool --input 10x10x10x10,NWHC --output 10x10x8x8
MIOpenDriver pool --input 10x10x10x10,1000x100x10x1 --doutput 10x10x8x8,NWHC
```

Some details on the argument format (NxC[xD]xHxW[,Layout|SNxSC[xSD]xSHxSW]):
- The minimum amount of the details required to include in the provided tensor argument are lengths. If nothing else has been provided, the default layout is presumed, and strides are calculated based on that.
- With layout string provided (like NCDHW or NHWC), strides are calculated based on that and lengths.
- With strides provided, they are used as-is.

Pooling tensors calculation logic has been preserved except for a new ability to override all the tensors and layouts/strides. Also, spatial dims parameter is now implicitly calculated from the format of the input tensor descriptor.


atamazov:
@ce1adon @zjing14 What do you think about this?

atamazov (approved):
No objections!

atamazov:
This blocks pooling testing.

junliume (dismissed):
Ping @zjing14 and @JehandadKhan for review :)

driver/pool_driver.hpp
```diff
@@ -137,29 +136,43 @@ int PoolDriver_impl<Tgpu, Tref, Index>::ParseCmdLineArgs(int argc, char* argv[])
...
 {
-    spatial_dim             = inflags.GetValueInt("spatial_dim");
-    std::vector<int> in_len = GetInputTensorLengthsFromCmdLine();
+    auto input  = inflags.GetValueTensor("input");
+    spatial_dim = input.lengths.size() - 2;
+
+    if(input.SetTensordDescriptor(inputTensor, data_type) != miopenStatusSuccess)
+        MIOPEN_THROW("Error parsing input tensor: " + inflags.GetValueStr("input") + ".");
+
+    auto dinput = inflags.GetValueTensor("dinput").FillMissing(input);
```
JehandadKhan, line 145: Why do we need a special `dinput` flag? Cant we reuse the `input` flag for `dinput` etc ?

JehandadKhan (dismissed):
I like the idea in general, one question in comments.

driver/pool_driver.hpp
DrizztDoUrden: To allow passing any values to the API via the driver. It is not needed for valid configurations but allows bizarre case testing. More flexibility of the driver allows more experimentation when reproducing issues. I can remove this, but it allows easy access to providing incorrect data. Also it is initialized from `input` if no value provided, so it may be ignored when not neeeded.

DrizztDoUrden: Strictly speaking, we don't need it right now, but it may save some time later, one never knows.

atamazov (approved):
Let's push it through CI and merge.

junliume (approved):

Issue #1311, opened 2021-11-30
Layout Parameters - Error Handling


technicalgrp89:
convfp16 -n 128 -c 512 -H 28 -W 28 -k 128 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 --in_layout nhwc --out_layout nhwc --fil_layout nhwc
MIOpenDriver convfp16 -n 128 -c 512 -H 28 -W 28 -k 128 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 --in_layout nhwc --out_layout nhwc --fil_layout nhwc

Results in following Error Message
```
terminate called after throwing an instance of 'miopen::Exception'
  what():  /data/MIOpen/src/include/miopen/tensor_layout.hpp:66: mismatched layout string, unexpect char: N
Aborted (core dumped)
```
Error message needs to be corrected with proper description.

atamazov:
[off-topic] @technicalgrp89 The formatting is a bit weird. Please pay some attention to markdown formatting rules, especially to:
- https://github.github.com/gfm/#example-50
- https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code

Formatting fixed.

atamazov:
@technicalgrp89 Please confirm that everything is clear in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1311#issuecomment-1027037206 or ask questions. I'll be happy to help.

ppanchad-amd:
@technicalgrp89 Can you please test with latest ROCm 6.0.2 (HIP 6.0.32831)? If resolved, please close ticket. Thanks!

ppanchad-amd:
Hi @atamazov. Please advise if we can close this ticket. Thanks!

ppanchad-amd:
@technicalgrp89 Closing ticket due to lack of activity. Please feel free to re-open if you still have the issue with the latest ROCm. Thanks!

PR #1316, opened 2021-12-02
layout_flag_error_handle

technicalgrp89:
This is regarding the fix for below issue
https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1311 

@JehandadKhan @atamazov @junliume   @cderb please review

### DIFF
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -199,6 +199,19 @@ class ConvDriver : public Driver
     int ParseCmdLineArgs(int argc, char* argv[]) override;
     InputFlags& GetInputFlags() override { return inflags; }
 
+    //function to validate the Layout type parameters.
+    //Layout types are -In,Out,Fil etc.This function validates the
+    //layout parameter value to std (NCHW/NHWC/NCDHW/NDHWC) values,
+    //defined in MIOpen lib.
+    //layout_type - input value supplied with MIOpen driver command.
+    void ValidateLayoutInputParameters(std::string layout_type);
+
+    //Helper function to check the Layout type short names
+    //Short names are defined as I,O,f. W.r.t In/Out/fil layout
+    //types.
+    int ChkLayout_ShortName();
+
+
     int GetandSetData() override;
     std::vector<int> GetInputTensorLengthsFromCmdLine();
     std::vector<int> GetWeightTensorLengthsFromCmdLine();
@@ -426,6 +439,7 @@ bool ConvDriver<Tgpu, Tref>::IsInputTensorTransform() const
 template <typename Tgpu, typename Tref>
 int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
 {
+
     inflags.Parse(argc, argv);
 
     // try to set a default layout value for 3d conv if not specified from cmd line
@@ -433,13 +447,40 @@ int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
 
     const std::string default_layout = (spatial_dim == 2) ? "NCHW" : "NCDHW";
 
+    //inflags value is empty, default value is used
+    //if it is supplied via cmd line, check the value.
     if(inflags.GetValueStr("in_layout").empty())
+    {
         inflags.SetValue("in_layout", default_layout);
+    }
+    else
+    {
+       std::string in_layoutValue = inflags.GetValueStr("in_layout");
+       ValidateLayoutInputParameters(in_layoutValue);
+    }
+    //fil layout argument value check
     if(inflags.GetValueStr("fil_layout").empty())
+    {
         inflags.SetValue("fil_layout", default_layout);
-    if(inflags.GetValueStr("out_layout").empty())
-        inflags.SetValue("out_layout", default_layout);
+    }
+    else
+    {
+       std::string fil_layoutValue = inflags.GetValueStr("fil_layout");
+       ValidateLayoutInputParameters(fil_layoutValue);
 
+    }
+    //out layout argument check
+    if(inflags.GetValueStr("out_layout").empty())
+    {
+        inflags.SetValue("out_layout", default_layout);
+    }
+    else
+    {
+       std::string out_layoutValue = inflags.GetValueStr("out_layout");
+       ValidateLayoutInputParameters(out_layoutValue);
+
+
+    }
     num_iterations = inflags.GetValueInt("iter");
     if(num_iterations < 1)
     {
@@ -489,6 +530,53 @@ int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
 
     return 0;
 }
+
+
+template <typename Tgpu, typename Tref>
+void  ConvDriver<Tgpu, Tref>::ValidateLayoutInputParameters(std::string layout_value)
+{
+     if ((ChkLayout_ShortName()))
+     {
+        std::cerr << " Invalid Layout Short Name = " << ChkLayout_ShortName() << std::endl;
+        exit(EXIT_FAILURE);
+     }
+     else
+     {
+        if((layout_value.compare("NCHW") == 0) ||
+          (layout_value.compare("NHWC") == 0) ||
+          (layout_value.compare("NCDHW") == 0) ||
+          (layout_value.compare("NDHWC") == 0))
+          {
+            //do nothing,Values are matching as defined in Lib.
+          }
+          else
+          {
+            std::cerr << "Invalid Layout Parameter Value - " << layout_value <<  std::endl;
+            exit(EXIT_FAILURE);
+          }
+    }
+}
+
+
+template <typename Tgpu, typename Tref>
+int ConvDriver<Tgpu, Tref>::ChkLayout_ShortName()
+{
+    //check for short name of layout type
+    if ( (inflags.FindShortName("in_layout") == 'I') && 
+         (inflags.FindShortName("out_layout") == 'O') &&
+         (inflags.FindShortName("fil_layout") == 'f'))
+    {
+         //do noting
+         //found valid short names
+         return 0;
+    }
+    else
+    {
+        std::cerr << "Error:Invalid Short Name!"<< std::endl;
+        exit(EXIT_FAILURE);
+    }
+}
+
 
 template <typename Tgpu, typename Tref>
 int ConvDriver<Tgpu, Tref>::GetandSetData()
--- a/src/include/miopen/tensor_layout.hpp
+++ b/src/include/miopen/tensor_layout.hpp
@@ -58,8 +58,8 @@ void tensor_layout_to_strides(const std::vector<T>& len,
                        auto pos = layout.find(cur_layout_char);
                        if(pos == std::string::npos)
                        {
-                           MIOPEN_THROW(std::string("mismatched layout string, unexpect char: ")
-                                            .append(1, cur_layout_char));
+                           MIOPEN_THROW(std::string("mismatched layout string - ")
+                                       .append(layout));
                        }
                        return std::accumulate(layout.begin() + pos + 1,
                                               layout.end(),


atamazov:
[off-topic] @technicalgrp89 Sorry but please expect some delays because I am very busy with 5.0 release issues.

driver/InputFlags.cpp
```diff
@@ -125,8 +128,20 @@ void InputFlags::Parse(int argc, char* argv[])
 
             char short_name = FindShortName(long_name);
 
-            MapInputs[short_name].value = args[i + 1];
-            i++;
+            //check layout flag values
+            if ((args[i + 1]) == ("NCHW")||
```
cderb, line 132: Layout flags aren't the only argument parsed by this function. short_name (line 129) should be checked to make sure error checking is for the correct argument.
Also @JehandadKhan @atamazov are there possible layouts outside of this set?

driver/InputFlags.cpp
```diff
@@ -146,6 +161,8 @@ void InputFlags::Parse(int argc, char* argv[])
             else
             {
                 MapInputs[short_name].value = args[i + 1];
+
+                std::cout << "Input Flag:::else:value of short name " << MapInputs[short_name].value << std::endl;
```
cderb, line 165: Is this a debug print? Perhaps should clean up.

cderb (dismissed):

driver/InputFlags.cpp
technicalgrp89: Ok. I will work on this check. Thank you. @cderb

technicalgrp89: Ok. I will work on this check. Thank you @cderb

driver/InputFlags.cpp
technicalgrp89: Ok. I will check and update the code. Thank you. @cderb

technicalgrp89:
@cderb review comments added.

driver/InputFlags.cpp
```diff
@@ -124,9 +127,30 @@ void InputFlags::Parse(int argc, char* argv[])
                 Print();
 
             char short_name = FindShortName(long_name);
-
-            MapInputs[short_name].value = args[i + 1];
-            i++;
+            
+            //check for short name of in layout flags
+            if ( (short_name == 'I') || (short_name == 'O') || (short_name == 'f'))
```
atamazov, line 132: This code implicitly depends on layout option definitions.

driver/InputFlags.cpp
```diff
@@ -124,9 +127,30 @@ void InputFlags::Parse(int argc, char* argv[])
...
+            {
+              //check layout flag values
+              if ((args[i + 1]) == ("NCHW")||
```
atamazov, line 135: The driver should not limit layouts. Only library knows which layouts are supported.

driver/InputFlags.cpp
```diff
@@ -124,9 +127,30 @@ void InputFlags::Parse(int argc, char* argv[])
...
+                  (args[i + 1]) == ("NCDHW")||
+                  (args[i + 1]) == ("NDHWC"))
+              {
+                  MapInputs[short_name].value = args[i + 1];
+                  i++;
+              }
+              else
+              {
+                std::cout << "Invalid " << args[i] <<" Flag Type :: " << args[i+1] <<  std::endl;
+                MIOPEN_THROW("In/Out/Fill Layout Flags supports only  NCHW/NHWC/NCDHW/NDHWC values.");
```
atamazov, line 146: `MIOPEN_THROW` is intended for use in the library and should not be used in the external programs (like this driver)

atamazov (changes requested):
I recommend improving the message the the library throws when invalid layout is passed to it.

JehandadKhan:
@technicalgrp89  Please address the issues raised by @atamazov

driver/InputFlags.cpp
technicalgrp89: I have added these short name checks to make sure, lib is looking at correct input strings. As these are supplied by user. String value "NHCW/NCHW" etc are w.r.t short name I/O/f which are related to in layout/out layout/fill layout flags.

driver/InputFlags.cpp
technicalgrp89: Ok. I will add only Error message.

driver/InputFlags.cpp
atamazov: [Note] If you change "short name" of an option, then you'll need to change this code. But this dependence is implicit: if you forget to do that, then compilation failure will NOT happen; but the driver will work incorrectly. That is why I said that "this code implicitly depends on layout option definitions."

Actually I do not think we need these checks in the driver at all, because "The driver should not limit layouts. Only library knows which layouts are supported" (see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1316#discussion_r767329617)

driver/InputFlags.cpp
atamazov: The driver should stop working and return non-zero.

atamazov: No, please.

https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1316#discussion_r767329617:
> The driver should not limit layouts. Only library knows which layouts are supported.

https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1316#pullrequestreview-829671079:
> I recommend improving the message that the library throws when invalid layout is passed to it.

That is, let the driver to pass any layout to the library (but convert to uppercase first). If something is wrong, the library (not the driver) fires MIOPEN_THROW, which contains a diagnostic message.

If that message does not suit us, then it is necessary to improve it.

The problem in #1311 is the library says "mismatched layout string, unexpect char: N" but actually passed layout was "nchw", which confuses the user. I believe you can propose and implement much better error message in the library.

driver/InputFlags.cpp
technicalgrp89: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1316#discussion_r762338663 
short name checks are included, to accommodate the review comment.

atamazov: This is very true, but what I am asking for is total redesign of this feature, for the sake of not adding new implicit dependencies in our code (maybe at the cost of UX, but out users are engineers, right?)

driver/InputFlags.cpp
atamazov: @cderb 
> are there possible layouts outside of this set?

Of course, in the future.

driver/InputFlags.cpp
atamazov: @technicalgrp89 
>> total redesign of this feature, for the sake of not adding new implicit dependencies in our code . 

> can you explain, what all should be redesigned?

This is very simple. This PR is intended to implement a new small feature mentioned in #1311, i.e. better handling of incorrect layouts. I think that the implementation of that feature (i.e. this PR) should be redesigned.

> our intension is to check layout flag value strings. so I did not understand, let me knw what is needed. thank you

The intention is good, but current implementation introduces dependencies between the driver and the library. If the library will be modified to support more (or less) layouts, then the driver needs to be modified as well. And this dependence is not expressed programmatically, which makes it _implicit_.

Another implicit dependence is mentioned [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1316#discussion_r768093501).

Each implicit dependence means a code quality issue. Does the improved diagnostics worth two more quality issues in the code? I suspect that this is questionable, and that is why I am asking for "total redesign" (which is described [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1316#discussion_r768160664)):

> ...I believe you can propose and implement much better error message in the library.

Maybe I am wrong, but who knows? @JehandadKhan what do you think?

technicalgrp89: @JehandadKhan , let me knw . thank you

JehandadKhan: @technicalgrp89 I agree with @atamazov. There should not be such dependency between the library and the driver and separation of function and responsibility should be ensured. Lets setup a meeting to go over your PR

junliume:
@technicalgrp89 Happy New Year! Could you help to address the reviewers' comments and push this PR through? Thanks!

technicalgrp89:
I accidently, pushed the branch with out committing the changes.

technicalgrp89:
@junliume @JehandadKhan review comments are incorporated. Pls do review changes.

JehandadKhan (dismissed):

JehandadKhan (approved):

junliume (approved):
CI has passed here:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/sp%2Flayout_error_handle/9/pipeline

Please re-review

atamazov:
@technicalgrp89 In the future, please do not use force-pushing without prior communication with @junliume or me. Thanks!

driver/InputFlags.cpp
atamazov: Resolved.

driver/conv_driver.hpp
```diff
@@ -490,6 +531,53 @@ int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
...
+}
+
+
+template <typename Tgpu, typename Tref>
+int ConvDriver<Tgpu, Tref>::ChkLayout_ShortName()
+{
+    //check for short name of layout type
+    if ( (inflags.FindShortName("in_layout") == 'I') && 
+         (inflags.FindShortName("out_layout") == 'O') &&
+         (inflags.FindShortName("fil_layout") == 'f'))
```
atamazov, line 567: What is the reason of checking this? https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1316#discussion_r767329529 seems unresolved.

driver/conv_driver.hpp
```diff
@@ -426,20 +439,48 @@ bool ConvDriver<Tgpu, Tref>::IsInputTensorTransform() const
...
+    //inflags value is empty, default value is used
+    //if it is supplied via cmd line, check the value.
     if(inflags.GetValueStr("in_layout").empty())
+    {
         inflags.SetValue("in_layout", default_layout);
+    }
+    else
+    {
+       std::string in_layoutValue = inflags.GetValueStr("in_layout");
+       ValidateLayoutInputParameters(in_layoutValue);
```
atamazov, line 459: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1316#discussion_r767329617 seems unresolved

src/include/miopen/tensor_layout.hpp
```diff
@@ -58,8 +58,8 @@ void tensor_layout_to_strides(const std::vector<T>& len,
                        auto pos = layout.find(cur_layout_char);
                        if(pos == std::string::npos)
                        {
-                           MIOPEN_THROW(std::string("mismatched layout string, unexpect char: ")
-                                            .append(1, cur_layout_char));
+                           MIOPEN_THROW(std::string("mismatched layout string - ")
+                                       .append(layout));
```
atamazov, line 62: [Recommendation] Maybe `std::string("Unsupported tensor layout: )" + layout` ?

atamazov: I recommend reverting all changes except this one.

atamazov (commented):
Some comments from previous review remain unresolved.

driver/conv_driver.hpp
technicalgrp89: @atamazov - Short names are part of command line argument list to MIOpen driver. This particular check is to make sure , driver has received the valid short names. Let me knw.

driver/conv_driver.hpp
technicalgrp89: @atamazov - "in_layout" name should be renamed?

driver/conv_driver.hpp
```diff
@@ -199,6 +199,19 @@ class ConvDriver : public Driver
     int ParseCmdLineArgs(int argc, char* argv[]) override;
     InputFlags& GetInputFlags() override { return inflags; }
 
+    //function to validate the Layout type parameters.
```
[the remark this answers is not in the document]
technicalgrp89: @atamazov ,Can you please share a sample of comment , It would be easy to follow the same format in future. thanks

Issue #1319, opened 2021-12-02
HIP/ASM Naive solver ignore debug ENV


asroy:
All naive solvers are not controlled by use_asm_kernels or use_hip_kernels, for example:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/037e006c19be0d810c5cf1ced53b02e10d99702d/src/solver/conv_direct_naive_conv_fwd.cpp#L37

@JehandadKhan @atamazov  @carlushuang

atamazov:
And in some cases Naive Solvers use ASM kernels (that shall be controlled by `use_asm_kernels`):
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/037e006c19be0d810c5cf1ced53b02e10d99702d/src/solver/conv_direct_naive_conv.cpp#L86-L95

PR #1323, opened 2021-12-03
Fix: Use `use_hip_kernels` and `use_asm_kernels` in Naive convolution Solvers

carlushuang:
this fix https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1319

### DIFF
--- a/src/solver/conv_direct_naive_conv_bwd.cpp
+++ b/src/solver/conv_direct_naive_conv_bwd.cpp
@@ -40,18 +40,8 @@ bool ConvDirectNaiveConvBwd::IsApplicable(const ConvolutionContext& ctx) const
        miopen::IsDisabled(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_BWD{}))
         return false;
 
-    const auto device_name = ctx.GetStream().GetDeviceName();
-    if((device_name == "gfx906" || device_name == "gfx908") && ctx.rmv.IsV3() &&
-       ctx.IsLayoutDefault())
-    {
-        if(!ctx.use_asm_kernels)
-            return false;
-    }
-    else
-    {
-        if(!ctx.use_hip_kernels)
-            return false;
-    }
+    if(!ConvDirectNaiveConvIsApplicableByKernelType(ctx))
+        return false;
 
     if(!ctx.IsLayoutDefault() && !ctx.IsLayoutNHWC())
         return false;
--- a/src/solver/conv_direct_naive_conv_bwd.cpp
+++ b/src/solver/conv_direct_naive_conv_bwd.cpp
@@ -40,6 +40,9 @@ bool ConvDirectNaiveConvBwd::IsApplicable(const ConvolutionContext& ctx) const
        miopen::IsDisabled(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_BWD{}))
         return false;
 
+    if(!ctx.use_hip_kernels)
+        return false;
+
     if(!ctx.IsLayoutDefault() && !ctx.IsLayoutNHWC())
         return false;
 
--- a/src/solver/conv_direct_naive_conv_bwd.cpp
+++ b/src/solver/conv_direct_naive_conv_bwd.cpp
@@ -40,8 +40,18 @@ bool ConvDirectNaiveConvBwd::IsApplicable(const ConvolutionContext& ctx) const
        miopen::IsDisabled(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_BWD{}))
         return false;
 
-    if(!ctx.use_hip_kernels)
-        return false;
+    const auto device_name = ctx.GetStream().GetDeviceName();
+    if((device_name == "gfx906" || device_name == "gfx908") && ctx.rmv.IsV3() &&
+       ctx.IsLayoutDefault())
+    {
+        if(!ctx.use_asm_kernels)
+            return false;
+    }
+    else
+    {
+        if(!ctx.use_hip_kernels)
+            return false;
+    }
 
     if(!ctx.IsLayoutDefault() && !ctx.IsLayoutNHWC())
         return false;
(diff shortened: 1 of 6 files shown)


src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -40,6 +40,9 @@ bool ConvDirectNaiveConvBwd::IsApplicable(const ConvolutionContext& ctx) const
        miopen::IsDisabled(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_BWD{}))
         return false;
 
+    if(!ctx.use_hip_kernels)
```
atamazov, line 43: In some cases Naive Solvers use ASM kernels (that shall be controlled by `use_asm_kernels`):
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/037e006c19be0d810c5cf1ced53b02e10d99702d/src/solver/conv_direct_naive_conv.cpp#L86-L95

atamazov (changes requested):

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -40,6 +40,19 @@ bool ConvDirectNaiveConvBwd::IsApplicable(const ConvolutionContext& ctx) const
        miopen::IsDisabled(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_CONV_BWD{}))
         return false;
 
+    const auto device_name = ctx.GetStream().GetDeviceName();
```
atamazov, line 43: This is obvious duplicate of code. Please factor out the common code into a function and use is everywhere (including ConvDirectNaiveConvKernelFile()).

atamazov (changes requested):

src/solver/conv_direct_naive_conv_bwd.cpp
atamazov: @carlushuang Thanks for fixing this!

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: OK, done in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1323/commits/15117e6569e045f63438f668c80374777e9fac5d

atamazov: > ...(including ConvDirectNaiveConvKernelFile()).

It is sad to say, but the above is not done. Now `ConvDirectNaiveConvKernelFile()` and `ConvDirectNaiveConvIsApplicableByKernelType()` contain duplicate code.

I recommend implementing function like `ConvDirectNaiveConvIsAssemblyKernel()` and using it in both `ConvDirectNaiveConvKernelFile()` and `ConvDirectNaiveConvIsApplicableByKernelType()`.

atamazov (changes requested):
see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1323#discussion_r767334319

src/solver/conv_direct_naive_conv_bwd.cpp
carlushuang: ok

carlushuang:
@atamazov fixed in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1323/commits/4b255460ba848f3f4af1ce444272512aafcc36b4

atamazov (dismissed):
LGTM!

junliume:
@JehandadKhan test_find_db somehow failed for this PR, no obvious errors in the log except `[GetKernels] 0 kernels for key: miopenIm2d2Col "c192i28_28w5_5p1_1s1_1d1_1t1"` I tried to see if it is related to specific nodes but now it is failing on both ixt-sjc2-11 and ixt-sjc2-17.
CC: @atamazov

junliume:
> @JehandadKhan test_find_db somehow failed for this PR, no obvious errors in the log except `[GetKernels] 0 kernels for key: miopenIm2d2Col "c192i28_28w5_5p1_1s1_1d1_1t1"` I tried to see if it is related to specific nodes but now it is failing on both ixt-sjc2-11 and ixt-sjc2-17. CC: @atamazov

@carlushuang  @atamazov  this test is passing on develop branch and consistently fails with this PR

atamazov:
Yeah, that is possible since OCL and HIP kernels are disabled

atamazov:
...because this PR disables the rest of HIP solvers (Naive ones) that remain enabled even after #1329 because of #1319. But find-db test needs some of HIP solvers enabled. I'll propose the fix in a minute.

atamazov:
Just after this:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/53f34c75c4bdf77b52fb75646f14b49d25fea06f/test/find_db.cpp#L231
add
```C++
#if WORKAROUND_SWDEV_292187
    setenv("MIOPEN_DEBUG_HIP_KERNELS", "1", 1);
    setenv("MIOPEN_DEBUG_OPENCL_CONVOLUTIONS", "1", 1);
#endif
```
You may also need to `#include <miopen/target_properties.hpp>` for the definition of WORKAROUND_SWDEV_292187.

atamazov:
:warning: This PR is expected to be NFC. But after merging #1329 (that adds the "_MIOPEN_DEBUG_HIP_KERNELS=0_" workaround for Navi21) this is not so.

Now the side effect of this PR is that it disables Naive Solvers, and we may encounter a number of unexpected issues later (the test_find_db is just an early sign).

It seems like we need to replace the "_MIOPEN_DEBUG_HIP_KERNELS=0_" workaround from #1329 with disabling HIP solvers on individual basis (and this way keep Naive Solvers enabled). I recommend implementing this right here.

After that, the W/A in test_find_db won't be needed.

atamazov:
Actually it may be even better to this PR on hold. I am going to investigate SWDEV-292187, narrow the existing W/A (from #1329) and create a new PR. At the very beginning of that work, I'll need HIP solvers individually disabled (instead of global switch). This is the same change as I just proposed to do in this PR. So let's save efforts, and wait several days.

atamazov (changes requested):
Block merging due to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1323#issuecomment-992981185

atamazov:
#1353 will unblock this. After merging it and updating this branch from `develop`, 15d695b3c8647bf05b59482adfea92e738b4ba2d and 9b8a200fdf038465ec011630e52b693f1b5251aa will not be necessary and should be reverted.

atamazov:
Now blocked only by #1398.

atamazov (approved):
LGTM

Issue #1317, opened 2021-12-02
 [OCL][MI100][MI200] iGemm ASM GTC XDLOPS failures with OCL backend (Staging 95b58f72f)


atamazov:
`make check` convolution failures. Preconditions:
- Mainline build 9148
- OpenCL backend
- NCHW configs
- MI200 (possibly MI100 as well -- not tested)

:warning: GPU is dead after the failure (`sudo cat /sys/kernel/debug/dri/<GPU_NUM>/amdgpu_gpu_recover` to recover).

```
../bin/test_conv2d --float --cmode conv --pmode default --group-count 3 --input 16 9 224 224 --weights 63 3 3 3 --batch_size 16 --input_channels 9 --output_channels 63 --spatial_dim_elements 224 224 --filter_dims 3 3 --pads_strides_dilations 1 1 1 1 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW
MIOpen(OpenCL): Info [ForwardGetWorkSpaceSize]
MIOpen(OpenCL): Info [GetForwardSolutions]
MIOpen(OpenCL): Info [HeuristicInit] 32,256,4,32,64,4,16,64,16,0,1,8
MIOpen(OpenCL): Info [FindConvFwdAlgorithm] requestAlgoCount = 1, workspace = 231217812
MIOpen(OpenCL): Info [GetForwardSolutions]
MIOpen(OpenCL): Info [TryLoad] Find-db regenerating.
MIOpen(OpenCL): Info [FindSolutionImpl] ConvBinWinogradRxSf3x2
MIOpen(OpenCL): Info [FindSolutionImpl] Perf Db: record not found for: ConvBinWinogradRxSf3x2
MIOpen(OpenCL): Info [GetPerformanceConfig] 34
MIOpen(OpenCL): Info [FindSolutionImpl] ConvBinWinogradRxSf2x3
MIOpen(OpenCL): Info [FindSolutionImpl] Perf Db: record not found for: ConvBinWinogradRxSf2x3
MIOpen(OpenCL): Info [GetPerformanceConfig] 34
MIOpen(OpenCL): Info [FindSolutionImpl] GemmFwdRest (not searchable)
MIOpen(OpenCL): Info [FindSolutionImpl] ConvDirectNaiveConvFwd (not searchable)
[...]

PR #1324, opened 2021-12-03
[NCHW][SWDEV-312112] fix local buffer alignment bug on transpose+nhwc kernel's solver

shaojiewang:
root cause fix for swdev312112:
[nchw]fix local buffer alignment bug on transpose+nhwc kernel's solver.

_added by carlus:_

````
original buffer       workspace
inp NCHW fp16  --> 1# inp NHWC fp16 buffer
wei NCHW fp16  --> 2# wei NHWC fp16 buffer
out NCHW fp16  --> 3# out NHWC fp16 buffer
                   4# out NHWC fp32 buffer  <- bug here
````
Above is what we are implementing in fwd when input is NCHW but use NHWC kernel with transpose. Basically there should be 4 buffers in workspace, where the first 3 buffers are for transpose the original inp/wei/out buffer. the last one is for some kernel that need use atomic fp32 to accumulate the result, then cast back to fp16. We always allocate a single workspace, and keep an offset inside each of the buffer.
So the problem here is the offset for 4# may not be aligned to multiple of 4 byte, since the previous 3 buffers are fp16. If 4# buffer are not aligned to 4 byte, fp32 atomic will crash.
This PR tries to solve this problem by adding some extra byte to make sure 4# offset is aligned to multiple of 4.

### DIFF
--- a/src/conv/invokers/impl_gemm_dynamic.cpp
+++ b/src/conv/invokers/impl_gemm_dynamic.cpp
@@ -579,7 +579,7 @@ InvokerFactory MakeImplGemmDynamicForwardXdlopsNHWCInvokerFactory(
             trans_output_idx = idx++;
     }
 
-    const size_t cast_offset = is_nchw ? (trans_output_offset + trans_output_size) : 0;
+    const size_t cast_offset = is_nchw ? (((trans_output_offset + trans_output_size + 3) >> 2) << 2) : 0;
     const size_t cast_size   = need_cast ? miopen::GetTypeSize(miopenFloat) * n * k * ho * wo : 0;
 
     const int kID_trans_start = isGfx90aFp16altSupport ? 2 : 1;
@@ -886,7 +886,7 @@ InvokerFactory MakeImplGemmDynamicBackwardDataXdlopsNHWCInvokerFactory(
             trans_output_idx = idx++;
     }
 
-    const size_t cast_offset = is_nchw ? (trans_output_offset + trans_output_size) : 0;
+    const size_t cast_offset = is_nchw ? (((trans_output_offset + trans_output_size + 3) >> 2) << 2) : 0;
     const size_t cast_size   = need_cast ? miopen::GetTypeSize(miopenFloat) * n * c * hi * wi : 0;
 
     const int kID_trans_start = isGfx90aFp16altSupport ? 2 : 1;
--- a/src/conv/invokers/impl_gemm_dynamic.cpp
+++ b/src/conv/invokers/impl_gemm_dynamic.cpp
@@ -579,6 +579,7 @@ InvokerFactory MakeImplGemmDynamicForwardXdlopsNHWCInvokerFactory(
             trans_output_idx = idx++;
     }
 
+    // 4 bytes alignment to do atomic add
     const size_t cast_offset = is_nchw ? (((trans_output_offset + trans_output_size + 3) >> 2) << 2) : 0;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/conv/invokers/impl_gemm_dynamic.cpp [hunk 1]
```diff
@@ -579,7 +579,7 @@ InvokerFactory MakeImplGemmDynamicForwardXdlopsNHWCInvokerFactory(
             trans_output_idx = idx++;
     }
 
-    const size_t cast_offset = is_nchw ? (trans_output_offset + trans_output_size) : 0;
+    const size_t cast_offset = is_nchw ? (((trans_output_offset + trans_output_size + 3) >> 2) << 2) : 0;
```
carlushuang, line 582: I think we need add some comments here, since this is quite hacky

shaojiewang: Right. Added.

junliume:
I can confirm that the specific command as listed in [SWDEV-312112] is fixed by this PR.

Meanwhile, we still get RuntimeError: test_nn.py failed! FAILED (failures=13, skipped=78, expected failures=4)

atamazov:
This seems Ok as a hotfix. However we should hide details in some suitable abstraction instead hacky code here and there. In this specific case, I can propose some class that can represent the workspace that logically the consists of two sub-buffers, each with its own size and alignment. Very basic example:
```C++
WorkspaceBuf2Traits wt(size_0, size_1, alignment_1);
...
wt.GetSize(); // total workspace size
wt.GetOffset1(); // offset of subbuffer 1
```

atamazov (approved):
Conditionally approved.

atamazov:
A complicated example of handling sub-buffers can be found in `ConvWinograd3x3MultipassWrW`. @shurale-nkn Your resommendations?

junliume (approved):
Post merge issue has been raised: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1326

atamazov:
@shaojiewang I assumed that #1324 would resolve #1317 and tried it. Unfortunately this is not so. It may worth double-checking #1324 -- maybe some of the fixes are still missing (and HIP is tolerant to these, but OCL doesn't)?

shaojiewang:
> @shaojiewang I assumed that #1324 would resolve #1317 and tried it. Unfortunately this is not so. It may worth double-checking #1324 -- maybe some of the fixes are still missing (and HIP is tolerant to these, but OCL doesn't)?

Yes, I'll check it ASAP.

shaojiewang:
> @shaojiewang I assumed that #1324 would resolve #1317 and tried it. Unfortunately this is not so. It may worth double-checking #1324 -- maybe some of the fixes are still missing (and HIP is tolerant to these, but OCL doesn't)?

AFAICS, OCL backend needs a large alignment(128bytes). I'll prepare another PR to solve both #1317 and #1326

atamazov:
@shaojiewang 
>  I'll prepare another PR to solve both #1317 and #1326

Good! Just in case -- the OCL problem is not of blocker priority anymore, so we have time))

atamazov:
The fins-db NCHW records for
- ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC
- ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC
- ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC
must be refreshed.

PR #1333, opened 2021-12-08
[tests] Fix the most of the tests to return well defined 0 on success.

atamazov:
This should kill the false `make check` failures when the tests actually pass!

### DIFF
--- a/test/w_supertensor.cpp
+++ b/test/w_supertensor.cpp
@@ -592,4 +592,8 @@ struct superTensorTest : test_driver
     }
 };
 
-int main(int argc, const char* argv[]) { test_drive<superTensorTest>(argc, argv); }
+int main(int argc, const char* argv[])
+{
+    test_drive<superTensorTest>(argc, argv);
+    return 0;
+}
--- a/test/activation.cpp
+++ b/test/activation.cpp
@@ -315,4 +315,8 @@ struct activation_driver : test_driver
     }
 };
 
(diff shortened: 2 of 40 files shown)


atamazov:
@pfultz2 Well, it seems like the most recent shown config has quietly failed... How this is possible...?

atamazov:
@asroy Maybe the reason is out of memory (Navi21)?

test/w_supertensor.cpp
```diff
@@ -592,4 +592,8 @@ struct superTensorTest : test_driver
     }
 };
 
-int main(int argc, const char* argv[]) { test_drive<superTensorTest>(argc, argv); }
+int main(int argc, const char* argv[])
+{
+    test_drive<superTensorTest>(argc, argv);
+    return 0;
```
pfultz2, line 598: I dont think this is necessary, `main()` should return `0` without needing an explicit return statement, From [cppreference](https://en.cppreference.com/w/cpp/language/main_function):

> The body of the main function does not need to contain the return statement: if control reaches the end of main without encountering a return statement, the effect is that of executing return 0;.

pfultz2:
> Well, it seems like the most recent shown config has quietly failed... How this is possible...?

Something is calling `abort()` causing it to exit early. Adding the `return 0` no-ops wont do anything.

test/w_supertensor.cpp
atamazov: Never thought about this ;)

atamazov:
@pfultz2 Yep, thanks for help))

PR #1243, opened 2021-10-27
[HOTFIX][COMGR] W/A for SWDEV-308265 "Precompiler headers are broken: __hipGetPCH() is not available.." etc

atamazov:
Excerpts from the `hiprtc` branch.

- W/A for [SWDEV-308265 "[COMGR] Precompiler headers are broken: __hipGetPCH() is not available.."](https://ontrack-internal.amd.com/browse/SWDEV-308265)
- Related by-products:
  - [tests] Removed `WORKAROUND_COMGR_WARNING_ISSUES` (no issue in 4.5 RC and in Mainline 8733)
  - [COMGR] Quality fix: Return binary size 0 when "get binary" fails.
  - `MIOPEN_DEBUG_COMGR_HIP_PCH_ENFORCE`: Removed possibility to enforce use of HIP PCH.
  - [Jenkinsfile] BUILD_SMOKE_FP32_AUX1 split to _FP32 and _AUX1

PR #1339, opened 2021-12-11
[HIP][COMGR] Re-enable PCH starting from 5.0 (SWDEV-308265 is fixed in ROCClr)

atamazov:
This switches off the W/A for [SWDEV-308265 "[COMGR] Precompiler headers are broken: __hipGetPCH() is not available.."](https://ontrack-internal.amd.com/browse/SWDEV-308265) introduced in #1243 starting from HIP version 5.0.0.

:warning: The fix in ROCClr present in Mainline since build 9292. However, the build reports HIP version 4.4.21493. In order to test PCH functionality in the Mainline, it is necessary to provide two additional options to MIOpen's CMake command line:
```
... -DMIOPEN_OVERRIDE_HIP_VERSION_MAJOR=5 -DMIOPEN_OVERRIDE_HIP_VERSION_MINOR=0 ...
```
Locally tested.

:warning: Similar PR needs to be prepared for 4.5.x (if 4.5.x with fix for missing `__hipGetPCH()` will be released).

### DIFF
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -106,9 +106,9 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_SRAM_EDC_DISABLED)
 #endif
 #endif
 
-// Temporary workaround for SWDEV-308265.
-// '__hipGetPCH' is not available since 4.4
-#if HIP_SUPPORTS_PCH && (HIP_PACKAGE_VERSION_FLAT > 4004000000)
+// '__hipGetPCH' is not available in [4.4, 5.0). See SWDEV-308265.
+#if HIP_SUPPORTS_PCH && (HIP_PACKAGE_VERSION_FLAT >= 4004000000) && \
+    (HIP_PACKAGE_VERSION_FLAT < 5000000000)
 #undef HIP_SUPPORTS_PCH
 #define HIP_SUPPORTS_PCH 0
 #endif


junliume:
@sunway513 should we cherry-pick it into 4.5.x as @atamazov  has mentioned above?

junliume (approved):

atamazov:
@junliume This will only be required if a decision is made to release 4.5.3. Or if a decision is made to release PyTorch Wheels based on 4.5.2 with a fixed ROCClr. Either of these decisions require participation/approval of release managers, I guess.

src/comgr.cpp
```diff
@@ -106,9 +106,9 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_SRAM_EDC_DISABLED)
 #endif
 #endif
 
-// Temporary workaround for SWDEV-308265.
-// '__hipGetPCH' is not available since 4.4
-#if HIP_SUPPORTS_PCH && (HIP_PACKAGE_VERSION_FLAT > 4004000000)
+// '__hipGetPCH' is not available in [4.4, 5.0). See SWDEV-308265.
+#if HIP_SUPPORTS_PCH && (HIP_PACKAGE_VERSION_FLAT >= 4004000000) && \
+    (HIP_PACKAGE_VERSION_FLAT < 5000000000)
```
atamazov, line 111: @junliume 
> @sunway513 should we cherry-pick it into 4.5.x as @atamazov has mentioned above?

Simple cherry-picking won't work. We'll need to also change 
```
(HIP_PACKAGE_VERSION_FLAT < 5000000000)
```
to
```
(HIP_PACKAGE_VERSION_FLAT < 4005<patch>)
```
where `<patch>` should be taken from HIP version of 4.5.x that contains fixed ROCClr.

sunway513:
Let's move the discussion to internal chats, thanks.

Issue #1322, opened 2021-12-03
MIOpen Build in Docker Image - tons of -Wreserved-identifier warnings


junliume:
While Building MIOpen within dockers, I have met tons of warnings caused by

`warning: identifier '_***' is reserved because it starts with '_' followed by a capital letter [-Wreserved-identifier]`

Is there a way to ~~disable~~ **properly fix** them?

UPDATE by @atamazov: To reproduce, pull and run one of the latest HIP mainline builds, e.g. 9148. Install MIOpen's dependencies (of not there already), and build MIOpen with HIP backend.

:warning: Tons of warnings can make your ssh console unresponsive so better redirect stderr and stdout to a file.

carlushuang:
seems similiar to https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1314
Is all the compiler bug is because rocm-5.0, the clang++ compiler is upgraded?

atamazov:
Those are valid warnings due to usage of reserved identifiers in MIOpen, e.g. addkernels produce them. Of course we can disable the warning, but let's make a proper fix.

atamazov:
Maybe @averinevg can you fix these if you have time?

averinevg:
@atamazov Ok, I'll see how to fix it

atamazov:
@averinevg :+1:

atamazov:
@averinevg The topmost comment updated with reproduce instructions. Details by mail or voice.

junliume:
@averinevg PR #1340 and #1341 fixed quite a lot of these issues while still some more are observed:
an easy way is to search for `NOLINT (bugprone-reserved-identifier)` within the codebase.
also ref: @shurale-nkn do you know why these NOLINT instructions are not respected and are still generating warnings?

atamazov:
@averinevg `__` is a replacement for ellipsis. The function answers the question: "Is parameter belongs to the sequence: {1,4,8,12,16...32}". Just use common sense and invent a new name. For ex, you can use `Is_1_4_8_12_to_32`.

averinevg:
All related PRs have been merged. Issue fixed.

atamazov:
@averinevg :1st_place_medal:

junliume:
@averinevg reopening this issue since if we run `make -j16 check` there are still tons of similar warnings :(
looks like mostly this line, but please check elsewhere too:
```
[...]

PR #1340, opened 2021-12-12
[NFC][addkernels] Fix compiler warnings [-Wreserved-identifier]

averinevg:
The problem is described in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1322

### DIFF
--- a/addkernels/addkernels.cpp
+++ b/addkernels/addkernels.cpp
@@ -104,7 +104,7 @@ void PrintHelp()
     std::cout << "           -n[o-recurse] : dont expand include files recursively. Default: off"
               << std::endl;
     std::cout << "           -m[ark-includes] : mark variables that represent include files with "
-                 "'_INC'. Default: off"
+                 "'_INCLUDE'. Default: off"
               << std::endl;
 }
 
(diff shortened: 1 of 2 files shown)


addkernels/addkernels.cpp
```diff
@@ -104,7 +104,7 @@ void PrintHelp()
     std::cout << "           -n[o-recurse] : dont expand include files recursively. Default: off"
               << std::endl;
     std::cout << "           -m[ark-includes] : mark variables that represent include files with "
-                 "'__INC'. Default: off"
+                 "'_INC'. Default: off"
```
atamazov, line 107: [Recommendation] This is fine, but maybe this:
```suggestion
                 "'_INCLUDE'. Default: off"
```
would be cleaner? Up to you.

atamazov (approved):
LGTM

addkernels/addkernels.cpp
averinevg: Changed as you suggested

atamazov (approved):
LGTM!

atamazov:
This may block migration to 4.5 in our CI.

junliume (approved):

JehandadKhan (approved):

PR #1341, opened 2021-12-12
[NFC] Reduce compiler warnings [-Wreserved-identifier]

averinevg:
The problem is described in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1322

### DIFF
--- a/src/include/miopen/mlo_internal.hpp
+++ b/src/include/miopen/mlo_internal.hpp
@@ -617,7 +617,7 @@ struct mlo_construct_pooling2D : mlo_construct_activ_lrn_pooling_common
         _pooling_method = MLO_POOLING_OP_MAX;
         _index_type     = miopenIndexUint8;
         _wsp_index      = miopenPoolingWorkspaceIndexMask;
-        NAN_option     = 0;
+        _nan_option     = 0;
     }
 
     inline void
@@ -641,7 +641,7 @@ struct mlo_construct_pooling2D : mlo_construct_activ_lrn_pooling_common
         _search_params.kernel_size_w   = windowWidth;
         _search_params.kernel_stride_h = stride_h;
         _search_params.kernel_stride_w = stride_w;
-        NAN_option                    = NAN_opt;
+        _nan_option                    = NAN_opt;
     }
 
     inline void getPoolingDescr(int& /*pooling_method*/,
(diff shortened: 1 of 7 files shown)


src/include/miopen/mlo_internal.hpp
```diff
@@ -617,7 +617,7 @@ struct mlo_construct_pooling2D : mlo_construct_activ_lrn_pooling_common
         _pooling_method = MLO_POOLING_OP_MAX;
         _index_type     = miopenIndexUint8;
         _wsp_index      = miopenPoolingWorkspaceIndexMask;
-        _NAN_option     = 0;
```
junliume, line 620: mmm ... this will change the naming convention (not saying it's a good one though IMHO) for others to keep track of it together with the (three) ones above.
@pfultz2 what would you recommend for the name of this member variable please? This is due to https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1322

atamazov: `_nan_option`

averinevg: Replaced `NAN_option` with `_nan_option`. Thanks.

atamazov (approved):
[Recommendation] Is it possible to move the guilty underscore from the beginning to the end of identifiers?

atamazov:
May block migration of CI to 4.5.

junliume (dismissed):
Another naming convention change but I guess we should do what is right and needed :)

src/include/miopen/mlo_internal.hpp
```diff
@@ -673,7 +673,7 @@ struct mlo_construct_pooling2D : mlo_construct_activ_lrn_pooling_common
     int _pooling_method;
     miopenIndexType_t _index_type;
     miopenPoolingWorkspaceIndexMode_t _wsp_index;
-    int _NAN_option; // NOLINT (bugprone-reserved-identifier)
```
junliume, line 676: @averinevg @atamazov  @shurale-nkn just curious, do we know why this NOLINT and others mentioned in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1322#issuecomment-995347664 are ignored by compiler and still generating warnings?

averinevg:
> [Recommendation] Is it possible to move the guilty underscore from the beginning to the end of identifiers?

Of course we can do it, but I don't know for what reason these underscores are there. What are they need for?

src/include/miopen/mlo_internal.hpp
atamazov: cppcheck does not recognize NOLINT

atamazov:
@averinevg 
> @atamazov Of course we can do it, but I don't know for what reason these underscores are there. It's unlikely that this is a calling convention.

I think to reduce probability of name conflicts.

atamazov (approved):

shurale-nkn (approved):
LGTM

junliume (approved):

JehandadKhan (approved):

PR #1350, opened 2021-12-17
[Quality][tidy] Add missing headers to MIOpen_source

atamazov:
Review only the 2nd commit - 96bc894860ce2211e73e25d891843c2a09758b17. The 1st one is simple sorting.

This affects tidy checks.


src/CMakeLists.txt
```diff
@@ -72,200 +72,288 @@ function(add_kernels FILE_NAME VAR_PREFIX VAR_SUFFIX KERNEL_FILES)
...
+    include/miopen/object.hpp
+    include/miopen/ocldeviceinfo.hpp
     include/miopen/oclkernel.hpp
-    include/miopen/tensor.hpp
-    include/miopen/tensor_layout.hpp
-    include/miopen/tensor_ops.hpp
+    include/miopen/op_kernel_args.hpp
+    include/miopen/par_for.hpp
+    include/miopen/perf_field.hpp
+    include/miopen/pooling
```
junliume, line 211:
```suggestion
```

Is this line accidentally added? :-P

atamazov: :/

atamazov:
@pfultz2 Could you please review this? I do not see any merits from adding headers except more work for clang-tidy.

pfultz2:
> I do not see any merits from adding headers except more work for clang-tidy.

I dont see the merits for this either(as the headers are checked when included by .cpp files). However, I dont see any issue with merging this in either.

pfultz2 (approved):

atamazov:
@pfultz2 maybe it is better to remove headers totally instead of this?

atamazov:
On hold until next promotion to Staging branch

atamazov:
> @pfultz2 maybe it is better to remove headers totally instead of this?

:ear:

DrizztDoUrden (approved):
LGTM

atamazov:
@DrizztDoUrden Thanks, but --

> > @pfultz2 maybe it is better to remove headers totally (_from CMakeFile_) instead of this?

What do you think?

pfultz2:
Yea that might be better as I don't really understand the reason to add them in the first place.

JehandadKhan:
> Yea that might be better as I don't really understand the reason to add them in the first place.

+1

junliume:
> > Yea that might be better as I don't really understand the reason to add them in the first place.
> 
> +1

Recommendation:
Let's merge this PR as it is an enhancement of current practice. Create an issue for removing all header files from tidy checks (they will be checked as parts of the files that include them anyway), and move on with resolving that issue later. :)

junliume (approved):

junliume:
Created https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1365

Issue #1359, opened 2021-12-23
[COMGR] The use of hipModuleLoadData requires some redesign


atamazov:
- [x] As a first step (workaround), #981 must be reverted.
  - Done in #1360 
- [ ] Full-blown solution TBD. More info: https://ontrack-internal.amd.com/browse/SWDEV-304151#comment-8282503

/cc @sunway513 for awareness. I do not know if this really affects COMGR build path used in PyTorch WHL.

ppanchad-amd:
@atamazov Is the 2nd task for this issue completed? If so, can we close this ticket. Thanks!

JonathanLichtnerAMD:
Closing since MIOpen is being migrated to [ROCm/rocm-libraries](https://github.com/ROCm/rocm-libraries).  Please re-open there if you believe this is still an issue.

PR #280, opened 2020-06-11
Write-through enabled ram Db

DrizztDoUrden:
This is revived PR https://github.com/AMDComputeLibraries/MLOpen/pull/1968

PR #1360, opened 2021-12-23
[WORKAROUND][COMGR] W/A for issue 1359: Use hipModuleLoad with interim file instead of hipModuleLoadData. By-product: Better names for tempfiles.

atamazov:
Quick W/A for issue #1359, almost identical to https://github.com/AMDComputeLibraries/MLOpen/pull/2501.

### DIFF
--- a/src/include/miopen/temp_file.hpp
+++ b/src/include/miopen/temp_file.hpp
@@ -38,21 +38,24 @@ class TempFile
     public:
     TempFile(const std::string& path_infix);
 
-    TempFile(TempFile&& other) noexcept : name(std::move(other.name)), dir(std::move(other.dir)) {}
+    TempFile(TempFile&& other) noexcept
+        : path_infix(std::move(other.path_infix)), dir(std::move(other.dir))
+    {
+    }
 
     TempFile& operator=(TempFile&& other) noexcept
     {
-        name = std::move(other.name);
-        dir  = std::move(other.dir);
+        path_infix = std::move(other.path_infix);
+        dir        = std::move(other.dir);
         return *this;
     }
 
-    const std::string& Name() const { return name; }
-    std::string Path() const { return (dir.path / name).string(); }
+    const std::string& GetPathInfix() const { return path_infix; }
+    std::string Path() const { return (dir.path / "file").string(); }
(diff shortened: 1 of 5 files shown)


src/temp_file.cpp
```diff
@@ -4,7 +4,7 @@
 #include <boost/filesystem.hpp>
 
 namespace miopen {
-TempFile::TempFile(const std::string& path_template) : name(path_template), dir("tmp")
+TempFile::TempFile(const std::string& path_infix) : name("file"), dir(path_infix)
```
atamazov, line 7: This changes directory name from "`miopen-tmp-...`" to "`miopen-<INFIX>-...`" which is more informative for the geeks who would like to inspect the content of the system temp directory.

atamazov:
CI passed, please review

src/include/miopen/temp_file.hpp
```diff
@@ -36,23 +36,26 @@ namespace miopen {
...
+        : path_infix(std::move(other.path_infix)), dir(std::move(other.dir))
+    {
+    }
 
     TempFile& operator=(TempFile&& other) noexcept
     {
-        name = std::move(other.name);
-        dir  = std::move(other.dir);
+        path_infix = std::move(other.path_infix);
+        dir        = std::move(other.dir);
```
pfultz2, line 49: Why is a custom assignment and move constructor necessary as there is no custom destructor in this class? Can't this be written as:

```cpp
TempFile(TempFile&& other) noexcept = default;
TempFile& operator=(TempFile&& other) noexcept = default;
```

atamazov: @pfultz2 Hmm... The default should work in this case.

@DrizztDoUrden This first appeared at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/280/files#r775664957. Any special considerations there?

atamazov: @pfultz2 Resolved

DrizztDoUrden: No, nothing special, I think.

DrizztDoUrden (approved):
LGTM

PR #1366, opened 2022-01-05
[Tuning][gfx90a] find DB update for variation

cderb:
This is a preliminary find db tuning across all configs in the tuning database.


junliume (approved):

junliume:
@cderb restarted CI with GPU target set to gfx90a. In the future, if only certain ASIC variations are updated, we can limit the CI to corresponding platforms only to save some time :)

PR #1377, opened 2022-01-14
[MI100][MI200] iGemm ASM GTC XDLOPS: Support for deterministic results (SWDEV-316705)

carlushuang:
This hope to fix [SWDEV-316705](https://ontrack-internal.amd.com/browse/SWDEV-316705)

`MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC` env var can be used to set deterministic. If `MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC` is set, then the iGemm ASM GTC XDLOPS Solvers produce only Solutions that contain deterministic kernels.

### DIFF
--- a/src/include/miopen/conv/wrw_invoke_params.hpp
+++ b/src/include/miopen/conv/wrw_invoke_params.hpp
@@ -38,18 +38,15 @@ struct WrWInvokeParams : InvokeParams
     Data_t workSpace;
     std::size_t workSpaceSize;
     bool gfx90aFp16alt;
-    bool deterministic;
 
     WrWInvokeParams(ConvWrwTensors tensors_,
                     Data_t workSpace_,
                     std::size_t workSpaceSize_,
-                    bool gfx90aFp16alt_,
-                    bool deterministic_)
+                    bool gfx90aFp16alt_)
         : tensors(tensors_),
           workSpace(workSpace_),
           workSpaceSize(workSpaceSize_),
-          gfx90aFp16alt(gfx90aFp16alt_),
-          deterministic(deterministic_)
+          gfx90aFp16alt(gfx90aFp16alt_)
     {
     }
 
@@ -57,14 +54,12 @@ struct WrWInvokeParams : InvokeParams
                     ConvWrwTensors tensors_,
                     Data_t workSpace_,
                     std::size_t workSpaceSize_,
-                    bool gfx90aFp16alt_,
-                    bool deterministic_)
+                    bool gfx90aFp16alt_)
         : InvokeParams{type_},
           tensors(tensors_),
           workSpace(workSpace_),
           workSpaceSize(workSpaceSize_),
-          gfx90aFp16alt(gfx90aFp16alt_),
-          deterministic(deterministic_)
+          gfx90aFp16alt(gfx90aFp16alt_)
     {
     }
 };
--- a/src/include/miopen/conv/wrw_invoke_params.hpp
+++ b/src/include/miopen/conv/wrw_invoke_params.hpp
@@ -38,15 +38,18 @@ struct WrWInvokeParams : InvokeParams
     Data_t workSpace;
     std::size_t workSpaceSize;
     bool gfx90aFp16alt;
+    bool deterministic;
 
     WrWInvokeParams(ConvWrwTensors tensors_,
                     Data_t workSpace_,
                     std::size_t workSpaceSize_,
-                    bool gfx90aFp16alt_)
+                    bool gfx90aFp16alt_,
+                    bool deterministic_)
         : tensors(tensors_),
           workSpace(workSpace_),
           workSpaceSize(workSpaceSize_),
-          gfx90aFp16alt(gfx90aFp16alt_)
+          gfx90aFp16alt(gfx90aFp16alt_),
+          deterministic(deterministic_)
     {
     }
 
@@ -54,12 +57,14 @@ struct WrWInvokeParams : InvokeParams
                     ConvWrwTensors tensors_,
                     Data_t workSpace_,
                     std::size_t workSpaceSize_,
-                    bool gfx90aFp16alt_)
+                    bool gfx90aFp16alt_,
+                    bool deterministic_)
         : InvokeParams{type_},
           tensors(tensors_),
           workSpace(workSpace_),
           workSpaceSize(workSpaceSize_),
-          gfx90aFp16alt(gfx90aFp16alt_)
+          gfx90aFp16alt(gfx90aFp16alt_),
+          deterministic(deterministic_)
     {
     }
 };
--- a/src/include/miopen/convolution.hpp
+++ b/src/include/miopen/convolution.hpp
@@ -44,6 +44,7 @@
 #include <unordered_map>
 
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONVOLUTION_ATTRIB_FP16_ALT_IMPL)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONVOLUTION_ATTRIB_DETERMINISTIC)
 
 namespace miopen {
 
@@ -91,6 +92,25 @@ struct ConvolutionAttribute
         inline bool GetBwd() const { return Get() != 0; } // true is the default.
         inline bool GetWrW() const { return Get() != 0; } // true is the default.
     } gfx90aFp16alt;
+
+    class deterministic
+    {
+        int value = -1;
+        friend struct ConvolutionAttribute;
+
+        public:
+        inline bool Get() const
+        {
+            int determ_value = value;
+            if(nullptr != miopen::GetStringEnv(MIOPEN_DEBUG_CONVOLUTION_ATTRIB_DETERMINISTIC{}))
+                determ_value = miopen::Value(MIOPEN_DEBUG_CONVOLUTION_ATTRIB_DETERMINISTIC{});
+#if MIOPEN_NDEBUG
+            return determ_value == 1; // false is the default.
+#else
+            return determ_value != 0; // true is the default.
+#endif
+        }
+    } deterministic;
 
     /// Tri-state attribute values:
     /// * -1: Default (attribute-specific).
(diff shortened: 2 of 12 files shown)


src/include/miopen/conv/wrw_invoke_params.hpp
```diff
@@ -38,28 +38,33 @@ struct WrWInvokeParams : InvokeParams
     Data_t workSpace;
     std::size_t workSpaceSize;
     bool gfx90aFp16alt;
+    bool deterministic;
```
atamazov, line 41: Why you would like to pass this attrib into Invoker? It's not used there except for `std::cout << ...`.

src/include/miopen/convolution.hpp
```diff
@@ -92,6 +93,25 @@ struct ConvolutionAttribute
...
+        inline bool Get() const
+        {
+            int determ_value = value;
+            if(nullptr != miopen::GetStringEnv(MIOPEN_DEBUG_CONVOLUTION_ATTRIB_DETERMINISTIC{}))
+                determ_value = miopen::Value(MIOPEN_DEBUG_CONVOLUTION_ATTRIB_DETERMINISTIC{});
+#if MIOPEN_NDEBUG
+            return determ_value == 1; // false is the default.
+#else
+            return determ_value != 0; // true is the default.
+#endif
```
atamazov, line 111: Why this is needed?

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -777,6 +777,9 @@ bool PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC::IsValid(const Convolution
         if(ctx.IsFp16() && gemm_k_global_split != 0 && vector_store != 1)
             return false;
 
+    if(ctx.conv_problem.GetConv().attribute.deterministic.Get() && gemm_k_global_split != 0)
```
atamazov, line 780: Please explain how it is expected to provide determinism.

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -995,6 +998,8 @@ ConvSolution ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC::GetSolution(
     result.construction_params[0].comp_options = opts_0.str();
     std::ostringstream msg;
 
+    msg << ", determ:" << (ctx.conv_problem.GetConv().attribute.deterministic.Get() ? "y" : "n");
```
atamazov, line 1001: Do you plan to remove this?

atamazov (changes requested):

src/include/miopen/conv/wrw_invoke_params.hpp
carlushuang: Since this is designed within ConvDescAttribute, I just following the same design with `gfx90aFp16alt`, so I passed this value to all the `InvokeParams`. Do you think this is actually needed or not?

src/include/miopen/convolution.hpp
carlushuang: I just follow this [comments](https://ontrack-internal.amd.com/browse/SWDEV-316705?focusedCommentId=8334110&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-8334110), so I give a default value in debug build or not

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
carlushuang: Actually this is to have some debug message inside solver, incase other developers want to double confirm this solver is deterministic or not

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
carlushuang: when `gemm_k_global_split` is not zero, it indicate the kernel will split gemm_k, and finally use atomic add to accumulate the result.

junliume:
@atamazov can we revisit this issue and move it out of hold status? Thanks!

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
atamazov: Thanks, but this was already clear. The question is how this check in `TuningConfig::IsValid()` ensures determinism. The answer is non-trivial, so I had to investigate the code myself:

- If the ProblemConfig is tuned, then
  - This check ensures that TuningConfig is refused when it requires non-zero `gemm_k_global_split`. Then the common machinery leverages `HeuristicInit()` to obtain the "default" TuningConfig, which shouldn't, presumably, require "atomic add" (the default TuningConfig should contain `gemm_k_global_split == 0`).
- Otherwise, `HeuristicInit()` is directly used to obtain the "default" TuningConfig.
- `HeuristicInit()` calls `IsValid()` to make sure that the result of the `for(const auto& config : GetBwdXdlopsNHWCConfigList())` loop contains `gemm_k_global_split == 0`.

However when all attempts in the loop fail (of when `((m_per_block == 0 && n_per_block == 0 && k_per_block == 0) || not_support_vector_store)` is true), `HeuristicInit()` uses the `find_with_gemm_k_pad` lambda. The lambda uses `kernel_param_list` to initialize the result. `kernel_param_list` contains non-zero initializers for `gemm_k_global_split`, and the `find_with_gemm_k_pad` lambda does not check this.

As a result, non-zero `gemm_k_global_split` can be produced.

src/include/miopen/conv/wrw_invoke_params.hpp
atamazov: Please do not pass this into Invokers.

atamazov: This is obviously not needed because you use it only for logging.

src/include/miopen/convolution.hpp
atamazov: That comment is unclear. Clarification [requested](https://ontrack-internal.amd.com/browse/SWDEV-316705?focusedCommentId=8378275&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-8378275).

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
atamazov: Please double-check the issue and provide a fix if necessary.

Please also add comment (to the source code) that explains the machinery, i.e. how the check ensures determinism.

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
atamazov: Please move this logging into GetSolution().

atamazov:
Maybe API changes are not required and some env var should be used instead (e.g. `MIOPEN_DEBUG_DETERMINISTIC`). Pending for feedback from the end user.

carlushuang:
@atamazov yes, I see the discussion... If API level change is not needed, I guess we'd better drop this PR or refactor this into only related to asm igemm fwd/bwd/wrw solver

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
carlushuang: @atamazov I think I understand your question here. Actually, there is an internal rule when we create the `kernel_param_list`. There always will be a config with `gemm_k_global_split=0`  in front of a almost the same config, with only `gemm_k_global_split=1`, so it will never choose the *=1 one. And indeed in this `HeuristicInit()` we never wish to find a config with `gemm_k_global_split=1`.
But yes, I think this need clarification or comments, or a more robust check inside the `HeuristicInit()`.

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -777,6 +777,11 @@ bool PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC::IsValid(const Convolution
         if(ctx.IsFp16() && gemm_k_global_split != 0 && vector_store != 1)
             return false;
 
+    // we always have 2 configs that are the same expect gemm_k_global_split=1 and 0
+    // so it's safe to disable the one with gemm_k_global_split=1 here
+    if(miopen::IsEnabled(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC{}) && gemm_k_global_split != 0)
```
atamazov, line 782: [Recommendation] Please fully explain the machinery in comments. You can re-use text from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1377#discussion_r790292417 and from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1377#discussion_r790705199. Otherwise I'll need to open a post-merge issue.

atamazov (dismissed):
Good to go as a hotfix, but some clarification is required.

atamazov:
In the PR description:

> by setting MIOPEN_FIND_MODE=1, MIOPEN_FIND_ENFORCE=4, all the kernel in above Jira will pick up deterministic kernel

I think this is more clear: "If `MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC` is set, then the Solver produces only Solutions that contain deterministic kernels." I.e. the effect does not depend on Search (Tuning) mode etc. Do you agree?

carlushuang:
> 

@atamazov yes, you are right. Let me refactor the PR's description

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
carlushuang: @atamazov OK I update the comments here.

atamazov:
@junliume This PR is NFC when `MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC` is not set, so nothing bad may happen in our CI. We can merge this.

atamazov:
@carlushuang Last couple of questions
- Has it been verified that setting MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC=1 has no side effects (other than a possible performance drop)?
- What do you think about testing?

PR #1378, opened 2022-01-14
[CI] Update docker and Jenkins CI

junliume:
update dockerfile only (not the base OS, ROCm etc)

Question to @jerryyin:
Would MLIR be updated with latest ROCm too?


Dockerfile
```diff
@@ -16,10 +16,10 @@ RUN dpkg --add-architecture i386
 
 RUN if [ "$USE_MLIR" = "ON" ] ; \
         then export ROCM_APT_VER=.apt_4.3.1;\
```
jerryyin, line 18: Please help also bump the MLIR ROCm dependency as well.

There's nothing I'm aware of in MLIR that is incompatible with latest ROCm. Thanks!

atamazov:
Let's continue with this. Even if we skip 4.5.2 and upgrade right to 5.0, we'll need the fixes from this PR. What I recommend:
1. Wait for #1407
1. Update from develop
1. Revert all new warning suppressions
1. Continue development of this PR

Issue #1362, opened 2021-12-28
Pooling API problems


DrizztDoUrden:
Right now, we have several problems with the current API design. I would post the as comments bellow to ease later referencing as suggested by @atamazov.

DrizztDoUrden:
X tensor descriptor, handle and direction are missing in GWSS. This leads to the following issues:
- It is impossible to properly reuse solver methods.
- We imply that x and y buffers have exactly same descriptors.
- Handle is required to properly fetch memory alignment from GPU to calculate offset of transpose regions in the workspace.

DrizztDoUrden:
Some arguments of forward/backward functions are not used and generally are misleading, e.g. #1357.

DrizztDoUrden:
Mask/indices in the workspace are interpreted as kind of a buffer everywhere but is positioned in the workspace. This leads to complications in the transpose branch but is avoidable as a problem to some extent.

ppanchad-amd:
@DrizztDoUrden Is this fixed with latest ROCm 6.0.2 (HIP 6.0.32831)? Thanks!

DrizztDoUrden:
@ppanchad-amd No and I am not sure we want to fix it in foreseeable future. This would be a breaking API change and would require at the very least significant value to do, like a massive work on pooling that would entirely support find 1.0 planned soon enough. The issue exists mostly for documentation reasons so that people could easily find it if they have described problems and confirm their susception without wasting much time reading through every related bit of code as I did when I was reimplementing its internals.

JonathanLichtnerAMD:
Closing since MIOpen is being migrated to [ROCm/rocm-libraries](https://github.com/ROCm/rocm-libraries).  Please re-open there if you believe this is still an issue.

PR #1381, opened 2022-01-18
Pooling trasposed forward solvers and transposed solver core

DrizztDoUrden:
This implements a universal transpose kernel as a stub. It is naive, straightforward, and slow, but theoretically transposes any valid stride configuration to any other. The universal transposing solver allows developers to override it or use different transpose kernels for different layout configurations.

The implementation makes several assumptions:
- For OpenCL subbuffer alignment is assumed to be `(0x400 / 8) * 2` bytes for safety margin when handle is unavailable (GWSS). AFAIK GPUs report 0x400 bits. I have decided to double that for a possible future rise of the value. The assumption is necessary due to API flaw described here: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1362#issuecomment-1002013287.
- The same API flaw leads to an assumption that `x` and `y` have same layouts in GWSS. This may lead to insufficient workspace size for the transpose.


junliume:
@DrizztDoUrden could you select a few labels so that reviewers can know its urgency level? The CI is close to pass, I had to restart a few times. Thanks!

src/subbuffers.cpp
```diff
@@ -0,0 +1,52 @@
...
+#if MIOPEN_BACKEND_OPENCL
+#include <miopen/handle.hpp>
+#include <miopen/ocldeviceinfo.hpp>
+#endif
+
+#include <tuple>
+
+namespace miopen {
+
+std::size_t GetSubbufferAlignment(const miopen::Handle* handle)
```
shaojiewang, line 39: Could it be possible to use ```struct MultiBufferWorkspaceTraits``` instead?
ref: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/54ac76851bfaedf8321f8b8ec0a3c9c74bb48f33/src/include/miopen/buffer_info.hpp#L311

src/kernels/UniversalTranspose.cl
```diff
@@ -0,0 +1,81 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include "float_types.h"
+
+__kernel void UniversalTranspose(const __global _FLOAT* in,
```
shaojiewang, line 29: Could ```_FLOAT ``` be either ```float``` or ```half```? I think this kernel might also be applicable for ```double``` and ```int/short/char/```. Could it be suitable for all datatype?

shaojiewang (commented):
@DrizztDoUrden Thanks for formatting the ```asm igemm``` code. Almost good for me.

DrizztDoUrden:
Igemm got formatted by accident. I am not sure, why have my format script touched it, I thought, I have configured it to only touch files changed from baseline. I guess that is due to merges. I could revert that if you would rather see it in a separate PR.

src/kernels/UniversalTranspose.cl
DrizztDoUrden: I guess so. I have just used a common header: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/kernels/float_types.h. I am not a pro in kernel development, but I could probably do that. Also, this kernel is kind of a stub. It is not performant, it just allows us to use the API for previously unsupported configurations and to implement a performant specific solution by just making a tiny, specialized transpose pseudo-solver, e.g. NCHW to NHWC.
It would probably make more sense for someone more experienced to look at the kernel and update it if needed. Maybe in a separate PR as it is not mandatory for other work to do with this code.

src/subbuffers.cpp
DrizztDoUrden: The `GetSubbufferAlignment` returns the hardware constant of subbuffer alignment if you pass a handle and a constant expected value otherwise. The reason is the following: OpenCL has a meaningful value there, meaning that you have to align a subbuffer when it is being created unless you want a runtime error. Handle is optional here due to the pooling API flaw I have mentioned in the PR descriptions.

averinevg (commented):
I'm a bit distracted by the large number of formatting changes. Some of them are optional in my opinion. Do we have a coding style for this library?

DrizztDoUrden:
I will revert unrelated formatting tomorrow.

src/subbuffers.cpp
shaojiewang: OK, got you.

src/kernels/UniversalTranspose.cl
shaojiewang: Agreed, maybe in a separate PR.

DrizztDoUrden:
wait, what did just happen in that last merge :/

DrizztDoUrden:
Awsum, my local copy got messed up :/
`Unlink of file '.git/objects/pack/pack-00921996ba0659457a17dd2a91054a5c8f0b7524.idx' failed`
I guess, I'll have to fix it again.

DrizztDoUrden:
Done

shaojiewang (approved):

averinevg (approved):

src/ocl/handleocl.cpp
```diff
@@ -1,4 +1,4 @@
-/*******************************************************************************
+﻿/*******************************************************************************
```
atamazov, line 1: Broken file encoding

PR #1353, opened 2021-12-19
[WORKAROUND][Navi21] Narrow down W/A for SWDEV-292187

atamazov:
WORKAROUND_SWDEV_292187 has been introduced in #1329.
- This change makes it more specific to SWDEV-292187:
  - Removes global disabling of HIP and OCL kernels for Navi21.
  - Only Solvers that used in the failing test remain disabled.
  - The W/A is applied to FP16 convolutions only.
- **_The goal of this PR is minimizing the performance impact in 5.0_**
  - FP32 and BF16 convolutions are not affected by the W/A anymore.
  - The overall effect of W/A for FP16 convolutions:
```bash
# FP16 && Navi21 only:
export MIOPEN_DEBUG_CONV_DIRECT_OCL_WRW2=0
export MIOPEN_DEBUG_CONV_DIRECT_OCL_WRW53=0
export MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW=0
export MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_FWD_V4R1=0
export MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_V4R1=0
```

~:information_source: This is the 1st part of narrowing down the W/A. It changes the scope from _"all possible OCL and HIP solvers"_ to "_OCL and HIP solvers used in the failing test"_.~

~:information_source: The 2nd part is planned: its goal is to find the culprit solvers, re-enable the rest AND to open the Solver-specific ticket(s) for further triage/fixing.~

PR #1390, opened 2022-01-25
[5.0][Navi21] Remove W/A for SWDEV-292187

atamazov:
**_This is alternative for #1389_** which removes W/A for SWDEV-292187 completely instead of narrowing it, thus restoring both FP16 and FP32 performance for Navi21.

Note on process: If this PR is to be included into 5.0, then **_it should be also ported to the recent develop and merged separately_**. #1353 won't needed in develop in such a case.


atamazov:
On hold until CQE testing is done.

junliume (approved):

atamazov:
Testing (by Gowtham from CQE team) succeeded.

atamazov:
> Note on process: If this PR is to be included into 5.0, then **_it should be also ported to the recent develop and merged separately_**. #1353 won't needed in develop in such a case.

TODO: porting to develop when release staging testing is done.

PR #1392, opened 2022-01-27
[gfx90a68] Removing ConvBinWinogradRxSf3x2 from pdb

cderb:
Removing ConvBinWinogradRxSf3x2 from mi200 104


atamazov (approved):

atamazov:
Reasoning:

> MIOpen(HIP): Warning [FindSolutionImpl] Invalid config loaded from Perf Db: ConvBinWinogradRxSf3x2: 109.
> ...it is seen on system with MI210 (but our other systems uses MI250x)

MI210 has 104 CUs. Winograd's tuning config value should not exceed the number of available CUs.T he reason of the warning is that tuning configs were obtained on MI250 and then copied to MI210's db. As a fix, I recommend removing Winograd's tuning configs from MI210's db. The HeuristicInit of Winograd is very capable so I do not expect any performance drops.

junliume (approved):

PR #1394, opened 2022-01-31
Replace deprecated HIP API (SWDEV-318154)

averinevg:
`hipCtxGetCurrent()` and `hipCtxSetCurrent()` were added to better support tensorflow which was using the context API. Since HIP context API were removed from tensorflow codebase (https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/pull/58), these calls can be removed.

Changes:
- `hipCtxGetCurrent()` and `hipCtxSetCurrent()` removed
- `hipHccModuleLaunchKernel()` replaced with `hipExtModuleLaunchKernel()`
- `-Wno-deprecated-declarations` removed
- `MIOPEN_WORKAROUND_ROCM_COMPILER_SUPPORT_ISSUE_30` for https://github.com/RadeonOpenCompute/ROCm-CompilerSupport/issues/30 is not necessary since ROCm 4.3.


src/hip/handlehip.cpp
```diff
@@ -66,24 +66,20 @@ namespace miopen {
 
 #if MIOPEN_WORKAROUND_ROCM_COMPILER_SUPPORT_ISSUE_30
 namespace {
-void toCallHipInit() __attribute__((constructor(1000)));
-void toCallHipInit() { hipInit(0); }
+static void toCallHipInit() __attribute__((constructor(1000)));
```
atamazov, line 69: This is not necessary in unnamed namespace.

src/hip/handlehip.cpp
```diff
@@ -66,24 +66,20 @@ namespace miopen {
...
+static void toCallHipInit() { hipInit(0); }
 } // namespace
 #endif
 
 // Get current context
 // We leak resources for now as there is no hipCtxRetain API
-hipCtx_t get_ctx()
+static hipCtx_t get_ctx()
```
atamazov, line 76: Please define local functions in unnamed namespace (instead of static)

atamazov (changes requested):

src/hip/handlehip.cpp
atamazov: Or keep this as is.

shurale-nkn (approved):
LGTM

cmake/EnableCompilerWarnings.cmake
```diff
@@ -102,7 +102,7 @@ else()
             endif()
             list(APPEND CMAKE_COMPILER_WARNINGS
                 -Wno-missing-field-initializers
-                -Wno-deprecated-declarations
+                -Wno-global-constructors
```
atamazov, line 105: Hmm... Why we need this new suppression?

src/hip/handlehip.cpp
```diff
@@ -58,30 +58,11 @@
 #include <chrono>
 #include <thread>
 
-#define MIOPEN_WORKAROUND_ROCM_COMPILER_SUPPORT_ISSUE_30 (MIOPEN_USE_COMGR && BUILD_SHARED_LIBS)
```
atamazov, line 61: Is it so that we no longer need a workaround for https://github.com/RadeonOpenCompute/ROCm-CompilerSupport/issues/30 ?

atamazov (changes requested):

src/hip/handlehip.cpp
atamazov: [process] @averinevg Let me "resolve" my review comments myself (unless the case is trivial), thanks!

atamazov (approved):
LGTM!

PR #1397, opened 2022-01-31
[Jenkins] Fix "clang-format" stage

averinevg:
"Clang format" stage was fixed
Additional changes: formatting all code according to the .clang-format.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -373,8 +373,8 @@ pipeline {
                                 -o -iname \'*.hpp.in\' \
                                 -o -iname \'*.cpp.in\' \
                                 -o -iname \'*.cl\' \
-                                | grep -v 'build/' \
-                                | xargs -n 1 -P 1 -I{} -t sh -c \'clang-format-10 -style=file {} | diff - {}\'"
+                                | grep -v -E '(build/)|(install/)' \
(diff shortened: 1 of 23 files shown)


atamazov:
@averinevg Now please fix formatting. Thanks!

Jenkinsfile
```diff
@@ -374,7 +374,7 @@ pipeline {
                                 -o -iname \'*.cpp.in\' \
                                 -o -iname \'*.cl\' \
                                 | grep -v -E '(build/)|(install/)' \
-                                | xargs -n 1 -P $(nproc) -I{} -t sh -c \'clang-format-10 -style=file {} | diff - {}\'"
```
atamazov: Note that this may affect readability of the log when stage fails. I do not mind because usually fixing of formatting problems is an automatic process, i.e. we do not need to understand the details from the log. From the other hand, the speed of clang-format stage is not very important unless it is slower than tidy check stages. Decide yourself.

atamazov (approved):
LGTM!

Jenkinsfile
averinevg: Changes in line 377 were reverted in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1397/commits/751a2d765c23eead5aeec8f153e0d45c0bf72192

atamazov:
@averinevg :birthday:

PR #1398, opened 2022-01-31
[Navi21] Remove W/A for SWDEV-292187

atamazov:
This is the clone of #1390 for develop.

**_This is alternative for #1353_** which removes W/A for SWDEV-292187 completely instead of narrowing it, thus restoring both FP16 and FP32 performance for Navi21.


Kirpich30000 (approved):
LGTM

DrizztDoUrden (approved):

PR #1400, opened 2022-02-02
[Quality] Fix potential verify error with NaN - IEEE-754 comparison rule

junliume:
A lot of places:
```
    if(error > tolerance)
    {
        std::cout << "verification Failed ";
    }
```
will have issue if error contains NaN; per IEEE-754 the comparison will return FALSE.

### DIFF
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -3287,7 +3287,7 @@ int ConvDriver<Tgpu, Tref>::VerifyForward()
     if(is_fwd_igemm)
         tolerance = tolerance * 10;
 
-    if(error > tolerance)
+    if(!(error < tolerance))
     {
         std::cout << "Forward Convolution Failed: " << error << " > " << tolerance << std::endl;
         return EC_VerifyFwd;
--- a/driver/lrn_driver.hpp
+++ b/driver/lrn_driver.hpp
@@ -474,7 +474,7 @@ int LRNDriver<Tgpu, Tref>::VerifyForward()
 
     auto error           = miopen::rms_range(outhost, out);
     const Tref tolerance = 1.5e-4; // 1e-6;
-    if(error > tolerance)
+    if(!(error < tolerance))
     {
         std::cout << "Forward LRN Failed: " << error << "\n";
     }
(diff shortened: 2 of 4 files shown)


JehandadKhan (approved):

driver/conv_driver.hpp
```diff
@@ -3287,7 +3287,7 @@ int ConvDriver<Tgpu, Tref>::VerifyForward()
     if(is_fwd_igemm)
         tolerance = tolerance * 10;
 
-    if(error > tolerance)
+    if(!(error < tolerance))
```
atamazov, line 3290: [Note] The equivalent construct should have `<=`.

atamazov (commented):
@junliume Just in case -- Have you checked if this able to catch NaNs? Some compiler optimizations may transform this to `if(error >= tolerance)`.

driver/conv_driver.hpp
atamazov: No need to fix, just in case.

atamazov:
@junliume Would you like have a separate fix for BN driver?

junliume:
> @junliume Would you like have a separate fix for BN driver?

@atamazov Yes, but only in one case:
before:
```
root@ixt-rack-185:/opt/rocm-5.1.0-9514/miopen# ./bin/MIOpenDriver lrn -H 7 -W 7 -n 1 -c 1056 -N 4 -K 1 -A 1 -B 1 -m within -F 2
MIOpenDriver lrn -H 7 -W 7 -n 1 -c 1056 -N 4 -K 1 -A 1 -B 1 -m within -F 2
Backward LRN Verifies on CPU and GPU (err=-nan)
```
after:
```
root@ixt-rack-185:~/MIOpen/build# ./bin/MIOpenDriver lrn -H 7 -W 7 -n 1 -c 1056 -N 4 -K 1 -A 1 -B 1 -m within -F 2
MIOpenDriver lrn -H 7 -W 7 -n 1 -c 1056 -N 4 -K 1 -A 1 -B 1 -m within -F 2
Backward LRN Failed: -nan
```

atamazov:
> @junliume Would you like have a separate fix for BN driver?

junliume:
> > @junliume Would you like have a separate fix for BN driver?

Yikes ... I replaced all "error > tolerance" but BN driver is using "diff > tolerance" :(

TODO:
1. driver/bn_driver.hpp:
    diff > tolerance
2. driver/conv_driver.hpp:
    error_**** > tolerance

PR #1408, opened 2022-02-07
[Quality] Fix more verify error with NaNs and Inf

junliume:
Update: fixed a few extra places where the comparison might have issues.

Original Comments about the root cause:

A lot of places:
```
    if(error > tolerance)
    {
        std::cout << "verification Failed ";
    }
```
will have issue if error contains NaN; per IEEE-754 the comparison will return FALSE.

### DIFF
--- a/driver/bn_driver.hpp
+++ b/driver/bn_driver.hpp
@@ -1180,10 +1180,10 @@ int BatchNormDriver<Tgpu, Tref, Tmix>::VerifyForward()
             runningVariance_dev->FromGPU(GetStream(), runningVariance.data());
 
             auto errorRunMean = miopen::rms_range(runningMean_host, runningMean);
-            if(errorRunMean > maxrms || std::isnan(errorRunMean))
+            if(!std::isfinite(errorRunMean) || errorRunMean > maxrms)
             {
-                std::cout << "Forward train batch norm verification failed on running mean: "
-                          << errorRunMean << "\n";
+                std::cout << "Forward train batch norm verification FAILED on running mean: "
+                          << errorRunMean << std::endl;
                 anError = true;
 #if(MIO_BN_DEBUG == 1)
                 for(int i = 0; i < runningMean.size() && i < runningMean_host.size() &&
@@ -1191,7 +1191,7 @@ int BatchNormDriver<Tgpu, Tref, Tmix>::VerifyForward()
                     i++)
                 {
                     diff = fabs(Tmix(fabs(runningMean[i]) - fabs(runningMean_host[i])));
-                    if(!(diff < tolerance))
+                    if(!std::isfinite(diff) || diff > tolerance)
                     {
                         std::cout << "rm[" << i << "]: " << runningMean[i];
                         std::cout << ", rm_host[" << i << "]: " << runningMean_host[i];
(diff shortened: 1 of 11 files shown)


qianfengz:
So    **!(diff < tolerance)** is equal to   **std::isnan(diff) || std::isinf(diff) || diff >= tolerance** ?   Should we use **std::isnan(diff) || std::isinf(diff) || diff >= tolerance**  directly to be more clear ? 

**std::isnan()** and **std::isinf()** is provided by **cmath**

atamazov:
> So `!(diff < tolerance)` is equal to `std::isnan(diff) || std::isinf(diff) || diff >= tolerance`? Should we use `std::isnan(diff) || std::isinf(diff) || diff >= tolerance` directly to be more clear ?

Yes, this is better than comparison because is does not depend on compiler's optimizations. It would work even if unsafe FP math is enabled.

`std::isfinite()` is equivalent of `(std::isnan() || std::isinf())`, so I recommend using it.

driver/bn_driver.hpp
```diff
@@ -1191,7 +1191,7 @@ int BatchNormDriver<Tgpu, Tref, Tmix>::VerifyForward()
                     i++)
                 {
                     diff = fabs(Tmix(fabs(runningMean[i]) - fabs(runningMean_host[i])));
-                    if(diff > tolerance)
+                    if(!(diff < tolerance))
```
atamazov, line 1194:
```suggestion
                    if(!std::isfinite(diff) || diff > tolerance)
```
I recommend these checks everywhere.

atamazov:
In bn_driver (and possibly other places):
```C++
auto errorRunVar = miopen::rms_range(runningVariance_host, runningVariance);
if(errorRunVar > maxrms || std::isnan(errorRunVar))
```
INF is not detected, let's use `!std::isnormal()`

atamazov:
In bn_driver (and possibly other places):
```C++
std::cout << "Forward train batch norm verification failed on running mean: "
<< errorRunMean << "\n";
```
This is not good. I highly recommend using uppercase `*FAILED*` and `std::endl` instead of `\n`.

atamazov:
@junliume And big thanks for working on this!

driver/reduce_driver.hpp
```diff
@@ -493,20 +493,20 @@ int ReduceDriver<Tgpu, Tref>::VerifyForward()
...
+        std::cout << "ReduceTensor() FAILED with error = " << error
+                  << " , tolerance = " << tolerance << std::endl;
     }
     else
     {
         if(out_indices.size() > 0)
         {
             auto error2 = miopen::rms_range(outhost_indices, out_indices);
 
-            if(static_cast<float>(error2) != 0.0f)
```
junliume, line 507: @qianfengz and @atamazov  please check if this change meets your design requirements.

qianfengz, line 507: The tolerance checking could not be used with error2, since error2 is the rms_range of two vector<int> representing the outputs of indices by Host and GPU reduction.  For outputs of indices, we want the GPU result be completely the same as the Host result.  
We can use **if(!std::isfinite(error2) && static_cast<float>(error2) != 0.0f )**

atamazov (dismissed):
LGTM

atamazov (changes requested):
Please remove [.gitignore](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1408/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947) from the patch.

junliume:
> Please remove [.gitignore](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1408/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947) from the patch.

That's a local stash that got accidentally got in ... will remove promptly

atamazov:
@junliume Yes. You can use script from https://github.com/ROCmSoftwarePlatform/MIOpen/wiki/How-to-format-code before each commit.

atamazov (approved):
LGTM!

junliume:
@atamazov We may have a persistent failure on Vega20 nodes (Fp16 Hip Install All Vega20) now, probably found by this change:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/ieee754_nan/11/pipeline/324

PR #1411, opened 2022-02-08
Tensor reorder

aska-0096:
test all cases & CI check.


src/kernels/gpu_general_tensor_reorder_kernel/order.hpp
```diff
@@ -31,9 +31,9 @@ struct order
 {
     static constexpr int m_size = sizeof...(Is);
 
-    __host__ __device__ static constexpr auto size() { return m_size; }
+    __host__ __device__ static constexpr int size() { return m_size; }
```
atamazov, line 34: Is `int` enough to hold all possible sizes? It's max is `2 GiB - 1`.

Issue #1410, opened 2022-02-07
[tests] `test_reduce_test --D 1024 30528` fails


atamazov:
Leftover from #1409. This works:
```bash
$ ./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 1024,30528,1 -I 0 -N 1 -O 0 -R 0,1,2
ReduceTensor() Verifies on CPU and GPU (err=0.000094)
```
But this doesn't:
```bash
$ ./bin/test_reduce_test --scales 1 0 --CompType 1 --D 1024 30528 1 --I 0 --N 1 ---ReduceOp 0 --R 0 1 2 -v
FAILED: 0.0272532
Max diff: 851954
Mismatch at 0: 3.04087e+07 != 3.12606e+07
verify_reduce_no_indices failed
Input Tensor 1024, 30528, 1
```
The `--tolerance` option does not help.

As a result, this case was not added to the tests. **_The fix should include adding it._**

qianfengz:
MIOpenDriver and test_reduce_test depend on different methods for initializing data.  Since test_reduce_test needs supports both batch testing and interactive testing, and its batch testing has all operations covered,  its data initializing is more complicated and it is hard for test_reduce_test to specify a suitable tolerance value for any unexpected testing case.    For example, MIOpenDriver initializes all input data be in the range of (0, 1), but when operation MUL is tested,    1024 * 30528 values in (0, 1) could lead to underflow/overflow when accumulated.   The target of test_reduce_test is to ensure the batch testing works well for various operation types and reduction situation using specific tensor case.

qianfengz:
The above case can be solved by using a similar initializing data as MIOpenDriver uses and enlarge the tolerance for ADD from 80 to 800.  But do we really need to do so ?

PR #1409, opened 2022-02-07
[SWDEV-321331] Fix ONNX runtime unit tests for reductions

qianfengz:
Quick fix for [SWDEV-321331](https://ontrack-internal.amd.com/browse/SWDEV-321331). 

The fix has been verified with the following script using MIOpenDriver. 
```bash
#!/bin/bash

./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 8,2,1 -I 0 -N 1 -O 0 -R 0,1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 160,10,1 -I 0 -N 1 -O 0 -R 0,1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 7,1024,1 -I 0 -N 1 -O 0 -R 0,1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 1024,30528,1 -I 0 -N 1 -O 0 -R 0,1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 3,1,1 -I 0 -N 1 -O 0 -R 0,1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 3,1,1 -I 0 -N 1 -O 1 -R 0,1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 3,1,1 -I 1 -N 1 -O 3 -R 0,1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 3,2,1 -I 1 -N 1 -O 3 -R 1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 6,2,1 -I 0 -N 1 -O 3 -R 1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 6,2,1 -I 0 -N 1 -O 2 -R 1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 2,2,1 -I 0 -N 1 -O 0 -R 1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 4,3,1 -I 0 -N 1 -O 3 -R 1,2
./bin/MIOpenDriver reduce -A 1 -B 0 -C 1 -D 3,4,1 -I 0 -N 1 -O 3 -R 1,2
[...]

PR #1415, opened 2022-02-09
[Tests] add test_reduce_test and follow up PR #1409

qianfengz:
Resolves [Issue 1410](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1410)

Update in test/reduce_test.cpp to make the leaked testing case mentioned in [PR.1409 ](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1409) as well as [Issue 1410](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1410) pass.  The fix has passed all MIOpenDriver and test_reduce_test cases used in [PR.1409 ](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1409) as well as the case

`./bin/test_reduce_test --scales 1 0 --CompType 1 --D 1024 30528 1 --I 0 --N 1 ---ReduceOp 0 --R 0 1 2 -v`

Also, the following default test_reduce_test batch tests passed 
`./bin/test_reduce_test --all`
`./bin/test_reduce_test --all --half`
`./bin/test_reduce_test --all --double`

### DIFF
--- a/test/reduce_test.cpp
+++ b/test/reduce_test.cpp
@@ -819,7 +819,7 @@ struct reduce_driver : test_driver
             max_value = 3;
         else
             max_value =
-                miopen_type<T>{} == miopenHalf ? 13 : miopen_type<T>{} == miopenInt8 ? 127 : 17;
+                miopen_type<T>{} == miopenHalf ? 13 : miopen_type<T>{} == miopenInt8 ? 127 : 999;
 
         // default data gneration (used by MIN/MAX)
         auto gen_value = [&](auto... is) {
@@ -833,7 +833,7 @@ struct reduce_driver : test_driver
             auto rand_value = tensor_elem_gen_integer{max_value}(is...);
             auto sign_value = tensor_elem_gen_checkboard_sign{}(is...);
 
-            return (sign_value * rand_value + 1.0);
+            return (sign_value * rand_value / max_value + 0.01);
         };
 
         // Special data generation for MUL, to avoid all-zero and large accumulative error in the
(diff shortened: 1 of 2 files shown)


test/reduce_test.cpp
```diff
@@ -833,7 +833,7 @@ struct reduce_driver : test_driver
             auto rand_value = tensor_elem_gen_integer{max_value}(is...);
             auto sign_value = tensor_elem_gen_checkboard_sign{}(is...);
 
-            return (sign_value * rand_value + 1.0);
+            return (sign_value * rand_value / max_value + 0.01);
```
atamazov, line 836: Are you sure that changes in this file do not affect the quality of testing?

atamazov:
@qianfengz PR description should mention #1410 as it is intended to fix it.

test/reduce_test.cpp
qianfengz: MIOpenDriver uses values in `(0,1)` to initialize the input data,  the change here use similar method.   The previous codes use integers in `[-16, 17]` for ADD/AVG.  But I believe the accumulative error can only be reduced by using more fine-grained values.  Here the max_value is improved to `999` and use rand_value/max_value will cause more fine-grained values.  `+ 0.01` is used to avoid the mean of zero.  I could not work out better method to conquer that case,  I don't want to add `--tolerance` option, since the tolerance value in `test_reduce_test` is hard to understand, it is a multiplier used to multiply with the `std::numeric_limits<float>::epsilon` to get a final tolerance threshold. 

In general,  `test_reduce_test` is similar to other `test/*.cpp`,  which depend on the same facility and are different from `driver/*.cpp`

qianfengz:
> @qianfengz PR description should mention #1410 as it is intended to fix it.

Yes, You are right!

junliume (approved):

atamazov (approved):
LGTM!

junliume:
@qianfengz  and @atamazov 
I am currently investigating an issue related to this custom test. It is failing the CI test and my local builds on gfx1030.

```
# ./bin/test_reduce_test --scales 1 0 --CompType 1 --D 1024 30528 1 --I 0 --N 1 ---ReduceOp 0 --R 0 1 2 -v
./bin/test_reduce_test --float --D 1024 30528 1 --R 0 1 2 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0 
error: 4.70829e-05
Max diff: 14.7188
Mismatch at 0: 312599 != 312614
verify_reduce_no_indices failed
Input Tensor 1024, 30528, 1

PR #1420, opened 2022-02-10
fix a bug in order.hpp for passing ocl debug+codecov

aska-0096:


junliume:
@aska-0096 : these merges should not appear on GitHub page. Are you doing them locally and accidentally pushed them to remote? Thanks!

PR #1421, opened 2022-02-10
fix bug in order.hpp to satisfy cpp11 standard

aska-0096:


junliume:
@aska-0096 : these merges should not appear on GitHub page. Are you doing them locally and accidentally pushed them to remote? Thanks!

Issue #1423, opened 2022-02-11
[Tests][Navi21][fp16] added test_reduce_test from #1415 still fails with nan


junliume:
**[Keywords]:**
test; Navi21; reduction; fp16
**URGENCY :HIGH** since it is already in the develop branch

**[Env]:**
Navi21
ROCm 5.0
**FP16**

**[Description]:**
related to #1415 and https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1415#issuecomment-1036515269

It is failing the CI test and my local builds on gfx1030 **for FP16**.
Problematic CI: http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/develop/492/pipeline/
```
# ./bin/test_reduce_test --scales 1 0 --CompType 1 --D 1024 30528 1 --I 0 --N 1 ---ReduceOp 0 --R 0 1 2 -v
./bin/test_reduce_test --float --D 1024 30528 1 --R 0 1 2 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0 
error: 4.70829e-05
Max diff: 14.7188
Mismatch at 0: 312599 != 312614
verify_reduce_no_indices failed
Input Tensor 1024, 30528, 1
```
```
# ./bin/test_reduce_test --half --D 1024 30528 1 --R 0 1 2 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0 
./bin/test_reduce_test --half --D 1024 30528 1 --R 0 1 2 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0 
FAILED: -nan
Iteration: 0
verify_reduce_no_indices failed
Input Tensor 1024, 30528, 1
Max diff: nan
Mismatch at 0: inf != inf
Non finite number found in cpu at 0: inf
Non finite num

junliume:
Do those large tensors make sense for FP16? `HALF_ENABLED` is applied here. @qianfengz

qianfengz:
I think we could not guarantee the data we provided for testing will not cause overflowing for half_t when the tensor rank/extent is very special.   For this `--half` case, both the Host and Gpu reduction lead to overflow, and the same is true for using MIOpenDriver. Since we using a fixed method to set up data,  we could not guarantee our data is suitable for any Reduction testing case given by the user. 

[...]

PR #1425, opened 2022-02-11
[Tests][Navi21][FP16] W/A #1423 bypass large tensor reduce test

junliume:
Issue described in #1423:

We still need to fix the issue, this PR is to unlock develop branch

Hence: not reverting #1415 since it worked for FP32


junliume:
@atamazov  what do you think about this PR or we simply revert the previous one?

JehandadKhan (approved):

Issue #1431, opened 2022-02-16
[Navi21][ROCm5.0] warpSize inconsistent between different HIP kernel building methods 


junliume:
> > > @atamazov @JehandadKhan @asroy @zjing14 @qianfengz could you take a look at CK related changes? The composable kernel-related changes in this PR might have affected accuracy:
> > > ```
> > > # ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0
> > > ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0 
> > > FAILED: 0.255031
> > > Iteration: 0
> > > verify_reduce_no_indices failed
> > > Input Tensor 64, 3, 280, 81
> > > Max diff: 2.27564
> > > Mismatch at 0: 1.00436 != 2.00873
> > > ```
> > 
> > 
> > Very strange! I could not reproduce the issue on MI100 and MI50 using hiprtc branch. From your test, the reduced result value get on Host is just half that of the GPU.
> 
> This might not be hiprtc related but instead ROCm 5.0 related. Removing the blocker for this PR for now.

This could be compiler issue.  I found that the `warpSize` used by my `DirectWarpSize` reduction kernels is 64, while the value of `handle.GetWavefrontWidth()` is 32.   My understand is that `warpSize` is provided as a constant by the compiler, while `handle.GetWavefrontWidth()` gets the warp size from the HIP Runtime device properties.  So the constant value maintained by the compiler under ROCM 5.0 is problematic on Navi.   The issue can be worked-around by passing the value of `handle.GetWavefrontWidth()` to the kernel and let the `DirectWarpWise` kernel to use it instead of the `warpSize`

_Originally posted by @qianfengz in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1237#issuecomment-1041386644_

junliume:
AFAIK Navi21 only support one warpSize. Should we enforce it for all kernels? ref to #1407 @atamazov

qianfengz:
The issue also occurred on ROCM-4.3 ROCM-4.3.1 with my testing.  But the following codes compiled with hipcc is able to output the warpSize of 32 on Navi.  
```c++
#include <hip/hip_runtime.h>
__global__ void kernel_show_warpsize()
{
[...]

PR #1436, opened 2022-02-18
Work-around Reduction issue due to inconsistent warpSize between different HIP kernel building methods

qianfengz:
This P.R is a work-around for the Reduction test issue mentioned in [1431](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1431). 

It is found that when the HIP kernel is compiled using the `comgr` methd (with `MIOPEN_USE_COMGR` enabled), the the value of HIP kernel `warpSize` builtin variable will be 64 on Navi, rather than an expected value of 32.  The inconsistent value of `warpSize` will cause wrong result while calling the `DirectWarpWise`  kernel for Reduction.  This P.R provides the work-around by passing the miopen runtime detected wavefront size to the reduction kernel to avoid using the incorrect `warpSize`.

### DIFF
--- a/src/composable_kernel/composable_kernel/include/tensor_operation/gridwise_generic_2d_reduction_direct_warpwise.hpp
+++ b/src/composable_kernel/composable_kernel/include/tensor_operation/gridwise_generic_2d_reduction_direct_warpwise.hpp
@@ -36,6 +36,7 @@
 namespace ck {
 
 template <index_t BlockSize,
+          index_t WavefrontSize,
           typename srcDataType,
           typename dstDataType,
           typename compType,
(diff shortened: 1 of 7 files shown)


junliume:
@qianfengz @atamazov fix verified on a local Navi21 bare metal with 5.0. Cannot reproduce in current MIOpen CI.

junliume (approved):

atamazov:
@qianfengz @junliume Do we have a regression test for this issue?

junliume:
> @qianfengz @junliume 🔴 Do we have a regression test for this issue?

@atamazov I think the following test already exist and can discover this issue:
> ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0

atamazov:
@junliume 
> > ./bin/test_reduce_test --float --D 64 3 280 81 --R 0 --ReduceOp 0 --CompType 1 --N 0 --I 0 --scales 1 0

Sadly, this PR does not add this test case, and I am unable to find it in ./test/CMakeLists.txt. Maybe we shall add it to `test_reduce_custom_fp32_fp16` or to `test_reduce_custom_fp32`?

Issue #1372, opened 2022-01-12
CI upgrade plan (base driver and ROCm)


atamazov:
_Originated from https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1148#issuecomment-998255498_

I recommend updating nodes to the most recent released ROCm version and corresponding kernel driver. 5.0 is not released yet, I would either wait or use 4.5.2.

> NOTE :warning: ROCm used for testing (in the docker container) must be of the same version as the base installation; otherwise failures are possible. Using newer ROCm than base driver is allowed; but an opposite combo must be avoided. That is why this process includes updates of the Dockerfile.

We must be careful to avoid CI malfunction due to, for example, failures of static checks.

## Proposed plan:
- [ ] Devote one Navi21 node (re-label it to `rocmtest-5.0`, for example)
- [ ] Upgrade base installation of ROCm and kernel driver
- [ ] Create new "upgrade" branch (e.g. `wip-rocmtest-5.0-upgrade`). Try it to see if it is passing on that node, as is.
  - :warning: Failures are possible (see note above)
- [ ] Modify "upgrade" branch: update Dockerfile to use new ROCm version and make sure it passes all the tests
- [ ] Modify "upgrade" branch: include static checks and fix MIOpen until it passes tests
- [ ] Devote one gfx906 node, one MI100 node and one MI200 node to the CI upgrade process. Upgrade base installation of ROCm and kernel driver on these nodes.
- [ ] Update vega10, vega20, mi200 and mi200 trial branches from `wip-rocmtest-5.0-upgrade` and make sure that all tests pass.
  - [Note] All fixes from these trial nodes must be integrated back to the "upgrade" branch (e.g. `wip-rocmtest-5.0-upgrade`).
- [ ] :red_circle: Inform everyone and shutdown CI (e.g this can be done by re-labeling all the nodes with `rocmtest-5.0`).
- [ ] Upgrade remaining nodes
- [ ] Merge `wip-rocmtest-5.0-upgrade` into `develop` and make sure it passes CI.
- [ ] Merge `develop` into all trial branches and test all nodes.
- [ ] :green_circle: Ask everyone to merge `develop` into their development branches.

/cc @junliume @JehandadKhan @pfultz2 @okakarpa @jbakhrai

junliume:
@atamazov 4.5.2 in dockerfiles seem to have no problems on the current CI pipelines (while base OS ROCm are not updated):
[...]

PR #1443, opened 2022-02-23
[MLIR] Add workspace to MLIR wrw solver in case needed.

whchung:
Teach MLIR wrw solver to allocate and use a workspace if needed.

An MIIR interface would be used to let MLIR decide whether a convolution configuration really uses a workspace. Currently, a workspace is required in the following combination:

- backward weight convolution (wrw)
- fp16
- XDLOPS
- not padding along gemm m/n/k dimension


src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -332,36 +365,82 @@ InvokerFactory MakeMlirWrWInvokerFactory(const ConvolutionContext& ctx)
                            weights_strides,
                            out_dims,
                            out_strides);
-    MlirConvArgs args =
-        MakeMlirConvArgs(in_dims, in_strides, weights_dims, weights_strides, out_dims, out_strides);
+    MlirConvArgs args = MakeMlirConvArgs(in_dims,
```
jerryyin, line 368: Looks like you need an if condition here: Based on whether a workspace is needed, invoke either `MakeMlirConvArgs()` or `MakeMlirConvArgsWithWorkspace()`.

jerryyin (commented):
Overall looks good, I have mostly nitpicks.

Biggest feedback is to add one test to track a config that requires a workspace, following the below pointer:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/df658f382c5af02133836a8fa5700a23b58c2ca0/test/CMakeLists.txt#L736-L743

src/solver/conv_mlir_igemm_wrw_xdlops.cpp
```diff
@@ -113,5 +114,20 @@ ConvSolution ConvMlirIgemmWrWXdlops::GetSolution(const ConvolutionContext& ctx,
...
+    else
+    {
+        const auto k = ConvolutionContextInterpreter::GetOutputChannelK(ctx);
+        const auto c = ConvolutionContextInterpreter::GetInputChannelC(ctx);
+        const auto y = ConvolutionContextInterpreter::GetFilterHeightY(ctx);
+        const auto x = ConvolutionContextInterpreter::GetFilterWidthX(ctx);
+
+        return k * c * y * x * miopen::GetTypeSize(miopenFloat);
+    }
+}
```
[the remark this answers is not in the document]
whchung: `MiirGetWorkspaceSize` takes a string, instead of `ConvolutionContext`. How do I populate such a string?

In `ConvMlirIgemmWrWXdlops::GetSolution`, I can see some logic to construct `construction_parameters.comp_options)`, but it also depends on an `PerformanceConvMlirIgemmXdlops` object which isn't available here.

whchung:
> Overall looks good, I have mostly nitpicks.
> 
> Biggest feedback is to add one test to track a config that requires a workspace, following the below pointer:
> 
> https://github.com/ROCmSoftwarePlatform/MIOpen/blob/df658f382c5af02133836a8fa5700a23b58c2ca0/test/CMakeLists.txt#L736-L743

@jerryyin Will do. How do I trigger the test manually at command line?

requirements.txt
```diff
@@ -1,4 +1,4 @@
 sqlite3@3.17 -DCMAKE_POSITION_INDEPENDENT_CODE=On
 boost@1.72 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build
 half,https://github.com/pfultz2/half/archive/1.12.0.tar.gz -X header -H sha256:0a08660b68abb176ebc2a0cdf8de46e3182a7f46c66443bb80dbfaaec98cf969 --build
-ROCmSoftwarePlatform/llvm-project-mlir@release/rocm-5.1 -DBUILD_FAT_LIBMLIRMIOPEN=1
+ROCmSoftwarePlatform/llvm-project-mlir@fp16-wrw-optimization-rebase -DBUILD_FAT_LIBMLIRMIOPEN=1
```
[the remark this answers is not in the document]
whchung: Will do.

jerryyin:
> @jerryyin Will do. How do I trigger the test manually at command line?

`make test_conv_igemm_mlir_xdlops` will compile and test all the configs highlighted (but I believe it is using cpu so is slow). In development time, you can comment out the irrelevant ones to speed it up..

requirements.txt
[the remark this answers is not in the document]
jerryyin: I might not make it clear. I don't think we need that comment there. Instead, since your wrw fp16 PR in MLIR has already been merged to the mainline, you can just pick up the latest commit in our HEAD and make sure this feature is included, such that:

> ROCmSoftwarePlatform/llvm-project-mlir@b04c00925237718e5d260efb2a1776ee6f99f5b4 -DBUILD_FAT_LIBMLIRMIOPEN=1

src/solver/conv_mlir_igemm_wrw_xdlops.cpp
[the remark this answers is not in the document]
whchung: @jerryyin I think I've figured out how to do it after examining the interface more clearly.

requirements.txt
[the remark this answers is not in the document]
whchung: Pushed another commit to address this.

src/solver/conv_mlir_igemm_wrw_xdlops.cpp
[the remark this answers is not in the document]
jerryyin: ConvolutionContext has every information we need to construct that string. To use it, you can:

```cpp
  MiirGetWorkspaceSize(mlir::ConstructBuildOptions(ctx, true));
```

Since I can see you already setup the workspace using below, I don't think you need this `GetWorkspaceSize()`

```cpp
    result.workspace_sz = MiirGetWorkspaceSize(construction_parameters.comp_options);
```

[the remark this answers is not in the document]
whchung: Pushed another commit to consolidate `MiirGetWorkspaceSize` call to a single place in `GetWorkspaceSize`.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -164,35 +167,65 @@ MlirConvArgs MakeMlirConvArgs(const std::vector<size_t>& in_dims,
     StridedMemRef5D output{nullptr, nullptr, 0, {0, 0, 0, 0, 0}, {0, 0, 0, 0, 0}};
     initDimStrides(out_dims, out_strides, output);
 
-    return {filter, input, output};
+    if(populateWorkspaceArg)
```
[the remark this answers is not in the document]
whchung: This would introduce duplicated logic so I'd like to avoid it.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -164,35 +167,65 @@ MlirConvArgs MakeMlirConvArgs(const std::vector<size_t>& in_dims,
...
 // way around is to call clSetKernelArg on a oclMemory object to pass
 // the device pointer to the kernel
 #if MIOPEN_BACKEND_HIP
-void SetMlirConvArgsPtr(ConstData_t in, ConstData_t out, ConstData_t w, MlirConvArgs& args)
+void SetMlirConvArgsPtr(ConstData_t in,
+                        ConstData_t out,
+                        ConstData_t w,
+                        ConstData_t wk,
+                        MlirConvArgs& args,
+                        bool populateWorkspaceArg = false)
```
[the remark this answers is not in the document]
whchung: This would also introduce duplicated logic so I'd like to avoid it.

src/conv/invokers/mlir_impl_gemm.cpp
whchung: Same as above. I'd like to avoid duplicated logic in boilerplate codes so I'm adding this optional `bool` field.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
jerryyin: This will actually consolidate the logic if you invoke the no-workspace overload from inside the workspace overload.

src/conv/invokers/mlir_impl_gemm.cpp
jerryyin: You can have `MakeMlirConvArgsWithWorkspace()` call `MakeMlirConvArgs()` that in the end result each function is simplified.

jerryyin: But then the workspace arg in `MlirConvArgs` will be uninitialized. 

This may not be a problem since we are not effectively using all fields of a memref but I prefer this to be addressed to be conservative.

atamazov (changes requested):
This PR should not change 42 files. Please merge recent develop into this branch.

src/conv/invokers/mlir_impl_gemm.cpp
whchung: @jerryyin I think I need to understand the architecture between a solver / an invoker / a ConvolutionContext better to see how to derive a better solution.

In a `ConvMlirIgemmWrWXdlops` solver, given a `ConvolutionContext`, we can use `MiirGetWorkspaceSize` to compute the size of workspace needed. How do we achieve the same here at `MakeMlirWrWInvokerFactor`?

whchung: @jerryyin So my main question is:

How to derive if a workspace is needed within this couple of lines?

```
InvokerFactory MakeMlirWrWInvokerFactory(const ConvolutionContext& ctx)
{
    assert((ctx.direction.IsBackwardWrW()));

    std::vector<size_t> in_dims, in_strides;
    std::vector<size_t> weights_dims, weights_strides;
    std::vector<size_t> out_dims, out_strides;
    ComputeMlirDimsStrides(ctx.conv_problem,
                           in_dims,
                           in_strides,
                           weights_dims,
                           weights_strides,
                           out_dims,
                           out_strides);
```

Can we invoke `ConvMlirIgemmWrWXdlops::GetWorkspaceSize` here? But `ConvMlirIgemmWrWXdlops::GetWorkspaceSize` is a member function, not a class function, and an invoker does not possess a reference to a solver.

whchung:
> This PR should not change 42 files. Please merge recent develop into this branch.
@atamazov It's done.

src/conv/invokers/mlir_impl_gemm.cpp
atamazov: @whchung `MakeMlirWrWInvokerFactory()` is called from `GetSolution()`. Compute workspace size in `GetSolution()` and pass it to `MakeMlirWrWInvokerFactory()` as a parameter.

[Informative] BTW you can instantiate Solver wherever you want and call its methods provided that you have necessary parameters on hand. Example: `ConvMlirIgemmWrWXdlops{}.GetWorkspaceSize(ctx)`.

whchung: Per discussion with @jerryyin, I'll add some comments to explain why the implementation is done in this fashion. Meanwhile, we have reached a consensus that the existing implementation is okay for the time being.

src/conv/invokers/mlir_impl_gemm.cpp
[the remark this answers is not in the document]
whchung: Updated in the latest commit.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -46,6 +47,7 @@ struct MlirConvArgs
     StridedMemRef5D filter;
     StridedMemRef5D input;
     StridedMemRef5D output;
+    boost::optional<StridedMemRef5D> workspace;
```
atamazov, line 50: Can you please show me where `boost::optional` semantics is used?

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -164,7 +167,14 @@ MlirConvArgs MakeMlirConvArgs(const std::vector<size_t>& in_dims,
...
 
-    return {filter, input, output};
+    if(populateWorkspaceArg)
+    {
+        StridedMemRef5D workspace{nullptr, nullptr, 0, {0, 0, 0, 0, 0}, {0, 0, 0, 0, 0}};
+        initDimStrides(weights_dims, weights_strides, workspace);
+        return {filter, input, output, workspace};
+    }
+
+    return {filter, input, output, {}};
```
whchung, line 177: @atamazov It's used here.

whchung: The optional `StridedMemRef5D` is only used by `MakeMlirWrWInvokerFactory` when invoking `MakeMlirConvArgs`. Other invokers would not use the field.

requirements.txt
```diff
@@ -1,4 +1,4 @@
 sqlite3@3.17 -DCMAKE_POSITION_INDEPENDENT_CODE=On
 boost@1.72 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build
 half,https://github.com/pfultz2/half/archive/1.12.0.tar.gz -X header -H sha256:0a08660b68abb176ebc2a0cdf8de46e3182a7f46c66443bb80dbfaaec98cf969 --build
-ROCmSoftwarePlatform/llvm-project-mlir@release/rocm-5.1 -DBUILD_FAT_LIBMLIRMIOPEN=1
+ROCmSoftwarePlatform/llvm-project-mlir@b04c00925237718e5d260efb2a1776ee6f99f5b4 -DBUILD_FAT_LIBMLIRMIOPEN=1
```
junliume, line 4: @whchung could you check the CMakeLists file in MLIR?
```
CMake Error at external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt:193 (message):

No non-CPU ROCm agents found on the system, and ROCM_TEST_CHIPSET is not
 defined

Comment (whchung) [reply requirements.txt:4]: @junliume This seems to be a test node-specific issue, from the error message it seems to denote no GPU is found on the test node.

Comment (whchung) [reply requirements.txt:4]: @junliume Wondering what's the ROCm version used in the CI of MIOpen? This particular commit was added today (2/24/2022) to migrate everything on MLIR CI infrastructure to ROCm 5.0 if I understand correctly.

Comment (junliume) [reply requirements.txt:4]: @whchung MIOpen CI has not yet upgraded to ROCm 5.0 (we are still using 4.3.1) but will do so soon.
@okakarpa Can we start testing on CI upgrade for a selected few nodes first, and plan on moving all CI nodes to 5.0.1 eventually? Thanks! If a meeting is needed between MIOpen and MIGraphX folks, please feel free to arrange one. Thanks!

Comment (jerryyin) [reply requirements.txt:4]: @whchung The commit is not an issue. That commit will upgrade our CI's ROCm version and has no impact on our source code or how MIOpen builds our static library.

Judging by that error, it points to the machine executor `rocm_agent_enumerator` return an empty agent list. The reason why happened is due to this is an opencl environment and that we shouldn't presume `rocm_agent_enumerator` to must exist in an client environment. The must have been an issue brought by the recent upstream merge and has to be reverted. The reason why we didn't catch it locally is due to we don't test opencl builds in our CI.

Comment (whchung) [reply requirements.txt]: @jerryyin Thanks. It's done.

Comment (jerryyin) [reply requirements.txt]: @jerryyin I'd actually prefer `MLIR_ENABLE_ROCM_RUNNER=0` to be set in MLIR's wildcard build with `BUILD_FAT_LIBMLIRMIOPEN` and regressed in MLIR CI first. If we just change it here it will create discrepancy in how MIOpen use our static library vs how our CI test it. So the right sequence to do this should be:
 1. Submit a new PR in MLIR that make the rocm runner turned off in the top level cmake
 2. Have that pass our CI, got it merged
 3. Update the commit hash in this PR

Comment (whchung) [reply requirements.txt]: MLIR PR filed: https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/529

Comment (whchung) [reply requirements.txt]: @jerryyin Per discussion and https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/blob/miopen-dialect/external/llvm-project/mlir/lib/ExecutionEngine/CMakeLists.txt#L185-L199 I think the earlier assessment was incorrect. It's still a node-specific issue, not related with the logic introduced in this PR. I've closed https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/529 .

Comment (whchung) [reply requirements.txt]: Now https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/pull/531 has been merged, I've moved the pointer in this PR.

**Comment by whchung** [comment]
jenkins: retest this please.

**Review by jerryyin — APPROVED** [review]
LGTM

**Comment by atamazov** [comment]
@whchung 
> [force-pushed](https://github.com/ROCmSoftwarePlatform/MIOpen/compare/fb78a741260237435062ba92b943258af457f29d..b9a14c30786da650e2aa9ab2c7a5a86471eeedaa) the mlir-workspace branch from fb78a74 to b9a14c3 4 days ago

Force-pushing is NOT recommended due to several reasons. It can be more or less safely used only if the branch has not yet been published for review.

File: src/conv/invokers/mlir_impl_gemm.cpp (hunk 6 above)
Comment (atamazov) [reply src/conv/invokers/mlir_impl_gemm.cpp:50]: This is not resolved.

File: src/conv/invokers/mlir_impl_gemm.cpp (hunk 7 above)
Comment (atamazov) [reply src/conv/invokers/mlir_impl_gemm.cpp:177]: @whchung But this simply calls a default constructor for the object. I do not see where we check if `workspace` exists using the `boost::optional` methods, like:
```C++
if(args.workspace)
{
    // Workspace exists; do something with *(args.workspace)
}
else
{
    // Optional args.workspace object does not exist and shouldn't be accessed
}
```
I recommend removing `boost::optional` if it is not used.

atamazov (changes requested):

src/conv/invokers/mlir_impl_gemm.cpp
atamazov, line 177: And this is how construction problem can be resolved without `boost::optional`:
```suggestion
    StridedMemRef5D workspace{nullptr, nullptr, 0, {0, 0, 0, 0, 0}, {0, 0, 0, 0, 0}};
    {
        initDimStrides(weights_dims, weights_strides, workspace);
        return {filter, input, output, workspace};
    }

    return {filter, input, output, workspace};
```

atamazov, line 177: And this is how construction problem can be resolved without `boost::optional`:
```suggestion
    StridedMemRef5D workspace{nullptr, nullptr, 0, {0, 0, 0, 0, 0}, {0, 0, 0, 0, 0}};
    if(populateWorkspaceArg)
        initDimStrides(weights_dims, weights_strides, workspace);
    return {filter, input, output, workspace};
```

atamazov: @junliume FYI I have limited privileges and unable to "unresolve" comments.

whchung: @atamazov I'm following the convention used in other parts of MIOpen when it comes to `boost::optional`. For example:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/target_properties.cpp#L82

I can remove the usage of `boost::optional` if you find it necessary. Please let me know if it's a recommendation or a requirement. Thanks.

whchung: @atamazov Also, we don't need to check for `args.workspace` because of:
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1443/files#diff-6a213cde36650f2a3771d5e595c4f5754984da9a3ba95e32e5d2634846d6fadcR364-R375
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1443/files#diff-6a213cde36650f2a3771d5e595c4f5754984da9a3ba95e32e5d2634846d6fadcL344-R391

whchung: I have just submitted a new commit to remove `boost::optional`.

atamazov: @whchung  
> @atamazov I'm following the convention used in other parts of MIOpen when it comes to `boost::optional`. For example: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/target_properties.cpp#L82

This is how we construct an optional object (and construction is done correctly in this PR). However if you construct an optional object, then let's use the "optional" semantics (otherwise we get overcomplicated code that looks suspicious).

> I can remove the usage of `boost::optional` if you find it necessary. Please let me know if it's a recommendation or a requirement. Thanks.

If optional semantics is not used, then it should be removed. In principle, this is a requirement, but not a hard one. For example, if it is difficult for you to rework the code, then I can do this for you, now or later.

@junliume can dismiss my blocking review, if necessary.

atamazov:
@wchung Thanks.

whchung:
CI failing at various targets under `test_gpu_reference_kernel`. I don't believe it's related to changes introduced in this PR.

jerryyin (approved):

atamazov:
@whchung How urgent is this? I would like to thoroughly review this (hopefully tomorrow), if we have time.

atamazov:
@whchung 
> CI failing at various targets under `test_gpu_reference_kernel`. I don't believe it's related to changes introduced in this PR.

Please check if the baseline of this branch has passed CI. Then restart CI from the failure point. If the issue is systematic, you'll see it again. (I am unable to access CI these days)

whchung:
> @whchung How urgent is this? I would like to thoroughly review this (hopefully tomorrow), if we have time.

@atamazov It's not urgent, but I'd like to have it land within a week. I'm working on 2 other derivative changes which eliminate dependency to OpenCL GPU kernels in MLIR solvers.

whchung:
@atamazov Also it'd be great if you could share some insights on how to improve test coverage. I have no idea how to interpret the results and improve it.

junliume:
@whchung @atamazov  I can help monitor the CI. While the last pipeline has just been kicked off, the last a few failed at "Fp16 Hip All gfx1030" due to "Memory access fault".
I am testing the branch on a local Navi21 with ROCm 5.0 installed to tell if it is rocm version issue.

atamazov:
@whchung 
> It's not urgent, but I'd like to have it land within a week.

I see.

whchung:
@junliume It seems the nodes employed in MIOpen CI may have some problem.

http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/mlir-wrw-no-set-zero-no-cast-tensor/6/pipeline

It can be observed a system without AMD GPU was selected for `Smoke Fp32 / Fp32 OpenCL gfx90a`.

```
[2022-03-04T20:03:42.432Z] NODE_NAME = dell113-r01-u27-28
[2022-03-04T20:03:42.432Z] + lsb_release -sd
[2022-03-04T20:03:42.432Z] Ubuntu 20.04.3 LTS
[2022-03-04T20:03:42.432Z] + uname -r
[2022-03-04T20:03:42.432Z] 5.13.0-30-generic
[2022-03-04T20:03:42.432Z] + cat /sys/module/amdgpu/version
[2022-03-04T20:03:42.432Z] cat: /sys/module/amdgpu/version: No such file or directory
```

On the other hand, the very same target was fine in the previous build:

http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/mlir-wrw-no-set-zero-no-cast-tensor/5/pipeline

junliume:
> @junliume It seems the nodes employed in MIOpen CI may have some problem.
> 
> http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/mlir-workspace/30/pipeline
> 
> It can be observed a system without AMD GPU was selected for `Smoke Fp16/Bf16/Int8 / Fp16 Hip gfx90a /opt/rocm`.

@whchung yep that's the same node that is causing issues in other pipelines. I have disabled it and restarted the pipeline.

atamazov:
[off-topic]
> It seems the nodes employed in MIOpen CI may have some problem.

@whchung @junliume Just to remind that:
(1) The AMDGPU kernel module is newer than ROCm in docker images used for testing. 
(2) Both AMDGPU kernel modules and ROCm are older than 5.0 and thus not "officially certified" for Navi21.

I recommend upgrading everything to ROCm 5.0. See #1372.

junliume:
> [off-topic]
> 
> > It seems the nodes employed in MIOpen CI may have some problem.
> 
> @whchung @junliume Just to remind that:
> 
> * (1) The AMDGPU kernel module is newer than ROCm in docker images used for testing.
> * (2) Both AMDGPU kernel modules and ROCm are older than 5.0 and thus not "officially certified" for Navi21.
> 
> I recommend upgrading everything to ROCm 5.0. See #1372.

@whchung and @atamazov Yes the upgrade plan is in process, and we have some issues with trials which is getting resolved.

The good news is that CI for this branch has passed :) So you can start with final reviewing.

whchung:
A gentle ping.

src/solver/conv_mlir_igemm_wrw_xdlops.cpp
```diff
@@ -105,6 +105,7 @@ ConvSolution ConvMlirIgemmWrWXdlops::GetSolution(const ConvolutionContext& ctx,
 
     result.invoker_factory = conv::MakeMlirWrWInvokerFactory(ctx);
     result.construction_params.push_back(construction_parameters);
+    result.workspace_sz = GetWorkspaceSize(ctx);
```
atamazov, line 108: Pass required workspace size to `MakeMlirWrWInvokerFactory()`

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -332,36 +360,84 @@ InvokerFactory MakeMlirWrWInvokerFactory(const ConvolutionContext& ctx)
...
         return [=](const Handle& handle, const AnyInvokeParams& primitive_parameters) mutable {
             const auto& wrw_invoke_params = primitive_parameters.CastTo<conv::WrWInvokeParams>();
             const auto& tensors           = wrw_invoke_params.tensors;
+            const auto& workspaceSize     = wrw_invoke_params.workSpaceSize;
+            float zero                    = 0.f;
 
-            // Only fp32 use atomic_add, which accumulate the result into dw.
-            // Therefore the need for setting to zero beforehand.
-            if(tensors.dwDesc.GetType() == miopenFloat)
+            if(workspaceSize > 0)
```
atamazov, line 390: It's sad to say, but unfortunately this is incorrect.

`workspaceSize` here is a size of workspace actually passed from the user's application. It may be smaller than required (so we must check and issue en error if so) or it can be bigger then required (and there should be no harm in this case).

Let's consider the latter. If kernel does not require workspace, but a user's app passes non-zero workspace, then this branch would take wrong path.

- Please use required workspace size here.
- Please also check if `workspaceSize` >= required workspace.

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -332,36 +360,84 @@ InvokerFactory MakeMlirWrWInvokerFactory(const ConvolutionContext& ctx)
...
+    //    and the pointer will be supplied by the solver.
+    // d) In case a workspace is not needed, the pointer of a workspace will
+    //    not be passed to MLIR-generated kernel.
+    MlirConvArgs args = MakeMlirConvArgs(in_dims,
+                                         in_strides,
+                                         weights_dims,
+                                         weights_strides,
+                                         out_dims,
+                                         out_strides,
+                                         /*populateWorkspaceArg=*/true);
```
atamazov, line 381: I recommend passing required workspace size here instead of boolean

src/conv/invokers/mlir_impl_gemm.cpp
```diff
@@ -164,7 +166,14 @@ MlirConvArgs MakeMlirConvArgs(const std::vector<size_t>& in_dims,
     StridedMemRef5D output{nullptr, nullptr, 0, {0, 0, 0, 0, 0}, {0, 0, 0, 0, 0}};
     initDimStrides(out_dims, out_strides, output);
 
-    return {filter, input, output};
+    StridedMemRef5D workspace{nullptr, nullptr, 0, {0, 0, 0, 0, 0}, {0, 0, 0, 0, 0}};
+    if(populateWorkspaceArg)
```
atamazov, line 170: If required workspace size > 0

atamazov (changes requested):
Wish I caught it sooner ;(

whchung:
@atamazov Thanks for the review. If I understand you correctly, you are suggesting I change the interface for `MakeMlirWrWInvokerFactory` so the required workspace is passed to the invoker factory, right?

Wondering could you help point other existing solvers/invokers so I can find good references?

src/solver/conv_mlir_igemm_wrw_xdlops.cpp
whchung: Will do. Do we still need to set `result.workspace_sz` here? And where inside `MakeMlirWrWInvokerFactor` should I keep track of the required workspace size?

src/conv/invokers/mlir_impl_gemm.cpp
whchung: @atamazov Please let me know if my understanding is correct. I'm not that familiar with the interaction between invoker factor, solver, and solution:

- change `MakeMlirWrWInvokerFactor` which takes another extra parameter about the required workspace size, which is returned by MLIR kernel generator.
- change `MakeMlirConvArgs` to use the required workspace size, instead of `bool`.
- inside the double lambda returned by the invoker factory, apply the two checks you mentioned.

src/solver/conv_mlir_igemm_wrw_xdlops.cpp
whchung: @atamazov I have submitted a patch. Please help review it again. Thanks. I'm still keeping `result.workspace_sz` be set here.

src/conv/invokers/mlir_impl_gemm.cpp
whchung: @atamazov Please help review the latest commit. I tried to address all the feedback based on my understanding.

atamazov: > @atamazov Please let me know if my understanding is correct.



> 
>     * change `MakeMlirWrWInvokerFactor` which takes another extra parameter about the required workspace size, which is returned by MLIR kernel generator.

Yes. Small clarification: required workspace size is returned by solver's `GetWorkspaceSize()` (maybe it asks MLIR kernel gen for it, but this is implementation detail).

>     * change `MakeMlirConvArgs` to use the required workspace size, instead of `bool`.

This is just a recommendation. I hope that there is no cases when GetWorkspaceSize() returns 0 but MLIR WrW kernel still need zero-sized workspace to be passed in.

>     * inside the double lambda returned by the invoker factory, apply the two checks you mentioned.

[Informative] Outer lambda is Invoker Factory. Inner one is Invoker.

In the Invoker, pass workspace-related arguments to the kernel when required workspace size is > 0.

Also check if passed workspace is of enough size.

```C++
if(invoke_params.workSpaceSize < required workspace size)
    MIOPEN_THROW("Not enough workspace for ...");
```

whchung: @atamazov Got it. I think I addressed your feedback in the latest patch.

src/solver/conv_mlir_igemm_wrw_xdlops.cpp
atamazov: >  Do we still need to set result.workspace_sz here? 

yes.

atamazov (dismissed):
LGTM!

atamazov (approved):
LGTM

whchung:
@atamazov Sorry the workstation I have access to build MIOpen was down for several hours, so my last commits were submitted without being able to build it. Now the system is back online and I've made sure this very last commit f5ad6ab could finish the build. Would you mind review this PR again. Sorry for your inconveniences.

atamazov (approved):
No problem))

junliume (approved):

whchung:
@junliume `Fp16 Hip All gfx1030` test failed at `test_conv_ck_igemm_fwd_v6r1_dlops_nchw`, which is not related with this PR. Could you advise help to proceed?

junliume:
> @junliume `Fp16 Hip All gfx1030` test failed at `test_conv_ck_igemm_fwd_v6r1_dlops_nchw`, which is not related with this PR. Could you advise help to proceed?

@whchung @atamazov I have restarted it and spawn another check on another gfx1030 with updated ROCm. If at least the latter has passed I will mark this PR as CI passed, and merge to develop accordingly. ETA: by EOD 03112022 PST.

junliume:
> > @junliume `Fp16 Hip All gfx1030` test failed at `test_conv_ck_igemm_fwd_v6r1_dlops_nchw`, which is not related with this PR. Could you advise help to proceed?
> 
> @whchung @atamazov I have restarted it and spawn another check on another gfx1030 with updated ROCm. If at least the latter has passed I will mark this PR as CI passed, and merge to develop accordingly. ETA: by EOD 03112022 PST.

I am still having issues with CI ... trying again now. The branch passed on my local system though.

junliume:
> > > @junliume `Fp16 Hip All gfx1030` test failed at `test_conv_ck_igemm_fwd_v6r1_dlops_nchw`, which is not related with this PR. Could you advise help to proceed?
> > 
> > 
> > @whchung @atamazov I have restarted it and spawn another check on another gfx1030 with updated ROCm. If at least the latter has passed I will mark this PR as CI passed, and merge to develop accordingly. ETA: by EOD 03112022 PST.
> 
> I am still having issues with CI ... trying again now. The branch passed on my local system though.

CI passed after disabling two problematic nodes.

PR #1433, opened 2022-02-17
[Solvers] Base class for tunable solvers

averinevg:
- Added base class for tunable solvers
- `GetSolution()`: `disableConfigOverrideFromEnv` has been removed
- `GetPerformanceConfig()` has been renamed to `GetDefaultPerformanceConfig()`

PR #1445, opened 2022-02-25
Handles for calling perf_db compile & eval through fin

cderb:
These are changes necessary for allowing Fin to execute perf_db compile and perf_db evaluate.

GetAllSolutions function in any_solver / generic_search to facilitate return of array of all possible solutions to fin.
Changes around generic_search and solvers to allow for generic_search constructs to be used from call originating outside the file.
GetPerfCfgParams function in any_solver to return the parameter string for supplied context.
Added LegacySolver test in any_solver.
Added db_update to execution context to allow for programmatic selection of update execution.


cderb:
Working to fix clang-format errors.

src/include/miopen/generic_search.hpp
```diff
@@ -296,6 +298,43 @@ using RunAndMeasure_t =
...
+
+    auto all_configs = GetAllConfigs(s, context);
+
+    std::vector<ConvSolution> solutions;
+    for(const auto& current_config : all_configs)
+    {
+        ConvSolution current_solution = s.GetSolution(context, current_config, true);
+        solutions.push_back(current_solution);
+    }
+    return solutions;
```
atamazov, line 335: This breaks the idea behind ComputedContainer. The point is to not occupy memory, thus making the potential number of tuning configs unlimited. That is why we are using ComputedContainer instead of a regular STL container. As soon as we copy the "virtual" content of ComputedContainer into a regular STL container, it starts to take up memory.

atamazov (changes requested):
[Note] This PR does not contain `fin` changes, so I can't understand the goal of this PR. If you explain the purpose, then I hope I can suggest a more appropriate design.

atamazov:
This should also be consistent with the upcoming refactoring designed by @DrizztDoUrden and @averinevg. I highly recommend assigning them as reviewers (I can't do that now).

src/include/miopen/generic_search.hpp
cderb: This is an addition for Fin performance db compile step. What I'm trying to accomplish here is to pass the list of all solutions up to Fin so that Fin can do a precompile kernel for each solution. This also allows Fin to select each kernel created and add them to a json output file.
Perhaps there is a way to pass the solutions virtually?

cderb:
> [Note] This PR does not contain `fin` changes, so I can't understand the goal of this PR. If you explain the purpose, then I hope I can suggest a more appropriate design.

Fin changes are in the following PR, I apologize if it isn't accessible at the moment.
https://github.com/ROCmSoftwarePlatform/fin/pull/42 
When I get it merged to Fin develop I'll bundle it in either this PR or a following PR.

averinevg (commented):
We are going to remake `AnySolver` so that it will only translate calls to `SearchableSolverBase`. See https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1433 for information about the implementation of  `SearchableSolverBase`. Therefore. it might be better to put all helper methods in modules where they are used.

CC @atamazov @DrizztDoUrden

src/include/miopen/solver.hpp
```diff
@@ -2321,6 +2321,18 @@ struct PerformanceConfigConvOclBwdWrw2
...
+bool PerformanceConfigConvOclBwdWrw2<N_BATCH_LOOPS>::operator==(
+    const PerformanceConfigConvOclBwdWrw2<N_BATCH_LOOPS>& other) const
+{
+    // clang-format off
+    return n_waves == other.n_waves
+        && read_size == other.read_size
+        && n_out_channels_per_tile == other.n_out_channels_per_tile
+        && n_out_channels_tiles == other.n_out_channels_tiles
+        && n_out_rows_in_lcl == other.n_out_rows_in_lcl; // clang-format on
+}
```
averinevg, line 2334: Why was this code moved here?

cderb: There were some issues with load order which cause a bunch of warnings to be issued when trying to call the GetAllSolvers function from AnySolver. This may have been solved by the declaration of the specializations for PerformanceConfigConvOclBwdWrw2. So, I'll check if this can be reverted and still compile cleanly.

cderb: @averinevg successfully reverted without affecting compilation warnings.

src/include/miopen/generic_search.hpp
atamazov: What about the following design.

Generic search checks for some callback (pointer to a function). If it is nullptr (default state), then generic search works as usual. If it is set, then, instead of building the Solution and benchmarking it, generic search calls that function (passing each Solution as a parameter).

`fin` should set the callback prior launching search. The callback may simply collect received Solutions into some container (for further handling) or immediately perform necessary actions with each Solution (compile and add to a json output).

atamazov: @cderb IMPORTANT QUESTION: What is the purpose of the compiled kernels? I am asking because the library includes only tunable kernels into Solutions when tuning is in progress. For example, if a Solution consists of two kernels (convolution + some transform) and only convolution kernel is tunable, then the library includes only convolutions into solutions for tuning.

atamazov:
@cderb Please set urgency label, if possible. Thanks!

src/include/miopen/generic_search.hpp
cderb: Interesting idea, It will take me some more development time to figure that out. I'd like to work on that in a separate PR since this method is stable currently.

The purpose of collecting the compiled kernels is entirely for Tuna's tuning process. In our compile step we store all possible kernel compilations in Tuna's mysql database, the evaluation step then consumes these kernels from mysql. It acts as a way of separating our compile machines and evaluation machines.

cderb:
De-serialize test is failing for solvers: ConvOclDirectFwd and ConvOclDirectFwd1x1 across all libraries it seems. I don't think the serialization / solver code was changed by this PR. @junliume was this introduced by one of the tuning PRs? There are some older perf libraries in here that weren't changed recently as well.

junliume:
> De-serialize test is failing for solvers: ConvOclDirectFwd and ConvOclDirectFwd1x1 across all libraries it seems. I don't think the serialization / solver code was changed by this PR. @junliume was this introduced by one of the tuning PRs? There are some older perf libraries in here that weren't changed recently as well.

I don't think codes related to those two solvers have changed for a while. Maybe the tuning issue? CC: @JehandadKhan

junliume:
@JehandadKhan  @cderb  yep Perf DB Deserialize Test is failing pretty consistently which is the issue.

JehandadKhan:
@cderb Can you please check what is going on with the failing test ?

atamazov:
If this is not urgent, then I recommend either postponing this until [refactoring is complete](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1445#issuecomment-1058446444) or re-implementing this [with callbacks](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1445#discussion_r819867212). Why:

- After ongoing refactoring done, the code changes from this PR won't work. In fact, @averinevg or @DrizztDoUrden will have to rewrite almost all of the code from this PR.
- Callback-based implementation involves smaller number of changes. And it does not have the drawback [mentioned here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1445#discussion_r819028492).

cderb:
@atamazov in short, we can still accomplish tuning without the change. This change is an enhancement to increase Tuna's productivity.
I'll attach the Fin extension of this PR once it has merged to develop on that side. And we can pause this branch here until the refactor is completed
In the mean time I can attempt the suggested callback implementation.

cderb:
I've added the Fin changes to this PR to give context.

JehandadKhan:
@DrizztDoUrden  Lets discuss this in our upcoming meeting

junliume:
> @DrizztDoUrden Lets discuss this in our upcoming meeting

@JehandadKhan and @DrizztDoUrden any follow ups? :)

cderb:
Merge resolution in new PR.

PR #1451, opened 2022-03-05
[HOTFIX][SWDEV-326460] check MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC in IsApplicable()

carlushuang:
This is an hot fix for [SWDEV-316705](https://ontrack-internal.amd.com/browse/SWDEV-316705)
Due to our current design, if a solver has multiple configs, and some of the config support deterministic(e.g. non-deterministic), some are not, we can not disable a single config in `IsValid()` function. But We need disable it in `IsApplicable()`, which means will disable this solver entirely if `MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC` env var is set.
This may have a performance impact, but since currently this env var is for test purpose (in this case performance impact is OK per our customer), so this approach is simple and fine.
If in future has performance requirement under `MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC`, need to seek for other approach and implementation.

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_bwd_v4r1.cpp
+++ b/src/solver/conv_hip_implicit_gemm_bwd_v4r1.cpp
@@ -749,6 +749,8 @@ bool ConvHipImplicitGemmBwdDataV4R1::IsApplicable(const ConvolutionContext& ctx)
 
     if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_BWD_V4R1{}))
         return false;
+    if(miopen::IsEnabled(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC{}))
+        return false;
 
     if(!IsComposableKernelSupportedHardware(ctx))
         return false;
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -2436,8 +2436,8 @@ miopenBatchNormalizationForwardTraining(miopenHandle_t handle,
  *
  * If either estimatedMean, or estimatedVariance are null pointers then the values for the mean and
  * variance will be calculated from input data and this calculated mean and variance will be used
- * to update input values. 
- * If variance is zero and epsilon is also zero, this function outputs NAN values.  Input espilon 
+ * to update input values.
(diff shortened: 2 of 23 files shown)


carlushuang:
blocked by https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1452

src/solver/conv_hip_implicit_gemm_bwd_v4r1.cpp
```diff
@@ -749,6 +749,8 @@ bool ConvHipImplicitGemmBwdDataV4R1::IsApplicable(const ConvolutionContext& ctx)
 
     if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_BWD_V4R1{}))
         return false;
+    if(miopen::IsEnabled(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC{}))
+        return false;
```
shaojiewang, line 753: I think hip igemm v4r1 is deterministic. Why do we ban this solver?

carlushuang: For debug purpose(not consider performance), need to disable a solver if any of the configs may use atomic. So I just disable all suspicious one.

shaojiewang (approved):
LGTM, waiting for CI.

atamazov:
@carlushuang I am Ok with this PR, but can you please comment out was is wrong with #1377?

> Due to our current design, if a solver has multiple configs, and some of the config support deterministic(e.g. non-deterministic), some are not, we can not disable a single config in IsValid() function.

We can, and the library will use the **_default_** tuning config instead. The cost is a warning message.

junliume:
@JehandadKhan @asroy @zjing14 @atamazov  last call for reviewers :) Plan to merge and cherry pick later today 03/08.

atamazov (approved):
I am OK with this provided that all the affected iGemm solvers use floating-point addition in order to compute sums of values calculated by different work-items.

atamazov:
I highly recommend renaming this PR in order to include "(SWDEV-316705)" into commit name.

PR #1457, opened 2022-03-10
[issue 1458]reduce multi-buffer size

### LINKED ISSUE #1458: There is no need to add padding for the most recent buffer because there is no buffer after it.
There is https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1354 which directly modifies fdb in order to resolve #1349 without retuning. It blindly adds 3*255 bytes to workspace size, which should be enough to align 4 sub-buffers.

**_There is no need to add padding for the most recent buffer because there is no buffer after it._**

I suspect that the reason of the issue is that `MultiBufferWorkspaceTraits::MultiBufferWorkspaceTraits()` wants too much padding. It adds padding for all four buffers, and total padding may exceed 3*255.

I think that the proper fix should remove adding padding for the last sub-buffer. `MultiBufferWorkspaceTraits::GetSize()` should return offset of the beginning of the last sub-buffer + the size of the last sub-buffer. **_Basically, no other fixes required._**

_Originally posted by @atamazov in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1429#issuecomment-1048305729_

shaojiewang:
Resolve #1458 
reduce multi buffer size by using offset of last buffer plus last buffer size

### DIFF
--- a/src/buffer_info.cpp
+++ b/src/buffer_info.cpp
@@ -164,7 +164,7 @@ MultiBufferWorkspaceTraits::MultiBufferWorkspaceTraits(std::initializer_list<siz
 
 size_t MultiBufferWorkspaceTraits::GetSize() const
 {
-    return v_offset[v_offset.size() - 2] + v_size.back();
+    return (&v_offset.back())[-1] + v_size.back();
 }
 
 size_t MultiBufferWorkspaceTraits::GetOffset(size_t index) const
--- a/src/buffer_info.cpp
+++ b/src/buffer_info.cpp
@@ -162,7 +162,10 @@ MultiBufferWorkspaceTraits::MultiBufferWorkspaceTraits(std::initializer_list<siz
     }
 }
 
-size_t MultiBufferWorkspaceTraits::GetSize() const { return v_offset[v_offset.size() - 2] + v_size.back(); }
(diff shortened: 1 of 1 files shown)


shaojiewang:
Issue #1458 is originally posted by @atamazov . So also needs Artem's review

atamazov (dismissed):
LGTM!

src/buffer_info.cpp
```diff
@@ -162,7 +162,10 @@ MultiBufferWorkspaceTraits::MultiBufferWorkspaceTraits(std::initializer_list<siz
     }
 }
 
-size_t MultiBufferWorkspaceTraits::GetSize() const { return v_offset.back(); }
+size_t MultiBufferWorkspaceTraits::GetSize() const
+{
+    return v_offset[v_offset.size() - 2] + v_size.back();
```
atamazov, line 167: [Recommendation] Maybe you'll find this nicer
```suggestion
    return (&v_offset.back())[-1] + v_size.back();
```

atamazov (approved):
LGTM

carlushuang (approved):
LGTM

junliume (approved):

PR #1461, opened 2022-03-14
[Tuning][MI100][MI200] Superbench

cderb:
tuning update for superbench, bkc 9740, miopen fa2d5895b
includes tuning results from #1459 for MI200 110
If merged first #1459 should be deprecated. With a new PR for MI200 104 update.

PR #1464, opened 2022-03-18
[Tuning][MI200] regression patching swdev 326970 part1

cderb:
patching MI200 110 for regressions

write-over for mi200 110 in #1461


cderb:
closing this in favor of another tuning

PR #1474, opened 2022-03-23
[MLIR][Tuning] Revise tuning search space and perf DB entries for MLIR XDLOPS solver.

whchung:
- all data types: reduce KPerBlock range to 4, 8. 1 and 2 were never used after empirical tuning, and would be invalid especially for fp16 as the minimal KPerBlock needs to be 4 to be fit into MFMA instruction.
- fp32: 
  - KPACK values could be 1, 2, 4 for forward convolution.
  - KPACK values could be 1, 2 for backward weight convolution.
  - KPACK values would be 1 for backward data convolution.
- fp16: KPACK values could be 1, 4, 8.
  - 2 is deliberately avoided because existing MFMA instructions takes a vector of 4 fp16 values. Going 2 would require additional unpacking and packing vectors in MLIR, which is not implemented as of yet, and likely will never be implemented in the near future.
- other data types: fix KPACK value be 1 for now.

This allows all resnet50 configs be tuned for fp32/fp16 after KPACK is enabled in backward weight convolution. Also it allows KPACK = 1/2 in various fp32 configs be selected as the fastest config. And it get rids of KPACK = 8 in fp32 path where all of them would be rejected by MLIR lowering pipeline.

MLIR commit pointer hasn't been modified in this PR as the changed tuning search space is also applicable to existing pointer.

perf DB entries for `fp32`/`wrw`/`gfx908` have also been eliminated as the default values contain empirically verified invalid `KPACK=4` which could cause MLIR solver render intermittently incorrect kernels in some configs.


src/solver/conv_mlir_igemm_fwd_xdlops.cpp
```diff
@@ -139,14 +139,33 @@ bool PerformanceConvMlirIgemmXdlops::SetNextValue(const ConvolutionContext& /*co
...
+                    GemmKPACKSize = 4;
+                }
+                break;
+            }
+        }
+        else
+        {
+            if(!NextTwoPower<1, 1>(GemmKPACKSize))
+                break;
+        }
```
atamazov, line 168: [Recommendation] To get rid of ambiguity
```suggestion
        ; // Nop. Leave GemmKPACKSize equal to 1.
```

atamazov (approved):
@whchung Does it require re-tuning of Forward MLIR XDLOPS solver?

atamazov:
@cderb @JehandadKhan @junliume **_This PR requires re-tuning of Forward MLIR XDLOPS solver._**

whchung:
@atamazov Thanks for the feedback. I'll revise the PR after I have a fully passing tuning run on MLIR CI first. This proposed revised search space have been passing on my local dev node, but not on MLIR CI. I'm still digging into why.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
whchung: Changed in the very recent commit.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
```diff
@@ -139,14 +151,35 @@ bool PerformanceConvMlirIgemmXdlops::SetNextValue(const ConvolutionContext& /*co
...
+            }
+            else if(config.direction.IsBackwardData())
+            {
+                if(!NextTwoPower<1, 1>(GemmKPACKSize))
+                    break;
+            }
+        }
+        else if(config.IsFp16())
+        {
+            if(!Next_1_4_8(GemmKPACKSize))
```
jerryyin, line 180: Could you add a comment here to link to the MLIR ticket and mention that this aim to be reverted?

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
```diff
@@ -139,14 +151,35 @@ bool PerformanceConvMlirIgemmXdlops::SetNextValue(const ConvolutionContext& /*co
...
+                    break;
+            }
+            else if(config.direction.IsBackwardWrW())
+            {
+                if(!NextTwoPower<1, 2>(GemmKPACKSize))
+                    break;
+            }
+            else if(config.direction.IsBackwardData())
+            {
+                if(!NextTwoPower<1, 1>(GemmKPACKSize))
```
jerryyin, line 174: I thought you have started another PR to enable kpack for bwd, if so, this condition is out-of-date?

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
```diff
@@ -139,14 +151,35 @@ bool PerformanceConvMlirIgemmXdlops::SetNextValue(const ConvolutionContext& /*co
...
             break;
         if(!NextTwoPower<16, 128>(GemmNPerWave))
             break;
-        if(!NextTwoPower<4, 8>(GemmKPACKSize))
-            break;
+        if(config.IsFp32())
+        {
+            if(config.direction.IsForward())
+            {
+                if(!NextTwoPower<1, 4>(GemmKPACKSize))
```
jerryyin, line 164: What about 8?

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
```diff
@@ -139,14 +151,35 @@ bool PerformanceConvMlirIgemmXdlops::SetNextValue(const ConvolutionContext& /*co
...
+                    break;
+            }
+            else if(config.direction.IsBackwardWrW())
+            {
+                if(!NextTwoPower<1, 2>(GemmKPACKSize))
```
jerryyin, line 169: What about 4/8?

jerryyin (commented):
I don't think MIOpen has yet branched on ROCm 5.2. Is that right? If so, we need to wait till after MIOpen branch to ship this PR.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
whchung: This PR does not aim to support enable KPACK for bwd yet, therefore KPACK is fixed as 1 for now.

Another PR will be filed once the MLIR PR gets merged.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
whchung: 4 has known intermittent issue.
8 is not valid.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
whchung: This does not aim to be reverted. I assume you are talking about fp32 wrw?

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
jerryyin: Gotcha, sounds good.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
jerryyin: That's right I didn't notice it is under fp32 condition, but what about 2? I do notice that 2 was previously not included but we can potentially include in the future?

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
jerryyin: Make sense.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
whchung: `NextTwoPower<1,4>` covers [1, 2, 4].

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
jerryyin: >  I assume you are talking about fp32 wrw?

Right, I got myself confused. Mainly because we skipped 2.

I think it will help explain a lot if you can put the PR description on why not include 2 here.

whchung: Revised the description on why KPACK=2 is not used for fp16.

jerryyin (dismissed):
LGTM

@junliume Please do not merge until ROCm 5.2 branch.

atamazov (approved):
LGTM but please fix merge conflicts.

atamazov:
> @junliume Please do not merge until ROCm 5.2 branch.

This can be expressed explicitly by setting the `ON_HOLD` label or by converting to draft.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
```diff
@@ -38,7 +38,19 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_MLIR_IGEMM_FWD_XDLOPS)
...
+    if(v == 1)
+        v = 4;
+    else if(v == 4)
+        v = 8;
+    else // if(v == 8)
+        v = 1;
+    return v == 1; // overflow
+}
+
+const PerformanceConvMlirIgemmXdlops& PerformanceConvMlirIgemmXdlops::MlirHeuristicInitRequest()
```
junliume, line 53: @whchung this seems to be an issue:
```
error: return type of out-of-line definition of 'miopen::solver::PerformanceConvMlirIgemmXdlops::MlirHeuristicInitRequest' differs from that in the declaration

[2022-04-07T21:06:41.304Z] const PerformanceConvMlirIgemmXdlops& PerformanceConvMlirIgemmXdlops::MlirHeuristicInitRequest()

[2022-04-07T21:06:41.304Z]       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                 ^

[2022-04-07T21:06:41.304Z] /var/jenkins/workspace/_mlir-revise-tuning-search-space/src/include/miopen/solver.hpp:949:44: note: previous definition is here

[2022-04-07T21:06:41.304Z]     static PerformanceConvMlirIgemmXdlops& MlirHeuristicInitRequest()

[2022-04-07T21:06:41.304Z]            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^

jerryyin:
@whchung To what degree is this PR still necessary:

For the code changes:
 - My understanding is that you have figured out how to revert the hotfix from MLIR. Do you plan to have a PR for it?
 - So even without this PR changes, nothing should fail 
 - Looks like the impact of changes is more of a matter of optimization to remove some tuning parameters. (But in reality not a big deal as long as we can reject inapplicable parameters.)

For the db changes:
 - Seems like the db get updated, is the new db still a problem?

I'm asking also because I'm wondering if we are ready to bring MLIR side of CI back to the `develop` branch? Otherwise we are sort of stuck with an older MIOpen `develop` which may not the best thing for our regression tests/ bring in new features.

junliume:
> @whchung To what degree is this PR still necessary:
> 
> For the code changes:
> 
> * My understanding is that you have figured out how to revert the hotfix from MLIR. Do you plan to have a PR for it?
> * So even without this PR changes, nothing should fail
> * Looks like the impact of changes is more of a matter of optimization to remove some tuning parameters. (But in reality not a big deal as long as we can reject inapplicable parameters.)
> 
> For the db changes:
> 
> * Seems like the db get updated, is the new db still a problem?
> 
> I'm asking also because I'm wondering if we are ready to bring MLIR side of CI back to the `develop` branch? Otherwise we are sort of stuck with an older MIOpen `develop` which may not the best thing for our regression tests/ bring in new features.

@jerryyin  ROCm 5.2 is cut, we should revisit this PR to see if it can be merged into develop

whchung:
@jerryyin / @junliume Sorry for late response. MLIR PR 586 was merged in yesterday ( https://github.com/ROCmSoftwarePlatform/llvm-project-mlir/commit/f26869d33a5052496889497c5fbac529c1bf4260 ).

Before the hiatus, I had already marked this PR as ROCm 5.3. We still need the PR to trim the tuning search space, but it's optional. And it's likely we don't need explicitly supply a modified perf DB.

So I'd suggest the followings:
- submit another PR to MIOpen to modify the tracking commit of MLIR : @jerryyin ?
- revise this PR : @whchung 
  - remove the MLIR tracking commit from this PR.
  - remove the explicitly-supplied perfDB.
  - preserve the tuning search space trimming logic.

jerryyin:
@whchung List look good to me.

> submit another PR to MIOpen to modify the tracking commit of MLIR : @jerryyin 

For sure, let me handle that. My immediate next PR will require this anyway.

Meanwhile could you also help move our CI's tracking branch to MIOpen's `develop`?

junliume:
@whchung and @jerryyin Do we need this in 5.3 release? If so could you take care of the conflicts? Thanks!

junliume:
@jerryyin and @whchung do you still plan to proceed with this PR? Please merge the latest `develop` and resolve the conflicts. Thanks!

atamazov:
> @jerryyin and @whchung do you still plan to proceed with this PR? Please merge the latest `develop` and resolve the conflicts. Thanks!

@junliume Can you please convert this to draft until the above is done. Thanks!

PR #1475, opened 2022-03-24
[BN] Fixed numeric checks for valid nullptr buffers

### LINKED ISSUE #1442: [BN] checkNumericsInput is called with nullptr
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/df658f382c5af02133836a8fa5700a23b58c2ca0/src/ocl/batchnormocl.cpp#L171-L172

These numeric checks must not be performed when `(estimatedMean == nullptr)` or `(estimatedVariance == nullptr)`.

I highly recommend checking for similar issues in the rest of BN code.

DrizztDoUrden:
As suggested, added conditions to every check where a buffer may be nullptr.

Fixes #1442


atamazov (approved):
LGTM

junliume (approved):

Issue #658, opened 2020-12-19
Build fails with boost 1.75


tpkessler:
When compiling MIOpen 4.0.0 from source with boost 1.75 on Arch Linux I get multiple error messages regarding `noinline`, for instance:

```
/usr/include/boost/container/vector.hpp:2722:4: error: type name does not allow function specifier to be specified
/usr/include/boost/container/detail/workaround.hpp:116:37: note: expanded from macro 'BOOST_CONTAINER_NOINLINE'
   #define BOOST_CONTAINER_NOINLINE BOOST_NOINLINE#      define BOOST_NOINLINE __attribute__ ((__noinline__))

                                    ^                                             ^

/usr/include/boost/config/detail/suffix.hpp:626:46: /opt/rocm/hip/include/hip/hcc_detail/host_defines.h:note75: :37expanded from macro 'BOOST_NOINLINE': 
note: expanded from macro '__noinline__'
#      define BOOST_NOINLINE __attribute__ ((__noinline__))
                                             ^
/opt/rocm/hip/include/hip/hcc_detail/host_defines.h:75:37: note: expanded from macro '__noinline__'
#define __noinline__ __attribute__((noinline))
                                    ^
#define __noinline__ __attribute__((noinline))

```

MIOpen compiles fine however when I downgrade boost to 1.72.

daniellowell:
Our requirements pin the boost version 1.72. I'll update our documentation to reflect that.

You can install the minimal requirements for MIOpen using the command:
`cmake -P install_deps.cmake --minimum --prefix /some/local/dir`

and then configure with:
`CXX=/opt/rocm/llvm/bin/clang++ cmake -DMIOPEN_BACKEND=HIP -DCMAKE_PREFIX_PATH="/some/local/dir" ..`

See more information here:
https://rocmsoftwareplatform.github.io/MIOpen/doc/html/install.html#installing-minimum-dependencies-in-rocm-environment

daniellowell:
Documentation has been updated in our develop branch. Please feel free to reopen if you have additional questions.

pfultz2:
Actually, this is a bug in HIP, as its defining a `__noinline__` macro and boost is trying to use the attribute `__noinline__` causing the error. I think we should file a ticket with the HIP team.

tpkessler:
> Documentation has been updated in our develop branch.

Thank you! Installing `boost 1.72` via `cmake` does the trick for me.

> Actually, this is a bug in HIP

[...]

PR #1490, opened 2022-03-30
Fix Headers for Boost v1.78+

### LINKED ISSUE #1489: Boost dependency can be (mostly) upgraded
Previously, there was an incompatibility between Boost's `noinline` and what was provided within the HIP compiler which was initially reported in #658.  Since then, the Boost library has been patched through boostorg/config#393 which was included in Boost version 1.78.  As a result, I have tried again to see if I can avoid having to recompile Boost and use the system version and it mostly works.

In particular, the compilation works fine for all files except for the following 9 files:
- `src/find_db.cpp`
- `src/binary_cache.cpp`
- `src/sqlite_db.cpp`
- `src/solver/conv_asm_3x3u.cpp`
- `src/ocl/gcn_asm_utils.cpp`
- `src/ocl/mloNeuron.cpp`
- `src/ocl/mloPooling.cpp`
- `src/ocl/mloNorm.cpp`
- `src/hip/hip_build_utils.cpp`

The error message is similar to the original issue:

```text
In file included from /tmp/makepkg/miopen-hip/src/MIOpen-rocm-5.0.2/src/find_db.cpp:29:
In file included from /tmp/makepkg/miopen-hip/src/MIOpen-rocm-5.0.2/src/include/miopen/handle.hpp:30:
In file included from /tmp/makepkg/miopen-hip/src/MIOpen-rocm-5.0.2/src/include/miopen/kernel_info.hpp:33:
In file included from /tmp/makepkg/miopen-hip/src/MIOpen-rocm-5.0.2/src/include/miopen/kernel.hpp:53:
In file included from /tmp/makepkg/miopen-hip/src/MIOpen-rocm-5.0.2/src/include/miopen/hipoc_kernel.hpp:34:
In file included from /tmp/makepkg/miopen-hip/src/MIOpen-rocm-5.0.2/src/include/miopen/op_kernel_args.hpp:8:
In file included from /usr/include/boost/container/small_vector.hpp:27:
/usr/include/boost/container/vector.hpp:2754:4: error: use of undeclared identifier 'noinline'; did you mean 'inline'?
   BOOST_CONTAINER_NOINLINE iterator priv_insert_forward_range_no_capacity
   ^
/usr/include/boost/container/detail/workaround.hpp:116:37: note: expanded from macro 'BOOST_CONTAINER_NOINLINE'
   #define BOOST_CONTAINER_NOINLINE BOOST_NOINLINE
                                    ^
/usr/include/boost/config/detail/suffix.hpp:639:46: note: expanded from macro 'BOOST_NOINLINE'
#      define BOOST_NOINLINE __attribute__ ((__noinline__))
                                             ^
/opt/rocm/hip/../include/hip/amd_detail/host_defines.h:50:37: note: expanded from macro '__noinline__'
#define __noinline__ __attribute__((noinline))
                                    ^
```

and in all of the above cases, the addition of `#include <hip/hip_version.h>` as an additional header in each `.cpp` file with no other change to the code fixes the compilation error.  I did also try adding this header to `hipoc_kernel.hpp` but that did not resolve the issue.

I'm happy to create a PR to address this issue, though I would like to understand why adding the missing `hip_version.h` header to `hipoc_kernel.hpp` does not appear to resolve the issue.

I have checked this against the latest develop commit, and the issue (at least when compiling `find_db.cpp`) is still there.

JP-Ellis:
Boost's issues with `noinline` (see #658) have been fixed upstream in
boostorg/config#393 which was included in v1.78 of Boost; however, in order for
it to work properly Boost library headers must be included after we include
headers which define `HIP_VERSION` (which come through `kernel.hpp` and headers
therein).

Whilst most files worked fine, there were a small number which included a Boost
library before `HIP_VERSION` was defined causing a compilation error.

This will also allow MIOpen to be built against the system Boost instead of
requiring it to be compiled specifically for MIOpen.

Fixes #1489.

### DIFF
--- a/src/include/miopen/mlo_internal.hpp
+++ b/src/include/miopen/mlo_internal.hpp
@@ -56,6 +56,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #include <miopen/config.h>
+#include <miopen/kernel.hpp>
 
 #if MIOPEN_BACKEND_OPENCL
 #include <miopen/oclkernel.hpp>
--- a/src/include/miopen/binary_cache.hpp
+++ b/src/include/miopen/binary_cache.hpp
@@ -28,6 +28,7 @@
 #define GUARD_MLOPEN_BINARY_CACHE_HPP
 
 #include <miopen/config.h>
+#include <miopen/kernel.hpp>
 #include <miopen/target_properties.hpp>
 #include <boost/filesystem/path.hpp>
 #include <string>
--- a/src/include/miopen/comgr.hpp
+++ b/src/include/miopen/comgr.hpp
@@ -27,6 +27,7 @@
 #define GUARD_COMGR_HPP
 
 #include <miopen/config.h>
+#include <miopen/kernel.hpp>
 #if MIOPEN_USE_COMGR
 
 #include <miopen/target_properties.hpp>
--- a/src/include/miopen/find_db.hpp
+++ b/src/include/miopen/find_db.hpp
@@ -27,6 +27,7 @@
 #ifndef GUARD_MIOPEN_FIND_DB_HPP_
 #define GUARD_MIOPEN_FIND_DB_HPP_
 
+#include <miopen/kernel.hpp>
 #include <miopen/db.hpp>
 #include <miopen/db_path.hpp>
 #include <miopen/db_record.hpp>
--- a/src/include/miopen/gcn_asm_utils.hpp
+++ b/src/include/miopen/gcn_asm_utils.hpp
@@ -27,6 +27,7 @@
 #define GCN_ASM_UTILS_H
 
 #include <miopen/config.h>
+#include <miopen/kernel.hpp>
 #include <miopen/target_properties.hpp>
 #include <sstream>
 #include <string>
--- a/src/include/miopen/hip_build_utils.hpp
+++ b/src/include/miopen/hip_build_utils.hpp
@@ -27,6 +27,7 @@
 #define MIOPEN_GUARD_MLOPEN_HIP_BUILD_UTILS_HPP
 
 #include <miopen/config.h>
+#include <miopen/kernel.hpp>
 #include <miopen/target_properties.hpp>
 #include <miopen/kernel.hpp>
 #include <miopen/tmp_dir.hpp>
--- a/src/include/miopen/mlir_build.hpp
+++ b/src/include/miopen/mlir_build.hpp
@@ -27,6 +27,7 @@
 #define MIOPEN_GUARD_MLIR_BUILD_HPP
 
 #include <miopen/config.h>
+#include <miopen/kernel.hpp>
 #if MIOPEN_USE_MLIR
 
 #include <miopen/target_properties.hpp>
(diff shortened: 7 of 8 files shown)


src/include/miopen/mlo_internal.hpp
```diff
@@ -500,8 +501,8 @@ struct mlo_construct_activ_lrn_pooling_common : mlo_construct_base
 
         int data_len = miopen::GetTypeSize(data_type);
         size_t size  = (layout == "NCHW")
-                          ? batch * channels * depth * height * width * data_len
-                          : batch * batch_stride * channel_stride * stride * w_stride * data_len;
+                           ? batch * channels * depth * height * width * data_len
+                           : batch * batch_stride * channel_stride * stride * w_stride * data_len;
```
atamazov, line 505: Please revert this.

src/include/miopen/mlo_internal.hpp
```diff
@@ -542,8 +543,8 @@ struct mlo_construct_activ_lrn_pooling_common : mlo_construct_base
...
```
atamazov, line 547: Ditto.

atamazov (approved):
No objections other than couple of formatting issues.

junliume:
@JP-Ellis  @atamazov  It's on roadmap that when we update to C++17 standard we may not need boost after all.

JP-Ellis:
Sorry for the extra formatting changes.  Perhaps that was the result of an automatic format-on-save.  The changes in the PR now only add new headers.

src/include/miopen/mlo_internal.hpp
JP-Ellis: Fixed in latest commit.

atamazov:
@JP-Ellis Nada problemas))

atamazov (approved):

atamazov:
> Boost's issues with `noinline` (see #658) have been fixed upstream in [boostorg/config#393](https://github.com/boostorg/config/pull/393) which was included in v1.78 of Boost; 

The reason of the above problem is incorrect fix in BOOST (https://github.com/boostorg/config/pull/393#discussion_r858808126). 

> however, in order for it to work properly Boost library headers must be included after we include headers which define `HIP_VERSION` (which come through `kernel.hpp` and headers therein).

kernep.hpp is too much, -- it imports too may declarations to the places where such declarations are not really required.

The recommendation is adding a workaround to `config.h.in` like this:
```C++
/// Workaround for https://github.com/boostorg/config/issues/392#issuecomment-1109889533
/// \todo Remove the W/A as soon we switch to the properly fixed boost.
#if MIOPEN_BACKEND_HIP
#include <hip/hip_version.h>
#endif
```
And then include config.h instead of kernel.hpp where required.

The W/A is to be removed when we switch to the properly fixed boost.

atamazov (changes requested):
Let's add W/A into config.h and revert (almost all) other changes. Please see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1490#issuecomment-1109928102 for more info.

junliume:
Hi @JP-Ellis, can you make the recommended changes? I have tested on dev machines and they are working properly. We just want to let it run through the whole set of CI first.
If you do not have time, I can try to create another PR for the purpose. Thanks!
CC: @atamazov

junliume:
Close this one and replaced with #1543

PR #1492, opened 2022-03-31
[SWDEV-329485] Remove clang-ocl dependency for miopen-hip

solaiys:
When HIPRTC is enabled the runtime compilation is performed through
hiprtc and comgr apis. So rocm-clang-ocl dependency can be removed.

However if HIPRTC is disabled it adds the clang-ocl dependency
to the miopen-hip package conditionally..

Signed-off-by: Saravanan Solaiyappan <saravanan.solaiyappan@amd.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,7 +253,7 @@ if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_B
         message(STATUS "OpenCL compiler: ${HIP_OC_COMPILER}")
         set(HIP_OC_COMPILER "${HIP_OC_COMPILER}")
     else()
-        message(FATAL_ERROR "OpenCL compiler not found")
+        message(STATUS "OpenCL compiler not found")
     endif()
 
         # Hcc's clang always defines __HCC__ even when not using hcc driver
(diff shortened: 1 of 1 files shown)


DrizztDoUrden (approved):

JehandadKhan (approved):

junliume (approved):

CMakeLists.txt
```diff
@@ -420,14 +420,24 @@ if(MIOPEN_USE_ROCBLAS)
...
 
-set(CPACK_DEBIAN_PACKAGE_DEPENDS "${MIOPEN_PACKAGE_REQS}, rocm-opencl-dev")
-set(CPACK_RPM_PACKAGE_REQUIRES "${MIOPEN_PACKAGE_REQS}, rocm-opencl-devel")
-
-# Make backends explicitly conflict
 if(MIOPEN_BACKEND STREQUAL "HIP")
+    # In HIP backend, there is a posibility of HIPRTC disabled.
+    # In this case add the clang-ocl as dependency for runtime kernel compilation
+    if(NOT MIOPEN_USE_HIPRTC )
+        set(MIOPEN_PACKAGE_REQS "${MIOPEN_PACKAGE_REQS}, rocm-clang-ocl")
```
atamazov, line 427: > When HIPRTC is enabled the runtime compilation is performed through hiprtc and comgr apis. 

The above is true, however please note:
- clang-ocl is used to build OpenCL kernels. 
- HIPRTC is used to build HIP kernels. Therefore, disabling HIPRTC does not affect OpenCL build path.
- HIPRTC requires COMgr to be enabled.
- Whenever COMgr is enabled, ASM and OpenCL go via COMgr.

This means that we do not need clang-ocl when we have COMgr. This is not a bug, but let's better fix this.

atamazov (commented):
Let's improve this patch.

Issue #1495, opened 2022-03-31
conv bwd-data error


shaojiewang:
conv bwd-data got error when hi=wi=2, stride=2. Error case is:
./bin/MIOpenDriver conv -n 256 -c 2048 -H 2 -W 2 -k 1024 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 2 -t 1 -w 1 .

PR #1496, opened 2022-03-31
[Bug fix][SWDEV-329914] fix stride compute bug in asm igemm bwd fwd and wrw

shaojiewang:
To solve #1495 .
In bwd solver, we re-compute the ```stride_h``` and ```stride_w``` by the following code:
```
const auto stride_h   = ctx.in_height > 1 ? ctx.kernel_stride_h : 1;
const auto stride_w   = ctx.in_width > 1 ? ctx.kernel_stride_w : 1;
```
However, MIOpen switches input size and output size for backward path. So we need the following code to get the right params. 
```
const auto stride_h   = ctx.out_height > 1 ? ctx.kernel_stride_h : 1;
const auto stride_w   = ctx.out_width > 1 ? ctx.kernel_stride_w : 1;
```

test case:
./bin/MIOpenDriver conv -n 256 -c 2048 -H 2 -W 2 -k 1024 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 2 -t 1 -w 1

### DIFF
--- a/src/conv/invokers/impl_gemm_dynamic.cpp
+++ b/src/conv/invokers/impl_gemm_dynamic.cpp
@@ -241,8 +241,8 @@ MakeImplGemmDynamicBackwardDataInvokerFactory<solver::TunableImplicitGemmGTCDyna
     int c                    = conv_problem.GetOutChannels();
     int ho                   = conv_problem.GetInHeight();
     int wo                   = conv_problem.GetInWidth();
-    int stride_h             = conv_problem.GetInHeight() > 1 ? conv_problem.GetKernelStrideH() : 1;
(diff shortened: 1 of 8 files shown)


junliume (approved):
Please also plan a PR picking this fix into release/rocm-rel-5.1-staging Thanks!

carlushuang (commented):
fwd/wrw NCHW/NHWC also need change the stride calculation like bwd

junliume (approved):

atamazov:
@junliume 
> Please also plan a PR picking this fix into release/rocm-rel-5.1-staging Thanks!

If/when cherry-picking is trivial, the Integrator can do it himself/herself, and no PR is needed. This is even better as commit would inherit the description and PR number from the original commit.

PR #1503, opened 2022-04-05
[API addition] Find 2.0

DrizztDoUrden:
This is a basic implementation of find 2.0 API.

The behaviour should be considered temporarily limited and to be extended. It basically routes find to 1.0 implementation for now. Later it should be done the other way around to extend the 2.0 functionality. Run works like immediate mode. Serialization is versioned, but basic. It doesn't store binary, so would use caches on the run call. If find was not skipped as usual it would only use invoker cache though. Performance should not differ from version 1.0 overall, when serialization has not been used. Otherwise, some benefit may be achieved by manual caching of solution and skipping find calls.

Future work:
Implement binary serialization for cache skips.
`miopenGetSolutionsNumber` - not required right now. May be added when needed. Right now, it is not trivial to make.
Parametrized serialization/deserialization - not required for basic use. May be added later as a separate pair of functions. The current pair would just call parametrized version with default options.


DrizztDoUrden:
Maybe we should add something like this?
```C
miopenStatus_t miopenGetSolutionsNumber(miopenHandle_t handle,
    miopenProblem_t problem,
    miopenSearchOptions_t options
    size_t* numSolutions);
```

src/solution.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+namespace miopen {
+
+std::size_t Solution::GetSize() const
+{
+    MIOPEN_THROW(miopenStatusNotImplemented);
+    return 0;
+}
+
+void Solution::Save(char* data) const
```
junliume, line 37: error: function 'Save' could be declared with attribute 'noreturn' [-Werror,-Wmissing-noreturn]

src/solution.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    std::ignore = data;
+
+    MIOPEN_THROW(miopenStatusNotImplemented);
+}
+
+void Solution::Load(const char* data, std::size_t size)
```
junliume, line 44: same as above

src/solution.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    std::ignore = data;
+    std::ignore = size;
+
+    MIOPEN_THROW(miopenStatusNotImplemented);
+}
+
+void Solution::Run(const Handle& handle,
```
junliume, line 52: same above

DrizztDoUrden:
@JehandadKhan This commit should work without solution serialization: [`b9ffa0b` (#1503)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1503/commits/b9ffa0b717ad734fe888730222f73f81002cc051)
I haven't done much testing yet though. I'm working on basic serialization right now. ~60% complete.

DrizztDoUrden:
Serialization should be working now.

src/include/miopen/problem.hpp
```diff
@@ -0,0 +1,130 @@
...
+                                                  : problem.operator_descriptor->GetPrimitive();
+
+        stream << primitive;
+
+        switch(primitive)
+        {
+        case solver::Primitive::Convolution:
+            if(stream.IsDeserializing())
+                problem.operator_descriptor = std::make_shared<ConvolutionDescriptor>();
+            stream << *static_cast<ConvolutionDescriptor*>(problem.operator_descriptor.get());
```
junliume, line 99: OCL tidy complaints that
```
rror: do not use static_cast to downcast from a base to a derived class; use dynamic_cast instead [cppcoreguidelines-pro-type-static-cast-downcast,-warnings-as-errors]
[2022-04-07T20:32:09.015Z]             stream << *static_cast<ConvolutionDescriptor*>(problem.operator_descriptor.get());
[2022-04-07T20:32:09.015Z]                        ^~~~~~~~~~~
[2022-04-07T20:32:09.015Z]                        dynamic_cast

File: src/problem.cpp [hunk 5]
```diff
@@ -0,0 +1,214 @@
...
+    const auto found = tensor_descriptors.find(name);
+    if(found == tensor_descriptors.end())
+        MIOPEN_THROW(miopenStatusInvalidValue,
+                     "Problem is missing " + name_str + " tensor descriptor.");
+    return found->second;
+}
+
+conv::ProblemDescription Problem::AsConvolution() const
+{
+    const auto& conv_desc = *static_cast<ConvolutionDescriptor*>(operator_descriptor.get());
```
junliume, line 86: Same above

src/problem.cpp
```diff
@@ -0,0 +1,214 @@
...
+               : conv::ProblemDescription(y_desc, w_desc, x_desc, conv_desc, conv_dir);
+}
+
+std::vector<Solution> Problem::FindConvSolutions(Handle& handle,
+                                                 const SearchOptions& options,
+                                                 std::size_t max_solutions) const
+{
+    auto ret = std::vector<Solution>{};
+
+    const auto& conv_desc = *static_cast<ConvolutionDescriptor*>(operator_descriptor.get());
```
junliume, line 107: same above

DrizztDoUrden:
Updated description. Ready for review. Functionally working.
I am thinking on making a dedicated test pass like conv2d/conv3d tests. Right now it is a one problem basic functionality test.

DrizztDoUrden:
A question: do we want to merge it in a state when it has different (limited to find 1.0 selected solvers) behaviour from the plan (all solvers)? Or have we lifted that plan and further changes would be more of reorganization and optimization type?

include/miopen/miopen.h
```diff
@@ -4666,6 +4666,120 @@ miopenReduceTensor(miopenHandle_t handle,
...
+// Universal problem should have named tensor fields rather than ordered.
+// Previously, different directions had different input orders, which may lead to confusion.
+// Implicitly ordered fields would also lead to a whole lot of magical ids all over the internals.
+miopenStatus_t miopenSetProblemTensorDescriptor(miopenProblem_t problem,
+                                                miopenTensorName_t name,
+                                                const miopenTensorDescriptor_t descriptor);
+
+miopenStatus_t miopenSetProblemOperatorDescriptor(miopenProblem_t problem,
+                                                  const void* operatorDesc,
+                                                  miopenProblemDirection_t direction);
```
pfultz2, line 4725: What is this function doing? Is this supposed to set the convolution descriptor? It would be more type-safe to have a function with the correct type such as `miopenSetProblemConvolutionDescriptor`. Although, I think its better to do this during construction of the problem.

include/miopen/miopen.h
```diff
@@ -4666,6 +4666,120 @@ miopenReduceTensor(miopenHandle_t handle,
...
+ * future, this can support setting different find mode options.
+ */
+MIOPEN_DECLARE_OBJECT(miopenSearchOptions);
+
+miopenStatus_t miopenCreateSearchOptions(miopenSearchOptions_t* options);
+miopenStatus_t miopenDestroySearchOptions(miopenSearchOptions_t options);
+miopenStatus_t miopenSetSearchOption(miopenSearchOptions_t options,
+                                     miopenSearchOptionName_t optionName,
+                                     size_t valueSize,
+                                     void* value);
```
pfultz2: You should use different functions for each attribute as using `void*` is error-prone.

pfultz2: We could also have search options in the future that takes multiple values.

include/miopen/miopen.h
```diff
@@ -4666,6 +4666,120 @@ miopenReduceTensor(miopenHandle_t handle,
...
+
+miopenStatus_t miopenSaveSolution(miopenSolution_t solution, char* data);
+
+miopenStatus_t miopenGetSolutionSize(miopenSolution_t solution, size_t* size);
+
+miopenStatus_t miopenGetSolutionAttribute(miopenSolution_t solution,
+                                          miopenSolutionAttribute_t solutionAttribute,
+                                          size_t valueSize,
+                                          void* value,
+                                          size_t* valueSizeRet);
```
pfultz2, line 4781: Same here. There should be a function for each attribute.

src/include/miopen/convolution.hpp
```diff
@@ -99,9 +109,18 @@ struct ConvolutionAttribute
...
+    {
+        stream << conv.gfx90aFp16alt;
+        return stream;
+    }
 };
 
-struct ConvolutionDescriptor : miopenConvolutionDescriptor
+struct ConvolutionDescriptor
+    : miopenConvolutionDescriptor,
+      OperatorDescriptorImpl<ConvolutionDescriptor, solver::Primitive::Convolution>
```
pfultz2, line 123: This should use the `Serializable` class, with a `Visit` method, like how did in #1290: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1290/files#diff-082c9699504013c0a45dd886dccf551d0b11f16b16677c597deff1471ee5e84eR138

src/include/miopen/binary_serialization.hpp
```diff
@@ -0,0 +1,229 @@
...
+public:
+    bool IsSerializing() const { return true; }
+    bool IsDeserializing() const { return false; }
+
+    BinarySerializationStream(char* buffer) : position(buffer) {}
+
+    template <class Type, std::enable_if_t<SerializationCriteria<Type>{}, bool> = true>
+    BinarySerializationStream& operator<<(const Type& value)
+    {
+        std::memcpy(position, &value, sizeof(Type));
```
pfultz2, line 88: This is not portable. We should not be creating our own binary serialization(which could expose security vulnerabilities as well).

Instead, we should use msgpack to serialize to binaries. If we use nlohmann/json, It also can serialize STL containers, and it has an easy interface to extend to your own data structures so we dont need to write our own serialization framework.

Also, msgpack makes it easier for us to read these buffers for debugging purposes.

In #1290, I added nlohmann/json as a dependency but it is header-only so if you dont like a dependency you can just copy the single header over to our repo.

src/include/miopen/convolution.hpp
```diff
@@ -437,6 +456,23 @@ struct ConvolutionDescriptor : miopenConvolutionDescriptor
                               miopenConvSolution_t* solutions) const;
 
     std::size_t GetSolutionCountFallback(Handle& handle, const ProblemDescription& problem) const;
+
+    template <class Stream, std::enable_if_t<IsBinarySerializationRelated<Stream>{}, bool> = true>
+    friend Stream& operator<<(Stream& stream, ConvolutionDescriptor& conv)
```
pfultz2, line 461: If this is serializing then `ConvolutionDescriptor` can be `const`. That is why the `Visit` takes templated `Self` parameter so it can be used for both.

src/include/miopen/operator_descriptor.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+template <class Derived, solver::Primitive primitive>
+struct OperatorDescriptorImpl : OperatorDescriptor
+{
+    solver::Primitive GetPrimitive() const final { return primitive; }
+
+    OperatorDescriptor* Clone() const final
+    {
+        return static_cast<OperatorDescriptor*>(
+            new Derived{reinterpret_cast<const Derived&>(*this)});
```
pfultz2, line 48: Whats the purpose of this?

src/include/miopen/operator_descriptor.hpp
```diff
@@ -0,0 +1,52 @@
...
+    virtual solver::Primitive GetPrimitive() const = 0;
+    virtual OperatorDescriptor* Clone() const      = 0;
+};
+
+template <class Derived, solver::Primitive primitive>
+struct OperatorDescriptorImpl : OperatorDescriptor
+{
+    solver::Primitive GetPrimitive() const final { return primitive; }
+
+    OperatorDescriptor* Clone() const final
```
pfultz2, line 45: This leaks memory as `OperatorDescriptor*` does not own the memory.

src/include/miopen/problem.hpp
```diff
@@ -0,0 +1,130 @@
...
+        default: MIOPEN_THROW(miopenStatusNotImplemented);
+        }
+
+        return stream;
+    }
+
+private:
+    miopenProblemDirection_t direction = miopenProblemDirectionForward;
+    std::unordered_map<miopenTensorName_t, TensorDescriptor> tensor_descriptors;
+    std::shared_ptr<OperatorDescriptor> operator_descriptor;
```
pfultz2, line 114: I think it would be better to store the descriptor as a variant. The number of descriptors used is close(whereas the serialized types is more open). 

Using a variant will simplify serialization as we dont need to pump serialization through some erased type which is more difficult.

src/solution.cpp
pfultz2: I can actually help implement Load/Save.

pfultz2: I think it would be better to get the API working correctly first, and then we can tackle Load/Save in a later PR.

src/include/miopen/convolution.hpp
pfultz2: Maybe we dont want to make all our descriptors inherit from `Serializable` as the `Serialize` methods are in the headers, so it could impact compile-time.

Just having a `Visit` method should be enough to implement serialization in a single .cpp file.

DrizztDoUrden: Didn't notice we have msgpack as dependency.

DrizztDoUrden: Hmmm, not yet. I c

DrizztDoUrden: @junliume would you mind having nlohmann/json as a dependency? If so, what way would you prefer: copying the header or externally referencing?

pfultz2:
There is a single header include which nlohmann/json provides:

https://github.com/nlohmann/json/blob/develop/single_include/nlohmann/json.hpp

We can copy this single header rather than copy a bunch of individual headers.

include/miopen/miopen.h
```diff
@@ -4667,6 +4667,119 @@ miopenReduceTensor(miopenHandle_t handle,
...
+    miopenProblemDirectionBackwardWeight = 2,
+} miopenProblemDirection_t;
+
+// Enums with no meaningful default value should probably have invalid default for fool-proofing.
+typedef enum
+{
+    miopenTensorNameInvalid  = 0,
+    miopenTensorConvolutionX = 1,
+    miopenTensorConvolutionW = 2,
+    miopenTensorConvolutionY = 3,
```
pfultz2, line 4690: Perhaps X, W, and Y are clear for most AI folks, although I think names like `miopenTensorConvolutionInput`, `miopenTensorConvolutionWeghts`, and `miopenTensorConvolutionOutput` would be clearer. What does everyone else think about the names?

include/miopen/miopen.h
```diff
@@ -4667,6 +4667,119 @@ miopenReduceTensor(miopenHandle_t handle,
...
+ * future, this can support setting different find mode options.
+ */
+MIOPEN_DECLARE_OBJECT(miopenSearchOptions);
+
+miopenStatus_t miopenCreateSearchOptions(miopenSearchOptions_t* options);
+miopenStatus_t miopenDestroySearchOptions(miopenSearchOptions_t options);
+miopenStatus_t miopenSetSearchOption(miopenSearchOptions_t options,
+                                     miopenSearchOptionName_t optionName,
+                                     size_t valueSize,
+                                     const void* value);
```
pfultz2, line 4739: There should be a function for each search option instead of using `void*`.

include/miopen/miopen.h
pfultz2: Yea we should try to avoid `void*` parameters in our code wherever possible. I dont see an issue with having a function instead of an enum for each attribute.

src/convolution.cpp
```diff
@@ -772,6 +774,16 @@ std::ostream& operator<<(std::ostream& stream, const ConvolutionDescriptor& c)
     return stream;
 }
 
+void to_json(nlohmann::json& json, const ConvolutionAttribute::Gfx90aFp16alt& attribute)
+{
+    json = {{"value", attribute.value}};
+}
```
pfultz2, line 780: Just an idea, rather than bringing in a json header for every descriptor we could have a templated friend function:

```cpp
class Gfx90aFp16alt
{
public:
    template<class Json>
    friend void to_json(Json& j, const ConvolutionAttribute::Gfx90aFp16alt& attribute)
    {
        j = {{"value", attribute.value}};
    }
};
```

I think this would still be found by ADL lookup, but I haven't tested this yet. What do you think?

include/miopen/miopen.h
DrizztDoUrden: input/output terminology is unclear for different directions. For example, which is which for backward weights? Technically weights buffer would be the output of the operation.

DrizztDoUrden:
Why would we want it to be single header? With forward declaration header it is much clearer what is used by fact and gives sane autocompletion without making it included in every cpp?

include/miopen/miopen.h
DrizztDoUrden: I would like to talk about that at today's host/api meeting.

src/convolution.cpp
DrizztDoUrden: I have brought only forward declaration in headers, so it should be fine and have nice autocompletion in cpps.

DrizztDoUrden:
@pfultz2 how would you solve the problem of automatic formatting by clang-tidy of the nlohmann/json included in the library? Or should I just format it? Technically that would not violate the MIT license AFAIK.

include/miopen/miopen.h
pfultz2: Ah yea makes sense.

include/miopen/miopen.h
pfultz2: What are some of your concerns with adding functions?

src/problem.cpp
```diff
@@ -0,0 +1,333 @@
...
+};
+
+} // namespace detail
+
+std::vector<Solution> Problem::FindSolutions(Handle& handle,
+                                             const SearchOptions& options,
+                                             std::size_t max_solutions) const
+{
+    const auto visitor = detail::ProblemFindVariantVisitor{this, &handle, &options, max_solutions};
+    auto ret           = boost::apply_visitor(visitor, operator_descriptor);
```
pfultz2, line 119: No need to use boost, `std::visit` should work.

src/problem.cpp
```diff
@@ -0,0 +1,333 @@
...
+    }
+};
+
+} // namespace detail
+
+std::vector<Solution> Problem::FindSolutions(Handle& handle,
+                                             const SearchOptions& options,
+                                             std::size_t max_solutions) const
+{
+    const auto visitor = detail::ProblemFindVariantVisitor{this, &handle, &options, max_solutions};
```
pfultz2, line 118: I think using an overloaded lambdas would be simpler. With boost, you can use [`boost::hof::match`](https://www.boost.org/doc/libs/1_72_0/libs/hof/doc/html/include/boost/hof/match.html) or [`boost::hof::first_of`](https://www.boost.org/doc/libs/1_72_0/libs/hof/doc/html/include/boost/hof/first_of.html). In C++17 without boost, you can add this:

```cpp
template <class... Fs>
struct overloaded : Fs...
{
    using Fs::operator()...;
    overloaded(Fs... fs) : Fs(fs)... {}
};

template <class... Fs>
overloaded<Fs...> overload(Fs... fs)
{
    return {fs...};
}
```

src/convolution.cpp
pfultz2: Sounds good.

pfultz2:
> Why would we want it to be single header? With forward declaration header it is much clearer what is used by fact and gives sane autocompletion without making it included in headers.

Well I thought it would be easier to manage as one file, but if you need the forward declarations then multiple files is probably better.

pfultz2:
> how would you solve the problem of automatic formatting by clang-tidy of the nlohmann/json included in the library? Or should I just format it? 

We should probably just format it. Thats what we do with the md5.cpp file from openssl.

pfultz2:
Actually, having it in the repo I think we will need to install the headers if we include the json headers(such as for forward declarations) in our headers as fin uses are our C++ headers. @JehandadKhan How does fin use our C++ headers? Is it using whats in the repo directly? Or is this using installed C++ headers?

Also, it looks like fin uses nlohmann/json:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/rbuild.ini#L26

This could cause ODR issues with different versions. @JehandadKhan  Is fin always built in MIOpen source tree? If so, fin needs to be updated to use MIOpen's copy of nlohmann/json.

pfultz2:
Actually there could be a potential ODR issue when statically linking with migraphx which uses nlohmann/json as well.

include/miopen/miopen.h
DrizztDoUrden: I don't mind both approaches in general, but the idea was not mine :)

src/problem.cpp
DrizztDoUrden: `std::variant` is C++17, I thought we do not plan to switch there

DrizztDoUrden:
I guess the best solution would be to reference it externally then?

test/find_2_conv.cpp
```diff
@@ -0,0 +1,253 @@
...
+        std::size_t found = 0;
+
+        solutions.resize(100);
+
+        const auto search_values = std::vector<int>({0, 1});
+        const auto workspace_limit_values =
+            std::vector<std::size_t>({std::numeric_limits<std::size_t>::max(), 0});
+
+        for(const auto search : search_values)
+            for(const auto workspace_limit : workspace_limit_values)
```
junliume, line 139: @DrizztDoUrden Tidy complaints:
```
 Value stored to 'workspace_limit' during its initialization is never read [clang-analyzer-deadcode.DeadStores,-warnings-as-errors]

File: test/find_2_conv.cpp [hunk 22]
```diff
@@ -0,0 +1,253 @@
...
+        EXPECT_OP(found, >=, 0);
+        solutions.resize(found);
+        return solutions;
+    }
+
+    void TestSolutionAttributes(const std::vector<miopenSolution_t>& solutions)
+    {
+        for(const auto& solution : solutions)
+        {
+            const auto checked_get_attr = [&](auto name, auto& value) {
```
junliume, line 169: @DrizztDoUrden tidy complaints:
```
unused variable 'checked_get_attr' [clang-diagnostic-unused-variable,-warnings-as-errors]

**Comment by junliume** [comment]
@DrizztDoUrden test_find_2_conv is failing pretty consistently for the latest update, could you take a look?

**Comment by junliume** [comment]
@DrizztDoUrden Now the issues are mostly on `test_conv3d_find2 ` :)
for example:
```
bin/test_conv3d_find2 --float --cmode conv --pmode default --group-count 1 --batch_size 1 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 0 0 0 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW

DrizztDoUrden:
Can't reproduce on the machine I have access to.

junliume:
> Can't reproduce on the machine I have access to.

@DrizztDoUrden it is still failing pretty consistently:
```

[2022-05-14T07:24:05.541Z]  44/103 Test  #17: test_conv3d_find2 .....................................***Failed  Error regular expression found in output. Regex=[FAILED]264.02 sec

[2022-05-14T07:24:05.541Z] /var/jenkins/workspace/MLLibs_MIOpen_ddu_find2.0/build/bin/test_conv3d_find2 --float --cmode conv --pmode default --group-count 1 --batch_size 1 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 0 0 0 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW 

[2022-05-14T07:24:05.541Z] FAILED: inf

[2022-05-14T07:24:05.541Z] Iteration: 80

[2022-05-14T07:24:05.541Z] Backward convolution: GemmBwd1x1_stride1

[2022-05-14T07:24:05.541Z] Input tensor: 1, 32, 4, 9, 9

[2022-05-14T07:24:05.541Z] Weights tensor: 32, 32, 1, 1, 1

[2022-05-14T07:24:05.541Z] Output tensor: 1, 32, 4, 9, 9

[2022-05-14T07:24:05.541Z] Filter: conv3d, miopenConvolution, miopenPaddingDefault, {0, 0, 0}, {1, 1, 1}, {1, 1, 1}, 

[2022-05-14T07:24:05.541Z] Max diff: 2.83701e+24

[2022-05-14T07:24:05.541Z] Mismatch at 0: 188 != 941471

**Comment by junliume** [comment]
> > Can't reproduce on the machine I have access to.
> 
> @DrizztDoUrden it is still failing pretty consistently:
> 
> ```
> 
> [2022-05-14T07:24:05.541Z]  44/103 Test  #17: test_conv3d_find2 .....................................***Failed  Error regular expression found in output. Regex=[FAILED]264.02 sec
> 
> [2022-05-14T07:24:05.541Z] /var/jenkins/workspace/MLLibs_MIOpen_ddu_find2.0/build/bin/test_conv3d_find2 --float --cmode conv --pmode default --group-count 1 --batch_size 1 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 0 0 0 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW 
> 
> [2022-05-14T07:24:05.541Z] FAILED: inf
> 
> [2022-05-14T07:24:05.541Z] Iteration: 80
> 
> [2022-05-14T07:24:05.541Z] Backward convolution: GemmBwd1x1_stride1
> 
> [2022-05-14T07:24:05.541Z] Input tensor: 1, 32, 4, 9, 9
> 
> [2022-05-14T07:24:05.541Z] Weights tensor: 32, 32, 1, 1, 1
> 
> [2022-05-14T07:24:05.541Z] Output tensor: 1, 32, 4, 9, 9
> 
> [2022-05-14T07:24:05.541Z] Filter: conv3d, miopenConvolution, miopenPaddingDefault, {0, 0, 0}, {1, 1, 1}, {1, 1, 1}, 
> 
> [2022-05-14T07:24:05.541Z] Max diff: 2.83701e+24
> 
> [2022-05-14T07:24:05.541Z] Mismatch at 0: 188 != 941471
> ```

BTW~ the platform is `Fp32 OpenCL`so you need to test on OpenCL backend @DrizztDoUrden

File: include/miopen/miopen.h [hunk 23]
```diff
@@ -4667,6 +4667,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+
+miopenStatus_t miopenDestroySearchOptions(miopenSearchOptions_t options);
+
+miopenStatus_t miopenSetSearchOptionTuning(miopenSearchOptions_t options, int value);
+
+miopenStatus_t miopenSetSearchOptionResultsOrder(miopenSearchOptions_t options,
+                                                 miopenSearchResultsOrder_t value);
+
+miopenStatus_t miopenSetSearchOptionWorkspaceLimit(miopenSearchOptions_t options, size_t value);
+
```
JehandadKhan, line 4729: Why not let the client filter this out, seems a little verbose ?

include/miopen/miopen.h
```diff
@@ -4667,6 +4667,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+ * used. This can include things like exhaustiveSearch, and ordering. In the
+ * future, this can support setting different find mode options.
+ */
+MIOPEN_DECLARE_OBJECT(miopenSearchOptions);
+
+miopenStatus_t miopenCreateSearchOptions(miopenSearchOptions_t* options);
+
+miopenStatus_t miopenDestroySearchOptions(miopenSearchOptions_t options);
+
+miopenStatus_t miopenSetSearchOptionTuning(miopenSearchOptions_t options, int value);
```
JehandadKhan, line 4723: Can we add a few lines about this call, as to what it does ?

JehandadKhan: Also, why `int`  ?

include/miopen/miopen.h
```diff
@@ -4667,6 +4667,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+
+miopenStatus_t miopenSetSearchOptionWorkspaceLimit(miopenSearchOptions_t options, size_t value);
+
+MIOPEN_DECLARE_OBJECT(miopenSolution);
+
+/**
+ * Finds solutions to a problem by running different applicable solutions. Memory is automatically
+ * allocated.
+ */
+miopenStatus_t miopenFindSolutions(miopenHandle_t handle,
```
JehandadKhan, line 4736: All the new API calls implicitly assume the operation to be convolution, however, it is not reflected in its name. I would prefer not to have to replicate all these functions for each op and have an operator argument. Even if that is not acceptable, we should at least have `Convolution` in the API name

JehandadKhan (changes requested):

include/miopen/miopen.h
DrizztDoUrden: To save time by not building, tuning and evaluating solvers that are not applicable.

include/miopen/miopen.h
DrizztDoUrden: There is no bool in C. I will add documentation after fixing bugs. This specifies whether or not to run tuning, aka exhaustive search.

include/miopen/miopen.h
DrizztDoUrden: They don't. What would be executed depends on the problem or solution.

include/miopen/miopen.h
DrizztDoUrden: I wouldn't mind changing this functions naming convention, but have no idea what to.

JehandadKhan:
@pfultz2  Can you please check if your reviews have been addressed ?

include/miopen/miopen.h
```diff
@@ -4701,6 +4701,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+
+miopenStatus_t miopenCreateConvProblem(miopenProblem_t* problem,
+                                       miopenConvolutionDescriptor_t operatorDesc,
+                                       miopenProblemDirection_t direction);
+
+miopenStatus_t miopenDestroyProblem(miopenProblem_t problem);
+
+// Universal problem should have named tensor fields rather than ordered.
+// Previously, different directions had different input orders, which may lead to confusion.
+// Implicitly ordered fields would also lead to a whole lot of magical ids all over the internals.
```
pfultz2, line 4741: This seems like a comment for internal developers. We should just have a comment for describing this for external users.

include/miopen/miopen.h
```diff
@@ -4701,6 +4701,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+ * descriptors are passed in so if a user would like to dynamically change
+ * the sizes, they can do so. If dynamic changes are not supported(or are
+ * invalid), we should return an error to the user.
+ */
+miopenStatus_t miopenRunSolution(miopenHandle_t handle,
+                                 miopenSolution_t solution,
+                                 size_t nInputs,
+                                 const miopenTensorName_t* names,
+                                 const miopenTensorDescriptor_t* descriptors,
+                                 void* const* buffers,
```
pfultz2, line 4788: I dont know if its too late to change this, but I think it would be better to have a struct to store these together:

```cpp
struct miopenTensorArgument {
    miopenTensorName_t name;
    miopenTensorDescriptor_t descriptor;
    void* buffer;
};
```

`void **` just accepts too many things that are not even close to correct.

src/solution.cpp
```diff
@@ -0,0 +1,244 @@
...
+    {
+        const auto header = json.at("header").get<Solution::SerializationMetadata>();
+        constexpr const auto check_header = Solution::SerializationMetadata::Current();
+
+        if(header.validation_number != check_header.validation_number)
+            MIOPEN_THROW(miopenStatusInvalidValue,
+                         "Invalid buffer has been passed to the solution deserialization.");
+        if(header.version != check_header.version)
+            MIOPEN_THROW(
+                miopenStatusInvalidValue,
```
pfultz2, line 235: We should probably have a different error code for version mismatch.

include/miopen/miopen.h
atamazov: @pfultz2 Let me argue a bit. Yes, description for external users is necessary, but comments that explain motivation ("rationale") are often very useful for them.

atamazov:
@DrizztDoUrden 
> miopenStatus_t miopenGetSolutionsNumber(miopenHandle_t handle,

miopenGetNumberOfSolutions looks less ambiguous.

include/miopen/miopen.h
atamazov, line 4788: The signature seems redundant. The user wants to run a Solution for some Problem. The Problem "knows" all TensorDescriptors as well as a number of tensors/buffers. So if the instance of a Problem still exists, the user can pass Problem and a set of {name, buffer_ptr} pairs.

include/miopen/miopen.h
```diff
@@ -4701,6 +4701,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+    miopenProblemDirectionBackwardWeight = 2,
+} miopenProblemDirection_t;
+
+// Enums with no meaningful default value should probably have invalid default for fool-proofing.
+typedef enum
+{
+    miopenTensorNameInvalid  = 0,
+    miopenTensorConvolutionX = 1,
+    miopenTensorConvolutionW = 2,
+    miopenTensorConvolutionY = 3,
```
atamazov, line 4724: What about Workspace buffer? How can the user pass it to miopenRunSolution?

include/miopen/miopen.h
```diff
@@ -4701,6 +4701,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+ * For now, this is only used for convolution, but could be used for other
+ * operators in the future(such as GEMM, Pooling, etc)
+ */
+MIOPEN_DECLARE_OBJECT(miopenProblem);
+
+typedef enum
+{
+    miopenProblemDirectionForward        = 0,
+    miopenProblemDirectionBackward       = 1,
+    miopenProblemDirectionBackwardWeight = 2,
```
atamazov, line 4715:
```suggestion
    miopenProblemDirectionBackwardWeights = 2,
```

include/miopen/miopen.h
```diff
@@ -4701,6 +4701,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+
+// Universal problem should have named tensor fields rather than ordered.
+// Previously, different directions had different input orders, which may lead to confusion.
+// Implicitly ordered fields would also lead to a whole lot of magical ids all over the internals.
+miopenStatus_t miopenSetProblemTensorDescriptor(miopenProblem_t problem,
+                                                miopenTensorName_t name,
+                                                const miopenTensorDescriptor_t descriptor);
+
+/**
+ * The miopenSearchOptions will allow the user to configure how search will be
```
atamazov, line 4747: Let's stop the confusion between "Search" and "Find" and use "Tuning" instead of "Search" everywhere in the new API.

include/miopen/miopen.h
```diff
@@ -4701,6 +4701,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+                                       miopenProblemDirection_t direction);
+
+miopenStatus_t miopenDestroyProblem(miopenProblem_t problem);
+
+// Universal problem should have named tensor fields rather than ordered.
+// Previously, different directions had different input orders, which may lead to confusion.
+// Implicitly ordered fields would also lead to a whole lot of magical ids all over the internals.
+miopenStatus_t miopenSetProblemTensorDescriptor(miopenProblem_t problem,
+                                                miopenTensorName_t name,
+                                                const miopenTensorDescriptor_t descriptor);
```
atamazov, line 4744: AFAICS this creates a copy of a tensor descriptor and associates that copy with a problem. It is what we really want? Or it is better to bind tensor descriptor to a problem (pass a pointer)? 

/cc @pfultz2

include/miopen/miopen.h
```diff
@@ -4701,6 +4701,106 @@ miopenReduceTensor(miopenHandle_t handle,
...
+} miopenProblemDirection_t;
+
+// Enums with no meaningful default value should probably have invalid default for fool-proofing.
+typedef enum
+{
+    miopenTensorNameInvalid  = 0,
+    miopenTensorConvolutionX = 1,
+    miopenTensorConvolutionW = 2,
+    miopenTensorConvolutionY = 3,
+} miopenTensorName_t;
```
atamazov, line 4725: Name is too common. `miopenTensorAssignment_t`?

atamazov (changes requested):

include/miopen/miopen.h
DrizztDoUrden: Tensor descriptors are optional here. That's not my idea originally, and a lot of solvers won't support mismatching tensor descriptors right now.

include/miopen/miopen.h
DrizztDoUrden: It is an argument of `miopenRunSolution`. It would be problematic to tensor because it doesn't have `miopenTensorDescriptor`, just size.

include/miopen/miopen.h
DrizztDoUrden: These are actually find options, but I see your point and agree with you.

include/miopen/miopen.h
DrizztDoUrden: Yes, we want a copy. Otherwise user would have to manage lifetime of a `TensorDescriptor` with a problem and a lot of bad things would happen if they would free it and user the problem afterwards, which would require additional internal complexity to produce a sane error. TensorDescriptor is not a large structure by the way.

include/miopen/miopen.h
DrizztDoUrden: "Assignment" would be confusing for both users and MIOpen developers.

include/miopen/miopen.h
pfultz2: We added the tensor descriptors to support dynamic tensors in the future, but maybe this doesn't really make sense.

include/miopen/miopen.h
pfultz2: I would prefer search as it encompasses both "Find" and "Tuning"(which is really just a larger search space).

include/miopen/miopen.h
pfultz2: To reduce the overhead of copying we could store the tensor descriptor data in a impl class, and make all public methods `const` so we can store the impl as `std::share_ptr<const impl>`, so copies are just ref count update. Thats what we do for the 'shape` class in migraphx:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/shape.hpp#L232

include/miopen/miopen.h
DrizztDoUrden: Actually, there are solvers which support it. What we need is a way to return that information to the user and to properly check for that before executing.

include/miopen/miopen.h
DrizztDoUrden: Nah, these are specifically find options. Tuning is actually an optional part of the find. And the only tuning related option there is whether to tune at all or not. Which is a parameter of find 1.0 method.

include/miopen/miopen.h
DrizztDoUrden: Yes, I was thinking about that as well. Though that is irrelevant from the API perspective and we can add it when we would think that actually matters. That may be planned as a future improvement, but I don't think that would be significant realistically.

include/miopen/miopen.h
atamazov: Maybe `miopenTensorBinding_t`? After all, here we tell which output or input of the primitive we bind the tensor to. I've always thought that a "name" is something the user can freely assign, but in this specific case this is impossible, and that confuses me...

include/miopen/miopen.h
atamazov: So, shall we name it like `miopenFindOptions`?

include/miopen/miopen.h
pfultz2: I see, I think something like `miopenTensorArgumentId_t`. We can use `ID` which doesn't mean arbitrary name the user gives. I think `miopenTensorBinding_t` is not very clear.

include/miopen/miopen.h
DrizztDoUrden: Yes, sure.

junliume:
@DrizztDoUrden the most recent static check issue:
```
[2022-06-23T08:56:41.599Z] /var/jenkins/workspace/MLLibs_MIOpen_ddu_find2.0/test/find_2_conv.cpp:233:52: error: pointer parameter 'names' can be pointer to const [readability-non-const-parameter,-warnings-as-errors]
[2022-06-23T08:56:41.599Z]                          miopenTensorArgumentId_t* names,
[2022-06-23T08:56:41.599Z]                                                    ^
[2022-06-23T08:56:41.599Z]                          const

**Comment by DrizztDoUrden** [comment]
@junliume @JehandadKhan I have reproduced the opencl vega20 gemm failure on develop and fixed the rest of the bugs.

**Comment by junliume** [comment]
> Will make the docs before tomorrow and it would be ready. The GEMM failure requires additional investigation, but I am more concerned about why it doesn't happen on Jenkins.

Jenkins is using the last version of ROCm in the docker, currently 5.1, and the base OS is using ROCm 5.0.2 which might have played a role in your findings.

**Comment by DrizztDoUrden** [comment]
I mean it did happen on Jenkins in find 2.0 test, just not in find 1.0. The same call to conv3d test on both develop and my branch on my pc did reproduce the failure.

**Comment by DrizztDoUrden** [comment]
Right now I don't believe current failures are caused by the PR. #1649 also failes CI, but it just doesn't changes anything which executes in tests, only debug priniting.

**Comment by junliume** [comment]
> Right now I don't believe current failures are caused by the PR. #1649 also failes CI, but it just doesn't changes anything which executes in tests, only debug priniting.

@DrizztDoUrden the following test `test_conv3d_find2` and the stage `Fp32 OpenCL Install All` has failed the CI pretty consistently:
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/ddu%2Ffind2.0/82/pipeline/1221
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/ddu%2Ffind2.0/84/pipeline/250

Hence they are both legacy ASIC and OpenCL backend, so I am fine with disabling this test on them.

**Comment by junliume** [comment]
@DrizztDoUrden if you have a gfx900/906 for a quick test:
```
CXX=g++ CXXFLAGS='-Werror'  cmake -DMIOPEN_TEST_FLAGS=' --disable-verification-cache ' -DCMAKE_BUILD_TYPE=release -DBUILD_DEV=Off -DCMAKE_INSTALL_PREFIX=../install -DMIOPEN_USE_MLIR=ON -DMIOPEN_GPU_SYNC=Off  -DMIOPEN_TEST_ALL=On   .. 

LLVM_PATH=/opt/rocm/llvm CTEST_PARALLEL_LEVEL=4 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0  dumb-init make -j$(nproc) install check
```
```
[2022-07-30T10:55:46.202Z] /home/jenkins/workspace/MLLibs_MIOpen_ddu_find2.0/build/bin/test_conv3d_find2 --float --cmode conv --pmode valid --group-count 1 --batch_size 8 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW 
[2022-07-30T10:55:46.202Z] FAILED: -nan
[2022-07-30T10:55:46.202Z] Iteration: 722
[2022-07-30T10:55:46.202Z] Backward convolution: GemmBwd1x1_stride1
[2022-07-30T10:55:46.202Z] Input tensor: 8, 32, 4, 9, 9
[2022-07-30T10:55:46.202Z] Weights tensor: 32, 32, 1, 1, 1
[2022-07-30T10:55:46.202Z] Output tensor: 8, 32, 4, 9, 9
[2022-07-30T10:55:46.202Z] Filter: conv3d, miopenConvolution, miopenPaddingValid, {0, 0, 0}, {1, 1, 1}, {1, 1, 1}, 
[2022-07-30T10:55:46.202Z] Max diff: inf
[2022-07-30T10:55:46.202Z] Mismatch at 0: -13 != 1.09577e+29
[2022-07-30T10:55:46.202Z] Non finite number found in gpu at 72576: inf

junliume:
> @DrizztDoUrden if you have a gfx900/906 for a quick test:
> 
> ```
> CXX=g++ CXXFLAGS='-Werror'  cmake -DMIOPEN_TEST_FLAGS=' --disable-verification-cache ' -DCMAKE_BUILD_TYPE=release -DBUILD_DEV=Off -DCMAKE_INSTALL_PREFIX=../install -DMIOPEN_USE_MLIR=ON -DMIOPEN_GPU_SYNC=Off  -DMIOPEN_TEST_ALL=On   .. 
> 
> LLVM_PATH=/opt/rocm/llvm CTEST_PARALLEL_LEVEL=4 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0  dumb-init make -j$(nproc) install check
> ```
> 
> ```
> [2022-07-30T10:55:46.202Z] /home/jenkins/workspace/MLLibs_MIOpen_ddu_find2.0/build/bin/test_conv3d_find2 --float --cmode conv --pmode valid --group-count 1 --batch_size 8 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW 
> [2022-07-30T10:55:46.202Z] FAILED: -nan
> [2022-07-30T10:55:46.202Z] Iteration: 722
> [2022-07-30T10:55:46.202Z] Backward convolution: GemmBwd1x1_stride1
> [2022-07-30T10:55:46.202Z] Input tensor: 8, 32, 4, 9, 9
> [2022-07-30T10:55:46.202Z] Weights tensor: 32, 32, 1, 1, 1
> [2022-07-30T10:55:46.202Z] Output tensor: 8, 32, 4, 9, 9
> [2022-07-30T10:55:46.202Z] Filter: conv3d, miopenConvolution, miopenPaddingValid, {0, 0, 0}, {1, 1, 1}, {1, 1, 1}, 
> [2022-07-30T10:55:46.202Z] Max diff: inf
> [2022-07-30T10:55:46.202Z] Mismatch at 0: -13 != 1.09577e+29
> [2022-07-30T10:55:46.202Z] Non finite number found in gpu at 72576: inf
> ```

It turns out that I cannot reproduce this issue on a gfx906 either. I guess we'd disable it first as a workaround.

DrizztDoUrden:
This is what I get on ***develop***:
```
$ bin/test_conv3d --float --cmode conv --pmode valid --group-count 1 --batch_size 8 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW
bin/test_conv3d --float --cmode conv --pmode valid --group-count 1 --batch_size 8 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW
FAILED: 0.0573837
Iteration: 0
Backward convolution: GemmBwd1x1_stride1
Input tensor: 8, 32, 4, 9, 9
Weights tensor: 32, 32, 1, 1, 1
Output tensor: 8, 32, 4, 9, 9
Filter: conv3d, miopenConvolution, miopenPaddingValid, {0, 0, 0}, {1, 1, 1}, {1, 1, 1},
Max diff: 2.63712e+06
Mismatch at 320: -695 != 0
```

shurale-nkn:
> This is what I get on _**develop**_:
> 
> ```
> $ bin/test_conv3d --float --cmode conv --pmode valid --group-count 1 --batch_size 8 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW
> bin/test_conv3d --float --cmode conv --pmode valid --group-count 1 --batch_size 8 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW
> FAILED: 0.0573837
> Iteration: 0
> Backward convolution: GemmBwd1x1_stride1
> Input tensor: 8, 32, 4, 9, 9
> Weights tensor: 32, 32, 1, 1, 1
> Output tensor: 8, 32, 4, 9, 9
> Filter: conv3d, miopenConvolution, miopenPaddingValid, {0, 0, 0}, {1, 1, 1}, {1, 1, 1},
> Max diff: 2.63712e+06
> Mismatch at 320: -695 != 0
> ```
> 
> It only reproduces in docker though, it doesn't happen for me outside. I don't know what causes the issue.

@junliume 
reproduced on navi21.  I think this is an architecture-independent problem.

MIOpen(OpenCL): Warning [PrintOpenCLDeprecateMsg] Please note that the OpenCL backend to MIOpen is being deprecated, 
MIOpen(OpenCL): Warning [PrintOpenCLDeprecateMsg] please port your application to the better supported and functional HIP backend. 
MIOpen(OpenCL): Warning [PrintOpenCLDeprecateMsg] If you have any questions, please reach out to the MIOpen developers at 
MIOpen(OpenCL): Warning [PrintOpenCLDeprecateMsg] https://github.com/ROCmSoftwarePlatform/MIOpen
bin/test_conv3d --float --cmode conv --pmode valid --group-count 1 --batch_size 8 --input_channels 32 --output_channels 32 --spatial_dim_elements 4 9 9 --filter_dims 1 1 1 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW 
FAILED: 0.0573837
Iteration: 0
Backward convolution: GemmBwd1x1_stride1
Input tensor: 8, 32, 4, 9, 9
Weights tensor: 32, 32, 1, 1, 1
Output tensor: 8, 32, 4, 9, 9
Filter: conv3d, miopenConvolution, miopenPaddingValid, {0, 0, 0}, {1, 1, 1}, {1, 1, 1}, 
Max diff: 2.63712e+06
Mismatch at 320: -695 != 0

DrizztDoUrden:
With OpenCL stages made optional this should pass even without #1683

junliume:
@atamazov  @pfultz2 @JehandadKhan 
CI has passed, please provide your final round of review for this PR. Thanks!

include/miopen/miopen.h
```diff
@@ -4704,6 +4706,231 @@ miopenReduceTensor(miopenHandle_t handle,
...
+ */
+miopenStatus_t miopenDestroyFindOptions(miopenFindOptions_t options);
+
+/*! @brief Sets the tuning find option. Default value is zero.
+ *
+ * @param options    Options object to upfate
+ * @param value      Value of zero means no tuning, value of one means tuning enabled
+ * @return           miopenStatus_t
+ */
+miopenStatus_t miopenSetFindOptionTuning(miopenFindOptions_t options, int value);
```
pfultz2, line 4805: It would be better `miopenSetFindOptionEnableTuning(miopenFindOptions_t options, int enabled)`.

pfultz2:
Is there a find option to time all solutions? I believe by default MIOpen will skip solutions if the config/problem is not found in the find_db(aka `DYNAMIC_HYBRID`).

DrizztDoUrden:
Right now 2.0 find call simply invokes 1.0 method. Find mode is read from the environment inside of that. As a future work we should probably allow overriding find-related env variables from find options. But I don't think it should be a part of this PR as it requires a lot of next steps to be implemented before.

junliume:
@DrizztDoUrden please merge the latest `develop` and we will push forward with this PR when the CI has passed. Thanks!
CC: @atamazov  @JehandadKhan

include/miopen/miopen.h
```diff
@@ -4704,6 +4706,231 @@ miopenReduceTensor(miopenHandle_t handle,
...
+MIOPEN_EXPORT miopenStatus_t miopenCreateConvProblem(miopenProblem_t* problem,
+                                                     miopenConvolutionDescriptor_t operatorDesc,
+                                                     miopenProblemDirection_t direction);
+
+/*! @brief Destroys a problem object.
+ *
+ * @param problem Problem to destroy
+ * @return        miopenStatus_t
+ */
+miopenStatus_t miopenDestroyProblem(miopenProblem_t problem);
```
umangyadav, line 4768: Does it have corresponding `miopenCreateProblem` ?

umangyadav: `miopenSolution_t` has API for destroy, does it have one for `create` ?

DrizztDoUrden: It is allocated by miopen `miopenFindSolutions` and `miopenLoadSolution`.

DrizztDoUrden: Basically, it can't be created and empty at the same time.

src/invoker_cache.cpp
```diff
@@ -72,6 +72,32 @@ boost::optional<const Invoker&> InvokerCache::GetFound1_0(const std::string& net
     return invoker->second;
 }
 
+boost::optional<const std::string&> InvokerCache::GetFound1_0Id(const std::string& network_config,
```
atamazov, line 75: This name is hardly understandable. Can we use something like GetFound1_0SolverName?

atamazov (approved):
This can be merged as is, but I still recommend considering comments.

src/include/miopen/solution.hpp
```diff
@@ -0,0 +1,109 @@
...
+        static constexpr SerializationMetadata Current() { return {0x123456789ABCDEF0, 1}; }
+
+        friend void to_json(nlohmann::json& json, const SerializationMetadata& metadata);
+        friend void from_json(const nlohmann::json& json, SerializationMetadata& metadata);
+    };
+
+    struct RunInput
+    {
+        boost::optional<TensorDescriptor> descriptor;
+        Data_t buffer;
```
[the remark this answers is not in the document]
DrizztDoUrden: Not in that sense. This is a generic operand. I may rename this if you want.

src/invoker_cache.cpp
DrizztDoUrden: Would you be okay with GetFound1_0SolverId?

src/include/miopen/solution.hpp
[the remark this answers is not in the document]
atamazov: @DrizztDoUrden Let's discuss by voice

src/invoker_cache.cpp
atamazov: Of course

atamazov: But please note that I think that distinguishing the difference between _numeric_ and _string_ solver identifiers is important because strings are much more expensive.

I still recommend using "ID" for the numeric identifiers, and "Name" for the strings.

DrizztDoUrden: I don't think there is much overhead, but I'd rather have them both replaced a bit later everywhere we can by `solver::Id`. That is as fast as number (as it is one), type-safe and trivially and cheaply enough (tiny map fetch) converted to string.

src/api/find2_0_commons.cpp
```diff
@@ -0,0 +1,259 @@
...
+}
+
+miopenStatus_t miopenRunSolution(miopenHandle_t handle,
+                                 miopenSolution_t solution,
+                                 size_t nInputs,
+                                 const miopenTensorArgument_t* tensors,
+                                 void* workspace,
+                                 size_t workspaceSize)
+{
+    MIOPEN_LOG_FUNCTION(handle, solution, nInputs, workspace, workspaceSize);
```
umangyadav, line 148: This one is not printing `tensors` argument.

include/miopen/miopen.h
```diff
@@ -4704,6 +4706,231 @@ miopenReduceTensor(miopenHandle_t handle,
...
+ * @param solution      Solution to execute
+ * @param nInputs       Amount to inputs for the solution
+ * @param tensors       Tensor arguments described by miopenTensorArgument_t
+ * @param workspace     Pointer to device buffer used as workspace. May be null when not required.
+ * Should not be less than expected
+ * @param workspaceSize Size of the workspace buffer
+ * @return              miopenStatus_t
+ */
+miopenStatus_t miopenRunSolution(miopenHandle_t handle,
+                                 miopenSolution_t solution,
```
umangyadav, line 4877: `solution`  here, can it be made const ?

DrizztDoUrden: Why? It's not a reference and unless it won't have any meaningful effect other than a warning. It will become `miopenSolution* const` rather than `const miopenColution*`.

umangyadav: ok. 
why would it turn to `miopenSolution* const` instead of `const miopenSolution*` ?

DrizztDoUrden: Because applying `const` to a type `miopenSolution_t`, which is `miopenSolution*`, would make it `miopenSolution* const`.

DrizztDoUrden: https://i.imgur.com/Rut5gqz.png

DrizztDoUrden: For further details you may check literature: https://www.cprogramming.com/tutorial/const_correctness.html

atamazov (approved):
LGTM!

include/miopen/miopen.h
DrizztDoUrden: `typedef` is not a macro. It defines a type. Application of `const` to a pointer makes the pointer const, not the value.

shurale-nkn (changes requested):
please make third-party content as installable external dependencies

pfultz2:
> Right now 2.0 find call simply invokes 1.0 method. Find mode is read from the environment inside of that.

So for now we can continue using the hidden api to set the find mode.

> As a future work we should probably allow overriding find-related env variables from find options. But I don't think it should be a part of this PR as it requires some of the next steps to be implemented before.

Sure it makes sense, but we shouldn't expose the API as modes, as that is very confusing for users. We should probably expose it in terms of feature, ie using finddb cache, disable checking kernels that need to be compiled, etc. We can think of better names for this in the API in the future, we dont need to fix it in this PR.

junliume:
@DrizztDoUrden could you help to make the changes? We wish to push this PR through asap. Thanks!

atamazov:
@pfultz2 @DrizztDoUrden Mostly +1 to to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1503#issuecomment-1238441712. This feature needs to be thoroughly thought through & analyzed prior implementation; it worth separate ticket.

> Sure it makes sense, but we shouldn't expose the API as modes, as that is very confusing for users.

I wouldn't claim that Find modes are _very confusing_, because these seem well documented to me. I would say it is _complicated_. What is important is that the notions of find-db (find cache), dynamic kernels etc. will present in the new feature as well, thus keeping the difficulty (to understand) at the same level at least.

The functional difference (as I see it) is that while Find modes are process-global, the proposed feature allows setting per API call, which may make implementation more complex.

include/miopen/miopen.h
atamazov: @DrizztDoUrden @pfultz2 [Recommendation] Use `miopenSetFindOptionTuningMode` and introduce enum for tuning mode (0 - disabled, 1- enabled).

atamazov:
@pfultz2 
> Is there a find option to time all solutions? I believe by default MIOpen will skip solutions if the config/problem is not found in the find_db(aka `DYNAMIC_HYBRID`).

In most Find modes (including DYNAMIC_HYBRID which is the default), MIOpen will skip benchmarking stage if the config/problem **_is found_** in the find-db (it simply takes cached timings from db). This happens in all Find modes except NORMAL.

_How_ MIOpen performs benchmarking is a different question. In NORMAL and HYBRID mode, it benchmarks **_all_** applicable Solutions. In DYNAMIC_HYBRID mode, it benchmarks only dynamic Solvers. In FAST mode, benchmarking is never performed. More info at https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/find_and_immediate.md#find-modes.

Please use hidden API to set NORMAL or HYBRID find mode if you want to benchmark all solutions (NORMAL to skip find-db caching, HYBRID to reuse results of previous benchmarking).

atamazov:
> @DrizztDoUrden could you help to make the changes? We wish to push this PR through asap. Thanks!

@junliume I do not see anything that needs to be changed in this PR (except https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1503#discussion_r949246229 which does not seem very important to me).

junliume:
> > @DrizztDoUrden could you help to make the changes? We wish to push this PR through asap. Thanks!
> 
> @junliume I do not see anything that needs to be changed in this PR (except [#1503 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1503#discussion_r949246229) which does not seem very important to me).

We discussed a bit earlier about installing json third party library instead of copying over their source files :)

atamazov:
@junliume 
> installing json third party library

I recommend doing that later in separate PR.

atamazov:
@pfultz2 Just to remind that we have a draft of Find Mode API that allows mode setting per convolution descriptor. It resides in https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/miopen_internal.h and can be easily made public.

pfultz2:
> I wouldn't claim that Find Modes feature is very confusing, because these seem well documented. I would say it is complicated, because the user has to understand the notions of benchmarking, find-db (find cache), dynamic kernels etc.

The problem is that Find mode is mixing different features together. Find-db caching is completely orthogonal to which kernels we will benchmark. Also, if we add some predictive capabilities a single mode could get more complicated.
It seems we need a want a setting to control caching and then a another setting to control/prune which kernels we will benchmark.

> Just to remind that we have a draft of Find Mode API that allows mode setting per convolution descriptor. It resides in https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/miopen_internal.h and can be easily made public.

That was a hack so we could set the find mode in migraphx without an env variable. Its hidden so we can remove it when we have find 2.0 support it. It should not be made public, it doesn't make sense for the convolution descriptor to have the find mode. If the user sets the find mode to `NORMAL` and then call immediate mode, then it seems the user could expect it to benchmark kernels(which wont happen). The find mode should only be set when the user is doing benchmarking through the `miopenFindSolutions`. That is why it should be set in the `miopenFindOptions` and not in a descriptor(and supporting both is even more confusing).

atamazov:
@pfultz2 
> If the user sets the find mode to `NORMAL` and then call immediate mode, then it seems the user could expect it to benchmark kernels(which wont happen).

This kind of expectation is a user's error. We clearly separate **_Find_** API and **_Immediate mode_** API in the documentation, and also we state there that **_Find modes_** only affect **_Find_** API. Maybe we need to be more clear in the docs, however.

> The find mode should only be set when the user is doing benchmarking through the `miopenFindSolutions`. That is why it should be set in the `miopenFindOptions` and not in a descriptor(and supporting both is even more confusing).

Yes, `miopenFindOptions` provides an explicit connection between Find calls and Find modes/options, while relying on documentation is a kind of implicit dependence, which is always worse. And supporting both is not good.

atamazov:
@junliume @pfultz2 Shall we open separate ticket for FindOptions?

junliume:
@DrizztDoUrden  @JehandadKhan
Updating dependency will require updating the base CI docker image. Can we settle on the version requirement, so that we can update the docker and let it pass through CI? Thanks!

atamazov:
@DrizztDoUrden @junliume Performance testing is in progress, ETA is several hours from now.

CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ if(MIOPEN_ENABLE_SQLITE)
     pkg_check_modules(SQLITE3 REQUIRED sqlite3)
 endif()
 find_package(BZip2)
-find_package(nlohmann_json 3.11.2 REQUIRED)
+find_package(nlohmann_json 3.9.1 REQUIRED)
```
atamazov, line 73: I am Ok with this, but just in case -- why this change is required?

DrizztDoUrden: So we won't have to update the image. It already contains 3.9.1

atamazov:
> @DrizztDoUrden @junliume Performance testing is in progress, ETA is several hours from now.

Tests passed, good to merge.

atamazov:
@shurale-nkn 
> please make third-party content as installable external dependencies

Please re-review or dismiss your blocking review.

atamazov:
@JehandadKhan https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1503#pullrequestreview-987885424

Please dismiss your blocking review or re-review, thanks.

DrizztDoUrden:
The image would still need to be updated, it should contain multiple-headers version of nlohmann::json rather than the usual one.

junliume:
> The image would still need to be updated, it should contain multiple-headers version of nlohmann::json rather than the usual one.

@JehandadKhan could you help to update the CI base docker? Thanks!
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/ddu%2Ffind2.0/100/pipeline

Dockerfile
```diff
@@ -59,6 +59,7 @@ DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated \
     rocm-opencl \
     rocm-opencl-dev \
     rocblas \
+    rpm \
```
shurale-nkn, line 62: why do we need it?

shurale-nkn (dismissed):

junliume (approved):
CI passed with updated CI docker (ROCm 5.2.3)

CMakeLists.txt
pfultz2: You need to link with `nlohmann_json::nlohmann_json` so the include paths are correctly added.

shurale-nkn (approved):
LGTM

junliume (approved):

src/problem.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+    for(const auto& pair : tensor_descriptors)
+    {
+        const auto& descriptor  = pair.second;
+        const auto element_size = get_data_size(descriptor.GetType());
+        auto buffer             = handle.Create(descriptor.GetElementSpace() * element_size);
+
+        visit_float(descriptor.GetType(), [&](auto as_float) {
+            const auto zero = as_float(0.f);
+            SetTensor(handle, descriptor, buffer.get(), &zero);
```
atamazov, line 107: ⚓

include/miopen/miopen.h
```diff
@@ -4704,6 +4706,231 @@ miopenReduceTensor(miopenHandle_t handle,
...
+ *
+ * @param handle       Handle to execute the kernels
+ * @param problem      Problem to solve
+ * @param options      Find options. When null default values would be used
+ * @param solutions    Pointer to the first result. Must not be null
+ * @param numSolutions Pointer to the amount of results. Ignored if null
+ * @param maxSolutions Limits the amount of results
+ * @return             miopenStatus_t
+ */
+miopenStatus_t miopenFindSolutions(miopenHandle_t handle,
```
atamazov, line 4840: @DrizztDoUrden @umangyadav Do we have a ticket intended for discussion of Find 2.0 API? Current API enforces the library to allocate buffers for tensors. I am afraid this approach is incompatible with frameworks that do manage GPU memory themselves, and may lead to memory starvation issues with really big tensors/networks.

DrizztDoUrden: @atamazov We could allow passing pre-allocated tensors via find options. That would only require one new universal API method and would keep the API flexible.

umangyadav: memory starvation looks like a real issue. I wasn't planning on opening a dedicated ticket for find 2.0. Do we have any existing ticket for find-2.0 that we can re-use?

DrizztDoUrden: I think separate issues should be confined to their own tickets. We can either tag them with some label or create an umbrella ticket with a list of other issues.

src/problem.cpp
```diff
@@ -0,0 +1,375 @@
...
+        const auto algo = ConvolutionAlgoToDirectionalString(
+            static_cast<miopenConvAlgorithm_t>(find1_solutions[i].fwd_algo), conv_dir);
+
+        auto solution = Solution{};
+        solution.SetTime(find1_solutions[i].time);
+        solution.SetWorkspaceSize(find1_solutions[i].memory);
+        solution.SetSolver(handle.GetFound1_0SolverId(netcfg, AlgorithmName{algo}).value());
+        solution.SetProblem(*this);
+
+        MIOPEN_LOG_I("Found solvution: " << solution.GetSolver().ToString() << " , "
```
atamazov, line 310: Typo

junliume: @DrizztDoUrden  please fix in later PRs. Thanks!

PR #1504, opened 2022-04-05
[CI] Update to ROCm 5.1 in dockerfile and remove miopentensile stage

junliume:
**[Purpose]**

- Update MIOpen CI dockerfile ROCm version to 5.1
- Rational: We are looking for a stable set of CI configurations. 5.1 is more stable during some of the tests.

**[TODOs]**

- Disabled a few OCL unit tests, which passed local tests if baseOS and docker ROCm are both 5.1 (unclear if hardware or baseOS issues)
- gfx90a xnack+ is disabled temporarily. Need to find the root cause of the issue.


src/include/miopen/utility/transposing_solver.hpp
```diff
@@ -591,7 +591,7 @@ struct TransposingSolver : Base
         auto transposed_problem = problem;
         for(const auto& transpose : Derived::GetTransposes())
             transpose.Transpose(problem, transposed_problem);
-        return std::move(transposed_problem);
+        return transposed_problem;
```
JehandadKhan, line 594: This might create a copy, what warning are we fixing here ?

src/ocl/handleocl.cpp
```diff
@@ -417,7 +417,7 @@ Program Handle::LoadProgram(const std::string& program_name,
         miopen::SaveBinary(
             path.string(), this->GetTargetProperties(), program_name, params, is_kernel_str);
 #endif
-        return std::move(p);
+        return p;
```
JehandadKhan, line 420: This might create a copy, what warning are we fixing here ?

test/ford.hpp
```diff
@@ -53,7 +53,7 @@ std::future<typename std::result_of<Function()>::type> detach_async(Function&& f
     std::packaged_task<result_type()> task(std::forward<Function>(f));
     auto fut = task.get_future();
     std::thread(std::move(task)).detach();
-    return std::move(fut);
+    return fut;
```
JehandadKhan, line 56: same as above two

JehandadKhan (approved):
A few minor comments, looks fine otherwise.

src/ocl/handleocl.cpp
junliume: mmm ... I'd think the compiler should automatically optimize out the copy in these cases (NRVO or not), nothing to gain by explicitly inhibit it and force the compiler.

Jenkinsfile
```diff
@@ -866,16 +870,17 @@ pipeline {
...
-                        beforeAgent true
-                        expression { params.TARGET_GFX90A && params.DATATYPE_FP32 }
-                    }
-                    agent{ label rocmnode("gfx90a") }
-                    steps{
-                        buildHipClangJobAndReboot(setup_flags: Full_test, gpu_arch: "gfx90a:xnack+", enforce_xnack_on: true)
-                    }
-                }
+                // #TODO Code Quality WORKAROUND ROCm 5.1 update
+                // stage('Fp32 Hip All gfx90a Xnack+') {
```
atamazov, line 874: Ditto.

cmake/FindOpenCL.cmake
```diff
@@ -75,7 +75,7 @@ endif( )
 mark_as_advanced( OPENCL_LIBRARIES )
 
 include( FindPackageHandleStandardArgs )
-find_package_handle_standard_args( OPENCL DEFAULT_MSG OPENCL_LIBRARIES OPENCL_INCLUDE_DIRS )
+find_package_handle_standard_args( OpenCL DEFAULT_MSG OPENCL_LIBRARIES OPENCL_INCLUDE_DIRS )
```
atamazov, line 78: Is this change backward compatible?

test/CMakeLists.txt
```diff
@@ -177,7 +177,7 @@ option( WORKAROUND_ISSUE_1334 "" ${WORKAROUND_ISSUE_1334_DEFAULT})
 if(NOT MIOPEN_TEST_MIOTENSILE)
 	if(MIOPEN_TEST_HALF)
 	    if(MIOPEN_BACKEND_OPENCL)
-		set(SKIP_TESTS test_gru test_rnn_vanilla test_lstm)
+		    set(SKIP_TESTS test_gru test_rnn_vanilla test_lstm)
 	    endif()
 	elseif(MIOPEN_TEST_INT8)
 	    set(SKIP_ALL_EXCEPT_TESTS
```
atamazov, line 183: [Notice] This block of code contains tabs. Formatting will look weird in editors that have tab size != 4.

test/CMakeLists.txt
```diff
@@ -223,6 +223,16 @@ if(MIOPEN_TEST_GFX1030)
...
 
+#TODO Code Quality WORKAROUND ROCm 5.1 update
+if(MIOPEN_BACKEND_OPENCL AND MIOPEN_TEST_ALL)
+    if(MIOPEN_TEST_GFX900 OR MIOPEN_TEST_GFX906)
+        list(APPEND SKIP_TESTS test_bn_3d_spatial_test test_conv3d test_immed_conv3d test_immed_conv2d)
+    endif()
+    if(MIOPEN_TEST_GFX1030)
+        list(APPEND SKIP_TESTS test_conv3d test_immed_conv3d test_immed_conv2d)
+    endif()
+endif()
```
atamazov, line 234: I highly recommend creating a ticket and mentioning the ticket number for this W/A here.

test/ford.hpp
atamazov: @JehandadKhan IIRC since C++11 std::move prevents copy elision. std::move is useless because when copy elision is not applicable, the compiler should try move constructor. Ditto two other cases.

atamazov (commented):
@junliume Some comments mostly related to process.

CMakeLists.txt
```diff
@@ -551,6 +551,18 @@ enable_clang_tidy(
...
+        -bugprone-easily-swappable-parameters
+        -bugprone-implicit-widening-of-multiplication-result
+        -cert-err33-c
+        -google-readability-casting
+        -hicpp-use-emplace
+        -modernize-use-emplace
+        -performance-unnecessary-copy-initialization
+        -readability-container-data-pointer
+        -readability-identifier-length
+        -readability-suspicious-call-argument
```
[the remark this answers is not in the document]
junliume: Yes this effort is to stabilize CI since it has been troublesome for a while when baseOS has been upgraded. I hope this change will at least let CI be functional, and I will create tickets regarding the TODOs.

PR #1505, opened 2022-04-06
[MLIR] Querying all sub-kernels applicability

jerryyin:
In MLIR solver, I discovered a bug where, applicability check and get solution ended to get different result. The reason ended up being we only query the first kernel's applicability in MLIR whereas second kernel is in-applicable.

In this PR, I fixed this issue by making performance config to mark as applicable only when all subsequent kernels are applicable.


jerryyin:
@junliume Gentle ping: Look like this has passed CI. Could you help with merge so as to make sure this will go to ROCm 5.2?

junliume (approved):

jerryyin:
@atamazov Thanks for visiting back on this recommendation :-)

atamazov:
@jerryyin the result of routine check of my mailbox ;)

PR #1509, opened 2022-04-08
Fix Warning - Update rocBLAS re-org PATH

arvindcheru:
- Updated rocblas include path. 
- Fix internal build warning (reorg deprecated version warnings)


junliume (approved):

junliume:
@arvindcheru This change has to be reverted since it breaks MIOpen CI.
**It is another backward Incompatible change which could be problematic with MIOpen CI, staging CI, and even mainline builds**
Could you let us know the coordinated sync up timing for this change to happen?

PR #1518, opened 2022-04-11
Make int8 convolution default output type to be int32

jerryyin:
This PR makes MIOpenDriver's int8 convolution output native type to be int32, based on:

 - MIOpen's gemm is based on rocblas, whose int8 output type is int32
   - There's an unnecessary `CastTensor` operation forced in every int8 validation
 - More fundamentally, AMD XDLOPS instructions output type for int8 calculation is int32
   - This implied high performance kernel will pick int32 as the native output type 
 - Numerical precision has been compromised by forcing conversion from int32 to fp32
   - Not all int32 can be precisely converted, see [here](https://stackoverflow.com/questions/12442685/can-a-ieee-754-real-number-cover-all-integers-within-its-range)
 - fp32 type incorrectly becomes the first class citizen for int8
   - If there's a solver who can support only int32 as the output, it will be marked as in-applicable

Locally I tested the change and make sure no regression appear on existing gemm kernel on int8 convolutions.

[TODO]: Make the same change for `conv2d_test`

### DIFF
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -320,7 +320,7 @@ private:
 
     std::vector<Tgpu> din;
     std::vector<Tgpu> dwei;
-    std::vector<float> out_int8;
+    std::vector<int32_t> out_int8;
     std::vector<Tgpu> db;
     std::vector<float> b_int8;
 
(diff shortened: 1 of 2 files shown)


DrizztDoUrden (approved):

JehandadKhan (approved):

junliume (approved):

atamazov (approved):
LGTM

PR #1520, opened 2022-04-12
[kernel references List]Development kernel and kernel include batches

DrizztDoUrden:
This provides separate lists for in development kernels and kernel includes. Using them instead of main lists should significantly reduce rebuild time when changing the kernel source. At the end of the kernel development cycle (before merging the PR) they should be moved to main lists to free up build time. This is a fix for our unity build like kernel inline process.

Additional build time reduction may be reached if kernel references were added to a separate map as the main map file would not require regeneration and rebuild.

This PR is probably worth being noted to kernel developers who are/would like to try iterating their sources in the MIOpen itself.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -442,8 +442,28 @@ if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN
         kernels/xform_bidirect_winograd_out.s
         kernels/UniversalTranspose.cl)
 
+    # Kernels in development lists.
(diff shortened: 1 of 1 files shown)


DrizztDoUrden:
@atamazov your review would be appreciated.

JehandadKhan:
@dmikushin Can you please check if this works for you ?

DrizztDoUrden:
I can add to this PR moving them to a separate map. That would reduce build time a bit as well, but not nearly as significant. May not even worth the effort.

atamazov (approved):
LGTM! Let's inform all kernel developers about this feature.

junliume (approved):
@asroy @zjing14 @carlushuang @qianfengz Please be noted about this change.

PR #1522, opened 2022-04-17
[CI] Enable `ccache` for Docker and Jenkins

JehandadKhan:
The PR 

* enables `ccache` for MIOpen in the Dockerfile and the Jenkins
* refactors the Dockerfile to improve the chances of cache hits


JehandadKhan:
@okakarpa Please let me know when the redis server is setup so we can test this in the CI

Dockerfile
```diff
@@ -94,21 +91,35 @@ ADD requirements.txt /requirements.txt
 ADD dev-requirements.txt /dev-requirements.txt
 # Install dependencies
 # TODO: Add --std=c++14
+ARG GPU_ARCH=";"
+ARG PREFIX=/usr/local
+ARG USE_FIN="OFF"
+ARG CCACHE_SECONDARY_STORAGE="redis://10.236.118.172"
+# RUN cget -p $PREFIX install https://github.com/ccache/ccache/archive/7f1572ae9ca958fa923a66235f6a64a360b03523.tar.gz -DZSTD_FROM_INTERNET=ON -DHIREDIS_FROM_INTERNET=ON
+RUN rm -rf /tmp/ccache* && mkdir /tmp/ccache && wget https://github.com/ccache/ccache/archive/7f1572ae9ca958fa923a66235f6a64a360b03523.tar.gz -O /tmp/ccache.tar.gz && \
```
pfultz2, line 99: I assume you are not using cget here because rbuild cleans out the packages before building, but you can list this in the rbuild.ini file so it will install for the develop and fin sessions.

Dockerfile
```diff
@@ -94,21 +91,35 @@ ADD requirements.txt /requirements.txt
 ADD dev-requirements.txt /dev-requirements.txt
 # Install dependencies
 # TODO: Add --std=c++14
+ARG GPU_ARCH=";"
+ARG PREFIX=/usr/local
+ARG USE_FIN="OFF"
+ARG CCACHE_SECONDARY_STORAGE="redis://10.236.118.172"
```
pfultz2, line 97: This should only be set in the jenkinsfile. We dont want users caching there builds here.

Dockerfile
```diff
@@ -94,21 +91,35 @@ ADD requirements.txt /requirements.txt
...
+    cd /tmp/ccache-7f1572ae9ca958fa923a66235f6a64a360b03523/build && \
+    cmake -DZSTD_FROM_INTERNET=ON -DHIREDIS_FROM_INTERNET=ON .. && make -j install
+RUN ccache -s 
+ARG COMPILER_LAUNCHER=""
 RUN if [ "$USE_FIN" = "ON" ]; then \
-        rbuild prepare -s fin -d $PREFIX -DAMDGPU_TARGETS=${GPU_ARCH}; \
+        rbuild prepare -s fin -d $PREFIX -DAMDGPU_TARGETS=${GPU_ARCH} -DCMAKE_CXX_COMPILER_LAUNCHER="${COMPILER_LAUNCHER}"; \
     else \
-        rbuild prepare -s develop -d $PREFIX -DAMDGPU_TARGETS=${GPU_ARCH}; \
+        rbuild prepare -s develop -d $PREFIX -DAMDGPU_TARGETS=${GPU_ARCH} -DCMAKE_CXX_COMPILER_LAUNCHER="${COMPILER_LAUNCHER}"; \
```
pfultz2, line 108: It would be nice to set `CMAKE_CXX_COMPILER_LAUNCHER` in the ini file, but its a little complicated as it needs to be set after installing ccache.

In MIGraphX, we do set this variable in the ini file but only for the project and not the dependencies which is not helpful here:

https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/rbuild.ini#L25

I need to update rbuild to better support such workflows.

Dockerfile
JehandadKhan: This is temporary

Dockerfile
JehandadKhan: The issue is that I want to use ccache while building the stuff in rbuild.ini that is why I install it directly

Dockerfile
JehandadKhan: Fixed

pfultz2 (approved):

junliume (approved):

PR #1524, opened 2022-04-21
Composable Kernel Integration

JehandadKhan:


junliume:
@JehandadKhan the test `test_conv_for_implicit_gemm` for `BF16` failed on MI200, maybe we should investigate it.

JehandadKhan:
@asroy  @atamazov  Please review when convenient

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -0,0 +1,256 @@
...
+    const auto args        = CKArgs{ctx};
+    result.invoker_factory = [=](const std::vector<Kernel>& kernels) {
+        std::ignore = kernels;
+        return [=](const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            std::vector<DeviceConvFwdPtr_t> conv_ptrs;
+            add_device_conv2d_fwd_xdl_nhwc_kyxc_nhwk_int8_instances_t(conv_ptrs);
+            auto& conv_ptr       = conv_ptrs.at(config.index);
+            const auto& data_ctx = primitive_parameters.CastTo<conv::DataInvokeParams>();
+            const auto& tensors  = data_ctx.tensors;
+            auto argument_ptr    = conv_ptr.MakeArgumentPointer(
```
atamazov, line 223: [Performance] AFAICS the most of computations of kernel arguments can be moved outside of both lambdas (Invoker and InvokerFactory). These arguments may be computed right in GetSolution (or in InvokerFactory) and captured by value by InvokerFactory (or by Invoker). This works because the Solver produces distinct instance of InvokerFactory in each Solution, and then the InvokerCache produces and maintains distinct instance of Invoker per each ProblemConfig. When the library wants to run a Solution, it runs Invoker, which simply initializes pointers to the actual I/O buffers and calls the kernel -- much less work than now.

Please look how `opArgs` is used in `MakeImplGemmDynamicBackwardDataXdlopsNHWCInvokerFactory()`, for example. In #680 you can find some review comments related to this topic.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
```diff
@@ -65,6 +65,8 @@ bool ConvMlirIgemmFwdXdlops::IsApplicable(const ConvolutionContext& ctx) const
         return false;
     if(!IsComposableKernelSupportedHardware(ctx))
         return false;
+    if(ctx.IsInt8())
+        return false;
```
atamazov, line 69: Can we remove this? This condition should be tested by `MiirIsConfigApplicable(...)`, one line below.

test/conv2d.cpp
```diff
@@ -56,6 +56,9 @@ struct conv2d_driver : conv_driver<T>
         this->add(this->in_layout, "in_layout", this->generate_data({"NCHW"}));
         this->add(this->fil_layout, "fil_layout", this->generate_data({"NCHW"}));
         this->add(this->out_layout, "out_layout", this->generate_data({"NCHW"}));
+        // Only valid for int8 input and weights
+        this->add(this->output_type, "output_type", this->generate_data({"int32"}));
+        this->add(this->int8_vectorize, "int8_vectorize", this->generate_data({false}));
```
atamazov, line 61: [Q] Can you please remind me what are the side effects of these lines? Thanks!

atamazov (changes requested):
Refactored tuning machinery in the Solver looks very good 👍

junliume (dismissed):

junliume (approved):

junliume:
@atamazov @JehandadKhan CI has passed and I plan to merge this PR first. Please summarize the review opinions above in another post-merge issue.

junliume:
@asroy @zjing14 @JehandadKhan  @atamazov @carlushuang It is a great milestone and thank you all for your efforts! Congratulations on the first step of integration!

atamazov:
🔴 It seems like this breaks develop.

@junliume 
> It is a great milestone and thank you all for your efforts!

Thanks, Jun. However, maybe I am wrong, but it seems like specifically my efforts were useless because none of my review comments were resolved ;)

src/include/miopen/solver.hpp
```diff
@@ -4196,6 +4196,59 @@ struct ConvAsmImplicitGemmGTCDynamicFwdDlopsNCHWC final
                 const PerformanceConfigAsmImplicitGemmGTCFwdDlopsNCHWC& config) const override;
 };
 
+struct PerformanceConfigHipImplicitGemmFwdXdlops
+    : Serializable<PerformanceConfigHipImplicitGemmFwdXdlops>
```
shurale-nkn, line 4200: Why based on Serializable? should be based on PerfConfig

junliume:
@atamazov @JehandadKhan yes this PR is reverted due to its breaking of develop. We are working on an urgent fix.

src/include/miopen/solver.hpp
shurale-nkn: For classes based on PerfConfig, we have defined output operator. But for Serializable we don't.

test/conv2d.cpp
DrizztDoUrden: This registers them as cmd line arguments and defines data generated for the full run.

src/convolution.cpp
```diff
@@ -265,8 +265,9 @@ ConvolutionDescriptor::GetForwardOutputTensorWithLayout(const TensorDescriptor&
     tensor_layout_to_strides(
         out_lens, default_layout, yLayout, xDesc.GetVectorLength(), out_strides);
     return {(xDesc.GetType() == miopenInt8 || xDesc.GetType() == miopenInt8x4
-                 ? (yType == miopenInt32 ? yType : miopenFloat)
-                 : xDesc.GetType()),
+                 ? (yType)
+                 : xDesc.GetType()), // TODO: This function overrides the output type with
+                                     // essentially the input which is incorrect.
```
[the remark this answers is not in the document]
JehandadKhan: I believe this was added as a workaround to support int8x4 without the rest of MIOpen supporting vectorized data types. Therefore for int8 convolutions we would use float as the "temporary" data type to hold the four elements.

src/solver/conv_mlir_igemm_fwd_xdlops.cpp
JehandadKhan: The `MiirIsConfigApplicable` is relying on a string definition of the context, so we would need to pass in the context as an additional argument to be able to call the `IsInt8` method which seems like shoe-horning something into an otherwise MLIR specific method?

test/conv2d.cpp
JehandadKhan: This adds two arguments to the conv_2d test driver which can then be set in the CMakeLists.txt file to execute test cases.

src/include/miopen/solver.hpp
JehandadKhan: This was a merge error since Evgenii's PR was merged in the interim, I have fixed the error.

shurale-nkn: oh, that explains why test failed only in develop.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
JehandadKhan: I agree, however, the challenge is that since the `composable_kernel` API requires `unique_ptr` which makes capture by value difficult. For more details [please see here](https://github.com/ROCmSoftwarePlatform/composable_kernel/blob/91d8b7d67ae9dbf8a6e691ea3e17c0b9705c6ba7/library/include/ck/library/host/host_interface.hpp)

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -0,0 +1,256 @@
...
+    return GenericSearch(*this, ctx, invoke_ctx);
+}
+
+size_t ConvHipImplicitGemmFwdXdlops::GetWorkspaceSize(const ConvolutionContext& ctx) const
+{
+    std::ignore = ctx;
+    return 0;
+}
+
+bool ConvHipImplicitGemmFwdXdlops::IsApplicable(const ConvolutionContext& ctx) const
```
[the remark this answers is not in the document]
JehandadKhan: The idea is to extend it further in the future, I would not like to pollute the solver id space with minor changes in the solver.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
JehandadKhan: Please let me know if you have any suggestions to improve it.

test/CMakeLists.txt
```diff
@@ -141,7 +141,7 @@ else()
 endif()
 
 find_package(Threads REQUIRED)
-add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -C ${CMAKE_CFG_INTDIR})
+add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} -VV --output-on-failure -C ${CMAKE_CFG_INTDIR})
```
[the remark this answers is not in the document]
JehandadKhan: Not needed, its been removed from the branch but somehow lingered in the Git Diff

src/solver/gemm.cpp
```diff
@@ -87,6 +87,13 @@ bool GemmFwdBase::IsApplicable(const ExecutionContext& ctx,
     const auto& xDesc = problem.GetIn();
     const auto& wDesc = problem.GetWeights();
     const auto& yDesc = problem.GetOut();
+    if(xDesc.GetType() == miopenInt8x4 || xDesc.GetType() == miopenInt8)
+    {
+        // rocBlas needs the output to be int32 always
```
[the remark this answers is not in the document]
JehandadKhan: Yes, thats what it means, and its not a bug, its a feature (no pun intended)

Issue #1516, opened 2022-04-09
[Navi21] test_conv_ck_igemm_fwd_v6r1_dlops_nchw fails


Slimakanzer:
FP16, HIP backend

> 59/99 Test #94: test_conv_ck_igemm_fwd_v6r1_dlops_nchw ................***Failed   40.40 sec
>
> build/bin/test_conv2d --half --cmode conv --pmode default --group-count 1 --disable-backward-data --disable-backward-weights --input 128 1024 14 14 --weights 2048 1024 1 1 --batch_size 128 --input_channels 1024 --output_channels 2048 --spatial_dim_elements 14 14 --filter_dims 1 1 --pads_strides_dilations 0 0 2 2 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW
> 
> Memory access fault by GPU node-1 (Agent handle: 0x10d4e30) on address 0x7f9e8fd8e000. Reason: Page not present or supervisor privilege.

atamazov:
@Slimakanzer Good finding! GPU type, ROCm version, Linux version, AMDGPU driver version? Is it reproducible in our CI?

junliume:
@Slimakanzer @atamazov This issue has been observed recently and not only on Navi21s. It happens occasionally and restart the pipeline most times it will pass. I tested it locally on a Navi21 with 5.1 for both base OS and docker, and it consistently passes.
@asroy @zjing14 @carlushuang AFAIK this is a legacy CK kernel which will be replaced soon?

atamazov:
@junliume 

> This issue has been observed recently and not only on Navi21s. It happens occasionally and restart the pipeline most times it will pass. 

This looks like a stability issue. Added to #1147.

If this error happens on our CI, and we still use 4.3.1 there, then I recommend postponing this until CI is upgraded.

PR #1530, opened 2022-04-27
[CI][Tests] Workaround #1516 by disable legacy CK kernel test

junliume:
test passed on local Navi21 with ROCm 5.1.0 installed as both base and docker

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1538,7 +1538,9 @@ set(CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW_ENV
     MIOPEN_FIND_MODE=normal
     MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvCkIgemmFwdV6r1DlopsNchw)
 
-add_custom_test(test_conv_ck_igemm_fwd_v6r1_dlops_nchw FLOAT_ENABLED HALF_ENABLED BF16_DISABLED GFX1030_ENABLED SKIP_UNLESS_ALL
+#TODO Code Quality WORKAROUND ROCm 5.1 update
+# WORKAROUND #1516 #1529
+add_custom_test(test_conv_ck_igemm_fwd_v6r1_dlops_nchw FLOAT_ENABLED HALF_ENABLED BF16_DISABLED GFX1030_DISABLED SKIP_UNLESS_ALL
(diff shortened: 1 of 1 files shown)


atamazov (changes requested):
I recommend re-implementing this so the test is automatically enabled with 5.1. In addition, I recommend disabling iGemm V6R1 until 5.1 as well. Detailed suggestions will follow soon.

atamazov:
> I recommend re-implementing this so the test is automatically enabled with 5.1.

In test/cmakelists.txt, near line 177:
```C++
set_var_to_condition(WORKAROUND_ISSUE_1516_DEFAULT ${MIOPEN_TEST_GFX1030} AND (${MIOPEN_hip_VERSION_FLAT} GREATER 500100000))
option( WORKAROUND_ISSUE_1516 "" ${WORKAROUND_ISSUE_1516_DEFAULT})
```
In add_custom_test():
```C++
    if(WORKAROUND_ISSUE_1516
        AND (${NAME} MATCHES "test_conv_ck_igemm_fwd_v6r1_dlops_nchw" ))
        set_tests_properties(${NAME} PROPERTIES DISABLED On)
    endif()
```

atamazov:
>  I recommend disabling iGemm V6R1 until 5.1 as well

In conv_ck_igemm_fwd_v6r1_dlops_nchw.cpp:
```C++
#include <config.h>
.
.
#define WORKAROUND_ISSUE_1516 (HIP_PACKAGE_VERSION_FLAT < 5001000000)
.
.
bool ConvCkIgemmFwdV6r1DlopsNchw::IsApplicable(const ConvolutionContext& ctx) const
{
#if WORKAROUND_ISSUE_1516
    if(!miopen::IsEnabled(MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW{}))
        return false;
#endif
    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW{}))
        return false;
.
.
```

atamazov:
@junliume BTW do we know what is the faulty component? For example, is issue reproducible with 5.0 base driver and ROCm 5.1?

junliume:
> @junliume BTW do we know what is the faulty component? For example, is issue reproducible with 5.0 base driver and ROCm 5.1?

@atamazov  thanks for the suggestions! The current issue is that `test_conv_ck_igemm_fwd_v6r1_dlops_nchw` fails randomly in CI (CI nodes have ROCm 5.0.2 installed as base OS and ROCm 5.1 in dockerfile); while I cannot reproduce it with another standalone dev machine (ROCM 5.1.0 installed as both base OS and docker). 

#1516 shows that we plan to "fix" it by CK integration which will happen with #1524. I think maybe I should put this PR on HOLD until CK tells us whether this test should be kept active. Thanks!

aska-0096:
Hi @junliume , has encountered this problem twice during the PR1572 CI.
http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/issue_1569__driver_support_int8conv/3/pipeline

PR #1523, opened 2022-04-18
[Solver & Kernel] Integration of MISA NCHWc fp16x4 and fp16x8 asm kernel .

aska-0096:
[Kernel side]
1. [New solver] conv_asm_implicit_gemm_gtc_fwd_nchwc solver support 
2. [New kernel] NCHWc x NCHWc x NCHWc,  NCHWc x CHWNc x NCHWc layout with fp16x8 and fp16x4 fwd kernels based on Dlops for gfx1030. 

usage:
For building the ctest: 

1. cmake --build . --config Release --target test_conv_igemm_dynamic_dlops_nchwc_nchwc_fwd_fp16x4
2. cmake --build . --config Release --target test_conv_igemm_dynamic_dlops_nchwc_nchwc_fwd_fp16x8
3. cmake --build . --config Release --target test_conv_igemm_dynamic_dlops_nchwc_chwnc_fwd_fp16x4
4. cmake --build . --config Release --target test_conv_igemm_dynamic_dlops_nchwc_chwnc_fwd_fp16x8                                     

For executing MIOpenDriver: 

1. ./bin/MIOpenDriver convfp16 -I NCHW -O NCHW -f NCHW -n 2 -W 32 -H 32 -c 32 -k 32 -x 5 -y 5 -p 2 -q 2 -Z 1 -L 4 -F 1 -V 1
2. ./bin/MIOpenDriver convfp16 -I NCHW -O NCHW -f NCHW -n 2 -W 32 -H 32 -c 32 -k 32 -x 5 -y 5 -p 2 -q 2 -Z 1 -L 8 -F 1 -V 1
3. ./bin/MIOpenDriver convfp16 -I NCHW -O NCHW -f CHWN -n 2 -W 32 -H 32 -c 32 -k 32 -x 5 -y 5 -p 2 -q 2 -Z 1 -L 4 -F 1 -V 1
4. ./bin/MIOpenDriver convfp16 -I NCHW -O NCHW -f CHWN -n 2 -W 32 -H 32 -c 32 -k 32 -x 5 -y 5 -p 2 -q 2 -Z 1 -L 8 -F 1 -V 1

cc: @atamazov

Issue #767, opened 2021-03-03
GetLayout() ambiguous while parsing tensor layout if a dimension is 1


carlushuang:
As in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/765, function [TensorDescriptor::GetLayout()](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/tensor.hpp#L188) will produce wrong result if a single dimension is 1.

e.g. if a depthwise weight length is `[63, 1, 3, 3]`, and we want to calculate the stride of `NHWC` layout, which is `[9, 1, 3, 1]`. The output of `GetLayout()` will be `NHCW` which is not wanted.
This is still true in 3d weight `[32,1,3,5,20]`, the stride of `NDHWC` will be `[300,1,100,20,1]` , which will produce wrong layout string as `NDHCW`.

~However, under this case, the stride of either `NDHCW` or `NDHWC` layout is indeed `[300,1,100,20,1]`, the `C` and `W` dimension seems exchangeable. So in such case (C=1) if we only look at the stride, it will have ambiguous meaning, not only one layout is valid.~

stride of `[9, 1, 3, 1]` and length of `[63, 1, 3, 3]` should be `NHWC` layout not `NHCW` layout.
And again, if we have multiple dimension of `1`, will lead to ambiguous layout result

However the ambiguous case would be very common if in `1x1` filter case.

e.g.
wei length: `[32, 3, 1, 1]`
1. we want `NHWC` -> stride `[3, 1, 3, 3]` -> valid layout could be `NHWC`/`NWHC` (2 ambiguous)
2. we want `CHWN` -> stride `[32, 32, 32, 1]` -> valid layout would be `CHWN`/`CWHN` (2 ambiguous)

wei length: `[32, 1, 1, 1]` (depth wise 1x1 filter)
1. we want `NHWC` -> stride `[1, 1, 1, 1]` -> valid layout would be `NCHW`/`NCWH`/`NHWC`/`NHCW`... (6 ambiguous)
2. we want `CNHW` -> stride `[1, 32, 1, 1]` -> valid layout would be `CNHW`/`CNWH` (2 ambiguous)

cc @jerryyin @pfultz2

jerryyin:
It's unfortunate that we didn't realize this in #343.

[...]

PR #1532, opened 2022-04-28
[API] NCHWc tensor layout support

aska-0096:
This PR is API only part of  #1523 

[New API]

```
extern "C" miopenStatus_t miopenSetNdTensorDescriptorWithLayout(miopenTensorDescriptor_t tensorDesc,
                                                                miopenDataType_t dataType,
                                                                miopenTensorLayout_t tensorLayout,
                                                                int* lens,
                                                                int  numLens)
```
NCHWc layout will be identified as a 4-D tensor with last dimension stride equal to vector length (Currentlly only vector_c).
This API create the 4-D tensor descriptor with
dim lengths of (n, c/vector_c, h, w)
strides of (chw, hw*vector_c, w*vector_c, vector_c).

[Host side]

1. New enum type miopenTensorLayout_t to support different user input tensor layout.
2. New command line flag vectorLength with short name `L`.
3. New member of tensorDescriptor, vectorLength and tensorLayout
4. New tensor and tensorDescriptor constructor include the tensorLayout parameter.

[CPU convolution]

1. A new way to get index of a pixel in vectorized tensor.
2. CPU convolution of NCHWc * NCHWc and NCHWc * CHWNc

For executing MIOpenDriver:

./bin/MIOpenDriver convfp -I NCHW -O NCHW -f NCHW -z 1 -L 4 -n 2 -W 32 -H 32 -c 32 -k 32 -x 5 -y 5 -p 2 -q 2 -F 1 -V 1
./bin/MIOpenDriver convfp -I NCHW -O NCHW -f NCHW -z 1 -L 8 -n 2 -W 32 -H 32 -c 32 -k 32 -x 5 -y 5 -p 2 -q 2 -F 1 -V 1
./bin/MIOpenDriver convfp -I NCHW -O NCHW -f CHWN -z 1 -L 4 -n 2 -W 32 -H 32 -c 32 -k 32 -x 5 -y 5 -p 2 -q 2 -F 1 -V 1
./bin/MIOpenDriver convfp -I NCHW -O NCHW -f CHWN -z 1 -L 8 -n 2 -W 32 -H 32 -c 32 -k 32 -x 5 -y 5 -p 2 -q 2 -F 1 -V 1

CC: @atamazov


junliume:
@aska-0096 CI passed with [16abc8d](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532/commits/16abc8dc7d635abd4257fccc8251f7a689fd00ab) but failed with [f242a9f](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532/commits/f242a9fabe0ec2cf8cfdc3b2112a6a7f901b9df7).

Hence
```
src/find_controls.cpp:127: Invalid value of MIOPEN_DEBUG_FIND_ONLY_SOLVER: ConvAsmImplicitGemmGTCDynamicFwdDlopsNCHWC

**Comment by aska-0096** [comment]
> @aska-0096 CI passed with [16abc8d](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532/commits/16abc8dc7d635abd4257fccc8251f7a689fd00ab) but failed with [f242a9f](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532/commits/f242a9fabe0ec2cf8cfdc3b2112a6a7f901b9df7).
> 
> Hence
> 
> ```
> src/find_controls.cpp:127: Invalid value of MIOPEN_DEBUG_FIND_ONLY_SOLVER: ConvAsmImplicitGemmGTCDynamicFwdDlopsNCHWC
> ```

I see. I will delete the nchwc kernel ctest part.

File: include/miopen/miopen.h [hunk 1]
```diff
@@ -337,8 +337,27 @@ typedef enum
...
+ * miopenTensorNCHW_VECT_C and miopenTensorCHWN_VECT_C layout are only supported with tensor data
+ * types miopenHalfx4 and miopenHalfx8. miopenTensorCHWN_VECT_C layout only support weight tensor.
+ */
+typedef enum
+{
+    miopenTensorNCHW        = 0, /*!< NCHW memory layout (Fully supported) */
+    miopenTensorNHWC        = 1, /*!< NHWC memory layout (Fully supported) */
+    miopenTensorCHWN        = 2, /*!< CHWN memory layout (Not supported) */
+    miopenTensorNCHW_VECT_C = 3, /*!< NCHW_VECT_C memory layout (Partially supported) */
+    miopenTensorCHWN_VECT_C = 4, /*!< CHWN_VECT_C memory layout (Partially supported) */
```
atamazov, line 358: [important] Unfortunately, API is cumbersome as @shurale-nkn mentions [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1523#discussion_r860952292). 

(1) Do we foresee vectorization other than VECT_C? If not, then we can avoid adding a new API call and derive vectorized tensor layout internally, from the datatype.

(2) Otherwise we definitely need a new API call. But in this case I highly recommend developing an orthogonal API where pack (vector) size (x8, x4) and vectorized dimension (VECT_C) are separated from the datatype and tensor layout. Which means that the new API call would have two new parameters added (pack size and the index of dimension to be vectorized).

(3) The 3rd approach is adding a call like `miopenSetTensorDescriptorAttribute()` instead of this new `miopenSet4dTensorDescriptorWithLayout()`.

Of course, the above doesn't apply if maximum cudnn compatibility is paramount.

@junliume @JehandadKhan @pfultz2 What do you think?

JehandadKhan: Thanks for the feedback @atamazov. I have posed similar questions to the frameworks team to get their input and will update here as soon as I hear back

atamazov:
[Notice] I have to pause my reviewing until there is a response to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532#discussion_r862108183

include/miopen/miopen.h
atamazov: @JehandadKhan Great!

aska-0096:
> [Notice] I have to pause my reviewing until there is a response to [#1532 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532#discussion_r862108183)

Thanks @atamazov, I got it. Let's wait JD's feed back from framework team.

include/miopen/miopen.h
```diff
@@ -337,8 +337,27 @@ typedef enum
     miopenBFloat16 = 5, /*!< 16-bit binary floating point (8-bit exponent, 7-bit fraction)
                            (Partially supported) */
     miopenDouble = 6,   /*!< 64-bit floating point (Partially supported) */
+    miopenHalfx4 = 7,   /*!<Pack of 4 16-bit floating point values. Suitable for vectorized tensor
+                           formats only.*/
+    miopenHalfx8 = 8,   /*!<Pack of 8 16-bit floating point values. Suitable for vectorized tensor
+                           formats only.*/
```
JehandadKhan, line 343: I suggest we remove these and use special layout enum members to specify the packing dimension. That way the data type enum would only specify what the data type is and the layout enum would specify what the layout and the packing dimension is.

include/miopen/miopen.h
JehandadKhan: @aska-0096  In place of the new members you created can we do something along the lines of 

```suggestion
    miopenTensorNCHWc4 = 3, /*!< NCHWc4 memory layout (Partially supported) */
    miopenTensorNCHWc8 = 4, /*!< NCHWc8 memory layout (Partially supported) */
    miopenTensorCHWNc4 = 4, /*!< CHWNc4 memory layout (Partially supported) */
    miopenTensorCHWNc8 = 4, /*!< CHWNc8 memory layout (Partially supported) */
```
cc/ @atamazov

JehandadKhan (changes requested):
@aska-0096 Upon further deliberation I think we should decouple the data type from the layout enum. This would reduce the API footprint of these changes and also provides better conceptual division of concerns resulting in a cleaner API design.

include/miopen/miopen.h
aska-0096: Thanks JD, I will modify this part.

include/miopen/miopen.h
atamazov: So the decision is to add new tensor layouts to the API. This seems acceptable.

src/conv/problem_description.cpp
```diff
@@ -64,7 +64,7 @@ void ProblemDescription::HeuristicUpdateLayouts()
 {
     const std::string labels = tensor_layout_get_default(in_layout.size());
 
-    static const std::vector<std::string> supported_layouts = {"NCHW", "NHWC", "NCDHW"};
+    static const std::vector<std::string> supported_layouts = {"NCHW", "NHWC", "CHWN", "NCDHW"};
```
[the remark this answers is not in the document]
aska-0096: In PR #1523 , convolution kernel with NCHWc(input tensor) x CHWNc (weight tensor) = NCHWc (output tensor) are implemented, that's why I include CHWN layout into supported layouts. I'm not sure a separate PR is possible or not, since we only use CHWNc layout as weight tensor.

include/miopen/miopen.h
```diff
@@ -579,6 +600,28 @@ MIOPEN_EXPORT miopenStatus_t miopenCreateTensorDescriptor(miopenTensorDescriptor
...
+ * @param dataType     MIOpen datatype (input)
+ * @param tensorLayout Tensor layout (input)
+ * @param n            Mini-batch size (input)
+ * @param c            Number of channels (input)
+ * @param h            Data height dimension size (input)
+ * @param w            Data width dimension size (input)
+ * @return             miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t
+miopenSet4dTensorDescriptorWithLayout(miopenTensorDescriptor_t tensorDesc,
```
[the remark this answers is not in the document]
aska-0096: I think this is a problem about when we package the information from command line
1. Artem suggested one: we get `(tensorLayout, vectorizedPosition, vectorLength) `three arguments from command line pass them into API then package them into a vectorized layout inside .
2. Original one in this PR: we get  `(tensorLayout, vectorizedPosition, vectorLength) `three arguments from command line and package them into a  vectorized layout outside then only pass this packaged arguments(vectorizedLayout) into API.

I'm not sure which one is better. JD, what do you think about it? @JehandadKhan

include/miopen/miopen.h
aska-0096: @JehandadKhan Hi, this change has been implemented, would you like to say something about Artem's question?

junliume:
CI has passed, ping reviewers: @JehandadKhan @DrizztDoUrden @shurale-nkn @carlushuang

carlushuang (dismissed):
LGTM, but some review from @JehandadKhan need resolve or double check

src/tensor.cpp
```diff
@@ -99,15 +154,25 @@ int TensorDescriptor::GetSize() const
 std::size_t TensorDescriptor::GetElementSize() const
 {
     assert(lens.size() == strides.size());
-    return std::accumulate(
-        lens.begin(), lens.end(), std::size_t{1}, std::multiplies<std::size_t>());
+    return std::accumulate(lens.begin(), lens.end(), vector_c, std::multiplies<std::size_t>());
 }
 miopenDataType_t TensorDescriptor::GetType() const { return this->type; }
+miopenTensorLayout_t TensorDescriptor::GetLayout_t() const { return this->tensorLayout; }
+int TensorDescriptor::GetVectorLength() const { return this->vector_c; }
```
DrizztDoUrden, line 161: I would suggest syncing the names. Either `GetVectorC` for the method or `vector_length` for the field.

src/tensor_api.cpp
```diff
@@ -48,6 +48,23 @@ extern "C" miopenStatus_t miopenSet4dTensorDescriptor(
     });
 }
 
+extern "C" miopenStatus_t miopenSet4dTensorDescriptorWithLayout(miopenTensorDescriptor_t tensorDesc,
```
DrizztDoUrden, line 51: Why not make it `...Nd...` with `int* lens` and `int num_lens` parameters?

src/tensor_api.cpp
```diff
@@ -91,6 +108,15 @@ extern "C" miopenStatus_t miopenGet4dTensorDescriptor(miopenTensorDescriptor_t t
 
 // Internal API
 // MD: This should not be required to be exported. Temporary hack
+MIOPEN_EXPORT miopenStatus_t
+miopenGet4dTensorDescriptorVectorLength(miopenTensorDescriptor_t tensorDesc, int* vectorLength)
```
DrizztDoUrden, line 112: This is definitely Nd rather than 4d

DrizztDoUrden (changes requested):

driver/conv_driver.hpp
```diff
@@ -594,6 +631,12 @@ int ConvDriver<Tgpu, Tref>::GetandSetData()
     SetConvDescriptorFromCmdLineArgs();
 
     std::vector<int> out_len = GetOutputTensorLengths();
+    if(miopen::deref(inputTensor).GetLayout_t() == miopenTensorNCHWc4 ||
+       miopen::deref(inputTensor).GetLayout_t() == miopenTensorNCHWc8)
+        out_len[1] *= miopen::deref(inputTensor).GetVectorLength();
```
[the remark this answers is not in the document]
aska-0096: Thanks for the suggestion. I'm not sure what "updating" actually mean: change or add the` .GetVectorLength() != 1`? We can not abandon the tensor layout related condition, because the vectorized dimension is at different position in NCHWc and CHWNc tensors.

driver/conv_driver.hpp
```diff
@@ -1201,20 +1244,44 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
...
-        wei = tensor<Tgpu>(miopen::deref(weightTensor).GetLengths(),
-                           miopen::deref(weightTensor).GetStrides());
-    if(is_fwd)
-        out = tensor<Tgpu>(miopen::deref(outputTensor).GetLengths(),
-                           miopen::deref(outputTensor).GetStrides());
-    if(is_bwd || is_wrw)
-        dout = tensor<Tgpu>(miopen::deref(outputTensor).GetLengths(),
-                            miopen::deref(outputTensor).GetStrides());
-
+    if(miopen::deref(inputTensor).GetTensorLayout().find("_VECT_") == std::string::npos)
```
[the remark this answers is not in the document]
aska-0096: good idea.

src/convolution.cpp
```diff
@@ -163,6 +163,12 @@ ConvolutionDescriptor::GetForwardOutputTensorWithLayout(const TensorDescriptor&
 
     auto wei_spatial = boost::adaptors::slice(wDesc.GetLengths(), 2, 2 + spatial_dim);
 
+    if(wDesc.GetLayout_t() == miopenTensorCHWNc4 || wDesc.GetLayout_t() == miopenTensorCHWNc8)
```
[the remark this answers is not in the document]
aska-0096: I see.

src/include/miopen/tensor_extra.hpp
```diff
@@ -31,6 +31,9 @@ MIOPEN_EXPORT int miopenGetTensorIndex(miopenTensorDescriptor_t tensorDesc,
 
 int miopenGetTensorDescriptorElementSize(miopenTensorDescriptor_t tensorDesc);
 
+MIOPEN_EXPORT miopenStatus_t
+miopenGet4dTensorDescriptorVectorLength(miopenTensorDescriptor_t tensorDesc, int* vectorLength);
```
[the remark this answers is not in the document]
aska-0096: Currently, we only support 4d vectorized tensor layout.

src/tensor.cpp
aska-0096: Updated in latest commit.

src/tensor_api.cpp
aska-0096: The tensor api designed here follow the fashion of other miopen tensor api style.

src/tensor_api.cpp
aska-0096: agree.

aska-0096:
@atamazov @JehandadKhan @DrizztDoUrden Any further review is appreciated. If there are nothing else remaining, I would like to ask @junliume to merge this PR after I resolve the conflicts.

src/convolution.cpp
JehandadKhan, line 166:
```suggestion
    if(wDesc.IsVectorized())
```

src/gemm_v2.cpp
```diff
@@ -845,7 +846,10 @@ miopenStatus_t CallGemmStridedBatched(const Handle& handle,
         }
         break;
         case miopenInt32: break;
+
         case miopenHalf: {
+            assert(gemm_desc.k % 4 == 0);
```
JehandadKhan, line 851: Please reconsider

src/include/miopen/conv/problem_description.hpp
```diff
@@ -178,7 +208,7 @@ struct ProblemDescription
     {
         if(GetSpatialDims() == 2)
         {
-            return in.GetLayout("NCHW");
+            return in.GetLayout(in.GetTensorLayout());
```
JehandadKhan, line 211: Perhaps it is no longer required

src/include/miopen/conv/problem_description.hpp
```diff
@@ -212,7 +242,7 @@ struct ProblemDescription
     {
         if(GetSpatialDims() == 2)
         {
-            return out.GetLayout("NCHW");
+            return out.GetLayout(out.GetTensorLayout());
```
JehandadKhan, line 245: Perhaps it is no longer required

src/include/miopen/conv/problem_description.hpp
```diff
@@ -245,7 +287,7 @@ struct ProblemDescription
     {
         if(GetSpatialDims() == 2)
         {
-            return weights.GetLayout("NCHW");
+            return weights.GetLayout(weights.GetTensorLayout());
```
JehandadKhan, line 290: Perhaps this is no longer required

src/include/miopen/tensor.hpp
```diff
@@ -147,12 +156,16 @@ struct TensorDescriptor : miopenTensorDescriptor
     }
 
     void CalculateStrides();
+    void CalculateVectorC();
```
JehandadKhan, line 159: Vectorization can be any dimension ( K/N etc) perhaps rename the method ?

src/include/miopen/tensor_layout.hpp
```diff
@@ -69,6 +69,42 @@ void tensor_layout_to_strides(const std::vector<T>& len,
                    });
 }
 
+template <typename T>
+void tensor_layout_to_strides_vect(const std::vector<T>& len,
```
JehandadKhan, line 73: Perhaps merge this function with the non-vectorized version since this function can subsume the other one

src/problem_description.cpp
```diff
@@ -44,6 +44,14 @@ bool ProblemDescription::IsLayoutNHWC() const
     }
 }
 
+bool ProblemDescription::IsLayoutNCHWC() const
```
JehandadKhan, line 47:
```suggestion
bool ProblemDescription::IsLayoutNCHWc() const
```

src/tensor_api.cpp
JehandadKhan: I second @DrizztDoUrden's opinion

test/conv_common.hpp
```diff
@@ -1712,6 +1713,27 @@ struct conv_driver : test_driver
...
+        {
+            auto found_vec = conv_mode.find("fp16x");
+            if(found_vec != std::string::npos)
+            {
+                std::string vec_str = conv_mode.substr(found_vec + 5);
+                ///\todo try catch()
+                vector_length = std::stoi(vec_str);
+                if(vector_length != 4 && vector_length != 8)
+                    MIOPEN_THROW("Unsupported vector length");
+            }
```
JehandadKhan, line 1729: Vector length should be its own argument to the test and not attached to the convolution mode.

test/conv_common.hpp
```diff
@@ -164,11 +164,7 @@ struct scalar_gen_random_integer
 
 struct tensor_elem_gen_one
 {
-    template <class... Ts>
-    double operator()(Ts...) const
-    {
-        return 1;
-    }
+    double operator()() const { return static_cast<double>(1); }
```
JehandadKhan, line 167: Why is the templating missing ?

test/conv_common.hpp
```diff
@@ -1750,11 +1776,36 @@ struct conv_driver : test_driver
...
-                              spatial_dim_elements.at(0),
-                              spatial_dim_elements.at(1)}
-                        .generate(tensor_elem_gen_integer{17});
+            if(in_layout == "NCHW_VECT_C")
+            {
+                batch_size     = input_dims.at(0);
+                input_channels = input_dims.at(1);
+                std::copy(input_dims.begin() + 2, input_dims.end(), spatial_dim_elements.begin());
+                if(vector_length == 4)
+                    input = tensor<T>{type,
```
JehandadKhan, line 1785: Same issue as the conv_driver.

test/conv_common.hpp
```diff
@@ -1953,24 +2098,33 @@ struct conv_driver : test_driver
...
+                 (filter.group_count > 1 &&
+                  (input.desc.GetLengths().at(1) % weights.desc.GetLengths().at(0) == 0)))))
             {
-                auto output = get_output_tensor(filter, input, weights, out_layout);
-
+                auto output             = get_output_tensor(filter, input, weights, out_layout);
                 auto gen_positive_value = [=](auto...) {
                     auto data_type    = input.desc.GetType();
-                    std::size_t v_max = is_int8 ? 16 : (data_type == miopenHalf) ? 4 : 16;
+                    std::size_t v_max = is_int8 ? 16 : data_type == miopenHalf ? 4 : 16;
```
JehandadKhan, line 2119: Why were the parenthesis removed ?

test/conv_common.hpp
```diff
@@ -1953,24 +2098,33 @@ struct conv_driver : test_driver
...
 
                     return gen_float ? scalar_gen_random_float{0, 1}()
                                      : scalar_gen_random_integer{1, v_max}();
                 };
 
                 auto gen_sign_value = [=](auto... is) {
                     auto data_type    = input.desc.GetType();
-                    std::size_t v_max = is_int8 ? 16 : (data_type == miopenHalf) ? 4 : 16;
+                    std::size_t v_max = is_int8 ? 16 : data_type == miopenHalf ? 4 : 16;
```
JehandadKhan, line 2127: same as above

test/conv_common.hpp
```diff
@@ -2037,9 +2190,9 @@ struct conv_driver : test_driver
                 input.generate(gen_positive_value);
                 output.generate(gen_positive_value);
                 weights.generate(gen_sign_value);
+                // weights.generate(gen_positive_value);
```
JehandadKhan, line 2193: Perhaps remove ?

test/cpu_conv.hpp
```diff
@@ -75,58 +74,165 @@ void cpu_convolution_forward_impl(const tensor<Tin>& in,
...
-                    {
-                        std::array<std::size_t, ConvDim + 2> in_id{};
-                        in_id[0] = out_n_id;
-                        in_id[1] = in_c_id;
-                        std::copy_n(in_spatial_id.begin(), ConvDim, in_id.begin() + 2);
-
-                        acc += double(in(in_id)) *
-                               double(wei(out_k_id, wei_c_id, wei_spatial_id_pack...));
-                    }
+    if(wei.desc.GetLayout_t() == miopenTensorCHWNc4 || wei.desc.GetLayout_t() == miopenTensorCHWNc8)
```
JehandadKhan, line 91: Perhaps merge the two branches since one contains the other ?

JehandadKhan (changes requested):

driver/conv_driver.hpp
```diff
@@ -538,7 +573,9 @@ void ConvDriver<Tgpu, Tref>::ValidateLayoutInputParameters(std::string layout_va
     else
     {
         if((layout_value.compare("NCHW") == 0) || (layout_value.compare("NHWC") == 0) ||
-           (layout_value.compare("NCDHW") == 0) || (layout_value.compare("NDHWC") == 0))
+           (layout_value.compare("NCHW_VECT_C") == 0) ||
+           (layout_value.compare("CHWN_VECT_C") == 0) || (layout_value.compare("NCDHW") == 0) ||
```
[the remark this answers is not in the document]
aska-0096: Yes. By adding a input flag called `vector_length `with short name `L `and existed input flag `tensor_vect`, we can fully determine vectorized tensor layout.

junliume:
@aska-0096 Let's ping the reviewers again and really want to see this PR moving forward soon

aska-0096:
> @aska-0096 Let's ping the reviewers again and really want to see this PR moving forward soon

@atamazov @DrizztDoUrden @JehandadKhan Previous review problems have been solved, would you like to give a re-review here?

junliume:
> > @aska-0096 Let's ping the reviewers again and really want to see this PR moving forward soon
> 
> @atamazov @DrizztDoUrden @JehandadKhan Previous review problems have been solved, would you like to give a re-review here?

Ping @atamazov @DrizztDoUrden @JehandadKhan CI tests have passed, please review so that we can include this PR in the next staging.

test/conv_common.hpp
atamazov: +1 to @JehandadKhan

src/conv/problem_description.cpp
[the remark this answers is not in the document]
atamazov: @aska-0096 
> ...we only use CHWNc layout as weight tensor.

Sorry but I do not understand still... There is no "N" in the weights tensor.

IIRC we derive weights layout from input & output layouts, implicitly. For example, for NCHW input and output, weights is of KCYX layout. Ditto NHWC convolutions. We do not require the user to explicitly specify weights layout for both NCHW and NHWC.

Why this is necessary for NCHWc layout?

include/miopen/miopen.h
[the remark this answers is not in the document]
atamazov: Let's mark it as resolved.

atamazov (commented):
The overall impression is that the code calls for generalization. However I think we can merge this (I feel a kind of urgency wave around this work) as is.

src/conv/problem_description.cpp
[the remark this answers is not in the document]
aska-0096: Hi @atamazov.
`CHWNc` means `CYXKc`  in weight tensor.

> IIRC we derive weights layout from input & output layouts, implicitly. For example, for NCHW input and output, weights is of KCYX layout.

Yes, it is, but not necessary. The convolution is all about multiply two tensors and get another one, in forward convolution, we always want input tensor and output tensor have same layout while weight tensor layout is arbitrary, because as you can imagine, tensor layout is just the order of data lied in computer memory space. The reason we introduce `NCHWc x CYXKc = NCHWc` (where weight tensor is different from input and output) is just because it's faster in some cases and we support it in MISA(implicit gemm generator).

> We do not require the user to explicitly specify weights layout for both NCHW and NHWC


Actually we need, but we did it implicitly in based on two MIOpen logics:

1. Default tensor layout is NCHW.
2. Input tensor, weight tensor and output tensor have same tensor layout.

These legacy only native support NCHW, which means, lack of flexibility.
NHWC support was did by carlushuang, and he put much effort on transform NHWC tensor to NCHW tensor.

> So the question is: Why setting the layout of weights is necessary for NCHWc?


It's natural logic in convolution if we want to support different tensor layouts. As I talked above, the biggest reason is in the forward convolution:
**weight tensor could be in arbitrary layout.**

driver/driver.hpp
```diff
@@ -139,10 +139,10 @@ inline void PadBufferSize(size_t& sz, int datatype_sz)
 [[gnu::noreturn]] inline void Usage()
 {
     printf("Usage: ./driver *base_arg* *other_args*\n");
-    printf(
-        "Supported Base Arguments: conv[fp16|int8|bfp16], CBAInfer[fp16], pool[fp16], lrn[fp16], "
-        "activ[fp16], softmax[fp16], bnorm[fp16], rnn[fp16], gemm, ctc, dropout[fp16], "
-        "tensorop[fp16], reduce[fp16,fp64]\n");
+    printf("Supported Base Arguments: conv[fp16|fp16x4|fp16x8|int8|bfp16], CBAInfer[fp16], "
```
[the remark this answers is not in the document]
aska-0096: Added

src/convolution.cpp
aska-0096: Actually, we can not only IsVectorized or not here to determine the following code, since weight tensor is not in default NCHW.

aska-0096:
> The overall impression is that the code calls for generalization. However I think we can merge this (I feel a kind of urgency wave around this work) as is.

Hi @JehandadKhan , What's your opinion?

junliume (approved):
Also ping @DrizztDoUrden  and @JehandadKhan  for review opinions :)

driver/conv_driver.hpp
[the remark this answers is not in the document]
DrizztDoUrden: Sorry, they just look very simmilar at fast glance.

src/include/miopen/tensor_extra.hpp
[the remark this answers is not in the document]
DrizztDoUrden: I mean should the api function be specifically 4d? If you would make Nd function, would it differ in signature in any way? I would rename it to Nd, and return miopenStatusNotImplemented for non 4d tensors.

src/tensor_api.cpp
DrizztDoUrden: I don't think we should spawn 3d/4d/5d function signatures and implementation, when there is additional complexity in making just one. Later each of them would require either support, or to be implemented via an internal Nd function.

DrizztDoUrden:
I think the new API functions should not be specifically 4d, but either return a correct value (for vector length getter), or notImplemented status code for the parameter combinations they are not implemented for. Otherwise at some point we would need to support a lot of additional signature permutations.

src/include/miopen/tensor_extra.hpp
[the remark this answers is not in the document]
aska-0096: Hi @DrizztDoUrden , these function has been changed to Nd already in some commits ago. Sorry for late inform.

src/tensor_api.cpp
aska-0096: Hi @DrizztDoUrden , these function has been changed to Nd already in some commits ago. Sorry for late inform.

aska-0096:
> I think the new API functions should not be specifically 4d, but either return a correct value (for vector length getter), or notImplemented status code for the parameter combinations they are not implemented for. Otherwise at some point we would need to support a lot of additional signature permutations.

Hi @DrizztDoUrden , these function has been changed to Nd already in some commits ago. Sorry for late inform.

DrizztDoUrden (approved):
LGTM

JehandadKhan (approved):

junliume (approved):

junliume:
@aska-0096 Congratulations! It's a lot of work.

include/miopen/miopen.h
```diff
@@ -340,6 +340,24 @@ typedef enum
...
+    miopenTensorNCHW   = 0, /*!< NCHW memory layout (Fully supported) */
+    miopenTensorNHWC   = 1, /*!< NHWC memory layout (Fully supported) */
+    miopenTensorCHWN   = 2, /*!< CHWN memory layout (Not supported) */
+    miopenTensorNCHWc4 = 3, /*!< NCHWc4 memory layout (Partially supported) */
+    miopenTensorNCHWc8 = 4, /*!< NCHWc8 memory layout (Partially supported) */
+    miopenTensorCHWNc4 = 5, /*!< CHWNc4 memory layout (Partially supported) */
+    miopenTensorCHWNc8 = 6, /*!< CHWNc8 memory layout (Partially supported) */
+    miopenTensorNCDHW  = 7, /*!< NCDHW memory layout (Fully supported) */
+    miopenTensorNDHWC  = 8, /*!< NCDHW memory layout (Fully supported) */
+} miopenTensorLayout_t;
```
jerryyin, line 359: @pfultz2 Request for comments.

src/include/miopen/tensor.hpp
```diff
@@ -213,8 +247,10 @@ struct TensorDescriptor : miopenTensorDescriptor
     std::vector<std::size_t> strides;
 
     bool packed;
+    int vector_length = 1;
 
-    miopenDataType_t type = miopenFloat;
+    miopenDataType_t type             = miopenFloat;
+    miopenTensorLayout_t tensorLayout = miopenTensorNCHW;
```
jerryyin, line 253: I disagree with this, on a number two aspects:
 - tensor layout shouldn't be of enum type, (rather it should be dynamically calculated and cached as string). Using a enum type will make it much harder to scale when we are to add a different layout. This has been concluded in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343
 - tensor layout shouldn't be cached in `tensor` type, but in `problem_description`

With it, I prefer the changes in here/constructor to be reverted such that a client of tensor descriptor doesn't need to use a layout to create a tensor descriptor. If I remember it right, @pfultz2 was against this opinion when I was trying to do (almost) the same change in "ambiguous layout" fix in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/767.

include/miopen/miopen.h
```diff
@@ -580,6 +598,22 @@ MIOPEN_EXPORT miopenStatus_t miopenCreateTensorDescriptor(miopenTensorDescriptor
 MIOPEN_EXPORT miopenStatus_t miopenSet4dTensorDescriptor(
     miopenTensorDescriptor_t tensorDesc, miopenDataType_t dataType, int n, int c, int h, int w);
 
+/*! @brief Set shape of ND tensor with specific layout
+ *
+ * Interface for setting N-D tensor shape. This interface support NHWC, NCHW, NCHWc*, CHWNc*
+ * @param tensorDesc   Tensor descriptor type (output)
+ * @param dataType     MIOpen datatype (input)
+ * @param tensorLayout Tensor layout (input)
```
jerryyin, line 606: @pfultz2 Request for comment: I remember you are against forcing client giving the layout from API.

jerryyin (commented):
Sorry for giving after-approved reviews, please don't take it personal. I want to point out a few aspect that I'm not sure we are on the same page.

include/miopen/miopen.h
junliume: @aska-0096 @carlushuang  for awareness

src/include/miopen/tensor.hpp
junliume: @aska-0096 @carlushuang  for awareness

include/miopen/miopen.h
junliume: @aska-0096 @carlushuang  for awareness

include/miopen/miopen.h
aska-0096: Got it. Maybe this problem should be discussed between paul and framework team, because this design is determined after framework team's review.

src/conv/problem_description.cpp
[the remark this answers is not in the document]
atamazov: @aska-0096 Thanks, agreed.

src/include/miopen/tensor.hpp
atamazov: @jerryyin 
> tensor layout shouldn't be cached in tensor type, but in problem_description

IIRC the modern convolution's problem description holds tensor descriptors, which, in turn, hold layouts. Is it acceptable?

include/miopen/miopen.h
atamazov: @jerryyin @pfultz2 @aska-0096 @carlushuang @junliume Yes, IIRC we generally want to set strides via an API and let MIOpen do the rest of the work (derive layouts from them). So I'm also wondering why this approach isn't used here (due to complexity?)

src/include/miopen/tensor.hpp
jerryyin: > convolution's problem description holds tensor descriptors, which, in turn, hold layouts

Yes, that sounds like the consistent approach with the infrastructure already exists before this PR.

pfultz2: Yea this is wrong. The layout is a property of the strides. If a user passes the layout with strides instead of an enum this will be broken.

include/miopen/miopen.h
```diff
@@ -580,6 +598,22 @@ MIOPEN_EXPORT miopenStatus_t miopenCreateTensorDescriptor(miopenTensorDescriptor
...
+ * @param lens         Tensor dimensions (input)
+ * @param num_lens     Tensor dimension size (input)
+ * @return             miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t
+miopenSetNdTensorDescriptorWithLayout(miopenTensorDescriptor_t tensorDesc,
+                                      miopenDataType_t dataType,
+                                      miopenTensorLayout_t tensorLayout,
+                                      int* lens,
+                                      int num_lens);
```
pfultz2, line 616: This API wont work for NCHWc. How does the user specify NCHWc when the number of channels is not divisible by c? For example, if the user has N5HW and convert to N2HW4, but there needs to be padding for c so it will only use 5 channels not 8(which would cause buffer overrun).

include/miopen/miopen.h
pfultz2: Well an enum is fine for an easy API fro some users as along as users can still set the strides as well to achieve the layout. The only way to do this correctly is not store a layout string or enum in the tensor descriptor.

aska-0096:
Hi @jerryyin @atamazov @pfultz2 , thank you for all of your discussions above, let me give a summary of this ticket.

### Summary
**Origin tensorDescriptor**
`tensorDescriptor ` hold the member `lens `and `strides`, length of tensor's each dimension is always stored in NCHW order in the `lens `, while the `strides ` has the information of each dimension's stride.
It has some drawbacks as:
- Derived tensor layout from `strides `and `lens `will be ambiguous in some cases.
- With the scope of supported tensor layout increasing, there must be many tensor layout checking, it's a kind of expensive to calculate/derive tensor layout more than once.

**Tensor layout added tensorDescriptor**
Add a enum member to help determine the unique tensor layout, it's external user-appointed and internal-checked, reorder lens and strides in the order of tensor layout.
`NHWC example: {N, C, H, W} = {16, 32, 5, 5},  lens = {16, 5, 5, 32}, strides = {800, 160, 32, 1}
`
It's truly flexible and without ambiguous problem, but

- Cumbersome, since we already have the tensor layout information in `strides`.
- More workload to add new tensor layout support.

### An acceptable way

- External user-appointed tensor layout information directly store into `ConvDescriptor` which based on `problemDescriptor`, avoid ambiguous problem.
- Keep the origin concise  `tensorDescriptor `design, without `tensorlayout ` enum.

aska-0096:
> Hi @jerryyin @atamazov @pfultz2 , thank you for all of your discussions above, let me give a summary of this ticket.
> 
> ### Summary
> **Origin tensorDescriptor** `tensorDescriptor ` hold the member `lens `and `strides`, length of tensor's each dimension is always stored in NCHW order in the `lens `, while the `strides ` has the information of each dimension's stride. It has some drawbacks as:
> 
> * Derived tensor layout from `strides `and `lens `will be ambiguous in some cases.
> * With the scope of supported tensor layout increasing, there must be many tensor layout checking, it's a kind of expensive to calculate/derive tensor layout more than once.
> 
> **Tensor layout added tensorDescriptor** Add a enum member to help determine the unique tensor layout, it's external user-appointed and internal-checked, reorder lens and strides in the order of tensor layout. `NHWC example: {N, C, H, W} = {16, 32, 5, 5}, lens = {16, 5, 5, 32}, strides = {800, 160, 32, 1} ` It's truly flexible and without ambiguous problem, but
> 
> * Cumbersome, since we already have the tensor layout information in `strides`.
> * More workload to add new tensor layout support.
> 
> ### An acceptable way
> * External user-appointed tensor layout information directly store into `ConvDescriptor` which based on `problemDescriptor`, avoid ambiguous problem.
> * Keep the origin concise  `tensorDescriptor `design, without `tensorlayout ` enum.

Hi @junliume.
If we decide the final design, an acceptable way to everyone, I will create a new PR to implement it.

pfultz2:
> Derived tensor layout from strides and lens will be ambiguous in some cases.

The #767 discusses the original design decisions we made. I think there is only an ambiguity for extreme cases like n, 1, 1, 1, which I dont think is ever used.

>  it's a kind of expensive to calculate/derive tensor layout more than once.

Sure we could store the permutation on construction, but this should be stored as `std::vector<uint8_t>` or `std::string`(SSO might be more efficient here) using integers such as `{0, 2, 3, 1}` or `"0231"`. We shouldn't use enums as there is too many combinations, and storing labels(such as `"NCHW"`) is dependent on where the tensor is used so shouldn't be used either.

> reorder lens and strides in the order of tensor layout.
> NHWC example: {N, C, H, W} = {16, 32, 5, 5}, lens = {16, 5, 5, 32}, strides = {800, 160, 32, 1}

Order of lens is always NCHW regardless of memory layout. The tensor should be `lens = {16, 32, 5, 5}, strides = {800, 32, 1, 160}`.

> It's truly flexible and without ambiguous problem

Well you traded one ambiguity for another ambiguity here though. 

> External user-appointed tensor layout information directly store into ConvDescriptor which based on problemDescriptor, avoid ambiguous problem.

We should not store the layout in the ConvDescriptor as this is independent of tensor inputs. Users could reuse the same ConvDescriptor for tensors with NCHW and NHWC if they wanted to.

We can store a layout in the tensor descriptor during construction, but we need to make sure that:

1) The order of lens is always NCHW regardless of memory layout
2) The tensor permutation should be stored as integers not as an enum
3) The strides should match the layout

pfultz2:
> Hi @junliume.
> If we decide the final design, an acceptable way to everyone, I will create a new PR to implement it.

Most of this discussion is about the internal represention, which we have discussed in the past(see [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/343#discussion_r473132394) and [here](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/767#issuecomment-802262860)). Either way, these changes can wait another discussion.

However, the C API changes are more urgent and needs to document how a user would specify a `N5HW` tensor with `NCHWc4` layout. If that is not possible or clear how it can be done then we should revert the C API changes before releasing otherwise we will be stuck supporting a broken API, and we will need to add `miopenSetNdTensorDescriptorWithLayout2` API function to correct it.

jerryyin:
@pfultz2 Thanks for giving thorough feedbacks, and +1 on your response.

@junliume @aska-0096 I think we want to mark this discussion as a blocking issue for ROCm 5.3 release. (which is less than 2 weeks till dev complete. Unless @aska-0096 plan to get everything implemented very soon) Agreed with @pfultz2 to revert C API changes to avoid this get released without consensus. Then we can shoot for another PR to fix/expose it in next release.

pfultz2:
> IIRC this is not supported, i.e. C must be evenly divisible by 4. Do we really need to support such cases (and on what grounds)?

We dont need to support it now in the backend, but our API should be able to support this so when we enable this in the future we dont need to add `miopenSetNdTensorDescriptorWithLayout2` to enable it.

We definitely should support this in the future, because:

* Channels are not always divisible by powers of 2
* Competitors(such as onednn) already support [this](https://oneapi-src.github.io/oneDNN/dev_guide_understanding_memory_formats.html#what-if-channels-are-not-multiples-of-8-or-16)
* It should be simple to support this in CK by adding padding to channels, as this is just another affine transform to the tensor

aska-0096:
> We definitely should support this in the future, because:
> 
> * Channels are not always divisible by powers of 2
> * Competitors(such as onednn) already support [this](https://oneapi-src.github.io/oneDNN/dev_guide_understanding_memory_formats.html#what-if-channels-are-not-multiples-of-8-or-16)
> * It should be simple to support this in CK by adding padding to channels, as this is just another affine transform to the tensor

Hi @pfultz2 , Thanks for explanation why we need to deal the non-divisible case.

> However, the C API changes are more urgent and needs to document how a user would specify a N5HW tensor with NCHWc4 layout. If that is not possible or clear how it can be done then we should revert the C API changes before releasing otherwise we will be stuck supporting a broken API, and we will need to add miopenSetNdTensorDescriptorWithLayout2 API function to correct it.

> Agreed with @pfultz2 to revert C API changes to avoid this get released without consensus. Then we can shoot for another PR to fix/expose it in next release.

Let me explain how this API work in the backend.
When we call the `miopenSetNdTensorDescriptorWithLayout` API
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/edae4e69f49059c7488fbe5d9fad087f07dae4df/src/tensor_api.cpp#L51-L63
It will call the tensor construct that utilize the tensorLayout_t, which store the memory layout of that tensor. In current design, we coupled the layout and vectorized length( or `blocking` in oneDNN style), for example, enum miopenNCHWc8 means the tensor layout is NCHWc and vector length is 8. ( I know you disagree with using of enum in tensor constructor, but please forgive me I use it for illustration here).
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/edae4e69f49059c7488fbe5d9fad087f07dae4df/src/tensor.cpp#L94-L104
In side the tensor constructor, apart from the direct copy arguments, we have two internal functions using to derive the vector length and strides:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/edae4e69f49059c7488fbe5d9fad087f07dae4df/src/tensor.cpp#L147-L154

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/edae4e69f49059c7488fbe5d9fad087f07dae4df/src/tensor.cpp#L125-L145

In the former one, we parse the vector length directly which will be utilized to modify the user-passing tensor dimension length and calculate the strides in the latter function `CalculateStrides()`.
For example, user input:  `... -I NCHW -n 2 -c 16 -h 5 -w 5 -z 1 -L 4...` will generate a tensor with lens of {2, 4, 5, 5} and strides of { 400, 100, 20, 4}, in layout `miopenNCHWc4`.
In my point of view, this API **ALLOW** user input N5HW like channel-non-divisible tensor, when we need to enable this kind of use case in the backend, we can add a internal function like `zero_padding()` in the tensor constructor.

### Summary

I understand why enum type tensor layout is not welcome after looking at the oneDNN's hundreds of `format_tag_t`, agree with @pfultz2 's opinion that we should make it easy to maintain and treat it as one of the core design of MIOpen tensor description.

However, I **DISAGREE** that we should revert the `miopenSetNdTensorDescriptorWithLayout` API, since it has the ability to fulfill the future expansion.

cc. @junliume . I think we need your voice.

atamazov:
@pfultz2 It seems that actually `miopenSetNdTensorDescriptorWithLayout` API call supports vectorized layouts that require padding.

And yes, internal representation it definitely something that can be improved (but we can do that later).

@junliume Can you please add the `request_for_comments` label to this PR and to #767, thanks!

jerryyin:
No offense but I still lean towards reverting based on that fundamentally I'm against:
 1. enum to represent layout, this will seriously limit MIOpen's capability to try kernel in non-defined layout
     - Reasoning here is that in MLIR providing a kernel of different layout is almost trivial. The only reason we don't support other layouts are mainly due to testing capacity rather than difficulty.
     - With a enum defined, MIOpen will be scoped on only a couple of defined layouts, and it will be hard even to experiment others
     - With the enum data structure, the extend-ability of different layout becomes increasingly difficult
       - For NCHW, there's total of permutation(4) = 24 possibilities; For NCDHW, there's total of permutation(5) = 120 possibilities; Should we deicde to give other layouts a shot, it is not realistic to enumerate all permutations.
 2. Layout is coupled with vectorization. They are two independent attributes and in my opinion shouldn't be coupled together. This will again be a limitation on non-defined vectorization value.
     - This exacerbate the limitation of enums. Shall we amend or extend the `miopenTensorLayout_t` everytime we bring up a different vectorization size?
     - Along with this: vectorization size should be a first-class citizen and numerical value, not part of information carried with enum. I dislike the approach taken with `CalculateVectorLength()`, and think that this is another example we pay overhead because of enum.

In my opinion, the minimal changes needs to happen is:
 - Decouple between layout and data type, (I believe they are decoupled in cudnn too) for example:
   - NCHW is a layout
   - int8x4 is a data type
 - Avoid enum type for either layout or data type
   - We can reject non-supported values, instead of enumerating all supported cases 
   - Internally layout get derived and cached in problem description, so there's very limited overhead

carlushuang:
@JehandadKhan @junliume please be noticed this API design related discussion. Any comments is welcome. If needed we can host a offline meeting to speed up communication.

jerryyin:
Personally I find onednn's way of defining everything coupled together as in [here](https://oneapi-src.github.io/oneDNN/enum_dnnl_format_tag_t.html) is absurd and an example of bad data structure design. If we continue this approach, we'd ended up a enum of their definition and spend effort dealing with those hard-coded values.

To reiterate my point, I'm going to give some examples of tensorflow below on why this new API may cause confusion too:

### Example 1

This echoes @pfultz2's [concern ](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532#issuecomment-1157844955) below: 

For none vectorized layout, its support is already in existing `miopenSetTensorDescriptor()` API and has been adopted by TensorFlow. Now this new `miopenSetNdTensorDescriptorWithLayout()` will just create more confusion because MIOpen API would leave more than 1 way to do it:
 - Refer to https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/pull/1256 on how tensorflow use just strides and lens to represent the layout
 - Now in this new API, the functionality overlap in terms of we provide this to do a similar thing. Except that this is performed by `miopenSetNdTensorDescriptorWithLayout()`, but also needs a redundant enum `miopenTensorNCHW` or `miopenTensorNHWC`

Concerns are:
 - Which API should they use? 
 - Why this new API requires a enum whereas previous one don't?
 - What if they defined the lens/strides sequence that are incoherent with layout enum provided? Do you have the error detection logic to handle the case?

### Example 2

This is to show the inefficient framework layout format that still exist today [here](https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/blob/84fa9ec65dec173bb1e88f05bd7d9c8ed317d1bd/tensorflow/core/kernels/conv_ops.cc#L1049-L1067). 
 - User creates a filter format in HWIO (or `HWCN`) format
 - If data layout is NCHW, TF conv op implementation will transform filter layout to OIHW (or, NCHW)
 - If data layout is NHWC, TF conv op implementation will transform filter layout to OHWI (or, NHWC)

Now, this is an example if the framework team decide to optimize the implementation, MIOpen has to support this new `HWCN` format by adding a enum definition. On the other hand, it'd be much easier if MIOpen don't define a whitelist enum but write a solver to support the format.

pfultz2:
> Decouple between layout and data type, (I believe they are decoupled in cudnn too) for example:

I dont think this can handle padding. Also using the data type its not clear which parameter is vectorized, and can change depending on the algorithm used(gemm vs convolution), so its even more confusing. OneDNN makes this clear by having a blocked dims parameter to set which axis is being vectorized in the tensor descriptor.

> Now this new miopenSetNdTensorDescriptorWithLayout() will just create more confusion because MIOpen API would leave more than 1 way to do it

I dont think its a problem to have an enum to set the layout in our C API as long as internally we represent the tensor the same way. Some simple users may want to set the tensor dimension without having to deal with strides so having an enum makes that easy. However, the enum does not have to be comprehensive, for more advance cases the strides can be used. It also makes sense that frameworks and migraphx will use strides instead of an enum since it already has strides.

Also for now, we will need to use an enum for NCHWc layout since neither affine transforms nor blocked dims have been implemented for MIOpen's tensor descriptor, and we want to proceed cautiously here. If we implement the blocked dims API we need to make sure it wont prevent us from using affine transforms in the future.

jerryyin:
> I dont think this can handle padding. Also using the data type its not clear which parameter is vectorized

I was thinking that another vector can represent where the source and vectorized dimensions are like `NCHW8c` represented as `[0, C, 0, 0, 8]`. From here we can derive the padding too. Do you see a counter example where this won't work?

> I dont think its a problem to have an enum to set the layout in our C API as long as internally we represent the tensor the same way. 

Taking one step back I still think:
 - Representation of non-vectorized layout like `NCHW` and `NHWC` should not be coupled with vectorized layout, it should stay to be represented by strides and dims
 - If we'd like clarity and compatibility with other places we can scope the use of enum to only vectorized layout.

pfultz2:
> Personally I find onednn's way of defining everything coupled together as in [here](https://oneapi-src.github.io/oneDNN/enum_dnnl_format_tag_t.html) is absurd and an example of bad data structure design.

To be clear, onednn uses the strides(and blocked dims) to represent the layout. The enum is used for a setting the strides automatically from the dimensions.

Also you get a format tag from the tensor(usually when logging is enabled), which why it tries to list out all possibilities. However, we in MIOpen dont enum to represent all layout. I think just showing the permutation as integers(ie `{0, 3, 1, 2}`) is good enough. As such we dont need enumerate all possible layouts(whereas OneDNN does).

pfultz2:
> I was thinking that another vector can represent where the source and vectorized dimensions are like NCHW8c represented as [0, C(padded), 0, 0, 8] (alternatively, use a pair to denote the source and destination of the vectorization dimension). From here we can derive the padding too (by comparing with dims). Do you see a counter example where this won't work?

I dont really follow this. How would the user create `N5HW` with `NCHWc4`?

Also how does the user control the dimension that is vectorized?  For convolution its the C dimension, but for gemms it would be the columns. 

> Representation of non-vectorized layout like NCHW and NHWC should not be coupled with vectorized layout, it should stay to be represented by strides and dims

Yes I agree, especially since this how MIGraphX will tell MIOpen the layout. However, this is somewhat unrelated to the  `miopenSetNdTensorDescriptorWithLayout` API, and more related to the implementation of `miopenSetNdTensorDescriptorWithLayout` which should properly set strides and not an enum. Of course, its not cleary documented what `miopenSetNdTensorDescriptorWithLayout` does, so this should *NOT* go into a release until the semantics is defined.

> If we'd like clarity and compatibility with other places we can scope the use of enum to only vectorized layout.

I dont see a problem having some common layouts in the enum like NHWC, but there is no reason for CHWN to be listed though.

jerryyin:
> I dont really follow this. How would the user create N5HW with NCHWc4?

Here are my thoughts:
 - If a user wants to create N5HW with NCHWc4, then
   - dims = [N, 5, H, W]; vec = [0, 2, 0, 0, 4]; There's 1:1 mapping from vec to dims, and a non-zero value represent the source vectorized dimension. The last dimension is the destination vectorized dimension.
   - With it, first step we can derive the vectorization dimension is C because only C dimension is set to 2
   - Then, since `5 < 2 x 4`, it means that only 5 out of 8 channels are used
 - If a user wants to create regular NCHWc4, then
   - dims = [N, 8, H, W]; vec = [0, 2, 0, 0, 4]; 

> Of course, its not cleary documented what miopenSetNdTensorDescriptorWithLayout does, so this should NOT go into a release until the semantics is defined.

Agreed

> I dont see a problem having some common layouts in the enum like NHWC, but there is no reason for CHWN to be listed though.

Also agreed especially since it is unsupported. I think @aska-0096, with the existing implementation would argue that there has to be `CHWN` to be able to represent `CHWNc`. This shows the downside of the existing implementation of coupling vectorized and non-vectorized format together.

src/include/miopen/conv/problem_description.hpp
```diff
@@ -223,8 +253,20 @@ struct ProblemDescription
...
     miopenDataType_t GetWeightsDataType() const { return weights.GetType(); }
     std::size_t GetWeightsDepth() const { return GetD5(GetSpatialDims(), weights.GetLengths()); }
-    std::size_t GetWeightsHeight() const { return GetH5(GetSpatialDims(), weights.GetLengths()); }
-    std::size_t GetWeightsWidth() const { return GetW5(GetSpatialDims(), weights.GetLengths()); }
+    std::size_t GetWeightsHeight() const
+    {
+        if(weights.GetLayout_str() == "CHWNc")
+            return GetHofCHWN(weights.GetLengths());
+        else
+            return GetH5(GetSpatialDims(), weights.GetLengths());
```
atamazov, line 261: @jerryyin Would this work correctly with NHWC layout? I am afraid that some change here is necessary but missed in #343.

jerryyin: This seems to be correct? Back in #343 We only supports NCHW or NCDHW alternatives. This additional branch to handle `CHWNc` seems to be what it should do in this PR.

atamazov: @jerryyin You're right. NHWC support was introduced in in #728. So does #728 missing changes in src/include/miopen/conv/problem_description.hpp, what do you think?

/cc @carlushuang

carlushuang: #728 is far before this PR at that time there is no `GetLayout_str()` API. I'm not sure what is missing you refer to?

atamazov: @carlushuang My original question is about `GetWeightsHeight()` and similar functions. These were introduced 3 yrs ago for NCHW, and I would expect that some modification is necessary to support NHWC properly (in #728).

AFAICS now, the library requires (and supports) only the following order of elements in the lengths vector: `{ N,C,H,W }` (or `{ N,C,D,H,W }`), **_regardless of the actual layout_**. If so, then `GetWeightsHeight()` and similar functions do indeed support NHWC (no modifications to these functions were required in #728 to support NHWC).

Strictly speaking, this is off-topic in this PR. I can move the discussion there, if you wish.

src/ocl/convolutionocl.cpp
```diff
@@ -111,16 +111,36 @@ static inline void ValidateGroupCount(const TensorDescriptor& xDesc,
...
     if(conv.group_count == 1)
     {
-        if(xDesc.GetLengths()[1] != wDesc.GetLengths()[1])
+        if((((wDesc.GetLayout_t() == miopenTensorNCHW) ||
+             (wDesc.GetLayout_t() == miopenTensorNCHWc4) ||
+             (wDesc.GetLayout_t() == miopenTensorNCHWc8)) &&
+            (xDesc.GetLengths()[1] != wDesc.GetLengths()[1])) ||
+           ((wDesc.GetLayout_t() == miopenTensorCHWNc4 ||
+             wDesc.GetLayout_t() == miopenTensorCHWNc8) &&
+            (xDesc.GetLengths()[1] != wDesc.GetLengths()[0])))
```
atamazov, line 120: This breaks transposed convolution support. Example:
```bash
MIOPEN_LOG_LEVEL=6 ./bin/MIOpenDriver conv -x 20 -y 5 -W 700 -H 161 -c 1 -n 4 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -m trans -s 0 -S 0 -w 2 -t 1 -i 1 -F 1 -w 1
MIOpenDriver conv -x 20 -y 5 -W 700 -H 161 -c 1 -n 4 -k 32 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -g 1 -F 0 -V 1 -m trans -s 0 -S 0 -w 2 -t 1 -i 1 -F 1 -w 1
MIOpen(HIP): Info [Handle] stream: 0x1d29a70, device_id: 0
MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, HIP version 5.4.22505, MIOpen version 3.00.0.331fc1012
MIOpen(HIP): Info [GetSolutionCount]
MIOpen(HIP): Info2 [GetInstalledPathFile] Found exact find database file: /root/github/MLOpen1/src/kernels/gfx1030_36.HIP.fdb.txt
MIOpen(HIP): Info [Measure] ReadonlyRamDb::Prefetch time: 56.7145 ms
MIOpen(HIP): Info [Measure] RamDb::Prefetch time: 0.012484 ms
MIOpen(HIP): Info2 [ValidateUnsafe] DB file is older than cache: 271540521524873, 271550751086576
MIOpen(HIP): Info2 [FindRecordUnsafe] Looking for key 32-325-1418-5x20-1-161-700-4-0x0-2x2-1x1-0-NCHW-FP32-B in cache for file /root/github/MLOpen1/src/kernels/gfx1030_36.HIP.3_00_0.ufdb.txt
MIOpen(HIP): Info2 [FindRecord] Looking for key 32-325-1418-5x20-1-161-700-4-0x0-2x2-1x1-0-NCHW-FP32-B in file /root/github/MLOpen1/src/kernels/gfx1030_36.HIP.fdb.txt
MIOpen(HIP): Info2 [Measure] Db::FindRecord time: 0.045436 ms
MIOpen Error: /root/github/MLOpen1/src/ocl/convolutionocl.cpp:82: Invalid filter channel number
RunForwardGPU() FAILED, rc = 0x3
Forward Convolution FAILED: 1.79769e+308 > 1.5e-06
```

src/include/miopen/tensor_layout.hpp
```diff
@@ -69,6 +69,42 @@ void tensor_layout_to_strides(const std::vector<T>& len,
...
+        [&base_layout, &vector, &dim_to_len](char cur_layout_char) {
+            auto pos = base_layout.find(cur_layout_char);
+            if(pos == std::string::npos)
+            {
+                MIOPEN_THROW(std::string("mismatched layout string - ").append(base_layout));
+            }
+            return std::accumulate(
+                base_layout.begin() + pos + 1,
+                base_layout.end(),
+                vector,
```
atamazov, line 103: 🐛 Same as https://github.com/ROCm/MIOpen/pull/765/files#r1534223104.

PR #1539, opened 2022-05-09
[MLIR] Adding int8 support and test to ConvMlirIgemmFwdXdlops solver

jerryyin:
This PR enables `ConvMlirIgemmFwdXdlops` solver for int8 convolutions on following types:
 - int8 x int8 x int32
 - int8 x int8 x fp32

This PR made the below changes:
 - Bumped MLIR commit id to the latest, which has support for int8 convolution
 - Added `IsInt8()` call in problem description, making int8 have parallel support to other types
   - This help fix the "Unsupported data types configuration" warnings from `mlo_dir_conv.cpp`
 - Relaxed the restriction of non-gemm solver for in8 type in `convolutionocl.cpp`
 - Added a `CastTensor` kernel upon MLIR int8 invoker such that it allows non-int32 output type
 - Amended `conv2d_test`, such that all fwd convolutions are int8 enabled
 - Got rid of the `conv2d_test` requirement that a solver supporting int8 must also support int8x4
   - TODO: Add an independent int8x4 test option
   - Note: This does not decrease test coverage because all current int8 related testing does not test its functional validness. MLIR solver's int8 test is the first group of int8 testing with `INT8_ENABLED` flag. 

CC: @atamazov

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ RadeonOpenCompute/rocm-cmake@839cf61c355f997283e3baaf245fa9acf2e3baa0 --build
 sqlite3@3.17 -DCMAKE_POSITION_INDEPENDENT_CODE=On
 boost@1.72 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build
 half,https://github.com/pfultz2/half/archive/1.12.0.tar.gz -X header -H sha256:0a08660b68abb176ebc2a0cdf8de46e3182a7f46c66443bb80dbfaaec98cf969 --build
-ROCmSoftwarePlatform/llvm-project-mlir@release/rocm-5.2 -DBUILD_FAT_LIBMLIRMIOPEN=1
+ROCmSoftwarePlatform/llvm-project-mlir@59905bc4e56b4d54dd416f00ecf08d1923bfb4e1 -DBUILD_FAT_LIBMLIRMIOPEN=1
--- a/src/include/miopen/problem_description.hpp
+++ b/src/include/miopen/problem_description.hpp
@@ -200,6 +200,7 @@ struct ProblemDescription
         return in_data_type == miopenBFloat16 && weights_data_type == miopenBFloat16 &&
                out_data_type == miopenBFloat16;
     }
+    bool IsInt8() const { return in_data_type == miopenInt8 && weights_data_type == miopenInt8; }
 
     ProblemDescription() = default;
 
--- a/src/include/miopen/problem_description.hpp
+++ b/src/include/miopen/problem_description.hpp
@@ -200,7 +200,7 @@ struct ProblemDescription
         return in_data_type == miopenBFloat16 && weights_data_type == miopenBFloat16 &&
                out_data_type == miopenBFloat16;
     }
-    bool IsInt8() const { return in_data_type == miopenInt8 && weights_data_type == miopenInt8; }
+    bool IsInt8() const { return conv_problem.IsInt8(); }
 
     ProblemDescription() = default;
 
--- a/src/ocl/convolutionocl.cpp
+++ b/src/ocl/convolutionocl.cpp
@@ -606,8 +606,7 @@ void ConvolutionDescriptor::ConvolutionForward(Handle& handle,
     ValidateConvTensors(tensors);
     ValidateAlphaBeta(alpha, beta);
 
-    if(algo != miopenConvolutionFwdAlgoGEMM &&
-       (xDesc.GetType() == miopenInt8 || xDesc.GetType() == miopenInt8x4))
+    if(algo != miopenConvolutionFwdAlgoGEMM && xDesc.GetType() == miopenInt8x4)
     {
         MIOPEN_THROW(miopenStatusBadParm);
     }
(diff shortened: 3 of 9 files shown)


carlushuang:
Thanks @jerryyin 

Regarding to vector type interpretation inside MIOpen, there was a PR https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1532 working on refactoring the vector type. I'll add @aska-0096 to review list in case some changes are needed here.

src/ocl/convolutionocl.cpp
```diff
@@ -606,8 +606,7 @@ void ConvolutionDescriptor::ConvolutionForward(Handle& handle,
     ValidateConvTensors(tensors);
     ValidateAlphaBeta(alpha, beta);
 
-    if(algo != miopenConvolutionFwdAlgoGEMM &&
-       (xDesc.GetType() == miopenInt8 || xDesc.GetType() == miopenInt8x4))
+    if(algo != miopenConvolutionFwdAlgoGEMM && xDesc.GetType() == miopenInt8x4)
```
aska-0096, line 609: Hi @jerryyin, as discussed in PR #1532 , we decide to decouple the data type and vector length, which means `miopenInt8x4`  may be discarded in the near future. Thus change the return value to `xDesc.GetType() == miopenInt8` here will be appreciate.

jerryyin: That sounds like an orthogonal change with the diff here. This diff touched only on the applicable range of `miopenInt8` and preserved the original logic for `miopenInt8x4`. 

> which means miopenInt8x4 may be discarded in the near future

There should be another future PR that get rid of `miopenInt8x4` rather than me doing patch work in this MLIR pull request. By the time #1532 landed, we should not touch any code preemptively in order not to break any existing functionalities.

jerryyin:
@carlushuang This PR added `MLIR + int8` capability (with a focus on MLIR backend) and preserved all pre-existing `int8x4` functionalities. In case anything needs to be done after #1532, it should be done in a separate PR, after #1532 merged.

src/include/miopen/problem_description.hpp
```diff
@@ -200,6 +200,7 @@ struct ProblemDescription
         return in_data_type == miopenBFloat16 && weights_data_type == miopenBFloat16 &&
                out_data_type == miopenBFloat16;
     }
+    bool IsInt8() const { return in_data_type == miopenInt8 && weights_data_type == miopenInt8; }
```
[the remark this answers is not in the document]
jerryyin: Good idea, done in 7d86ac4adcf8919a57a3ebd0ae0842895b2bda24

jerryyin:
CC: @JehandadKhan Not sure if you have noticed but looks like some permission related issue is coming up for ccache (refer to [here](http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/add-mlir-int8/8/pipeline)):

> {"message":"Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization.","documentation_url":"https://docs.github.com/articles/authenticating-to-a-github-organization-with-saml-single-sign-on/"}

junliume:
> CC: @JehandadKhan Not sure if you have noticed but looks like some permission related issue is coming up for ccache (refer to [here](http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/add-mlir-int8/8/pipeline)):
> 
> > {"message":"Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization.","documentation_url":"[https://docs.github.com/articles/authenticating-to-a-github-organization-with-saml-single-sign-on/"}](https://docs.github.com/articles/authenticating-to-a-github-organization-with-saml-single-sign-on/%22%7D)

CI pipeline restarted ...

junliume (approved):

aska-0096 (approved):
LGTM

requirements.txt
```diff
@@ -2,4 +2,4 @@ RadeonOpenCompute/rocm-cmake@839cf61c355f997283e3baaf245fa9acf2e3baa0 --build
 sqlite3@3.17 -DCMAKE_POSITION_INDEPENDENT_CODE=On
 boost@1.72 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build
 half,https://github.com/pfultz2/half/archive/1.12.0.tar.gz -X header -H sha256:0a08660b68abb176ebc2a0cdf8de46e3182a7f46c66443bb80dbfaaec98cf969 --build
-ROCmSoftwarePlatform/llvm-project-mlir@release/rocm-5.2 -DBUILD_FAT_LIBMLIRMIOPEN=1
+ROCmSoftwarePlatform/llvm-project-mlir@59905bc4e56b4d54dd416f00ecf08d1923bfb4e1 -DBUILD_FAT_LIBMLIRMIOPEN=1
```
atamazov, line 5: Is it so that this breaks backward compatibility with MLIR from ROCm 5.2?

IIRC MLIR backend is not an experimental feature anymore. Maintenance of backward compatibility with a reasonably selected range of previous versions is a standard requirement for MIOpen.

We can use MLIR version information from at CMake level (and forward it into C++ sources as macros, like we already do for COMGR, for example) to maintain backward compatibility at full, including tests.

atamazov (commented):
@jerryyin What is your opinion?

/cc @junliume

requirements.txt
jerryyin: > Can you please remind me, is MLIR Backend an experimental feature yet?

Now with MIOpen release MLIR by default, it shouldn't be an experimental feature.

> If not, then we have a problem.

I think you are right, It indeed never occurred to me to maintain backward compatibility before.

> We can use MLIR version information from at CMake level

Thanks for the pointer. Let's move the discussion into the ticket you opened. I'll likely need a few more pointers and do whatever necessary.

PR #1470, opened 2022-03-22
[Mainline][SWDEV-327397][HotFix] disable hipRTC by default

junliume:
disable hipRTC by default in mainline

PR #1545, opened 2022-05-14
Enable hipRTC as default

junliume:
Issue with SRIOV should have been resolved and root cause has identifies to be not MIOpen.
We should re-enable hipRTC as default

The original PR which disabled hipRTC is https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1470

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,7 +228,7 @@ math(EXPR MIOPEN_hip_VERSION_FLAT "(${MIOPEN_hip_VERSION_MAJOR} * 1000 + ${MIOPE
 
 # Do not enable HIPRTC by default for older ROCm versions in order to avoid
 # build time errors, because HIPRTC is a relatively new component.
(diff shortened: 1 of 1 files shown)


junliume:
@atamazov add as a reviewer if you have time. Any suggestions/recommendations are highly appreciated.

averinevg:
I think it is better to add link to the reverted PR (https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1470) to the description for clarity.

averinevg (commented):
I think it is better to add link to the reverted PR (https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1470) to the description for clarity.

averinevg (approved):

atamazov (approved):
[Gentle hint] In case of more complex reverts, I recommend finding the necessary commit in develop and reverting it using git. This is very convenient for large reverts. The bonus is that git would automatically generate a good draft of commit message that contains the `#` of PR being reverted ))

@averinevg Special thanks for very important notice!

PR #1546, opened 2022-05-16
Fin Perf Eval handles

cderb:
#1445 
With merge resolutions.

PR #1548, opened 2022-05-18
Fin Perf Eval handles (reboot)

cderb:
rehash of #1546 
Changes:
generic_search
GetAllSolutions function added to generic_search, this function is made available through any_solver to the outside. This allows Fin to get the full list of solutions so that it may manage compilation manually.
Related function GetAllConfigs is also introduced for code modularity.

any_solver has functions added
GetAllSolutions - returns list of all solutions, used for kernel compiling in Fin.
GetPerfCfgParams - returns the parameter string for the solver's performance configuration. This is used for db recording in Fin. 
LegacySolver - creates a bool at compile time that identifies legacy solvers

execution_context & find_solutions
add db_update bool to context - This allows Fin to programmatically initiate a db update tuning.

solver.hpp + <solver>.cpp files
changes necessary for new functions in generic_search

binary_cache
spelling fix


cderb:
@averinevg consolidated the changes to those needed in enabling perf eval functionality for fin.

src/solver/conv_asm_3x3u.cpp
```diff
@@ -70,8 +70,7 @@ PerformanceConfigConvAsm3x3U::PerformanceConfigConvAsm3x3U(int lwc, int fpw, int
 {
 }
 
-inline bool
-PerformanceConfigConvAsm3x3U::operator==(const PerformanceConfigConvAsm3x3U& other) const
+bool PerformanceConfigConvAsm3x3U::operator==(const PerformanceConfigConvAsm3x3U& other) const
```
JehandadKhan, line 73: Can we move the inlined operators to the headers ? That would keep them inline and avoid a function call cost?

JehandadKhan (dismissed):
A minor comment, otherwise looks ok

src/solver/conv_asm_3x3u.cpp
cderb: Sure, I'll move those around.

src/include/miopen/solver.hpp
```diff
@@ -2429,8 +2516,15 @@ struct ConvOclBwdWrW2 : ConvTunableSolver<PerformanceConfigConvOclBwdWrw2<N_BATC
...
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wweak-template-vtables"
+#pragma clang diagnostic ignored "-Wunneeded-internal-declaration"
 #endif
 
+extern template struct PerformanceConfigConvOclBwdWrw2<1>;
+extern template struct PerformanceConfigConvOclBwdWrw2<2>;
+extern template struct PerformanceConfigConvOclBwdWrw2<4>;
+extern template struct PerformanceConfigConvOclBwdWrw2<8>;
+extern template struct PerformanceConfigConvOclBwdWrw2<16>;
```
averinevg, line 2525: @cderb Could you please explain why do you need this code to be here?

src/solver/conv_ocl_dir2D_bwdWrW_2.cpp
```diff
@@ -745,6 +733,12 @@ ConvOclBwdWrW2<N_BATCH_LOOPS>::Search(const ConvolutionContext& context,
 /// We need to instantiate required classes implicitly.
 /// The reason is that we do not define the whole template class
 /// in the header, only declaring it there.
+template struct PerformanceConfigConvOclBwdWrw2<1>;
+template struct PerformanceConfigConvOclBwdWrw2<2>;
+template struct PerformanceConfigConvOclBwdWrw2<4>;
+template struct PerformanceConfigConvOclBwdWrw2<8>;
+template struct PerformanceConfigConvOclBwdWrw2<16>;
```
averinevg, line 740: @cderb Could you please explain why do you need this code to be here?

averinevg (changes requested):
This PR contain some changes not related to the main purpose, especially in files where there are no other changes.

I would suggest moving the non-library code to where it is being used as much as possible.

averinevg:
@atamazov Please review if you have time. Thanks.

src/include/miopen/solver.hpp
```diff
@@ -246,7 +246,11 @@ struct PerformanceConfigConvAsm3x3U : Serializable<PerformanceConfigConvAsm3x3U>
     bool IsValidValue() const;
     bool SetNextValue(const ConvolutionContext& config);
     bool IsValid(const ConvolutionContext& config) const;
-    bool operator==(const PerformanceConfigConvAsm3x3U& other) const;
+    inline bool operator==(const PerformanceConfigConvAsm3x3U& other) const
```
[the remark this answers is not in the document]
cderb: Because of the changes in generic_search, inline can't be used when the operator is defined in the cpp file. 
The function definition could stay in the cpp file, but it would be declared as a standard function.
To use inline we would need to move the function to the header.
Per Jehandad's comment I moved the function to the header to use inline to avoid function call cost.

src/include/miopen/any_solver.hpp
```diff
@@ -140,22 +154,69 @@ struct AnySolver
...
+
+        bool TestSysDbRecord_(const DbRecord& record, std::true_type) const
         {
             using PerformanceConfig = decltype(
                 value.GetDefaultPerformanceConfig(std::declval<const ConvolutionContext&>()));
             PerformanceConfig config{};
             return record.GetValues(value.SolverDbId(), config);
         }
-        bool TestSysDbRecord(const DbRecord& record, std::false_type) const
+        bool TestSysDbRecord_(const DbRecord& record, std::false_type) const
```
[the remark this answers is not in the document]
cderb: I'll take some time to revise the naming. I meant to distinguish the functions that were doing compile time branching.

src/include/miopen/solver.hpp
cderb: Changes in generic_search required forward declaration of solvers to avoid compiler warnings.

src/include/miopen/solver.hpp
```diff
@@ -345,7 +374,19 @@ struct PerformanceConfigConvBiasActivAsm1x1U : PerformanceConfigConvAsm1x1U
...
+        // clang-format off
+                return read_size == other.read_size
+                    && k_mult == other.k_mult
+                    && chunks_per_wave == other.chunks_per_wave
+                    && chunk_size == other.chunk_size
+                    && n_mult == other.n_mult
+                    && c_mult == other.c_mult
+                    && waves_c_in_group == other.waves_c_in_group
+                    && waves_k_in_group == other.waves_k_in_group
+                    && use_spare_set == other.use_spare_set; // clang-format on
```
averinevg, line 388: Please, fix formatting here

src/include/miopen/solver.hpp
```diff
@@ -2400,8 +2494,15 @@ struct ConvOclBwdWrW2 : ConvTunableSolver<PerformanceConfigConvOclBwdWrw2<N_BATC
 #if defined(__clang__) && defined(CONV_OCL_DIR2D_BWDWRW_2_CPP)
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wweak-template-vtables"
+#pragma clang diagnostic ignored "-Wunneeded-internal-declaration"
```
averinevg, line 2497: Why do you need to suppress this warning?

src/include/miopen/solver.hpp
```diff
@@ -2366,7 +2451,16 @@ struct PerformanceConfigConvOclBwdWrw2
...
     bool IsValid(const ConvolutionContext& params) const;
-    bool operator==(const PerformanceConfigConvOclBwdWrw2<N_BATCH_LOOPS>& other) const;
+    inline bool operator==(const PerformanceConfigConvOclBwdWrw2<N_BATCH_LOOPS>& other) const
+    {
+        // clang-format off
+	    return n_waves == other.n_waves
+		&& read_size == other.read_size
+		&& n_out_channels_per_tile == other.n_out_channels_per_tile
+		&& n_out_channels_tiles == other.n_out_channels_tiles
+		&& n_out_rows_in_lcl == other.n_out_rows_in_lcl; // clang-format on
```
averinevg, line 2461: Please, fix formatting here

averinevg (changes requested):

src/include/miopen/solver.hpp
```diff
@@ -242,11 +243,26 @@ struct PerformanceConfigConvAsm3x3U : Serializable<PerformanceConfigConvAsm3x3U>
...
     }
 
+    // clang-format off
+    auto PerfFieldRules() const
+    {
+        return seq::MakeRuleSet(
+            std::make_tuple(seq::Span<int, 0, 9>{}, &PerformanceConfigConvAsm3x3U::limit_wave_cnt),
+            std::make_tuple(seq::Span<int, 1, 8>{}, &PerformanceConfigConvAsm3x3U::filters_per_wave),
+            std::make_tuple(seq::Span<int, 1, 8>{}, &PerformanceConfigConvAsm3x3U::output_lines_per_wave)
+        );
```
JehandadKhan, line 253: Why did you have to add this ?

JehandadKhan (approved):
One minor comment, looks good otherwise

src/include/miopen/solver.hpp
DrizztDoUrden: What do you mean? This is a declarative way to make an iterable pref config without tons of code duplication for generic search.

src/include/miopen/solver.hpp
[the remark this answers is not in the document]
DrizztDoUrden: It shouldn't have noticeable performance difference in any meaningful case (unless 0 configs were valid). For a couple calls to this operator on average you would have a compilation and kernel launch. Couple int comparison would have no noticeable effect even with high precision clock. At the same time the header is already massive. And these calls would probably be virtualized later anyway.
Also modern day linkers actually can inline stuff from other cpp files sometimes.
Overall I would like you to revert the inlining of comparison operators for perf config.

src/solver/conv_asm_3x3u.cpp
DrizztDoUrden: This is ONLY called from generic search. It would not have ANY noticeable difference.

src/include/miopen/solver.hpp
DrizztDoUrden: I see what has happened here. The whole RuleSet concept is basically consteval std::integral_constant based computation.

DrizztDoUrden (changes requested):
LGTM besides the == operators should not be in the header.

src/include/miopen/solver.hpp
cderb: removed

src/include/miopen/solver.hpp
cderb: removed

DrizztDoUrden (dismissed):
LGTM

src/include/miopen/generic_search.hpp
```diff
@@ -296,6 +296,43 @@ using RunAndMeasure_t =
...
+    using PerformanceConfig = decltype(s.GetDefaultPerformanceConfig(context));
+
+    ComputedContainer<PerformanceConfig, Context> primary(context);
+    const int main_size = std::distance(primary.begin(), primary.end());
+    ComputedContainer<PerformanceConfig, Context> spare(context, true);
+    const int spare_size = std::distance(spare.begin(), spare.end());
+    const bool useSpare  = (main_size == 0);
+
+    ComputedContainer<PerformanceConfig, Context> all_configs = useSpare ? spare : primary;
+    const int n_runs_total                                    = useSpare ? spare_size : main_size;
```
averinevg, line 312:
```suggestion
    const int primary_size = std::distance(primary.begin(), primary.end());
    ComputedContainer<PerformanceConfig, Context> spare(context, true);
    const int spare_size = std::distance(spare.begin(), spare.end());
    const bool useSpare  = (primary_size == 0);

    ComputedContainer<PerformanceConfig, Context> all_configs = useSpare ? spare : primary;
    const int n_runs_total                                    = useSpare ? spare_size : primary_size;
```

averinevg (dismissed):

atamazov (changes requested):
Let me remind about re-implementing this [with callbacks](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1445#discussion_r819867212). Why:
- Callback-based implementation involves smaller number of changes.
- It does not have the drawback [mentioned here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1445#discussion_r819028492).

cderb:
> Let me remind about re-implementing this [with callbacks](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1445#discussion_r819867212). Why:
> 
> * Callback-based implementation involves smaller number of changes.
> * It does not have the drawback [mentioned here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1445#discussion_r819028492).

I'll draft that up in a separate pr.

atamazov:
@cderb After second thought, I guess maybe it's time to kill ComputedContainers and replace them with STL.

Yes, ComputedContainers are virtually endless, but this feature is not really used. After all, now we build kernels for all possible PerformanceConfigs before benchmarking. If there are too many PerformanceConfigs, then the system will first run out of disk space. Usual STL container of PerformanceConfigs will run out of space much later.

So I am not against brute-force approach. And removal of ComputedContainers should make our code simpler.

atamazov:
@cderb So please decide yourself. If you decide to use STL, then go ahead. I will get rid of ComputedContainers after that, in a separate PR.

cderb:
@atamazov I'm leaning more towards this implementation (using STL). It looks more clean from an interface level to have a function that specifically returns the array of solutions. The callback implementation I've tested and is functional, but I have some concerns about expanding the function signature of GenericSearch / what unintended uses that function pointer could enable.

atamazov:
@cderb Ok.

junliume:
@atamazov  @DrizztDoUrden @JehandadKhan  last round call for the reviewers, we plan to proceed with this one.

junliume (dismissed):

JehandadKhan (dismissed):

averinevg (approved):

src/include/miopen/any_solver.hpp
```diff
@@ -166,12 +235,46 @@ struct AnySolver
...
         float GetWti(const ConvolutionContext& ctx) const override { return value.GetWti(ctx); }
+
         ConvSolution FindSolution(const ConvolutionContext& ctx,
                                   Db& db,
                                   const miopen::AnyInvokeParams& invoke_ctx) const override
         {
             return miopen::solver::FindSolution(value, ctx, db, invoke_ctx);
         };
+
+        std::string GetPerfCfgParams(const ConvolutionContext& ctx, Db& db, std::true_type) const
```
atamazov, line 246: The body of this function looks like a modified copy of very similar code from `FindSolutionImpl` (src/include/miopen/find_solution.hpp). Can we generalize this?

cderb: The objective of this function is to return the parameter string from the performance config. I see the portion that is common here, and most likely it was the example for this addition. They are entirely similar in that they 
1. use the same type declaration
2. load from the db
3. check validity of the config loaded
4. do a function x with the loaded config
5. log progress
The main difference being #4 the way to minimize this code would be to create an additional function providing a callback in place of #4.
Disregarding log statements and syntactical spacing there are a total of 4 shared lines. Not sure if it's worth complicating with a callback.

src/include/miopen/any_solver.hpp
```diff
@@ -158,6 +188,45 @@ struct AnySolver
...
         }
 
+        // tunable legacy solver
+        std::vector<ConvSolution>
+        GetAllSolutions(const ConvolutionContext& ctx, std::true_type, std::true_type) const
+        {
+            std::vector<ConvSolution> solutions;
+            solutions.push_back(value.GetSolution(ctx, value.GetDefaultPerformanceConfig(ctx)));
+            return solutions;
+        }
```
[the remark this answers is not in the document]
atamazov: > we don't expect to call this for LegacySolvers.

Exactly! Then let's simply MIOPEN_THROW here.

src/include/miopen/any_solver.hpp
```diff
@@ -158,6 +188,45 @@ struct AnySolver
...
+        // tunable solver, not legacy
+        std::vector<ConvSolution>
+        GetAllSolutions(const ConvolutionContext& ctx, std::true_type, std::false_type) const
+        {
+            return miopen::solver::GetAllSolutions(value, ctx);
+        }
+
+        // non tunable solver
+        std::vector<ConvSolution>
+        GetAllSolutions(const ConvolutionContext& ctx, std::false_type, std::true_type) const
```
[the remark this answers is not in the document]
atamazov: Let's MIOPEN_THROW and be ensured that unwanted calls do not happen.

src/include/miopen/solver/implicitgemm_util.hpp
```diff
@@ -607,6 +607,11 @@ template <typename T>
 T gcd(T x, T y)
 {
     assert(!(x == 0 && y == 0));
+    // for clang
+    if(x == 0 && y == 0)
+    {
+        return 1;
```
atamazov, line 613: gcd is undefined when `(x == 0 && y == 0)` so it is incorrect to return 1.

Can you please clarify why this is necessary? IIRC we were sure that `(x == 0 && y == 0)` would never happen, and therefore decided to remove run-time checks from release builds for the sake of performance. Finally we decided to add assert so debug builds can catch the problem (see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/243). So far, clang hasn't complained.

If we really need to handle this case in release builds, then let's replace assert with MIOPEN_THROW.

atamazov (changes requested):

src/include/miopen/any_solver.hpp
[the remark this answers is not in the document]
cderb: I'll change this to a throw.

src/include/miopen/solver/implicitgemm_util.hpp
cderb: There is an issue with clang. When clang is performing branch exploration it is certain that this function can return zero (because x and y are both zero). This obviously cannot happen, because of the assert. Possibly because of the recursive logic here, the compiler abbreviates its check and assumes this function returns a zero, in which case there is a divide by 0 exception in a caller to this function. This unreachable statement here resolves the issue for clang.

src/include/miopen/any_solver.hpp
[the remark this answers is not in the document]
cderb: I found a use case for this. It will be useful for this to return the solution for a non-tunable solver. This will allow gathering of timing information (fdb) at the same time as pdb.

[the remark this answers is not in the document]
atamazov: > This will allow gathering of timing information (fdb) at the same time as pdb.

But fdb update should happen after **_all tuning_** (tuning of all Solvers) is done. And if you would like to collect both fdb and pdb at once, how would you prevent configs from non-tunable solvers from getting into pdb?

src/include/miopen/solver/implicitgemm_util.hpp
atamazov: I see. Do you mean clang-tidy checks or warning during compilation with clang?

src/include/miopen/any_solver.hpp
cderb: If we remove this function, I'll need a way to retrieve the serialized performance parameters. Could be the winner from the solver or from a solution. I'm searching for another method, but nothing yet.

src/include/miopen/solver/implicitgemm_util.hpp
cderb: It was a make check error.

src/include/miopen/any_solver.hpp
atamazov: Can we add this function to a PR that will actually use it? This will allow us to expedite the review of this one.

src/include/miopen/solver/implicitgemm_util.hpp
cderb: I'll add this.

src/include/miopen/any_solver.hpp
atamazov: I see. If you know SolverId, then it is possible to construct an instance of the Solver and then follow the design pattern of:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/72d89b40ee97666efaac0a0979a48680bd02f1da/src/include/miopen/find_solution.hpp#L123-L125
I can do this for you in a future PR where you will implement Fin changes. Or you can ask @DrizztDoUrden to assist.

But let's avoid adding unused code in this PR. Is it acceptable for you?

src/include/miopen/any_solver.hpp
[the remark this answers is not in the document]
atamazov: Now I see, thanks. Resolved!

src/include/miopen/any_solver.hpp
```diff
@@ -158,6 +188,44 @@ struct AnySolver
             return TestSysDbRecord(record, std::integral_constant<bool, TunableSolver::Is>());
         }
 
+        // tunable legacy solver
+        std::vector<ConvSolution>
+        GetAllSolutions(const ConvolutionContext& ctx, std::true_type, std::true_type) const
+        {
+            std::ignore = ctx;
+            MIOPEN_THROW("No solutions returned for Legacy Solvers.");
+        }
```
atamazov, line 197:
```suggestion
        GetAllSolutions(const ConvolutionContext&, std::true_type, std::true_type) const
        {
            MIOPEN_THROW("No solutions returned for Legacy Solvers.");
        }
```

src/solver/conv_hip_implicit_gemm_bwd_v4r1.cpp
```diff
@@ -218,7 +218,7 @@ PerformanceImplicitGemmBwdDataV4R1::CalculateGemmABlockCopyPerformanceParameters
         // decide threadwise copy lengths
         const auto a_data_per_thread_copy_gemmm = SrcDataPerRead_GemmM;
         const auto a_data_per_thread_copy_gemmk =
-            a_data_per_thread_copy / a_data_per_thread_copy_gemmm;
+            a_data_per_thread_copy / a_data_per_thread_copy_gemmm; // NOLINT
```
atamazov, line 221: It was not necessary before your PR, why it is necessary now? Does clang-tidy on CI blame this?

src/solver/conv_hip_implicit_gemm_fwd_v4r4.cpp
```diff
@@ -316,7 +316,7 @@ PerformanceImplicitGemmV4R4Fwd::CalculateGemmBBlockCopyPerformanceParameters(
         DstDataPerWrite_GemmN = gcd(DstDataPerWrite_GemmN, b_data_per_thread_copy_gemmn);
 
         // calculate blockwise copy thread cluster lengths
-        ClusterLengths_GemmK = GemmKPerBlock / b_data_per_thread_copy_gemmk;
+        ClusterLengths_GemmK = GemmKPerBlock / b_data_per_thread_copy_gemmk; // NOLINT
```
atamazov, line 319: Ditto.

src/solver/conv_hip_implicit_gemm_bwd_v4r1.cpp
cderb: http://micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLibs/pipelines/MIOpen/branches/cderb%252Ffin_perf_solver_mergres2/runs/38/nodes/13/steps/87/log/?start=0
yes

atamazov: [Informative] @cderb I can't access CI (micimaster.amd.com)

cderb: > [Informative] @cderb I can't access CI (micimaster.amd.com)

Oh, Sorry!
```
[2022-06-08T18:16:45.191Z] /var/jenkins/workspace/n_cderb_fin_perf_solver_mergres2/src/solver/conv_hip_implicit_gemm_bwd_v4r1.cpp:221:36: error: Division by zero [clang-analyzer-core.DivideZero,-warnings-as-errors]
[2022-06-08T18:16:45.191Z]             a_data_per_thread_copy / a_data_per_thread_copy_gemmm;
```
``` 
[2022-06-08T18:16:49.973Z] /var/jenkins/workspace/n_cderb_fin_perf_solver_mergres2/src/solver/conv_hip_implicit_gemm_fwd_v4r4.cpp:319:46: error: The result of the '/' expression is undefined [clang-analyzer-core.UndefinedBinaryOperatorResult,-warnings-as-errors]
[2022-06-08T18:16:49.973Z]         ClusterLengths_GemmK = GemmKPerBlock / b_data_per_thread_copy_gemmk;
```
which was just this prior commit [9563a5e](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1548/commits/9563a5e49ae0f3ed218de11168ce6b33a90bd71f)

cderb: I'll branch it and see of removing the nolints would work.

cderb: Confirmed that after merge with develop, removing the nolint will still result in a static analyzer error.
Note that this error started after making functions in generic_search callable from any_solver.

atamazov: Ok

src/include/miopen/any_solver.hpp
atamazov: Please send me proposed Fin changes (where you call `GetPerfCfgParams`) and I will offer a substitute. This is necessary, because

> from the architectural point of view, getting tuning config from the database is not a Solver's job.

src/solver/conv_hip_implicit_gemm_bwd_v4r1.cpp
atamazov: [Recommendation] But I would prefer in this case replacing NOLINT with
```C++
if(b_data_per_thread_copy_gemmk == 0)
   MIOPEN_THROW("DIV/0 with b_data_per_thread_copy_gemmk");
```
Up to you.

src/include/miopen/solver/implicitgemm_util.hpp
atamazov: It seems like cppcheck-suppress is not necessary anymore. Resolved.

src/include/miopen/any_solver.hpp
cderb: I'll email the code section to your amd address. Do you still have access?

atamazov: @cderb No, unfortunately. But you can send me mail to DXC (please check your amd inbox).

src/include/miopen/execution_context.hpp
```diff
@@ -67,6 +67,7 @@ struct ExecutionContext
 {
     // Operation modes & environment
     bool do_search               = false;
+    bool db_update               = false;
```
atamazov, line 70: This is only initialized but not updated anywhere. Can we make this change when it's really needed (thus following the YAGNI principle)?

src/solver/conv_hip_implicit_gemm_bwd_v4r1.cpp
atamazov: Resolved!

src/solver/conv_hip_implicit_gemm_fwd_v4r4.cpp
atamazov: Resolved!

src/include/miopen/execution_context.hpp
cderb: It's used in fin, at conv_fin.hpp:559

atamazov: Yes, but the problem is that this PR does NOT include changes in conv_fin.hpp.

cderb: This could be resolved by changes in #1588

atamazov (approved):
Let's move forward with whatever we already have. I will try to refactor the code after some ongoing rework (that involves reworking of the PerformanceConfig class hierarchy) is done.

/cc @averinevg

junliume (approved):

PR #1551, opened 2022-05-19
[API] Add the deterministic attribute to the convolution descriptor

JehandadKhan:


junliume (dismissed):

src/gemm_v2.cpp
```diff
@@ -121,6 +121,34 @@ std::ostream& operator<<(std::ostream& stream, const GemmDescriptor& gemm_desc)
...
+    }
+    return cur_mode;
+}
+
+inline void SetRocblasAtomics(const miopen::Handle& handle, rocblas_atomics_mode mode)
+{
+    MIOPEN_LOG_I2("");
+    rocblas_status status = rocblas_set_atomics_mode(handle.rhandle().get(), mode);
+    assert(status == rocblas_status::rocblas_status_success);
+    (void)status; // WA till C++17 [[maybe_unused]]
```
atamazov, line 147: [Note] If WG21 continue to litter the language with syntactic sugar, then termites will come and gobble everything 🐜 

BTW `std::ignore = status` can be used instead of `(void)`, I guess (`std::ignore` is defined in `<tuple>` IIRC)

src/gemm_v2.cpp
```diff
@@ -495,6 +523,10 @@ miopenStatus_t CallGemm(const Handle& handle,
         {
             ProfilingRecordStart(handle, start, stop);
         }
+        rocblas_atomics_mode cur_mode =
+            rocblas_atomics_mode::rocblas_atomics_allowed; // default value from rocblas
+        if(gemm_desc.deterministic)
+            cur_mode = DisableRocblasAtomics(handle);
```
atamazov, line 529: [Recommendation]
If we assume that the default mode or rocBLAS is non-deterministic, then maybe we can simply
```suggestion
        if(gemm_desc.deterministic)
            DisableRocblasAtomics(handle);
```
and then 
```C++
        if(gemm_desc.deterministic)
            EnableRocblasAtomics(handle);
```
at the end?

And we do not need to use `rocblas_get_atomics_mode` at all because we assume `rocblas_atomics_allowed`.

Then we can wrap this into a class, like this AutoEnableProfiling
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2eb766cdfe4afbc1d62ca6d972cdbd9c0016cfa8/src/include/miopen/handle.hpp#L271
and keep only one call (constructor) in `CallGemm()`, like
```C++
    AutoDisableRocblsAtomics disableAtomics{handle, gemm_desc.deterministic};
```
Dtor that enables atomics will be called automatically.

Or course it is not necessary to assume `rocblas_atomics_allowed` by default. `AutoDisableRocblsAtomics` may call `rocblas_get_atomics_mode` in ctor and do nothing of it returns `rocblas_atomics_not_allowed`.

atamazov (commented):
Partial review

src/gemm_v2.cpp
JehandadKhan: I like this idea, will update the PR

JehandadKhan: Upon further thought, I realized what happens if TensorFlow sets the flag already and breaks the assumption ? 

cc @atamazov

junliume (approved):

PR #1555, opened 2022-05-24
Bump up MIOpen version to 2.18 (after cutting rocm-rel-5.2) and reduce Jenkinsfile size

### LINKED ISSUE #1542: MIOPEN jenkinsfile fails with method code too large after reboot
solution is to add this in jenkins script console but doesn't survive reboot need to make sure these can be a permanent fix
org.jenkinsci.plugins.pipeline.modeldefinition.parser.RuntimeASTTransformer.SCRIPT_SPLITTING_TRANSFORMATION=true

junliume:
1. Bump up MIOpen version to 2.18 (after cutting rocm-rel-5.2);
2. Reduce Jenkinsfile size by removing legacy and outdated stages
also trying to resolve #1542


DrizztDoUrden (approved):
LGTM

Issue #1564, opened 2022-05-27
Conv Driver fails with default `vector_length` parameter


Slimakanzer:
Conv Driver throws an error with #1532 PR.

## Example
```
./bin/MIOpenDriver convfp16 -c 1 -H 700 -W 700 -y 5 -x 20 -k 32 -n 2 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -F 1 -V 1 -i 1 -t 1

MIOpenDriver convfp16 -c 1 -H 700 -W 700 -y 5 -x 20 -k 32 -n 2 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -F 1 -V 1 -i 1 -t 1
MIOpen(HIP): Info [get_device_name] Raw device name: gfx1030
MIOpen(HIP): Info [Handle] stream: 0x12a6060, device_id: 0
Invalid Tensor Vectorization Parameter Value - vector_dim:0vector_length:1
```

Seems the error in the `conv_driver` and the check should be `(vector_length == 1 && vector_dim == 0)`: 
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/93d1476463d281ff731e8aea394c44b09cdc1da7/driver/conv_driver.hpp#L575-L588

aska-0096:
Thanks @Slimakanzer for report the problem, I made a stupid mistake.
Thanks @carlushuang for implement the hot fix.

PR #1567, opened 2022-05-28
[Hotfix][Issue 1564] Fix Conv Driver Failure with default vector_length parameter

carlushuang:
hot fix issue #1564 

Without this PR, our `MIOpenDriver` will fail every time you launch the driver

cc @aska-0096 @atamazov @Slimakanzer


aska-0096 (approved):
Thanks for implement hot fix. LGTM.

junliume (approved):

Issue #1566, opened 2022-05-27
[GFX10] verification issues


Slimakanzer:
Verification fails after #1544 PR.

## Local setup
- rocm 5.1.0
- gfx1030

## Failed cases

```bash
$ export MIOPEN_FIND_MODE=1
$ ./bin/MIOpenDriver conv -c 1 -H 24 -W 24 -y 3 -x 3 -k 1 -n 1 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -F 1 -V 1 -i 1 -t 1
    MIOpen Forward Conv. Algorithm: 1, Solution: 11/ConvOclDirectFwd
    Forward Convolution FAILED: 0.501405 > 1.5e-06
$ ./bin/MIOpenDriver convfp16 -c 1 -H 24 -W 24 -y 3 -x 3 -k 1 -n 1 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -F 1 -V 1 -i 1 -t 1
    MIOpen Forward Conv. Algorithm: 1, Solution: 11/ConvOclDirectFwd
    Forward Convolution FAILED: 0.545767 > 0.0082

$ export MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvBinWinogradRxSf2x3g1
$ ./bin/MIOpenDriver conv -c 1 -H 24 -W 24 -y 3 -x 3 -k 1 -n 1 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -F 1 -V 1 -i 1 -t 1
    Forward Convolution FAILED: 0.501405 > 1.5e-06
$ ./bin/MIOpenDriver convfp16 -c 1 -H 24 -W 24 -y 3 -x 3 -k 1 -n 1 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -F 1 -V 1 -i 1 -t 1
    Forward Convolution FAILED: 0.545809 > 0.0082

$ export MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvBinWinogradRxSf3x2
$ ./bin/MIOpenDriver conv -c 1 -H 24 -W 24 -y 3 -x 3 -k 1 -n 1 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -F 1 -V 1 -i 1 -t 1
    Forward Convolution FAILED: 0.501405 > 1.5e-06

[...]

PR #1568, opened 2022-05-28
[HotFix][Issue 1566][gfx10] fix verification issues

carlushuang:
fix issue https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1566

Basically this is due to acc_type selection is not correct, introduced by PR:#1544.
This will block all `MIOpenDriver` cpu validation.


aska-0096 (approved):
Thanks for the fast hotfix! The problem is accumulator type of int8 cpu conv. we should use double for int8xint8xfloat and int32 for int8xint8xint32.

junliume (approved):

Issue #1576, opened 2022-06-02
Regression in direct naive gpu convolution caused false positive test breakages


jerryyin:
I believe gpu reference convolution kernel for bwd is broken after #1532. I found this via bisecting into MIOpen commit history.

### Before #1532 

Command
```bash
$ git checkout f3b4f1898
$ MIOPEN_FIND_MODE=1 MIOPEN_DRIVER_USE_GPU_REFERENCE=1 MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvDirectNaiveConvBwd ./bin/MIOpenDriver convfp16 --forw 2 --in_layout NCHW --out_layout NCHW --fil_layout NCHW -n 256 -c 1024 -H 14 -W 14 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
```

Output
```bash
MIOpenDriver convfp16 --forw 2 --in_layout NCHW --out_layout NCHW --fil_layout NCHW -n 256 -c 1024 -H 14 -W 14 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
MIOpen Backward Data Conv. Algorithm: 1, Solution: 86/ConvDirectNaiveConvBwd
GPU Kernel Time Backward Data Conv. Elapsed: 245.581238 ms (average)
stats: name, n, c, ho, wo, x, y, k, flopCnt, bytesRead, bytesWritten, GFLOPs, GB/s, timeMs
stats: bwdd-conv1x1u1, 256, 1024, 1, 1, 256, 14, 14,  26306674688, 134742016, 25690112, 107, 1, 245.581238
Backward Convolution Data Verifies OK on GPU reference (0)
```

### After #1532 + #1567 + #1568

Command
```bash
$ git checkout a22af06
[...]

PR #1578, opened 2022-06-03
[HotFix][Issue 1576] Fix bwd conv gpu reference bug

aska-0096:
Problem and solution described in #1576 
Summary: caused by the way to construct the host tensor which used to hold the gpu reference data.
cc: @atamazov

### DIFF
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -3151,7 +3151,7 @@ int ConvDriver<Tgpu, Tref>::RunBackwardDataGPUReference()
         din_dev->FromGPU(GetStream(), din_host.data.data());
     else
     {
-        auto din_tmp = tensor<Tgpu>(miopen::deref(inputTensor).GetType());
+        auto din_tmp = tensor<Tgpu>(miopen::deref(inputTensor));
         din_dev->FromGPU(GetStream(), din_tmp.data.data());
         for(int i = 0; i < din_tmp.data.size(); i++)
         {
--- a/driver/driver.hpp
+++ b/driver/driver.hpp
@@ -139,7 +139,7 @@ inline void PadBufferSize(size_t& sz, int datatype_sz)
 [[gnu::noreturn]] inline void Usage()
 {
     printf("Usage: ./driver *base_arg* *other_args*\n");
-    printf("Supported Base Arguments: conv[fp16|fp16x4|fp16x8|int8|bfp16], CBAInfer[fp16], "
+    printf("Supported Base Arguments: conv[fp16|int8|bfp16], CBAInfer[fp16], "
            "pool[fp16], lrn[fp16], "
            "activ[fp16], softmax[fp16], bnorm[fp16], rnn[fp16], gemm, ctc, dropout[fp16], "
            "tensorop[fp16], reduce[fp16,fp64]\n");


driver/conv_driver.hpp
```diff
@@ -3151,7 +3151,7 @@ int ConvDriver<Tgpu, Tref>::RunBackwardDataGPUReference()
         din_dev->FromGPU(GetStream(), din_host.data.data());
     else
     {
-        auto din_tmp = tensor<Tgpu>(miopen::deref(inputTensor).GetType());
```
jerryyin, line 3154: Good catch, talking out loud on what's happening here: 
 - On the right hand side, it is trying to create a tensor from the data type (`miopenDataType_t`). 
 - Looking at the constructor of the tensor, it "thought" that it is creating a new tensor of size of the enum value.

Reasoning this through I believe implicitly, it was doing an implicit cast from `miopenDataType_t` to `size_t`.  C++ does allow it to happen silently.

(Now, independent from your PR), to prevent this from happening, I think those enum type can be instead amended to `enum class` type. However, since those miopen type definitions happen from a C header, it is probably not possible... So nothing we can do to convert this to a compilation error.

jerryyin (approved):
LGTM

jerryyin:
Actually can we tract the use case of gpu reference kernel in `test/CMakeList.txt` as a smoke test?

aska-0096:
> Actually can we track the use case of gpu reference kernel in `test/CMakeList.txt` as a smoke test?

Good idea. Let me add it to this pr.

junliume (approved):

aska-0096:
> > Hi @jerryyin I remembered that we actually have a ctest to track the gpu reference kernel..
> 
> @aska-0096 How come that one didn't break MIOpen CI between #1532 and now?

Hi @jerryyin , I think this is the drawback of MIOpen ctest design: 
we have a separate `test driver` for convolution, see` test/conv_common`, that means every convolution API is called by test driver instead of MIOpen convolution driver, as the result, our primary convolution driver not covered by the ctest.

jerryyin:
@aska-0096 Okay thanks for the explanation. Though gpu refernece kernel does not necessarily get tracked in `conv_common`, it can be its own test target in ctest. To have it covered, I believe the workaround is to add it into `test/CMakeLists.txt` as a independent test target. I'll create a ticket to track it later.

atamazov:
Guys, AFAICS this is a bug in _**convolution validation**_ (that employs GPU reference kernels instead of CPU) in **_MIOpenDriver_**.

So far, we do not consider MIOpenDriver as a "product" (it's just a "tool") and therefore we do not have tests for it.

@junliume @JehandadKhan @pfultz2 Shall we add some tests for MIOpenDriver, what do you think? I can open a ticket and implement.

junliume:
> Guys, AFAICS this is a bug in _**convolution validation**_ (that employs GPU reference kernels instead of CPU) in **_MIOpenDriver_**.
> 
> So far, we do not consider MIOpenDriver as a "product" (it's just a "tool") and therefore we do not have tests for it.
> 
> @junliume @JehandadKhan @pfultz2 Shall we add some tests for MIOpenDriver, what do you think? I can open a ticket and implement.

Yes, that would be a good idea. :) @atamazov

atamazov:
> @JehandadKhan @pfultz2 Shall we add some tests for MIOpenDriver, what do you think? I can open a ticket and implement.

Guys, what do you think?

jerryyin:
@atamazov From MLIR end, this sounds like a good idea because we apply MIOpenDriver tests on the fly (from a moving MLIR branch) to validate the correctness. So if there's basic sanity guarantee it will make our life easier. From MIOpen end, it sounds fair as long as this is tested in a smoke environment like single stage + single hardware + hip + conv only setup. (just to avoid duplication from ctest)

On the scope of this ticket/PR, I think it is okay to add gpu reference kernel validation either in ctest or in MIOpenDriver test.

PR #1580, opened 2022-06-03
[HotFix] fix macro guard issue in config.h.in

junliume:
As per https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1543/files#r889131708

This is an error introduced in #1543


junliume:
Close since implemented in #1581

atamazov:
@junliume This problem does not look too serious. The fix is provided in #1581, so this PR can be dropped, I guess. Up to you ;)

PR #1587, opened 2022-06-07
[Tuning][MI100]  pdb duplication fix

cderb:
Duplicate config entry raised during merge, corrected here.


junliume (approved):
I think we should have some perf tests on it?

PR #1588, opened 2022-06-08
Add fin as a submodule

JehandadKhan:
This PR removes the old fin code which was added verbatim and replaces it with a git submodule. This allows better tracking with the upstream repo which is developed independently.


JehandadKhan:
In draft state until I verify behavior of Jenkins with git submodules.

pfultz2:
Why is fin developed independently? If it needs to be then why doesn't MIOpen use fin with `find_package`?

Or is fin more tightly coupled with MIOpen that it needs to build with MIOpen? If so, then why is this in a separate repo?

junliume:
@JehandadKhan "Tuna Fin Build Test" stage has failed, and I guess that's expected?

cderb:
Completed work with converting Fin to a submodule.

JehandadKhan:
@cderb @junliume I am unable to approve this PR, since I am the original author.

cderb:
> @cderb @junliume I am unable to approve this PR, since I am the original author.

Curiously enough I can approve it even though I was the last to make changes.

cderb:
@junliume This removes the Fin code copy from MIOpen, and instead it is tracked as a submodule. This change would remove the diffs we see in MIOpen when we change Fin. MIOpen changes here are for how the Jenkinsfile interacts with and builds the Fin repo.

Jenkinsfile
```diff
@@ -130,9 +186,8 @@ def getDockerImage(Map conf=[:])
     def miotensile_version = conf.get("miotensile_version", "default") // deprecated
     def target_id = conf.get("target_id", "OFF") // deprecated
     def mlir_build = conf.get("mlir_build", "ON") // always ON
-    def build_fin = conf.get("build_fin", "OFF") // forcing this to be always on means fewer docker containers since this ensures all dependencies are present in the docker image
     def no_cache = conf.get("no_cache", false)
-    def dockerArgs = "--build-arg BUILDKIT_INLINE_CACHE=1 --build-arg PREFIX=${prefixpath} --build-arg GPU_ARCH='${gpu_arch}' --build-arg MIOTENSILE_VER='${miotensile_version}' --build-arg USE_TARGETID='${target_id}' --build-arg USE_MLIR='${mlir_build}' --build-arg USE_FIN='${build_fin}' "
+    def dockerArgs = "--build-arg BUILDKIT_INLINE_CACHE=1 --build-arg PREFIX=${prefixpath} --build-arg GPU_ARCH='${gpu_arch}' --build-arg MIOTENSILE_VER='${miotensile_version}' --build-arg USE_TARGETID='${target_id}' --build-arg USE_MLIR='${mlir_build}' "
```
atamazov, line 190: Remove USE_FIN from Dockerfile.

atamazov: @cderb Let's remove unused code.

atamazov: @JehandadKhan oops

junliume: @atamazov should `USE_FIN` be removed from everywhere in the repo? I may need to update dockerfile too.

atamazov: @junliume I do not know. But removal of `USE_FIN='${build_fin}'` at this line in Jenkinsfile makes USE_FIN unused in the Dockerfile, so it should be removed.

PR #1589, opened 2022-06-09
[MLIR] Adjust int8 kPerBlock tuning range

jerryyin:
For int8 xdlops algorithm, all values between 1, 2, 4 will be rejected because the k/block size is too narrow for supported xdlops instruction. Adjusting the tuning range such that MIOpen can tune xdlops kernel correctly.

Note, this tuning range has already been tested through MLIR CI.


junliume:
@jerryyin Do you need this PR in 5.3? If so please mark as a blocker. Thanks!

junliume (approved):

jerryyin:
@junliume That's right! Thanks for reminding.

atamazov:
@jerryyin @JehandadKhan This https://github.com/ROCmSoftwarePlatform/MIOpen/labels/IMPACTS_PERFORMANCE_DB

jerryyin:
@atamazov This is compatible with https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1539. Just so that the right tuning range is applied to int8. Previous range would work but will yield sub-optimal kernels because the k/block values are too small.

jerryyin:
FYI guys, all testing has passed for this PR.

atamazov:
> @atamazov This is compatible with #1539. Just so that the right tuning range is applied to int8. Previous range would work but will yield sub-optimal kernels because the k/block values are too small.

So it looks like the set of _configs used during tuning_ for INT8 is smaller than the set of _valid configs_ (`MiirIsConfigApplicable`)? This is unusual mismatch, but seems acceptable.

atamazov (approved):
LGTM!

jerryyin:
> set of configs used during tuning for INT8 is smaller than the set of valid configs (MiirIsConfigApplicable)?

More precisely speaking, the previous tuning range was there because they fit for fp32 and fp16. Now with the int8 enabled, that tuning range originally targeted for other data types no long make sense in this new scenario. I can make the tuning range to be [1, 32], however that means:
 - For int8, all kernels generated between [1,8] are doomed to be slower ones. (In my local testing of [1,32], not one case does tuning select a k/block less than 8). So it saved tuning time not to include this range. 
 - For fp32/fp16, all kernels generated between [8,32] may mostly be rejected, or likely slow because it is too large.

PR #1593, opened 2022-06-15
[CI][Jenkins] Store wrapper credentials in Env Var

JehandadKhan:


junliume (approved):
Hot Patch to unblock CI pipelines. Thanks!

PR #1594, opened 2022-06-15
[Enhancement] Fix potential issues in MT environments. Resolve warning and code clean ups.

atamazov:
This PR is a result of reviewing of the `cppcoreguidelines-avoid-non-const-global-variables` warning suppressions. Many of them were added in PR #761. Some of them camouflage potential issues in multi-threaded environments; some can be easily resolved by proper initialization or code cleanup.

- 🔴 Fixed MT-unsafe construct in `get_type_name<Solver>`
- 🔴 Fixed MT-unsafe construct related to static id in `MDGraph_vertex`
- [Style] Properly fix the warning -- use `static const` and initialize by lambda
- [Style] Moved testing_find_db_* into debug namespace
- [Style] Moved StaticContainer to the place where it is used (into tests)
- [Cleanup] Removed mlo_utils.hpp (contains only unreachable code). Removed some extra license headers.

@junliume @JehandadKhan @shurale-nkn @DrizztDoUrden Please have a look.


JehandadKhan (approved):

junliume (dismissed):

atamazov:
> @atamazov How does moving the function to the `debug` namespace help with the issue ? I believe there is a gap in my understanding here.

@JehandadKhan This change is listed as "[Style] Moved testing_find_db_* into debug namespace" in the description. It does not resolve the potential issue, but clearly indicates that the issue may happen only when debug functionality is used. This informs the reader that the importance of this problem is low. So this is about readability of our code.

atamazov:
@junliume @DrizztDoUrden can you please help me with CI? I see that some of the checks are not successful.

atamazov:
@junliume @DrizztDoUrden @shurale-nkn Tidy issues cleaned up. Can you please push it via our CI and let me know if any issues, thanks.

src/include/miopen/type_name.hpp
```diff
@@ -34,11 +34,8 @@ namespace miopen {
 template <class MIOpen_Private_TypeName_>
 const std::string& get_type_name()
 {
-    // NOLINTNEXTLINE (cppcoreguidelines-avoid-non-const-global-variables)
-    static std::string name;
-
-    if(name.empty())
-    {
+    static const std::string ret = []() {
```
pfultz2, line 37: Since this relies on `__PRETTY_FUNCTION__` to get the type name then this can't be put into a lambda. I think you will need to make a seperate `compute_type_name` function and then call it. Thats what we do [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/type_name.hpp#L34).

atamazov: Good catch, Paul! but let me do that separately. This is a bit out of scope of this PR and I would like to look how this stuff is used.

pfultz2: Sure, but we have unit tests here that could fail:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/test/type_name.cpp#L24

If the tests do pass then you can ignore my comment.

atamazov: Indeed, in spite of `__PRETTY_FUNCTION__` outputs full signature, including the caller's name (https://reviews.llvm.org/D33497) the test fails. Let me fix this.

atamazov: Resolved!

atamazov:
Guys, let's re-review, thanks!

shurale-nkn (approved):
LGTM

atamazov:
@junliume Thank you!

PR #1595, opened 2022-06-15
[Enhancement] FIND_MODE 1 - remove workaround for SWDEV_227826

JehandadKhan:
This workaround causes `MIOPEN_FIND_MODE` to behave incorrectly when set to `1`. The underlying issue of slow startup times is no longer relevant since the default find mode for MIOpen is `5` resulting in a search only along dynamic solvers.

### DIFF
--- a/src/mlo_dir_conv.cpp
+++ b/src/mlo_dir_conv.cpp
@@ -52,15 +52,6 @@
 #include <miopen/sqlite_db.hpp>
 #endif
 
-// Only select the first applicable igemm solver due to long compilation time
-// (JIRA SWDEV-227826)
-/// \todo enable all applicable solvers of igemm after fixing slow compilation
-#define WORKAROUND_SWDEV_227826 1
-
-#if WORKAROUND_SWDEV_227826
-MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS)
-#endif
-
 miopen::PerformanceDb mlo_construct_base::GetDb() const
 {
     return {db_path(), _search_params.GetUserPerfDbPath()};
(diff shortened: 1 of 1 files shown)


JehandadKhan:
@atamazov  Can you please take a look?

atamazov:
@JehandadKhan 
> This workaround causes MIOPEN_FIND_MODE to behave incorrectly when set to 1. 

Can you please elaborate -- why if works incorrectly?

atamazov (approved):
The change is good. Okay to go if you are 100% sure that performance won't be affected. Or you can simply redefine the macro to 0 (just to keep a quick way to get back in case of issues)

JehandadKhan:
> @JehandadKhan
> 
> > This workaround causes MIOPEN_FIND_MODE to behave incorrectly when set to 1.
> 
> Can you please elaborate -- why if works incorrectly? I mean, what is happening?

When a user sets `MIOPEN_FIND_MODE=1` They expect performance to either be on par or better than out of the box performance. However, due to this workaround in place, the performance is actually lower when compared to OOB performance.  Since the env var is widely communicated and documented it creates the impression that a bug exists.

junliume (approved):

atamazov:
@JehandadKhan I would be for Jun's solution. Note that we do not have parallel builds of HIP kernels with COMGR/hipRTC. However, I am not against this PR, especially if this problem irritates end users or hinders QA to triage performance issues.

atamazov (approved):
LGTM

DrizztDoUrden (approved):
LGTM

pfultz2:
This is good. We use `MIOPEN_FIND_MODE=1` in migraphx through MIOpen's hidden API. We definitely dont want to require the user to set an env variable(like `MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS`) to get good performance from MIOpen.

PR #1607, opened 2022-06-26
Normalise sphinx doc

v01dXYZ:
Small inconsequential MR that reorganises and normalises the doc (rename files, prefer rst over md when possible).


atamazov (changes requested):
Is it necessary to switch from markdown to restructured text? I see quite a few formatting issues. For example, in `debug_and_logging.rst`:
- Throughout all text
  - Missing italic, bold and bold italic formatting.
- Formatting of Notes looks broken
  - Missing vertical line on the left
  - Unnecessary line breaks after "Note"
  - Missing note numbers and colons.
- Under "ASM Implicit GEMM" and "HIP Implicit GEMM"
  - Formatting of lists looks wrong (unnecessary vertical line on the left)
- Under "Filtering out all Solutions except one"
  - References to Note 4 look broken (but maybe this is how github shows the references in the browser)
  - Typo: "nfail"

For quick comparison, you can use the following links:
- [Original .md](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/DebugAndLogging.md)
- [.rst from this PR after commit ad955c8ac](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ad955c8ace634641eb6c2ce18272b937107ec48a/doc/src/debug_and_logging.rst).

v01dXYZ:
Thank you for your review.

It is not necessary to enforce rst. But I thought it would have been an amelioration to only have one markup language for the documentation (there are currently two). We can also enforce `MyST` instead which is as much as powerful as `rst`. I didn't take this path but actually it is the most efficient one as the `api_reference` files are the easiest to change. 

In the two cases, GitHub won't be able to nicely render them.

The aim of the MR is to simplify as much as possible the docs:

* use only one markup
* have an arborescence of files instead of a flat directory
* use sphinx directives to ameliorate the way the documentation is rendered.

I hope it would be helpful if new topics are added to the documentation in the future (such as documenting the assembly files in `src/kernels` or LLVM/MLIR for code generation with custom passes...).

(also it is a way for me to introduce me to the code base by solving an easy problem, so if it is not necessary, let's forget it)

PR #1611, opened 2022-06-28
Sl/google test

xinlipn:
The project is to move tests from CTest to GoogleTest, which takes advantage of the same input data e.g. for testing on CPU and GPU, and saves the intermediate results for later comparison. This would avoid computing repetitively on the same data set and thus reduce testing time.

- Moved cache.cpp from /test to google test folder /test/gtest
- Replace main() in cache.cpp with google test Macro TEST()

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@
 # SOFTWARE.
 #
 ################################################################################
-cmake_minimum_required( VERSION 3.5 )
+cmake_minimum_required( VERSION 3.11 )
 
 macro(set_var_to_condition var)
     if(${ARGN})
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ echo $ROCM_APT_VER &&\
 sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/$ROCM_APT_VER/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
 RUN sh -c "echo deb http://mirrors.kernel.org/ubuntu bionic main universe | tee -a /etc/apt/sources.list"
 
+
 #Add gpg keys
 # Install dependencies
 RUN apt-get update && \
@@ -30,10 +31,13 @@ DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated \
     apt-utils && \
 apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 9386B48A1A693C5C && \
 wget -q -O - https://repo.radeon.com/rocm/rocm.gpg.key | apt-key add - && \
+wget --no-check-certificate -qO - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | apt-key add - && \
+sh -c "echo deb https://apt.kitware.com/ubuntu/ bionic main | tee -a /etc/apt/sources.list" && \
 apt-get update && \
 DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated \
     build-essential \
-    cmake \
+    cmake-data=3.15.1-0kitware1 \
+    cmake=3.15.1-0kitware1 \
     comgr \
     clang-format-10 \
     doxygen \
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,6 @@ echo $ROCM_APT_VER &&\
 sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/$ROCM_APT_VER/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
 RUN sh -c "echo deb http://mirrors.kernel.org/ubuntu bionic main universe | tee -a /etc/apt/sources.list"
 
-
 #Add gpg keys
 # Install dependencies
 RUN apt-get update && \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -305,6 +305,7 @@ def buildDocker(install_prefix)
     }
 
     echo "Build Args: ${dockerArgs}"
+    /*Force to rebuild a docker file to include higher version of CMake, will revert shortly
     try 
     {
         echo "Checking for image: ${image_name}"
@@ -317,6 +318,9 @@ def buildDocker(install_prefix)
         retimage = docker.build("${image_name}", dockerArgs + ' .')
         retimage.push()
     }
+    */
+    retimage = docker.build("${image_name}", dockerArgs + ' .')
+    retimage.push()
 }
 
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -305,7 +305,6 @@ def buildDocker(install_prefix)
     }
 
     echo "Build Args: ${dockerArgs}"
-    /*Force to rebuild a docker file to include higher version of CMake, will revert shortly
     try 
     {
         echo "Checking for image: ${image_name}"
@@ -317,10 +316,7 @@ def buildDocker(install_prefix)
         echo "Unable to locate image: ${image_name}. Building image now"
         retimage = docker.build("${image_name}", dockerArgs + ' .')
         retimage.push()
-    }
-    */
-    retimage = docker.build("${image_name}", dockerArgs + ' .')
-    retimage.push()
+    } 
 }
 
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -316,7 +316,7 @@ def buildDocker(install_prefix)
         echo "Unable to locate image: ${image_name}. Building image now"
         retimage = docker.build("${image_name}", dockerArgs + ' .')
         retimage.push()
-    } 
+    }
 }
 
 
--- a/cmake/googletest.cmake
+++ b/cmake/googletest.cmake
@@ -1,0 +1,39 @@
+include(FetchContent)
+
+set(GOOGLETEST_DIR "" CACHE STRING "Location of local GoogleTest repo to build against")
+
+if(GOOGLETEST_DIR)
+  set(FETCHCONTENT_SOURCE_DIR_GOOGLETEST ${GOOGLETEST_DIR} CACHE STRING "GoogleTest source directory override")
+endif()
+
+message(STATUS "Fetching GoogleTest")
+
+list(APPEND GTEST_CMAKE_CXX_FLAGS 
+     -Wno-undef
+     -Wno-reserved-identifier
+     -Wno-global-constructors
+     -Wno-missing-noreturn
+     -Wno-disabled-macro-expansion
+     -Wno-used-but-marked-unused
+     -Wno-switch-enum
+     -Wno-zero-as-null-pointer-constant
+     -Wno-unused-member-function
+     -Wno-comma
+     -Wno-old-style-cast
+)
+message(STATUS "Suppressing googltest warnings with flags: ${GTEST_CMAKE_CXX_FLAGS}")
+
+FetchContent_Declare(
+  googletest
+  GIT_REPOSITORY https://github.com/google/googletest.git
+  GIT_TAG        e2239ee6043f73722e7aa812a459f54a28552929
+)
+
+# Will be necessary for windows build
+# set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
+
+target_compile_options(gtest PRIVATE ${GTEST_CMAKE_CXX_FLAGS})
+target_compile_options(gtest_main PRIVATE ${GTEST_CMAKE_CXX_FLAGS})
+target_compile_options(gmock PRIVATE ${GTEST_CMAKE_CXX_FLAGS})
+target_compile_options(gmock_main PRIVATE ${GTEST_CMAKE_CXX_FLAGS})
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1737,3 +1737,4 @@ add_custom_test(test_regression_half_mi200 GFX900_DISABLED GFX906_DISABLED GFX90
     # Issue-internal #4
     COMMAND	${ENVS_FIND_ONLY_HIP_IGEMM_V4R4XDLOPS} $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --cmode conv --pmode default --input 120 64 75 75 --weights 128 64 1 1 --pads_strides_dilations 0 0 2 2 1 1 ${ARGS_ENABLE_FORWARD_ONLY}
 )
+add_subdirectory(gtest)
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -213,7 +213,7 @@ else()
 endif()
 
 if (MIOPEN_NO_GPU)
-    set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params
+    set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params test_lstm test_lstm_dropout
             test_test_errors test_type_name test_tensor_test test_sqlite_perfdb test_sequences
             test_pooling3d test_perfdb)
 endif()
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -213,7 +213,7 @@ else()
 endif()
 
 if (MIOPEN_NO_GPU)
-    set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params test_lstm test_lstm_dropout
+    set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params
             test_test_errors test_type_name test_tensor_test test_sqlite_perfdb test_sequences
             test_pooling3d test_perfdb)
 endif()
--- a/test/gtest/CMakeLists.txt
+++ b/test/gtest/CMakeLists.txt
@@ -7,12 +7,16 @@ include(GoogleTest)
 
 function(add_gtest TEST_NAME)
   message("Adding Test: " ${TEST_NAME})
-  add_executable(test_${TEST_NAME} ${TEST_NAME}.cpp)
+  add_executable(test_${TEST_NAME} ${TEST_NAME}.cpp )
   add_dependencies(tests test_${TEST_NAME})
   add_dependencies(check test_${TEST_NAME})
   target_compile_options(test_${TEST_NAME} PRIVATE -Wno-global-constructors -Wno-undef)
-  target_link_libraries(test_${TEST_NAME} gtest_main MIOpen ${Boost_LIBRARIES} hip::host $<BUILD_INTERFACE:roc::rocblas>)
-  gtest_discover_tests(test_${TEST_NAME})  #Enable CMake to discover the test binary
+  #Get the path for upper directory
+  get_filename_component(DIR_TWO_ABOVE ../../ ABSOLUTE)
+  #Depency on cache object file for linker, refactor absolute path?
+  target_link_libraries(test_${TEST_NAME} gtest_main MIOpen ${Boost_LIBRARIES} hip::host $<BUILD_INTERFACE:roc::rocblas> ${DIR_TWO_ABOVE}/build/test/CMakeFiles/test_cache.dir/cache.cpp.o)
+  #Enable CMake to discover the test binary
+  gtest_discover_tests(test_${TEST_NAME})
 endfunction()
 
 file(GLOB TESTS *.cpp)
--- a/test/gtest/CMakeLists.txt
+++ b/test/gtest/CMakeLists.txt
@@ -1,0 +1,22 @@
+include(googletest)
+enable_testing()
+include(GoogleTest)
+
+#add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -C ${CMAKE_CFG_INTDIR})
+#add_custom_target(tests)
+
+function(add_gtest TEST_NAME)
+  message("Adding Test: " ${TEST_NAME})
+  add_executable(test_${TEST_NAME} ${TEST_NAME}.cpp)
+  add_dependencies(tests test_${TEST_NAME})
+  add_dependencies(check test_${TEST_NAME})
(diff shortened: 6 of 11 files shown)


atamazov:
@xinlipn Is this PR ready for review? If not, then can you please convert this PR into draft?

Could you please also add some motivation to the PR description? This is important for sharing info with external collaborators who do not have access to AMD Jira. Thanks.

atamazov:
@xinlipn 
> GoogleTest, which takes advantage of the same input data e.g. for testing on CPU and GPU, and saves the intermediate results for later comparison.

Interesting feature. Can you please provide me with a link to the description of it?

test/gtest/cache.cpp
```diff
@@ -145,13 +147,19 @@ void check_cache_str()
...
+    ++marker;
+    EXPECT_EQ(marker, 1);
 #if MIOPEN_ENABLE_SQLITE
-    check_bz2_compress();
-    check_bz2_decompress();
+    check_bz2_compress();    
+    check_bz2_decompress();    
     check_kern_db();
+    ++marker;
+    EXPECT_EQ(marker, 2);
```
atamazov, line 163: Out tests (including this cache.cpp) use our own macro `EXPECT` throughout the sources for validation:
(https://github.com/ROCmSoftwarePlatform/MIOpen/blob/f09132964d628ffb649f4a6ee14dadfc933a7a69/test/test.hpp#L76-L82)
If we want to port it to GoogleTest, then shall we also update all these checks with the ones provided by the GoogleTest framework?

atamazov (changes requested):

atamazov:
[Note] Please also note that CTest is able to test whatever the user wants, while IIRC GoogleTest is suitable for testing of C++ programs only. Therefore all the "custom" tests will have to be left under CTest.

atamazov:
@pfultz2 Can you please glance at this PR? Thanks!

cmake/googletest.cmake
```diff
@@ -0,0 +1,39 @@
...
+     -Wno-comma
+     -Wno-old-style-cast
+)
+message(STATUS "Suppressing googltest warnings with flags: ${GTEST_CMAKE_CXX_FLAGS}")
+
+FetchContent_Declare(
+  googletest
+  GIT_REPOSITORY https://github.com/google/googletest.git
+  GIT_TAG        e2239ee6043f73722e7aa812a459f54a28552929
+)
```
pfultz2, line 30: Dont use FetchContent to pull down dependencies. We do not want to download and build gtest everytime we configure with cmake. This should be listed under dev-requirements.txt file so its installed automatically with rest of the dependencies.

test/gtest/hello_test.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+    // Check if we can access MIOpen internals
+    auto idx = 0;
+    for(const auto& solver_id :
+        miopen::solver::GetSolversByPrimitive(miopen::solver::Primitive::Convolution))
+    {
+        std::ignore = solver_id;
+        ++idx;
+    }
+    EXPECT_GT(idx, 0);
```
pfultz2, line 28: Why cant we write `EXPECT(idx > 0)`?

atamazov:
@pfultz2 Thank you, Paul.

pfultz2:
Just curious, how does gtest handle passing parameters from CLI to tests? 

Its not feasible to create tests cases for every combination of convolution, so we can enumerate the ones that are important and then we should be able to pass flags to run the combinations that were not listed. That is how the current test driver works, but it does a cartesian product of all configurations so it creates a lot of test cases.

From looking at the docs in gtest, it looks like we will still need argument parsing as gtest doesn't take care of that for us whereas catch2 lets us easily add our own command-line arguments: 

https://github.com/catchorg/Catch2/blob/devel/docs/own-main.md#adding-your-own-command-line-options

What the rational for choosing gtest over catch? gtest/catch will work well for our unit tests, but not so well for our acceptance tests. I am not sure either one can do what are current acceptance test drivers do though, but it looks like catch might get us closer there(we could at least take advantage of Clara).

> which takes advantage of the same input data e.g. for testing on CPU and GPU, and saves the intermediate results for later comparison

The current test driver uses the same data for the CPU and GPU as well and it can save the CPU results too, but it is too large for all the test cases we run(which why it is not enabled by default). I did try to use bzip2 but it seems uncompressing was just about as slow(or even slower) for most cases as recomputing the CPU version(zstd might be a better choice for this).

There are few cases where the same config produces different input values because we are calling the test differently and using `std::rand` in a lot of places.

To improve our current test drivers we need:

* Enumerate the test config we want instead of cartesian product of all values
* Stop using `std::rand` to generate values(an [`xorshf96_generator`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/generate.hpp#L72) like what we use in migraphx is quite fast).

But using gtest/catch wont help fix any of these issues expect maybe forcing a refactoring of the tests(which is definitely needed).

test/gtest/hello_test.cpp
pfultz2: But will `EXPECT_TRUE` decompose the conditions(so you can see what value of `idx` was in this case)? Thats how it works in Catch, Boost.Test and even in migraphx's [test.hpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/generate.hpp#L72) header.

atamazov:
@pfultz2
> To improve our current test drivers we need:...

We can also use naive GPU implementations instead of CPU to compute reference data. This feature is implemented in the driver, but not in conv tests yet.

> maybe forcing a refactoring of the tests(which is definitely needed).

😉

xinlipn:
> @xinlipn
> 
> > GoogleTest, which takes advantage of the same input data e.g. for testing on CPU and GPU, and saves the intermediate results for later comparison.
> 
> Interesting feature. Can you please provide me with a link to the description of it?

Thanks everyone for reviewing and suggestions.
@atamazov, here's a link about GoogleTest FYI:
http://google.github.io/googletest/primer.html

atamazov:
@xinlipn If the idea behind re-using the same input and output data is based on [fixture classes](http://google.github.io/googletest/primer.html#same-data-multiple-tests), then it requires serious redesign on existing tests.

test/gtest/hello_test.cpp
xinlipn: > But will `EXPECT_TRUE` decompose the conditions(so you can see what value of `idx` was in this case)? Thats how it works in Catch, Boost.Test and even in migraphx's [test.hpp](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/generate.hpp#L72) header.

@pfultz2 Seems EXPECT_TRUE can output error message, e.g. below. Will update the shortly 
EXPECT_TRUE(marker == 1) << "test_activation failed";

xinlipn:
> 

Thank you @atamazov and @pfultz2 for your time and input.  Yes, TEST_F() offers the same data configurations for different tests.

I pushed some changes as below:

1. Macro EXPECT_EQ is replaced with EXPECT_TRUE per Paul's suggestion. 

2. I also started refactoring activation.cpp for a simple "--float" test case, which may need further improvement to include the logic in test/CMakeLists.txt,

pfultz2:
> Macro EXPECT_EQ is replaced with EXPECT_TRUE per Paul's suggestion.

So doing `EXPECT_TRUE(x == y)` gtest will show what the values for the variables `x` and `y` are?

> I also started refactoring activation.cpp for a simple "--float" test case, which may need further improvement to include the logic in test/CMakeLists.txt,

Is the plan to just have our test cases be gtests that just invoke the test driver? So we just keep the test driver and remove the `--all` option and lists in ctests?

Enumerating the tests in C++(with gtest or catch) does provide some nice filtering for running different subsets of tests. However, there is a compile-time cost that can grow as we add more tests or change existing tests.

A better option would be to have a json file with the different test configurations(its actually faster to parse json than to parse and compile C++). We would just need to provide our own filtering for this.

atamazov:
@xinlipn 
>  Yes, TEST_F() offers the same data configurations for different tests.

Can you please elaborate the machinery you are going to use in the fixture classes to enable the "re-using the data" feature in tests? For example, what is the expected lifetime of the instances of fixture classes (and, consequently, what is expected lifetime of the re-usable data they own)?

test/gtest/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+#add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -C ${CMAKE_CFG_INTDIR})
+#add_custom_target(tests)
+
+function(add_gtest TEST_NAME)
+  message("Adding Test: " ${TEST_NAME})
+  add_executable(test_${TEST_NAME} ${TEST_NAME}.cpp )
+  add_dependencies(tests test_${TEST_NAME})
+  add_dependencies(check test_${TEST_NAME})
+  target_compile_options(test_${TEST_NAME} PRIVATE -Wno-global-constructors -Wno-undef)
```
JehandadKhan, line 13: Adding the include directory would remove the requirement of up-leveling each directory
```
target_include_directories(test_${TEST_NAME} PRIVATE ../)
```

cmake/googletest.cmake
xinlipn: Thanks Paul, this is based on the example provided by google test. Let me take a look into it.

https://google.github.io/googletest/quickstart-cmake.html

xinlipn:
> Just curious, how does gtest handle passing parameters from CLI to tests?

Here's a thread discussing passing CLI parameters to gtest 
https://github.com/google/googletest/issues/765

test/gtest/CMakeLists.txt
xinlipn: Thanks and done

xinlipn:
> Just curious, how does gtest handle passing parameters from CLI to tests?
> 
> Its not feasible to create tests cases for every combination of convolution, so we can enumerate the ones that are important and then we should be able to pass flags to run the combinations that were not listed. That is how the current test driver works, but it does a cartesian product of all configurations so it creates a lot of test cases.
> 
> From looking at the docs in gtest, it looks like we will still need argument parsing as gtest doesn't take care of that for us whereas catch2 lets us easily add our own command-line arguments:
> 
> https://github.com/catchorg/Catch2/blob/devel/docs/own-main.md#adding-your-own-command-line-options
> 
> Whats the rational for choosing gtest over catch? gtest/catch will work well for our unit tests, but not so well for our acceptance tests. I am not sure either one can do what are current acceptance test drivers do though, but it looks like catch might get us closer there(we could at least take advantage of Clara).
> 
> > which takes advantage of the same input data e.g. for testing on CPU and GPU, and saves the intermediate results for later comparison
> 
> The current test driver uses the same data for the CPU and GPU as well and it can save the CPU results too, but it is too large for all the test cases we run(which why it is not enabled by default). I did try to use bzip2 but it seems uncompressing was just about as slow(or even slower) for most cases as recomputing the CPU version(zstd might be a better choice for this).
> 
> There are few cases where the same config produces different input values because we are calling the test differently and using `std::rand` in a lot of places.
> 
> To improve our current test drivers we need:
> 
> * Enumerate the test config we want instead of cartesian product of all values
> * Stop using `std::rand` to generate values(an [`xorshf96_generator`](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/include/migraphx/generate.hpp#L72) like what we use in migraphx is quite fast).
> 
> But using gtest/catch wont help fix any of these issues expect maybe forcing a refactoring of the tests(which is definitely needed).

xinlipn:
> @xinlipn If the idea behind re-using the same input and output data is based on [fixture classes](http://google.github.io/googletest/primer.html#same-data-multiple-tests), then it requires serious redesign on existing tests.

Yes, tests need to be redesigned.

xinlipn:
> @xinlipn
> 
> > Yes, TEST_F() offers the same data configurations for different tests.
> 
> Can you please elaborate the machinery you are going to use in the fixture classes to enable the "re-using the data" feature in tests? For example, what is the expected lifetime of the instances of fixture classes (and, consequently, what is expected lifetime of the re-usable data they own)?

@atamazov Please correct me if I'm wrong. my understanding is current tests are individual executables and  the test data e.g. for activation test, exists during the lifetime of the test after they are initialized, e.g. get_input_tensor(tensor_elem_gen_integer{miopen_type<T>{} == miopenHalf ? 5 : 17})); //activation.cpp

I'm thinking the lifetime of fixture class instances along with the data would also be the same of the test. Does this make sense?

test/gtest/cache.cpp
xinlipn: @atamazov , Seems EXPECT() is doing similar things, my understanding is yes, we may need to update the Macros with those provided by GoogleTest, does this make sense?

xinlipn:
> For example, in migraphx doing:
> 
> ```c++
> TEST_CASE(expect)
> {
>     int x = 1;
>     int y = 2;
>     EXPECT(x == y);
> }
> ```
> 
> It will print out:
> 
> ```
> [   RUN    ] expect
> void expect()
> AMDMIGraphX/test/main.cpp:30:
>     FAILED: x == y [ 1 == 2 ]
> [  FAILED  ] expect: Test failure
> [==========] 1 tests ran
> [  FAILED  ] 1 tests failed
> [  FAILED  ] expect
> ```
> 
> Does gtest show something similiar?

@pfultz2, yes, it will have serial spew, here's an example for EXPECT_GT(0, 1);
...
MIOpen/test/gtest/hello_test.cpp:23: Failure
Expected: (0) > (1), actual: 0 vs 1
...

xinlipn:
> Just curious, how does gtest handle passing parameters from CLI to tests?

@pfultz2  Here's an example for passing parameters :) 

https://stackoverflow.com/questions/9496730/can-i-pass-parameters-to-googletest-test-function

JehandadKhan:
@xinlipn  Please make sure that the CI passes and that the newly added tests are run

JehandadKhan (dismissed):

xinlipn:
The CI issue appears to be related to no response from nc -N jenkins-raptor-11.amd.com 6379
[2022-07-21T05:25:32.015Z] + nc -N jenkins-raptor-11.amd.com 6379
script returned exit code 1

Compared with previous build, below is expected
FOUND CCACHE SERVER: jenkins-raptor-11.amd.com

Jun has ping'd Nyamodi, Leon  and kakarparthi, omkar...

atamazov:
@xinlipn 
> my understanding is current tests are self contained individual executables and the test data e.g. for activation test, exists during the lifetime of the test after they are initialized

This is correct.

> I'm thinking the lifetime of fixture class instances along with the data would also be the same of the test. Does this make sense?

Yes, it makes sense. Especially if you have tests that internally may re-use the same _test data_ more than once. But currently all our tests use _test data_ only once.

For example, the test of a convolution generates input tensors and reference output tensor -- these comprise the _test data_. Then the test runs the convolution from the library with the inputs, gets actual output and compares it with the reference. Then the test throws out all the tensors. If it wants to test some other convolution (the convolution that has different dimensions), then previous _test data_ is useless and needs to be regenerated.

So maybe I am wrong but I do not see how the the "re-using the data" feature can improve the situation.

JehandadKhan:
@atamazov  The rationale for using the GTest is to use a standardized testing framework which is thoroughly documented and comprehensive, allowing new team members to understand and contribute more easily. Furthermore, I believe it is more comprehensive and would reduce the size of our code base. However, the current testing framework is not going anywhere, however, new tests would only be added using the GTest to keep things uniform. 

This PR only adds the dependency and a sample test as the stepping stone before we can add more tests or port some of the existing tests to GTest framework. Since GTest has a clean setup and tear down mechanism for test data I am hoping that porting some of the tests to GTest would enable us to correctly cache the test data and avoid recomputing it allowing us to extend our code coverage without increasing the testing time substantially. Attempts to do the same via existing framework have not been successful since the code paths there are non deterministic and I am hoping GTest would not suffer the same issue. 

I hope this answers some of the questions you might have and help expedite the review.

xinlipn:
> @xinlipn [Note about the process] If you do not think that PR is ready for review/merging, then please convert it to draft. Thanks.

Thank you Artem and Jehandad for the input. The CI currently failed at an older version of CMake. Should be able to merge once this is fixed.

atamazov:
@JehandadKhan I am by no means opposed to continuing this research))

junliume:
@xinlipn clang format issues can be fixed by the following script: https://github.com/ROCmSoftwarePlatform/MIOpen/wiki/How-to-format-code

xinlipn:
> @xinlipn clang format issues can be fixed by the following script: https://github.com/ROCmSoftwarePlatform/MIOpen/wiki/How-to-format-code

@junliume Thank you. Code from other branch was accidentally checked in to this branch. I'm cleaning up and will commit the changes soon.

junliume:
@xinlipn Thanks! Some other tidy issues before the static check can run through: http://[micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLibs/pipelines/MIOpen/branches/sl%252Fgoogle_test/runs/20/nodes/97/steps/311/log/?start=0](http://micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLibs/pipelines/MIOpen/branches/sl%252Fgoogle_test/runs/20/nodes/97/steps/311/log/?start=0)

such as 
```
/var/jenkins/workspace/MLLibs_MIOpen_sl_google_test/test/gtest/cache.cpp:157:24: warning: Argument 'marker==1' to function EXPECT_TRUE is always 1 [constArgument]

**Comment by xinlipn** [comment]
> @xinlipn Thanks! Some other tidy issues before the static check can run through: http://[micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLibs/pipelines/MIOpen/branches/sl%252Fgoogle_test/runs/20/nodes/97/steps/311/log/?start=0](http://micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLibs/pipelines/MIOpen/branches/sl%252Fgoogle_test/runs/20/nodes/97/steps/311/log/?start=0)
> 
> such as
> 
> ```
> /var/jenkins/workspace/MLLibs_MIOpen_sl_google_test/test/gtest/cache.cpp:157:24: warning: Argument 'marker==1' to function EXPECT_TRUE is always 1 [constArgument]
> ```

@junliume , thank you for the reminder. This is actually caused by the code accidently checked in. It has been reverted.

**Comment by junliume** [comment]
@xinlipn CI has passed, please ping all reviewers for their opinions. Thanks!

**Comment by xinlipn** [comment]
CI has passed. Could @JehandadKhan @atamazov @pfultz2  take another look? Thank you

http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/sl%2Fgoogle_test/27/pipeline

**Review by JehandadKhan — APPROVED** [review]

File: Jenkinsfile [hunk 5]
```diff
@@ -316,7 +316,7 @@ def buildDocker(install_prefix)
         echo "Unable to locate image: ${image_name}. Building image now"
         retimage = docker.build("${image_name}", dockerArgs + ' .')
         retimage.push()
-    }
```
junliume, line 319: Format issue, no need to trigger another round of CI

CMakeLists.txt
```diff
@@ -23,7 +23,7 @@
 # SOFTWARE.
 #
 ################################################################################
-cmake_minimum_required( VERSION 3.5 )
+cmake_minimum_required( VERSION 3.11 )
```
junliume, line 26: Ubuntu 18.04 has cmake 3.10. Would this potentially cause a problem if releasing team build from source? (I see in our own docker file we have installed 3.15, so where is 3.11 coming from?)

xinlipn: Hi @junliume , 3.11 is the minimal version that supports FetchContent.  If Docker is used on Ubuntu 18.04, I believe the following changes in Dockerfile *should* cover the lower cMake version issue. 

wget --no-check-certificate -qO - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | apt-key add - && \
sh -c "echo deb https://apt.kitware.com/ubuntu/ bionic main | tee -a /etc/apt/sources.list" && \

junliume (approved):

test/CMakeLists.txt
```diff
@@ -213,7 +213,7 @@ else()
 endif()
 
 if (MIOPEN_NO_GPU)
-    set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params test_lstm test_lstm_dropout
+    set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params
```
atamazov, line 216: This change seems to be out of the scope of this PR.

Jenkinsfile
```diff
@@ -316,7 +316,7 @@ def buildDocker(install_prefix)
...
+    }
```
atamazov, line 319: Whitespace-only change, please revert.

Dockerfile
```diff
@@ -18,6 +18,7 @@ echo $ROCM_APT_VER &&\
 sh -c 'echo deb [arch=amd64 trusted=yes] http://repo.radeon.com/rocm/apt/$ROCM_APT_VER/ ubuntu main > /etc/apt/sources.list.d/rocm.list'
 RUN sh -c "echo deb http://mirrors.kernel.org/ubuntu bionic main universe | tee -a /etc/apt/sources.list"
 
+
```
atamazov, line 21: Ditto

atamazov (changes requested):

test/CMakeLists.txt
xinlipn: @atamazov , you are correct. This is for the following PR. Since it's minor, I put it here too to avoid merge conflict. Would you suggest to revert it? Thanks

https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1628

Jenkinsfile
xinlipn: Done. Thanks

Dockerfile
xinlipn: Done. Thanks

test/CMakeLists.txt
atamazov: Yes, please revert.

xinlipn: Reverted. Thanks :)

atamazov (approved):
LGTM

junliume (approved):

PR #1615, opened 2022-06-30
Integrate Kernel DBs into MIOpen repo

JehandadKhan:
This PR enables GIT LFS for large binary code objects in MIOpen
Packaging for each Arch specific code objects is added. 
CMake emits a warning if the file is a link and skips adding it to the package.
Test to detect whether MIOpen's binary loading is able to detect empty git file links
Update the kernels package installation to reflect the new naming.
Git LFS config is setup such that default cloning of the repo would not clone the large files. Two scenarios exsit:

  - No Git LFS installed: Git only creates the soft link and no files are downloaded
  - Git LFS exists: `.lfsconfig` forces git to exclude the large files from downloading code objects. To pull the controlled lfs files the user need to explicitly request so by issuing: `git lfs pull --exclude= origin` 

cc @atamazov @DrizztDoUrden


test/sqlite.cpp
```diff
@@ -0,0 +1,86 @@
...
+int main(int argc, char* argv[])
+{
+    std::ignore = argc;
+    std::ignore = argv;
+#if MIOPEN_ENABLE_SQLITE
+    CHECK(test_lfs_db(
+        true)); // System DB should pass, since the lfs file was installed in the sys directory
+    CHECK(throws(std::bind(
+        test_lfs_db, false))); // User db should fail since MIOpen should not create such a file
+                               // ever, if it exists its a corrupt file which should be reported.
```
atamazov, line 81: Maybe I do not understand things very well, but I have a Q about "should not create such a file ever". Do you mean that MIOpen should not create this file during tests (because all tests must use popular configs)?

But what if the user manually ran some unusual configs and then wants to run tests? Or if a developer will add a regression test with unusual config (and that test will create a user db file during tests)?

test/sqlite.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <fstream>
+#include <mutex>
+#include <random>
+#include <string>
+#include <thread>
+#include <vector>
+#if MIOPEN_ENABLE_SQLITE
+const char* const lfs_db = R"(version https://git-lfs.github.com/spec/v1
+oid sha256:cc45c32e44560074b5e4b0c0e48472a86e6b3bb1c73c189580f950f098d2a8d7
+size 357490688)";
```
atamazov, line 53: What does this depend on (in other words, when should it be updated)?

test/sqlite.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <miopen/db.hpp>
+#include <miopen/db_record.hpp>
+#include <miopen/lock_file.hpp>
+#include <miopen/temp_file.hpp>
+
+#include <boost/filesystem/operations.hpp>
+#include <boost/filesystem/path.hpp>
+#include <boost/optional.hpp>
+#include <boost/thread.hpp>
+
```
atamazov, line 40: This test does not use boost. This test does not use boost. Most likely, it is necessary to revise other headers, apparently there are too many of them.

test/sqlite.cpp
```diff
@@ -0,0 +1,86 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2019 Advanced Micro Devices, Inc.
```
atamazov, line 5: 2022

test/sqlite.cpp
JehandadKhan: This is a sample `git-lfs` file when the actual data has not been downloaded. This is test data and should not need updating.

test/sqlite.cpp
JehandadKhan: Thanks for pointing this out, I have removed the unnecessary headers.

JehandadKhan:
@atamazov  Ping

test/sqlite.cpp
JehandadKhan: Fixed!

junliume (approved):
@JehandadKhan we need to update the documentation mentioning how to download the KDBs too. Thanks!

PR #1616, opened 2022-06-30
[SWDEV-330460][W/A][gfx90a][FP32] disable ConvAsmBwdWrw3x3 for numerical stability

JehandadKhan:
Workaround for [SWDEV-330460](https://ontrack-internal.amd.com/browse/SWDEV-330460)


junliume (approved):

atamazov:
[Recomendation] @junliume Maybe it is better to use explicit "[WORKAROUND]" in commit messages (instead of abbreviations like "[W/A]" etc), just to emphasize workarounds in the history.

atamazov:
@JehandadKhan Note that this PR may require find-db update for FP32 WrW convolutions for the gfx90a target (in order to avoid find-db misses and falling back to GEMM if we have something better). Of course this is necessary only if we have substantial number of ConvAsmBwdWrw3x3 wins))

PR #1618, opened 2022-07-02
[SWDEV-330460] Improve workaround for ConvAsmBwdWrw3x3

atamazov:
Improvement of W/A introduced in #1616: this allows to enable ConvAsmBwdWrw3x3 explicitly by means of `MIOPEN_DEBUG_CONV_DIRECT_ASM_WRW3X3`. Might be useful for experimentation when developing the full-blown fix.

### DIFF
--- a/src/solver/conv_asm_dir_BwdWrW3x3.cpp
+++ b/src/solver/conv_asm_dir_BwdWrW3x3.cpp
@@ -373,7 +373,8 @@ bool ConvAsmBwdWrW3x3::IsApplicable(const ConvolutionContext& params) const
         return false;
 
 #if WORKAROUND_SWDEV_330460
-    if(name == "gfx90a" && params.IsFp32())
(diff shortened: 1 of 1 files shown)


atamazov:
@JehandadKhan One detail I've forgot during review of #1616.

JehandadKhan (approved):

junliume (approved):

JehandadKhan:
@junliume  CI passes

atamazov:
@junliume @JehandadKhan BTW what is the current status of SWDEV-330460? Have we decided to keep this solver disabled for MI200 (and thus transform the W/A into the persistent part of the design of this solver)?

junliume:
> SWDEV-330460

Hi @atamazov  so far we are keeping this WA, since it is for FP32 we are not fixing it at kernel level at high urgencies.

PR #1625, opened 2022-07-07
int8 quantization apis

iq136boy:
Implement the quantizaiton apis in miopen hostside, which is prepared for the integration with the quantization composable kernel.

### DIFF
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -2706,10 +2706,8 @@ miopenCreateQuantizationDescriptor(miopenQuantizationDescriptor_t* quantiDesc);
  * @param bias         bias for quantization (input)
  * @return             miopenStatus_t
  */
-MIOPEN_EXPORT miopenStatus_t
-miopenSetQuantizationDescriptor(const miopenQuantizationDescriptor_t quantiDesc,
-                                double scaler,
-                                double bias);
+MIOPEN_EXPORT miopenStatus_t miopenSetQuantizationDescriptor(
+    const miopenQuantizationDescriptor_t quantiDesc, double scaler, double bias);
 
 /*! @brief Gets the quantization descriptor details
  *
@@ -2720,10 +2718,8 @@ miopenSetQuantizationDescriptor(const miopenQuantizationDescriptor_t quantiDesc,
  * @param bias         bias for quantization (output)
  * @return             miopenStatus_t
  */
-MIOPEN_EXPORT miopenStatus_t
-miopenGetQuantizationDescriptor(const miopenQuantizationDescriptor_t quantiDesc,
-                                double* scaler,
-                                double* bias);
+MIOPEN_EXPORT miopenStatus_t miopenGetQuantizationDescriptor(
+    const miopenQuantizationDescriptor_t quantiDesc, double* scaler, double* bias);
 
 /*! @brief Destroys the quantization descriptor object
  *
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -2725,6 +2725,14 @@ miopenGetQuantizationDescriptor(const miopenQuantizationDescriptor_t quantiDesc,
                                 double* scaler,
                                 double* bias);
 
+/*! @brief Destroys the quantization descriptor object
+ *
+ * @param activDesc   Quantization tensor descriptor type (input)
+ * @return            miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t
+miopenDestroyQuantizationDescriptor(miopenQuantizationDescriptor_t quantiDesc);
+
 /** @} */
 // CLOSEOUT QUANTIZATION DOXYGEN GROUP
 
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -2681,6 +2681,19 @@ MIOPEN_EXPORT miopenStatus_t miopenActivationBackward(miopenHandle_t handle,
 MIOPEN_EXPORT miopenStatus_t
 miopenDestroyActivationDescriptor(miopenActivationDescriptor_t activDesc);
 
+// Quantization APIs
+/** @addtogroup quantization
+ *
+ *  @{
+ */
+/*! @brief Creates the Quantization descriptor object
+ *
+ * @param quantiDesc Pointer to an quantization tensor descriptor type
+ * @return          miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t
+miopenCreateQuantizationDescriptor(miopenQuantizationDescriptor_t* quantiDesc);
+
 /** @} */
 // CLOSEOUT ACTIVATION DOXYGEN GROUP
 
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -2694,6 +2694,34 @@ miopenDestroyActivationDescriptor(miopenActivationDescriptor_t activDesc);
 MIOPEN_EXPORT miopenStatus_t
 miopenCreateQuantizationDescriptor(miopenQuantizationDescriptor_t* quantiDesc);
 
+/*! @brief Sets the quantization descriptor details
+ *
+ * Sets all of the descriptor details for the quantization
+ *
+ * @param quantiDesc   Pointer to a quantization descriptor (output)
+ * @param scaler       scaler for quantization (input)
+ * @param bias         bias for quantization (input)
+ * @return             miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t
+miopenSetQuantizationDescriptor(const miopenQuantizationDescriptor_t quantiDesc,
+                                double scaler,
+                                double bias);
+
+/*! @brief Gets the quantization descriptor details
+ *
+ * Retrieves all of the descriptor details for the quantization.
+ *
+ * @param quantiDesc   Pointer to a quantization descriptor (input)
+ * @param scaler       scaler for quantization (output)
+ * @param bias         bias for quantization (output)
+ * @return             miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t
+miopenGetQuantizationDescriptor(const miopenQuantizationDescriptor_t quantiDesc,
+                                double* scaler,
+                                double* bias);
+
 /** @} */
 // CLOSEOUT ACTIVATION DOXYGEN GROUP
 
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -2727,7 +2727,7 @@ miopenGetQuantizationDescriptor(const miopenQuantizationDescriptor_t quantiDesc,
 
 /*! @brief Destroys the quantization descriptor object
  *
- * @param activDesc   Quantization tensor descriptor type (input)
+ * @param quantiDesc   Quantization tensor descriptor type (input)
  * @return            miopenStatus_t
  */
 MIOPEN_EXPORT miopenStatus_t
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -303,6 +303,14 @@ MIOPEN_DECLARE_OBJECT(miopenLRNDescriptor);
  */
 MIOPEN_DECLARE_OBJECT(miopenActivationDescriptor);
 
+/*! @ingroup quantization
+ * @brief Creates the miopenQuantizationDescriptor_t type
+ *
+ * Quantization descriptor is an object that allows the user triger quantization.
+ *
+ */
+MIOPEN_DECLARE_OBJECT(miopenQuantizationDescriptor);
+
 /*! @ingroup RNN
  * @brief Creates the miopenRNNDescriptor_t type
  */
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -2681,6 +2681,9 @@ MIOPEN_EXPORT miopenStatus_t miopenActivationBackward(miopenHandle_t handle,
 MIOPEN_EXPORT miopenStatus_t
 miopenDestroyActivationDescriptor(miopenActivationDescriptor_t activDesc);
 
+/** @} */
+// CLOSEOUT ACTIVATION DOXYGEN GROUP
+
 // Quantization APIs
 /** @addtogroup quantization
  *
@@ -2723,7 +2726,7 @@ miopenGetQuantizationDescriptor(const miopenQuantizationDescriptor_t quantiDesc,
                                 double* bias);
 
 /** @} */
-// CLOSEOUT ACTIVATION DOXYGEN GROUP
+// CLOSEOUT QUANTIZATION DOXYGEN GROUP
 
 // Softmax APIs
 /** @addtogroup softmax
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -83,6 +83,7 @@ endfunction()
 set( MIOpen_Source
     activ/problem_description.cpp
     activ_api.cpp
+    quanti_api.cpp
     batch_norm.cpp
     batch_norm_api.cpp
     batchnorm/problem_description.cpp
@@ -470,6 +471,7 @@ if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN
     configure_file(db_path.cpp.in ${PROJECT_BINARY_DIR}/db_path.cpp)
     list(APPEND MIOpen_Source
         activ.cpp
+        quanti.cpp
         kernel_cache.cpp
         lrn.cpp
         mlo_dir_conv.cpp
--- a/src/include/miopen/quanti.hpp
+++ b/src/include/miopen/quanti.hpp
@@ -47,14 +47,6 @@ struct QuantizationDescriptor : miopenQuantizationDescriptor
     double GetScaler() const;
     double GetBias() const;
 
-    miopenStatus_t Quantize(Handle& handle,
-                           const TensorDescriptor& inDesc,
-                           ConstData_t in,
-                           const void* scaler,                           
-                           const void* bias,
-                           const TensorDescriptor& outDesc,
-                           Data_t out);
-
     friend std::ostream& operator<<(std::ostream& stream, const QuantizationDescriptor& x);
 
 private:
--- a/src/include/miopen/quanti.hpp
+++ b/src/include/miopen/quanti.hpp
@@ -1,0 +1,65 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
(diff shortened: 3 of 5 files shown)


atamazov:
@iq136boy Is it ready for review? Please mark this PR with the `IMPACTS_API` label, add also urgency label and indicate target milestone (MIOpen release), thanks.

junliume:
@iq136boy  I see CI results for static checks only. Is it intentional? When you convert to ready for review, we would need the whole set of CI I guess.

iq136boy:
> @iq136boy Is it ready for review? Please mark this PR with the `IMPACTS_API` label, add also urgency label and indicate target milestone (MIOpen release), thanks.
@atamazov Not yet. JD and I will go over this PR on our 1:1 meeting today (pre-review). Then I will added you as a reviewer when it is ready. Thanks.

iq136boy:
> @iq136boy I see CI results for static checks only. Is it intentional? When you convert to ready for review, we would need the whole set of CI I guess.

@junliume Yes. That is intentional. The changes still need  pre-review from JD. We will do this today on the 1:1 meeting. I will request review from you after it is ready (all the futures added and CI passed). Thanks.

include/miopen/miopen.h
```diff
@@ -583,6 +583,46 @@ typedef enum
...
+
+/*! @brief Set the quantization bias of Tensor Descriptor
+ *
+ * API for setting the quantization bias of tensor descriptor.
+ * @param tensorDesc  Pointer to a tensor descriptor type
+ * @param quantiBias  quantization bias
+ * @return            miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t miopenSetQuantizationBias(miopenTensorDescriptor_t tensorDesc,
+                                                       double quantiBias);
```
JehandadKhan, line 604:
```suggestion
MIOPEN_EXPORT miopenStatus_t miopenSetQuantizationBias(miopenTensorDescriptor_t tensorDesc,
                                                       double* quantiBias, size_t len);
```

junliume:
On hold and waiting for kernels @zjing14

src/include/miopen/tensor.hpp
```diff
@@ -251,6 +256,9 @@ struct TensorDescriptor : miopenTensorDescriptor
 
     miopenDataType_t type             = miopenFloat;
     miopenTensorLayout_t tensorLayout = miopenTensorNCHW;
+
+    std::vector<double> quantScales{std::vector<double>(1, 1.0f)};
+    std::vector<double> quantBiases{std::vector<double>(1, 0.0f)};
```
jerryyin, line 261: I have some doubts on whether we want to put the quantization states in such a generic place at `tensor.hpp`. The current design seem to indicate the following:
 - We can only do uniform quantization
 - Quantization will and only be able to carry on a single axis
 - That axis is "implicitly" set as the channel axis
 - The fact that we use 2 APIs added the chance of a client to mis-use it and leave invalid state fed to MIOpen

Are we okay with all of those assumptions?

Also, the title seems to be pointing at supporting int8 quantization only. If so I think we need some asserts on:
 - This tensor to be int8 data type, when quantization information is fed into it
 - The size of quantization scale and bias equivalent with channel size
 - They must both exist or both empty. MIOpen shouldn't allow empty scale and a filled bias
 - For convolution, this must only exist on the output tensor, but must not exist on filter and input tensor

Lastly, what does the framework plan to do with those APIs? Can I request their unit test examples?

include/miopen/miopen.h
```diff
@@ -583,6 +583,50 @@ typedef enum
...
+/*! @brief Set the quantization bias of Tensor Descriptor
+ *
+ * API for setting the quantization bias of tensor descriptor.
+ * @param tensorDesc  Pointer to a tensor descriptor type
+ * @param quantBiases Quantization biases
+ * @param len         Length of quantization biases buffer
+ * @return            miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t miopenSetQuantizationBiases(miopenTensorDescriptor_t tensorDesc,
+                                                         double* quantBiases,
```
jerryyin, line 607: How does MIOpen want to kernel to compute with the precision level? I see double precision as input argument but realistically, is this really necessary?

I'm guessing this shall be fp32 (or needs to be converted to fp32), such that int32 convolution result can cast up also to fp32 for computation.

include/miopen/miopen.h
```diff
@@ -583,6 +583,50 @@ typedef enum
...
+/*! @brief Set the quantization scale of Tensor Descriptor
+ *
+ * API for setting the quantization scale of tensor descriptor.
+ * @param tensorDesc  Pointer to a tensor descriptor type
+ * @param quantScales Quantization scales
+ * @param len         Length of quantization scales buffer
+ * @return            miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t miopenSetQuantizationScales(miopenTensorDescriptor_t tensorDesc,
+                                                         double* quantScales,
```
jerryyin, line 595: The impression I get for the element-wise operation is that we use the following equation to compute the result elements:

`int8_res = (int32_res / scale) + bias`

However, due the the overhead of division in gpu, I'd like us to be passed with a multiplication for scaling factor, such that:

`int8_res = (int32_res * new_scale) + bias`

Now, does framework guarantee that they will do the conversion of `new_scale = 1 / scale` when the pass this argument to MIOpen? If we don't have that guarantee, this needs to carry out the division right in the API.

jerryyin (dismissed):
I'm putting a blocker for this API as I'm not clear about some key aspects from the kernel perspective. I'd need all the question addressed before this PR is merged.

include/miopen/miopen.h
```diff
@@ -583,6 +583,50 @@ typedef enum
...
+
+/*! @brief Get the quantization scale of tensor descriptor
+ *
+ * Interface for querying tensor quantization scale.
+ * @param tensorDesc   Tensor descriptor type (input)
+ * @param quantScales  Tensor quantization scales (output)
+ * @return             miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t miopenGetQuantizationScales(miopenTensorDescriptor_t tensorDesc,
+                                                         double* quantScales);
```
JehandadKhan, line 618: How does the user know how much memory to allocate for `quantScales` ?

include/miopen/miopen.h
```diff
@@ -583,6 +583,50 @@ typedef enum
...
+
+/*! @brief Get the quantization scale of tensor descriptor
+ *
+ * Interface for querying tensor quantization Bias.
+ * @param tensorDesc   Tensor descriptor type (input)
+ * @param quantBiases  Tensor quantization biases (output)
+ * @return             miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t miopenGetQuantizationBiases(miopenTensorDescriptor_t tensorDesc,
+                                                         double* quantBiases);
```
JehandadKhan, line 628: Same as above

junliume:
@iq136boy Is this PR on track for the next sprint? If we are at risk, let me know and I'll help.

iq136boy:
> @iq136boy Is this PR on track for the next sprint? If we are at risk, let me know and I'll help.

@junliume Yes, it is on track. In summary, the int8 quantization task can be divided to three parts: 1) the APIs in this PR that for the upper level apps/users. 2) the ck int8 quantization kernel integration in MIOpen and 3) the test of the APIs and integration.
The progresses now are: item 1) is almost done with some minor modifications based on the reviewers comments; item 2) was done early this week; For 3), unfortunately we do not have any naïve convolution kernel with quantization operation in MIOpen to test the newly-integrated ck quantization kernel. So now I am working on adding a new naive convolution kernel with quantization operation in MIOpen.  This is the current status of int8 quantization task.

include/miopen/miopen.h
iq136boy: > How does the user know how much memory to allocate for `quantScales` ?

I will add another parameter that controls the memory size that users need to allocate.

atamazov:
@iq136boy 
> So now I am working on adding a new naive convolution kernel with quantization operation in MIOpen. 

Maybe it is faster to add INT8 QNT at the host side (less work)

iq136boy:
> @iq136boy
> 
> > So now I am working on adding a new naive convolution kernel with quantization operation in MIOpen.
> 
> Maybe it is faster to add INT8 QNT at the host side (less work)

Hi @atamazov , what do you mean by add on "host side"? Di you mean int8 kernel on CPU? Could you give some more details on your thought? Thanks so much.

atamazov:
@iq136boy 
> Hi @atamazov , what do you mean by add on "host side"? Di you mean int8 kernel on CPU? Could you give some more details on your thought? Thanks so much.

Well, maybe I do not understand the intent very well, but if you only need to **_test_** CK QNT kernels, then you can compute reference data on the host. We have host-side implementations of convolutions. This is still used by default for validation in MIOpenDriver. And (which is more important) we use it to validate _naive_ solvers.

Of course, if you need to close any coverage gaps that CK implementation of INT8 QNT likely has, then you need the naive kernels, indeed. And these can be reused to test CK QNT kernels. But some host side implementation is still necessary -- because what will test the naive kernels otherwise? ;)

include/miopen/miopen.h
atamazov: [Recommendation] I recommend describing the operation in comments, e.g.:
```
tensor[n,k,h,w] = quantScale[k] * tensor[n,k,h,w] + quantBias[k]
```
This is expected to resolve some abmiquities mentioned at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1625#discussion_r1095897476

include/miopen/miopen.h
atamazov: [Recommendation] COMgr-like stuff:
```C++
MIOPEN_EXPORT miopenStatus_t miopenGetQuantizationScales(miopenTensorDescriptor_t tensorDesc,
                                                         size_t* numQuantScales);
                                                         double* arrayQuantScales);
...
size_t n;
miopenGetQuantizationScales(td, &n, nullptr); // get number of elements in scales
double *p = ... // allocate array as necessary
n = n / 2 ; // assume we want to read only half of elements
miopenGetQuantizationScales(td, &n, p); // write n elements of scales into *p
```
- *numQuantScales - On entry, the number of elements to get. On return, if arrayQuantScales is NULL, set to the number of scales' elements.
- arrayQuantScales - If not NULL, then the first *numQuantScales elements of the scales array is copied there. If NULL, no data is copied, but *numQuantScales is updated to the number of elements in scales array.

atamazov:
- (1) This PR adds new INT8 QNT data members to the tensor descriptor. These members do not describe the _**tensor**_ itself, they actually describe some _**operation**_ (quantization in this case) _**to be performed**_ with the tensor. I am afraid this does not match the design of the MIOpen API that we have for now.

- (2) Another question is when the quantization should be performed. If I understand correctly from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1625#discussion_r1095897476, this should happen after convolution is computed, right? (please correct me if I am wrong)

Taking all this into account, I think I worth moving INT8 QNT data members to the convolution descriptor.

Note that (if I understand the intent correctly) this is similar to fusing convolution with INT8 quantization. Maybe some different API design approach should be used.

Last but not least -- does INT8 QNT implies allocating memory for a new quantized INT8 tensor?

atamazov:
Or maybe this feature is for [quantization-aware training (QAT)](https://developer.nvidia.com/blog/achieving-fp32-accuracy-for-int8-inference-using-quantization-aware-training-with-tensorrt/) where the user wants to perform "fake" INT8-quantization on all tensors while staying in FP32 domain? 

UPDATE: NO, this is for inference only for now.

src/include/miopen/tensor.hpp
iq136boy: @jerryyin  Hi Zhuoyan, thanks for your valuable comments. I set up a meeting with you, Jun, Jing, Jehandad to discuss about these concerns.

jerryyin:
> And these can be reused to test CK QNT kernels. But some host side implementation is still necessary

@iq136boy This seems to be a very reasonable comment, could you follow-up on this point? In you summarized 3-part, I don't see this point.

> Taking all this into account, I think I worth moving INT8 QNT data members to the convolution descriptor.

@atamazov  Sounds more reasonable than the existing `tensor.hpp`, thanks for your suggestion.

>  does INT8 QNT implies allocating memory for a new quantized INT8 tensor?

According to my understanding, it will not be a newly allocated memory for int8, but rather we replace the int32 output tensor as a int8 output tensor.

Your comments also reminded me about: Both bias and scale are coming in as host vectors. It may need to be copied to be device tensors passed to the kernel. (As kernel will need and dereference those data at compute time)

src/include/miopen/tensor.hpp
jerryyin: @iq136boy Sure, we can discuss about those, but please also include MIGraphX side people. I myself is not all clear about the implementation details they are going to adopt now. If you are to implement the MIOpen/CK implementation, it is best that you also talk with @turneram who work with MIGraphX/CK implementation on this as well.

jerryyin: Besides, could you also follow-up closely with @atamazov's suggestions? He is much more experienced than me, and it is important that you address and engage with his comments.

iq136boy: > Besides, could you also follow-up closely with @atamazov's suggestions? He is much more experienced than me, and it is important that you address and engage with his comments.

In fact, I have already discussed with Jehandad about all the comments from Artem and we thought we need to clarify/discuss more about these comments. That is why we bring all people together to discuss. But you are right, I should leave some comments like "this comment is under discussion" or so to make it clear.

jerryyin: @iq136boy After you have offline discussions, please make sure to sync up your conclusions/confusions to this review. This way the reviewers will have an idea what you have talked about.

Also, don't hesitate to reach out to me if you want to clarify things. This is a interactive process. For myself personally, it only become clear to me what the requirements are until I sync up with stakeholders.

iq136boy: > @iq136boy After you have offline discussions, please make sure to sync up your conclusions/confusions to this review. This way the reviewers will have an idea what you have talked about.
> 
> Also, don't hesitate to reach out to me if you want to clarify things. This is a interactive process. For myself personally, it only become clear to me what the requirements are until I sync up with stakeholders.

OK. I see. I will put more offline discussions in this PR to make it clear to everyone. Thanks for your suggestion.

atamazov:
@iq136boy @JehandadKhan What is required length of bias/scale used with weights tensor? `(C*K)` or... ?

include/miopen/miopen.h
atamazov: BTW how does the the library know how many elements from `quantScales` should be used? This is necessary to know because we need to copy data from CPU to GPU.

If this is hardwired to `C`, then the API becomes rigid in terms of future extension (but the answer to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1625/files#r1128369353 becomes very simple: `K`).

I would prefer flexibility. Let's add another parameter, `numQuantScales`.

This would also work for output bias/scale. There is no need to initialize them. The user can simply provide NULL as `quantScales` and the library will skip initialization (copying) -- it will simply reserve space for output bias/scale.

atamazov (changes requested):

atamazov:
All, please look at #2045. AFAICS this PR has been written with something like (60.10) in mind.

atamazov:
[Q] Do we want to stay with (60.10.50) or we would like to optimize out some host-to-host copying with (60.10.60)?

atamazov (changes requested):
I highly recommend implementing (60.10.20) from #2045.

src/include/miopen/tensor.hpp
atamazov: @jerryyin @iq136boy Please look at #2045 and at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1625#issuecomment-1483457193.

iq136boy:
Hi @atamazov, I would adjust the APIs based on your [comments](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1625#discussion_r1147905811). I am now modified the quantization solver based on the discussion of last Friday. @JehandadKhan can give you some latest update of the discussion that we did on last Friday.

iq136boy:
> I highly recommend implementing (60.10.20) from #2045.

@atamazov I would discuss with @JehandadKhan about this and get back to asap.

iq136boy:
> @iq136boy @JehandadKhan What is required length of bias/scale used with weights tensor? `(C*K)` or... ?

I think it should be K. @JehandadKhan correct me if it is not.

iq136boy:
@jerryyin Based on the meeting last Friday, this [concers](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1625#issuecomment-1479710288) will be addressed in the new APIs and new fusion conv quantization solvers. I am current working on the solver as requested.

iq136boy:
> Or maybe this feature is for [quantization-aware training (QAT)](https://developer.nvidia.com/blog/achieving-fp32-accuracy-for-int8-inference-using-quantization-aware-training-with-tensorrt/) where the user wants to perform "fake" INT8-quantization on all tensors while staying in FP32 domain?
> 
> UPDATE: NO, this is for inference only for now.

Yes, correct. Based on the meeting on last Friday. It is for inference only.

jerryyin:
@iq136boy I have dismissed my review after last Friday meeting. I'm assuming you will drop API change and instead focus on fusion plan side changes. I will abstain myself from reviewing due to my unfamiliarity with fusion side of world.

atamazov:
@jerryyin It's a pity my friend.

iq136boy:
> @iq136boy I have dismissed my review after last Friday meeting. I'm assuming you will drop API change and instead focus on fusion plan side changes. I will abstain myself from reviewing due to my unfamiliarity with fusion side of world.

@jerryyin Yes. I am current focusing on the fusion plan solver implementation base on the discussion on Friday's meeting.

atamazov:
@iq136boy @JehandadKhan Please note that regardless of the API used, we have to analyze the strategy of allocating/passing bias and scale (see https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2045#issuecomment-1483409134 and below) and **_clearly specify_** the approach we are going to implement (either (60.10), (60.20), (60.30) or something else)

iq136boy:
@junliume @JehandadKhan Based on the discussion in the Inference performance meeting, this feature is no longer a request from Migraphx team. At least not a P1 request for Rocm5.6.

JehandadKhan:
@iq136boy It looks like this PR is not going to make it in ROCm 5.6, if you agree, please update the milestone

JehandadKhan:
Closing since it has been dormant for a while, lets revisit this when other more urgent issues are addressed.

PR #1655, opened 2022-07-27
[Enhancement] Tensor op changes when beta is zero

muralinr:
Added Tensor op changes when beta is zero.  Changes are limited to one file only.


junliume:
@atamazov  FYI

junliume (approved):

PR #1662, opened 2022-08-01
    [LWPTUNA-63] : Log MIOpenDriver commands when Find API is called

bghimireamd:
* Log MIOpenDriver commands when Find API is called
        * Introduced a new environment variable MIOPEN_ENABLE_LOGGING_CMD_FIND
        * added tests to verify the log msg


test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,170 @@
...
+    miopenTensorDescriptor_t desc{};
+    void* data;
+    size_t data_size;
+
+    Tensor(int n, int c, int h, int w)
+    {
+        EXPECT_EQ(miopenCreateTensorDescriptor(&desc), 0);
+        EXPECT_EQ(miopenSet4dTensorDescriptor(desc, miopenFloat, n, c, h, w), 0);
+        data_size = n * c * h * w * sizeof(float);
+        EXPECT_EQ(hipMalloc(&data, data_size), hipSuccess);
```
junliume, line 95: @bghimireamd  Some issues with static checks:

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    ~Tensor()
+    {
+        miopenDestroyTensorDescriptor(desc);
+        hipFree(data);
```
iq136boy, line 101: Try this. Otherwise the OpenCL test will complain. 
#if MIOPEN_BACKEND_OPENCL
            clReleaseMemObject(data);
#elif MIOPEN_BACKEND_HIP
            hipFree(data);
#endif

iq136boy (changes requested):

test/gtest/log_test_helper.hpp
bghimireamd: pushed a new commit to fix the error.

DrizztDoUrden:
I don't like the approach of duplicating arg definitions, but I don't see a cheap way to achieve the goal otherwise. Maybe we should think about how we define arguments in the driver to make the definitions asfe and reusable for the printout.

DrizztDoUrden (approved):
lgtm

bghimireamd:
> I don't like the approach of duplicating arg definitions, but I don't see a cheap way to achieve the goal otherwise. Maybe we should think about how we define arguments in the driver to make the definitions safe and reusable for the printout. Although I don't think it is necessary right now.

@DrizztDoUrden  You mean the args that are being passed around in ConvArgsForMIOpenDriver, LogCmdConvolution and LogCmdFindConvolution? If so, I do see something similar in a struct that packs those args as (https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/conv/tensors.hpp#L34). Do you mean we should have something similar in Driver?

DrizztDoUrden:
No idea. I haven't thought about the actual design. That may be a worthy separate task though.
But in general I am talking about command line arguments, not calls. To describe and store them in a way that may be later used for the printing without duplicate logic. That way when someone changes parsing, printing would not require separate update.

atamazov (changes requested):
Indeed, this kind of logging is missing in `miopenFindConvolutionBackwardDataAlgorithm` and in `miopenFindConvolutionForwardAlgorithm`. But I am afraid that this should be a one liner fix. Please look how logging of driver commands is implemented in `miopenFindConvolutionBackwardWeightsAlgorithm()` and do in the same way. No need to introduce new env vars, for example. Thanks.

junliume:
@bghimireamd could you follow up with the review opinions above? Thanks!

src/logger.cpp
```diff
@@ -48,6 +48,11 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_ENABLE_LOGGING)
 /// Allows to reproduce library use cases using the driver instead of the actual application.
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_ENABLE_LOGGING_CMD)
 
+/// Prints driver command lines into log when Find API is called.
+/// Works from any application which uses the library.
+/// Allows to reproduce library use cases using the driver instead of the actual application.
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_ENABLE_LOGGING_CMD_FIND)
```
atamazov, line 54: [Question] Let's agree on what is Find API for convolutions. Which API calls constitute it, what do you think?

atamazov:
@bghimireamd I appreciate your work on tests (👍 !), but I have some serious doubts if we really need tests for debugging functionality.

@junliume @JehandadKhan What is your opinion?

bghimireamd:
> Indeed, this kind of logging is missing in `miopenFindConvolutionBackwardDataAlgorithm` and in `miopenFindConvolutionForwardAlgorithm`. But I am afraid that this should be a one liner fix. Please look how logging of driver commands is implemented in `miopenFindConvolutionBackwardWeightsAlgorithm()` and do in the same way. No need to introduce new env vars, for example. Thanks.

`miopenFindConvolutionBackwardWeightsAlgorithm()` called `LogCmdConvolution` to log driver command. I introduced `LogCmdConvolutionFind` so that it would be easier to distinguish that the driver command is being logged from one of these  Find API.  

for eg: 

old : 

MIOpen(HIP): Command [**LogCmdConvolution**] ./bin/MIOpenDriver conv -n 100 xxxx

new : 

MIOpen(HIP): Command [**LogCmdFindConvolution**] ./bin/MIOpenDriver conv -n 100 xxx

I do agree that your approach would be an easier fix. Also, `MIOPEN_LOG_FUNCTION` function that is called just before `LogCmdConvolution` in `miopenFindConvolutionBackwardWeightsAlgorithm()` does log `miopenFindConvolutionBackwardWeightsAlgorithm()`.
eg: 
MIOpen(HIP): miopenStatus_t miopenFindConvolutionBackwardWeightsAlgorithm xxxx
                      xxx
MIOpen(HIP): Command [**LogCmdConvolution**] ./bin/MIOpenDriver conv -n xxxx

So user can find out if the log of driver command is coming from one of the find API by just looking one line above. Should we just go with one liner approach then?

src/logger.cpp
bghimireamd: Regarding which API calls constitute it? If this env variable is enabled the driver command for find APIs( `miopenFindConvolutionBackwardDataAlgorithm`, `miopenFindConvolutionForwardAlgorithm` and `miopenFindConvolutionBackwardWeightsAlgorithm`) is logged. 

eg: 
`MIOpen(HIP): Command [LogCmdFindConvolution] ./bin/MIOpenDriver conv -n 100 xxx`

src/include/miopen/convolution.hpp
```diff
@@ -73,6 +73,27 @@ using ExtraKernelArgs = std::tuple<int /*N*/,
...
+};
+
+void LogCmdConvolution(const miopenTensorDescriptor_t& xDesc,
+                       const miopenTensorDescriptor_t& wDesc,
+                       const miopenConvolutionDescriptor_t& convDesc,
+                       const miopenTensorDescriptor_t& yDesc,
+                       const ConvDirection& conv_dir,
+                       bool is_immediate);
+
+void LogCmdFindConvolution(const miopenTensorDescriptor_t& xDesc,
```
atamazov, line 90: These global declarations are used only in tests. Can you please copy declarations directly into tests and remove them from the public header?

src/logger.cpp
atamazov: I do not think we need dedicated `MIOPEN_ENABLE_LOGGING_CMD_FIND` (unless there is some use for it right now).

JehandadKhan: I think it is useful since the Find call results in kernels being invoked. If there is a bug and the kernel is crashing during the Find Stage then its difficult to infer which kernel is actually causing the crash since the Driver command only gets logged when  the actual convolution is run. 

With a separate environment variable we can also detect if the framework is called find over and over and the count that we would get for the actual convolution operations would be accurate as well

JehandadKhan:
> @bghimireamd I appreciate your work on tests (👍 !), but I have some serious doubts if we really need tests for debugging functionality.
> 
> @junliume @JehandadKhan What is your opinion?

I dont think there is any harm in being through.

src/include/miopen/convolution.hpp
bghimireamd: Actually `LogCmdConvolution` and `LogCmdFindConvolution` are used in [convolution_api.cpp](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/LWPTUNA-63/src/convolution_api.cpp#L420). I had to place them in header file so that I could test them. I was not sure where would be the best header file to place for these function declaration. So, I just placed it inside convolution.hpp. I am open to suggestions :)

src/logger.cpp
atamazov: @JehandadKhan Sounds reasonable, especially the latter.

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,196 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include "miopen/logger.hpp"
+#include <gtest/gtest.h>
+#include <miopen/convolution.hpp>
+
+namespace Env {
```
atamazov, line 30: Everything in the `Env` namespace is expected to be related to the environment. Let's avoid duplication: either remove namespace or rename functions to something like `SetVariable` and `UnsetVariable`

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,196 @@
...
+#ifdef _WIN32
+    std::string env_var(name + "=" + value);
+    ret = _putenv(env_var.c_str());
+#else
+    ret = setenv(name.c_str(), value.c_str(), 1);
+#endif
+    EXPECT_EQ(ret, 0);
+}
+
+void unSetEnvironmentVariable(const std::string& name)
```
atamazov, line 45: "unset" is a single word in this case.

test/gtest/log_test_neg.cpp
```diff
@@ -0,0 +1,36 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include "log_test_helper.hpp"
```
atamazov, line 26: Do you plan to use this header in several tests? If not, then we do not need header at all. Otherwise more reviewing is needed (e.g. we have to make header compliant with some basic modular programming principles).

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,196 @@
...
+ *
+ *******************************************************************************/
+#include "miopen/logger.hpp"
+#include <gtest/gtest.h>
+#include <miopen/convolution.hpp>
+
+namespace Env {
+
+void setEnvironmentVariable(const std::string& name, const std::string& value)
+{
```
atamazov, line 33: Non-template functions should not be defined in headers.

atamazov (changes requested):

test/gtest/log_test_neg.cpp
atamazov: Oops, I see.

src/include/miopen/convolution.hpp
atamazov: @bghimireamd 
> I am open to suggestions :)

Here is the suggestion: 
> Can you please copy declarations directly into tests and remove them from the public header?

atamazov: You insert copies of declarations of `LogCmdFindConvolution` and `LogCmdConvolution` into `test/gtest/log_test_helper.hpp`

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,196 @@
...
+         false);
+
+    // get the captured string
+    std::string str = capture_cerr.getString();
+    // now do the assertions
+    if(set_env)
+        ASSERT_TRUE(isSubStr(str, sub_str)) << "str     : " << str << "str_sub : " << sub_str;
+    else
+        ASSERT_FALSE(isSubStr(str, sub_str)) << "str     : " << str << "str_sub : " << sub_str;
+}
```
atamazov, line 196: Actually, AFAICS this file should contain only declarations of:
```
TestLogFun
envConv
envFindConv
logConv
logFindConv
```
The definitions of all entities shall be be moved to log_test_helper.cpp.

atamazov (changes requested):

atamazov:
@JehandadKhan 
> I dont think there is any harm in being thorough.

I'm just trying to balance efforts and the benefits they bring ))

src/include/miopen/convolution.hpp
bghimireamd: modified code as per request.

test/gtest/log_test_helper.hpp
bghimireamd: I removed the `Env` namespace and just placed `setEnvironmentVariable` and `unSetEnvironmentVariable`. I was thinking if `setEnvironmentVariable` would be more descriptive then `SetVariable` :)

test/gtest/log_test_helper.hpp
bghimireamd: I was thinking if `unSetEnvironmentVariable` would be more descriptive.

test/gtest/log_test_helper.hpp
bghimireamd: modified code as per request.

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,196 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2017 Advanced Micro Devices, Inc.
```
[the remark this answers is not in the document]
bghimireamd: modified code as per request.

test/gtest/log_test_helper.hpp
bghimireamd: modified code as per request.

junliume (approved):

PR #1663, opened 2022-08-01
PerfDB entry fallback

JehandadKhan:
This PR adds a method `AltSolverDbId` to the solver interface. In the event that a solver does not have a perf db entry, this allows a fallback entry which may be loaded instead. This also allows using perf db entries from non-fused solvers to be used for fused solvers where a specific entry does not exist.


JehandadKhan:
@DrizztDoUrden Can you please go over the suggestion and provide some early feedback.

averinevg (commented):
I think working with DB from several units is not the best idea.

Maybe it is better to add `AltSolverDbId()` for such cases. It may be empty by default. If It is not empty, then we use its value. Or it can be used as a second possible option in case of config loading failure. However, in this case you need to do something with the loading of missing fields into configs. Current implementation does not offer such feature.

On the other hand we can move all DB iteraction into solvers.

Also it seems that you need base class for PerfDb for this PR because of using another DB implementation in tests/solver.cpp. It's in my todo list.

DrizztDoUrden:
I think you should just override this method: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/492eea5934ae0dbf9b36195c2a5e6bede9f6215a/src/include/miopen/solver.hpp#L83
If I understand correctly, the PR change is identical in functionality but also moves the `db.Load` call inside the solver and duplicates it. So I'd rather override id getter to return id of whichever solver you want for such cases.

JehandadKhan:
> I think you should just override this method:
> 
> https://github.com/ROCmSoftwarePlatform/MIOpen/blob/492eea5934ae0dbf9b36195c2a5e6bede9f6215a/src/include/miopen/solver.hpp#L83
> 
> If I understand correctly, the PR change is identical in functionality but also moves the `db.Load` call inside the solver and duplicates it. So I'd rather override id getter to return an id of whichever solver you want for such cases.

The challenge is that its not simply overriding the solver name, there is some action involved first. This means that the solver would only want to override the solver name when there is no exact match, however, if there is a match then we do not want to override it. 

If the preference here is that we should move this logic to `find_solution.hpp` (to avoid DB access in the solver) then I can do that and make progress. 

@DrizztDoUrden and @averinevg  please confirm if that is what you suggest

JehandadKhan:
@DrizztDoUrden and @averinevg  Ping for response

averinevg:
> The challenge is that its not simply overriding the solver name, there is some action involved first. This means that the solver would only want to override the solver name when there is no exact match, however, if there is a match then we do not want to override it.
> 
> If the preference here is that we should move this logic to `find_solution.hpp` (to avoid DB access in the solver) then I can do that and make progress.

I suggest doing the following:
- Solver may have alternative `DbId` and returns it by `AltSolverDbId()`, otherwise empty string is returned.
- Alternative `DbId` should be used if there is no exact match.
- All logic should be implemented in `find_solution.hpp`.

JehandadKhan:
@averinevg and @DrizztDoUrden Please review

src/include/miopen/solver.hpp
```diff
@@ -82,6 +82,15 @@ struct SolverBase
     /// overriden to keep the name to avoid DB corruption.
     virtual const std::string& SolverDbId() const = 0;
 
+    // In some instances ( particularly fusions) the fused solver might like to
+    // fallback to the non-fused variant for performance parameters, this information
+    // is returned via AltSolverDbId
```
averinevg, line 87:
```suggestion
    /// In some instances ( particularly fusions) the fused solver might like to
    /// fallback to the non-fused variant for performance parameters, this information
    /// is returned via AltSolverDbId
```

src/include/miopen/find_solution.hpp
```diff
@@ -82,6 +82,17 @@ auto FindSolutionImpl(
...
+            {
+                MIOPEN_LOG_I("Perf Db: alternate record loaded: " << s.AltSolverDbId());
+                if(s.IsValidPerformanceConfig(context, config))
+                {
+                    return s.GetSolution(context, config);
+                }
+                MIOPEN_LOG_WE("Invalid alternate record loaded from Perf Db: "
+                              << s.AltSolverDbId() << ": " << config
+                              << ". Performance may degrade.");
+            }
```
averinevg, line 95: @JehandadKhan You should also add similar code in `AnySolver::GetPerfCfgParams()`

src/include/miopen/find_solution.hpp
```diff
@@ -82,6 +82,17 @@ auto FindSolutionImpl(
                 MIOPEN_LOG_WE("Invalid config loaded from Perf Db: "
                               << s.SolverDbId() << ": " << config << ". Performance may degrade.");
             }
+            else if(!s.AltSolverDbId().empty() && db.Load(context, s.AltSolverDbId(), config))
+            {
+                MIOPEN_LOG_I("Perf Db: alternate record loaded: " << s.AltSolverDbId());
```
averinevg, line 87:
```suggestion
                MIOPEN_LOG_I2("Perf Db: alternate record loaded: " << s.AltSolverDbId());
```

src/include/miopen/find_solution.hpp
averinevg, line 95: @JehandadKhan You may also generalize this code do avoid duplication for first and second DB loading, but I don't insist.

averinevg (changes requested):

averinevg:
@JehandadKhan Please change the description

src/include/miopen/find_solution.hpp
JehandadKhan: I suggest we keep the higher log level, so that the developer is aware that such a heuristic is functioning

src/include/miopen/solver.hpp
JehandadKhan: I am not sure what is the advantage of the suggested change.

JehandadKhan:
@averinevg CI passes and reviews have been addressed

src/include/miopen/any_solver.hpp
```diff
@@ -263,6 +263,17 @@ struct AnySolver
...
+            else if(!value.AltSolverDbId().empty() && db.Load(ctx, value.AltSolverDbId(), config))
+            {
+                MIOPEN_LOG_I("Perf Db: alternate record loaded: " << value.AltSolverDbId());
+                if(value.IsValidPerformanceConfig(ctx, config))
+                {
+                    return config.ToString();
+                }
+                MIOPEN_LOG_WE("Invalid alternate record loaded from Perf Db: "
+                              << value.AltSolverDbId() << ": " << config
+                              << ". Performance may degrade.");
```
averinevg, line 275: I think `"Perf Db:"` prefix should be added here for consistency.
Do you need to print config here? It is not printed for the first loading from Db. Also do we need to print warning about performance in this case?

averinevg (changes requested):

src/include/miopen/any_solver.hpp
JehandadKhan: I have added the prefix and reduced the log level, printing the config can be beneficial for debugging.

averinevg (approved):
LGTM

junliume (approved):

Issue #1670, opened 2022-08-03
[pooling][2d] Asymmetric configs fail validation when padding != 0x0


atamazov:
Unlike #1660, this is not a regression caused by #1381. It is likely that the issue persists for a long time.
```bash
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 1 2 --strides 1 1 --pads 1 0
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 1 2 --strides 1 1 --pads 1 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 1 2 --strides 1 2 --pads 1 0
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 1 2 --strides 1 2 --pads 1 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 1 2 --strides 2 1 --pads 1 0
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 1 2 --strides 2 1 --pads 1 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 1 2 --strides 2 2 --pads 1 0
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 1 2 --strides 2 2 --pads 1 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 2 1 --strides 1 1 --pads 0 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 2 1 --strides 1 1 --pads 1 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 2 1 --strides 1 2 --pads 0 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 2 1 --strides 1 2 --pads 1 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 2 1 --strides 2 1 --pads 0 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 2 1 --strides 2 1 --pads 1 1
./bin/test_pooling2d --float --verify_indices 0 --input 1 4 4 4 --lens 2 1 --strides 2 2 --pads 0 1
[...]

PR #1659, opened 2022-07-28
[SWDEV-345750] Fix Tensorflow pooling test failure

DrizztDoUrden:
This fixes a typo in the original PR causing incorrect values to be passed for any asymmetric parameter combinations in one of backward solvers.

Issue #1642, opened 2022-07-22
[Tensorflow] observed kernel_tests:pool_test_gpu failure after #1296


junliume:
It seems that PR #1296 has caused this issue in Tensorflow Unit Test Failure

**[Symptom]**
```
 # bazel test //tensorflow/python/kernel_tests:pool_test_gpu
```
shows that the following test has failed:
```
 [  FAILED  ] PoolingTest.testGradient1D
```
However, none of the MIOPEN environmental variables seem to be effective in the log output. The only info available in the log is:
```
 AssertionError: 0.500274658203125 not less than 0.01
```
**[Steps to Reproduce]:**
```
alias ptdrun=' sync && sudo /sbin/sysctl vm.drop_caches=3 && sudo docker run -it --network=host --device=/dev/kfd --device=/dev/dri --ipc=host --group-add video --cap-add=SYS_PTRACE --security-opt seccomp=unconfined -v /data:/data --shm-size=64G'
	
ptdrun compute-artifactory.amd.com:5000/rocm-plus-docker/framework/compute-rocm-dkms-no-npi-hipclang:10405_ubuntu_py3_tensorflow_develop-upstream-QA-rocm53_cf1e7e6a
	
cd tensorflow/
	
bazel test //tensorflow/python/kernel_tests:pool_test_gpu

**Comment by junliume** [comment]
Here are the collected commands:
```

./bin/MIOpenDriver pool -M 0 -n 2 -c 2 -H 1 -W 5 -y 1 -x 1 -p 0 -q 0 -v 1 -u 1 -m avg -F 1 -t 1
./bin/MIOpenDriver pool -M 0 -n 2 -c 2 -H 1 -W 5 -y 1 -x 2 -p 0 -q 0 -v 1 -u 1 -m avg -F 1 -t 1
./bin/MIOpenDriver pool -M 0 -n 2 -c 2 -H 1 -W 5 -y 1 -x 2 -p 0 -q 0 -v 1 -u 2 -m avg -F 1 -t 1
./bin/MIOpenDriver pool -M 0 -n 1 -c 1 -H 1 -W 4 -y 1 -x 1 -p 0 -q 0 -v 1 -u 1 -m avg -F 1 -t 1
./bin/MIOpenDriver pool -M 0 -n 1 -c 1 -H 1 -W 4 -y 1 -x 2 -p 0 -q 0 -v 1 -u 1 -m avg -F 1 -t 1
[...]

PR #1671, opened 2022-08-03
[tests][pooling][2d] Added asymmetric tests (regression test for #1642) and W/A for #1670. [tests][general] Added support for multiple datasets.

atamazov:
- Added generic support for multiple datasets (--dataset N) in tests.
- test_pooling2d:
  - Added dataset 1 with asymmetric configs.
  - Added W/A for #1670.
- Added test_pooling2d_asymmetric (which is basically `test_pooling2d --dataset 1 --limit 0`).
  - This is regression test for #1642 (that is missing in #1659).

Manually tested on gfx906 and gfx1030 with and without #1659 - results as expected.

@junliume @DrizztDoUrden `urgency_high` as this is supplement to the blocker PR.


junliume (approved):

atamazov:
@junliume Thank you!

@JehandadKhan @pfultz2 @DrizztDoUrden @shurale-nkn @asroy Please notice the new `--dataset N` option. The new functionality is supported in the base class, and can be used in any of our tests, if/when necessary.

PR #1619, opened 2022-07-04
[SWDEV-306318][WORKAROUND][gfx908] Disable ConvAsmImplicitGemmGTCDynamicFwdXdlops

shaojiewang:
w/a swd306318:
kernel produce wrong result using:
```
MIOpenDriver conv -n 16 -c 76 -H 9 -W 9 -k 32 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 -w 1
```
As kernel was developed about 2 years ago and I was not the developer, I need more time to identity the faulty code block. In order to move the test process forward, I give a w/a at this moment and I will continue checking the root cause.

As MIOpen has nhwc+transpose kernel as backup, it will not broke anything by disabling this solver.

PR #1675, opened 2022-08-05
[SWDEV-306318][WORKAROUND][gfx908] Limit WA PR:1619 by only ignore 1x1 conv where C is not 8x

carlushuang:
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1619 originally try to fix SWDEV-306318, but introduced large performance regression.
 This PR try to narrow down the non-applicable range, only when 1x1 conv, C is not multiple of 8, this solver is not applicable. This is true for both fp32 and fp16

cc @atamazov

### DIFF
--- a/src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp
@@ -1504,11 +1504,6 @@ bool ConvAsmImplicitGemmGTCDynamicFwdXdlops::IsApplicable(const ConvolutionConte
     if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_FWD_GTC_XDLOPS{}))
         return false;
 
-#if WORKAROUND_SWDEV_306318
-    if(!miopen::IsEnabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_FWD_GTC_XDLOPS{}))
-        return false;
-#endif
-
     const auto device_name = ctx.GetStream().GetDeviceName();
     if(device_name != "gfx908")
         return false;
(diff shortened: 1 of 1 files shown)


DrizztDoUrden (approved):
lgtm

src/solver/conv_asm_implicit_gemm_gtc_fwd.cpp
```diff
@@ -1536,6 +1531,11 @@ bool ConvAsmImplicitGemmGTCDynamicFwdXdlops::IsApplicable(const ConvolutionConte
         return false;
     }
 
+#if WORKAROUND_SWDEV_306318
+    if((ctx.kernel_size_h == 1) && (ctx.kernel_size_w == 1) && (ctx.n_inputs % 8 != 0))
+        return false;
```
atamazov, line 1536: It is recommended to insert
```C++
if(!miopen::IsEnabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_FWD_GTC_XDLOPS{}))
```
just above `return false`. This would allow explicit enabling the solver for `1x1 C%8 != 0` configs for triaging, without rebuilding the library.

qianfengz: @atamazov Do you mean with the current commits,  `1x1 c%8 !=0` configs could not be tested, even after the kernels being improved?   and by inserting 
`if(!miopen::IsEnabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_FWD_GTC_XDLOPS{}))` 
 we are able to test `1x1 c%8 !=0` by using 
`export MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_FWD_GTC_XDLOPS=1` ? 

So my understand is that the env var  `MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_FWD_GTC_XDLOPS` should have three states:
1) Disabled ( defined from env and have false value) 
2) Enabled ( defined from env and have true value)
3) Neither Disabled nor Enabled ( not defined from env) 

right?

atamazov: @qianfengz Yes, and in the latter case both IsEnabled and IsDisabled return false. This is useful in cases like this one.

qianfengz: > It is recommended to insert
> 
> ```c++
> if(!miopen::IsEnabled(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_FWD_GTC_XDLOPS{}))
> ```
> 
> just above `return false`. This would allow explicit enabling the solver for `1x1 C%8 != 0` configs for triaging, without rebuilding the library.

@carlushuang  please take a look at the atamazov recommendation, so we will approve after your update

Issue #1681, opened 2022-08-09
Performance drops after PR #1634


atamazov:
> @junliume This PR fixes Winograd applicability for grouped convolutions; tested on gfx906 using `MIOPEN_DEBUG_FIND_ONLY_SOLVER` environment for the following Winograd solvers: `ConvBinWinogradRxSf2x3` and `ConvBinWinogradRxSf3x2`.
> 
> Was the FindDB that has been used during perf tests generated after this PR? If it was generated before this PR, but after #1533, where bug was introduced, then Winograd solvers may not be in FindDB for grouped convolutions.

Hi @Slimakanzer thank you for getting back to me during your vacation. The following cases have confirmed performance regressions right after this PR (ranked by the impacted lvel):
```
python3 micro_benchmarking_pytorch.py --network inception_v3  --batch-size 128  --iterations 10 --dataparallel --device_ids 0  --fp16 1 && \
python3 micro_benchmarking_pytorch.py --network alexnet  --batch-size 1024  --iterations 10 --dataparallel --device_ids 0  --fp16 1 && \
python3 micro_benchmarking_pytorch.py --network SqueezeNet  --batch-size 128  --iterations 10 --dataparallel --device_ids 0  --fp16 1 && \
python3 micro_benchmarking_pytorch.py --network densenet121  --batch-size 64  --iterations 10 --dataparallel --device_ids 0  --fp16 1 && \
python3 micro_benchmarking_pytorch.py --network alexnet  --batch-size 64  --iterations 10 --dataparallel --device_ids 0  --fp16 0 && \
python3 micro_benchmarking_pytorch.py --network resnet50  --batch-size 128  --iterations 10 --dataparallel --device_ids 0  --fp16 1 && \
python3 micro_benchmarking_pytorch.py --network resnet152  --batch-size 64  --iterations 10 --dataparallel --device_ids 0  --fp16 1 && \
python3 micro_benchmarking_pytorch.py --network vgg16  --batch-size 128  --iterations 10 --dataparallel --device_ids 0  --fp16 1 && \
python3 micro_benchmarking_pytorch.py --network vgg19  --batch-size 128  --iterations 10 --dataparallel --device_ids 0  --fp16 1 && \
python3 micro_benchmarking_pytorch.py --network densenet121  --batch-size 64  --iterations 10 --dataparallel --device_ids 0  --fp16 0
```
[...]

PR #1634, opened 2022-07-13
[BugFix] Winograd ConvBinWinogradRxSf2x3 applicability

Slimakanzer:
This PR fixes incorrect `ConvBinWinogradRxSf2x3`  applicability introduced in #1533.

PR #1682, opened 2022-08-09
[WORKAROUND][HOTFIX] W/A for #1681, "Performance drops after PR #1634"

atamazov:
This is expected to work around issue #1681.

@junliume `urgency_blocker`, `performance, bug`. `workaround`. Can you please also double check if this resolves the performance issue, thank you!

### DIFF
--- a/src/solver/conv_winoRxS.cpp
+++ b/src/solver/conv_winoRxS.cpp
@@ -41,6 +41,17 @@
 
 #include <tuple>
 
+// ConvBinWinoRxS<2,3> is intended to handle group convolutions, but
+// it is able to work with non-group convolutions and shows quite good performance,
+// comparable with ConvBinWinogradRxSf2x3g1. Due to the issue in #1533 (fixed in #1634)
+// AND some inaccuracy (jitter) of benchmarking, it is often shows the "best" performance
+// with non-group convolutions, and therefore unexpected records where ConvBinWinoRxS<2,3>
+// is the "best" leaked into find-db. That is why disabling ConvBinWinoRxS<2,3> for non-group
+// convolutions leads to performance drops. Let's enable ConvBinWinoRxS<2,3> for non-group
+// in order to quickly W/A the perf issue. When find-db fix is ready,
+// we will keep ConvBinWinoRxS<2,3> for group convolutions only.
+#define WORKAROUND_ISSUE_1681 1
+
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3_PERF_VALS)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3_G1)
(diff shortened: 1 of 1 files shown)


junliume:
@atamazov @Slimakanzer my tests show that the perf regressions on the 10 selective cases have "recovered" with this workaround. We need the Winograd team's review on the changes.

src/solver/conv_winoRxS.cpp
```diff
@@ -513,14 +524,22 @@ static bool IsApplicableBase(const ConvolutionContext& params)
...
-        return false;
-    if(IS3X2 && params.kernel_stride_w == 2) // f3x2 stride 2 not implemented yet
-        return false;
-    if(IS2X3 && params.group_counts == 1)
-        return false;
+    if(IS2X3)
+    {
+        if(miopen::IsDisabled(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3{}))
+            return false;
+#if !WORKAROUND_ISSUE_1681
```
junliume, line 531: @Slimakanzer is it so that group convo should not be disabled with 3x2? Is it by design or by coincidence that it was supported?

Slimakanzer: `ConvBinWinogradRxSf3x2` support grouped convolution and should not be disabled for cases with `group_counts == 1` as it also support non-grouped convolutions. `ConvBinWinogradRxSf2x3` solver can also support grouped and non-grouped convolutions, but for non-grouped convolutions we know best tuning parameters, and to reduce tuning time, we introduced `ConvBinWinogradRxSf2x3g1` solver with no performance tuning option.

Slimakanzer (approved):
@atamazov 
Thank you, LGTM!

src/solver/conv_winoRxS.cpp
```diff
@@ -41,6 +41,17 @@
...
 
+// ConvBinWinoRxS<2,3> is intended to handle group convolutions, but
+// it is able to work with non-group convolutions and shows quite good performance,
+// comparable with ConvBinWinogradRxSf2x3g1. Due to the issue in #1533 (fixed in #1634)
+// AND some inaccuracy (jitter) of benchmarking, it is often shows the "best" performance
+// with non-group convolutions, and therefore unexpected records where ConvBinWinoRxS<2,3>
+// is the "best" leaked into find-db. That is why disabling ConvBinWinoRxS<2,3> for non-group
+// convolutions leads to performance drops. Let's enable ConvBinWinoRxS<2,3> for non-group
+// in order to quickly W/A the perf issue. When find-db fix is ready,
+// we will keep ConvBinWinoRxS<2,3> for group convolutions only.
```
atamazov, line 52: ⚓

src/solver/conv_winoRxS.cpp
junliume: Since this is a workaround I would hope there will be a full-blown implementation to follow :) I will try to merge this one first.

atamazov:
> @atamazov @Slimakanzer my tests show that the perf regressions on the 10 selective cases have "recovered" with this workaround. We need the Winograd team's review on the changes.

Good! Have you seen performance regression on both MI50 (60 CUs) and MI60 (64 CUs)? I am asking because I do not see "unexpected" records in gfx906_60.HIP.fdb.txt, while gfx906_64.HIP.fdb.txt has lost of them. I guess that MI50 is unaffected by the performance regression.

**_Please note that find-db for gfx906/60 and /64 remain unchanged for a long time_** (10 and 14 months respectively). However IIRC we update databases before almost each release. Maybe we update databases only in release branches; if this is the case, then the missing step in the process is merging updated databases from the release branches back to develop (I think the best time for this is immediately after the release). @JehandadKhan What is your opinion?

src/solver/conv_winoRxS.cpp
atamazov, line 52: ⚓

src/solver/conv_winoRxS.cpp
atamazov: @junliume Yes; please see added comment here: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1682#discussion_r941903884

junliume (approved):

atamazov:
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1682#issuecomment-1210005719 moved to #1681 and hidden.

atamazov:
@junliume Now I can claim that this W/A restores perf drops on MI60 (confirmed by experiment, see https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1681#issuecomment-1214207069). Let's push this through CI and merge.

atamazov:
The full blown fix will come later. I think that this W/A is enough for release.

PR #1698, opened 2022-08-15
Copyright updates part #1

alexandraBara:
Added copyrights to missing files

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 #
 # MIT License
 #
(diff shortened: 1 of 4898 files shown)


junliume:
@alexandraBara @JehandadKhan 
When the license headers are missing, sure I support adding them 

However, I highly doubt if revising 2017 to 2022 is really necessary.

junliume:
@alexandraBara  could you rework on these three PRs by only adding copyright to the files which have missed it?

PR #1714, opened 2022-08-20
[CI] Update dockerfile ROCm version and bump MIOpen version

junliume:
@atamazov @shurale-nkn @JehandadKhan  please review:

1. Update docker ROCm version to ROCm 5.2.3 (latest public release);
2. Bump MIOpen version to 2.19
3. Make a few CI stages "Any GPU" to utilize CI resources more efficiently


Dockerfile
```diff
@@ -10,13 +10,13 @@ RUN dpkg --add-architecture i386
 # unless MLIR library is incompatible with current ROCm.
 
 RUN if [ "$USE_MLIR" = "ON" ] ; \
```
shurale-nkn, line 12: Maybe we will get rid of this meaningless condition already

shurale-nkn (changes requested):

Dockerfile
junliume: @jerryyin I think MLIR wishes to retain the flexibility of using a different ROCm version? However indeed I would prefer to consolidate them.

jerryyin: This branch condition was there because in the first MLIR release it is attached to a specific ROCm release.  (The enablement took our exclusive attention and supporting a newer ROCm become secondary priority.) Because of this, MIOpen is blocked by MLIR's old ROCm version. This condition make it such that MIOpen fall back to older ROCm release when testing MLIR related stages.

Agreed here that we can get rid of this condition because:
 - MLIR is always up to date with latest ROCm version, the risk of falling behind is minimal
 - MLIR has been tested in (almost) all MIOpen stage, this fallback will cause MIOpen to also effectively fallback to a previous ROCm (making the version update meaningless)

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
```
atamazov, line 1: [Q] Which Ubuntu version is installed on CI machines? Is it safe to run ubuntu 20 docker on ubuntu 18 machine?

Jenkinsfile
```diff
@@ -510,19 +513,22 @@ pipeline {
             }
         }
         stage("Smoke Fp32") {
+            options {
+                retry(2)
+            }
```
atamazov, line 518: I guess this is already in develop. Please update the branch.

atamazov (dismissed):
No objections

Jenkinsfile
junliume: Actually it's an addition (retry on parallel stages instead of on individual ones).

Dockerfile
junliume: That's indeed a TODO item after changes in this PR looks stable for the CI: we need to update CI machines to 20.04.

Jenkinsfile
atamazov: @junliume If "instead" then shall we remove retries in individual stages?

shurale-nkn (approved):
LGTM

JehandadKhan (approved):
LGTM

PR #1727, opened 2022-08-31
[bg/LWPMIOPEN-40] : Dump out inpput/output and weights tensor when a NaN/Inf occurs in Convolution

bghimireamd:
* Overloading Handle::ReadTo to take raw pointer.
* Added API DumpTensorToFileFromDevice to dump tensor when nan is encountered and env variable MIOPEN_DUMP_TENSOR_PATH is set.
* Addtest test for verification.


driver/conv_verify.hpp
```diff
@@ -119,7 +119,7 @@ void RunBackwardWeightsCPUVerify(std::vector<Tcheck_>& dwei_host,
         assert(stride_h == 1);
         assert(stride_w == 1);
 
-        std::fill(dwei_host.begin(), dwei_host.end(), (static_cast<Tcheck_>(0));
+        std::fill(dwei_host.begin(), dwei_host.end(), static_cast<Tcheck_>(0));
```
junliume, line 122: Good catch!

junliume (approved):

PR #1533, opened 2022-04-28
[enhancement] ConvBinWinogradRxSf2x3 / f3x2 solvers merge

johnny-keker:
This PR merges almost similar `ConvBinWingradRxSf2x3` and `ConvBinWingradRxSf3x2` solvers into one `ConvBinWinoRxS` solver, which reduces the amount of duplicated code. In order to provide similar functionality, new solver uses templates:

```
    ConvBinWinoRxS<2, 3>;
    ConvBinWinoRxS<3, 2>;
```

@atamazov please review

PR #1735, opened 2022-09-06
[Proactive][Fix][Performance] Winograd applicability for WrW direction

Slimakanzer:
WrW direction is only supported by grouped Winograd kernels (`ConvBinWinogradRxSf2x3` solver) and must be disabled for `ConvBinWinogradRxSf2x3g1` solver.

Perf/Find DB are not affected, but they will be affected once the full blown fix for #1681 is implemented.

### DIFF
--- a/src/solver/conv_winoRxS.cpp
+++ b/src/solver/conv_winoRxS.cpp
@@ -531,7 +531,7 @@ bool ConvBinWinoRxS<Winodata, Winofilter>::IsApplicable(const ConvolutionContext
         if(miopen::IsDisabled(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3{}))
             return false;
 #if !WORKAROUND_ISSUE_1681
-        if(params.problem.group_counts == 1)
+        if(params.problem.group_counts == 1 && !params.problem.direction.IsBackwardWrW())
             return false;
 #endif
     }
@@ -913,6 +913,10 @@ bool ConvBinWinogradRxSf2x3g1::IsApplicable(const ConvolutionContext& params) co
 {
     if(miopen::IsDisabled(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3_G1{}))
         return false;
+#if !WORKAROUND_ISSUE_1681
+    if(params.problem.direction.IsBackwardWrW())
+        return false;
+#endif
     return IsApplicableBase(params) && params.problem.group_counts == 1;
 }
 


Slimakanzer:
@atamazov pls review

atamazov:
@Slimakanzer 
> WrW direction is only supported by grouped Winograd

What are the side effects of the issue? This is important to know to categorize the patch.

atamazov:
@Slimakanzer Actually this is not a bug fix right now. This is a fix of #1682 that proactively resolves the future issue when WORKAROUND_ISSUE_1681 will be disabled.

Actual problem existed in our code even before #1533. What happens: Non-grouped kernels is unable to handle WrW by design, but G1 solver is not disabled. For non-grouped WrW convolution the G1 solver selects proper (grouped) kernel, but tuning is not engaged (because G1 solver is not tunable).

The possible side effects of the issue is: the performance of Winograd 2x3 for non-grouped WrW convolutions may suffer.

atamazov (approved):
LGTM!

DrizztDoUrden (approved):
LGTM

atamazov:
@junliume This PR is currently NFC, but it blocks bull-blown fix for #1681. Please merge this ASAP, if you can (or I will be forced to merge this into my PR and close this one -- this is acceptable scenario, but requires some more work). Thanks!

junliume (approved):

junliume:
> @junliume This PR is currently NFC, but it blocks bull-blown fix for #1681. Please merge this ASAP, if you can (or I will be forced to merge this into my PR and close this one -- this is acceptable scenario, but requires some more work). Thanks!

I am mostly concerned about perf impacts, but since it is a functional fix, I think we just need it earlier then.

atamazov:
@junliume This PR is No Functional Changes (until WORKAROUND_ISSUE_1681 is removed). Thanks!

src/solver/conv_winoRxS.cpp
```diff
@@ -913,6 +913,10 @@ bool ConvBinWinogradRxSf2x3g1::IsApplicable(const ConvolutionContext& params) co
 {
     if(miopen::IsDisabled(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3_G1{}))
         return false;
+#if !WORKAROUND_ISSUE_1681
+    if(params.problem.direction.IsBackwardWrW())
+        return false;
```
atamazov, line 918: [Potential problem] This would lead to huge WrW perf regressions when WORKAROUND_ISSUE_1681 is disabled with current system perf-db. Reason: Bad tuning values for ConvBinWinogradRxSf2x3 in the current system perf-db.

Example for `./bin/MIOpenDriver conv -x 1 -y 7 -W 17 -H 17 -c 128 -n 128 -k 128 -p 3 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 -F 0 -V 1 -s 0 -w 2 -t 1 -i 6` (excerpt). When g1 solver disabled:
```
MIOpen(HIP): Info [GetValues] =ConvBinWinogradRxSf2x3:1
...
MIOpen(HIP): Info [EvaluateInvokers] Selected: ConvBinWinogradRxSf2x3: ... 50.591, workspace_sz = 0
```
When g1 solver enabled:
```
MIOpen(HIP): Info [EvaluateInvokers] Selected: ConvBinWinogradRxSf2x3g1: ... 1.71519, workspace_sz = 0
```

junliume: @Slimakanzer @atamazov  were the previously enabled user cases just by accident then? If the solver should not be applicable in the beginning.
If it appears to be working nevertheless, should we extend the applicability?

atamazov: @junliume The "f2x3" solver should be applicable for WrW from the beginning. Therefore we do not need the "g1" solver for WrW because f2x3 can do the same job with the same performance even without tuning (and if tuned, it can show better performance for WrW). That is why we wanted to disable "g1" for non-group WrW in the full-blown fix.

However, some tuning values for f2x3 are wrong and result in bad performance. The "g1" solver can easily camouflage the issue with minimal overhead.

I will provide necessary changes in the upcoming "full fix" PR.

> If it appears to be working nevertheless, should we extend the applicability?

Basically, yes, unless we can _reduce host-size overhead by narrowing the applicability_. Of course this should be only done if we 100% sure that some other solver can do the same job with the same or better performance. For the f2x3 and f2x3g1 solvers this is guaranteed programmatically.

atamazov: @junliume @Slimakanzer 
> [Potential problem] This would lead to huge WrW perf regressions when WORKAROUND_ISSUE_1681 is disabled with current system perf-db. Reason: Bad tuning values for ConvBinWinogradRxSf2x3 in the current system perf-db.

Perhaps it was is a kind of glitch. Re-testing in progress...

atamazov: The database issues confirmed. I was unable to reproduce the issue manually due to #1753

PR #1739, opened 2022-09-08
[NFC] convolution.cpp and convolution.ocl clean-up

DrizztDoUrden:
WIP

The point is to remove massive code duplications from the specified files. This would ease find 2.0 internals work and updating other primitives to reuse find-db. Right now, the focus is only on methods related to `find` internals. To stop the propagation of the deprecated code in these files and ease Evgeny's work a bit most of the new methods use `ExecutionContext` with `conv::ProblemDescription` rather than `ConvolutionContext` up to the very deep parts where it becomes too cumbersome without updating the rest of the codebase.
In the next PR `FindConvolution` function would be extended to support new find 2.0 functionality like binary serialization.
As a future work, immediate mode methods should be cleaned up as well. Most of them are just almost identical copies for each direction. Some of them may not require any additional scaffolding.
That has not been done to minimize the number of changes, while keeping the total amount of work sane.

This blocks the next Find 2.0 PR. The draft of that PR has not been created yet.

Fixes UB in test_main (DEV builds only). miopen handle that employs _explicitly created HIP stream_ should be used in the same thread where is has been created. Otherwise we run into weird problems like ["Device mismatch from where this module is loaded"](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1739#issuecomment-1320210095) with DEV builds.


atamazov (commented):
I do not see anything suspicious. Testing for performance regressions is ongoing; stay tuned.

atamazov:
❗ Heads up: Some problems detected; details tomorrow, please do not merge.

atamazov:
Navi21, ROCm 5.2.3, all FP32 and FP16 conv tests failed. Log of FP32 test and the list of configs attached.

- [smoke-fp32-93.sh.hip-clang.7d9bb2911.ddu_conv-cleanup.0.fail.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9706338/smoke-fp32-93.sh.hip-clang.7d9bb2911.ddu_conv-cleanup.0.fail.txt)
- [smoke-fp32-93.sh.hip-clang.7d9bb2911.ddu_conv-cleanup.0.zip](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9706341/smoke-fp32-93.sh.hip-clang.7d9bb2911.ddu_conv-cleanup.0.zip)

atamazov:
@DrizztDoUrden 
> Blocked by https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1791.

Does this mean that this PR is **_not_** ready for reviewing & testing?

DrizztDoUrden:
>>Blocked by https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1791.
>
>Does this mean that this PR is not ready for reviewing & testing?

This has some overlapping changes and it would be easier and more appropriate to merge that PR first as that is inherently more urgent and passing the CI.
This one is still a bit in the works, but the general concept and approach shouldn't change noticeably.

src/ocl/convolutionocl.cpp
```diff
@@ -174,7 +174,7 @@ static inline std::vector<PerfField> FindConvolution(const ExecutionContext& ctx
         const auto id = solver::Id{sol->solution_id};
         CompileSolution(id, ctx, problem);
         results.push_back(
-            {id.GetAlgo(conv::Direction::Forward), id.ToString(), sol->time, sol->workspace_size});
+            {id.GetAlgo(problem.GetDirection()), id.ToString(), sol->time, sol->workspace_size});
```
atamazov, line 177: Re-testing in progress...

atamazov:
[preliminary] @DrizztDoUrden It seems like warm-up is broken, which affects precision of benchmarking in the driver (-w 2)...

atamazov:
@DrizztDoUrden Issue: **_Gemm algo is used instead of Winograd algo, Direct algo is not used_** (during Fwd Find at least). That affects everything including warmup.

Example config: `conv -x 1 -y 7 -W 17 -H 17 -c 128 -n 128 -k 128 -p 3 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 -F 1`

<details><summary><b><i>EvaluateInvokers sequence in the baseline</i></b></summary>

```bash
MIOpen(HIP): Info [EvaluateInvokers] GemmFwdRest: : 6.79936 < 3.40282e+38
MIOpen(HIP): Info [EvaluateInvokers] Selected: GemmFwdRest: : 6.79936, workspace_sz = 1035776
MIOpen(HIP): Info [EvaluateInvokers] ConvBinWinogradRxSf3x2: miopenSp3AsmConv_v21_1_3_gfx10_f3x2_fp32_stride1_group: 2.52284 < 3.40282e+38
MIOpen(HIP): Info [EvaluateInvokers] ConvBinWinogradRxSf2x3: miopenSp3AsmConv_v21_1_3_gfx10_fp32_stride1: 0.94072 < 2.52284
MIOpen(HIP): Info [EvaluateInvokers] ConvBinWinogradRxSf2x3g1: miopenSp3AsmConv_v21_1_3_gfx10_fp32_stride1: 1.21312 >= 0.94072
MIOpen(HIP): Info [EvaluateInvokers] Selected: ConvBinWinogradRxSf2x3: miopenSp3AsmConv_v21_1_3_gfx10_fp32_stride1: 0.94072, workspace_sz = 0
MIOpen(HIP): Info [EvaluateInvokers] ConvDirectNaiveConvFwd: naive_conv_fwd_nchw_float_double_float: 34.9975 < 3.40282e+38
MIOpen(HIP): Info [EvaluateInvokers] Selected: ConvDirectNaiveConvFwd: naive_conv_fwd_nchw_float_double_float: 34.9975, workspace_sz = 0
MIOpen(HIP): Info [EvaluateInvokers] ConvHipImplicitGemmV4R1Fwd: gridwise_convolution_implicit_gemm_v4r1_nchw_kcyx_nkhw_lds_double_buffer: 0.69296 < 3.40282e+38
MIOpen(HIP): Info [EvaluateInvokers] ConvHipImplicitGemmV4R4Fwd: gridwise_convolution_implicit_gemm_v4r4_nchw_kcyx_nkhw: 0.664161 < 0.69296
MIOpen(HIP): Info [EvaluateInvokers] ConvCkIgemmFwdV6r1DlopsNchw: convolution_forward_implicit_gemm_v6r1_dlops_nchw_kcyx_nkhw_prepare/convolution_forward_implicit_gemm_v6r1_dlops_nchw_kcyx_nkhw: 29.6836 >= 0.664161
MIOpen(HIP): Info [EvaluateInvokers] Selected: ConvHipImplicitGemmV4R4Fwd: gridwise_convolution_implicit_gemm_v4r4_nchw_kcyx_nkhw: 0.664161, workspace_sz = 0
```
</details>
<details><summary><b><i>EvaluateInvokers sequence in the branch</i></b></summary>

```
MIOpen(HIP): Info [EvaluateInvokers] GemmFwdRest: : 6.7584 < 3.40282e+38
MIOpen(HIP): Info [EvaluateInvokers] Selected: GemmFwdRest: : 6.7584, workspace_sz = 1035776
MIOpen(HIP): Info [EvaluateInvokers] ConvHipImplicitGemmV4R1Fwd: gridwise_convolution_implicit_gemm_v4r1_nchw_kcyx_nkhw_lds_double_buffer: 0.75952 < 3.40282e+38
MIOpen(HIP): Info [EvaluateInvokers] ConvHipImplicitGemmV4R4Fwd: gridwise_convolution_implicit_gemm_v4r4_nchw_kcyx_nkhw: 0.6632 < 0.75952
MIOpen(HIP): Info [EvaluateInvokers] ConvCkIgemmFwdV6r1DlopsNchw: convolution_forward_implicit_gemm_v6r1_dlops_nchw_kcyx_nkhw_prepare/convolution_forward_implicit_gemm_v6r1_dlops_nchw_kcyx_nkhw: 29.5662 >= 0.6632
MIOpen(HIP): Info [EvaluateInvokers] Selected: ConvHipImplicitGemmV4R4Fwd: gridwise_convolution_implicit_gemm_v4r4_nchw_kcyx_nkhw: 0.6632, workspace_sz = 0
MIOpen(HIP): Info [EvaluateInvokers] GemmFwdRest: : 6.17984 < 3.40282e+38
MIOpen(HIP): Info [EvaluateInvokers] Selected: GemmFwdRest: : 6.17984, workspace_sz = 1035776
```
</details>

Reproduce command for Navi21:
```
$ MIOPEN_FIND_MODE=1 \
MIOPEN_LOG_LEVEL=5 \
./bin/MIOpenDriver conv -x 1 -y 7 -W 17 -H 17 -c 128 -n 128 -k 128 -p 3 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 \
-F 1 -V 0 -s 0 -w 1 -t 1 -i 1 \
|& tee <LOGFILE>
```
Full logs:
- [config01.fm1_ll5_w1.1.log.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9719361/config01.fm1_ll5_w1.1.log.txt)
- [config01.fm1_ll5_w1.2.log.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9719362/config01.fm1_ll5_w1.2.log.txt)

atamazov:
@DrizztDoUrden Let me re-test this. This would take some time, excuse me in advance. Please set urgency label, if it is high.

DrizztDoUrden:
No need yet, I am fixing small stuff

atamazov:
@DrizztDoUrden Is it time to ~die~ re-test?
time-to-die-1_small

DrizztDoUrden:
@atamazov yes, I think so.

atamazov:
@DrizztDoUrden The recent merge from develop was done using outdated commit (~10 days ago). Please pull fresh develop and merge again.

I did the same locally before testing (the merge is trivial)

DrizztDoUrden:
My bad

atamazov:
No problem ;)

atamazov:
All tests passed.

atamazov (approved):
LGTM!

src/execution_context.cpp
```diff
@@ -203,21 +206,37 @@ bool IsHipKernelsEnabled()
...
+        use_asm_kernels = !IsDisabled(MIOPEN_DEBUG_GCN_ASM_KERNELS{}) && ValidateGcnAssembler();
 #ifndef HIP_OC_FINALIZER
-        use_binaries = !miopen::IsDisabled(MIOPEN_DEBUG_AMD_ROCM_PRECOMPILED_BINARIES{});
+        use_binaries = !IsDisabled(MIOPEN_DEBUG_AMD_ROCM_PRECOMPILED_BINARIES{});
 #endif
     }
+    return *this;
+}
+
+ExecutionContext& ExecutionContext::SetupFloats(const conv::ProblemDescription& problem)
```
averinevg: @DrizztDoUrden `ExecutionContext` is also used with activation, pooling and batchnorm solvers, and they have their own `ProblemDescription`

averinevg (changes requested):

src/convolution.cpp
```diff
@@ -336,19 +423,12 @@ ConvolutionDescriptor::WrwGetValidWorkSpaceSizeGemm(const TensorDescriptor& dyDe
...
 {
     MIOPEN_LOG_I("");
 
-    auto ctx = ConvolutionContext{xDesc, wDesc, yDesc, *this, conv::Direction::Forward};
-    ctx.SetStream(&handle);
-    ctx.DetectRocm();
-    ctx.SetupFloats();
-    ctx.do_search             = false;
-    ctx.disable_perfdb_access = true;
+    auto ctx = ctx_.WithTuning(false).WithPerfdbAccess(false);
```
averinevg: @DrizztDoUrden Here is a double copy. Is that how it's intended?

src/fusion.cpp
```diff
@@ -205,11 +205,13 @@ miopenStatus_t FusionPlanDescriptor::GetWorkspaceSizeImmed(Handle& handle,
     {
         if(op->kind() == miopenFusionOpConvForward)
         {
-            auto ptr = std::dynamic_pointer_cast<ConvForwardOpDescriptor>(op);
+            auto& ptr = dynamic_cast<ConvForwardOpDescriptor&>(*op);
```
averinevg: @DrizztDoUrden It is not a pointer anymore, right?

src/convolution.cpp
DrizztDoUrden: The first copy should get elided. I was going to add const there, but it's changed a bit later.

src/fusion.cpp
DrizztDoUrden: Will rename

src/convolution.cpp
DrizztDoUrden: My bad here, will fix.

src/execution_context.cpp
DrizztDoUrden: Will move to `conv::ProblemDescription`

atamazov:
@averinevg @JehandadKhan Please review.

@DrizztDoUrden When reviewing done I will re-test once more.

averinevg (dismissed):

atamazov:
🌀 Testing in progress...

junliume:
@DrizztDoUrden could you track the CI and ensure it passes? it should be an NFC PR but I guess we will wait for @atamazov 's tests too.

atamazov:
## 🌀 Testing DONE, no regressions, good to go!

atamazov:
@averinevg Please review the latest comment && re-approve, thanks!

averinevg (dismissed):
LGTM

DrizztDoUrden:
I don't understand why test_main fails and can't reproduce anywhere I have access to.

atamazov:
@DrizztDoUrden Please send me the failing log and also tell me which Jenkins stage is failing.

junliume:
> @DrizztDoUrden Please send me the failing log and also tell me which Jenkins stage is failing.

> [2022-11-18T13:33:27.625Z] "Device mismatch from where this module is loaded"

Fp32 Hip Debug gfx90a Happened more than once on different nodes

DrizztDoUrden:
It also writes the following all the time:
```
[2022-11-18T13:17:29.157Z] + cat /sys/module/amdgpu/version
[2022-11-18T13:17:29.157Z] cat: /sys/module/amdgpu/version: No such file or directory
```
I don't think it's related to the PR.

JehandadKhan:
@bghimireamd Please review this PR for your awareness.

atamazov:
@junliume @DrizztDoUrden 
> Fp32 Hip Debug gfx90a Happened more than once on different nodes

Does this happen on MI200 only?

src/problem.cpp
```diff
@@ -225,24 +225,10 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
...
-                return conv_desc.ForwardGetWorkSpaceSize(handle, w_desc, y_desc, x_desc);
-            return conv_desc.BackwardDataGetWorkSpaceSize(handle, w_desc, y_desc, x_desc);
-        case miopenProblemDirectionBackwardWeights:
-            if(conv_desc.mode == miopenTranspose)
-                return conv_desc.BackwardWeightsGetWorkSpaceSize(handle, x_desc, y_desc, w_desc);
-            return conv_desc.BackwardWeightsGetWorkSpaceSize(handle, y_desc, x_desc, w_desc);
-        default: MIOPEN_THROW(miopenStatusNotImplemented);
-        }
-    }();
+    const auto conv_problem =
```
atamazov, line 237: This leads to build error at line 335

atamazov (changes requested):
This fails:
```
ctest --output-on-failure -C ./test -R test_conv_trans
```

atamazov (commented):
Please also fix typo https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1503/files#r1042694684

atamazov:
> This fails:
> 
> ```
> ctest --output-on-failure -C ./test -R test_conv_trans
> ```

Specific case:
```
root@morpheus:~/github/MLOpen1/build/ci_2004# ./bin/test_conv2d --verbose --input 100 3 64 64 --weights 3 3 1 1 --pads_strides_dilations 2 2 1 1 1 1 --cmode trans --pmode default --group-count 3 --disable-verification-cache
MIOpen(HIP): Info [get_device_name] Raw device name: gfx906:sramecc+:xnack-
MIOpen(HIP): Info [Handle] stream: 0x2013270, device_id: 0
MIOpen(HIP): Info [GetWorkSpaceSize]
MIOpen(HIP): Info [GetSolutions]
MIOpen(HIP): Info [Measure] ReadonlyRamDb::Prefetch time: 49.2743 ms
MIOpen(HIP): Info [Measure] RamDb::Prefetch time: 1.38711 ms
  what():  /root/github/MLOpen1/src/ocl/convolutionocl.cpp:107: Invalid filter channel number
Aborted (core dumped)
```

junliume:
@DrizztDoUrden  still a WIP for this PR?

atamazov:
@DrizztDoUrden +1 to Jun. Is it time to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1739#issuecomment-1306385932 again? If so, then please merge latest `develop` and remove `ON_HOLD`, thanks.

test/main.cpp
```diff
@@ -294,18 +304,21 @@ struct conv_forward : output_tensor_fixture
...
+            STATUS(miopenDestroy(used_handle));
+
+            if(Profile)
+            {
+                CHECK(time > 0.0);
+            }
+            else
+            {
+                CHECK(time == 0.0);
+            }
```
atamazov: [Recommendation] Can we keep this in the parent thread?

test/main.cpp
```diff
@@ -294,18 +304,21 @@ struct conv_forward : output_tensor_fixture
...
-            CHECK(time > 0.0);
-        }
-        else
-        {
-            CHECK(time == 0.0);
-        }
+            float time;
+
+            STATUS(miopenGetKernelTime(used_handle, &time));
+            STATUS(miopenDestroy(used_handle));
```
atamazov:
```suggestion
#if MIOPEN_BUILD_DEV
            STATUS(miopenDestroy(used_handle));
#endif
```

atamazov (changes requested):
Don't destroy the handle you haven't created.

atamazov:
@DrizztDoUrden Please update PR description like this:

Fixes UB in test_main (DEV builds only). miopen handle that employs _explicitly created HIP stream_ should be used in the same thread where is has been created. Otherwise we run into weird problems like ["Device mismatch from where this module is loaded"](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1739#issuecomment-1320210095) with DEV builds.

/cc @pfultz2

pfultz2:
The handle should work across different threads without a device mismatch. These tests were added to ensure we support this.

atamazov:
@pfultz2 We create non-NULL HIP streams in DEV builds. **_Is it really safe to explicitly create such a stream in other thread and use it in other thread?_** We see "Device mismatch" errors on MI200 with this, and these happen randomly.

**_Do you think that this PR contains not a fix but actually a workaround?_**

If so, then I think we shall open a ticket for the HIP runtime, because this PR does not change any miopen-handle related code in the library. I think that it changes some timing relationships and _that_ is the implicit reason for the bug in the HIP runtime to manifest itself.

DrizztDoUrden:
> The handle should work across different threads without a device mismatch. These tests were added to ensure we support this.

This should work across different threads, but not with how tests create handles in BUILD_DEV=On configuration.

DrizztDoUrden:
By the way, what is a correct way to proceed with changes to fin? I guess I should create a PR there as well. Should that PR get merged after this one?

atamazov:
> @DrizztDoUrden +1 to Jun. Is it time to [#1739 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1739#issuecomment-1306385932) again? If so, then please merge latest `develop` and remove `ON_HOLD`, thanks.

Testing is in progress...

> By the way, what is a correct way to proceed with changes to fin? 

I think that fin changes should go first, then update fin submodule hash right in this PR.

DrizztDoUrden:
The thing is this pr contains fin on a commit it can be built with. If you check out the pr, it would be built and executed correctly. That's why it passes checks. At the same time fin doesn't pass. I haven't checked why yet, but 99% that it tries to use miopen from develop.

atamazov:
## 🌀 Testing

- FP32 conv, Radeon VII, ROCm 5.3
- FP16 conv, Radeon VII, ROCm 5.3
- FP32 conv, Navi 21, CI docker (ROCm 5.3 + libMLIR for 5.5)
- FP16 conv, Navi 21, CI docker (ROCm 5.3 + libMLIR for 5.5)

🟢 No correctness or performance regressions, good to go!

junliume:
Hi @DrizztDoUrden , please resolve the conflict and we can push this PR through CI.

DrizztDoUrden:
@junliume done. What is the right order of merging Fin changes? Should they go before or after this PR?

DrizztDoUrden:
@junliume this has no conflicts and passes CI

atamazov:
@DrizztDoUrden @junliume I think this is on hold until `release/rocm-rel-5.6-staging` is cut, correct?

src/convolution.cpp
```diff
@@ -58,6 +58,100 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_FFT)
 
 namespace miopen {
 
+std::size_t GetMaxWorkSpaceSize(const std::vector<std::pair<std::string, std::size_t>>& values)
+{
+    try
```
JehandadKhan: When would an exception be thrown here?

src/ocl/convolutionocl.cpp
```diff
@@ -479,79 +235,38 @@ void ConvolutionDescriptor::FindConvFwdAlgorithm(Handle& handle,
...
 
-    if(IsEnabled(MIOPEN_DEBUG_COMPILE_ONLY{}))
-        MIOPEN_THROW(
-            miopenStatusGpuOperationsSkipped,
-            "MIOPEN_DEBUG_COMPILE_ONLY is enabled, escaping forward convolution. Search skipped.");
+    const auto results = FindConvolution(ctx, problem, invoke_ctx);
 
-    if(perf_db.empty())
+    if(results.empty())
         MIOPEN_THROW("Forward Convolution cannot be executed due to incorrect params");
```
JehandadKhan: Can we update this error message to something more descriptive, such as "No suitable algorithm was found to execute the required convolution"

src/ocl/convolutionocl.cpp
```diff
@@ -1003,115 +661,18 @@ std::size_t ConvolutionDescriptor::GetForwardSolutionWorkspaceSize(Handle& handl
...
-    if(invoker)
-        return *invoker;
-    return PrepareInvoker(handle, ctx, problem, config, solver_id, dir);
-}
-
 static bool CheckInvokerSupport(const solver::Id solver_id, conv::Direction dir)
 {
     const auto& algo = solver_id.GetAlgo(dir);
     return CheckInvokerSupport(algo);
 }
```
JehandadKhan: All solvers support invokers, perhaps this and other such logic can be removed ?

src/ocl/convolutionocl.cpp
```diff
@@ -1560,124 +956,37 @@ void ConvolutionDescriptor::FindConvBwdWeightsAlgorithm(Handle& handle,
...
-                     "MIOPEN_DEBUG_COMPILE_ONLY is enabled, "
-                     "escaping backwards convolution. Search "
-                     "skipped.");
-
-    if(perf_db.empty())
+    const auto problem =
+        conv::ProblemDescription{dyDesc, dwDesc, xDesc, *this, conv::Direction::BackwardWeights};
+    const auto ctx = [&] {
+        auto tmp = ExecutionContext{&handle};
+        tmp.DetectRocm();
```
JehandadKhan, line 974: I see this in many places, why cant this call be moved to the ctor ?

src/convolution.cpp
DrizztDoUrden: Artifact of old times :) Will remove.

src/ocl/convolutionocl.cpp
DrizztDoUrden: Yes, sure.

src/ocl/convolutionocl.cpp
DrizztDoUrden: Yes, but it's been removed in a PR by @atamazov, I think. I can do it here as well if that's needed.

src/ocl/convolutionocl.cpp
DrizztDoUrden: I am not sure, but I'd rather do it in a separate PR.
And this is slow on the first call, it actually invokes stuff and caches the result in a static variable: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ec1fb7ad54d22fe5d2934e3396c8bcc20375a049/src/ocl/gcn_asm_utils.cpp#L141-L141
We may want to not do it where not needed, so that say GWSS won't look slower than it is in profiler.

JehandadKhan (approved):

src/ocl/convolutionocl.cpp
JehandadKhan: Perhaps we should move such initializations to the handle creation. The context is used every where in the library and it seems clunky to call this function everytime.

src/include/miopen/conv/problem_description.hpp
```diff
@@ -151,16 +153,16 @@ struct ProblemDescription : ProblemDescriptionBase
...
     ProblemDescription() = default;
 
-    ProblemDescription(const TensorDescriptor& in_,
-                       const TensorDescriptor& weights_,
-                       const TensorDescriptor& out_,
-                       const ConvolutionDescriptor& conv_,
+    ProblemDescription(TensorDescriptor in_,
+                       TensorDescriptor weights_,
+                       TensorDescriptor out_,
+                       ConvolutionDescriptor conv_,
```
averinevg: The arguments are copied when passed by value, in this case using `std::move` has no advantages

DrizztDoUrden: With this pattern they can be either copied or moved from outside.
E.g.: `ProblemDescription(args...)` would copy them, while `ProblemDescription(std::move(args)...)` would move.

averinevg: `TensorDescriptor` and `ConvolutionDescriptor` both contains a few small vectors that could be moved and several variables that are copied anyway. I think that this will not affect the speed, since memory allocation here takes much longer than copying.

averinevg: I think in this case it's worth adding a separate constructor that takes rvalue references

src/conv/solver_finders.cpp
```diff
@@ -0,0 +1,335 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2022 Advanced Micro Devices, Inc.
```
averinevg:
```suggestion
 * Copyright (c) 2023 Advanced Micro Devices, Inc.
```

averinevg (approved):

DrizztDoUrden:
I don't get it. It reproduces on my PC on develop too.

junliume (approved):

junliume:
Ping @JehandadKhan @averinevg and @atamazov before merging.

JehandadKhan:
@DrizztDoUrden  Can you open a PR in fin as well so that we can move the hash forward rather than making changes to fin here which would make it challenging to keep fin in sync

JehandadKhan (approved):

DrizztDoUrden:
@JehandadKhan Yes, there is one already: https://github.com/ROCmSoftwarePlatform/MIFin/pull/74

src/convolution.cpp
```diff
@@ -338,18 +424,11 @@ ConvolutionDescriptor::WrwGetValidWorkSpaceSizeGemm(const TensorDescriptor& dyDe
...
+std::size_t ConvolutionDescriptor::GetWorkSpaceSize(ExecutionContext ctx,
+                                                    const conv::ProblemDescription& problem) const
 {
     MIOPEN_LOG_I2("");
 
-    const auto problem = ProblemDescription{xDesc, wDesc, yDesc, *this, conv::Direction::Forward};
-    auto ctx           = ConvolutionContext{};
-    ctx.SetStream(&handle);
-    ctx.DetectRocm();
-    ctx.SetupFloats(problem);
```
atamazov, line 352: @DrizztDoUrden It seems that changes like this lead to [SWDEV-404537](https://ontrack-internal.amd.com/browse/SWDEV-404537), which is fixed in #2288. I am wondering if similar problems are still hiding somewhere in our code even after the fix.

As you can see, in the old code we do not fully construct ExecutionContext from handle and manually call DetectRocm when necessary. The refactored code is missing this. Shall we always call DetectRocm in ctor that accepts handle (`ExecutionContext::ExecutionContext(Handle* stream_)`) to prevent this problem? And #2288 can be reverted at once.

Oops, I see that #2289 is opened.

atamazov (commented):
missing DetectRocm calls

src/ocl/convolutionocl.cpp
```diff
@@ -111,326 +107,122 @@ static inline void ValidateGroupCount(const TensorDescriptor& xDesc,
...
-    try
-    {
-        return FindAllWinogradSolutions(ctx, problem, invoke_ctx);
-    }
-    catch(miopen::Exception& ex)
-    {
-        MIOPEN_LOG_WE(ex.what());
-        return {};
-    }
+    ctx.DetectRocm();
```
atamazov, line 115: This won't be necessary

src/convolution.cpp
atamazov: @DrizztDoUrden BTW I am wondering if we still need the default ctor `ExecutionContext::ExecutionContext()`. If we can get rid of it, then I think we can make `DetectRocm()` the private member of `ExecutionContext` and then never call it manually.

DrizztDoUrden: Yes, I think that would be a proper way to do this.

atamazov: [resolved]

PR #1742, opened 2022-09-09
Lwptuna 63

bghimireamd:
* remove cpp files that are part of GLOB's lists in test/gtest/CMakeLists.txt
*  fixed merge conflict with latest develop branch


src/convolution_api.cpp
```diff
@@ -295,105 +295,140 @@ miopenConvolutionForwardGetWorkSpaceSize(miopenHandle_t handle,
...
+}
+
+void LogCmdFindConvolution(const miopenTensorDescriptor_t& xDesc,
+                           const miopenTensorDescriptor_t& wDesc,
+                           const miopenConvolutionDescriptor_t& convDesc,
+                           const miopenTensorDescriptor_t& yDesc,
+                           const miopen::ConvDirection& conv_dir,
+                           bool is_immediate)
+{
+    if(miopen::IsLoggingCmdFind())
```
atamazov, line 427: We do not need separate env variable because it is easy to distinguish FindConvolution from other conv calls looking at function name. This is output of FindConvolution:
```
MIOpen(HIP): Command [LogCmdFindConvolution] ./bin/MIOpenDriver ...
```
This is output of other convolution calls:
```
MIOpen(HIP): Command [LogCmdConvolution] ./bin/MIOpenDriver ...
```
You can `grep -v LogCmdFindConvolution` to exclude FindConvolution calls from the log or you can `grep -v LogCmdConvolution` to exclude other convolution calls.

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#pragma once
+
+#include <gtest/gtest.h>
+
+#include "miopen/logger.hpp"
```
atamazov, line 30:
```suggestion
#include <miopen/logger.hpp>
```

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,73 @@
...
+#include <miopen/convolution.hpp>
+
+#if MIOPEN_BACKEND_OPENCL
+#define BKEND "OpenCL"
+#elif MIOPEN_BACKEND_HIP
+#define BKEND "HIP"
+#endif
+
+static const std::string& logConv =
```
atamazov, line 39: Please move definition to .cpp, keep only declaration here.

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,73 @@
...
+    "MIOpen(" BKEND
+    "): Command [LogCmdConvolution] ./bin/MIOpenDriver conv -n 128 -c 3 -H 32 -W 32 -k "
+    "64 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1";
+static const std::string& logFindConv =
```
atamazov, line 43: Ditto.

test/gtest/log_test_helper.hpp
```diff
@@ -0,0 +1,73 @@
...
+    "MIOpen(" BKEND
+    "): Command [LogCmdFindConvolution] ./bin/MIOpenDriver conv -n 128 -c 3 -H 32 -W 32 "
+    "-k 64 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1";
+
+static const std::string& envConv     = "MIOPEN_ENABLE_LOGGING_CMD";
+static const std::string& envFindConv = "MIOPEN_ENABLE_LOGGING_CMD_FIND";
```
atamazov, line 49: Ditto

test/gtest/log_test_helper.cpp
```diff
@@ -0,0 +1,171 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include "log_test_helper.hpp"
+
+void setEnvironmentVariable(const std::string& name, const std::string& value)
```
atamazov, line 28: Please move this into unnamed namespace or declare static.

test/gtest/log_test_helper.cpp
```diff
@@ -0,0 +1,171 @@
...
+private:
+    std::streambuf* old_stream_buf;
+    std::stringstream str_buffer;
+};
+
+struct Tensor
+{
+    miopenTensorDescriptor_t desc{};
+    size_t data_size;
+#if MIOPEN_BACKEND_OPENCL
```
atamazov, line 82: Where this macro is defined?

test/gtest/log_test_helper.cpp
```diff
@@ -0,0 +1,171 @@
...
+    Conv() : input(128, 3, 32, 32), output(128, 64, 32, 32), weights(64, 3, 3, 3)
+    {
+        EXPECT_EQ(miopenCreateConvolutionDescriptor(&convDesc), 0);
+        EXPECT_EQ(miopenInitConvolutionDescriptor(convDesc, miopenConvolution, 1, 1, 1, 1, 1, 1),
+                  0);
+    }
+    ~Conv() { miopenDestroyConvolutionDescriptor(convDesc); }
+};
+
+bool isSubStr(const std::string& str, const std::string& sub_str)
```
atamazov, line 133: Please move this into unnamed namespace or declare static.

test/gtest/log_test_helper.cpp
```diff
@@ -0,0 +1,171 @@
...
+{
+    return str.find(sub_str) != std::string::npos;
+}
+
+void TestLogFun(std::function<void(const miopenTensorDescriptor_t&,
```
atamazov, line 138: [Humor] No fun 😄

src/include/miopen/convolution.hpp
```diff
@@ -74,6 +74,13 @@ using ExtraKernelArgs = std::tuple<int /*N*/,
 struct ConvFwdTensors;
 struct ConvWrwTensors;
 
+enum class ConvDirection
+{
+    Fwd = 1,
+    Bwd = 2,
+    WrW = 4
+};
+
```
atamazov, line 83: ### It seems like the current design of unit tests requires moving entities from the local scope (implementation, e.g. *.cpp) to global scope (interface, e.g. *.hpp). This would lead to littering the code with unnecessary things and should be avoided.

For now, please revert this change. Put a copy of `ConvDirection` into log_test_helper.hpp instead -- just before declarations of `LogCmdConvolution()` and `LogCmdFindConvolution()`.

atamazov (changes requested):

test/gtest/log_test_helper.cpp
bghimireamd: Macro defined [here ](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/include/miopen/config.h.in#L29)

atamazov: How do you import macro definition from there?

src/convolution_api.cpp
bghimireamd: @JehandadKhan had some input regarding this [here ](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1662#discussion_r947029795).

test/gtest/log_test_helper.cpp
bghimireamd: I include [logger.hpp](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/LWPTUNA-63/test/gtest/log_test_helper.hpp#L30) and it eventually includes `config.h` . I guess I could replace `#include <miopen/logger.hpp>` with `#include <miopen/config.h>`

src/convolution_api.cpp
bghimireamd: But I do agree with you. We just have to mindful while grepping.

test/gtest/log_test_helper.hpp
bghimireamd: Since `log_test.cpp` and `log_test.cpp` both need these string I made then extern in `log_test_helper.hpp`

PR #1749, opened 2022-09-12
[API] Dedicated API for Conv+Bias+Activ Fusion

JehandadKhan:


JehandadKhan:
In draft pending addition of  more tests

test/gtest/api_convbiasactiv.cpp
```diff
@@ -0,0 +1,188 @@
...
+        std::fill(output.begin(), output.end(), 0.0f);
+        std::fill(ref_out.begin(), ref_out.end(), 0.0f);
+        std::fill(bias.begin(), bias.end(), 0.0f);
+        int bias_mode = 1; // zero disables bias
+        convHostForward(input, ref_out, weights, bias_mode, bias, conv_desc);
+        activationHostInfer(cba_config.activ_mode,
+                            double_zero,
+                            double_zero,
+                            double_zero,
+                            ref_out.data,
```
xinlipn, line 107: Should this be input.data instead?

xinlipn, line 107: Use non-zero alpha, beta, gamma value here?

junliume: @atamazov @DrizztDoUrden could you take a look while @JehandadKhan is on vacation?

junliume (approved):

PR #1752, opened 2022-09-13
bg/LWPMIOPEN-40 :  Dump out input/output and weights tensor when a NaN/Inf occurs in Convolution

bghimireamd:
* Overloading Handle::ReadTo to take raw pointer.
* Added API DumpTensorToFileFromDevice to dump tensor when nan is encountered and env variable MIOPEN_DUMP_TENSOR_PATH is set.
* Addtest test for verification.


junliume (approved):

src/ocl/convolutionocl.cpp
```diff
@@ -588,12 +639,23 @@ static void ConvForwardCheckNumerics(const Handle& handle,
         return;
     }
 
-    miopen::checkNumericsInput(handle, tensors.xDesc, tensors.x);
-    miopen::checkNumericsInput(handle, tensors.wDesc, tensors.w);
+    bool flag = false;
+
+    flag |= miopen::checkNumericsInput(handle, tensors.xDesc, tensors.x);
```
atamazov, line 644: Why do you use bitwise OR?

src/include/miopen/convolution.hpp
```diff
@@ -481,6 +481,11 @@ Invoker LoadOrPrepareInvoker(Handle& handle,
 
 std::ostream& operator<<(std::ostream& stream, const ConvolutionDescriptor& c);
 
+void DumpTensorToFileFromDevice(const miopen::Handle& handle,
```
atamazov, line 484: Please move this function into the `miopen::debug` namespace.

src/ocl/convolutionocl.cpp
```diff
@@ -578,6 +579,56 @@ void ValidateAlphaBeta(const void* alpha, const void* beta)
...
+
+    if(!file_stream.is_open())
+    {
+        MIOPEN_LOG_E("Cannot write to file : " << file_name_with_path_str);
+        return;
+    }
+
+    // read tensor data from gpu
+    size_t num_bytes = tDesc.GetNumBytes();
+    MIOPEN_LOG_I2("Start bringing tensor from device to host");
```
atamazov, line 621: I highly recommend moving this to Trace level or removing this logging.

src/ocl/convolutionocl.cpp
```diff
@@ -578,6 +579,56 @@ void ValidateAlphaBeta(const void* alpha, const void* beta)
...
+    std::vector<char> hdata(num_bytes);
+    handle.ReadTo(hdata.data(), dData, num_bytes);
+    MIOPEN_LOG_I2("Done bringing tensor from device to host");
```
atamazov, line 624: Ditto

src/ocl/convolutionocl.cpp
```diff
@@ -578,6 +579,56 @@ void ValidateAlphaBeta(const void* alpha, const void* beta)
...
+    // write tensor data to file
+    const char* pointer = reinterpret_cast<const char*>(&hdata[0]);
+    file_stream.write(pointer, num_bytes);
+    file_stream.close();
+    MIOPEN_LOG_I("Dumping tensor to file : " << file_name_with_path_str);
```
atamazov, line 629: Please lower level to Info2. It may worth printing num_bytes.

src/ocl/convolutionocl.cpp
```diff
@@ -588,12 +639,23 @@ static void ConvForwardCheckNumerics(const Handle& handle,
...
+    flag |= miopen::checkNumericsOutput(handle, tensors.yDesc, tensors.y);
+
+    const char* file_name = miopen::GetStringEnv(MIOPEN_DUMP_TENSOR_PATH{});
+    if(flag && static_cast<bool>(file_name))
+    {
+        std::string file_name_str = file_name;
+        DumpTensorToFileFromDevice(handle, tensors.xDesc, tensors.x, file_name_str + "_x.bin");
+        DumpTensorToFileFromDevice(handle, tensors.wDesc, tensors.w, file_name_str + "_w.bin");
+        DumpTensorToFileFromDevice(handle, tensors.yDesc, tensors.y, file_name_str + "_y.bin");
+    }
```
atamazov, line 658: It is definitely worth moving dumping functionality into `checkNumericsOutput()` and `checkNumericsOutput()`. This would enable dumping functionality in all primitives.

Getting `file_name_str` can be moved to a function in check_numerics.cpp

`DumpTensorToFileFromDevice()` should be moved there as well.

src/ocl/convolutionocl.cpp
```diff
@@ -588,12 +639,23 @@ static void ConvForwardCheckNumerics(const Handle& handle,
...
+    flag |= miopen::checkNumericsInput(handle, tensors.xDesc, tensors.x);
+    flag |= miopen::checkNumericsInput(handle, tensors.wDesc, tensors.w);
 
     worker();
 
-    miopen::checkNumericsOutput(handle, tensors.yDesc, tensors.y);
+    flag |= miopen::checkNumericsOutput(handle, tensors.yDesc, tensors.y);
+
+    const char* file_name = miopen::GetStringEnv(MIOPEN_DUMP_TENSOR_PATH{});
+    if(flag && static_cast<bool>(file_name))
```
atamazov, line 652: This should be coded like this, no static_cast please:
```C++
const auto file_name = miopen::GetStringEnv(MIOPEN_DUMP_TENSOR_PATH{});
if(... && file_name != nullptr)...
```

test/gtest/dumpTensorTest.cpp
```diff
@@ -0,0 +1,168 @@
+#include <iostream>
```
atamazov, line 1: Please use only lowercase letters and underscore in filenames. Please prepend test name with `debug_` as it tests debugging functionality.

test/gtest/dumpTensorTest.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include <limits>
```
atamazov, line 2: Please add license header

src/ocl/convolutionocl.cpp
```diff
@@ -65,6 +65,7 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_FFT)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEVICE_ARCH)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_IMMED_FALLBACK)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_COMPILE_ONLY)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DUMP_TENSOR_PATH)
```
atamazov, line 68: Please rename to `MIOPEN_CHECK_NUMERICS_DUMP_TENSOR_PATH` and add documentation to https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/DebugAndLogging.md#numerical-checking

test/gtest/dumpTensorTest.cpp
```diff
@@ -0,0 +1,168 @@
...
+    miopen::Handle handle{};
+    // before writing to gpu we set one of the element
+    // in the vector to nan.
+    // host_tensor.data[nan_index] = std::numeric_limits<T>::quiet_NaN();
+    populateWithNAN(host_tensor.data);
+
+    // write the tensor to GPU
+    auto gpu_tensor_addr = handle.Write(host_tensor.data);
+
+    miopen::DumpTensorToFileFromDevice(
```
atamazov, line 122: [Question] This is a unit test for the `DumpTensorToFileFromDevice()` only, and it does not test env variables that controls it, right?

atamazov:
@bghimireamd @junliume Please let me know if you need a ticket to track https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1752#pullrequestreview-1133632691. Proposed labels: `urgency_high`, `enhancement`, `quality`.

junliume:
> @bghimireamd @junliume Please let me know if you need a ticket to track [#1752 (review)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1752#pullrequestreview-1133632691). Proposed labels: `urgency_high`, `enhancement`, `quality`.

@bghimireamd could you track the above review opinions? Thanks! @atamazov

src/include/miopen/convolution.hpp
bghimireamd: will do

src/ocl/convolutionocl.cpp
bghimireamd: I will remove it.

src/ocl/convolutionocl.cpp
bghimireamd: I will remove it.

src/ocl/convolutionocl.cpp
bghimireamd: will make it `MIOPEN_LOG_I2` and add num_bytes (good point).

src/ocl/convolutionocl.cpp
bghimireamd: @atamazov If any of these tensor (`tensors.xDesc`, `tensors.wDesc` and `tensors.yDesc`) have `nan`, we need to dump all three and not just one that has `nan`. If we directly print in `checkNumericsOutput()` and `checkNumericsOutput()` we would  end up dumping only the tensor that has `nan`. 

eg: 
Move dumping functionality into `checkNumericsOutput()` and `checkNumericsOutput()`:

`tensors.xDesc`  : does not have `nan`
`tensors.wDesc` : does not have `nan`
`tensors.yDesc`  : have `nan`

Only `tensors.yDesc` is dumped. But users might want to see all three tensors and not just one. 

I agree moving `DumpTensorToFileFromDevice()` to `check_numerics.cpp`
I also agree moving ` file_name_str` to `check_numerics.cpp`

I am open for suggestion :smiley:

src/ocl/convolutionocl.cpp
bghimireamd: Because either `tensors.x` or `tensors.w` or `tensors.y` could contain `nan`. If any of these three tensor contain `nan` we dump all of them and not just one that encountered `nan`.  Please see [here ](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1752#discussion_r990218840)for more detail.

src/ocl/convolutionocl.cpp
bghimireamd: agreed!

test/gtest/dumpTensorTest.cpp
bghimireamd: Agree! I will make it `debug_dump_tensor_test.cpp`

test/gtest/dumpTensorTest.cpp
bghimireamd: will do

src/ocl/convolutionocl.cpp
bghimireamd: will do

test/gtest/dumpTensorTest.cpp
bghimireamd: Good question. I will actually add the env variable in the test too

bghimireamd:
@atamazov @junliume @JehandadKhan I will create a new PR as per Artem's suggestions.

src/ocl/convolutionocl.cpp
atamazov: ~Resolved.~

atamazov: But why _bitwise_ (`|`) instead of logical (`||`)?

bghimireamd: ah! I agree I will replace it with `||`

src/ocl/convolutionocl.cpp
atamazov: Please notice values 4 and 8 at https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/DebugAndLogging.md#numerical-checking.
- If the user does not want to abort, then only the last set of problematic buffers will remain in disk. This is because each conv operation with NaN/Inf overwrites existing dumps. **_Such dumps are likely useless._** It is the fist set of dumps that is of real interest.
- If the user wants to abort, then **_dumps won't be written at all_**.

🟡 This must be redesigned.

atamazov: [Proposal]
- Create a class with internal state like `check_failed` flag, addresses of buffers (x, w, y) and more.
- Convert `checkNumericsInput()` and `checkNumericsInput()` into class members.
  - Do not THROW immediately; collect stats instead.
  - These methods should, in addition, register addresses of buffers in the class instance.
  - At the same time I would register strings that identify buffers (e.g. `dy`, `w`, `dx`) for the dump filenames.
- Create instance of class wherever you would like to perform checks.
- Call `instance.checkNumerics()` methods.
- At the end, call `instance.TearDown()` or smth alike. It should dump buffers and, if necessary, THROW.
  - Note: Using dtor for this looks attractive, but dtors should never fail (throw).

atamazov: Note that the design is still not MT/MP-safe. If MIOpen is used in multiple threads and failure happens simultaneously (which is often the case!) then the same files will be written from different threads, causing mess. The simplest way to avoid that is encoding thread id in the filenames.

atamazov: If the user wants to continue execution after NaN/Inf detected, and have distinct dumps for each failure, then this is possible. Example design:
- Count failures in some integer member. The member should be static in the class (one counter per process).
- The value of counter should be encoded in the filename when dumping.

All the operations with this counter should be protected (mutex?) to achieve MT-safety.

Encoding of thread id (see previous comment) is still needed to correctly work in multi-process (MP) environment.

test/gtest/dumpTensorTest.cpp
atamazov: I'm still of the opinion that tests for _debug functionality_ are not required. Debugging is an optional part of the library API. And, I guess, we simply don't have the resources for optional things. That is why my initial review comment was marked with just [Question].

I recommend keeping the tests for debugging to a minimum.

src/ocl/convolutionocl.cpp
atamazov: But `|` is actually perfectly portable with booleans. And there is no `||=` form. Any advantage of `||`?

atamazov: (you'll find the answer is very interesting)

bghimireamd: If we use `||`
`bool flag = miopen::checkNumericsInput(abc) || miopen::checkNumericsInput(xyz)`  
`miopen::checkNumericsInput(xyz)` will not execute if `miopen::checkNumericsInput(abc)` is true.

solution 1: 
`bool flag1 = miopen::checkNumericsInput(abc);`
`bool flag2 = miopen::checkNumericsInput(xyz);`

`bool flag = flag1 || flag2;`

solution 2:
`bool flag = false;`
`flag |= miopen::checkNumericsInput(abc);`
`flag |= miopen::checkNumericsInput(xyz);`

src/ocl/convolutionocl.cpp
bghimireamd: @atamazov  Thank you for design solution. One thing I would like to point out is: 

Lets say I am running MIOpen (multi threaded) in two parallel process:

file name format: `<file_name>` _ `<thread_id>` _ `<failure_count>`.txt . Thread id will be unique even across multiple process after using `GetProcessAndThreadId()`. 

process1 : `file_name_123_1.txt`, `file_name_456_0.txt`, `file_name_789_2.txt`  

process2 : `file_name_132_1.txt`, `file_name_4516_0.txt`, `file_name_113_2.txt`

See how the counts (`xxx_1.txt`, `_0.txt`, `_2.txt`) are not in order. Current design does guarantee uniqueness (for MT/MP) but I am not  sure if it guarantee the order. Do we even care about the order?

PR #1759, opened 2022-09-15
[HotFix] Fix MIOpen dependencies build issues and sync CK version

junliume:
@JehandadKhan Thanks for the branch, this PR is needed to let `develop` passing builds on the docker images provided by CQE
1. Fix `pcre` dependency issue: the `cppcheck` version in Composable Kernel depends on an outdated URL of it, so we are ignoring it here; (BTW~ CK should be removing that dependency?)
2. Fix CK sync issue, when enabled, CK should provide the APIs expected from MIOpen.

### DIFF
--- a/install_deps.cmake
+++ b/install_deps.cmake
@@ -110,6 +110,7 @@ if(DEFINED CMAKE_TOOLCHAIN_FILE)
 endif()
 # Initialize directory
 cget(init ${TOOLCHAIN_FLAG} -DCMAKE_INSTALL_RPATH=${PREFIX}/lib ${PARSE_UNPARSED_ARGUMENTS})
+cget(ignore pcre)
 
 # Install dependencies
 cget(install -U pfultz2/rocm-recipes)
--- a/rbuild.ini
+++ b/rbuild.ini
@@ -1,6 +1,7 @@
 [main]
 cxx = ${rocm_path}/llvm/bin/clang++
 cc = ${rocm_path}/llvm/bin/clang
+ignore = pcre
 deps =
     ROCmSoftwarePlatform/rocm-recipes
     -f requirements.txt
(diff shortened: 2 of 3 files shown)


illsilin:
I think the dev-requirements.txt file in the CK with the cppcheck version is obsolete and is not actually used anywhere. I am planning to remove it from CK repo (along with other obsolete stuff).

pfultz2:
cppcheck is a dev dependency only(in order to use `make analyze`) and should not be listed in the requirements.txt.

>  I am planning to remove it from CK repo (along with other obsolete stuff).

The requirements.txt file can be removed, but dev-requirements.txt should not be removed. It should list cppcheck(if you are using static analysis), googletest(we should not be doing FetchContent), and rocm-cmake version(since you need a newer version).

junliume:
@illsilin and @pfultz2 Thanks! That's another issue with CK repo, for now could you review and check this PR? It is needed to resolve the dependency build issues from current MIOpen `develop`

install_deps.cmake
```diff
@@ -110,6 +110,7 @@ if(DEFINED CMAKE_TOOLCHAIN_FILE)
 endif()
 # Initialize directory
 cget(init ${TOOLCHAIN_FLAG} -DCMAKE_INSTALL_RPATH=${PREFIX}/lib ${PARSE_UNPARSED_ARGUMENTS})
+cget(ignore pcre)
```
pfultz2, line 113: The ignore should be added to the main session in the rbuild.ini file as well. We do ignore it for develop and fin:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/rbuild.ini#L11
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/rbuild.ini#L21

We just need to add it to main as well, otherwise we will see the same issue when building the package.

junliume: Review opinion adopted, should be resolved now, please re-review.

illsilin (approved):

pfultz2 (approved):

shurale-nkn:
@junliume 
There is one problem with this PR. You rolled back CK to a version that uses hipcc, but to build a dependency from requirements.txt we use clang++. Thus, our Dockerfile is invalid, it cannot be assembled.  And an attempt to install the dependencies specified in requirements.txt will result in an error.

atamazov:
@junliume 
> Hi @shurale-nkn I am trying to fix this issue in this following PR #1790 ...
> We hesitate to merge it since we want to split it into smaller parts.

A good reason to create a separate fix for the CK build issue soon. The issue seems urgent enough as it prevents our users from building MIOpen from scratch, right?

junliume:
> @junliume
> 
> > Hi @shurale-nkn I am trying to fix this issue in this following PR #1790 ...
> > We hesitate to merge it since we want to split it into smaller parts.
> 
> A good reason to create a separate fix for the CK build issue soon. The issue seems urgent enough as it prevents our users from building MIOpen from scratch, right?

@atamazov yes, https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1761 is marked as blocker for now and should be completed by tomorrow.

PR #1765, opened 2022-09-19
[tests] activation ELU forward

xinlipn:
Activation ELU forward test. 
- To check execution at RunActivation()


junliume:
@xinlipn could you resolve the conflict by merging the latest `develop` branch to yours?

xinlipn:
> @xinlipn could you resolve the conflict by merging the latest `develop` branch to yours?

@junliume Fixed the conflicts.

test/gtest/activation.cpp
```diff
@@ -0,0 +1,243 @@
...
+        return miopenStatusSuccess;
+    else if(fwdStatus != miopenStatusSuccess)
+        return fwdStatus;
+    else
+        return bwdStatus;
+}
+
+INSTANTIATE_TEST_SUITE_P(ActivationTestSuite,
+                         TestActivation,
+                         ::testing::Values(ActivationConfig{16, 32, 8, 8, miopenActivationELU}));
```
JehandadKhan, line 224: Please add more test cases to cover 
* A few other sizes
* All supported activations. 
One test vector seems to be under-whelming.

JehandadKhan (changes requested):

test/gtest/activation.cpp
xinlipn: > Please add more test cases to cover
> 
>     * A few other sizes
> 
>     * All supported activations.
>       One test vector seems to be under-whelming.

Updated with 7 cases now and 9 activations. The following haven't passed though, miopenActivationLOGISTIC, miopenActivationTANH, miopenActivationPOWER, miopenActivationCLIPPEDRELU

format.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
junliume, line 1: While it is very useful, this file should not be needed to check in :)

xinlipn:
> @xinlipn can we take a look at the CI issues? It seems that a few unit tests are failing. Hence failing with non-COMGR stage.

@junliume , yes, one assert failed. I have been working on it.

test/gtest/activation.cpp
```diff
@@ -227,7 +227,7 @@ INSTANTIATE_TEST_SUITE_P(ActivationTestSuite,
                                                 // miopenActivationLOGISTIC,
                                                 // miopenActivationTANH,
                                                 miopenActivationRELU,
-                                                miopenActivationSOFTRELU,
+                                                // miopenActivationSOFTRELU,
```
junliume: If this is an error tolerance issue, I would recommend setting a tolerance level instead of disabling the tests.
CC: @JehandadKhan

xinlipn: > If this is an error tolerance issue, I would recommend setting a tolerance level instead of disabling the tests. CC: @JehandadKhan

Yes, this is a temporary while still trying to reproduce this issue reliably to root cause.

driver/CMakeLists.txt
```diff
@@ -27,9 +27,11 @@ cmake_minimum_required( VERSION 3.5)
 
 find_package(Threads REQUIRED)
 
-add_executable(MIOpenDriver main.cpp InputFlags.cpp)
+add_library(InputFlags OBJECT InputFlags.cpp)
+add_executable(MIOpenDriver main.cpp $<TARGET_OBJECTS:InputFlags> )
```
JehandadKhan: Why was this required ?

test/gtest/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ function(add_gtest TEST_NAME)
   add_dependencies(tests test_${TEST_NAME})
   add_dependencies(check test_${TEST_NAME})
   target_compile_options(test_${TEST_NAME} PRIVATE -Wno-global-constructors -Wno-undef)
-  target_include_directories(test_${TEST_NAME} PRIVATE ../)
+  target_include_directories(test_${TEST_NAME} PRIVATE ../ PRIVATE ${PROJECT_SOURCE_DIR}/driver)
```
JehandadKhan: What code are tests using from the driver

test/gtest/activation.cpp
```diff
@@ -0,0 +1,262 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2017 Advanced Micro Devices, Inc.
```
JehandadKhan:
```suggestion
 * Copyright (c) 2022 Advanced Micro Devices, Inc.
```

test/gtest/activation.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include <miopen/miopen.h>
+#include <miopen/stringutils.hpp>
+#include <miopen/tensor.hpp>
+#include <utility>
+
+#include <fusionHost.hpp>
+#include "activ_driver.hpp"
+#include "InputFlags.hpp"
+#include "get_handle.hpp"
+//#include "tensor_holder.hpp"
```
JehandadKhan: Please remove if not required

test/gtest/activation.cpp
```diff
@@ -0,0 +1,262 @@
...
+        dinput_cpu = input;
+        dinput_gpu = input;
+
+        // TODO: use miopen API?
+        miopenCreateActivationDescriptor(&activ_desc);
+        miopenSetActivationDescriptor(activ_desc, activ_mode, alpha, beta, gamma);
+
+        std::size_t n, c, h, w;
+        auto&& handle        = get_handle();
+        std::tie(n, c, h, w) = miopen::tien<4>(input.desc.GetLengths());
```
JehandadKhan: You can also check if 

``` C++
n == activ_config.N
```

etc

test/gtest/activation.cpp
```diff
@@ -0,0 +1,262 @@
...
+        size_t total_mem =
+            6 * input.desc.GetNumBytes(); // estimate based on both forward and backward passes
```
JehandadKhan: Why does it have to be 6 ?

test/gtest/activation.cpp
```diff
@@ -0,0 +1,262 @@
...
+    // miopenActivationMode_t activ_mode;
+};
+
+template <class T1, class T2>
+void CompareTensors(T1&& t1, T2&& t2)
+{
+    EXPECT_FALSE(miopen::range_zero(t1)) << "CPU data is all zeros";
+    EXPECT_FALSE(miopen::range_zero(t2)) << "GPU data is all zeros";
+    const auto maxDiff = miopen::max_diff(t1, t2);
+    std::ignore        = maxDiff;
```
JehandadKhan:
```suggestion
    std::ignore = miopen::max_diff(t1, t2);
```

test/gtest/activation.cpp
```diff
@@ -0,0 +1,262 @@
...
+    // For debugging purpose only, remove later
+    std::cerr << "idx = " << idx << "; miopen::range_distance(t1) = " << miopen::range_distance(t1)
+              << std::endl;
+    EXPECT_FALSE(idx < miopen::range_distance(t1));
+    return;
+}
+
+struct TestActivation
+    : public ::testing::TestWithParam<std::tuple<miopenActivationMode_t, ActivationConfig>>
+{
```
JehandadKhan: Lets split the test into forward and backward tests.

test/gtest/activation.cpp
```diff
@@ -0,0 +1,262 @@
...
+    if(fwdStatus == miopenStatusSuccess && bwdStatus == miopenStatusSuccess)
+        return miopenStatusSuccess;
+    else if(fwdStatus != miopenStatusSuccess)
+        return fwdStatus;
+    else
+        return bwdStatus;
+}
+
+INSTANTIATE_TEST_SUITE_P(ActivationTestSuite,
+                         TestActivation,
```
JehandadKhan: Please move the instantiation to the end of the file.

JehandadKhan (changes requested):

atamazov:
- Is this ready for review?
  - Yes please  --@xinlipn
- How can we make sure that the new test checks the same configs as the old one?
  - New tests take combination of data types, activation mode, configs. Will cross check and update the configs/coverage to old one  --@xinlipn
- If a failure occurs with some config, what diagnostics does the test print? How to re-test the failing config in isolation?
  - Googletest macros i.e. ASSERT_ , EXPECT_ will spew error info along with google test logs for each test case. This could help isolate. No auto re-test mechanism is integrated yet as far as I know  --@xinlipn

driver/CMakeLists.txt
xinlipn: Will revert

test/gtest/CMakeLists.txt
xinlipn: #include "activ_driver.hpp" would need this line...

test/gtest/activation.cpp
xinlipn: > * Is this ready for review?
Yes please
> * How can we make sure that the new test checks the same configs as the old one?
New checks take the combination of data type, activation functions/mode and configurations. For the same input, same results/coverage is expected
> * If a failure occurs with some config, what diagnostics does the test print? How to re-test the failing config in isolation?
EXPECT_ or ASSERT_ macros in the code provided by googletest would spew info when failure happens at that point

atamazov:
@xinlipn Please clearly mark your edits in others' comments. Please also note that nobody receives notifications about edited comments. It is better to quote into a new comment and answer there.

atamazov:
@xinlipn 
>> * How can we make sure that the new test checks the same configs as the old one?
>   * New tests take combination of data types, activation mode, configs. Will cross check and update the configs/coverage to old one  --@xinlipn

Good.

 
>> * If a failure occurs with some config, what diagnostics does the test print? How to re-test the failing config in isolation?
>   * Googletest macros i.e. ASSERT_ , EXPECT_ will spew error info along with google test logs for each test case. This could help isolate. No auto re-test mechanism is integrated yet as far as I know  --@xinlipn

Unfortunately this does not answer my question. Are you aware of how re-testing of failed configs can be performed with current activation test? If not, please begin with:
```bash
$ ./bin/test_activation --verbose
$ ./bin/test_activation --help
```

atamazov (changes requested):
This PR drops existing functionality that allows the engineer to reproduce single failures and easily create regression tests.

JehandadKhan:
> 🔴 This PR drops existing functionality that allows the engineer to reproduce single failures and easily create regression tests.

The user can easily use the driver to reproduce the error! I am not sure there is value in reproducing the same functionality twice in the library.

atamazov:
@JehandadKhan Let's print failing configs at least then. It would be also nice to have a `--verbose` option that prints all configs under test.

atamazov:
@JehandadKhan What I can't accept is a situation when some PR quietly introduces limitations in functionality. I believe you agree that all the introduced changes must be carefully documented in the title and description or PR.

/cc @junliume

atamazov:
> @JehandadKhan Let's print failing configs at least then...

🟡 But please note that the driver and the test are not the same things. For example, the driver might use different thresholds or different order of library calls. Also the driver uses public API calls while the test is not limited to that. This might lead to situations where reproducing is impossible.

JehandadKhan:
> @JehandadKhan Let's print failing configs at least then. It would be also nice to have a `--verbose` option that prints all configs under test.

@xinlipn Failing configs should print automatically if you implement the `<<` operator for your test case. Please confirm that it is indeed the case.

xinlipn:
> > @JehandadKhan Let's print failing configs at least then. It would be also nice to have a `--verbose` option that prints all configs under test.
> 
> @xinlipn Failing configs should print automatically if you implement the `<<` operator for your test case. Please confirm that it is indeed the case.

@JehandadKhan , yes, latest updates have "<<" implemented and will output configs automatically

xinlipn:
> @JehandadKhan Let's print failing configs at least then. It would be also nice to have a `--verbose` option that prints all configs under test.

@atamazov  configs will be printed out as part of the test title as well as when it fails.

xinlipn:
>

driver/CMakeLists.txt
xinlipn: Reverted

test/activation.cpp
```diff
@@ -1,318 +0,0 @@
...
-#include "tensor_holder.hpp"
-#include "verify.hpp"
-
-std::string to_name(miopenActivationMode_t m)
-{
-#define STRING_CASE(x) \
-    case x: return #x; break;
-    switch(m)
-    {
-        STRING_CASE(miopenActivationPASTHRU)
```
atamazov: Where is this mode in the new test?

test/activation.cpp
```diff
@@ -1,318 +0,0 @@
...
-        STRING_CASE(miopenActivationLOGISTIC)
-        STRING_CASE(miopenActivationTANH)
-        STRING_CASE(miopenActivationRELU)
-        STRING_CASE(miopenActivationSOFTRELU)
-        STRING_CASE(miopenActivationABS)
-        STRING_CASE(miopenActivationPOWER)
-        STRING_CASE(miopenActivationCLIPPEDRELU)
```
atamazov: ...and this?

atamazov (changes requested):
I do not see where activation config is printed.

atamazov (changes requested):
The test must support the same data types as the original test supports. Right now I see only miopenFloat support. Please correct me if I am wrong.

test/activation.cpp
xinlipn: @atamazov  the above code seems to be from old activation.cpp

Configs are spewed out by << operator as below

struct ActivationConfig
{
    size_t N;
    size_t C;
    size_t H;
    size_t W;
    friend std::ostream& operator<<(std::ostream& outStream, const ActivationConfig& activConfig)
    {
        return outStream << "N:" << activConfig.N << " C:" << activConfig.C
                         << " H:" << activConfig.H << " W:" << activConfig.W;
    }
};

Here's an example
 EXPECT_EQ(status, miopenStatusSuccess)
        << "Backward activation failed. Config: " << activ_config;

xinlipn: > ...and this?
@atamazov the above code seems to be from old activation.cpp

Configs are spewed out by << operator as below
`struct ActivationConfig
{
size_t N;
size_t C;
size_t H;
size_t W;
friend std::ostream& operator<<(std::ostream& outStream, const ActivationConfig& activConfig)
{
return outStream << "N:" << activConfig.N << " C:" << activConfig.C
<< " H:" << activConfig.H << " W:" << activConfig.W;
}
};
`

Here's an example

`EXPECT_EQ(status, miopenStatusSuccess)
<< "Backward activation failed. Config: " << activ_config;`

xinlipn:
> The test must support the same data types as the original test supports. Right now I see only miopenFloat support. Please correct me if I am wrong.

@atamazov Not all types's test is stable yet. Will update other types.

test/activation.cpp
xinlipn: @atamazov this is indeed missing. Updated with this mode.

xinlipn:
> I do not see where activation config is printed.

> ...and this?
@atamazov the branch is sl/test_activation FYI. configs are printed out by << operator below

Configs are spewed out by << operator as below
`
return outStream << "N:" << activConfig.N << " C:" << activConfig.C
<< " H:" << activConfig.H << " W:" << activConfig.W;
`

Here's an example

`EXPECT_EQ(status, miopenStatusSuccess)
<< "Backward activation failed. Config: " << activ_config;`

test/activation.cpp
atamazov: It prints dimensions of one tensor, but the activation primitive has more parameters.

atamazov:
@xinlipn It prints dimensions of one tensor, but the activation primitive has more parameters.

test/activation.cpp
atamazov: Not resolved. miopenActivationCLIPPEDRELU is missing. Please "unresolve" this review comment.

/cc @JehandadKhan

atamazov:
@xinlipn 
> @atamazov Not all types's test is stable yet. Will update other types.

If so, then please convert this PR into draft. Or, if you would like to merge this test with FP32 support only, then please restore the original activation test.

xinlipn:
> @xinlipn It prints dimensions of one tensor, but the activation primitive has more parameters.

@atamazov Understood there are any benefits with current CTest. In comparison, GoogleTest may provide easy-to-read/modify, flexibility, more granularity in terms of unit tests, and compatibility with customers and other business units. Hope this makes sense

xinlipn:
@JehandadKhan Previous activation has been added back per discussion. Could you take a look?
@atamazov , please feel free to leave comments. Also is there a way to dump all test_activation's configs? I tried ctest -V, which doesn't spew out configs. Also tried ./test_activation -V, all configs give "Warning: Both CPU and GPU data is all zero"
[ctest-V.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9916701/ctest-V.txt)
[test_activation-V.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9916702/test_activation-V.txt)

atamazov:
@xinlipn 
> Also is there a way to dump all test_activation's configs? 

You should read and try instructions from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1765#issuecomment-1285673760. Example (this tests one default config):
```bash
$ ./bin/test_activation --verbose
./bin/test_activation --float --input 16, 32, 8, 8 --alpha 0.95 --beta 2.3 --gamma 3.4 --mode ELU --packed 1
error: 0
Max diff: 0
Forward Activation: miopenActivationELU
Input tensor: 16, 32, 8, 8
./bin/test_activation --float --input 16, 32, 8, 8 --alpha 0.95 --beta 2.3 --gamma 3.4 --mode ELU --packed 1
error: 0
Max diff: 0
Backwards Activation: miopenActivationELU
Input tensor: 16, 32, 8, 8
```

atamazov:
@xinlipn @junliume 
> I tried ctest -V, which doesn't spew out configs. Also tried ./test_activation -V, all configs give "Warning: Both CPU and GPU data is all zero"

It seems like test_activation is missing input tensor initialization. I am going to provide a fix soon.

xinlipn:
> ```shell
> ./bin/test_activation --verbose
> ```

@atamazov , Thanks for the reminder. Actually the 2nd file i.e. test_activation-V.txt I have attached earlier is  the output of 

./bin/test_activation --verbose

Which gives Warning: Both CPU and GPU data is all zero for all the configs. GPU uses is gfx1030

test/gtest/gtest_activation.cpp
```diff
@@ -0,0 +1,347 @@
...
+        miopenCreateActivationDescriptor(&activ_desc);
+        miopenSetActivationDescriptor(activ_desc, activ_mode, alpha, beta, gamma);
+
+        std::size_t n, c, h, w;
+        auto&& handle        = get_handle();
+        std::tie(n, c, h, w) = miopen::tien<4>(std::get<tensor<T>>(input).desc.GetLengths());
+        EXPECT_EQ(n, activ_config.N) << "Config: " << activ_config;
+        EXPECT_EQ(c, activ_config.C) << "Config: " << activ_config;
+        EXPECT_EQ(h, activ_config.H) << "Config: " << activ_config;
+        EXPECT_EQ(w, activ_config.W) << "Config: " << activ_config;
```
JehandadKhan: This seems redundant ?

test/gtest/gtest_activation.cpp
```diff
@@ -0,0 +1,347 @@
...
+                      static_cast<size_t>(h),
+                      static_cast<size_t>(w)};
+        std::get<tensor<T>>(output_cpu_ref) =
+            tensor<T>{static_cast<size_t>(n), // n from miopenGetConvolutionForwardOutputDim ?
+                      static_cast<size_t>(c),
+                      static_cast<size_t>(h),
+                      static_cast<size_t>(w)};
+
+        std::fill(
+            std::get<tensor<T>>(output_gpu).begin(), std::get<tensor<T>>(output_gpu).end(), NULL);
```
JehandadKhan: Instead of NULL its better to initialize with Nans

test/gtest/gtest_activation.cpp
```diff
@@ -0,0 +1,347 @@
...
+        std::fill(std::get<tensor<T>>(output_cpu_ref).begin(),
+                  std::get<tensor<T>>(output_cpu_ref).end(),
+                  NULL);
```
JehandadKhan: Same as above

JehandadKhan (changes requested):

test/gtest/gtest_activation.cpp
xinlipn: > This seems redundant ?

Removed

test/gtest/gtest_activation.cpp
xinlipn: > Instead of NULL its better to initialize with Nans

Initialized with std::numeric_limits<T>::quiet_NaN()

test/gtest/gtest_activation.cpp
xinlipn: Also initialized with std::numeric_limits<T>::quiet_NaN()

atamazov:
@junliume I seems like this PR was accidentally (and automatically) closed because of the text `"Resolves https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1765#issuecomment-1300868609"` in the just merged #1861. Can you please reopen it?

xinlipn:
Updated per @JehandadKhan's most recent three comments

test/gtest/gtest_activation.cpp
```diff
@@ -0,0 +1,344 @@
...
+                                            // miopenDouble),
+
+                                            ::testing::Values(miopenActivationPASTHRU,
+                                                              miopenActivationLOGISTIC,
+                                                              miopenActivationTANH,
+                                                              miopenActivationRELU,
+                                                              miopenActivationSOFTRELU,
+                                                              miopenActivationABS,
+                                                              miopenActivationPOWER,
+                                                              // miopenActivationCLIPPEDRELU,
```
atamazov, line 334: This is enabled in the original test.

atamazov (changes requested):
The test is LONG (~10 minutes or more) and therefore should not be used with testing datatypes other than MIOPEN_TEST_FLOAT to avoid CI performance regression. **_This is the only blocker that I see now provided that we keep the original `test_activation`._** Maybe the fastest way to resolve it is disabling this test (because we have test_activation that does the same job) until all the problems in the refactored test are resolved.

[Recommendations] These needs to be fixed sooner or later (at latest when we remove the original test_activation):
- Print configs https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1765#issuecomment-1300643170
- Clipped ReLU https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1765#discussion_r1015986535
- Reuse input tensor initialization from #1861 (this is proved to work correctly with all datatypes)

test/gtest/gtest_activation.cpp
xinlipn: @atamazov , I believe the rational is to migrate to GoogleTest frame work once more test cases are covered. That's why these activation modes are there.

junliume:
@xinlipn could you review the conflicts? We want to push this PR forward and merge it eventually.

test/gtest/gtest_activation.cpp
atamazov: The full Q is: Why miopenActivationCLIPPEDRELU is disabled here while it is enabled in the original test?

xinlipn: Let me re-run this test with miopenActivationCLIPPEDRELU enabled.

xinlipn:
> @xinlipn could you review the conflicts? We want to push this PR forward and merge it eventually.

@junliume, will re-run the test to address Artem's comment he just left then merge

junliume (approved):

PR #1776, opened 2022-09-21
ConvBinWinoRxS GetSolverDbId naming fix

alexandraBara:
Fixes the issue of getting the 
ConvBinWinoRxS solver DbId as:

- ConvBinWinoRxS<2, 3>
- ConvBinWinoRxS<3, 2> 

after rename in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1533

### DIFF
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -1937,15 +1937,9 @@ struct PerformanceConfigConvBinWinogradRxS : PerfConfigBase<PerformanceConfigCon
 template <int Winodata, int Winofilter>
 struct ConvBinWinoRxS final : ConvTunableSolver<PerformanceConfigConvBinWinogradRxS>
 {
-    const std::string& SolverDbId() const override { return GetSolverDbId(); }
-
-    static const std::string& GetSolverDbId()
+    const std::string& SolverDbId() const override
     {
-        static const std::string dbId = std::string("ConvBinWinogradRxSf")
-                                            .append(std::to_string(Winodata))
-                                            .append("x")
-                                            .append(std::to_string(Winofilter));
-        return dbId;
+        return GetSolverDbId<ConvBinWinoRxS<Winodata, Winofilter>>();
     }
 
     PerformanceConfigConvBinWinogradRxS


junliume (approved):

averinevg (changes requested):
It is not a bug. Custom `GetSolverDbId()` function has been implemented here in order to keep solver names.

CC @atamazov @johnny-keker

DrizztDoUrden (approved):
I agree with Evgenii. This should not be done without serious reasoning. Changing the id would make older db files easily fixable by a script, but containing invalid data.

alexandraBara:
> It is not a bug. Custom `GetSolverDbId()` function has been implemented here in order to keep old solver names.
> 
> CC @atamazov @johnny-keker

I see, thanks for explaining. We can address this in a different way in MITuna then, without having to address it here. I will close this PR.

alexandraBara:
> I agree with Evgenii. This should not be done without serious reasoning. Changing the id would make older db files easily fixable by a script, but containing invalid data.

Back to this discussion after talking to my team mate @cderb 
This seems like a workaround. It is confusing to the user that getting the solver name returns an old inexisting solver. I understand trying to keep historical DBs functional, but can that not be addressed in a different way? 

@junliume  any thoughts?

junliume:
> This seems like a workaround. It is confusing to the user that getting the solver name returns an old inexisting solver. I understand trying to keep historical DBs functional

Requesting input from @atamazov and @JehandadKhan. This PR is blocking Tuna's work while let's form a unified opinion before moving forward.

DrizztDoUrden:
> can that not be addressed in a different way?

Yes, it can. By generating new dbs or fixing them by basically a careful grep.

> It is confusing to the user that getting the solver name returns an old inexisting solver.

Solver ids are not guaranteed to be the same as class names. Correct me if I am wrong, but AFAIK there is no official way to get the solver id from the public API. DB files are also not meant to be transparent, even though they are de facto easily human-readable. So, the users should not be affected.

atamazov:
@junliume I do not see any problem with current implementation rather than it may confuse the users who are not fully aware of the library design. But we are here ready to explain ;)

To all: simple renaming of solvers is basically not allowed. If you need to rename a solver, then old SolverId (the numerical one) must be marked as unused, and the new SolverId must be occupied. Then new databases should be produced. These changes are allowed to be split to several PRs, but we shall expect performance regressions during transition period.

atamazov (changes requested):
Please see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1776#issuecomment-1256100682

Issue #1769, opened 2022-09-20
CU check from solvers causes fail during compile step


alexandraBara:
Some solvers check the local CU count in the IsValid function and this causes a segfault.
```
MIOpen(HIP): Info2 [Measure] Db::Load time: 117.739 ms
MIOpen(HIP): Info2 [GetPerfCfgParams] Perf Db: Record Loaded: ConvBinWinogradRxSf2x3

Program received signal SIGSEGV, Segmentation fault.
0x00007f2d0e1f9da5 in std::__uniq_ptr_impl<miopen::HandleImpl, std::default_delete<miopen::HandleImpl> >::_M_ptr (this=0x8)
    at /usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/unique_ptr.h:154
154	      pointer    _M_ptr() const { return std::get<0>(_M_t); }
```

Steps to reproduce:

1. launch the docker from ginger: ci-tuna:cderb_ubuntu20_dockerbuild_16
2. run fin with the input file attached ```/opt/rocm/bin/fin -i input_file.json -o some_file.json``` where input_file.json is the attached json file. The output file you can name to <whatever>.json just needs to exist locally.

[fin_input.json.zip](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9611388/fin_input.json.zip)

cderb:
This same error is also showing up during performance evaluation for MLIR. It seems to be a result of the reference to the **stream handler** deteriorating (cause yet unknown).

The error is observable by calling Fin on the attached input file. 

The entry point into miopen from fin is through a call to GetPerfCfgParams which takes the **ConvolutionContext**. This eventually calls down to mlir xdlop's IsValid function which then calls ConstructBuildOptions. The **segfault** happens when the ConvolutionContext, which has been passed in from Fin, is asked for its **stream handler**: ctx.GetStream().GetMaxComputeUnits(). This is peculiar, because if the same call is made in fin before the call to GetPerfCfgParams, it will pass without issue.

A backtrace of the Segfault follows:
``` Thread 1 "fin" received signal SIGSEGV, Segmentation fault.
0x00007ffff494712e in miopen::Handle::GetMaxComputeUnits() const () from /opt/rocm/lib/libMIOpen.so.1
(gdb) bt
#0  0x00007ffff494712e in miopen::Handle::GetMaxComputeUnits() const () from /opt/rocm/lib/libMIOpen.so.1
[...]

PR #1778, opened 2022-09-22
Fix for GetPerfCfgParams call

cderb:
ctx.problem should be ctx for GetPerfCfgParams passing method
#1769


junliume (dismissed):

averinevg (approved):
LGTM

jerryyin (approved):
Nice catch! 

It took me a while to wrap my head around why the typo didn't cause a compilation failure.... Until I see that you can basically construct a convolution context using a problem description. 

So what happened here is basically that we implicitly constructed a new convolution context from the problem description which itself naturally did not contain a valid execution context therefore had just invalid stream handles and all the headaches that we see later.

I think going forward we may want to get rid of that constructor to disallow this unexpected side effect to happen, and good job catching this!

PR #1788, opened 2022-09-26
[Find2.0] Fix direction for transpose convolution

umangyadav:
Fix for #1780

PR #1791, opened 2022-09-27
Fixed find 2.0 to work correctly with transposed convolution

### LINKED ISSUE #1780: [Find2.0][MIGraphX] Deconvolution support in Find 2.0 API
Is Deconvolution supported with Find 2.0 in MIOpen? 

MIGraphX has simple deconvolution test [here](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/test/verify/test_deconv.cpp). 

When I updated MIGraphX deconvolution Op to use Find.2.0 Interface of MIOpen, the test fails. 

Attached is the Log of MIOpen and MIGraphX for that particular test. 

[log.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9636495/log.txt)

Based on my debugging, It seems to me that MIOpen is not constructing the network configuration correctly for the Transpose/BackwardData convolution. 

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/7dbff4c14c9d0bf0f9f67d366bdf6abad7694694/src/problem.cpp#L302

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/7dbff4c14c9d0bf0f9f67d366bdf6abad7694694/src/problem.cpp#L157

DrizztDoUrden:
Fixes #1780.
Intended as a replacement for #1788.


src/problem.cpp
```diff
@@ -293,8 +332,15 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
...
 
-    const auto conv_dir = static_cast<conv::Direction>(direction);
-    const auto netcfg   = AsConvolution().BuildConfKey();
+    const auto conv_dir = ([&]() {
+        const auto dir = static_cast<conv::Direction>(direction);
+        if(dir == conv::Direction::BackwardWeights || conv_desc.mode != miopenTranspose)
+            return dir;
+        return dir == conv::Direction::Forward ? conv::Direction::BackwardData
+                                               : conv::Direction::Forward;
+    })();
```
umangyadav, line 341: I think this logic is also needed inside `AsConvolution()`.

umangyadav: Without that, i am still seeing error that it cann't find Invoker. And invoker it is trying to find has `Forward` convolution configuration but invoker was registered as `BackwardData`.

umangyadav: I required change similar to this in `AsConvolution` to make this PR work with MIGrpahX. 

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ef15b6aa7cc29638055f999d52e9ab129f29a8b9/src/problem.cpp#L157

umangyadav:
I find inconsistencies in `ProblemDescription` and `netcfg` for the miopenTranpose in some place and because of this transpose convolution requires additional handling. Perhaps there is a  better fix than this . 

e.g. 
1) Here `dy` and `dx` order is flipped. 
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ad9779dede0a36d86013ddf86fafe31998c3fc5c/src/ocl/convolutionocl.cpp#L1076

2) While constructing problem description, it is flipped again, as it passes `dx` as input
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ad9779dede0a36d86013ddf86fafe31998c3fc5c/src/ocl/convolutionocl.cpp#L1104

3)  https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ad9779dede0a36d86013ddf86fafe31998c3fc5c/src/problem_description.cpp#L107 

It is flipped one more time here

DrizztDoUrden:
Yes, legacy struct `miopen::ProblemDescription` is a mess, and we are going to get rid of it a bit later. On the other hand, I have fixed the incorrect network config used at the end of FindSolutions in [38b14ea](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1791/commits/38b14ea39c0d5ca3ed2635538e15e1fe1f30ab25). It was generated from non-transposed problem, so it had a wrong direction.

Unit test code related to transpose has been removed due to complication of properly adding it there and pre-existing ability to test transposed configs.
I have successfully tested with this config:
```
> bin/test_conv2d_find2 --float --cmode transpose --pmode default --group-count 1 --batch_size 1 --input_channels 32 --output_channels 32 --spatial_dim_elements 14 14 --filter_dims 1 1 --pads_strides_dilations 0 0 1 1 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW
```

DrizztDoUrden:
There is a test for convolution in general, which covers it. API calls in general are not covered by unit tests as the effort to cover every single branch inside would be massive.
But I don't know why did find 2.0 API PR has passed the CI while not being able to process any transposed config. I guess our transposed convolution test coverage may be literally zero. But that would be a wider issue separate from the bug. Of course, it would require further investigation, but I'd rather merge the fix and investigate the test right after that.

umangyadav:
>  But I don't know why did find 2.0 API PR has passed the CI while not being able to process any transposed config.

If you are talking about CI of this PR https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/pull/1346
It only coves the regular convolution not transposed and even for that regular one, it is not currently using Find2.0 in the "CI", i test it locally.  

Transposed convolution is called Deconvolution in MIGraphX and for that, I have a local branch.

umangyadav (dismissed):

DrizztDoUrden:
In MIOpen CI there are tests for both regular and transposed convolutions as there are for both find 1.0, find 2.0 and immediate mode.

umangyadav (approved):

src/problem.cpp
```diff
@@ -142,6 +144,43 @@ const TensorDescriptor& Problem::GetTensorDescriptorChecked(miopenTensorArgument
     return found->second;
 }
 
+Problem Problem::Transpose() const
```
atamazov, line 147:
```suggestion
Problem Problem::MakeTransposed() const
```
or maybe even `ConstructTransposed`. `Transpose` suggests that this method modifies existing object, which is not so.

src/problem.cpp
```diff
@@ -165,21 +204,21 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
...
 
+    // These are not swapped for now to preserve argument order in calls
     const auto& x_desc =
         GetTensorDescriptorChecked(miopenTensorConvolutionX, "miopenTensorConvolutionX");
     const auto& w_desc =
         GetTensorDescriptorChecked(miopenTensorConvolutionW, "miopenTensorConvolutionW");
     const auto& y_desc =
         GetTensorDescriptorChecked(miopenTensorConvolutionY, "miopenTensorConvolutionY");
 
-    ValidateGroupCount(x_desc, w_desc, conv_desc);
```
atamazov, line 181: Why this is removed?

src/problem.cpp
```diff
@@ -165,21 +204,21 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
...
 {
-    auto ret = std::vector<Solution>{};
+    const auto& actual = conv_desc.mode == miopenTranspose ? Transpose() : *this;
+    auto ret           = std::vector<Solution>{};
 
     if(tensor_descriptors.size() != 3)
         MIOPEN_THROW(miopenStatusInvalidValue,
                      "Convolution problem should have exactly three tensor descriptors.");
 
+    // These are not swapped for now to preserve argument order in calls
```
atamazov, line 214: Wow. It seems like we do not have a proper design for transposed convolutions. Let's discuss.

atamazov (changes requested):
Intermediate reivew

src/problem.cpp
DrizztDoUrden: This would get cleaned up in the next find 2.0 PR using the new `FindConvolution` method from #1739 rather than `Find...Convolution`, which are other API mappings and should only be used as such.
These are not swapped to make calls structure exactly same as in other API mappings until the whole approach would be changed very soon.

src/problem.cpp
atamazov: Resolved!

atamazov (dismissed):
🟢 Navi21 FP32 performance/correctness testing Ok: no regressions.

atamazov (approved):
LGTM!

umangyadav (approved):

junliume (approved):

PR #1801, opened 2022-10-03
Fixed MLIR kernel count retrieval

manupak:
This commit fixes the kernel count retrieval
to include perf_config used as part of the string
that is generated to pass to the Miir API.

Note this does not change the API per se but
includes extra information (additive) to passed
on to rocMLIR side.

### DIFF
--- a/src/solver/conv_mlir_igemm_bwd.cpp
+++ b/src/solver/conv_mlir_igemm_bwd.cpp
@@ -90,7 +90,7 @@ ConvSolution ConvMlirIgemmBwd::GetSolution(const ConvolutionContext& ctx,
 {
 #if MIOPEN_USE_MLIR
     ConvSolution result;
-    int kernel_count = MiirGetKernelCount(mlir::ConstructBuildOptions(ctx, false));
+    int kernel_count = MiirGetKernelCount(mlir::ConstructBuildOptions(ctx, config, false));
 
     for(int kernel_id = 0; kernel_id < kernel_count; ++kernel_id)
     {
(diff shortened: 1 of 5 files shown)


manupak:
This is my first PR for MIOpen and not familiar with the process. 
Any pointers on the next stepas to incorporate these changes to MIOpen would be really appreciated.

jerryyin:
Your changes looks good to me. I think one solver is missing from the list: src/solver/conv_mlir_igemm_wrw.cpp

atamazov (changes requested):
Please resolve https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1801#issuecomment-1265411098

manupak:
@jerryyin ,

I could not find a call to MiirGetKernelCount there...

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/solver/conv_mlir_igemm_wrw.cpp#L89-L119

jerryyin (approved):
I just double checked our wrw kernel count decision and you are right nonxdlops kernel will always be 1. I think we are good with this PR since we are consistent. https://github.com/ROCmSoftwarePlatform/rocMLIR/blob/d24b416757331193d1fec548e5ad6d32ffb0dfad/mlir/lib/Dialect/Rock/Generator/Conv2dGenerator.cpp#L305-L338

atamazov (approved):
LGTM!

manupak:
I just rebased and pushed the branch as there were some CI failure that seemed unrelated.

junliume (approved):

jerryyin:
@manupak Now that this is merged, could you file the follow-up PR to dial our tracking commit of MIOpen back to develop?

manupak:
@jerryyin will do!

PR #1827, opened 2022-10-17
[SWDEV-358495] Exclude gtest from all Target

JehandadKhan:


junliume (approved):
manually verified twice

Issue #1731, opened 2022-09-01
MIOpen using clang-14 fails 4 tests: "invalid HSA metadata", 1 


littlewu2508:
I'm packeging MIOpen 5.1.3 for gentoo, and noticing there are test failures.

All errors are "invalid HSA metadata" when compiling assemblies:

```
AMD_COMGR_SAVE_TEMPS=1 AMD_COMGR_REDIRECT_LOGS=stdout AMD_COMGR_EMIT_VERBOSE /test_immed_conv2d --float --cmode conv --pmode default --group-count 1 --enable-fdb 0 --batch_size 8 --input_channels 16 --output_channels 64 --spatial_dim_elements 28 28 --filter_dims 5 5 --pads_strides_dilations 1 1 1 1 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW

...

COMGR::executeInProcessDriver argv: clang "-cc1as" "-triple" "amdgcn-amd-amdhsa" "-filetype" "obj" "-main-file-name" "Conv_Winograd_v21_1_3_fp32_stride1.s" "-target-cpu" "gfx1031" "-target-feature" "+wavefrontsize64" "-target-feature" "+cumode" "-I" "/tmp/comgr-1d8e33/include" "-fdebug-compilation-dir=/home/selene" "-dwarf-debug-producer" "clang version 14.0.6" "-I" "/tmp/comgr-1d8e33/include" "-dwarf-version=5" "-mrelocation-model" "pic" "-defsym" "ROCM_METADATA_VERSION=5" "--mrelax-relocations" "-o" "/tmp/comgr-1d8e33/output/Conv_Winograd_v21_1_3_fp32_stride1.s.o" "/tmp/comgr-1d8e33/input/Conv_Winograd_v21_1_3_fp32_stride1.s"

<instantiation>:67:21: error: invalid HSA metadata
.end_amdgpu_metadata
    |   |   |   |   ^
<instantiation>:1:1: note: while in macro instantiation
METADATA %total_sgpr_count, %.amdgcn.next_free_vgpr, %workgroup_size_x, miopenSp3AsmConv_v21_1_3_gfx10_fp32_stride1
^
<instantiation>:93:1: note: while in macro instantiation
METADATA_WRAPPER total_sgpr_count,.amdgcn.next_free_vgpr,workgroup_size_x, <miopenSp3AsmConv_v21_1_3_gfx10_fp32_stride1>
^
[...]

PR #1816, opened 2022-10-07
[Workaround] bug in the llvm::msgpack when N is recognized as "false", a boolean type instead of string type.

atamazov:
Fixes #1731.

@junliume Please attribute this with:
- Reviewers: @littlewu2508, @Slimakanzer, @Kirpich30000
- `non-miopen-bug`, `urgency_high` (borrowed from #1731)

PR #1830, opened 2022-10-17
[Workaround] bug in the llvm::msgpack when "n" and "y" are recognized as a boolean instead of a string. 

atamazov:
More fixes to resolve issue #1731 (why: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1816#issuecomment-1272733599)

@junliume Please attribute this with:
- Reviewers: @littlewu2508, @carlushuang @shaojiewang 
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/non-miopen-bug)https://github.com/ROCmSoftwarePlatform/MIOpen/labels/non-miopen-bug https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_high (borrowed from #1731)


junliume:
Ping Reviewers: @littlewu2508, @carlushuang @shaojiewang

CI has passed

junliume (approved):
@atamazov although very simple ones, it is a change to ~4k files after all. Is there any other alternatives to this workaround?

shaojiewang (approved):
I don't have enough time to check every single file. But code looks good. It means that we should also change the codegen part of MISA.

littlewu2508:
I'll go test this with upstream clang

littlewu2508:
Tests passed for Gentoo miopen-5.1.3 package, compiled using clang-14, so I suppose this patch fixes https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1731

I'll test miopen-5.3.0 upon clang-15 in the future.

atamazov:
@junliume 
> @atamazov although very simple ones, it is a change to ~4k files after all. Is there any other alternatives to this workaround?

No. And we also need to update the generator of igemm gtc kernels so it won't produce `N`, `Y`, `n`, `y` as names of kernel parameters (it is recommended to produce `N_`, `Y_`, `n_`, `y_` instead). That is why @shaojiewang and @carlushuang were invited here.

atamazov:
@junliume Thanks!

PR #1839, opened 2022-10-21
[Find 2.0] Added support for attaching pre-allocated buffers to find options

DrizztDoUrden:
This should ease transitioning from find 1.0 to find 2.0 for the consumers and reduce memory usage for the ones who always allocate buffers.
Adds two new API methods to allow attaching allocated tensor and workspace buffers to the find options. They would be used instead of allocating memory by the find call itself. Any buffer skipped would be allocated instead as before.


include/miopen/miopen.h
```diff
@@ -4848,6 +4848,27 @@ miopenStatus_t miopenSetFindOptionResultsOrder(miopenFindOptions_t options,
  */
 miopenStatus_t miopenSetFindOptionWorkspaceLimit(miopenFindOptions_t options, size_t value);
 
+/*! @brief Attachs the preallocated workspace to find options. Allocated by the library by default.
```
atamazov:
```suggestion
/*! @brief Attaches the preallocated workspace to find options. Allocated by the library by default.
```

test/find_2_conv.cpp
```diff
@@ -160,30 +160,79 @@ struct Find2Test : test_driver
...
-                EXPECT_EQUAL(
-                    miopenFindSolutions(
-                        handle, problem, options, solutions.data(), &found, solutions.size()),
-                    miopenStatusSuccess);
-
-                EXPECT_EQUAL(miopenDestroyFindOptions(options), miopenStatusSuccess);
-            }
+        for(const auto search : {0, 1})
+            for(const auto workspace_limit : {std::numeric_limits<std::size_t>::max(), 0ul})
+                for(const auto prealloc : {true, false})
```
atamazov: I am afraid this is too much, as this doubles testing work (and time spent for CI). Maybe we can use pre-allocated buffers in the half of the test passes, e.g. in the even passes?

atamazov (changes requested):
Maybe we can we have a `--preallocate` option for the test (and print its value with `--verbose`)?

atamazov:
@DrizztDoUrden Please add `[Find2.0]` tag to the title.

test/find_2_conv.cpp
DrizztDoUrden: This test only runs for one config

atamazov: Good. Resolved.

atamazov (dismissed):
LGTM!

JehandadKhan (dismissed):
This PR adds new API which requires design update !

pfultz2:
This also looks good to me, and is the correct approach to extending options to the API.

atamazov:
@DrizztDoUrden 

> Maybe we can we have a `--preallocate` option for the test (and print its value with `--verbose`)?

This is necessary in order to precisely reproduce failures, if any. Otherwise fine.

atamazov (dismissed):
LGTM!

DrizztDoUrden:
Ping for reviews

JehandadKhan (dismissed):

JehandadKhan:
@junliume This can go in

include/miopen/miopen.h
```diff
@@ -4857,6 +4857,27 @@ miopenStatus_t miopenSetFindOptionResultsOrder(miopenFindOptions_t options,
  */
 miopenStatus_t miopenSetFindOptionWorkspaceLimit(miopenFindOptions_t options, size_t value);
 
+/*! @brief Attaches the preallocated workspace to find options. Allocated by the library by default.
+ *
+ * @param options    Options object to upfate
```
umangyadav:
```suggestion
 * @param options    Options object to update
```

include/miopen/miopen.h
```diff
@@ -4857,6 +4857,27 @@ miopenStatus_t miopenSetFindOptionResultsOrder(miopenFindOptions_t options,
...
+ * @param buffer     Specifies the workspace for find call
+ * @param size       Specifies the size of the buffer passed
+ * @return           miopenStatus_t
+ */
+miopenStatus_t
+miopenSetFindOptionPreallocatedWorkspace(miopenFindOptions_t options, void* buffer, size_t size);
+
+/*! @brief Attaches a preallocated tensor to find options. Allocated by the library by default.
+ *
+ * @param options    Options object to upfate
```
umangyadav:
```suggestion
 * @param options    Options object to update
```

include/miopen/miopen.h
```diff
@@ -4857,6 +4857,27 @@ miopenStatus_t miopenSetFindOptionResultsOrder(miopenFindOptions_t options,
...
+ * @param buffer     Specifies the workspace for find call
+ * @param size       Specifies the size of the buffer passed
+ * @return           miopenStatus_t
+ */
+miopenStatus_t
+miopenSetFindOptionPreallocatedWorkspace(miopenFindOptions_t options, void* buffer, size_t size);
+
+/*! @brief Attaches a preallocated tensor to find options. Allocated by the library by default.
```
umangyadav:
```suggestion
/*! @brief Attaches a preallocated tensor to find options. If not used, buffers are allocated by MIOpen internally by default.
```

src/problem.cpp
```diff
@@ -111,7 +120,8 @@ Problem::FindSolutions(Handle& handle, const FindOptions& options, std::size_t m
             SetTensor(handle, descriptor, buffer.get(), &zero);
         });
 
-        buffers.emplace(pair.first, std::move(buffer));
+        buffers.emplace(pair.first, buffer.get());
```
umangyadav, line 123: Why does this one (owned buffer) not require TensorID as key v/s other one does?

include/miopen/miopen.h
atamazov: UPFATE! 👍 💣 
Why can't we keep this? )
professor_fate_100x200

include/miopen/miopen.h
atamazov: Maybe also replace "by default" with "(not recommended)"?

src/include/miopen/sequences.hpp
```diff
@@ -32,10 +32,10 @@
 #include <boost/range/algorithm/find.hpp>
 
 #include <algorithm>
+#include <array>
 #include <cassert>
 #include <tuple>
 #include <vector>
-#include <array>
```
junliume, line 38: @DrizztDoUrden may I ask why this line is removed?

DrizztDoUrden: I don't remember how it got two lines of `#include <array>`, but looks like at some point I have accidentally added a second one. Than I have removed the one which placed in wrong line (not alphabetically sorted). I can revert it to how it has been before if you want.

junliume: Ah I see now. Thanks! I took a glance at it and thought it was removed. I recall adding similar edit due to Ubuntu 22.04 update, but it is not the same issue.

atamazov (approved):
LGTM!

src/include/miopen/sequences.hpp
DrizztDoUrden: Maybe I've got the second line from merge. Any way, it got sorted alphabetically by me. xD

JehandadKhan (approved):

DrizztDoUrden:
ping @junliume

atamazov:
@junliume @JehandadKhan @pfultz2 Thank you.

Issue #1800, opened 2022-09-30
[Find2.0][MIGraphX] Quant Convolution fails with Find 2.0


umangyadav:
For the quant convolution output tensor descriptor is of type int32 and innputs are either int8 or int8x4. 

Find2.0 uses "MIOpenSubTensorOpWithScalarKernel.cl" program which doesn't support int32 type. 

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/5e78e715b174edc2119560f41dcababeb1957fac/src/ocl/tensorocl.cpp#L1438

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/5e78e715b174edc2119560f41dcababeb1957fac/src/kernels/MIOpenSubTensorOpWithScalarKernel.cl#L55

I was able to make MIGraphX quant convolution work by adding INT32 type in that file.

```
#if MIOPEN_USE_INT32 == 1
#define _FLOAT int
#ifndef FLT_MAX 2,147,483,647
#define MAX_VAL 
#else 
#define MAX_VAL FLT_MAX
#endif
#endif
```

Even though it  works for me, i am not sure if this is the right fix. Can we get proper fix in MIOpen to handle this case ?

junliume:
@DrizztDoUrden @atamazov can we take a quick look at this issue?

atamazov:
@umangyadav Does it work with Find 1.0 without your fix? If yes, then it is @DrizztDoUrden who can provide the proper fix soon.

Can we reproduce the issue using our tests or driver? Easy reproduction is important for speed.

umangyadav:
@atamazov Yes it works with find1.0 without my fix.

umangyadav:
@atamazov For the production, i am not much aware with unit-tests or driver  on MIOpen side. Does miopen-driver suppport find.2.0 ? 

Otherwise, i can locally test it with MIGraphX once the PR is up.

atamazov:
@umangyadav 

> @atamazov For the production, i am not much aware with unit-tests or driver on MIOpen side. Does miopen-driver suppport find.2.0 ?

In #1503 (that you have partially reviewed) you can see new Find2.0 specific tests (so now we have `test_con2d_find2` and `test_conv3d_find2`). If you have difficulties with those (e.g. I am not sure that INT32 is supported), just let us know failing convolution configs. You can use MIOpenDriver's syntax as if you use Find 1.0. Or whatever format that you like.

[...]

PR #1850, opened 2022-10-26
[API change] Removed automatic buffer allocation from find 2.0

DrizztDoUrden:
Changes the signature of `miopenFindSolutions` to accept the buffers allocated by the user. This allows us not to allocate them inside the library.
This is a better way to implement #1839, solving #1800. It requires changing an existing API function in a breaking way. This should not affect many customers as find 2.0 is not widely used yet.

### DIFF
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -4877,10 +4877,10 @@ struct miopenTensorArgument_t
  * @param problem       Problem to solve
  * @param options       Find options. When null default values would be used
  * @param tensors       Buffers to use for kernel profiling. "descriptor" field is not used
- * @param workspace     Workspace to use for kernel profiling
+ * @param workspace     Workspace to use for kernel benchmarking
  * @param workspaceSize Workspace size
  * @param solutions     Pointer to the first result. Must not be null
- * @param numSolutions  Pointer to the amount of results. Ignored if null
+ * @param numSolutions  Pointer to the number of the returned results. Ignored if null
  * @param maxSolutions  Limits the amount of results
  * @return              miopenStatus_t
  */
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -4852,24 +4852,6 @@ miopenStatus_t miopenSetFindOptionWorkspaceLimit(miopenFindOptions_t options, si
  */
 MIOPEN_DECLARE_OBJECT(miopenSolution);
 
-/*! @brief Finds solutions to a problem by running different applicable solutions. Memory is
- * automatically allocated.
- *
- * @param handle       Handle to execute the kernels
- * @param problem      Problem to solve
- * @param options      Find options. When null default values would be used
- * @param solutions    Pointer to the first result. Must not be null
- * @param numSolutions Pointer to the amount of results. Ignored if null
- * @param maxSolutions Limits the amount of results
- * @return             miopenStatus_t
- */
-miopenStatus_t miopenFindSolutions(miopenHandle_t handle,
-                                   miopenProblem_t problem,
-                                   miopenFindOptions_t options,
-                                   miopenSolution_t* solutions,
-                                   size_t* numSolutions,
-                                   size_t maxSolutions);
-
 /*! @brief Values of a tensor argument for the miopenRunSolution function.
  */
 struct miopenTensorArgument_t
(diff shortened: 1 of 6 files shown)


atamazov:
[HINT] 💡 @DrizztDoUrden 
> This should not affect many customers as find 2.0 is not widely used yet.

We can explicitly assign an experimental status to 2.0 API by enclosing its declarations into

```C++
#ifndef MIOPEN_API_EXPERIMENTAL_FIND_2_0_ENABLE
...
#endif
```
All users who want to use the new API have to explicitly enable it by defining the macro. This way we can force them to clearly understand that the API is experimental and thus might be changed without notice.

include/miopen/miopen.h
```diff
@@ -4888,6 +4870,31 @@ struct miopenTensorArgument_t
...
 };
 
+/*! @brief Finds solutions to a problem by running different applicable solutions. Memory is
+ * automatically allocated.
+ *
+ * @param handle        Handle to execute the kernels
+ * @param problem       Problem to solve
+ * @param options       Find options. When null default values would be used
+ * @param tensors       Buffers to use for kernel profiling. "descriptor" field is not used
+ * @param workspace     Workspace to use for kernel profiling
```
atamazov: [Recommendation]
```suggestion
 * @param workspace     Workspace to use for kernel benchmarking
```

include/miopen/miopen.h
```diff
@@ -4888,6 +4870,31 @@ struct miopenTensorArgument_t
...
+ * @param workspaceSize Workspace size
+ * @param solutions     Pointer to the first result. Must not be null
+ * @param numSolutions  Pointer to the amount of results. Ignored if null
```
atamazov: [Recommendation]
```suggestion
 * @param numSolutions  Pointer to the number of the returned results. Ignored if null
```

atamazov (commented):
(Intermediate review)

atamazov (approved):
LGTM!

atamazov:
@umangyadav @JehandadKhan @junliume My approval does not allow merging. Please approve.

junliume (approved):

umangyadav (approved):
Looks good to me.

JehandadKhan:
@junliume  This can go in

pfultz2:
This breaks binary compatibility. The SO major version should be bumped to 2.0:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/CMakeLists.txt#L59

atamazov:
@pfultz2 Is it really important? IIRC this feature is experimental and not released yet. Or there are some internal users like MIGraphX for whom this is important?

pfultz2:
> IIRC this feature is experimental and not released yet.

It was released in rocm 5.4 and MIGraphX is using this API in 5.4 as well. It was never mentioned that this API was "experimental" until now. If it was meant to be experimental than all the functions and types should be prefixed with "experimental". Otherwise, you will have more users than just migraphx start using this API expecting it to be stable.

Either way, the major version should be bumped since a newer MIOpen is incompatible with the migraphx binaries in 5.4. Usually we wait until a major bump in rocm for this(such as rocm 6.0).

atamazov:
@pfultz2 Do you recommend implementing #1892 ?

atamazov:
@DrizztDoUrden Please implement https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1850#issuecomment-1329801094

pfultz2:
> Do you recommend implementing https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1892 ?

That wont help binary compatibility. A newer SO version is what is needed to prevent the runtime from loading MIOpen with an older migraphx.

junliume:
> @DrizztDoUrden Please implement [#1850 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1850#issuecomment-1329801094)

@DrizztDoUrden could you have a discussion with @JehandadKhan and @atamazov @pfultz2 on the change? we want to make sure it will not break any user cases.

atamazov:
@pfultz2 
> That wont help binary compatibility. A newer SO version is what is needed to prevent the runtime from loading MIOpen with an older migraphx.

This is obvious. The question is would you like to keep new migraphx compatible with old miopen at the source code level. For example, you may wish to perform A/B comparison of the newest migraphx with old and new miopen.

DrizztDoUrden:
@junliume the idea was originally by @JehandadKhan. When we discussed that we came to that the whole allocation stuff was an oversight at the original design phase.

> we want to make sure it will not break any user cases.

This depends on what you call a use case here. It would break binary and source compatibility with user applications originally using older version of MIOpen AND using find 2.0. The idea is that right now people don't have much benefit from converting existing applications from find 1.0 outside of very specific cases and few would do so. As time passes and features arrive more would do so.

pfultz2:
This is part of UIF releases and so cant be treated as "experimental". I think this should be rejected and backwards-compatible way should be added for this. #1839 looks like it would provide this backwards-compatible change. 

The opaque find options object was designed so we could easily extend the API with new options in a backwards-compatible way. So #1839 seems to be the correct approach here.

atamazov:
@pfultz2 What is UIF release?

I think that API where the library allocates buffers itself is conceptually flawed. Such an API should not exist and be used at all. Therefore, I think that spending resources on maintaining compatibility with it requires business justification at least.

pfultz2:
> I think that API where the library allocates buffers itself is conceptually flawed. Such an API should not exist and be used at all.

But its already being used. 

> Therefore, I think that spending resources on maintaining compatibility with it requires business justification at least.

The API doesnt need to support allocating buffers long-term. It can throw an error if no buffer is passed.

Although, short-term it should at least support this behavior during a deprecation period, as we cant make breaking changes for APIs that are being actively used. 

Either way, it should be done in a backwards-compatible way for both the API and ABI.

atamazov:
Agreed.

JehandadKhan:
@DrizztDoUrden I think we can close this in lieu of #1839

PR #1865, opened 2022-11-07
[Clang-16][D131307] suppress -Wenum-constexpr-conversion

junliume:
https://reviews.llvm.org/D131307

It is also causing problems in MIOpen and MIOpen dependencies:
```
[2022-10-29T14:44:59.883Z] ./boost/mpl/aux_/integral_wrapper.hpp:73:31: error: integer value -1 is outside the valid range of values [0, 3] for this enumeration type [-Wenum-constexpr-conversion]
```

### DIFF
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -463,10 +463,10 @@ static void PrintVersion()
     std::ignore            = once;
 }
 
-static std::string GetStatusText(const amd_comgr_status_t status)
+static std::string GetStatusText(const amd_comgr_status_t status, const bool unknown_error = false)
 {
     const char* reason = nullptr;
-    if(AMD_COMGR_STATUS_SUCCESS != amd_comgr_status_string(status, &reason))
+    if(unknown_error || AMD_COMGR_STATUS_SUCCESS != amd_comgr_status_string(status, &reason))
         reason = "<Unknown>";
     return std::string(reason) + " (" + std::to_string(static_cast<int>(status)) + ')';
 }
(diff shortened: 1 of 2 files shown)


src/comgr.cpp
```diff
@@ -876,7 +876,11 @@ void BuildHip(const std::string& name,
             action.Do(AMD_COMGR_ACTION_LINK_RELOCATABLE_TO_EXECUTABLE, relocatable, exe);
         }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wenum-constexpr-conversion"
         constexpr auto INTENTIONALY_UNKNOWN = static_cast<amd_comgr_status_t>(0xffff);
```
junliume: @atamazov any other ways to fix these similar issues than to suppress them?

atamazov:
@junliume Just in case, according to https://reviews.llvm.org/D131307, it should be possible to use `-Wno-error=enum-constexpr-conversion` for transition period even with clang-16

> any other ways to fix these similar issues than to suppress them?

I am going to provide a patch with a proper fix for COMGR. WRT other cases (if any) it depends on design. Basically, it should be always possible.

atamazov:
@junliume Please apply this patch to your branch and see if this helps:

* Revert COMGR changes
* COMGR: proper fix for enum-constexpr-conversion (pass "unknown problem" flag explicitly)

[comgr-fix-enum-constexpr-conversion.diff.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/9955768/comgr-fix-enum-constexpr-conversion.diff.txt)

atamazov:
@junliume ...alternatively, you can merge into this branch from https://github.com/atamazov/MIOpen/tree/clang_16_fix_ata-temp-0

atamazov (approved):
LGTM!

averinevg (approved):
LGTM

PR #1868, opened 2022-11-08
[solvers] SolverBase: split ConvolutionContext and ProblemDescription (stage 3)

averinevg:
PR splits `ConvolutionContext` and `ProblemDescription` inside `SolverBase`

Additional changes:
- Activ solvers: `OldStyleProblemDescription` was removed
- Batchnorm solvers: `OldStyleProblemDescription` was removed
- Pooling solvers: `OldStyleProblemDescription` was removed
- Missing headers were added


src/convolution.cpp
```diff
@@ -305,7 +305,7 @@ bool ConvolutionDescriptor::IsWinograd3x3SupportedAndFast(miopen::ConvolutionCon
     if(!(ctx.problem.n_outputs >= 16 && ctx.problem.n_outputs % 2 == 0))
         return false;
 
-    return solver::ConvBinWinograd3x3U{}.IsApplicable(ctx);
+    return solver::ConvBinWinograd3x3U{}.IsApplicable(ctx, ctx.problem);
```
JehandadKhan, line 308: This seems rather pointless ? Can we get some background on how this is helpful?

JehandadKhan (changes requested):

src/convolution.cpp
atamazov: @JehandadKhan This is a leftover from old good heuristics. Please look at PR #1952 in the private repo, maybe it contains some explanations. IIRC it covers cases when WTI of Winograd >= 100% (or even close to 200%), so I would better keep it for now. However, nothing prevents us from removing it (but let's do that separately and thoroughly test before merging).

/cc @Kirpich30000

JehandadKhan: @atamazov  or @averinevg  Can you please educate me on the usefulness of these changes and what is the larger goal here. Superficially, these PRs dont seem to add any value however, they touch many many files and once merged they induce a lot of churn for any inprogress work.

atamazov: @JehandadKhan I guess you mean usefulness of the series of the refactoring PRs (and this specific comment thread related to `IsWinograd3x3SupportedAndFast()` can be considered resolved).

@DrizztDoUrden @averinevg Do we have a ticket or comment where the reasons and expected values are described? Let's share our values with the rest of the team ;)

averinevg: @JehandadKhan My main goal is to make common interface for all solvers. There are several difficulties on the way to completing this task. Current `ConvolutionContext` implementation is one of them. This comment inspired me:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/88aee19b4a152f4f29fbea940607cba456067db6/src/include/miopen/conv/context.hpp#L41-L44
I absolutely agree with this. These objects should not be combined into one, because this only adds spaghetti code to the library. This work was started before my PRs, many solvers (activ, pooling, batchnorm, some of ocl and others) already had separated `Context` and `ProblemDescription`.
As for my work, such _combined_ implementation makes it impossible to complete this task.

About this line:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ce5bb0e8760cfb22654d733d748c2a861e342829/src/convolution.cpp#L308
This is a temporary step. I have split this task into several sub-tasks. `IsWinograd3x3SupportedAndFast()` is not used inside solvers and has not been changed yet.

One last thing, just FYI, after completing this task, `ConvolutionContext` and `ProblemDescription` will become completely independent.

DrizztDoUrden: @JehandadKhan There are two main reasons for the changes:
- ConvolutionContext shouldn't contain handle as it's not part of a convolution and there are a lot of methods it shouldn't be passed to.
- Having solver methods in form of `F(const ExecutionContext&, const AnyProblem&)` makes the solvers API easy to understand, unify and reuse between primitives. The public wrapper methods would be easily changed to type-erased problems and would be reuseable by non-templated code anywhere.

When the approach has been designed, there were not that many solvers. In fact, there were literally zero, because current way of writing them (like in this PR series and further) is very close the original, which was switched to templated for "performance" reasons and blocked from changing it back until a massive amount of work piled up to revert that. The amount of work has risen significantly since a lot of people adopt deprecated approaches. For example, mlo_construct_... have been used and updated up until lately despite being heavily deprecated for many years now.

DrizztDoUrden (dismissed):
LGTM

src/convolution.cpp
JehandadKhan: @averinevg and @DrizztDoUrden 

> My main goal is to make common interface for all solvers.

Isnt Solver base such an interface ? If not, did we define such an interface and discuss it so that an agreement can be reached ? Is such an interface documented somewhere providing the aim to which you are working ? 

> One last thing, just FYI, after completing this task, ConvolutionContext and ProblemDescription will become completely independent.

Lets assume this is done, what did we achieve ? What functionality did not exist before and is now viable ? 

> When the approach has been designed, there were not that many solvers. In fact, there were literally zero, because current way of writing them (like in this PR series and further) is very close the original, which was switched to templated for "performance" reasons and blocked from changing it back until a massive amount of work piled up to revert that

I do not agree with this assertion, in fact looking at the solver infrastructure I can see that there is a base virtual class from which all solvers derive now, I believe the CRTP pattern written on top of that pure virtual base class was more due to avoid making large code changes, instead of a some design choice. 

Furthermore, I always assumed that the solver registry would form the pinnacle of this design, however, after considerable work and changes merged into MIOpen, the registry is still only functional for the Convolution primitives and nothing else? The `AnySolver` type erasure and templated `SolverContainer` is still necessary to drive find. What I fail to understand is how are these changes getting us close to the final design, if there is one.

averinevg: @JehandadKhan 

>> My main goal is to make common interface for all solvers.

> Isnt Solver base such an interface ?

Initially yes. But for now, I'm going to use it to determine whether it should be cast to tunable or non-tunable solver base class. The difficulty is that tunable and non-tunable solvers have different sets of methods. If we combine them into one, then this will introduce even more uncertainty.

> If not, did we define such an interface and discuss it so that an agreement can be reached ? Is such an interface documented somewhere providing the aim to which you are working ?

I've discussed the design in general terms with @DrizztDoUrden and shared my plans with @atamazov. In short, my aim now is no make base classes for all types used in the interface. It would be clearer than using things like `boost::any`. We can discuss details in private.

>> ConvolutionContext and ProblemDescription will become completely independent.

> Lets assume this is done, what did we achieve ? What functionality did not exist before and is now viable ?

Well, first of all, they are completely different types. Second, we have several types of contexts that are designed in two different ways, and because of this, they cannot be generalized. This is not a question of functionality, but of clarity and unification.

> I do not agree with this assertion, in fact looking at the solver infrastructure I can see that there is a base virtual class from which all solvers derive now, I believe the CRTP pattern written on top of that pure virtual base class was more due to avoid making large code changes, instead of a some design choice.

This is not quite CRTP, performance config class is passed as a template parameter, not a solver class itself. But I am not against CRTP. All previous changes are related to the separation of classes and do not break the current inheritance hierarchy in any way.

> ... What I fail to understand is how are these changes getting us close to the final design, if there is one.

It is impossible to build a roof if there is no basement. It is not possible to use SolverBase when it isn't fully implemented.

JehandadKhan: From your comments above, I still dont follow what is the final design. I look forward to seeing the final design that you have envisioned.

atamazov: [Note] @JehandadKhan We want more flexibility. For example, we want to combine different kind of Problems into a vector in order to form a Fused problem (simplest case).

JehandadKhan: @averinevg and @atamazov  Thanks for your patience and support. Please create an issue to help us understand, what the overall plan is. More specifically: 
* How do these changes relate the over all goal of being able to put all Problems in a vector?
* How is this helping us getting rid of any solver ? 
* How is this helping us make the solver registry more effective? (Currently its only useful for Convolutions)
* What is lacking in the current [`SolverBase`](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/9fc10593bb61abd39b7e2e5419eeca90e18c4427/src/include/miopen/solver.hpp#L76) to achieve the goals listed above and how does this change help us. 

Lets move this discussion to a separate issue to help us document better.

averinevg: @JehandadKhan Please take a look at https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1887

junliume:
@averinevg could you please fix the conflicts and review opinions above? Thanks! :)

src/convolution.cpp
junliume: @atamazov  @averinevg : is there an issue with which we can discuss the goal of the solver refactoring? If non-exist yet could you please create one? Thanks!

averinevg: > @atamazov @averinevg : is there an issue with which we can discuss the goal of the solver refactoring? If non-exist yet could you please create one? Thanks!

@junliume https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1887

averinevg:
PR closed in favor of https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2009

PR #1872, opened 2022-11-10
Fix for identifying LegacySolver

cderb:
Fix for identifying LegacySolver in any_solver.hpp
Previous template was returning true for non LegacySolver types

### DIFF
--- a/src/include/miopen/any_solver.hpp
+++ b/src/include/miopen/any_solver.hpp
@@ -159,10 +159,9 @@ struct AnySolver
         {
             template <typename U>
             static constexpr auto Test(U*) ->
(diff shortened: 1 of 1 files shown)


junliume:
@cderb will this bug impact user-side behavior? I'm trying to identify if it should be cherry-picked.

JehandadKhan (approved):

cderb:
> @cderb will this bug impact user-side behavior? I'm trying to identify if it should be cherry-picked.

@junliume as of now it only affects tuning with Fin. Users would not normally encounter this.

junliume (approved):

PR #1873, opened 2022-11-10
[MI100][MI200][Fambench][SWDEV-365737] Retune

cderb:
rocm: 5.5.0-11080
miopen: https://github.com/ROCmSoftwarePlatform/MIOpen/commit/88aee19b4a152f4f29fbea940607cba456067db6

Leverages #1872 to recover solvers in tuning.
Significant performance recovery in cases where missing solver was fastest.


jataylo (approved):
LGTM, this seems to get rid of the large inefficient kernels that I observed. 

From a brief observation there are some inefficiencies still remaining but they are much fewer now. I will analyse and get back to you, meanwhile this change seems good!

junliume (approved):

Issue #216, opened 2020-05-13
Design of Invokers


JehandadKhan:
_The internals of our library went some major changes recently with solver registry, invokers etc. It would be helpful to have a design document to understand what the new design looks like._

## Motivation and basic concepts

See [Extend Solver/Solution architecture with Invoker objects. #1127](https://github.com/AMDComputeLibraries/MLOpen/issues/1127)

## Design of Invokers

Initial design is explained in the description of [PR #2192, Invokers core and Forward Direct](https://github.com/AMDComputeLibraries/MLOpen/pull/2192)

## Further reading (more code than text :eyes:)
- Evolution of low-level design in detail can be found in the same PR, beginning from [this comment](https://github.com/AMDComputeLibraries/MLOpen/pull/2192#issuecomment-548544077) and below. There are _lots_ of informative comments that touch different aspects like:
  - Why we want to hold Invokers separately from the kernels
  - Measuring kernel execution time
  - Invokers and moving towards Find API 2.0
  - A notion of _Invoker Factory_
  - ...
- Core Bwd Invokers + implementation for Backward/ Direct algo can be found at [PR #2366](https://github.com/AMDComputeLibraries/MLOpen/pull/2366). 
- Core of WrW Invokers + implementation for WrW/ Direct algo: see [PR #2454](https://github.com/AMDComputeLibraries/MLOpen/pull/2454)
- Why and how computations should be moved from Invoker to InvokerFactory: see [discussion in PR #441](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/441#discussion_r501377143). 
- Capturing values in Invoker/Factory labdas: be careful when using `slice` and similar things, see [note in PR #800](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/800/files#r606448462).

_Please do not hesitate to ask specific questions._

atamazov:
Basic information added. Please do not hesitate to ask specific questions.

atamazov:
"Why and how computations should be moved from Invoker to InvokerFactory" added to the topmost comment.

atamazov:
"Capturing values in Invoker/Factory labdas: be careful when using `slice` and similar things" added.

atamazov:
## On FP16_ALT support in Invokers

FP16_ALT requires support for selection of "FP16 mode" during kernel invocation. The necessary parameter is passed to the Invoker and it is the Invoker who must properly handle it. How it should be done is a design choice:

[...]

PR #1874, opened 2022-11-10
[CK API] 2d backward data convolution composable kernel integration

iq136boy:
1. Integrated 2d backward data convolution composable kernel
2. Adding test cases.


test/CMakeLists.txt
```diff
@@ -753,6 +753,22 @@ add_custom_test(test_conv_hip_igemm_xdlops SKIP_UNLESS_ALL OCL_DISABLED HALF_DIS
...
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-forward --disable-backward-weights --verbose --input 256 128  28 28 --weights 128  128  3 3 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 1 1 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-forward --disable-backward-weights --verbose --input 128 512  7  7  --weights 512  512  3 3 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 1 1 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-forward --disable-backward-weights --verbose --input 128 64   56 56 --weights 64   64   1 1 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 0 0 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-forward --disable-backward-weights --verbose --input 256 256  56 56 --weights 256  64   1 1 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 0 0 1 1 1 1
```
JehandadKhan, line 771: Can we add a GTest instead ?

JehandadKhan (changes requested):

src/include/miopen/solver.hpp
```diff
@@ -5740,6 +5740,110 @@ struct ConvHipImplicitGemmFwdXdlops final
...
+    {
+        return IsApplicable(ctx.problem);
+    }
+    bool IsDynamic() const override { return true; }
+    ConvSolution GetSolution(const ConvolutionContext& ctx,
+                             const PerformanceConfigHipImplicitGemmBwdXdlops& config) const override
+    {
+        return GetSolution(ctx, ctx.problem, config);
+    }
+    float GetWti(const ConvolutionContext&) const override { return 0.02f; };
```
averinevg, line 5831: @iq136boy I think it's worth adding a comment here with an explanation, for example: We use here a value bigger than 0.01f, e.g. 0.02f. See comment above ConvHipImplicitGemmFwdXdlops::GetWti()

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,392 @@
...
+    return GenericSearch(*this, ctx, problem, invoke_ctx);
+}
+
+size_t ConvHipImplicitGemmBwdXdlops::GetWorkspaceSize(const ConvolutionContext& ctx) const
+{
+    std::ignore = ctx;
+    return 0;
+}
+
+bool ConvHipImplicitGemmBwdXdlops::IsApplicable(const ProblemDescription& problem) const
```
averinevg: @iq136boy Does this solver support all GPUs?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,392 @@
...
+                                     const AnyInvokeParams& invoke_ctx) const
+{
+    return GenericSearch(*this, ctx, problem, invoke_ctx);
+}
+
+size_t ConvHipImplicitGemmBwdXdlops::GetWorkspaceSize(const ConvolutionContext& ctx) const
+{
+    std::ignore = ctx;
+    return 0;
+}
```
averinevg: @iq136boy Declaration and definition of `ConvHipImplicitGemmBwdXdlops::GetWorkspaceSize()` can be omitted because default `GetWorkspaceSize()` returns 0

averinevg:
@atamazov Please review

src/include/miopen/solver.hpp
```diff
@@ -5740,6 +5740,110 @@ struct ConvHipImplicitGemmFwdXdlops final
...
+    {
+        return IsValidPerformanceConfig(ctx.problem, config);
+    }
+    PerformanceConfigHipImplicitGemmBwdXdlops
+    Search(const ConvolutionContext& ctx, const AnyInvokeParams& invoke_ctx) const override
+    {
+        return Search(ctx, ctx.problem, invoke_ctx);
+    }
+    size_t GetWorkspaceSize(const ConvolutionContext& ctx) const override;
+    bool MayNeedWorkspace() const override { return false; }
```
averinevg: @iq136boy This line can be omitted. Default `MayNeedWorkspace()` from `SolverBase` returns `false`

src/include/miopen/solver.hpp
iq136boy: Hi @averinevg , you are right.  We do need to add similar comments here.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: No. Only mi100 and mi200. I will add the constraint to the IsApplicable function.

test/CMakeLists.txt
iq136boy: As discussed, this will be addressed in a separate PR.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,391 @@
...
+    case miopenDouble: break;
+    }
+    return false;
+#endif
+}
+
+bool PerformanceConfigHipImplicitGemmBwdXdlops::operator==(
+    const PerformanceConfigHipImplicitGemmBwdXdlops& other) const
+{
+    return this->index == other.index;
```
averinevg: @iq136boy Shouldn't other fields of performance config also be compared?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,391 @@
...
+}
+#endif
+
+void PerformanceConfigHipImplicitGemmBwdXdlops::HeuristicInit(const ProblemDescription& problem)
+{
+    this->index = 0;
+#if !MIOPEN_BACKEND_HIP || !MIOPEN_USE_COMPOSABLEKERNEL
+    std::ignore = problem;
+#else
+    this->index      = 0;
```
bghimireamd: `this->index = 0;`  is initialized twice.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
bghimireamd: `this->kernel_id == other.kernel_id && this->total_size == other.total_size;`

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,391 @@
...
+       problem.conv_problem.GetInDataType() != problem.conv_problem.GetOutDataType())
+        return false;
+    if(!problem.direction.IsBackwardData())
+        return false;
+    if(!problem.Is2d())
+        return false;
+    if(!problem.IsLayoutNHWC())
+        return false;
+    const std::string& arch = ctx.GetStream().GetDeviceName();
+    if(arch != "gfx908" && arch != "gfx90a")
```
bghimireamd: What about `gfx1030`?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,391 @@
...
+    case miopenBFloat16:
+    case miopenDouble: break;
+    }
+#endif
+}
+
+bool PerformanceConfigHipImplicitGemmBwdXdlops::SetNextValue(const ProblemDescription& problem)
+{
+    if(total_size == -1)
+        this->HeuristicInit(problem);
```
atamazov: This looks suspicious. AFAICS `HeuristicInit()` assigns valid value to `index` but we do not use it but increment it instead at line 258.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,391 @@
...
+#endif
+
+void PerformanceConfigHipImplicitGemmBwdXdlops::HeuristicInit(const ProblemDescription& problem)
+{
+    this->index = 0;
+#if !MIOPEN_BACKEND_HIP || !MIOPEN_USE_COMPOSABLEKERNEL
+    std::ignore = problem;
+#else
+    this->index      = 0;
+    this->total_size = 0;
```
bghimireamd: Is it ok to `this->total_size = 0;` even if `Init` fails?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: What is strange is that the only member of PerformanceConfig that is serialized is `kernel_id`, so it seems like actual performance config data is namely the `kernel_id`. But I do not see how (where) `kernel_id` is used after read from database.

What is used in `GetSolution` (well, actually in `RunCKSolution()` called from the Invoker) is `index`.

Can someone remind me how this works?

iq136boy: > @iq136boy Shouldn't other fields of performance config also be compared?

We might need to add the comparison of total_size too. But the kernel_id is not necessary since ck stored kernel_id based on index. If index is the same, kernel_id must be the same.

iq136boy: > What is strange is that the only member of PerformanceConfig that is serialized is `kernel_id`, so it seems like actual performance config data is namely the `kernel_id`. But I do not see how (where) `kernel_id` is used after read from database.
> 
> What is used in `GetSolution` (well, actually in `RunCKSolution()` called from the Invoker) is `index`.
> 
> Can someone remind me how this works?

CK stored the kernel_id info based on the index in a vector. The index is used to access the vector to get all the ck info. Then use the info to run/invoke the kernel. One thing I should mentioned is that CK use their own invoker to invoke the kernel.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: Thanks. I will fix that. Forget to delete it after debugging.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: No, currently CK does not support navi.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: Yes, you are right. I will fix it.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: No, it should be -1. I will delete it. Thanks for pointing this out.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,391 @@
...
+    }
+}
+
+template <typename DataType>
+bool PerformanceConfigHipImplicitGemmBwdXdlops::CheckIsSupportCKArgs(
+    const ProblemDescription& problem) const
+{
+    const auto args      = CKArgsBwd{problem};
+    const auto conv_ptrs = DeviceOpBwdPtrs<DataType>::GetInstances();
+    auto argument_ptr    = conv_ptrs[this->index]->MakeArgumentPointer(nullptr,
```
bghimireamd: Do we need to check if `this->index < conv_ptrs` ?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
bghimireamd: May be miopen log saying CK does not support this `arch`?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: @iq136boy Sure thing; I've already inspected the code. But what I do not understand: when MIOpen reads perf-db, it constructs an instance of PerformanceConfig with `kernel_id` read from the database, right? I see that Invoker needs `index` to run proper kernel. But I do **_not_** see where `index` is initialized from the information read from the database, i.e. from the `kernel_id`. What I am missing?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: @iq136boy I see the same problem in the `PerformanceConfigHipImplicitGemmFwdXdlops`. I recommend fixing it at once.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: Please also check Fwd counterpart of this solver. It uses the same pattern.

/cc @JehandadKhan

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: @bghimireamd Basically this is impossible because the value of `index` should be either validated (somewhere else) or produced from valid information read from perf-db.

@iq136boy If in doubt, please add an assertion (not MIOPEN_THROW)

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: @iq136boy Please also check Fwd counterpart of this solver. It may use the same pattern.

[Recommendation] Implement the fixes for `ConvHipImplicitGemmFwdXdlops` (if any) in a separate PR.

/cc @JehandadKhan

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: Do you think it should say that?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: [Important] @iq136boy 
> We might need to add the comparison of total_size too. 

Why?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -0,0 +1,391 @@
...
+        assert(total_size != -1);
+        return true;
+    }
+    else if((index + 1) < total_size)
+    {
+        ++index;
+        return true;
+    }
+    else
+        return false;
```
atamazov: [Recommendation] We do not need "else" in this func I guess
```suggestion
    if((index + 1) < total_size)
    {
        ++index;
        return true;
    }
    return false;
```

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -287,7 +287,7 @@ bool PerformanceConfigHipImplicitGemmBwdXdlops::IsValid(const ProblemDescription
 bool PerformanceConfigHipImplicitGemmBwdXdlops::operator==(
     const PerformanceConfigHipImplicitGemmBwdXdlops& other) const
 {
-    return this->index == other.index;
+    return this->index == other.index && this->total_size == other.total_size;
```
atamazov: _Topic continued from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1023495820_

[Important] @iq136boy 
> We might need to add the comparison of total_size too. 

Why?

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -335,7 +336,7 @@ bool ConvHipImplicitGemmBwdXdlops::IsApplicable(const ConvolutionContext& ctx,
     if(!problem.IsLayoutNHWC())
         return false;
     const std::string& arch = ctx.GetStream().GetDeviceName();
-    if( arch != "gfx908" || arch != "gfx90a" )
+    if(arch != "gfx908" && arch != "gfx90a")
```
atamazov: [Recommendation] I think this improves readability, but the final decision is up to you.
```suggestion
    if(!(arch == "gfx908" || arch == "gfx90a"))
```

atamazov (changes requested):

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: [Informative] @iq136boy 
> One thing I should mentioned is that CK use their own invoker to invoke the kernel.

Note on terminology. 

Invoker is an object (usually, a lambda) produced by InvokerFactory. InvokerFactory is an object (usually, a lambda as well) produced by `GetSolution()`. We often write lambdas right in `GetSolution()`, so InvokerFactory is an outer lambda, while Invoker is an inner one. More info at #216

Although it is namely `conv_ptr->MakeInvokerPointer()->Run(...)` that calls the kernel, it's only a part of the Invoker.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
bghimireamd: @atamazov  Not sure, it might be overkill, since we have to do the same for all the other solvers's `IsApplicable`.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: > @iq136boy Sure thing; I've already inspected the code. But what I do not understand: when MIOpen reads perf-db, it constructs an instance of PerformanceConfig with `kernel_id` read from the database, right? I see that Invoker needs `index` to run proper kernel. But I do **_not_** see where `index` is initialized from the information read from the database, i.e. from the `kernel_id`. What I am missing?

@atamazov In fact, you are right, we did not initialize/find the index based on the kernel info reading from the perf-db. I will fix it.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: > Please also check Fwd counterpart of this solver. It uses the same pattern.
> 
> /cc @JehandadKhan

@atamazov Got it. I will fix it too.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: > @iq136boy Please also check Fwd counterpart of this solver. It may use the same pattern.
> 
> [Recommendation] Implement the fixes for `ConvHipImplicitGemmFwdXdlops` (if any) in a separate PR.
> 
> /cc @JehandadKhan

@atamazov Sure. I will implement the fixes for ConvHipImplicitGemmFwdXdlops in a separate PR.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
iq136boy: > _Topic continued from [#1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1023495820)_
> 
> [Important] @iq136boy
> 
> > We might need to add the comparison of total_size too.
> 
> Why?

@atamazov This is a bit confusing here. Now I think only to compare the kernel_id since that will guarantee to get the kernel instance we want.

JehandadKhan (approved):

averinevg (approved):

bghimireamd (approved):

junliume (approved):

atamazov:
@iq136boy Were these _important_ comments resolved? I don't see the "resolved" marks (usually I do not look into code changes until a comment is masked as "resolved")

- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026616606
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024432254
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026617305
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024439404
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024280238
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026624355

/cc @junliume

junliume:
> @iq136boy Were these _important_ comments resolved? I don't see the "resolved" marks (usually I do not look into code changes until a comment is masked as "resolved")
> 
> * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026616606)
> * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024432254)
> * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026617305)
> * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024439404)
> * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024280238)
> * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026624355)
> 
> /cc @junliume

ping @iq136boy

test/CMakeLists.txt
```diff
@@ -753,6 +753,22 @@ add_custom_test(test_conv_hip_igemm_xdlops SKIP_UNLESS_ALL OCL_DISABLED HALF_DIS
...
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-backward-data --disable-backward-weights --verbose --input 256 256  56 56 --weights 256  64   1 1 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 0 0 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-forward --disable-backward-weights --verbose --input 256 128  28 28 --weights 128  128  3 3 --output_type fp32 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 1 1 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-forward --disable-backward-weights --verbose --input 128 512  7  7  --weights 512  512  3 3 --output_type fp32 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 1 1 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-forward --disable-backward-weights --verbose --input 128 64   56 56 --weights 64   64   1 1 --output_type fp32 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 0 0 1 1 1 1
```
atamazov, line 766: @iq136boy Please note that these tests are not guaranteed to test CK kernels. The reason is that the library selects the "fastest known" kernels or use Immediate Mode Fallback heuristics to select "likely the best" kernel, and therefore we do not know which Solver will be used. Please employ `MIOPEN_FIND_MODE=normal` and `MIOPEN_DEBUG_FIND_ONLY_SOLVER=...` in order to limit library to work with a known set of Solvers (if you need).

test/CMakeLists.txt
```diff
@@ -753,6 +753,22 @@ add_custom_test(test_conv_hip_igemm_xdlops SKIP_UNLESS_ALL OCL_DISABLED HALF_DIS
...
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-forward --disable-backward-weights --verbose --input 256 256  56 56 --weights 256  64   1 1 --output_type fp32 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 0 0 1 1 1 1
```
atamazov, line 767: @iq136boy Another note, this won't test the new solver on MI200 due to `GFX90A_DISABLED` attribute above.

atamazov (commented):
Tests do not ~guarantee verification of a~ **_test_** the new solver (https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#pullrequestreview-1248682823)

test/CMakeLists.txt
```diff
@@ -753,6 +753,22 @@ add_custom_test(test_conv_hip_igemm_xdlops SKIP_UNLESS_ALL OCL_DISABLED HALF_DIS
...
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-backward-data --disable-backward-weights --verbose --input 256 128  28 28 --weights 128  128  3 3 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 1 1 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-backward-data --disable-backward-weights --verbose --input 128 512  7  7  --weights 512  512  3 3 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 1 1 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-backward-data --disable-backward-weights --verbose --input 128 64   56 56 --weights 64   64   1 1 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 0 0 1 1 1 1
+    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --disable-backward-data --disable-backward-weights --verbose --input 256 256  56 56 --weights 256  64   1 1 --output_type fp16 --in_layout NHWC --fil_layout NHWC --out_layout NHWC --pads_strides_dilations 0 0 1 1 1 1
```
atamazov, line 763: This config is invalid. Silently skipped due to #1934

atamazov (commented):
[tests] The added tests only test the new solver with INT8 input datatype. But this is useless because INT8 is only supported for Forward convolutions. Testcase is silently skipped due to #1934.

test/CMakeLists.txt
atamazov: Expected to be resolved by #1936

iq136boy:
> > @iq136boy Were these _important_ comments resolved? I don't see the "resolved" marks (usually I do not look into code changes until a comment is masked as "resolved")
> > 
> > * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026616606)
> > * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024432254)
> > * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026617305)
> > * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024439404)
> > * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1024280238)
> > * [[CK API] 2d backward data convolution composable kernel integration #1874 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1874#discussion_r1026624355)
> > 
> > /cc @junliume
> 
> ping @iq136boy

@junliume All the comments has been addressed in the codes. I have mark these conversation as 'resolved'.

PR #1875, opened 2022-11-10
Rework Fusion API to use solvers

JehandadKhan:
This PR removes the MetaData Graph logic in MIOpen used to determine the applicable solvers for various fusions. The API and the associated flows remain the same. A few tests have been added to make sure that we are able to ascertain whether solvers work as expected.


atamazov:
@averinevg Can you please begin reviewing? At top level first; let's understand the general design ideas. Thanks!

src/include/miopen/find_solution.hpp
```diff
@@ -137,7 +137,7 @@ template <class Solver, class Context, class Db>
 ConvSolution
 FindSolution(Solver s, const Context& context, Db& db, const AnyInvokeParams& invoke_ctx)
 {
-    static_assert(sizeof(Solver) == sizeof(SolverBase), "Solver must be stateless");
+    // static_assert(sizeof(Solver) == sizeof(SolverBase), "Solver must be stateless");
```
daniellowell: Why remove this check?

JehandadKhan: reverted the error

src/include/miopen/fusion.hpp
```diff
@@ -33,16 +33,22 @@
...
 #include <miopen/op_kernel_args.hpp>
 #include <miopen/fusion_ops.hpp>
+#include <miopen/fusion/fusion_invoke_params.hpp>
 #include <miopen/activ.hpp>
 
 #include <set>
 #include <vector>
 #include <unordered_map>
 
+#define MIOPEN_CHECK(x)          \
```
averinevg: @JehandadKhan Is this macro used anywhere? If so, let's move it to some common header.

```suggestion
#define MIOPEN_CHECK_STATUS(x)          \
```

src/include/miopen/fusion/solvers.hpp
```diff
@@ -0,0 +1,356 @@
...
+#pragma once
+
+#include <miopen/solver.hpp>
+#include <miopen/fusion.hpp>
+#include <miopen/fusion_plan.hpp>
+#include <utility>
+
+namespace miopen {
+
+struct FusionDescription : SQLiteSerializable<FusionDescription>
```
averinevg: I suggest moving `FusionDescription` to its own header file inside fusion directory so as not to overload solvers.hpp

src/include/miopen/fusion/solvers.hpp
```diff
@@ -0,0 +1,356 @@
...
+                    {} /*bn_op.base_desc*/,
+                    not_used,
+                    bn_op.useBatchStats /*useSaved*/};
+        }
+        else
+            MIOPEN_THROW(miopenStatusNotImplemented);
+    }
+};
+
+struct FusionContext : miopen::ExecutionContext
```
averinevg: I suggest moving `FusionContext` to its own header file inside fusion directory so as not to overload solvers.hpp

src/include/miopen/fusion/solvers.hpp
```diff
@@ -0,0 +1,356 @@
...
+    /// The int parameter is needed only to not change the name of the
+    /// function in the derived class. Function declarations that differ
+    /// only by its return type cannot be overloaded.
+    virtual boost::any
+    Search(const FusionContext& ctx, const AnyInvokeParams& invoke_ctx, int) const = 0;
+
+    /// Tunable solvers provide a GetSolution that takes a Context and PerformanceConfig
+    virtual ConvSolution GetSolution(const FusionContext& ctx, const boost::any& config) const = 0;
+};
+template <class PerformanceConfig>
```
averinevg: Please add blank line here

src/include/miopen/fusion/solvers.hpp
```diff
@@ -0,0 +1,356 @@
...
+    PerformanceConfigConvBiasActivAsm1x1U(const bool spare) : PerformanceConfigConvAsm1x1U(spare) {}
+    PerformanceConfigConvBiasActivAsm1x1U()
+        : PerformanceConfigConvAsm1x1U(-1, -1, -1, -1, -1, -1, -1, -1, false)
+    {
+    }
+    void HeuristicInit(const FusionContext& ctx);
+    bool SetNextValue(const FusionContext& problem);
+    bool IsValid(const FusionContext& problem) const;
+};
+struct ConvBiasActivAsm1x1U : FusionTunableSolver<PerformanceConfigConvBiasActivAsm1x1U>
```
averinevg: Please add blank line here

src/include/miopen/fusion_plan.hpp
```diff
@@ -60,36 +65,23 @@ struct FusionPlanDescriptor : miopenFusionPlanDescriptor
...
-    auto GetLocalWGSz();
-    auto GetGlobalWGSz();
-    std::vector<Exec_arg_t> CalcArgOrder(const Handle& handle);
-    bool GetEnumVal(const std::string& sym, int& val) const;
-    OpKernelArg GetDevAttribute(const std::string& k, const Handle& handle) const;
-    OpKernelArg GetTensorAttr(const std::string& sym) const;
-    bool GetTensorAttr(const std::string& sym, int& val) const;
-
-private:
+    // private:
```
averinevg: I think it's worth adding a comment or TODO here

src/include/miopen/problem_description_base.hpp
```diff
@@ -0,0 +1,35 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#pragma once
```
averinevg, line 27: [Notice] Probably, we should use only either `#include guard` or `#pragma once`. Now we use both, but the first more often.

src/include/miopen/solver.hpp
```diff
@@ -38,6 +38,7 @@
 #include <miopen/miopen.h>
 #include <miopen/buffer_info.hpp>
 #include <miopen/performance_config.hpp>
+#include <miopen/batchnorm/problem_description.hpp>
```
averinevg: solver.hpp header contains base class for solvers and convolution solvers. I think this line should be removed.

src/include/miopen/solver.hpp
```diff
@@ -364,7 +365,7 @@ struct PerformanceConfigConvAsm1x1U : PerfConfigBase<PerformanceConfigConvAsm1x1
     bool operator==(const PerformanceConfigConvAsm1x1U& other) const;
 };
 
-struct ConvAsm1x1U final : ConvTunableSolver<PerformanceConfigConvAsm1x1U>
+struct ConvAsm1x1U : ConvTunableSolver<PerformanceConfigConvAsm1x1U>
```
averinevg: Why did you remove `final` here?

src/include/miopen/solver.hpp
```diff
@@ -420,6 +421,7 @@ struct ConvAsm1x1U final : ConvTunableSolver<PerformanceConfigConvAsm1x1U>
                                         const AnyInvokeParams& invoke_ctx) const;
 };
 
+#if 0
```
averinevg: If this code is not used, why should we keep it? I think it's worth adding a comment or TODO here

src/include/miopen/solver_base.hpp
```diff
@@ -0,0 +1,74 @@
+#pragma once
```
averinevg: I suggest not to put this in a separate file for now, but to do it later, when the implementation of the new interface for solvers is completed.

src/ocl/fusionopbiasbnactivocl.cpp
```diff
@@ -340,6 +166,7 @@ void BatchNormBwdTrainFusionOpDescriptor::calcBNParams(Handle& handle,
 miopenStatus_t BatchNormBwdTrainFusionOpDescriptor::GetNetworkConfig(std::string& network_config,
                                                                      Handle& handle)
 {
+
```
averinevg: Redundant blank line

averinevg (changes requested):
@JehandadKhan It is difficult to conduct quality review for such large PRs as it's hard to track links between files with lots of changes. Let's limit the number of changed lines to 1000 for non-trivial PRs.

src/include/miopen/fusion/solvers.hpp
atamazov: Yes, just like we do with problem descriptions of other primitives.

src/include/miopen/problem_description_base.hpp
atamazov: @averinevg This is out of the scope of this PR, I recommend closing this comment or marking it with explicit "[Notice]"

src/fusion.cpp
```diff
@@ -261,115 +227,11 @@ miopenStatus_t ConvForwardOpDescriptor::SetArgs(OperatorArgs& args,
                                                 const void* /*beta*/,
                                                 ConstData_t w)
 {
-    auto w_any = OpKernelArg(w);
-    args.ins_arg("weights" + std::to_string(GetIdx()), w_any);
-
+    const auto op_args = std::make_shared<fusion::ConvolutionOpInvokeParam>(w);
```
DrizztDoUrden: It's possible that I miss something, but should this guys be shared? I mean, are they really copied? Shouldn't they be unique_ptr's? Not that it would matter performance-wise, but ref counting would get eliminated. And you can still pass them around as raw ptrs anyway.

src/fusion.cpp
```diff
@@ -810,562 +378,79 @@ miopenStatus_t BiasFusionOpDescriptor::SetArgs(OperatorArgs& args,
...
-    std::map<std::pair<size_t, size_t>, std::vector<std::string>> size_map;
-    // A map between argument pointers (buffers) and argument names
-    std::map<size_t, std::vector<std::string>> ptr_map;
-
-    for(size_t idx = 0; idx < op_map.size(); idx++)
-    {
-        auto op   = op_map.at(idx);
-        auto keys = op->GetArgs();
-        for(auto&& key_arg : keys)
+    const auto solvers    = solver::SolverContainer<solver::fusion::ConvBiasActivAsm1x1U,
```
DrizztDoUrden: You should move the SolverContainer definition to a separate function and just call it here. You can mark it as inline if you want to although it would probably get inlined anyway.

src/fusion.cpp
```diff
@@ -810,562 +378,79 @@ miopenStatus_t BiasFusionOpDescriptor::SetArgs(OperatorArgs& args,
...
-                running_sz += arg_keys[idx].size;
-            }
-            arg_keys = std::move(padded_args);
-        }
-
-        if(arg_keys.empty())
-        {
-            MIOPEN_THROW("Kernel arguments not setup properly");
+            const auto strAlgo =
+                miopen::solver::Id{sol.solver_id}.GetAlgo(miopen::conv::Direction::Forward);
```
DrizztDoUrden: Save solver::Id{...} to a local.

src/fusion.cpp
```diff
@@ -810,562 +378,79 @@ miopenStatus_t BiasFusionOpDescriptor::SetArgs(OperatorArgs& args,
...
+            const auto algo = miopen::StringToConvolutionFwdAlgo(strAlgo);
+            MIOPEN_LOG_I2(algo);
+            if(algo == *conv_fwd_algo)
+                sols.push_back(sol);
         }
     }
+    else
+        sols = tmp_sols;
+    auto net_config =
+        input_desc.ToString() + ((input_desc.GetType() == miopenHalf) ? "FP16" : "FP32");
```
DrizztDoUrden: Generating a network config should be done in a separate method and via a string stream rather than concatenation. Also it should probably happen after the "UsupportedOp" check, but I may be wrong here.

src/fusion.cpp
```diff
@@ -810,562 +378,79 @@ miopenStatus_t BiasFusionOpDescriptor::SetArgs(OperatorArgs& args,
...
-            case OutputTensorDesc:
-                auto tensor_arg = GetTensorAttr(arg.key);
-                arg_keys.emplace_back(arg.key, Default, tensor_arg.size(), tensor_arg);
-                break;
-            }
+            if(!sol.invoker_factory)
+                MIOPEN_THROW(miopenStatusInternalError,
+                             "Invoker missing from solver " + sol.solver_id);
+            const auto invoker = handle.PrepareInvoker(
+                *sol.invoker_factory, sol.construction_params); // force compilation
```
DrizztDoUrden: Note: this doesn't force a compilation, kernels may be read from the binary cache or from HIP/OCL program handle cache in KCache object in the handle. This only forces the creation an invoker. Shouldn't you check if one already exists with your netcfg?

src/fusion.cpp
```diff
@@ -1388,51 +468,21 @@ miopenStatus_t FusionPlanDescriptor::Execute(const Handle& handle,
...
-        }
-    }
-    if(args.empty())
-    {
-        MIOPEN_THROW("Operator args not populated properly");
-    }
-    kernel(args);
+    const auto invoker = handle.GetInvoker(network_config, solver::Id{solution.solver_id}, {});
+    const auto& tmp    = op_args.params[1];
+    const auto& tmp2   = std::dynamic_pointer_cast<fusion::ActivationOpInvokeParam>(tmp);
```
DrizztDoUrden: Do you need the cast here or you can just pass AnyInvokeParams around? They were designed to be passed around like that and only have specific type on creation and inside of the invoker.
Plus you can do `dynamic_cast<...>(*tmp)`. That would eliminate ref count updates.

src/fusion.cpp
```diff
@@ -1388,51 +468,21 @@ miopenStatus_t FusionPlanDescriptor::Execute(const Handle& handle,
...
+    const auto tmp3    = op_args.params[1];
```
DrizztDoUrden: Are all this values not used at all? Is there any reason to create one set of "refs" (tmp2 is not really a ref as it owns the data it gets initialized from) and one of copies?

src/include/miopen/conv/problem_description.hpp
```diff
@@ -142,9 +143,10 @@ constexpr TElement GetW5(int spatial_dims, const std::vector<TElement>& data)
 
 namespace conv {
 
-struct ProblemDescription
+struct ProblemDescription : public ProblemDescriptionBase
```
DrizztDoUrden: Please, use public everywhere either explicitly or implicitly.

src/include/miopen/fusion.hpp
```diff
@@ -52,13 +58,15 @@ enum FusionKernelSourceType
...
+    std::vector<std::shared_ptr<fusion::FusionOpInvokeParamBase>> params;
     friend std::ostream& operator<<(std::ostream& stream, const OperatorArgs& x);
-    std::vector<OpKernelArg> args_vec;
-    std::unordered_map<std::string, OpKernelArg> args_map;
+    void SetArg(size_t idx, const std::shared_ptr<fusion::FusionOpInvokeParamBase>& arg)
+    {
+        if(params.size() < (idx + 1))
+            params.resize(idx + 1);
+        params[idx] = arg;
+    }
```
DrizztDoUrden:
```suggestion
    void SetArg(size_t idx, std::shared_ptr<fusion::FusionOpInvokeParamBase> arg)
    {
        if(params.size() < (idx + 1))
            params.resize(idx + 1);
        params[idx] = std::move(arg);
    }
```
This allows both move and copy. Also one may want to enforce move here with &&. And I'd rather use unique_ptr here anyway.

src/include/miopen/fusion/fusion_invoke_params.hpp
```diff
@@ -0,0 +1,187 @@
...
+struct FusionInvokeParams : InvokeParams
+{
+    FusionInvokeParams(){};
+    FusionInvokeParams(std::vector<std::shared_ptr<FusionOpInvokeParamBase>> op_invokers_,
+                       TensorDescriptor in_desc,
+                       ConstData_t in_,
+                       TensorDescriptor out_desc,
+                       Data_t out_,
+                       bool gfx90aFp16alt_)
+        // : op_invokers(std::move(op_invokers_)),
```
DrizztDoUrden: This should be move. You copy it two times in a row.

src/include/miopen/fusion/fusion_invoke_params.hpp
```diff
@@ -0,0 +1,187 @@
...
+          inDesc(in_desc),
+          in(in_),
+          outDesc(out_desc),
+          out(out_),
+          gfx90aFp16alt(gfx90aFp16alt_)
+    {
+    }
+
+    FusionInvokeParams(InvokeType type_,
+                       std::vector<std::shared_ptr<FusionOpInvokeParamBase>> op_invokers_,
```
DrizztDoUrden: This should probably be unique_ptr's

src/include/miopen/fusion/solvers.hpp
DrizztDoUrden: I would also suggest changing it from legacy style. For old style API you can just make std::tuple<ExecutionContext, FusionDescription> and use it as a legacy context type.

src/include/miopen/fusion/solvers.hpp
```diff
@@ -0,0 +1,356 @@
...
+        }
+        MIOPEN_LOG_I2(net_config);
+    }
+
+    ConvolutionContext GetConvContext(size_t idx, conv::Direction dir) const
+    {
+        const auto conv_prob = problem.GetConvProblem(idx, dir);
+        if(dir == conv::Direction::Forward)
+        {
+            auto ctx                       = ConvolutionContext{conv_prob};
```
DrizztDoUrden: Field-by-field copy is not a good approach. This may lead to very hard to find errors in the future, e.g. if some crucial field would be added to the original type.

src/include/miopen/fusion/utils.hpp
```diff
@@ -0,0 +1,126 @@
...
+                    miopenFusionOp_t op)
+{
+    int idx = 0;
+    for(const auto& ptr : op_map)
+    {
+        if(ptr->kind() == op)
+            return idx;
+        ++idx;
+    }
+    return -1;
```
DrizztDoUrden:
```suggestion
    auto it = std::find_if(op_map.cbegin(), op_map.cend(), [op](auto&& item){ return item->kind() == op; });
    return it == op_map.cend() ? -1 : std::distance(op_map.cbegin(), it);
```
Or if you crave for cpu ticks, you should use a regular for loop instead, like `for (auto idx = 0; idx < op_map.size(); ++idx)`. Maybe even caching op_map.size(). But std::find_if should not differ by a measurable amount inside the library.

src/include/miopen/fusion/utils.hpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+    const auto activ_idx = [&]() {
+        int idx = 0;
+        for(const auto& prim : desc.op_map)
+        {
+            if(prim->kind() == miopenFusionOpActivForward)
+                return idx;
+            ++idx;
+        }
+        return -1;
```
DrizztDoUrden: std::find_if and std::distance

src/include/miopen/problem_description_base.hpp
DrizztDoUrden: Still as it's been talked around a lot already, we don't want any new files to have include guards, #pragma once should be used instead.

src/solver.cpp
```diff
@@ -192,8 +194,11 @@ miopenConvAlgorithm_t Id::GetAlgo() const
     return it->second.convAlgo;
 }
 
-inline bool
-Register(IdRegistryData& registry, uint64_t value, Primitive primitive, const std::string& str)
+inline bool Register(IdRegistryData& registry,
+                     uint64_t value,
+                     Primitive primitive,
+                     const std::string& str,
+                     miopenConvAlgorithm_t algo = miopenConvolutionAlgoDirect)
```
DrizztDoUrden: This is also used by solvers, which are not convolution. It should not set convAlgo for them.

src/solver/batchnorm/backward_per_activation.cpp
```diff
@@ -32,11 +32,208 @@
...
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_BN_BWDTRG_ACTIV_FUSED)
 
 namespace miopen {
 
 namespace solver {
 
+namespace fusion {
+
+bool BnBwdTrgActivationFused::IsApplicable(const FusionContext& context) const
```
DrizztDoUrden: I would suggest moving fused solver to a separate file.

src/solver/batchnorm/forward_inference.cpp
```diff
@@ -32,11 +32,146 @@
...
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_BN_FWDINFER_ACTIV_FUSED)
 
 namespace miopen {
 
 namespace solver {
 
+namespace fusion {
+
+bool BnFwdInferActivationFused::IsApplicable(const FusionContext& context) const
```
DrizztDoUrden: I would suggest moving fused solver to a separate file.

src/solver/batchnorm/forward_per_activation.cpp
```diff
@@ -32,10 +32,216 @@
...
+#include <miopen/fusion/solvers.hpp>
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_BN_FWDTRG_ACTIV_FUSED)
 
 namespace miopen {
 
 namespace solver {
+namespace fusion {
+
+bool BnFwdTrgActivationFused::IsApplicable(const FusionContext& context) const
```
DrizztDoUrden: I would suggest moving fused solver to a separate file.

src/solver/conv_bin_winoRxS.cpp
```diff
@@ -497,25 +502,64 @@ ConvSolution ConvBinWinogradRxS::GetSolution(const ExecutionContext& ctx,
 
     return result;
 }
+namespace fusion {
 
-bool ConvBinWinogradRxSFused::IsApplicable(const ExecutionContext& ctx,
-                                           const ProblemDescription& problem) const
+bool ConvBinWinogradRxSFused::IsApplicable(const FusionContext& params) const
```
DrizztDoUrden: I would suggest moving fused solver to a separate file.

src/solver/conv_ocl_dir2Dfwd.cpp
```diff
@@ -484,12 +489,191 @@ ConvSolution ConvOclDirectFwd::GetSolution(const ConvolutionContext& ctx,
     return result;
 }
 
-ConvSolution ConvOclDirectFwdFused::GetSolution(const ConvolutionContext& ctx,
-                                                const ProblemDescription& problem,
-                                                const LegacyPerformanceConfig& config) const
+namespace fusion {
```
DrizztDoUrden: I would suggest moving fused solver to a separate file.

src/solver/conv_ocl_dir2Dfwd.cpp
```diff
@@ -484,12 +489,191 @@ ConvSolution ConvOclDirectFwd::GetSolution(const ConvolutionContext& ctx,
...
+    kernel_info.kernel_file = "MIOpenConvDirBatchNormActiv.cl";
+    kernel_info.kernel_name = "MIOpenConvUniBatchNormActiv";
+    const auto& desc        = *context.problem.fusion_plan_desc;
+
+    const int bias_idx  = GetOpIdx(desc.op_map, miopenFusionOpBiasForward);
+    const int activ_idx = GetOpIdx(desc.op_map, miopenFusionOpActivForward);
+    const int bn_idx    = GetOpIdx(desc.op_map, miopenFusionOpBatchNormInference);
+
+    if(bias_idx != -1)
+        kernel_info.comp_options += " -DMLO_CONV_BIAS=" + std::to_string(1);
```
DrizztDoUrden: Please use `KernelBuildParams` in all new code.

src/solver/conv_winoRxS.cpp
```diff
@@ -948,19 +950,43 @@ ConvSolution ConvBinWinogradRxSf2x3g1::GetSolution(const ConvolutionContext& ctx
     return tunable.GetSolution(ctx, problem, tunable.GetDefaultPerformanceConfig(ctx, problem));
 }
 
-bool ConvBinWinogradRxSf2x3g1Fused::IsApplicable(const ConvolutionContext&) const
+namespace fusion {
```
DrizztDoUrden: I would suggest moving fused solver to a separate file.

DrizztDoUrden (changes requested):

src/include/miopen/fusion.hpp
JehandadKhan: This macro is only used in `fusion.cpp` to construct fusion plans to support dedicated CBA API. I have moved it back to the CPP file since its not used anywhere else.

src/include/miopen/fusion/solvers.hpp
JehandadKhan: @DrizztDoUrden I am not sure I understand the requirement here. If you would like to remove `FusionDescription` as a member of the `FusionContext` then that would cascade into changes which are beyond the scope of this PR. For example, I would need to rewrite the `FindSolution` function in `find_solution.hpp` which expects the context to have a problem member. Moreover, this design follows the methodology of the `ConvolutionContext`.

Please elaborate as to the target design that you would like to achieve here.

src/include/miopen/fusion/solvers.hpp
JehandadKhan: I dont understand the rationale for that ? I dont believe a blank line here is a standard followed in MIOpen.

averinevg: This is just a recommendation for a better visual experience.

.githooks/pre-commit
```diff
@@ -4,7 +4,7 @@
 # are installed, and if so, uses the installed version to format
 # the staged changes.
 
-base=clang-format-10
+base=clang-format-12
```
JehandadKhan, line 7: [Note]: This is leftover from previous commit and may be moved to a separate PR if needed.

src/fusion.cpp
JehandadKhan: I am not sure I fully understand. The call to `PrepareInvoker` includes a call to `AddKernel` which can force the creation of a `Program` object. Therefore, I was under the impression that the `AddKernel` logic you are referring to is not necessary. Can you please elaborate what am I missing here ?

src/include/miopen/problem_description_base.hpp
JehandadKhan: I believe we can mark this comment resolved ?

src/include/miopen/fusion/solvers.hpp
JehandadKhan: Done!

src/include/miopen/fusion/solvers.hpp
JehandadKhan: `FusionContext` moved to its own file as requested.

src/include/miopen/fusion_plan.hpp
JehandadKhan: The commented code has been removed!

src/include/miopen/solver.hpp
JehandadKhan: has been removed.

src/include/miopen/solver.hpp
JehandadKhan: Added back

src/include/miopen/solver.hpp
JehandadKhan: Has been removed.

src/include/miopen/solver_base.hpp
JehandadKhan: The file was redundant and has been removed.

src/ocl/fusionopbiasbnactivocl.cpp
JehandadKhan: has been removed.

src/fusion.cpp
JehandadKhan: I agree, this has been converted to unique_ptrs throughout with a few changes to accomodate the update.

src/fusion.cpp
JehandadKhan: Done

src/fusion.cpp
JehandadKhan: Done

src/fusion.cpp
JehandadKhan: Agreed on both, I have updated the code.

src/fusion.cpp
JehandadKhan: due to the unique_ptr update, this has been fixed.

src/fusion.cpp
JehandadKhan: This was redundant code and has been removed

src/include/miopen/conv/problem_description.hpp
JehandadKhan: removed `public`

src/include/miopen/fusion.hpp
JehandadKhan: Changed to unique_ptr

src/include/miopen/fusion/fusion_invoke_params.hpp
JehandadKhan: removed since incompatible with `unique_ptr`

src/include/miopen/fusion/fusion_invoke_params.hpp
JehandadKhan: Done

src/include/miopen/fusion/solvers.hpp
JehandadKhan: Fixed.

src/include/miopen/fusion/utils.hpp
JehandadKhan: Fixed.

src/include/miopen/fusion/utils.hpp
JehandadKhan: Fixed

src/solver.cpp
JehandadKhan: Fixed.

src/solver/batchnorm/backward_per_activation.cpp
JehandadKhan: Done

src/solver/batchnorm/forward_inference.cpp
JehandadKhan: Done

src/solver/batchnorm/forward_per_activation.cpp
JehandadKhan: Done

src/solver/conv_bin_winoRxS.cpp
JehandadKhan: Done

src/solver/conv_ocl_dir2Dfwd.cpp
JehandadKhan: Done

src/solver/conv_ocl_dir2Dfwd.cpp
JehandadKhan: Done

src/solver/conv_winoRxS.cpp
JehandadKhan: Done

src/include/miopen/fusion/solvers.hpp
JehandadKhan: > I would also suggest changing it from legacy style. For old style API you can just make std::tuple<ExecutionContext, FusionDescription> and use it as a legacy context type.

@DrizztDoUrden  The issue with trying to do that is then I cant use `FindSolution` in the `find_solution.hpp` which requires `problem` to be a member of the Context. Since `ConvolutionContext` follows the same design as `FusionContext` I am not sure why I would want to deviate unless we would like to change both, which is best addressed in a separate PR.

Please elaborate if I have missed the point.

JehandadKhan:
@DrizztDoUrden , @averinevg  Thanks for your review. I have addressed all the comments, please take a look so we can push this PR through. 

cc @junliume

src/include/miopen/fusion/solvers.hpp
DrizztDoUrden: @JehandadKhan What I suggest is the following (like convolution and e.g. activation does it): solvers have methods with "old-style" signatures for temporary compatibility with old `find_solution.hpp` functions which take some type like `std::tuple<ExecutionContext, FusionDescription>` and just map it to methods with correct signatures. Actual implementations have parameters in form of `const ExecutionContext&, const FusionDescription&`. In fact, there are functions which require no mapping at all like this one https://github.com/ROCmSoftwarePlatform/MIOpen/blob/bc74d11db5187873204b623eb2d541d40a501745/src/include/miopen/find_solution.hpp#L227 an you don't need to map anything if you don't use perf db.

src/fusion.cpp
DrizztDoUrden: `AddKernel` does not necessarily compile anything: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/bc74d11db5187873204b623eb2d541d40a501745/src/kernel_cache.cpp#L130
So this may call compilation, but it may as well not and does not "force" it.
And by checking if anything is pre-existing I was talking of adding something like this:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/bc74d11db5187873204b623eb2d541d40a501745/src/include/miopen/find_solution.hpp#L350
at the start of the method to prevent duplicate find and re-creation of invokers if compile has been called twice unless that is intended and wished behaviour.

src/solver/conv_bin_winoRxS_fused.cpp
```diff
@@ -0,0 +1,240 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <miopen/solver.hpp>
```
averinevg: @JehandadKhan This `#include` directive can be removed

src/solver/conv_bin_winoRxS.cpp
```diff
@@ -32,6 +32,9 @@
 #include <miopen/kernel_build_params.hpp>
 #include <miopen/conv/data_invoke_params.hpp>
 #include <miopen/conv/tensors.hpp>
+#include <miopen/fusion_plan.hpp>
+#include <miopen/fusion/solvers.hpp>
+#include <miopen/fusion/utils.hpp>
```
averinevg: Redundant ~~headers~~ `#include` directives

src/solver/conv_bin_winoRxS.cpp
```diff
@@ -42,6 +45,8 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_WRW)
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_FWD_BWD)
 /// \todo Detect at runtime and remove this var:
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_SRAM_EDC_DISABLED)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_FUSED_WINOGRAD)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_GCN_ASM_KERNELS)
```
averinevg: Redundant env variables

src/solver/conv_asm_1x1u_bias_activ.cpp
```diff
@@ -35,170 +35,243 @@
 #include <miopen/generic_search.hpp>
```
averinevg: @JehandadKhan I suggest to rename source file to `conv_asm_1x1u_bias_activ_fused.cpp` using `git mv`

src/solver/conv_ocl_dir2Dfwd_fused.cpp
```diff
@@ -0,0 +1,233 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <miopen/handle.hpp>
+#include <miopen/legacy_exhaustive_search.hpp>
+#include <miopen/solver.hpp>
```
averinevg: @JehandadKhan This `#include` directive can be removed

src/solver/conv_ocl_dir2Dfwd.cpp
```diff
@@ -267,9 +267,9 @@ bool ConvOclDirectFwd::IsValidPerformanceConfig(const ProblemDescription& proble
     return true;
 }
 
-static ConvSolution BaseGetSolution(const ConvolutionContext& ctx,
-                                    const ProblemDescription& problem,
-                                    const LegacyPerformanceConfig& config)
+ConvSolution BaseGetSolution(const ConvolutionContext& ctx,
+                             const ProblemDescription& problem,
+                             const LegacyPerformanceConfig& config)
```
averinevg: @JehandadKhan This function related to `ConvOclDirectFwd` solver, I suggest to rename it or made a static member of the class.

src/include/miopen/fusion/solvers.hpp
```diff
@@ -0,0 +1,206 @@
...
+    bool SetNextValue(const FusionContext& problem);
+    bool IsValid(const FusionContext& problem) const;
+};
+
+struct ConvBiasActivAsm1x1U : FusionTunableSolver<PerformanceConfigConvBiasActivAsm1x1U>
+{
+    const std::string& SolverDbId() const override { return GetSolverDbId<ConvBiasActivAsm1x1U>(); }
+
+    bool IsApplicable(const FusionContext& problem) const override;
+    ConvSolution
```
averinevg: @JehandadKhan Probably `GetWorkspaceSize()` and `MayNeedWorkspace()` are missing

averinevg (changes requested):

src/fusion.cpp
```diff
@@ -810,562 +380,91 @@ miopenStatus_t BiasFusionOpDescriptor::SetArgs(OperatorArgs& args,
...
-            }
-            arg_keys = std::move(padded_args);
-        }
-
-        if(arg_keys.empty())
-        {
-            MIOPEN_THROW("Kernel arguments not setup properly");
+            const auto id      = miopen::solver::Id{sol.solver_id};
+            const auto strAlgo = id.GetAlgo(miopen::conv::Direction::Forward);
+            MIOPEN_LOG_I2(miopen::solver::Id{sol.solver_id}.ToString());
```
DrizztDoUrden: Duplicate constructor

src/include/miopen/fusion/context.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+struct FusionContext : miopen::ExecutionContext
+{
+    FusionDescription problem;
+    FusionContext(FusionPlanDescriptor* ptr_desc, Handle& handle)
+        : ExecutionContext(&handle), problem(ptr_desc)
+    {
+    }
+    void GetNetworkConfig(std::string& net_config) const
+    {
```
DrizztDoUrden: I would suggest using a string stream here.

DrizztDoUrden (changes requested):

src/fusion.cpp
atamazov: @DrizztDoUrden @JehandadKhan 

👍 Yeah, we need to check if Invoker already exists before creating it.

Some clarifications. This is true:
> The call to PrepareInvoker includes a call to AddKernel... 

This is not:
> which can force the creation of a Program object

Creation of a Program object happens only if the same Program object does not exist. If it exists, it's simply reused.

Otherwise the library creates it by calling `handle.LoadProgram(...)`. The latter method may reuse a code object from the binary cache. 

That is, compilation happens only if:
- Program object does not exist (not registered in the program_map)
- Code object does not reside in the binary cache.

I recommend removing the `// force compilation` comment.

src/solver/conv_bin_winoRxS.cpp
atamazov: @averinevg I guess you mean not _headers_ but `#include` directives? 😉

src/solver/conv_asm_1x1u_bias_activ.cpp
atamazov: @averinevg This PR does not add this file, only modifies it. Therefore your recommendation is good but out of the scope. Let's not add extra work to this big PR ;)

src/solver/conv_ocl_dir2Dfwd.cpp
atamazov: Yeah, BaseGetSolution is too generic. Let's make it a static member of one of these classes (`ConvOclDirectFwd` or `ConvOclDirectFwdFused`). Or prepend its name with `ConvOclDirectFwd`.

src/solver/conv_bin_winoRxS.cpp
averinevg: @atamazov Yes, you are right

src/solver/conv_asm_1x1u_bias_activ.cpp
averinevg: @atamazov This PR also splits fused and non-fused solvers with adding some new files with `_fused` ending. Renaming one file is not much work compared to the rest.

src/include/miopen/fusion/solvers.hpp
atamazov: @averinevg If WS is not needed, then the solver can safely derive them from `SolverMixin<FusionContext>`, right?

src/solver/conv_asm_1x1u_bias_activ.cpp
atamazov: @averinevg Ok, let's hope that github will not show this as one removed file plus one added file (which happens from time to time when you rename and edit a file in the same PR)))

averinevg: @atamazov This is what `git mv` is for :)

src/include/miopen/fusion/solvers.hpp
averinevg: @atamazov Yes, but this solver comes from `src/include/miopen/solver.hpp`, and these functions are there

JehandadKhan: The Fusion solvers are derived from `SolverMixin<FusionContext>` and inherit both of these methods since it derives from `SolverBase`

src/solver/conv_bin_winoRxS_fused.cpp
JehandadKhan: Done

src/solver/conv_bin_winoRxS.cpp
JehandadKhan: Removed

src/solver/conv_bin_winoRxS.cpp
JehandadKhan: Removed

src/solver/conv_asm_1x1u_bias_activ.cpp
JehandadKhan: Renamed

src/solver/conv_ocl_dir2Dfwd_fused.cpp
JehandadKhan: Removed

src/solver/conv_ocl_dir2Dfwd.cpp
JehandadKhan: Converted to a static method

src/fusion.cpp
JehandadKhan: Fixed

src/include/miopen/fusion/context.hpp
JehandadKhan: Updated the `GetNetworkConfig` to use `std::stringstream` everywhere

src/include/miopen/fusion/solvers.hpp
averinevg: @JehandadKhan `ConvBiasActivAsm1x1U` uses `GetSolution` from `ConvAsm1x1U`, and `ConvAsm1x1U` has these methods

src/solver/conv_asm_1x1u_bias_activ.cpp
averinevg: @atamazov You were right, github shows this as renaming in a single commit and as 2 files for the whole PR

src/include/miopen/fusion/solvers.hpp
JehandadKhan: I get your point, let me rework it.

src/fusion.cpp
JehandadKhan: Removed

src/include/miopen/fusion/solvers.hpp
JehandadKhan: @DrizztDoUrden As discussed, this change would not go with the Find Machinery implemented in `find_solution.hpp` therefore I have implemented in the same style as the tunable convolution solvers. Once the Find machinery is generalized, we can remove the shim between the old style fusion context and "two-parameter" methods. To minimize the changes, I have implemented the two paramter methods in the convolution based fusion solvers.

JehandadKhan:
@DrizztDoUrden @averinevg @atamazov Please let me know if there are any other issues.

src/include/miopen/any_solver.hpp
```diff
@@ -27,6 +27,7 @@
 #ifndef MIOPEN_GUARD_MLOPEN_ANY_SOLVER_HPP
 #define MIOPEN_GUARD_MLOPEN_ANY_SOLVER_HPP
 
+#include <miopen/problem_description_base.hpp>
```
DrizztDoUrden, line 30: Not critical, but is this used in the header?

DrizztDoUrden (approved):
LGTM (my comments resolved)

averinevg (approved):

junliume (approved):

PR #1876, opened 2022-11-11
[Find-2.0] Serialize kernel perf configs

umangyadav:
Serialize optimal perf configs into `miopenSolution`. It helps  running kernel at better performance when userDB entry is not  present or FIND_ENFORCE=3 is not  set.


src/include/miopen/find_solution.hpp
```diff
@@ -94,6 +94,15 @@ auto FindSolutionImpl(
                               << s.AltSolverDbId() << ": " << config
                               << ". Performance may degrade.");
             }
+            else if(not perf_cfg.empty()) {
```
DrizztDoUrden: Please don't use alternative operators. That's not how 99% of code is written here, they are not required and most people have no idea what it means and how it works.

src/problem.cpp
```diff
@@ -341,6 +343,11 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
     })();
 
     const auto netcfg = actual.AsConvolution().BuildConfKey();
+    auto conv_ctx = ConvolutionContext{actual.AsConvolution(), {&handle}};
```
DrizztDoUrden:
```suggestion
    const auto conv_problem = actual.AsConvolution();
    const auto netcfg = conv_problem.BuildConfKey();
    const auto conv_ctx = ConvolutionContext{{&handle}, conv_problem};
```

src/include/miopen/find_solution.hpp
```diff
@@ -46,7 +46,7 @@ namespace solver {
 
 template <class Solver, class Context, class Db>
 auto FindSolutionImpl(
-    rank<1>, Solver s, const Context& context, Db& db, const AnyInvokeParams& invoke_ctx)
+    rank<1>, Solver s, const Context& context, Db& db, const AnyInvokeParams& invoke_ctx, const std::string& perf_cfg)
```
DrizztDoUrden: I'd rather wrap it in `std::optional`. That wouldn't require a guarantee that empty string is not a valid value.

src/include/miopen/any_solver.hpp
```diff
@@ -250,6 +262,13 @@ struct AnySolver
             return miopen::solver::FindSolution(value, ctx, db, invoke_ctx);
         };
 
+        ConvSolution FindSolutionCfg(const ConvolutionContext& ctx,
```
DrizztDoUrden: This name is missleading.

DrizztDoUrden (dismissed):

umangyadav:
@DrizztDoUrden  Can you take over this PR ? I raised this so that it can show what I am trying to achieve and would  provide a little help.

atamazov:
@DrizztDoUrden Update to our discussion. This feature is actually necessary, even if kernel binaries are included into miopenSolution (which means that MIOpen does not even need to build kernels).

Why: Generally, GetSolution (which builds InvokerFactory) needs to have both problem config and tuning config in order to work properly. An InvokerFactory is free to use values from TuningConfigs. For example, `ConvBinWinoRxS` passes value from it to the kernel as a parameter (n_groups).

Please set https://github.com/ROCmSoftwarePlatform/MIOpen/labels/enhancement https://github.com/ROCmSoftwarePlatform/MIOpen/labels/performance https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_blocker as this is expected to become a blocker for Find 2.0 performance.

atamazov:
@umangyadav 
> @DrizztDoUrden Can you take over this PR ? I raised this so that it can show what I am trying to achieve and would provide a little help.

Please assign Vasilii to this PR 
image

DrizztDoUrden:
> Can you take over this PR ? I raised this so that it can show what I am trying to achieve and would provide a little help.

Sure, will do. We have a holyday today in Serbia, but I can take it over on Monday if you wish.

umangyadav:
> Sure, will do. We have a holyday today in Serbia, but I can take it over on Monday if you wish.

Sure, Monday is good. Enjoy the  weekend.

atamazov:
@DrizztDoUrden Please let me know do you agree with https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1876#issuecomment-1311793341

DrizztDoUrden:
@atamazov Yes, I agree.

junliume:
@umangyadav @DrizztDoUrden could you keep track of this PR? Pushing it through CI is the first step.
CC: @JehandadKhan 

@umangyadav if we need it soon, please add urgency and importance label.

DrizztDoUrden:
Fixed a pre-existing fin build error, now this passes. @junliume @umangyadav

src/include/miopen/any_solver.hpp
```diff
@@ -243,51 +246,59 @@ struct AnySolver
...
-        std::string GetPerfCfgParams(const ConvolutionContext& ctx, Db& db, std::true_type) const
+        std::optional<std::string>
+        GetPerfCfgParams(const ConvolutionContext& ctx, PerformanceDb& db, std::true_type) const
         {
             using PerformanceConfig = decltype(value.GetDefaultPerformanceConfig(ctx));
             PerformanceConfig config{};
             if(db.Load(ctx.problem, value.SolverDbId(), config))
             {
-                MIOPEN_LOG_I2("Perf Db: Record Loaded: " << value.SolverDbId());
+                MIOPEN_LOG_I2("Perf PerformanceDb: Record Loaded: " << value.SolverDbId());
```
atamazov: "Perf PerformanceDb" -- what dos it mean?

fin/src/include/conv_fin.hpp
```diff
@@ -691,7 +694,8 @@ int ConvFin<Tgpu, Tref>::MIOpenPerfEval()
                     std::cerr << solver_name << " Finished Search WRW" << std::endl;
                     kern_objs = BuildJsonKernelList(h, solution.construction_params);
                     SolutionHasProgram(h, solution);
-                    params = s.GetPerfCfgParams(ctx, db);
+                    // todo: proper handling of empty values
+                    params = s.GetPerfCfgParams(ctx, db).value_or("");
```
atamazov: @cderb I do not know how to properly handle empty `params` here. But what is clear is that your approval for fin changes is required.

atamazov (commented):
@cderb Please review fin changes.

fin/src/include/conv_fin.hpp
cderb: It is ok if params are set to empty string here.

Since Fin is a separate repository from MIOpen and will be setup as a submodule, it is preferable for these changes to be submitted to https://github.com/ROCmSoftwarePlatform/MIFin if they are necessary.

src/include/miopen/any_solver.hpp
cderb: Perhaps "Perf PerformanceDb" could be shortened to "PerformanceDb". Which seems equivalent to the previous string here, though I assume the change is to reflect the name of the new C class.

src/include/miopen/any_solver.hpp
```diff
@@ -243,51 +246,59 @@ struct AnySolver
...
+        std::optional<std::string> GetPerfCfgParams(const ConvolutionContext& ctx,
+                                                    const PerformanceDb& db,
+                                                    std::false_type) const
         {
-            MIOPEN_LOG_I2("Perf Db: No Config: " << value.SolverDbId());
+            MIOPEN_LOG_I2("Perf PerformanceDb: No Config: " << value.SolverDbId());
             std::ignore = ctx;
             std::ignore = db;
-            return "";
+            return std::nullopt;
```
cderb: This change necessitates the multiline changes to conv_fin.hpp. Is there a reason to return null here? I would suggest reverting this line so the Fin changes aren't necessary.

cderb (dismissed):

src/include/miopen/any_solver.hpp
DrizztDoUrden: If we use an empty string as no-value marker, we need to guarantee that there would be no solver which may have an empty string as a valid perf config value, but it is not default. This may lead to weird bugs which would be difficult to debug. I thought, I have outlined it somewhere before and it just got overlooked, but I can't find that.
But you are probably right. At the moment that would require a lot of changes to be meaningful. I should change it to `""` with no optional.

junliume (approved):

cderb (approved):

Issue #1843, opened 2022-10-24
Google tests to obey MIOPEN_TEST_HALF etc


atamazov:
(1) When the tests are run under CTest's control (i.e. in our CI):
- (1.1) The tests in gtest folder must obey the MIOPEN_TEST_HALF, MIOPEN_TEST_INT8, MIOPEN_TEST_BFLOAT16 etc CMAKE variables. For example FP16 tests should be run only if MIOPEN_TEST_HALF was set to On when the library has been configured.
- (1.2) Ditto `MIOPEN_TEST_ALL`. 
- (1.3) The refactored test should obey `SKIP_ALL_EXCEPT_TESTS` just like the original test does. Explanation: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1652#issuecomment-1306341535

(2) When the tests are run autonomously (in other words, "_in standalone mode_"),
- (2.1) All the datatypes needs to be tested
- (2.2) "Full Testing" should be performed

### How to implement detection of "standalone mode":

**_Query `MIOPEN_TEST_ALL` to detect if the test is run under CTest._** If the variable is unset, then assume the standalone mode. Otherwise, assume that the test is run under CTest's control.

Example of proper implementation of checks:
```C++
TEST_P(...)
{
    if(!(IsTestSupportedForDevice()                      //
         && (miopen::IsUnset(ENV(MIOPEN_TEST_ALL))       // standalone run
             || (miopen::IsEnabled(ENV(MIOPEN_TEST_ALL)) // or --half full tests enabled
                 && miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG)) == "--half"))))
    {
        GTEST_SKIP();
    }

    Run....
};
```

@junliume https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_blocker

JehandadKhan:
I do not agree with this design since it requires MIOpen run CMake again and again just to change the test cases which are run. Instead each of those precision can be a different test and be run directly. If manual running of tests requires filtering one over the other that can be done using the GTest filter feature which allows tests to run selectively. 

I would like to point out that due to this "feature" part of the test information ( precision and in some tests even test cases ) live in the CMake file while the code that is actually run lives in the CPP file which to my mind is unnecessary complexity. 

[...]

PR #1877, opened 2022-11-14
[Test][gtest] Sl/test solver

xinlipn:
ConvAsm3x3 solver test. Passed CI after disabling Navi21 which is not supported by this solver.


test/test.hpp
```diff
@@ -93,7 +93,13 @@ inline void expect_op(const TLeft& left,
 #define EXPECT_EQUAL(LEFT, RIGHT) EXPECT_OP(LEFT, ==, RIGHT)
 #define STATUS(...) EXPECT((__VA_ARGS__) == 0)
 
+// Conflicts with GoogleTest FAIL()
+#ifdef FAIL
+#undef FAIL
```
atamazov: How it could be so that test/test.hpp is used in gtest?

atamazov (changes requested):

atamazov:
This can't be merged in due to Navi21 issues. I recommend converting this into draft.

atamazov:
@junliume @JehandadKhan I highly recommend pausing with new GoogleTests until https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1843#issuecomment-1307952826 is resolved.

atamazov:
@xinlipn 
> ConvAsm3x3 solver test. 

Please describe which aspects of the solver's behavior this test is intended to check. Thanks.

by @xinlipn:
_@atamazov the intension is to bring test down to solver level for more granularity, i.e. each solver test will have its own gtest. Hope this makes sense._

test/test.hpp
xinlipn: @atamazov test.hpp isn't used in GTest, however, there's a warning from compiler about FAIL() being already defined in GoogleTest. This change probably shouldn't go to develop

xinlipn:
> This can't be merged in due to Navi21 issues. I recommend converting this into draft.

@atamazov ConvAsm3x3 doesn't support Navi21, only supports gfx8* and gfx9*.

atamazov:
@xinlipn That is why this test won't pass on Navi21. If this PR is merged in, CI will always fail. Which means that this PR, in its current state, shouldn't be merged.

atamazov:
@xinlipn 
> @atamazov the intension is to bring test down to solver level for more granularity, i.e. each solver test will have its own gtest. Hope this makes sense.

Yes, this makes sense. However, this does not answer my question.

pfultz2:
If we are writing tests for solvers then there should be unit tests for `IsApplicable`, `GetSolution`, and `IsValidPerformanceConfig` to get full branch coverage. For example, there is a lot of checking that is in `IsApplicable`, I would expect a unit test for each one of these branches. These unit tests should be very fast so it should easy to run through many configs to check all the branches.

test/gtest/solver_convasm3x3u.cpp
```diff
@@ -0,0 +1,90 @@
...
+    auto ctx = miopen::ConvolutionContext{
+        inputDesc, wDesc, outputDesc, convDesc, miopen::conv::Direction::Forward};
+
+    ctx.SetStream(&handle);
+    ctx.DetectRocm();
+
+    if(!solv.IsApplicable(ctx))
+    {
+        test_skipped = true;
+        GTEST_FAIL() << solv.SolverDbId() << "ConvAsm3x3U Not Applicable for this problem"
```
JehandadKhan:
```suggestion
        GTEST_SKIP() << solv.SolverDbId() << "ConvAsm3x3U Not Applicable for this problem"
```

test/test.hpp
JehandadKhan: Perhaps we can refactor these defs out into their own header, so that they are only included when they are needed

JehandadKhan:
> This can't be merged in due to Navi21 issues. I recommend converting this into draft.

That is a false issue, if the solver is not applicable on Navi, then the test should skip instead of failing, since there is nothing to test! Please see my [suggestion here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1877#discussion_r1024567216)

> @junliume @JehandadKhan I highly recommend pausing with new GoogleTests until https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1843#issuecomment-1307952826 is resolved.

I dont agree with that, since that would mean a new test which would enhance test coverage in MIOpen would be withering in our PR section. Since we wish to get to a place where we test each solver individually.

> If we are writing tests for solvers then there should be unit tests for IsApplicable, GetSolution, and IsValidPerformanceConfig to get full branch coverage. For example, there is a lot of checking that is in IsApplicable, I would expect a unit test for each one of these branches. These unit tests should be very fast so it should easy to run through many configs to check all the branches.

I agree, however, that is not the intent of this PR. The purpose of this PR to increase numerical accuracy coverage to a new solver, which previously does not exists. The current testing infra uses the API to test numerical accuracy on the API level where individual solvers might get skipped since they were not fast enough. Once we have numerical coverage over all the solvers, we can start the work on unit testing each solver as you propose.

cc @xinlipn

test/gtest/solver_convasm3x3u.cpp
xinlipn: @JehandadKhan replaced with GTEST_SKIP()

test/test.hpp
xinlipn: > Perhaps we can refactor these defs out into their own header, so that they are only included when they are needed

@JehandadKhan separated Macros defined in test.hpp to a separate header test_macro.hpp and updated affected code.

test/test_macro.hpp
```diff
@@ -0,0 +1,58 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#pragma once
+
+#include "test.hpp"
+
+#define CHECK(...)                                    \
```
atamazov: Why this is copied from test.hpp?

test/test_macro.hpp
```diff
@@ -0,0 +1,58 @@
...
+    do                                                \
+    {                                                 \
+        if(!(__VA_ARGS__))                            \
+            failed(#__VA_ARGS__, __FILE__, __LINE__); \
+                                                      \
+    } while(false)
+
+#define EXPECT(...)                                         \
```
atamazov: Ditto

test/test_macro.hpp
```diff
@@ -0,0 +1,58 @@
...
+    do                                                      \
+    {                                                       \
+        if(!(__VA_ARGS__))                                  \
+            failed_abort(#__VA_ARGS__, __FILE__, __LINE__); \
+                                                            \
+    } while(false)
+
+#define EXPECT_OP(LEFT, OP, RIGHT)                                 \
```
atamazov: Ditto

test/handle_test.cpp
```diff
@@ -39,7 +39,7 @@
 #include "get_handle.hpp"
 #include <vector>
 #include <thread>
-#include "test.hpp"
+#include "test_macro.hpp"
```
atamazov: Any way to resolve the "FAIL" problem without touching existing tests?

test/gtest/solver_convasm3x3u.cpp
atamazov: Good. This change resolves https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1877#issuecomment-1315461491.

test/test_macro.hpp
xinlipn: > Why this is copied from test.hpp?

@atamazov  this idea is to separate FAIL() out in a separate header which conflicts with the same definition in GoogleTest. It seems to be strange if just move FAIL() out, that's I put all Macros in the new header file

test/test_macro.hpp
xinlipn: Please see comments above

test/handle_test.cpp
xinlipn: @atamazov another way is only separate FAIL in to a new header file or remove this definition for now if no code is currently using it.

test/test_macro.hpp
xinlipn: Please see comments above

test/handle_test.cpp
JehandadKhan: The PR can be simplified, if we just remove the FAIL macro if its not being used anywhere

test/gtest/solver.hpp
```diff
@@ -0,0 +1,166 @@
...
+        EXPECT_FALSE(miopen::find_idx(ref_out, miopen::not_finite) >= 0)
+            << "Non finite number found in the CPU data";
+
+        EXPECT_TRUE(error < threshold)
+            << "Error beyond tolerance Error:" << error << ",  Threshold: " << threshold;
+
+        miopenDestroyConvolutionDescriptor(conv_desc);
+    }
+    ConvTestCase conv_config;
+    miopenConvolutionDescriptor_t conv_desc;
```
JehandadKhan: If we use the MIOpen C++ objects instead, then you would not need to manage the lifetime of these objects.

test/test_macro.hpp
atamazov: @xinlipn But then why test.hpp (that contains the same CHECK) is included here?

xinlipn: > @xinlipn But then why test.hpp (that contains the same CHECK) is included here?

@atamazov separate out only one Macro FAIL() from the rest and just create a header file for it seems to be a temp workaround and looks weird logically, at least to me. 

FAIL() isn't used anywhere and I removed it for now to minimize the impact to other code.
Consequently test_macro.hpp is removed and changes in other headers anyway are reverted.

test/gtest/solver.hpp
xinlipn: > If we use the MIOpen C++ objects instead, then you would not need to manage the lifetime of these objects.

@JehandadKhan , done

test/handle_test.cpp
xinlipn: > The PR can be simplified, if we just remove the FAIL macro if its not being used anywhere

@JehandadKhan done
cc @atamazov

JehandadKhan (approved):

atamazov (approved):
No objections!

junliume (approved):

junliume (approved):

PR #1885, opened 2022-11-21
Rework for any_solver perf params test function.

cderb:
Rename function TestSysDbRecord to TestPerfCfgParams
Function now takes a string instead of a db record, and now deserializes the string in the function.

Creating the db record to pass to TestSysDbRecord required translating the param string into a performance config with a deserialize method. This should be handled inside MIOpen. Renaming function to reflect that it now tests on a param string.


cderb:
Syncing with Fin.

junliume:
@cderb there are lots of changes to review :( Is it fixing the perfDB issue? Have we tested the DB generated after this PR, if it matches the "good" samples?

cderb:
> @cderb there are lots of changes to review :( Is it fixing the perfDB issue? 

@junliume sorry for the large diff.
Actually the changes relevant to MIOpen are just in this file: **src/include/miopen/any_solver.hpp**
The rest of the diff is fast-forwarding MIOpen's copy of Fin.
Unfortunately I also experienced some difficulties with the CI and how MIOpen tries to install with Fin. The commit log here has been me attempting to split those builds in the CI, because their CMake dependencies have directory conflicts.

cderb:
> Have we tested the DB generated after this PR, if it matches the "good" samples?

@junliume Unfortunately It won't be possible to test this change against a "good" sample, because there are none. Though I am working towards verifying the changes.
The issue with this function -was- that it eliminated entire solvers because of how the param string was passed. I've verified now that with the same databases, this update does not flag invalid entries across entire solvers. Though a large number of entries are flagged for some databases / datatypes.

atamazov (commented):
> Function now takes a string instead of a db record, and now deserializes the string in the function.

Can you please explain why this is necessary (in the description of the PR)?

/cc @DrizztDoUrden @averinevg

atamazov:
@cderb 
> The issue with this function -was- that it eliminated entire solvers because of how the param string was passed. I've verified now that with the same databases, this update does not flag invalid entries across entire solvers. Though a large number of entries are flagged for some databases / datatypes.

Missed this; no need to answer my previous question. Just update the PR's description, thanks.

atamazov:
@cderb Am I correct that `TestSysDbRecord` is used only in fin? If yes, then most likely I will have no objections.

cderb:
> @cderb Am I correct that `TestSysDbRecord` is used only in fin? If yes, then most likely I will have no objections.

@atamazov as far as I know, this function is only used by Fin.

junliume (dismissed):

fin/CMakeLists.txt
```diff
@@ -63,8 +63,8 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 endif()
 
 ############################################################
-# require C++17
-add_compile_options(-std=c++17)
+# require C++14
+add_compile_options(-std=c++14)
```
averinevg: MIOpen uses C++17. Why do you need C++14?

Jenkinsfile
```diff
@@ -78,13 +79,19 @@ def cmake_build(Map conf=[:]){
         setup_args = " -DCMAKE_CXX_COMPILER_LAUNCHER='ccache' -DCMAKE_C_COMPILER_LAUNCHER='ccache' " + setup_args
     }
 
+    if ( build_fin == "ON" )
+    {
+        setup_args = " -DMIOPEN_INSTALL_CXX_HEADERS=On " + setup_args
+    }
+
```
averinevg: Can we move FIN-related stuff to fin directory?

fin/CMakeLists.txt
cderb: This is in Fin's CmakeList and cpp 17 was crashing the build there.

Jenkinsfile
cderb: I modified the build order here because using MIOpen's CMakeList to build Fin was causing conflicts. This should be here so that we can use Fin in the MIOpen CI. This has the effect it had before of ensuring Fin is compatible with MIOpen and enables features like the perf db validity check.

averinevg: It is not necessary to use MIOpens CMakeList to build Fin, you can use CMakeList from fin directory. Also I googled a bit and it's possible to include one Jenkinsfile to another. You can store Fin Jenkinsfile in fin directory and import from it anything you need: functions, variables and even whole stages.

fin/src/include/tensor.hpp
```diff
@@ -57,6 +57,31 @@ using status_t = cl_int;
 using status_t = int;
 #endif
 
+inline miopenTensorLayout_t GetMemLayout(const std::string& s)
+{
+
+    if(s == "NCHW")
```
DrizztDoUrden: This should be a switch.

fin/tests/mem_layout_test.cpp
```diff
@@ -0,0 +1,94 @@
...
+                                      std::string(command["config"]["in_layout"]) +
+                                      ". Batch norm only support default NCHW"));
+            }
+        }
+    }
+}
+
+TEST(MemoryLayoutTest, TestGetMemLayout)
+{
+    miopenTensorLayout_t nchw_layout = fin::GetMemLayout("NCHW");
```
DrizztDoUrden: This should test ALL the cases.

src/include/miopen/any_solver.hpp
```diff
@@ -172,29 +171,31 @@ struct AnySolver
...
             PerformanceConfig config{};
-            bool success = record.GetValues(value.SolverDbId(), config);
+
+            bool success = config.Deserialize(params);
+            if(!success)
+                MIOPEN_LOG_WE("Perf params are obsolete or corrupt: "
+                              << params << ". Performance may degrade.");
+
             if(success)
                 success = value.IsValidPerformanceConfig(ctx, config);
```
DrizztDoUrden:
```suggestion
            if(!success)
            {
                MIOPEN_LOG_WE("Perf params are obsolete or corrupt: "
                              << params << ". Performance may degrade.");
                return false;
            }

            success = value.IsValidPerformanceConfig(ctx, config);
```

DrizztDoUrden (changes requested):

fin/src/include/tensor.hpp
cderb: iirc switches only work for integer values. This string would need to be recast as an enum.

fin/tests/mem_layout_test.cpp
cderb: I'd like to keep discussions on code from fin/ within the https://github.com/ROCmSoftwarePlatform/MIFin project.

Jenkinsfile
cderb: @averinevg here has been altered so that the cmakes are no longer intertwined.
As for loading the fin jenkinsfile, I don't think that will work. The load functionality in Jenkinsfile appears to work well for groovy files, but because a pipeline{...} definition is present in fin/Jenkinsfile, attempts to load the file will cause an error, as multiple definitions of these are not allowed.
If it would be sufficient to move the fin related build steps into a separate cmake_build function, I can provide that.

cderb:
Re-arranged Fin build in CI pipeline.

Jenkinsfile
```diff
@@ -38,10 +39,11 @@ def cmake_build(Map conf=[:]){
         config_targets = "package"
     }
 
+    env.MIOPEN_INSTALL = "${env.WORKSPACE}/install"
```
averinevg:
```suggestion
    def miopen_install_path = "${env.WORKSPACE}/install"
```

Jenkinsfile
```diff
@@ -99,9 +106,23 @@ def cmake_build(Map conf=[:]){
...
-            ${execute_cmd}
         """)
 
+    if ( build_fin == "ON" )
+    {
+        def fin_build = cmake_build_fin(env.MIOPEN_INSTALL)
+        cmd += """
+            export RETDIR=\$PWD
+            cd ${env.WORKSPACE}/fin 
+            ${fin_build}
```
averinevg:
```suggestion
        def fin_build_cmd = cmake_build_fin(env.MIOPEN_INSTALL)
        cmd += """
            export RETDIR=\$PWD
            cd ${env.WORKSPACE}/fin 
            ${fin_build_cmd}
```

Jenkinsfile
```diff
@@ -110,6 +131,30 @@ def cmake_build(Map conf=[:]){
         archiveArtifacts artifacts: "build/*.deb", allowEmptyArchive: true, fingerprint: true
     }
 }
+
+def cmake_build_fin(prefixpath){
```
averinevg:
```suggestion
def cmake_fin_build_cmd(prefixpath){
```

averinevg (changes requested):

src/include/miopen/any_solver.hpp
```diff
@@ -56,10 +56,10 @@ struct AnySolver
         assert(ptr_value != nullptr);
         return ptr_value->IsTunable();
     };
-    bool TestSysDbRecord(const ConvolutionContext& ctx, const DbRecord& record) const
+    bool TestPerfCfgParams(const ConvolutionContext& ctx, std::string params) const
```
averinevg:
```suggestion
    bool TestPerfCfgDeserialize(const ConvolutionContext& ctx, const std::string& params) const
```

averinevg (changes requested):

src/include/miopen/any_solver.hpp
cderb: This is testing if the params are a valid performance config, so this function name is partially misleading.
True, Deserialize is called to get the config, IsValidPerformanceConfig is afterwards called on the config to validate the params.

cderb: Update formal parameters to require a const string pointer.

src/include/miopen/any_solver.hpp
```diff
@@ -171,30 +171,36 @@ struct AnySolver
...
+            }
+
+            success = value.IsValidPerformanceConfig(ctx, config);
 
             return success;
         }
-        bool TestSysDbRecord(const ConvolutionContext& ctx,
-                             const DbRecord& record,
-                             std::false_type) const
+        bool TestPerfCfgParams(const ConvolutionContext&, std::string, std::false_type) const
```
averinevg:
```suggestion
        bool TestPerfCfgParams(const ConvolutionContext&, const std::string&, std::false_type) const
```

averinevg (changes requested):

src/include/miopen/any_solver.hpp
cderb: good catch.

averinevg (dismissed):

JehandadKhan (dismissed):

cderb:
Let's push this PR #1588 first. This should simplify this PR by reducing the diff.

junliume:
> Let's push this PR #1588 first. This should simplify this PR by reducing the diff.

After #1588 merged this PR has lots of conflicts which need to be resolved first :)

cderb:
Closing in favor of #1970

PR #1886, opened 2022-11-22
[NFC] Bump MLIR commit hash for ROCm 5.5 features

jerryyin:
This commit hash update include the fix for https://ontrack-internal.amd.com/browse/SWDEV-356492. After merging this PR, 
pytorch will be able to pull in latest MIOpen and be able to work with MLIR solvers.

CC: @jithunnair-amd Would you mind to use the branch and verify that the issue is gone after this bump?

PR #1804, opened 2022-10-04
[MLIR] Bumping release tag to rocm-5.4.0

yiqian1:
All regression tests have passed for the rocm-5.4.0 tag.

PR #1897, opened 2022-12-05
[mlir] Enable gfx103x for mlir non-xdlops tests

zhanglx13:
Retuning of the following MLIR solvers is required
- conv_igemm_mlir_fwd
- conv_igemm_mlir_bwd_wrw


jerryyin (dismissed):

atamazov:
@zhanglx13 These tests are not _regression_ tests. Please rename PR.

atamazov:
@jerryyin @zhanglx13 It seems like support for gfx103X was recently added to MLIR solvers. Can you please remind which PR did this?

atamazov:
...and what about gfx908 and gfx90a?

zhanglx13:
@atamazov This PR enabled gfx1030 from MLIR side: https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/778
I think mlir already supports gfx908 and 90a. @jerryyin, please correct me if I am wrong.

zhanglx13:
@atamazov There is the error messages "xxx convolution cannot be executed due to incorrect params" for f16 and f32 tests in the Full Tests stage for gfx1030. And such error messages come from convolutionocl.cpp and it seems that perf_db is empty. Does it mean the perf_db for gfx1030 is empty for the mlir solver since this is the first time gfx1030 is enabled for mlir sovlers? If so, how should we enable the tests without a perf_db?

jerryyin:
>  And such error messages come from convolutionocl.cpp and it seems that perf_db is empty.

@zhanglx13 I'm a bit doubtful about this. Typically when I get the conv cannot be executed it is usually due to that particular problem config wasn't enabled. Could you double check `release/rocm-5.4` incorporated all necessary changes for navi? (The other possibility is that we might need to pull a switch from solver side of applicability check) Since this is what MIOpen is using currently, you may need to checkout to that branch and try to reproduce the problem to see why it is still disabled. 

@atamazov gfx1030 is the most recently enabled arch but was enabled a couple of months back in MLIR. On reviewing this feature, we figured that the corresponding tests were not up-to-date therefore this PR. (This doesn't impact gfx908 or gfx90a and they'll continue to be supported)

atamazov:
@zhanglx13 
> @atamazov This PR enabled gfx1030 from MLIR side: [ROCmSoftwarePlatform/rocMLIR#778](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/778)

There should be a PR in MIOpen that changes the version of the MLIR library used in MIOpen so that MLIR includes https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/778. Most likely, that PR made changes to https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/requirements.txt. That is the PR I am interested in.

_If I am right_, then the problem is that it quietly made significant changes to MIOpen, but did not provide the necessary tests. We should prevent similar situations in the future.

@jerryyin _If I am right_, then please mark this PR with https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_blocker https://github.com/ROCmSoftwarePlatform/MIOpen/labels/testing

atamazov:
@jerryyin 
>> ...and what about gfx908 and gfx90a?

> (This doesn't impact gfx908 or gfx90a and they'll continue to be supported)

This PR updates non-xdlops MLIR solvers, any my question is related to these solvers only. IIRC these solvers were not applicable for gfx908/90a (and, therefore, `test_conv_igemm_mlir_fwd` and `test_conv_igemm_mlir_bwd_wrw` are disabled for these targets). Do we need to enable these tests for gfx908/90a?

atamazov:
@zhanglx13 
> @atamazov There is the error messages "xxx convolution cannot be executed due to incorrect params" for f16 and f32 tests in the Full Tests stage for gfx1030...

These messages mean that there is no applicable solvers.

zhanglx13:
@jerryyin I managed to reproduce the error using rocm/miopen:ci image. [Here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/537f7a55fcd667a332c2c00271cd6a85ea93d50d/src/ocl/convolutionocl.cpp#L528) is the only place where such error message could be generated. And it seems that the applicability check for the solver prints other messages like "invalid" or "empty" on the MIOpen side, and prints nothing on MLIR side.

atamazov:
@jerryyin 
> you may need to checkout to that branch and try to reproduce the problem to see why it is still disabled.

Maybe the reason is outdated docker image used for testing on CI. MLIR remains the same for long (til #1804 at least and even that PR seems to be NFC).

Please note that updating requirements does NOT update docker image on CI: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/537f7a55fcd667a332c2c00271cd6a85ea93d50d/Jenkinsfile#L113-L123
Hopefully @JehandadKhan can provide you with instructions on how to update MLIR library in our CI.

zhanglx13:
Alright, when I enabled the log, I confirmed that the solver is not applicable
```
MIOpen(HIP): Info2 [SearchForAllSolutions] ConvMlirIgemmFwd: Not applicable
```
The rocm/miopen:ci image was pushed 2 months ago, which seems to be the date that [PR#1804](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1804) was merged. 
Anyway, I'll try to rebuild the image with a newer version of MLIR to see if the error goes away.

Jenkinsfile
```diff
@@ -435,6 +438,12 @@ pipeline {
         NOCOMGR_flags   = " -DMIOPEN_USE_COMGR=Off"
     }
     stages{
+        stage("build docker") {
+            agent{ label rocmnode("nogpu") }
+            steps{
+                buildDocker("/opt/rocm")
+            }
+        }
```
atamazov: I do not think we need to test docker build functionality on each CI run. Anyway, a dedicated PR is required for such a change.

Jenkinsfile
```diff
@@ -309,6 +309,9 @@ def buildDocker(install_prefix)
     catch(Exception ex)
     {
         echo "Unable to locate image: ${image_name}."
+        echo "Unable to locate image: ${image_name}. Building image now"
+        retimage = docker.build("${image_name}", dockerArgs + ' .')
+        retimage.push()
```
atamazov: Not in this PR please.

Jenkinsfile
```diff
@@ -113,7 +113,7 @@ def cmake_build(Map conf=[:]){
 def getDockerImageName(prefixpath)
 {
     def branch =  sh(script: "echo ${scm.branches[0].name} | sed 's/[^a-zA-Z0-9]/_/g' ", returnStdout: true).trim()
-    def image = "${env.MIOPEN_IMAGE_URL}"
+    def image = "${env.MIOPEN_IMAGE_URL}_mlir266ad"
```
atamazov: This must be reverted.

### Proposal of the process of updating the CI docker image
- (A) Build CI docker image candidate and upload it to dockerhub (suggested naming: `rocm/miopen:ci_prXXXX_comment` where XXXX is number of PR that contains changes of requirements.txt and/or Dockerfile, and comment is optional)
- (B) Test develop with the candidate. Run the job manually. Use `DOCKER_IMAGE_OVERRIDE` to override standard image name with the name of the candidate.
- (C) Test PR with candidate image in the same way.

When all tests pass, MIOPEN_IMAGE_URL environment variable can be globally updated for the whole CI.

@junliume @JehandadKhan What do you think about this proposal? If no objections, I can document this in the dedicated ticket.

atamazov (changes requested):
General Jenkinsfile changes should be moved to a different PR.

zhanglx13:
@atamazov I do not mean to merge any changes to the Jenkinsfile in this PR. The last commit will be reverted after everything is working. 
- I need to rebuild a docker image with a newer rocMLIR commit id and [build#11](http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/enable-gfx1030-mlir-test/11/pipeline/) shows that it's working.
- However, the docker image the CI pulls needs to be updated for this PR to pass all the test after the last commit is reverted.

@jerryyin I think you may also want to invite @atamazov to the meeting about docker image update.

atamazov:
@zhanglx13 
> I do not mean to merge any changes to the Jenkinsfile in this PR. The last commit will be reverted after everything is working.

So this is misunderstanding, okay. Please convert this PR into draft until it is ready for final review and merging, thanks!

Jenkinsfile
junliume: @atamazov thanks and yes it should be documented. Step (B) can be set manually on Jenkins webpage for the CI docker candidate, if one prefer not to change Jenkinsfile

JehandadKhan: @atamazov that makes sense to me

jerryyin: Thanks for the proposal, it makes the opaque process with much more clarity. Though, I need a couple of clarifications:
 - (A) This seems to be a manual process. I'd prefer that MIOpen end of CI provides a docker build job dedicated to build image according to a particular branch. This way, we can easily have an URL to refer to that image is successfully built.
 - (B) This step seems unnecessary because ultimately changing the docker environment requires a pull request. A dependency such as rocMLIR doesn't have to have two passing record: one for develop branch and another for PR. (Occasionally when MIOpen change is required, the new docker environment would not work with the pre-existing develop branch)
 - (C) Sounds good, so we basically manually trigger CI and fill in that `DOCKER_IMAGE_OVERRIDE`. However, the current implementation of this seems to be always reverting to an empty string after each run, which seems to be a big inconvenience and can potentially waste MIOpen CI resources. A developer need to manually go and kick off CI with the override for every commit.

> When all tests pass, MIOPEN_IMAGE_URL environment variable can be globally updated for the whole CI.

Where do we set this? I only see this variable referenced from dockerfile but didn't see it elsewhere in the code base.

atamazov: @jerryyin Yes, my proposal assumes that some manual work is necessary when a developer wants to update CI docker image. However, it seems workable out of the box, and we can begin using it right away. The process can be improved/better automated in the future. Let's figure out the drawbacks clearly and open a ticket.

> "(A)" seems to be a manual process. I'd prefer that MIOpen end of CI provides a docker build job dedicated to build image according to a particular branch. This way, we can easily have an URL to refer to that image is successfully built.

Building docker image on each CI job could be a waste of resources, because the majority of PRs do not affect the image. But we can add the "Docker Build" step to the Jenkinsfile (and make it switched off by default). A developer can enable this step manually, then download the built image (artifact) from CI, tag it properly and upload to dockerhub.
 
> "(B)" seems unnecessary because ultimately changing the docker environment requires a pull request...

AFAIU pull request is not required to change the docker environment. It is enough to upload a new docker image to `rocm/miopen:ci`, or upload some other docker image and change `MIOPEN_IMAGE_URL` env variable on the nodes.

The purpose of (B) is to check if the new docker image can be used on CI with current develop. This could be necessary, for example, if we decide to revert the PR due to some reason.

> "(C)" sounds good, so we basically manually trigger CI and fill in that DOCKER_IMAGE_OVERRIDE. However... a developer need to manually go and kick off CI with the override for every commit.

When (B) and (C) pass, a developer should update `MIOPEN_IMAGE_URL` and then all the further jobs will use the new image. A developer need to manually trigger CI only twice, during steps (B) and (C).

>> MIOPEN_IMAGE_URL environment variable can be globally updated for the whole CI.

> Where do we set this?

IIRC the Jenkins' Web UI should allow this, but I can't give a definitive answer because I do not have access to our CI as of today. I believe that @JehandadKhan or @junliume know the exact answer.

jerryyin: @atamazov Agreed, we'll use the manual way to bump docker environment in this PR, and improve it over time.

> we can add the "Docker Build" step to the Jenkinsfile (and make it switched off by default). A developer can enable this step manually

Sounds prefect. Yes, I was thinking about another jenkins job that tracks dependency changes but I think your proposal is better. This is just so that we have the passing record of the docker build somewhere and can refer to when things go wrong.

> AFAIU pull request is not required to change the docker environment.

Just speaking from the rocMLIR integration, we have a commit tracking from `requirements.txt` and it will be very bad if the docker used a different commit id with the one from the `requirements.txt`. So MIOpen should not tolerate us implicitly "swap out" a docker environment that used a different commit#. I think I get it now the purpose of step (B) is to make it clear about whether a dependency can have breaking changes. Since rocMLIR is backward compatible with MIOpen this seems a reasonable step to have.

> a developer should update MIOPEN_IMAGE_URL and then all the further jobs will use the new image

I actually cannot find this `MIOPEN_IMAGE_URL` field but instead can only find `DOCKER_IMAGE_OVERRIDE`. So it'd be a nice clarification if there's a way to persist this to an environment variable.

zhanglx13:
@junliume I built a docker image (`rocm/miopen:ci_mlir_branch_rocm5.5`) based on the tip of mlir release branch and it is being tested. Once all tests pass, do you need to update `rocm/miopen:ci` before merging this PR or vise versa?

jerryyin (approved):
Action items after this PR is merged:
 1. Rename `rocm/miopen:ci_mlir_branch_rocm5.5` to `rocm/miopen:ci` to pull in latest testing for mlir static library
 2. Start ROCm 5.5 release tuning for all MLIR solvers, xdlops and nonxdlops included.

atamazov:
@jerryyin

> 1. Rename rocm/miopen:ci_mlir_branch_rocm5.5 to rocm/miopen:ci to pull in latest testing for mlir static library

This does not match the proposal at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1897#discussion_r1044798898. Shall we agree on the process of updating docker image for CI first? We can open a discussion or an issue to discuss, agree, and document it.

> 2. Start ROCm 5.5 release tuning for all MLIR solvers, xdlops and nonxdlops included.

Why this is necessary? Do you expect any perf regressions with 5.5? ~Basically all the perf-db changes necessary for 5.5 should be already merged into develop (from the 5.5 release branch)~

/cc @cderb @JehandadKhan

jerryyin:
@atamazov 

> This does not match the proposal

I think I missed step B, correct? Reason I skip it is due to the amount of change in this PR is straightforward and all it did is to relax the testing range. I don't quite see we need to run the tip of develop again in this PR (with a narrower coverage). Please let me know what else concerns you.

> Why this is necessary? Do you expect any perf regressions with 5.5?

We have made fairly significant changes on the kernel side over past couple of months that will make past selected perf config obsolete and cause perf regressions. 

> Basically all the perf-db changes necessary for 5.5 should be already merged into develop

ROCm 5.5 is the upcoming release. Did you mean to say ROCm 5.4 changes are already merged?

atamazov:
@jerryyin 

> I think I missed step B, correct? Reason I skip it is due to the amount of change in this PR is straightforward and all it did is to relax the testing range...

~The purpose of (B) is to check if the **_new docker image_** can be used on CI with **_current_** develop. We want to know for sure if there are any breaking changes in the new docker image. In this specific case, I would like to ensure that "old" (current) MIOpen is able to work with newer MLIR library (just in case if we will revert MLIR solver changes in the future due to some reason).~

~Another~ The mismatch is step (A) where I recommend keeping distinct name for the new docker image. This allows us to step back to any previous docker image is need arise.

> We have made fairly significant changes on the kernel side over past couple of months that will make past selected perf config obsolete and cause perf regressions.

Thanks, this is important to know.

@zhanglx13 The PR should be marked with https://github.com/ROCmSoftwarePlatform/MIOpen/labels/IMPACTS_PERFORMANCE_DB. Description should say that retuning of MLIR solvers is required (please include explicit list of the affected solvers).

> ROCm 5.5 is the upcoming release...

Oops, now I see, sorry for confusion. `rocm/miopen:ci_mlir_branch_rocm5.5` means "Current CI image updated with ROCm 5.5 release candidate MLIR library".

[Important] Are you aware of which version of MLIR library is included into Mainline? (Or it is not included there and only installed for MIOpen?) This is important to know, because, for example, if Mainline uses older version of MLIR library (e.g. `rocm-5.4.0 -H sha256:3823f455e`) **_and_** we tune MIOpen for the rocm5.5-RC MLIR `56a820ee3`, then QA will see perf regressions in the Mainline (which is interpreted as a blocker for promotion)

atamazov:
@jerryyin After second look at the PR I agree that in this specific case step (B) can be safely skipped.

atamazov:
@jerryyin I was under the impression that this PR makes changes in the MLIR solvers

atamazov:
@jerryyin Another small topic --

> There should be a PR in MIOpen that changes the version of the MLIR library used in MIOpen so that MLIR includes [ROCmSoftwarePlatform/rocMLIR#778](https://github.com/ROCmSoftwarePlatform/rocMLIR/pull/778)...

That PR is #1886...

> @jerryyin _If I am right_, then please mark this PR with https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_blocker ...

...and it have been "reverted" in #1901, so this PR is not a blocker anymore from my point of view. It adds MLIR support for gfx1030 (which urgency I am not aware of). I recommend updating the urgency label and adding https://github.com/ROCmSoftwarePlatform/MIOpen/labels/enhancement.

jerryyin:
@atamazov No worries.

>  Are you aware of which version of MLIR library is included into Mainline?

If you are talking about ROCm 5.4 mainline, then the commit hash of MLIR library is `ROCmSoftwarePlatform/rocMLIR@rocm-5.4.0`. I don't think we have bumped the (cmake) version for MLIR, is that right @zhanglx13? Or @atamazov is that a requirement to bump the cmake version? Just want to make those clarifications to make sure we are talking about the same thing.

I think ROCm 5.4 has been cut-off a long time ago and this PR is in preparation for ROCm 5.5 in coming weeks. So if we merge it now, it will not make its way back to ROCm 5.4 mainline if that's your concern.

> If I am right, then the problem is that it quietly made significant changes to MIOpen, but did not provide the necessary tests.

As it turned out in from the testing situation, that didn't happen. The commit pointer bump bring in the ability to do Navi test, this capability didn't exist in the current ROCm 5.4.0 tag from develop branch.

> and it have been "reverted" in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1901, so this PR is not a blocker anymore from my point of view.

It is still an urgency blocker from a number of aspects:
 1. Release point of view. We want this PR to be merged as soon as possible such that Tuna can do the tuning for ROCm 5.5 release, and push the corresponding perf db changes in time. 
 2. This new commit pointer bump incorporated all changes necessary to workaround the breakage identified from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1901. Once this PR is merged, we can ask Pytorch to do a new round of MIOpen integration test to make sure they can build MIOpen.

Once all testing has passed in MIOpen and Pytorch, we'd tag the current commit # as the ROCm-5.5.0 and mark this as our release for ROCm 5.5.

zhanglx13:
@jerryyin rocMLIR is still in version 1.0.0, which is also the version that MIOpen is looking for.

atamazov:
@jerryyin 
> If you are talking about ROCm 5.4 mainline...

No. IIRC there is no such thing as ROCm 5.4 mainline (and I guess you mean ROCm 5.4 _release_). Conceptually, the release pipeline looks like this: `Development` -> `Staging` -> `Mainline` -> `Release-ROCm-x.y`. Releases are created by "taking a copy" of Mainline state (often implemented by creating release branches). After Release is cut off, adding new features to it requires special approvals.

❓ **_Is MLIR library included into ROCm directly (as ROCm component) or it comes there only as a dependence of MIOpen?_**

> I don't think we have bumped the (cmake) version for MLIR, is that right @zhanglx13? Or @atamazov is that a requirement to bump the cmake version? 

There is no such requirement. IIRC versioning of ROCm components is independent of versioning of the ROCm itself. For ex., the same version of MLIR library can be included in several different versions of ROCm. 

> Just want to make those clarifications to make sure we are talking about the same thing.

👍

atamazov (approved):
LGTM but we need to update docker image for CI first.

atamazov:
> @jerryyin rocMLIR is still in version 1.0.0, which is also the version that MIOpen is looking for.

We want rocMLIR which is _compatible_ with 1.0.0 (because we do not use `EXACT` in `find_package()`)

jerryyin:
@atamazov Thanks for the clarifications, yes I think I meant to say that release was already cut off for the ROCm 5.4 and this is considered to be new feature for ROCm 5.5.

>  we need to update docker image for CI first

Are you talking about an additional commit in this PR that renames the docker image name in Jenkinsfile from `rocm/miopen:ci` to `rocm/miopen:ci_1897`? @zhanglx13 and if @atamazov confirms that this is the only thing left to do, could you help take a copy of existing docker image and upload according to the suggested naming convention? Then add a commit to take the docker image with new name.

@zhanglx13 This is a side topic but from now on we should think of a frequency to bump the cmake version. This will be important once we decide to give up support to an old ROCm with its compatibility code. 

/CC: @junliume This is to give you a note that all CI tests have passed and we are likely only needs to do a rename till it can merge.

zhanglx13:
I would suggest we keep the newly built docker image as `rocm/miopen:ci_mlir_branch_rocm5.5` so that we can tell from the tag why we built this image.
If I understand correctly, @atamazov wants to update the `rocm/miopen:ci` image with the above tag on docker hub before merging this PR. Therefore, the develop branch can be tested with the latest MLIR release. 
And since @junliume is going to update the `rocm/miopen:ci` image, could you please also rename the tag of the newly built one (from `ci_mlir_branch_rocm5.5` to `ci_1897`) if everyone else agrees that it is necessary?

atamazov:
@zhanglx13 
> I would suggest we keep the newly built docker image as rocm/miopen:ci_mlir_branch_rocm5.5 so that we can tell from the tag why we built this image.

The reference to PR (`rocm/miopen:ci_1897`) allows to get full information what is changed in the image and why. Maybe format like `rocm/miopen:ci_prXXXX_comment` (`rocm/miopen:ci_pr1897_mlir_branch_rocm5.5` in this specific case) would satisfy both of us?

atamazov:
@zhanglx13 
> If I understand correctly, @atamazov wants to update the rocm/miopen:ci image with the above tag on docker hub before merging this PR. Therefore, the develop branch can be tested with the latest MLIR release.

Not exactly. My (updated) proposal is to upload docker named like `rocm/miopen:ci_prXXXX_comment` to dockerhub and then update `MIOPEN_IMAGE_URL` (where Jenkins takes the docker URL from) with this name on all CI nodes. Then all the further jobs will use the new image.

atamazov:
@zhanglx13 @jerryyin WRT https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1897#issuecomment-1362051352 - please consult @JehandadKhan or @junliume to find if this is acceptable and how this can be done.

jerryyin:
> My (updated) proposal is to upload docker named like rocm/miopen:ci_prXXXX_comment to dockerhub and then update MIOPEN_IMAGE_URL (where Jenkins takes the docker URL from) with this name on all CI nodes. 

@junliume It looks like @atamazov's suggestion is to use `MIOPEN_IMAGE_URL` instead of overriding `rocm/miopen:ci`. I don't have a good idea of how to use that variable but just want to point out. (It looks like we ended up with that overriding approach).

atamazov:
@junliume @jerryyin At https://hub.docker.com/r/rocm/miopen/tags I see both `rocm/miopen:ci` and `rocm/miopen:ci_pr1897_mlir_branch_rocm5.5`, having the same digest.

Functionally, this partially matches my proposal: you can quickly check the digest of `rocm/miopen:ci` and find another image with the same digest. The latter gives the idea were current CI image came from. However, switching from one image to another image is not as simple as updating the `MIOPEN_IMAGE_URL` env var on the nodes. You need to upload new image with the `rocm/miopen:ci` tag for that.

PR #1900, opened 2022-12-09
[SWDEV-358170][Workaround] Specify -std=c++14 so boost builds using latest Clang and devtoolset-7

junliume:
"It's probably a bug in devtoolset-7's C++ library support for C++17 that got exposed by clang's change of its default to C++17. "
**and updating devtoolset is not an option**

Error:
```
./boost/asio/detail/memory.hpp:89:15: error: no member named 'aligned_alloc' in namespace 'std'; did you mean simply 'aligned_alloc'?
  void* ptr = std::aligned_alloc(align, size);
              ^~~~~~~~~~~~~~~~~~
              aligned_alloc
/usr/include/stdlib.h:508:14: note: 'aligned_alloc' declared here extern void *aligned_alloc (size_t __alignment, size_t __size)
             ^
1 error generated.
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 sqlite3@3.17 -DCMAKE_POSITION_INDEPENDENT_CODE=On
-boost@1.79 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build -DCMAKE_CXX_FLAGS=" -Wno-enum-constexpr-conversion "
+boost@1.79 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build -DCMAKE_CXX_FLAGS=" -std=c++14 -Wno-enum-constexpr-conversion "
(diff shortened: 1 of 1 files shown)


scchan (approved):
LGTM
For reference, here's the C++ implementation status of libstdc++.  C++17 is only stable after version 9.1 so enabling C++17 with earlier version of the toolchain should be avoided. 

https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017

JehandadKhan (approved):

junliume:
> LGTM For reference, here's the C++ implementation status of libstdc++. C++17 is only stable after version 9.1 so enabling C++17 with earlier version of the toolchain should be avoided.
> 
> https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017

Hi @scchan thanks! However, C++17 is enabled by default for the latest clang. Do you think the issue you mentioned should be fixed or patched at the compiler level instead? This is a non-MIOpen bug as you can see, the included version of boost is very recent.

Issue #1913, opened 2022-12-26
[tests][MLIR][gfx103x] test_conv_igemm_mlir failures with old libMLIR


atamazov:
@jerryyin The reason is #1897.

E.g. with stock ROCm 5.2.3, these tests fail:
```
test_conv_igemm_mlir_fwd (Failed)
test_conv_igemm_mlir_bwd_wrw (Failed)
```

We can disable MLIR tests on GFX10 with old libMLIR but we need to know the libMLIR version in tests/CMakeLists.txt (and that version should be > than version of older libMLIR).

@junliume https://github.com/ROCmSoftwarePlatform/MIOpen/labels/testing https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_normal

The W/A is using `-DMIOPEN_TEST_MLIR=Off` when running CMake on GFX10.

atamazov:
@jerryyin
> We can disable MLIR tests on GFX10 with old libMLIR but we need to know the libMLIR version in tests/CMakeLists.txt (and that version should be > than version of older libMLIR).

FYI the fix is partially implemented in #1914.

atamazov:
@jerryyin Another option is do not make any changes and use the W/A (see description of this ticket) when necessary. Why: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1914#discussion_r1058530284.

atamazov:
Finally decided to provide a fix, see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1914#discussion_r1060687570

ppanchad-amd:
@atamazov Can we close this ticket? Thanks!

PR #1914, opened 2022-12-27
[Navi21][tests] Disable MLIR tests on gfx103x when libMLIR <= 1.0.0. [CI][Navi21][WORKAROUND] Enable MLIR tests on gfx103x with current docker image

atamazov:
This partially resolves issue #1913.

- MLIR tests on gfx103x are disabled when libMLIR reports version <= 1.0.0
  - `cmake ... -DMIOPEN_LIBMLIR_SUPPORTS_GFX103X=On` can be used to override.
- Added WORKAROUND_ISSUE_1913 to Jenkinsfile. It enables MLIR testing on GFX10 with current CI docker image.

@junliume https://github.com/ROCmSoftwarePlatform/MIOpen/labels/testing https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_normal

@jerryyin Can you please review this and take care about implement the remaining fixes (proposal):
- Make changes in libMLIR so its version > 1.0.0
- Bump libMLIR commit hash so the above change is included for MIOpen.
- Update docker image for CI with new libMLIR.
- Remove WORKAROUND_ISSUE_1913.

⚠️ Jenkinsfile changes from forks are not used in CI (due to security reasons). Therefore we need to check if WORKAROUND_ISSUE_1913 really works (enables `test_conv_igemm_mlir_fwd` and `test_conv_igemm_mlir_bwd_wrw` on Navi21) after merging this PR. @junliume @jerryyin can you please take care about this?


CMakeLists.txt
```diff
@@ -345,10 +345,14 @@ if(MIOPEN_USE_MLIR)
...
         else()
             message(STATUS "Build with library libMLIRMIOpen: " ${LIBMLIRMIOPEN})
+            set(rocMLIR_VERSION_MAJOR 0)
+            set(rocMLIR_VERSION_MINOR 0)
+            set(rocMLIR_VERSION_PATCH 1)
         endif()
     else()
         message(STATUS "Build with rocMLIR::rockCompiler ${rocMLIR_VERSION}")
     endif()
+    math(EXPR rocMLIR_VERSION_FLAT "(${rocMLIR_VERSION_MAJOR} * 1000 + ${rocMLIR_VERSION_MINOR}) * 100000 + ${rocMLIR_VERSION_PATCH}")
```
jerryyin, line 356: Do you mean 

> `${rocMLIR_VERSION_MAJOR} * 100000 + ${rocMLIR_VERSION_MINOR}) * 1000 + ${rocMLIR_VERSION_PATCH}`

I typically see major's number comes first so want to double check.

test/CMakeLists.txt
```diff
@@ -123,6 +122,13 @@ if(NOT (MIOPEN_TEST_GFX900 OR MIOPEN_TEST_GFX906 OR MIOPEN_TEST_GFX908 OR MIOPEN
         set(MIOPEN_TEST_GPU_DETECTION_FAILED TRUE)
     endif()
 endif()
+
+set_var_to_condition(MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT ${MIOPEN_USE_MLIR} AND (${rocMLIR_VERSION_FLAT} GREATER 100000000))
+option( MIOPEN_LIBMLIR_SUPPORTS_GFX103X "libMLIR capability to support GFX10. Enables MLIR tests on GFX103X." ${MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT})
+
+set_var_to_condition(MIOPEN_TEST_MLIR_DEFAULT ${MIOPEN_USE_MLIR} AND NOT (${MIOPEN_TEST_GFX103X} AND NOT ${MIOPEN_LIBMLIR_SUPPORTS_GFX103X}))
+option( MIOPEN_TEST_MLIR "Test for MLIR compilation backend" ${MIOPEN_TEST_MLIR_DEFAULT} )
```
jerryyin, line 135: Okay, so this looks good. the 103x tests will be skipped when `${MIOPEN_LIBMLIR_SUPPORTS_GFX103X}` is Off.

I want to point out that in the past, we don't maintain backward compatibility of the MIOpen tests when we add new capabilities to MLIR backend. Do we want to change that now? It means that list of conditions appended to `MIOPEN_TEST_MLIR_DEFAULT` will always be incrementing to disable stages when condition doesn't satisfy.

CMakeLists.txt
```diff
@@ -345,10 +345,14 @@ if(MIOPEN_USE_MLIR)
             Refer to https://github.com/ROCmSoftwarePlatform/MIOpen#installing-the-dependencies")
         else()
             message(STATUS "Build with library libMLIRMIOpen: " ${LIBMLIRMIOPEN})
+            set(rocMLIR_VERSION_MAJOR 0)
+            set(rocMLIR_VERSION_MINOR 0)
+            set(rocMLIR_VERSION_PATCH 1)
```
jerryyin, line 351: We can file and backport MLIR commit to bump it from 1.0.0 to 1.1.0 in the static lib.

Why set the version to be 0.0.1 now? Is it a local workaround to disable MLIR tests? I think the best way to fix MLIR test failures is not this but rather actually bump to a new docker environment. Then MLIR will continue to be tested and problem "fixed" by the new package.

Jenkinsfile
```diff
@@ -23,6 +23,8 @@ def cmake_build(Map conf=[:]){
     def build_envs = "CTEST_PARALLEL_LEVEL=4 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0 " + conf.get("build_env","")
     def prefixpath = conf.get("prefixpath","/opt/rocm")
     def mlir_args = " -DMIOPEN_USE_MLIR=" + conf.get("mlir_build", "ON")
+    // WORKAROUND_ISSUE_1913:
+    mlir_args = mlir_args + " -DMIOPEN_LIBMLIR_SUPPORTS_GFX103X=ON"
```
jerryyin, line 27: I can remove those two lines in a follow-up PR, but I don't quite get why set this to ON here? This seems to be conflict with the version set to 0.0.1 (indicating MIOPEN_LIBMLIR_SUPPORTS_GFX103X=Off`).

test/CMakeLists.txt
```diff
@@ -123,6 +122,13 @@ if(NOT (MIOPEN_TEST_GFX900 OR MIOPEN_TEST_GFX906 OR MIOPEN_TEST_GFX908 OR MIOPEN
         set(MIOPEN_TEST_GPU_DETECTION_FAILED TRUE)
     endif()
 endif()
+
+set_var_to_condition(MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT ${MIOPEN_USE_MLIR} AND (${rocMLIR_VERSION_FLAT} GREATER 100000000))
+option( MIOPEN_LIBMLIR_SUPPORTS_GFX103X "libMLIR capability to support GFX10. Enables MLIR tests on GFX103X." ${MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT})
```
jerryyin, line 132: I'm not sure this option to make it overridable is necessary? I think `MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT` as a condition to check 103x capability would be sufficient.

jerryyin (commented):
Thanks for staying on top of MLIR integration! I leave mostly clarification comments. gfx103x is one example of the new capability we have added but this can serve as a template in future releases. :thumbsup:

test/CMakeLists.txt
atamazov: @jerryyin 
> I want to point out that in the past, we don't maintain backward compatibility of the MIOpen tests...

Hmm... Good point. It was correct provided that MLIR is a part of MIOpen. Can you please [answer this Q](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1897#issuecomment-1360366088):

> ❓ **_Is MLIR library included into ROCm directly (as ROCm component) or it comes there only as a dependence of MIOpen?_**

Another question: do we want/require users of old ROCm to install the newest version of libMLIR? If not, then tests shall pass with vanilla old ROCm.

For now it looks like we want to maintain bwd compatibility (including tests).

jerryyin: > ❓ Is MLIR library included into ROCm directly (as ROCm component) or it comes there only as a dependence of MIOpen?

MLIR library is not a ROCm component. It only ship as a dependence of MIOpen and MIGraphX. The only way a client like Pytorch can pull in MLIR is through MIOpen's `requirements.txt`.

> do we want/require users of old ROCm to install the newest version of libMLIR?

No one require us to give that guarantee. (and we don't test that scenario because of it.) Realistically speaking we haven't observed any version mismatch issues in the past too. Though most of the cases we don't have a dependency on ROCm version. ROCm update will only rarely break our internal test, there's a even smaller chance it break the static library.

CMakeLists.txt
atamazov: You've missed inner braces, around `(${rocMLIR_VERSION_MAJOR} * 1000 + ${rocMLIR_VERSION_MINOR})`. I maintain the same format as used for `MIOPEN_hip_VERSION_FLAT`, see line 236.

CMakeLists.txt
atamazov: @jerryyin 
> We can file and backport MLIR commit to bump it from 1.0.0 to 1.1.0 in the static lib.

👍 

> Why set the version to be 0.0.1 now? 

This is for old libMLIRMIOpen, when `find_package(rocMLIR...)` fails.

When `find_package` succeeds, it sets rocMLIR_VERSION_MAJOR/MINOR/PATCH to 1/0/0 respectively + more, see https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-version-selection

Jenkinsfile
atamazov: @jerryyin From PR description:
> Added WORKAROUND_ISSUE_1913 to Jenkinsfile. It enables MLIR testing on GFX10 with current CI docker image.

This is necessary to continue testing of MLIR solvers with GFX10 in our CI with current docker image.

Please do not forget to update CI docker image on dockerhub before removing this W/A.

test/CMakeLists.txt
atamazov: @jerryyin Overriding is necessary for WORKAROUND_ISSUE_1913 to work (and may be handy in the future). But please feel free to remove `option()` together with W/A in your PR, if you like.

test/CMakeLists.txt
atamazov: @jerryyin 
> MLIR library is not a ROCm component. It only ship as a dependence of MIOpen and MIGraphX. The only way a client like Pytorch can pull in MLIR is through MIOpen's requirements.txt.

If this is the case, then both #1913 and this PR can be closed.

Then we should require some actions from the end users who use old libMLIR but want to update MIOpen to the most recent version: _they must install the new version of libMLIR_ (or globally disable MLIR).

❗  Please let me know your opinion.

jerryyin: @atamazov Yes, I think enforcing them to always install the correct version of MLIR sounds like the right way to do it. Thinking about it, I think we can close #1913 and this PR but instead:
 - Bump rocMLIR version to 1.1.0 (and make sure that we bump at least the minor version in each ROCm release, like MIOpen do)
 - Amend https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2d90671f257244fd0d17b686f4ac5c7ba4ebfbfd/CMakeLists.txt#L337 to `find_package(rocMLIR 1.0.0 EXACT CONFIG)`

Let me if you agree with it.

CMakeLists.txt
jerryyin: Ah thanks for your explanation, this is very defensive cmake definitions 👍

CMakeLists.txt
jerryyin: Gotcha now thanks for the explanation.

test/CMakeLists.txt
jerryyin: It is still unclear to me why the existing condition will always block any GFX103X stages no matter what. I think `MIOPEN_TEST_MLIR_DEFAULT` is meant to be on when:
 - It isn't a gfx103x testing stage (because they will always be supported), Or
 - It is a gfx103x testing stage AND MLIR supports gfx103x

Translating those to conditionals:
> `set_var_to_condition(MIOPEN_TEST_MLIR_DEFAULT ${MIOPEN_USE_MLIR} AND` 
    `((NOT ${MIOPEN_TEST_GFX103X}) OR (${MIOPEN_TEST_GFX103X} AND ${MIOPEN_LIBMLIR_SUPPORTS_GFX103X})))`

Jenkinsfile
jerryyin: Backing off, the current condition as is, MLIR version is 1.0.0, that means `MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT` is ON, therefore `MIOPEN_LIBMLIR_SUPPORTS_GFX103X` is ON.

With above it seems to me that `MIOPEN_LIBMLIR_SUPPORTS_GFX103X=ON` is a natural derived decision from your cmake logic and doesn't need to be overridden again.

jerryyin: Backing off, the current condition as is, MLIR version is 1.0.0, that means `MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT` is ON, therefore `MIOPEN_LIBMLIR_SUPPORTS_GFX103X` is ON.

With above it seems to me that `MIOPEN_LIBMLIR_SUPPORTS_GFX103X=ON` is a natural derived decision from your cmake logic and doesn't need to be overridden again.

test/CMakeLists.txt
atamazov: @jerryyin On the other hand, it's good to have as much flexibility as possible. This may become handy in the future. For example, if the latest version of MIOpen can work with an outdated version of libMLIR, then we can use this possibility to triage issues.

Another imaginary example is when QA finds a complex bug in a release candidate branch, and that bug can be resolved by reverting to a previous version of libMLIR. Forcing MIOpen to always use the "correct" version of libMLIR would make this impossible.

❓ Of course, maintaining flexibility takes effort, but it's often worth it. Your opinion?

atamazov: [Informative] @jerryyin 
> make sure that we bump at least the minor version in each ROCm release, like MIOpen do

We don't actually do this. Yes, we bump MIOpen version with each new release of _MIOpen_. And usually we release new MIOpen with each new ROCm release, but from time to time we include the same MIOpen (with the same version) into a new ROCm.

atamazov: @jerryyin 
> Bump rocMLIR version to 1.1.0

I agree 🟢

Jenkinsfile
atamazov: > Backing off, the current condition as is, MLIR version is 1.0.0, that means `MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT` is ON, therefore `MIOPEN_LIBMLIR_SUPPORTS_GFX103X` is ON.

No, this PR _by default_ disables MLIR testing on GFX10 for 1.0.0. MIOpen wants `${rocMLIR_VERSION_FLAT} *GREATER* 100000000` to enable it, see line 126. Tested.

test/CMakeLists.txt
atamazov: The conditional proposed by you is correct. But `(!a || (a && b))` is equivalent to `(!a || b)` which is equivalent to `!(a && !b)`. Therefore, the proposed conditional is equivalent to the existing conditional at line 129. The construct at line 129:

- `...AND NOT (${MIOPEN_TEST_GFX103X} AND NOT ${MIOPEN_LIBMLIR_SUPPORTS_GFX103X})...`

...should be read as: Let's _enforce_ the result to FALSE if we test on GFX10 _and_ libMLIR does _not_ support GFX10. Otherwise do not change the result.

Jenkinsfile
jerryyin: Ahh it was my bad that I didn't get to see the GREATER condition. Thanks for the explanation again. I see why you need the override now.

test/CMakeLists.txt
jerryyin: Thanks for the interpretation, make sense to me! Sorry for taking this many round of back and forth!

jerryyin (dismissed):
LGTM

test/CMakeLists.txt
jerryyin: @atamazov I'd say this is a good example on making tests backward compatible so let's ship it.

> Another imaginary example is when QA finds a complex bug in a release candidate branch, and that bug can be resolved by reverting to a previous version of libMLIR.

Realistically speaking this will probably never happen 😄. First off, QA doesn't find complex bug, they find the bare minimal to reproduce bug (usually build related). Secondly, they don't do any more triaging than what is necessary, and MLIR related issues usually ended up re-assigned to me. - end of the rant. Since those issues come back to me usually, I wasn't as considerate in MIOpen end as I should. But you are right, this can be convenient if in my own triage to test lastly released package and narrow the scope here.

atamazov: [Informative] @jerryyin I mean issues like network non-convergence (or performance drops) that QA finds in MIOpen from time to time and assigns to MIOpen.

test/CMakeLists.txt
atamazov: No problem at all ;)

junliume:
[fail.log](https://github.com/ROCmSoftwarePlatform/MIOpen/files/10390224/fail.log)
@atamazov the CI fails with the following cmake error (log attached above):
```
[2023-01-11T08:25:18.807Z] CMake Error at CMakeLists.txt:29 (if):

[2023-01-11T08:25:18.807Z]   if given arguments:

[2023-01-11T08:25:18.807Z] 

[2023-01-11T08:25:18.807Z]     "OFF" "AND" "(" "GREATER" "100000000" ")"

[2023-01-11T08:25:18.807Z] 

[2023-01-11T08:25:18.807Z]   Unknown arguments specified

[2023-01-11T08:25:18.807Z] Call Stack (most recent call first):

[2023-01-11T08:25:18.807Z]   test/CMakeLists.txt:126 (set_var_to_condition)

[2023-01-11T08:25:18.807Z] 

[2023-01-11T08:25:18.807Z] 

[2023-01-11T08:25:18.807Z] -- Configuring incomplete, errors occurred!
```
to reproduce"
```
CXX=/opt/rocm/llvm/bin/clang++ CXXFLAGS='-Werror'  cmake -DMIOPEN_TEST_FLAGS=' --disable-verification-cache ' -DCMAKE_BUILD_TYPE=release -DBUILD_DEV=On -DMIOPEN_USE_MLIR=OFF -DMIOPEN_LIBMLIR_SUPPORTS_GFX103X=ON -DMIOPEN_GPU_SYNC=Off -DBUILD_SHARED_LIBS=Off -DCMAKE_PREFIX_PATH=/opt/rocm    ..
```

test/CMakeLists.txt
```diff
@@ -123,6 +122,13 @@ if(NOT (MIOPEN_TEST_GFX900 OR MIOPEN_TEST_GFX906 OR MIOPEN_TEST_GFX908 OR MIOPEN
         set(MIOPEN_TEST_GPU_DETECTION_FAILED TRUE)
     endif()
 endif()
+
+set_var_to_condition(MIOPEN_LIBMLIR_SUPPORTS_GFX103X_DEFAULT ${MIOPEN_USE_MLIR} AND (${rocMLIR_VERSION_FLAT} GREATER 100000000))
```
junliume, line 131: @atamazov maybe we should check if `rocMLIR_VERSION_FLAT` is set here? It is only set when `MIOPEN_USE_MLIR` is true

atamazov: Thanks for informing. I missed this case, will fix tomorrow.

atamazov: @junliume I thought that CMake employs _short-circuiting_ for AND and OR, but this is not so, and `${MIOPEN_USE_MLIR} AND ...` does not break evaluation when MIOPEN_USE_MLIR is Off.

atamazov: Resolved.

atamazov:
@junliume @jerryyin PR is ready for next round of review & CI testing.

jerryyin (approved):

junliume (approved):

PR #1928, opened 2023-01-12
[TensorDescriptor] Add input parameters validation

averinevg:
- Validation of input parameters has been added for all constructors
- Constructors with pointers have been replaced with constructors with vectors
- TensorDescriptor code has been refactored


junliume:
@averinevg we have a few find2.0 tests failed in CI, @DrizztDoUrden could you also take a look and work with @averinevg ? Thanks!

averinevg:
@junliume conv2d also fails. CI seems to be ok, there is no fails with other PRs. I am working on it.

junliume:
@averinevg is there any issue or tickets associated with this change? I may consider the urgency level of this PR if not required by the next release. Thanks!

averinevg:
> @averinevg is there any issue or tickets associated with this change? I may consider the urgency level of this PR if not required by the next release. Thanks!

@junliume I decided to add this while working on https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1929. I think it's worth adding to the next release.

averinevg:
@atamazov Please review

cmake/EnableCompilerWarnings.cmake
```diff
@@ -67,7 +67,6 @@ else()
             -Wunused
             -Wno-ignored-qualifiers
             -Wno-sign-compare
-            -Wno-deprecated
```
junliume: @averinevg why this change? I recall we added it due to Ubuntu 22.04 and Clang-16. Has the issue been fixed?

src/include/miopen/tensor.hpp
```diff
@@ -124,45 +124,56 @@ std::ptrdiff_t integer_division_ceil(X x, Y y)
 struct TensorDescriptor : miopenTensorDescriptor
 {
     TensorDescriptor();
-    TensorDescriptor(miopenDataType_t t, std::initializer_list<std::size_t> plens);
+
+    // This constructor is only used in test/tensor_holder.hpp
+    // clang-format off
+    [[deprecated("Use constructor with lengths instead")]]
```
JehandadKhan, line 130: [anchor]

cmake/EnableCompilerWarnings.cmake
averinevg: @junliume I need this to be able to mark deprecated functions. I've reverted it back here and created https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1942. Thanks!

src/include/miopen/tensor.hpp
averinevg: @JehandadKhan Could you please clarify if this is good or bad? :smile:

junliume (approved):

src/include/miopen/tensor.hpp
JehandadKhan: I think this is good, I put the anchor to share it with Jun

atamazov:
@averinevg [Important] Does it fix any issue in the library or in the tests (or only adds a check)? I suspect the former, but please confirm.

src/include/miopen/tensor.hpp
atamazov: @averinevg This is good practice, I usually add `:anchor:` which is rendered as :anchor:

atamazov:
@averinevg As per our today's talk, let's try to MIOPEN_THROW when BUILD_DEV && the default-initialized TensorDescriptor is being read (and see what happens). Special case is a partially initialized TensorDescriptor that has only datatype initialized, a dirty trick.

src/rnn.cpp
```diff
@@ -845,8 +845,8 @@ void RNNDescriptor::GetLayerBiasOffset(const int layer,
         return;
     }
 
-    auto bdim = int(hsize);
-    biasDesc  = miopen::TensorDescriptor(dataType, &bdim, 1);
+    auto bdim = hsize;
+    biasDesc  = miopen::TensorDescriptor(dataType, {bdim});
```
atamazov, line 849: Please make sure that the expected ctor is called in this case. Candidates:
```C++
    TensorDescriptor(miopenDataType_t t, const std::initializer_list<int>& lens_in);
    TensorDescriptor(miopenDataType_t t, const std::vector<int>& lens_in);
    TensorDescriptor(miopenDataType_t t, const std::initializer_list<std::size_t>& lens_in);
    TensorDescriptor(miopenDataType_t t, const std::vector<std::size_t>& lens_in);
```
According to https://stackoverflow.com/questions/68328782/constructor-taking-stdinitializer-list-is-preferred-over-other-constructors/68329048, the 3rd one should be used. But please double-check and confirm.

atamazov (commented):
overall, very good.

src/rnn.cpp
averinevg: @atamazov Since `hsize` is of type `size_t`, `bdim` is also of this type, `{bdim}` is of type `std::initializer_list<size_t>`. In this case, the 3rd constructor is called.

PR #1929, opened 2023-01-12
[SWDEV-366865] Add check for convolution API that all tensors are packed

averinevg:
Only fully packed tensors are currently supported for convolutions. This check is needed to make sure that all tensors are packed and throw an error otherwise.


junliume (approved):

atamazov:
@averinevg Good job!

averinevg:
@atamazov Thanks!

PR #1935, opened 2023-01-16
[SWDEV-375617] Fix 3d convolution Host API bug

aska-0096:
Fix the function that block the 3D convolution as mentioned in [MIOpenDriver driver test fail](https://ontrack-internal.amd.com/browse/SWDEV-375617). Meanwhile, a MIOpen driver test for 3D convolution added.
This ticket contain two issues actually:
1. Host side API (fixed in this PR)
2. Kernel side, GemmBwdRest.

@junliume  Please help me find the owner of GemmBwdRest solver to fix kernel side bug.
cc. @atamazov

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -650,10 +650,18 @@ if(${MIOPEN_TEST_WITH_MIOPENDRIVER})
         # Issue #1697: this is large test which has to run in serial and not enabled on gfx900/gfx906
         COMMAND MIOPEN_DRIVER_USE_GPU_REFERENCE=1 $<TARGET_FILE:MIOpenDriver> ${MIOPENDRIVER_MODE_CONV} -W 5078 -H 4903 -c 24 -n 5 -k 1 --fil_w 3 --fil_h 3 --pad_w 6 --pad_h 4 -F 1
     )
+
+    add_custom_test(test_miopendriver_conv3d_half_gfx9 SKIP_UNLESS_ALL GFX900_DISABLED GFX906_DISABLED GFX908_DISABLED GFX90A_ENABLED GFX103X_DISABLED HALF_ENABLED
+        # Regression test for:
+        #   [SWDEV-375617] Fix 3d convolution Host API bug
+        #   https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1935
+        COMMAND MIOPEN_DRIVER_USE_GPU_REFERENCE=1 $<TARGET_FILE:MIOpenDriver> convfp16 -n 2 -c 64 --in_d 128 -H 128 -W 128 -k 32 --fil_d 3 -y 3 -x 3 --pad_d 1 -p 1 -q 1 --conv_stride_d 1 -u 1 -v 1 --dilation_d 1 -l 1 -j 1 --spatial_dim 3 -m conv -g 1 -F 1 -t 1
+    )
+
     set_tests_properties(test_miopendriver_big_tensor
         PROPERTIES RUN_SERIAL On)
     set_tests_properties(test_miopendriver_big_tensor
-        PROPERTIES TIMEOUT 600) 
+        PROPERTIES TIMEOUT 600)
 endif()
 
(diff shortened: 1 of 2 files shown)


junliume (approved):

test/CMakeLists.txt
```diff
@@ -650,10 +650,18 @@ if(${MIOPEN_TEST_WITH_MIOPENDRIVER})
         # Issue #1697: this is large test which has to run in serial and not enabled on gfx900/gfx906
         COMMAND MIOPEN_DRIVER_USE_GPU_REFERENCE=1 $<TARGET_FILE:MIOpenDriver> ${MIOPENDRIVER_MODE_CONV} -W 5078 -H 4903 -c 24 -n 5 -k 1 --fil_w 3 --fil_h 3 --pad_w 6 --pad_h 4 -F 1
     )
+
+    add_custom_test(test_miopendriver_conv3d_half_gfx9 SKIP_UNLESS_ALL GFX900_DISABLED GFX906_DISABLED GFX908_DISABLED GFX90A_ENABLED GFX103X_DISABLED HALF_ENABLED
```
atamazov, line 654: [Quality] AFAICS the test is enabled only for MI200, for both `float` and `half` types. Using it with `float` datatype is not needed, Please add `FLOAT_DISABLED`.

[Quality] `GFX103X_DISABLED` is not needed because testing for this target is disabled by defaut. Please remove this attribute.

[Quality] Please rename the test to `test_miopendriver_half_mi200`. This would indicate that the test will be executed only for HALF datatype and only on MI200, and suggests the developer to add here any other tests that require MIOpenDriver and should be run only for FP16 on MI200.

Note that `conv3d` in the test name is superfluous (because other, non-conv3d related tests could be added here), `gfx9` is ambiguous.

test/CMakeLists.txt
```diff
@@ -650,10 +650,18 @@ if(${MIOPEN_TEST_WITH_MIOPENDRIVER})
...
+        # Regression test for:
+        #   [SWDEV-375617] Fix 3d convolution Host API bug
+        #   https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1935
+        COMMAND MIOPEN_DRIVER_USE_GPU_REFERENCE=1 $<TARGET_FILE:MIOpenDriver> convfp16 -n 2 -c 64 --in_d 128 -H 128 -W 128 -k 32 --fil_d 3 -y 3 -x 3 --pad_d 1 -p 1 -q 1 --conv_stride_d 1 -u 1 -v 1 --dilation_d 1 -l 1 -j 1 --spatial_dim 3 -m conv -g 1 -F 1 -t 1
```
atamazov, line 658: [Recommendation][Quality] Replace `convfp16` with `${MIOPENDRIVER_MODE_CONV}`. Up to you.

test/CMakeLists.txt
```diff
@@ -650,10 +650,18 @@ if(${MIOPEN_TEST_WITH_MIOPENDRIVER})
         # Issue #1697: this is large test which has to run in serial and not enabled on gfx900/gfx906
         COMMAND MIOPEN_DRIVER_USE_GPU_REFERENCE=1 $<TARGET_FILE:MIOpenDriver> ${MIOPENDRIVER_MODE_CONV} -W 5078 -H 4903 -c 24 -n 5 -k 1 --fil_w 3 --fil_h 3 --pad_w 6 --pad_h 4 -F 1
     )
+
```
atamazov, line 653: [Recommendation][Quality] It would be better to add this testcase to the existing `test_regression_half_mi200`.
- Pros:
  - We do not like to use MIOpenDriver in tests and do it only when there is no other way.
- Cons:
  - You need to convert MIOpenDriver parameters to test_conv3d parameters, and check that test_conv3d fails without #1935 and passes with it. Which means some additional work.

Up to you.

atamazov (commented):
https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_low I only found this while merging `develop` into my local branch.

atamazov:
@aska-0096 
> Please help me find the owner of GemmBwdRest solver to fix kernel side bug.

IIRC the original GEMM code in the library was written mostly by @ce1adon. Then the code was heavily refactored by @DrizztDoUrden.

However, if the issue is at "kernel side", then MIOpen has nothing to do with it, because GEMM algorithm leverages kernels from rocBLAS.

aska-0096:
Hi @atamazov , Thanks for the code quality advice, I will create a PR this week.

PR #1933, opened 2023-01-13
[HOTFIX][MI200][FP16] Disabled ConvHipImplicitGemmBwdXdlops when FP16_ALT is requried

atamazov:
Resolves the most important problem from ticket #1932.

[Attribution] @junliume @johnny-keker https://github.com/ROCmSoftwarePlatform/MIOpen/labels/correctness https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_blocker https://github.com/ROCmSoftwarePlatform/MIOpen/labels/bug

Suggested reviewers: @iq136boy @JehandadKhan

PR #1936, opened 2023-01-16
Enable ConvHipImplicitGemmFwdXdlops for MI200 and hotfix for fp16_alt issue

iq136boy:
Enable ConvHipImplicitGemmFwdXdlops for MI200.
We expected the similar issue might raise with datatype fp16_alt, therefore we applied the same hotfix as in #1933. The support of fp16_alt will be addressed in a separate PR after discussion with CK group.

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -339,7 +339,10 @@ bool ConvHipImplicitGemmFwdXdlops::IsApplicable(const ConvolutionContext& ctx,
         return false;
     if(!problem.Is2d())
         return false;
-    if(ctx.GetStream().GetDeviceName() != "gfx908")
+    const std::string& arch = ctx.GetStream().GetDeviceName();
+    if(!(arch == "gfx908" || arch == "gfx90a"))
+        return false;
+    if(arch == "gfx90a" && problem.conv_problem.IsGfx90aFp16altRequired())
         return false;
     if(!problem.IsLayoutNHWC())
         return false;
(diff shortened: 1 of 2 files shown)


junliume:
adding @atamazov for review.

atamazov (changes requested):
Almost LGTM provided ConvHipImplicitGemmFwdXdlops has been extensively tested on the MI200 machine, but please enable `test_conv_hip_igemm_xdlops` for MI200.

atamazov:
@iq136boy 
> The support of fp16_alt will be addressed in a separate PR after discussion with CK group.

Good! Just to remind that FP16_ALT requires support for selection of "FP16 mode" during kernel invocation. The necessary parameter is passed to the Invoker and it is the Invoker who should properly handle it. How it should be handled is a design choice.

For example, the programmer may add ALT precision mode to some existing kernel and pass the "ALT" parameter to it. Another method is creating a new kernel that implements the ALT mode and run it instead of the "normal" kernel. In the latter case Invoker looks at the "ALT" parameter and runs appropriate kernel; what is important here is that both "normal" and "ALT kernels must be included into a Solution.

iq136boy:
> Almost LGTM provided ConvHipImplicitGemmFwdXdlops has been extensively tested on the MI200 machine, but please enable `test_conv_hip_igemm_xdlops` for MI200.

The tests are enabled for MI200.

iq136boy:
> @iq136boy
> 
> > The support of fp16_alt will be addressed in a separate PR after discussion with CK group.
> 
> Good! Just to remind that FP16_ALT requires support for selection of "FP16 mode" during kernel invocation. The necessary parameter is passed to the Invoker and it is the Invoker who should properly handle it. How it should be handled is a design choice.
> 
> For example, the programmer may add ALT precision mode to some existing kernel and pass the "ALT" parameter to it. Another method is creating a new kernel that implements the ALT mode and run it instead of the "normal" kernel. In the latter case Invoker looks at the "ALT" parameter and runs appropriate kernel; what is important here is that both "normal" and "ALT kernels must be included into a Solution.

@atamazov Thanks so much for the reminder and explanation.

atamazov (approved):
LGTM!

junliume (approved):
watching the last round of CI :)

PR #1940, opened 2023-01-21
[NFC] Add missing header guards.

xinlipn:
Added an extra line after #endif at the end of each file. But it seems format.sh removes it.


iq136boy (commented):
We do need to add extra line at very end of each file to pass the CI format check. So adding this extra line is for same reason?

xinlipn:
> We do need to add extra line at very end of each file to pass the CI format check. So adding this extra line is for same reason?

@iq136boy ,yes, the extra line was added to pass the format check

bghimireamd:
LGTM.  This works and `#pragma once` would also work.

alexandraBara (commented):
If i remember correctly, the copyright year will only need to be updated if the file was written in that year (not modified), otherwise it should be left as is. @junliume  can confirm this

iq136boy (dismissed):
LGTM

junliume:
> If i remember correctly, the copyright year will only need to be updated if the file was written in that year (not modified), otherwise it should be left as is. @junliume can confirm this

@alexandraBara @xinlipn @JehandadKhan
```
Standard notice for new file:
Copyright(C) 2023 Advanced Micro Devices, Inc. All rights reserved.
```
```
Standard notice for modified AMD files:
Copyright (C) [year development began – year of most recent substantive update] Advanced Micro Devices, Inc. All rights Reserved.
```

xinlipn:
> If i remember correctly, the copyright year will only need to be updated if the file was written in that year (not modified), otherwise it should be left as is. @junliume can confirm this

@alexandraBara , thanks for pick this up. Confirmed with Jun and these indeed don't need to be changed. I have reverted these changes.

alexandraBara (dismissed):

JehandadKhan (changes requested):
@xinlipn a more modern method would be to just add 

``` 
# pragma once
```

to the top of the file.

xinlipn:
> @xinlipn a more modern method would be to just add
> 
> ```
> # pragma once
> ```
> 
> to the top of the file.

@JehandadKhan thanks for the comments. There's a discussion about the difference between #pragma once vs include guards as below. Not sure if this would be a concern, but #pragma once may affect portability since not all compilers support it. I'll update with #pragma once shortly

https://stackoverflow.com/questions/1143936/pragma-once-vs-include-guards

junliume:
@xinlipn @JehandadKhan Yes we are seeing the side effects now:
```
 /tmp/comgr-f62e8f/input/MIOpenDropout.cl:7575:9: error: #pragma once in main file [-Werror,-Wpragma-once-outside-header]
```
Maybe the legacy way is also safer, let's revert to the old way.

JehandadKhan:
> @xinlipn @JehandadKhan Yes we are seeing the side effects now:
> 
> ```
>  /tmp/comgr-f62e8f/input/MIOpenDropout.cl:7575:9: error: #pragma once in main file [-Werror,-Wpragma-once-outside-header]
> ```
> 
> Maybe the legacy way is also safer, let's revert to the old way.

Looks like it, on the host side we have not seen this issue with gcc or clang.

xinlipn:
> 



> 

@JehandadKhan @junliume reverted  #pragma once to include guards

JehandadKhan (approved):

iq136boy (approved):

junliume (approved):

PR #1942, opened 2023-01-24
[NFC] Remove -Wno-deprecated warning suppression

averinevg:
- Redundant `static constexpr` declarations have been removed. If a `static` data member is declared `constexpr`, it is implicitly `inline` and does not need to be redeclared at namespace scope (since C++17)
- `clCreateCommandQueue()` has been replaced with `clCreateCommandQueueWithProperties()` ( `clCreateCommandQueue()` is deprecated by OpenCL 2.0)


junliume:
@averinevg could we also add tests to #1928 and #1929 to this PR?

averinevg:
> @averinevg could we also add tests to #1928 and #1929 to this PR?

@junliume I'm working on test for https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1928, it is almost ready. Sure, we can, but I think it is better to add them in separated PRs

atamazov (approved):
LGTM

junliume (approved):
Always a great idea to remove global warning suppression :)

PR #1943, opened 2023-01-25
[NFC] Comgr include header reorg

junliume:
some residue which causes warning messages like
```
In file included from /root/MIOpen/src/comgr.cpp:41:
/opt/rocm-5.5.0-11454/include/amd_comgr.h:27:2: warning: "This file is deprecated. Use file from include path /opt/rocm-ver/include/ and prefix with amd_comgr" [-W#warnings]
#warning "This file is deprecated. Use file from include path /opt/rocm-ver/include/ and prefix with amd_comgr"
 ^
1 warning generated when compiling for gfx1030.
```
This PR resolve issue mentioned in #1945

### DIFF
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -38,7 +38,7 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/stringutils.hpp>
 
-#include <amd_comgr.h>
+#include <amd_comgr/amd_comgr.h>
 #include <hip/hip_runtime_api.h>
 #if MIOPEN_USE_HIPRTC
 #include <miopen/manage_ptr.hpp>
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -38,7 +38,11 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/stringutils.hpp>
 
+#if HIP_PACKAGE_VERSION_FLAT >= 4004000000ULL
 #include <amd_comgr/amd_comgr.h>
+#else
+#include <amd_comgr.h>
+#endif
 #include <hip/hip_runtime_api.h>
 #if MIOPEN_USE_HIPRTC
 #include <miopen/manage_ptr.hpp>


atamazov (changes requested):
```
/home/atamazov/github/MLOpen1/src/comgr.cpp:41:10: fatal error: 'amd_comgr/amd_comgr.h' file not found
#include <amd_comgr/amd_comgr.h>
         ^~~~~~~~~~~~~~~~~~~~~~~
```

- ROCm 5.2.3 - build failed
- ROCm 5.3.0 - Ok
- ROCm 5.4.1 - Ok

⚠️ COMgr version everywhere is 2.4.0. @junliume The change of the location of the public headers is a public API change. *Please inform COMgr developers that they MUST bump version when API of their library is changed.** As a workaround, we can use ROCm version to select correct path.

I'll provide W/A recommendation soon

src/comgr.cpp
```diff
@@ -38,7 +38,7 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/stringutils.hpp>
 
-#include <amd_comgr.h>
+#include <amd_comgr/amd_comgr.h>
```
atamazov, line 42:
```suggestion
#if HIP_PACKAGE_VERSION_FLAT >= 5004000000ULL
#include <amd_comgr/amd_comgr.h>
#else
#include <amd_comgr.h>
#endif
```

atamazov (changes requested):

src/comgr.cpp
atamazov: The new location will be used in releases starting from 5.4.0, and in mainline starting from 5.5.

atamazov (approved):
LGTM

atamazov:
@junliume I made a typo in my original suggestion (4004000000ULL instead of 5004000000ULL) and fixed it after a couple of hours or so, but you were too fast and incorrect number was merged. My bad.

atamazov:
@junliume please expect a PR with fix soon.

Issue #1956, opened 2023-01-30
MIOpendriver Config Failure: GemmBwdRest MIOpenCol2Im3d.cl


junliume:
@atamazov  this config has been failing even after #1935 

```
./bin/MIOpenDriver convfp16 -n 2 -c 64 --in_d 128 -H 128 -W 128 -k 32 --fil_d 3 -y 3 -x 3 --pad_d 1 -p 1 -q 1 --conv_stride_d 1 -u 1 -v 1 --dilation_d 1 -l 1 -j 1 --spatial_dim 3 -m conv -g 1 -F 2 -t 1
```

Kernel side, GemmBwdRest. and the following file might need to be changed to adapt the sizes:

```
src/kernels/MIOpenCol2Im3d.cl

junliume:
Ping @atamazov  and @zjing14

atamazov:
@junliume I see. Maybe this is blocker issue?

atamazov:
- gfx906/60
  - miopen-image_rocm-5.4.1 
  - ROCm 5.3.0              
  - miopen-image_rocm-5.2.3

The problematic config fails with `Memory access fault by GPU node-1` everywhere. Stay tuned.

atamazov:
[Info] Workspace size is 6.75 GiB which is beyond max addressable limit for 32-bit address math.

junliume:
> [Info] Workspace size is 6.75 GiB which is beyond max addressable limit for 32-bit address math.

I tried to modify a few "int" in this file to "long" but it did not work :( The file has been robust for a long time.
src/kernels/MIOpenCol2Im3d.cl

atamazov:
Yep, I am looking there.

atamazov:
@junliume Now we have W/A in #1965 and I have time to look into look into the kernel and develop a full-blown fix.

atamazov:
Now #1966 fixes the most important things, but we still need to resolve https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1966#discussion_r1108216563. Let's close the ticket when this is done.

PR #1966, opened 2023-02-06
[Fix issue #1956] use 64bit index in im2col3d.cl

zjing14:
Full-blown fix for issue #1956.

### DIFF
--- a/src/kernels/MIOpenCol2Im3d.cl
+++ b/src/kernels/MIOpenCol2Im3d.cl
@@ -72,17 +72,17 @@ __kernel void Col2Im3d(global _FLOAT* col,
     int start_d = (im_d < dilation_d * (wei_d - 1) + 1)
                       ? 0
                       : (im_d - (dilation_d * (wei_d - 1) + 1)) / stride_d + 1;
-    int end_d = min(col_d, im_d / stride_d + 1);
+    int end_d   = min(col_d, im_d / stride_d + 1);
 
     int start_h = (im_h < dilation_h * (wei_h - 1) + 1)
                       ? 0
                       : (im_h - (dilation_h * (wei_h - 1) + 1)) / stride_h + 1;
-    int end_h = min(col_h, im_h / stride_h + 1);
+    int end_h   = min(col_h, im_h / stride_h + 1);
 
     int start_w = (im_w < dilation_w * (wei_w - 1) + 1)
                       ? 0
                       : (im_w - (dilation_w * (wei_w - 1) + 1)) / stride_w + 1;
-    int end_w = min(col_w, im_w / stride_w + 1);
+    int end_w   = min(col_w, im_w / stride_w + 1);
 
 #if MIOPEN_USE_64BIT_INDEX
     long ch_offset = (long)im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h;
--- a/src/kernels/MIOpenCol2Im3d.cl
+++ b/src/kernels/MIOpenCol2Im3d.cl
@@ -72,19 +72,24 @@ __kernel void Col2Im3d(global _FLOAT* col,
     int start_d = (im_d < dilation_d * (wei_d - 1) + 1)
                       ? 0
                       : (im_d - (dilation_d * (wei_d - 1) + 1)) / stride_d + 1;
-    int end_d   = min(col_d, im_d / stride_d + 1);
+    int end_d = min(col_d, im_d / stride_d + 1);
 
     int start_h = (im_h < dilation_h * (wei_h - 1) + 1)
                       ? 0
                       : (im_h - (dilation_h * (wei_h - 1) + 1)) / stride_h + 1;
-    int end_h   = min(col_h, im_h / stride_h + 1);
+    int end_h = min(col_h, im_h / stride_h + 1);
 
     int start_w = (im_w < dilation_w * (wei_w - 1) + 1)
                       ? 0
                       : (im_w - (dilation_w * (wei_w - 1) + 1)) / stride_w + 1;
-    int end_w   = min(col_w, im_w / stride_w + 1);
+    int end_w = min(col_w, im_w / stride_w + 1);
 
+#if MIOPEN_USE_64BIT_INDEX
     long ch_offset = (long)im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h;
+#else
+    int ch_offset = (long)im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h;
+#endif
+
     col += ch_offset;
 
     _FLOAT_ACCUM tmp = (_FLOAT_ACCUM)0;
@@ -103,10 +108,15 @@ __kernel void Col2Im3d(global _FLOAT* col,
                     int y = (im_h - cy * stride_h) / dilation_h;
                     int x = (im_w - cx * stride_w) / dilation_w;
 
+#if MIOPEN_USE_64BIT_INDEX
                     long col_off =
                         ((((((long)z * wei_h) + y) * wei_w + x) * col_d + cz) * col_h + cy) *
                             col_w +
                         cx;
+#else
+                    int col_off =
+                        (((((z * wei_h) + y) * wei_w + x) * col_d + cz) * col_h + cy) * col_w + cx;
+#endif
 
                     tmp += CVT_FLOAT2ACCUM(col[col_off]);
                 }
--- a/src/kernels/MIOpenCol2Im3d.cl
+++ b/src/kernels/MIOpenCol2Im3d.cl
@@ -53,7 +53,7 @@ __kernel void Col2Im3d(global _FLOAT* col,
                        const int height,
                        const int width,
                        global _FLOAT* im,
-                       const int im_offset)
+                       const unsigned long im_offset)
 {
     global _FLOAT* im_off = im + im_offset;
     int gid               = (int)get_global_id(0);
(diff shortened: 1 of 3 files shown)


atamazov (commented):
@zjing14 I suspect this affects performance. Can we introduce a macro, like `MIOPEN_USE_64BIT_INDEX`, define it as 1 when workspace size > 4 GiB or as 0 otherwise, and use it in the kernel? If you agree but do not have time, then I can do it for you.

Perhaps we shall use 2 GiB threshold because im_offset and ch_offset in the original code are of type `int`?

zjing14:
@atamazov Yes, it may impact performance for small problem size. Let me fix it.

atamazov:
@zjing14 
> @atamazov Yes, it may impact performance for small problem size. Let me fix it.

[Recommendation] Use the following code in GemmBwdRest::GetSolution:
```C++
const bool use_64_bit_index =
    problem.GetConv().GetSpatialDimension() == 3
    && GetWorkspaceSize(context, problem) > 0xffffffffULL);
```
Then pass `use_64_bit_index` via `Col2ImGPU()` to both `Col2Im3dGPU()` and `Col2Im2dGPU()`. In the latter two, pass it to `GetDataTypeKernelParams()` and implement there support for generating the `MIOPEN_USE_64BIT_INDEX` macro. Then use the macro in MIOpenCol2Im3d.cl.

AFAICS the current implementation of 2D Col2Im does not use it, but that's not a problem; we can engage it in the future if need arise.

Hope this helps.

atamazov:
@zjing14 Can you please update the PR description with:

```
Full-blown fix for issue #1956.
```

Please also set the following labels on this PR: https://github.com/ROCmSoftwarePlatform/MIOpen/labels/bug https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_high

zjing14:
@atamazov Thanks very much for your suggestions. I will fix it ASAP.

zjing14:
@atamazov Added a switch that only enables 64bit for tensor larger than 4GB. Please review it.

junliume:
@zjing14 @atamazov some other (potential) issues listed here: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1974 the original issue mentioned in #1956 is indeed fixed by this PR.

atamazov:
Will review/test tomorrow

junliume (dismissed):
Ping @atamazov this issue and #1974 are for the same model, and #1974 might be different since 5D tensor is not supported in batchnorm OCL code

src/kernels/MIOpenCol2Im3d.cl
```diff
@@ -84,7 +84,12 @@ __kernel void Col2Im3d(global _FLOAT* col,
                       : (im_w - (dilation_w * (wei_w - 1) + 1)) / stride_w + 1;
     int end_w   = min(col_w, im_w / stride_w + 1);
 
-    int ch_offset = im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h;
+#if MIOPEN_USE_64BIT_INDEX
+    long ch_offset = (long)im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h;
+#else
+    int ch_offset = (long)im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h;
```
atamazov:
```suggestion
    int ch_offset = im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h;
```
This is faster and matches the original code.

BTW we can use typedef, and then avoid conditional compilation, like this:
```C++
#if MIOPEN_USE_64BIT_INDEX
typedef int offset_t;
#else
typedef long offset_t;
#endif
...
offset_t ch_offset = (offset_t)im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h;
...
offset_t col_off = ((((((offset_t)z * wei_h) + y) * wei_w + x) * col_d + cz) * col_h + cy) * col_w + cx;
```

atamazov (changes requested):

src/kernels/MIOpenCol2Im3d.cl
junliume: @zjing14 could you help to revise the PR accordingly? Thanks!

atamazov:
Confirmed that this PR resolves issue #1956.

atamazov:
@junliume Unfortunately I am unable to edit this branch directly. But I can copy it into my fork, update there, and open a new PR, if necessary due to urgency reasons.

junliume:
> @junliume Unfortunately I am unable to edit this branch directly. But I can copy it into my fork, update there, and open a new PR, if necessary due to urgency reasons.

If the suggested change above is the only place, I can commit it directly :)

src/ocl/utilocl.cpp
```diff
@@ -565,8 +565,15 @@ float Col2Im3dGPU(const Handle& handle,
     }
     else
     {
+        std::size_t index_size = static_cast<size_t>(in_c) * out_d * out_h * out_w * wei_d * wei_w *
+                                 wei_h * sizeof(ConstData_t);
+
+        const bool use_64_bit_index = index_size > 0xffffffffULL;
```
atamazov, line 571: [Performance]

1. Hmm, why `* sizeof(ConstData_t)`, which is `sizeof(void*) == 8` for HIP backend? In the shader, 
```
ch_offset = im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h
```
where max value of im_ch is in_c, right?

Multiplying this to 8 looks like an overkill.

2. Another question is about `0xffffffffULL` (4 GiB - 1). In the kernel, computations are performed as `int`, which has max value of `0x7fffffffULL` (2 GiB - 1).

atamazov (changes requested):

junliume:
@zjing14 @atamazov  Changing urgency level since we are in release mode now.

src/ocl/utilocl.cpp
junliume: @zjing14 please comment. Thanks!

atamazov: @zjing14 WRT question 1. ~AFAICS `sizeof(ConstData_t)` it should be `GetTypeSize(type)`, which matches `sizeof(_FLOAT)` in the kernel, right? It should be either 4 (fp32) or 2 (fp16), I think.~

UPDATE: Please ignore point 1. We do not need to take the size of element into account because `col_off` is actually not an offset but an index in the array `col`. Index is multiplied by the `sizeof(col[0])` automatically.

Question 2 remains the same.

atamazov (approved):
🟢 This can be merged as is, provided that https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1966#discussion_r1108216563 will be fixed later. The impact of unresolved comment is small: it affects performance for the GemmBwdRest in some relatively narrow range or FP16 problem configs (and also affects clarity/quality of the code).

junliume (approved):

atamazov:
@zjing14 What do you think about https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1966#pullrequestreview-1301204519? You can delegate the actual work to somebody else (even to me), if you do not have time, but we need your opinion prior starting the actual work.

junliume:
> @zjing14 What do you think about [#1966 (review)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1966#pullrequestreview-1301204519)? You can delegate the actual work to somebody else (even to me), if you do not have time, but we need your opinion prior starting the actual work.

Ping @zjing14 ...

@atamazov I briefly discussed with Jing on this issue and the main concern is about performance impacts.

zjing14:
@atamazov Sorry for the late reply. Yes, that will be great if someone else can take care of it. If any help is needed, just ping me.

atamazov:
Okay, let's track this in #1985

src/ocl/utilocl.cpp
zjing14: > [Performance]
> 
> 1. Hmm, why `* sizeof(ConstData_t)`, which is `sizeof(void*) == 8` for HIP backend? In the shader,
> 
> ```
> ch_offset = im_ch * col_d * col_w * col_h * wei_d * wei_w * wei_h
> ```
> 
> where max value of im_ch is in_c, right?
> 
> Multiplying this to 8 looks like an overkill.

Yes, it should be  `sizeof(ConstData_t)`.

> 2. Another question is about `0xffffffffULL` (4 GiB - 1). In the kernel, computations are performed as `int`, which has max value of `0x7fffffffULL` (2 GiB - 1).

Yes, `0x7fffffffULL` (2 GiB - 1) makes sense or we change it to `unsigned int`.

atamazov: @zjing14 Thanks for answering!
> Yes, it should be sizeof(ConstData_t).

Excuse me, but I do not understand why. AFAIU the goal is to prevent overflows in the following computations in the shader:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/7ad167d2ecdc29ccebdf965055e1b1a8b9a835a8/src/kernels/MIOpenCol2Im3d.cl#L87-L91

and
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/7ad167d2ecdc29ccebdf965055e1b1a8b9a835a8/src/kernels/MIOpenCol2Im3d.cl#L111-L119
For computation of `ch_offset` we do not need the `* sizeof(ConstData_t)` multiplier at all.

Is this multiplier necessary to prevent overflow during computation of `col_off`?

atamazov:
@zjing14 There is something that is not clear to me, pls look at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1966#discussion_r1160911478. Thanks!

src/ocl/utilocl.cpp
zjing14: Yes. You are right, the goal of `int64_t` is to prevent index overflow. We need to make sure all indices `ch_offset `and `col_off ` do not overflow.

zjing14: We may change indices in the kernel to `unsigned int` for 32bit. It may improve performance for some cases where indices > 16bit but < 32bit.

atamazov: @zjing14 Thanks for feedback. And yes, I'll change to unsigned. But what about removal of `* sizeof(ConstData_t)` from the computation of index_size? Is it Ok to do that or I am missing something?

PR #1970, opened 2023-02-07
Rework for any_solver perf params test function. (Repost)

cderb:
Opening to continue #1885. Needed to clear workspace on Jenkins.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "fin"]
     path = fin
     url = https://github.com/ROCmSoftwarePlatform/MIFin.git
-    branch = cderb/param_check
+    branch = develop
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "fin"]
     path = fin
     url = https://github.com/ROCmSoftwarePlatform/MIFin.git
-    branch = develop
+    branch = cderb/param_check
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -313,7 +313,7 @@ def buildHipClangJobAndReboot(Map conf=[:]){
     }
 }
 
-def CheckDeserializePerfDb(Map conf=[:]){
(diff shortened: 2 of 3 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "fin"]
     path = fin
     url = https://github.com/ROCmSoftwarePlatform/MIFin.git
-    branch = develop
+    branch = cderb/param_check
```
averinevg: @cderb Why do you need this branch?

junliume: Guess it might need some cleaning up?

cderb: There is a chicken V egg problem.  The 'TestPerfCfgParams' function was renamed so the Fin PR that changes the function name can't pass, because the MIOpen function doesn't exist.
At the same time, this PR which renames the function can't pass, because Fin has the old function name.
This branch has the new function. It will be returned to develop shortly after this PR settles.

averinevg: Does this mean that every time the API changes, we have to do this trick? If yes, then you need to do something with mutual dependencies.

It might be worth adding the new function to the `MIOpen` first, while keeping the old one, then enabling it in the `Fin`, and then deleting the old function in the `MIOpen`.

cderb: It would happen anytime a function from the API was renamed, yes.

I'd prefer to remove the old function in this PR as opposed leaving bloat. Alternatively I can also disable the fin build phase until this is resolved.

averinevg: @cderb I'm asking because I'm going to do the same thing with `AnySolver`. Its API will be changed, but I plan to keep methods with old signatures before they are implemented in the `Fin`.

cderb: Resolved this on the side of Fin.

averinevg (approved):

JehandadKhan (approved):

junliume (approved):

PR #1984, opened 2023-02-15
[ROCm 5.5] Remove MLIR entries from Find db for gfx90a/gfx908

cderb:
Perf db entries are still present.
Timing information is still being tracked by Tuna.


atamazov:
@cderb Why this is necessary?

junliume:
> @cderb Why this is necessary?

This is to fix some perf regression issues.

atamazov:
@cderb Can we sort fin-db files prior committing? That would make diffs smaller and better readable.

atamazov:
Changes looks good.

cderb:
> @cderb Can we sort fin-db files prior committing? That would make diffs smaller and better readable.

@atamazov The find dbs are sorted by kernel time and algorithm. (algorithm only if the same kernel time)

junliume (approved):

atamazov:
@cderb 
> The find dbs are sorted by kernel time and algorithm. (algorithm only if the same kernel time)

It is good to sort data within individual records (lines) like you've described. But it doesn't help to make the diff smaller. In order to see something meaningful in the diff, I had to: 
```
sort gfx90878.HIP.fdb.txt > gfx90878.HIP.fdb.txt.sorted 
```
and then apply diff to the .sorted files.

cderb:
@atamazov That is peculiar. As Tuna exports the fdb it executes a string sort using the fdb key to order the rows. They should be sorted.
I'm want to use diffs to compare these as well. So I'm also trying to impose an ordering.

atamazov:
@cderb Oh, I see. It is correctly sorted (i.e. as if sorted by `sort`, except for entries whose key ends with `_gNNN` which is not very important). But it seems that the _set of configs_ is very different (try diff and see), and I was confused by this.

Is it expected that the set of configs is very different?

cderb:
@atamazov I think there may be some extra configs that sit in Tuna's MySQL db over what there are in MIOpen. Usually I would not expect a big expansion of the config space. I think you'll notice that all of the configs added only have MLIR entries.

atamazov:
This may lead to performance regressions even if there is other fast iGemm solver exists in the library. This potential regression is fixed by resolving issue #1987.

Issue #1987, opened 2023-02-17
Updating fdb to have multiple solvers


cderb:
### Objective: 
We would like to be able to store multiple solvers from the same algorithm in a single find database entry. This is so that if we need to disable a solver, and the next best solver is in the same algorithm grouping, then we would still find it.

### Problem:
Currently our DbRecords will print with 
```cpp
pair.first + ':' + pair.second
```
Which is the key followed by a comma separated list of values.

MIOpen is now keying find db entries by algorithm, 
So printing a find db record looks like this:
```
algorithm:sover,kernel_time,workspace_size,algorithm,network_config
```

In order to store multiple solvers from the same algorithm, find db would need to be re-keyed by the solver name.
The new print would then look like:
```
solver:sover,kernel_time,workspace_size,algorithm,network_config
```

### Solution:
The fdbs need to be updated to the new format. This can be done without much difficulty.

Alternatively we could change the format to:
```
solver:kernel_time,workspace_size,algorithm,network_config
```
Which would be more clean, but would require updating 
a) FindDbData::Visit so that solver_id is not printed
b) the print function DbRecord::WriteIdsAndValues to exclude the first value in pair.second

Another format would be:
```
sover,kernel_time,workspace_size,algorithm,network_config
```
Which would be done easily by removing (pair.first + ':' +) from the output string in DbRecord::WriteIdsAndValues.

JehandadKhan:
@atamazov  As discussed, there is the proposal that @cderb  is working on implementing. Please share your thoughts to help with the PR review process later.

atamazov:
First, a few general remarks. The FDB functionality is part of the lib's core that is engaged in all the Find modes and in Immediate mode. Therefore, changing the structure of FDB might have nontrivial side effects. So first of all, we need to evaluate possible side effects wrt this:
- All places where the FDB is read or written should be reviewed.
  - Special care for MT/MP safety
    - Example of possible problems: #219
- There are some implicit dependencies in our API that must be taken into account
- All tests that count on certain FDB structure must be revised
- Let's also consider possible directions of future development of the lib.

Implementation should be only the second step.

[Recommendation] This is best done by the same people who designed and implemented the existing FDB and other parts of the lib's core, i.e. me, @DrizztDoUrden, @averinevg.

atamazov:
This ticket can be considered as a sub-task of #374, which, among other changes, assumes changing the FDB structure in the same way as proposed here.

[...]

PR #1988, opened 2023-02-17
Solver keyed FDB

cderb:
Implementation of #1987 
FDB format
```
solver:sover,kernel_time,workspace_size,algorithm,network_config
```


cderb:
Set as draft until find dbs are formatted.

DrizztDoUrden:
I don't see any immediate issues if this does not make immediate mode inoperable. That has to be tested, but looks like it would either work or not, so should not be complicated.

atamazov:
I am working on this.

atamazov:
@cderb Now this can be closed.

atamazov:
Let's close this.

PR #1997, opened 2023-02-22
Limit Tuning by Time 

JehandadKhan:
This PR makes updates to the search algorithm in MIOpen:

* The search sequence is randomized. This helps in increasing the chances of finding an optima even when a time budget is set. 
* An environment variable (`MIOPEN_TUNING_TIME_MS_MAX`) which sets the number of milliseconds MIOpen can spend to tune each solver.
* Updates to selection of multi-threading level based on the compilation back-end [in light of collected data](https://amdcloud-my.sharepoint.com/:x:/g/personal/jkhan_amd_com/EfxrlXYTabhNujw3zGJuieoBmGcEpjnlOHFMkYofl0-5vA?e=XCYyR2).
* Add a multi threaded queue which enables multiple threads to write to a queue and the main thread to consume work from it. This struct is accompanied with a unit test to go along with it. 
* Add an MIOpen env variable to override tuning parameters for solvers. (`MIOPEN_DEBUG_PERFDB_OVERRIDE`)


JehandadKhan:
@atamazov As discussed, here is the PR which limits tuning by time.

src/generic_search.cpp
```diff
@@ -29,16 +29,38 @@
...
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_COMPILE_PARALLEL_LEVEL)
 
 std::size_t GetTuningIterationsMax()
 {
     return Value(MIOPEN_DEBUG_TUNING_ITERATIONS_MAX{}, std::numeric_limits<std::size_t>::max());
 }
 
+std::chrono::milliseconds GetTuningTimeMax()
+{
+    const auto fallback =
```
DrizztDoUrden: Shouldn't this be static as well? It's not used in non-static context. Or you could wrap the calculation of `res` value in a lambda.

src/include/miopen/mt_queue.hpp
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        cond_var.notify_one();
+    }
+
+    T& front()
+    {
+        std::unique_lock<std::mutex> lock(mutex);
+        cond_var.wait(lock, [&] { return !queue.empty(); });
+        return queue.front();
```
DrizztDoUrden: Can't reference get messed up on push? This should probably return by value.

DrizztDoUrden (changes requested):

averinevg:
> * The search sequence is randomized. This helps in increasing the chances of finding an optima even when a time budget is set.

@JehandadKhan How could randomization increase the chances of finding an optima?

JehandadKhan:
> @JehandadKhan How could randomization increase the chances of finding an optima?

@averinevg When there is a limited time budget, then randomizing the search space is required. If the search space is traversed in order ( as is currently the case) then a time budget would limit the parts of space searched.

averinevg:
> @averinevg When there is a limited time budget, then randomizing the search space is required. If the search space is traversed in order ( as is currently the case) then a time budget would limit the parts of space searched.

@JehandadKhan ~~Let's imagine some list of numbers in an unknown order. Will the numbers with the highest values end up at the top of this list after randomization?~~

~~Randomization would make it possible to search among some elements from a limited part of the space, but this would not increase the probability of finding the optimal solution. I mean there is no correlation here.~~

Discussed with @atamazov. At this stage I have no objection about randomization.

src/include/miopen/generic_search.hpp
```diff
@@ -298,6 +300,50 @@ auto GenericSearch(const Solver s,
...
+    const auto time_budget        = GetTuningTimeMax();
+    auto context                  = context_; // Not sure if context is thread safe
+    context.is_for_generic_search = true;
+    const auto& profile_h         = context.GetStream();
+    // start the counter
+    for(auto idx = thread_index; idx < data_size; idx += total_threads)
+    {
+        // Check if we are out of time
+        const auto current_time = std::chrono::time_point_cast<std::chrono::milliseconds>(
+            std::chrono::system_clock::now());
```
averinevg: `std::chrono::system_clock` may not be monotonic, it is better to use `std::chrono::steady_clock`. Also MIOpen has `steady_clock` based timer implementation in `timer.hpp`

averinevg (changes requested):

src/generic_search.cpp
```diff
@@ -29,16 +29,38 @@
...
 #include <cstddef>
 #include <limits>
+#include <chrono>
 
 namespace miopen {
 namespace solver {
 
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_TUNING_ITERATIONS_MAX)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TUNING_TIME_MS_MAX)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_COMPILE_PARALLEL_LEVEL)
```
averinevg: It is not clear from the name of the variable what needs to be passed to it

src/generic_search.cpp
```diff
@@ -29,16 +29,38 @@
...
+        std::chrono::milliseconds{Value(MIOPEN_TUNING_TIME_MS_MAX{}, fallback.count())};
+    return res;
+}
+
+std::size_t GetTuningThreadsMax()
+{
+#if MIOPEN_USE_COMGR
+    const auto def_max = 1; // COMGR is not parallelizable
+#else
+    const auto def_max = 20;
```
averinevg: It might be better to calculate default `def_max` value based on the number of CPU threads.

src/generic_search.cpp
```diff
@@ -29,16 +29,38 @@
...
+#endif
+    return Value(MIOPEN_COMPILE_PARALLEL_LEVEL{}, def_max);
```
averinevg, line 58: If COMGR is used, 1 should be returned.

src/include/miopen/generic_search.hpp
```diff
@@ -323,10 +369,18 @@ auto GenericSearch(const Solver s, const Context& context_, const AnyInvokeParam
...
+    // For random access
+    std::vector<PerformanceConfig> all_configs;
+    std::copy(tmp_all_configs.begin(), tmp_all_configs.end(), std::back_inserter(all_configs));
+    // shuffle the configs
+    std::random_device rd{};
+    auto rng = std::default_random_engine{rd()};
+    std::shuffle(all_configs.begin(), all_configs.end(), rng);
     const std::size_t n_runs_total =
         std::min(static_cast<std::size_t>(std::distance(all_configs.begin(), all_configs.end())),
                  GetTuningIterationsMax());
```
averinevg:
```suggestion
    const std::size_t n_runs_total =
        std::min(all_configs.size(),
                 GetTuningIterationsMax());
```

src/include/miopen/generic_search.hpp
```diff
@@ -335,45 +389,54 @@ auto GenericSearch(const Solver s, const Context& context_, const AnyInvokeParam
     HeartBeat<PerformanceConfig> heartbeat;
     heartbeat.Start();
 
-    if(!miopen::IsCacheDisabled()) // Otherwise precompilation is useless.
+    const auto total_threads = GetTuningThreadsMax();
+    ThreadSafeQueue<std::tuple<PerformanceConfig, ConvSolution, bool>> solution_queue;
```
averinevg, line 391: It may be worth using `boost::lockfree::queue`

averinevg (changes requested):

src/generic_search.cpp
JehandadKhan: Fixed

src/include/miopen/mt_queue.hpp
JehandadKhan: Since its a queue, the push would happen at the other end of the underlying container. The object has only one consumer so peeking at the front with a reference saves the copy which a return by value would entail. Therefore, the semantics are to get a reference to the front, use the object and once you are done, you pop it off the queue.

src/include/miopen/generic_search.hpp
JehandadKhan: I would prefer not to add additional dependency on boost, it is something we are striving to get rid of.

src/generic_search.cpp
JehandadKhan: I am not sure I follow, do you mean to say that the name `MIOPEN_TUNING_TIME_MS_MAX` is not expressive enough? My thought process for choosing it is that it sets the maximum tuning time in milliseconds. Which I believe it captures.

src/generic_search.cpp
JehandadKhan: Even if COMGR is used, we still want the env variable to override the concurrency level.

src/include/miopen/generic_search.hpp
JehandadKhan: Thanks for the tip, done

src/generic_search.cpp
JehandadKhan: We default to half the hardware concurrency

src/include/miopen/generic_search.hpp
JehandadKhan: Fixed!

src/generic_search.cpp
averinevg: @JehandadKhan My comment was about `MIOPEN_COMPILE_PARALLEL_LEVEL`. In my opinion the word "level" should be replaced with the word "threads"

JehandadKhan: The `MIOPEN_COMPILE_PARALLEL_LEVEL` variable was not added by this PR. It has been around for a while. I would prefer not to change an existing an env var.

DrizztDoUrden: If this exist in some other place as well it is better to move it to some common (maybe even new) header. Otherwise it may have difficult to notice consequences like renaming only one at some point.

src/include/miopen/generic_search.hpp
```diff
@@ -298,6 +300,50 @@ auto GenericSearch(const Solver s,
...
+                  const Solver& s,
+                  const Context& context_,
+                  std::vector<PerformanceConfig>& data,
+                  ThreadSafeQueue<std::tuple<PerformanceConfig, ConvSolution, bool>>& comp_queue)
+{
+    const auto start_time =
+        std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now());
+    const auto data_size          = data.size();
+    const auto time_budget        = GetTuningTimeMax();
+    auto context                  = context_; // Not sure if context is thread safe
```
DrizztDoUrden: Note: Context should be thread-safe for every method that takes it as const. You can set is_for_generic_search in the main thread before passing it. That way no copy should be required.

src/include/miopen/generic_search.hpp
```diff
@@ -298,6 +300,50 @@ auto GenericSearch(const Solver s,
...
+    const auto& profile_h         = context.GetStream();
+    // start the counter
+    for(auto idx = thread_index; idx < data_size; idx += total_threads)
+    {
+        // Check if we are out of time
+        const auto current_time = std::chrono::time_point_cast<std::chrono::milliseconds>(
+            std::chrono::steady_clock::now());
+        if(current_time - start_time > time_budget)
+        {
+            MIOPEN_LOG_I2("Thread: " << thread_index << " Done");
```
DrizztDoUrden: I would suggest logging that it is a time-out finish and logging correct finishing separately at the end of the method.

src/include/miopen/mt_queue.hpp
DrizztDoUrden: This approach inherently relies on the implementation of std::queue not reallocating its internal container and invalidating the reference. And it probably can. Also it may get changed in the future. Thus even if safe right now this is a liability. I still suggest returning by value. Combining this with pop in a single method would remove one mutex lock.

DrizztDoUrden (changes requested):

src/include/miopen/mt_queue.hpp
DrizztDoUrden: This is what I am talking about: https://stackoverflow.com/a/16075550

JehandadKhan:
@DrizztDoUrden and @averinevg I have addressed your reviews.

src/include/miopen/mt_queue.hpp
```diff
@@ -0,0 +1,59 @@
...
+            queue.push(item);
+        }
+
+        cond_var.notify_one();
+    }
+    T pop()
+    {
+        std::unique_lock<std::mutex> lock(mutex);
+        cond_var.wait(lock, [&] { return !queue.empty(); });
+        T ret = queue.front();
```
DrizztDoUrden, line 55:
```suggestion
        T ret = std::move(queue.front());
```
Wouldn't save much time (relatively speaking) in generic search, but who knows where else this may get used.

DrizztDoUrden (dismissed):
It would be nice to add one change, but this is safe to merge right now.

averinevg (dismissed):

junliume:
@averinevg could you please resolve the conflict? It is caused by merging #2009 first.

averinevg:
> @averinevg could you please resolve the conflict? It is caused by merging #2009 first.

@junliume Merge conflict has been resolved

junliume (approved):

PR #1998, opened 2023-02-24
[SWDEV-385188] Guard HIP tests to enable building on OpenCL 

JehandadKhan:
These tests link with hip libraries such as `rocblas` which should not happen in the OpenCL backend. Therefore, we disable these tests on the OpenCL backend.

### DIFF
--- a/test/gtest/api_convbiasactiv.cpp
+++ b/test/gtest/api_convbiasactiv.cpp
@@ -23,6 +23,9 @@
  * SOFTWARE.
  *
  *******************************************************************************/
+#include <miopen/miopen.h>
+
+#if MIOPEN_BACKEND_HIP
 #include <gtest/gtest.h>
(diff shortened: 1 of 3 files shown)


junliume (approved):

bghimireamd:
do we need similar guard for other `*.cpp` files in `test/gtest` folder? Or may be we will do it in other PR eventually?

junliume:
> do we need similar guard for other `*.cpp` files in `test/gtest` folder? Or may be we will do it in other PR eventually?

Ping @JehandadKhan

JehandadKhan:
> > do we need similar guard for other `*.cpp` files in `test/gtest` folder? Or may be we will do it in other PR eventually?
> 
> Ping @JehandadKhan 

It is only necessary when we are using HIP code in the test.

bghimireamd (approved):

PR #2000, opened 2023-02-25
[NFC][Doc] Add note about local compiled kernel cache

### LINKED ISSUE #1999: Missing kernel database package for 6900XT
While working with PyTorch, I was presented with a warning that precompiled kernels aren't installed for the system's GPU architecture:
```
MIOpen(HIP): Warning [SQLiteBase] Missing system database file: gfx1030_40.kdb Performance may degrade. Please follow instructions to install: https://github.com/ROCmSoftwarePlatform/MIOpen#installing-miopen-kernels-package
```

This machine has a 6900XT, which apparently is a gfx1030 with 80 compute units.
However the package repository only contains a package for gfx1030 with 72 CUs, miopenkernels-gfx1030-36kdb, which is for the 6800XT.

Is it intended that precompiled kernels are only available for a limited set of GPUs?
If yes, I think the readme should also explain how to build and install precompiled kernel databases manually.

Or perhaps I'm using the wrong repository?
My apt source list contains this:
```
deb [arch=amd64 signed-by=/usr/share/keyrings/rocm.gpg] https://repo.radeon.com/rocm/apt/debian/ jammy main
```

I should mention that PyTorch is working fine, but the warning suggest that installing the database can improve startup times.

onitake:
This adds a note to the readme that compiled kernels will be cached, and not installing the precompiled kernel package will only make a difference on the first run.

Closes: #1999

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ The script `utils/install_precompiled_kernels.sh` provided as part of MIOpen aut
 
 The above script depends on the __rocminfo__ package to query the GPU architecture.
 
+More info can be found [here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/cache.md#installing-pre-compiled-kernels).
+
 ## Installing the dependencies
 
 The dependencies can be installed with the `install_deps.cmake`, script: `cmake -P install_deps.cmake`
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ Currently both the backends cannot be installed on the same system simultaneousl
 
 MIOpen provides an optional pre-compiled kernels package to reduce the startup latency. These precompiled kernels comprise a select set of popular input configurations and will expand in future release to contain additional coverage.
 
+Note that all compiled kernels are locally cached in the folder `$HOME/.cache/miopen/`, so the performance impact will only apply to the first execution of a new compute kernel. There is no startup time penalty on subsequent runs.
+
 To install the kernels package for your GPU architecture, use the following command:
 
 ```
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Currently both the backends cannot be installed on the same system simultaneousl
 
 MIOpen provides an optional pre-compiled kernels package to reduce the startup latency. These precompiled kernels comprise a select set of popular input configurations and will expand in future release to contain additional coverage.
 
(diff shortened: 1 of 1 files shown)


JehandadKhan (dismissed):

README.md
```diff
@@ -45,6 +45,8 @@ Currently both the backends cannot be installed on the same system simultaneousl
 
 MIOpen provides an optional pre-compiled kernels package to reduce the startup latency. These precompiled kernels comprise a select set of popular input configurations and will expand in future release to contain additional coverage.
 
+Note that all compiled kernels are locally cached in the folder `$HOME/.cache/miopen/`, so the performance impact will only apply to the first execution of a new compute kernel. There is no startup time penalty on subsequent runs.
```
atamazov:
```suggestion
Note that all compiled kernels are locally cached in the folder `$HOME/.cache/miopen/`, so precompiled kernels reduce the startup latency only for the first execution of a neural network. Precompiled kernels do not reduce startup time on subsequent runs.
```

atamazov (changes requested):

README.md
atamazov: I hope this looks clearer.

atamazov:
Actually we already have a note about this at https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/cache.md#installing-pre-compiled-kernels

atamazov (changes requested):
I recommend updating the paragraph that begins with "Not installing these packages would not impact the functioning of MIOpen..." instead of this change.

Plus, provide there a link (`More info can be found [here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/cache.md#installing-pre-compiled-kernels)`)

junliume:
> I recommend updating the paragraph that begins with "Not installing these packages would not impact the functioning of MIOpen..." instead of this change.
> 
> Plus, provide there a link (`More info can be found [here](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/doc/src/cache.md#installing-pre-compiled-kernels)`)

@onitake are you interested in taking this in this PR? We can do it separately too.

junliume (dismissed):

onitake:
Great, thanks for the review!
I'll add the link as suggested.

junliume (approved):

PR #1927, opened 2023-01-11
[All targets] Winograd v21_1_3 -> v30_2_6 update

Slimakanzer:
This PR updates Winograd kernels to the `v30_2_6` version.
Summary of the Winograd changes compared to #1255:
- Unified binaries for gfx90a and other gfx9 ASICs
- Unified binaries for grouped/non-grouped convolutions
- Added stride2/dilation2 support for Winograd F(3,2)

Summary of the Winograd solver changes:
- Extended applicability of the Winograd solver

## Local verification
- gfx90a, rocm 5.3
- gfx908, rocm 5.3
- gfx1030, rocm 5.3

## Performance improvements

Machine configuration: gfx90a, 104CUs, rocm 5.3

<details>
<summary>FP16 cases of interest for Winograd solver</summary>

W | H | c | n | k | x | y | pad_w | pad_h | stride_w | stride_h | type | dir | v30 time, ms | v30 solver | v21 time, ms | v21 solver | speedup
-- | -- | -- | -- | -- | -- | -- | --| -- | -- | -- | -- | -- | --| -- | -- | -- | --
17 | 17 | 128 | 32 | 128 | 1 | 7 | 0 | 3 | 1 | 1 | fp16 | fwd | 0.16 | 37/ConvBinWinogradRxSf3x2 | 0.19 | 37/ConvBinWinogradRxSf3x2 | 1.17
17 | 17 | 128 | 32 | 192 | 1 | 7 | 0 | 3 | 1 | 1 | fp16 | fwd | 0.23 | 37/ConvBinWinogradRxSf3x2 | 0.28 | 37/ConvBinWinogradRxSf3x2 | 1.18
17 | 17 | 160 | 32 | 160 | 1 | 7 | 0 | 3 | 1 | 1 | fp16 | fwd | 0.20 | 37/ConvBinWinogradRxSf3x2 | 0.22 | 37/ConvBinWinogradRxSf3x2 | 1.13
17 | 17 | 160 | 32 | 192 | 1 | 7 | 0 | 3 | 1 | 1 | fp16 | fwd | 0.28 | 37/ConvBinWinogradRxSf3x2 | 0.32 | 37/ConvBinWinogradRxSf3x2 | 1.14
17 | 17 | 192 | 32 | 192 | 1 | 7 | 0 | 3 | 1 | 1 | fp16 | fwd | 0.33 | 37/ConvBinWinogradRxSf3x2 | 0.37 | 37/ConvBinWinogradRxSf3x2 | 1.13
[...]

PR #2002, opened 2023-02-27
Fused Winograd solver update

Slimakanzer:
Continuation of #1927. This PR updates the Fused Winograd solver.

## Summary of code changes
- Added FP16 data type support
- Added Fused Winograd F(3,2) support
- The `weight` heuristic used in the Fusion Plan is now calculated based on the granularity loss model

## Local verification
- gfx906, rocm 5.4.2
- gfx90a, rocm 5.4.2

### DIFF
--- a/src/include/miopen/fusion/utils.hpp
+++ b/src/include/miopen/fusion/utils.hpp
@@ -81,8 +81,6 @@ inline bool WinoCommonIsApplicable(const FusionContext& params)
 
     if(!conv_ctx.problem.Is2d())
         return false;
-    if(!conv_ctx.problem.IsFp32())
-        return false;
     if(!conv_ctx.problem.IsLayoutDefault())
         return false;
     if(!conv_ctx.problem.direction.IsForward())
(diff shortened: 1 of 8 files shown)


junliume:
@Kirpich30000 @Slimakanzer are we still working on this PR? If we need it please resolve the conflicts. Thanks!

src/include/miopen/fusion/utils.hpp
```diff
@@ -81,8 +81,6 @@ inline bool WinoCommonIsApplicable(const FusionContext& params)
 
     if(!conv_ctx.problem.Is2d())
         return false;
-    if(!conv_ctx.problem.IsFp32())
```
junliume, line 84: @Slimakanzer a few lines below, do we know if XNACK settings could be enabled?
```
if(target.Xnack() && *target.Xnack())
```

atamazov: @junliume IIRC Winograd kernels work correctly only if XNACK is not enabled on target. Perhaps this is possible to create Winograd kernels that support XNACK-enabled targets.

@Kirpich30000 Is it possible, what do you think?

BradPepersAMD:

PR #2003, opened 2023-02-28
[WIP] Enable the transform tensor API to convert any NCHW tensor to NHWC tensor and vice versa

dmikushin:
LWPMIOPEN-125

We substitute the TensorTransfom implementation with a more generic GPU-based gpuTT library calls.

Few things still to be implemented:

* Support non-equal input and output types
* Support for alpha and beta in the case of a simple copy (i.e., perm = identity) is still missing
* The test should randomly transpose the input tensors, so that TensorTransform could transpose them back


src/ocl/tensorocl.cpp
```diff
@@ -2120,6 +2122,28 @@ void CastTensor(const Handle& handle,
...
+        permMap[dst_layout[i]].second = i;
+
+    std::vector<std::pair<int, int>> permVec;
+    for(const auto& kv : permMap)
+        permVec.push_back(kv.second);
+    std::sort(permVec.begin(), permVec.end(), [](const auto& a, const auto& b) -> bool {
+        return a.first < b.first;
+    });
+
+    printf("Permutation for %s -> %s: ", src_layout.c_str(), dst_layout.c_str());
```
atamazov, line 2159: Use `MIOPEN_LOG_I2/I/W/E` macros, no printf please.

atamazov (changes requested):
I guess that gputt cannot be used with OCL backend. If so then please use conditional compilation (see config.h) as needed.

BradPepersAMD:

PR #2006, opened 2023-03-01
[Tuning][MI100][MI200] DB version update

cderb:
Initial PR for golden version 13
Contains tunings for conv1d kernels.
Contains find mode tunings for pytorch and tensorflow.


junliume (approved):
Tested a few previously regressed cases, *most* are recovered, except for densenet121-bs128-4gpu-fp16. It's not recoverable with FIND_MODE=1 though, likely authentic one.

PR #2023, opened 2023-03-14
[Tuning][MI100][MI200] DB version 13 kdb update

cderb:
Re-export golden 13 with accompanying kdb files
export no longer includes Tuna entries which were generated using the wrong layout


junliume:
> Re-export golden 13 with accompanying kdb files export no longer includes Tuna entries which were generated using the wrong layout

@cderb could you also update ginger?

cderb:
> > Re-export golden 13 with accompanying kdb files export no longer includes Tuna entries which were generated using the wrong layout
> 
> @cderb could you also update ginger?

Sure thing.

junliume (approved):

PR #2028, opened 2023-03-16
bg/LWPMIOPEN-149 : add log command in conv fusion

bghimireamd:
* Currently driver command for fusions are not logged in MIOpen. This PR enables logging of driver command for fusions. The logging is done so that tuna can consume the driver command for its tuning. 
* Fusion plan descriptor now requires a int that describes the fusion mode. It is used in the logging. This variable could also be used to know the general state of the fusion plan that was being created. 
* Also fixed spelling mistake in code. 

Final output :

$$ **./bin/MIOpenDriver CBAInfer --fusion_mode 4 --in_h 224 --in_w 7 --out_channels 112 --pad_h 3 --pad_w 3 --fil_w 7 --fil_h 7**
MIOpenDriver CBAInfer --fusion_mode 4 --in_h 224 --in_w 7 --out_channels 112 --pad_h 3 --pad_w 3 --fil_w 7 --fil_h 7
MIOpen(HIP): Info [Handle] stream: 0x1118410, device_id: 0
Running fusion: Convolution+Bias+Activation
**MIOpen(HIP): Command [LogCmdFusion] ./bin/MIOpenDriver CBAInfer -F 4 -n 32 -c 3 -H 224 -W 7 -k 112 -y 7 -x 7 -p 3 -q 3 -u 1 -v 1 -l 1 -j 1 -g 1**
...


bghimireamd:
The test will pass we have updated the CK hash in requirements.

src/fused_api.cpp
```diff
@@ -35,6 +35,24 @@
...
+enum class ConvDirection
+{
+    Fwd = 1,
+    Bwd = 2,
+    WrW = 4
+};
+
+namespace miopen {
+namespace debug {
+void LogCmdConvolution(const miopenTensorDescriptor_t& xDesc,
```
cderb: This is a forward declaration to use the function in convolution_api.cpp?

bghimireamd: yes! Should I write this in comment to make it more clear?

junliume (approved):

junliume:
Ping @JehandadKhan before merging

include/miopen/miopen.h
```diff
@@ -2820,11 +2820,14 @@ typedef enum
...
  * @param inputDesc     Descriptor to tensor for the input (input)
+ * @param fusion_mode   id to represent fusion plan's mode. eg: for conv + bias + activ would be 0
+ * (miopen_fusion_cba)
  * @return              miopenStatus_t
  */
 MIOPEN_EXPORT miopenStatus_t miopenCreateFusionPlan(miopenFusionPlanDescriptor_t* fusePlanDesc,
                                                     const miopenFusionDirection_t fuseDirection,
-                                                    const miopenTensorDescriptor_t inputDesc);
+                                                    const miopenTensorDescriptor_t inputDesc,
+                                                    int fusion_mode);
```
JehandadKhan, line 2830: This changes the API, I am not sure why this is necessary ?

JehandadKhan (changes requested):

include/miopen/miopen.h
JehandadKhan: These contant/enums can be used internally in MIOpen to check for specific case/ logging etc. However, such an enum in the public facing API defeats its purpose for composability.

JehandadKhan:
This PR needs to be reworked. All that was required was to write a function which can take a fusion plan and return the value of the fusion plan type enum for logging purposes. All of this needs to be internal detail. exposing it to the API is no go .

JehandadKhan:
Closing in favor of #2189

PR #2036, opened 2023-03-21
initialize rocblas handle to avoid slow first GEMM call

JehandadKhan:
The MIOpen handle creates the rocBlas handle, however, it is initialized on the first GEMM call which involves loading code objects for the target GPU. This results in incorrect timing for GEMM based convolutions. 

This PR fixes the above issue by calling the `rocblas_initialize` API call to properly init rocBlas at the handle creation time.


src/hip/handlehip.cpp
```diff
@@ -593,6 +593,7 @@ rocblas_handle_ptr Handle::CreateRocblasHandle() const
     rocblas_create_handle(&x);
     auto result = rocblas_handle_ptr{x};
     rocblas_set_stream(result.get(), GetStream());
```
JehandadKhan: [anchor]

atamazov:
@JehandadKhan The library uses lazy initialization in many places. For example, we use large external libraries for compilation (LLVM) and loading them into memory has a similar effect.

Explicit initialization (while lazy initialization is already implemented) leads to additional complexities.

Finally, there may be cases where a handle is created but never used with rocblas, so explicit initialization just wastes time and resources.

For the above reasons, I think it's better to use the "warm up" approach when we need precise timings.

What is the specific problem we need to solve right now?

JehandadKhan:
> @JehandadKhan The library uses lazy initialization in many places. For example, we use large external libraries for compilation (LLVM) and loading them into memory has a similar effect.
> 
> Explicit initialization (while lazy initialization is already implemented) leads to additional complexities.

@atamazov  For example ? IIRC our discussion, all we need to do is explicitly call the `*Impl` functions at `handle` creation time? 

> 
> Finally, there may be cases where a handle is created but never used with rocblas, so explicit initialization just wastes time and resources.
> 

In my opinion this cost is minimum. 

> For the above reasons, I think it's better to use the "warm up" approach when we need precise timings.
> 
> What is the specific problem we need to solve right now?

This design leads to random delays in the first iteration and makes performance assessments different to make.

atamazov:
@JehandadKhan 
> This design leads to random delays in the first iteration and makes performance assessments different to make.

I see. We had the similar problem during performance measurements with MIOpenDriver and that is why I've implemented the "warm-up" feature in the driver (`-w 2`). Am I understanding correctly that it is impossible to implement the warm-up in the tools used for performance assessment you've mentioned?

As we discussed, I am not against explicit initialization provided that lazy init feature is untouched (even if unused when explicit init is engaged).

BTW maybe it is possible to control explicit init via environment variable and set that variable only when performance assessment is required? That way we'll avoid any (even minimal) overhead during normal work, what do you think?

JehandadKhan:
> I see. We had the similar problem during performance measurements with MIOpenDriver and that is why I've implemented the "warm-up" feature in the driver (-w 2). Am I understanding correctly that it is impossible to implement the warm-up in the tools used for performance assessment you've mentioned?

Indeed, why would any client for MIOpen want to implement such a machinery. Infact, also such machinery in the driver has made it very complicated with numerous code paths in the driver. If we dont have lazy initializing, we would not need that complexity in the driver and simplify it quite a bit.

> As we discussed, I am not against explicit initialization provided that lazy init feature is untouched (even if unused when explicit init is engaged).

I dont see the benefit of that, but I dont know of any harm either.

> BTW maybe it is possible to control explicit init via environment variable and set that variable only when performance assessment is required? That way we'll avoid any (even minimal) overhead during normal work, what do you think?

I dont think that is such a great idea. Firstly, we have more than enough env vars already, we are unable to test all the combinations of all the environment variables. Secondly, MIOpen should work like a production libarary out of the box and not depend on various env vars to be set over and over again.

atamazov:
@JehandadKhan 
> Infact, also such machinery in the driver has made it very complicated with numerous code paths in the driver.

I do not think so. Basically, there is one function `RunWarmupFindForwardGPU()`, one code block in `ConvDriver<Tgpu, Tref>::GetandSetData()` within `if(warmup_enabled) {...} ` and another one in `ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()`. I remember that implementation was trivial. But of course, it adds complexity (but it's worth it because we need a flexible tool to measure the performance of individual convolutions.)

> Firstly, we have more than enough env vars already, we are unable to test all the combinations of all the environment variables. Secondly, MIOpen should work like a production libarary out of the box and not depend on various env vars to be set over and over again.

As I said, it's all up to you. However, I would like to remind that such an env var is a debugging tool that is intended to be used by qualified engineers only, and I do not think we need a test for it (we only need to guarantee that if variable is not defined then the behavior of the library should not change -- but this is ensured because of simplicity of the proposed design). And another use for it can be for triaging problems that (God forbid) could theoretically arise after the introduction of explicit initialization.

Thanks for feedback!

JehandadKhan:
> As I said, it's all up to you. However, I would like to remind that such an env var is a debugging tool that is intended to be used by qualified engineers only, and I do not think we need a test for it (we only need to guarantee that if variable is not defined then the behavior of the library should not change -- but this is ensured because of simplicity of the proposed design). And another use for it can be for triaging problems that (God forbid) could theoretically arise after the introduction of explicit initialization.

We can add an env var if it helps, please make sure that the env var has the `MIOPEN_DEBUG` prefix and document that env vars with such a prefix are only for development and should not be used in production. 

@atamazov  Would you be able to fix this PR up and make it ready for merging ?

atamazov:
@JehandadKhan 
> @atamazov Would you be able to fix this PR up and make it ready for merging ?

Partially, because I don't have permission to change branches in this repo. But I can clone this branch into my fork and make the necessary changes there. Then you just merge the changes from there to here. Should I proceed?

JehandadKhan:
> @JehandadKhan
> 
> > @atamazov Would you be able to fix this PR up and make it ready for merging ?
> 
> Partially, because I don't have permission to change branches in this repo. But I can clone this branch into my fork and make the necessary changes there. Then you just merge the changes from there to here. Should I proceed?

Yes please!

JehandadKhan:
Closing until @atamazov prepares the finished version

atamazov:
@JehandadKhan I think that https://github.com/atamazov/MIOpen/blob/jd__rblas_init_updated is ready for merging here. 
- Added and engaged the `MIOPEN_DEBUG_PREFER_LAZY_INITIALIZATION` env var.
- [Docs] Added documentation about `MIOPEN_DEBUG_PREFER_LAZY_INITIALIZATION`.
  - Can be seen right away at https://github.com/atamazov/MIOpen/blob/jd__rblas_init_updated/docs/DebugAndLogging.md#initialization-order
- [Docs] Added general warning about `MIOPEN_DEBUG` variables. Some formatting.

Hope this helps.

atamazov:
@JehandadKhan Ping

JehandadKhan:
Updated with fixes from @atamazov  and reopened

DrizztDoUrden:
I would suggest the following:
1. Make some type like
```C++
template <class Value, bool thread_safe = true>
struct Lazy
{
    Lazy(std::function<Value()> init_) : init(std::move(init_)) {}
    Lazy(Value init_) : value(std::move(init_)) {}

    const Value* operator->() const
    {
        if (!value)
        {
            // some mutex can be added here when thread_safe == true
            // No double init can be ensured with the following: if (value) return *value;
            TryInit();
        }
        return *value;
    }

    void TryInit() const
    {
        value = init;
        init = nullptr;
    }

private:
    mutable std::optional<Value> value;
    mutable std::function<Value()> init;
}
```
2. We can store a bunch of fields in handle with it. This makes adding and using lazily initialized data trivial. Non-MT-safe option is for lazy locals.
3. We can add an API function like miopenForceInitHandle(miopenHandle_t), which would call TryInit on every lazy field of the handle specified.

I am not sure if this PR is still alive, but got asked to comment on this.

In fact, I'd do this class a bit later anyway if it would not be done yet to use it for at least the following: perf db instance in find, InvokerFactory, kernel info as they are not always consumed after being created and therefor can waste some time.

BradPepersAMD:

PR #2039, opened 2023-03-22
[tests] Test for #1928: [TensorDescriptor] Add input parameters validation

averinevg:
This PR adds test for https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1928

Additional changes:

- Old `tensor_test` has been removed, new test covers all the functionality of the old one
- Logging for `miopenSetNdTensorDescriptorWithLayout()` has been fixed
- Logging for negative `num_lens` has been fixed


test/gtest/tensor_api.cpp
```diff
@@ -0,0 +1,437 @@
...
+            {
+                const TestConfig config = {false,
+                                           {datatype,
+                                            layout,
+                                            ndims,
+                                            dims + (max_ndims - ndims),
+                                            use_strides ? (strides + (max_ndims - ndims)) : nullptr,
+                                            use_strides},
+                                           true};
+                configs.push_back(config);
```
DrizztDoUrden: This would be more performant and less memory consuming with a visitor pattern: rather than storing them to a vector this could take a `std::function<void(const TestConfig&)>` and just call it to consume the result. So from  `TEST_F(TestTensorApi, SetTensor)` it could be called like this
```C++
    RunValidTestConfigs<true>([](auto&& config){ TestTensorApi{}.RunTest(config)); });
    RunValidTestConfigs<false>([](auto&& config){ TestTensorApi{}.RunTest(config)); });
```
So only one config is actually in memory at any given time.

DrizztDoUrden (changes requested):
One minor-ish issue.

test/gtest/tensor_api.cpp
junliume: @averinevg could you follow up on this one?

averinevg: > @averinevg could you follow up on this one?

@junliume Sure

DrizztDoUrden (approved):

JehandadKhan (approved):

junliume (approved):

PR #2042, opened 2023-03-23
Faster batchsize split

msaudulhassan:
Batchsize splitting done during the verification of igemm_kernels takes too long. This is because the calculation for max_n in https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/conv/asm_implicit_gemm.hpp#LL253-L266C22 is slow. In theory, max_n can have a value anywhere from 4294967296 (`0xfffffffful`) to 1. The current algorithm starts with a max_n of 4294967296 / image_size, and it decrements it by 1 until max_n becomes a multiple of batch_size. For a small image_size (which is often the case for neural networks), this linear time search can be prohibitively slow, and can be improved significantly using the algorithm in this PR.

Note that KernelTuningNet for iGEMM does beam search. If there are N tuning parameters to predict, and we have a beam width of K, then constraint checking needs to happen N*K times, and  any slowdowns in constraint checking can slowdown KernelTuningNet by a lot.


carlushuang (approved):
seems reasonable LGTM

junliume (approved):

atamazov:
@msaudulhassan The idea is very good. 👍

Issue #2049, opened 2023-03-26
limits file not found


Epliz:
Hi,

When running the "AI Benchmark" form https://ai-benchmark.com/ranking_deeplearning.html for the first time, I got a crash with the following info:
```
MIOpen(HIP): Error [Compile] 'hiprtcCompileProgram(prog.get(), c_options.size(), c_options.data())' naive_conv.cpp: HIPRTC_ERROR_COMPILATION (6)
MIOpen(HIP): Error [BuildHip] HIPRTC status = HIPRTC_ERROR_COMPILATION (6), source file: naive_conv.cpp
MIOpen(HIP): Warning [BuildHip] /tmp/comgr-e64065/input/naive_conv.cpp:39:10: fatal error: 'limits' file not found
#include <limits> // std::numeric_limits
         ^~~~~~~~
1 error generated when compiling for gfx908.
terminate called after throwing an instance of 'miopen::Exception'
  what():  /long_pathname_so_that_rpms_can_package_the_debug_info/data/driver/MLOpen/src/hipoc/hipoc_program.cpp:304: Code object build failed. Source: naive_conv.cpp
Aborted (core dumped)
```

It seems like I resolved it by installing the `libstdc++-12-dev` on ubuntu 22.04.2 LTS .
I have rocm 5.4.3 installed through packages.
I guess one of your packages should declare that one or the right one as dependency.

Best regards,
Epliz

atamazov:
@Epliz Thanks for reporting. It seems like you do not have the binary kernel cache package installed.

This is HIPRTC-specific issue. naive_conv.cpp needs to be fixed like this:

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/d3ee8a87fa6e7b0a9db1b41f1d7c11acb529786a/src/kernels/static_composable_kernel/include/utility/static_kernel_reduction_operator.hpp#L29-L31

More details can be found at #627.

[Attribution] @junliume @johnny-keker https://github.com/ROCmSoftwarePlatform/MIOpen/labels/bug  https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_normal, Proposed assignee: @carlushuang

junliume:
@Epliz @atamazov this is a duplicated issue of https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1921 at least with similar root cause

Epliz:
thank you for looking into this.
Regarding:

[...]

PR #2050, opened 2023-03-27
guard header with MIOPEN_DONT_USE_HIP_RUNTIME_HEADERS

carlushuang:
this is to fix issue https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2049 by adding `MIOPEN_DONT_USE_HIP_RUNTIME_HEADERS` around headers, as suggested https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2049#issuecomment-1484242502


atamazov (approved):
LGTM! However I do not see if any entities from `<limits>` are used in the kernel.

junliume (approved):

PR #2058, opened 2023-03-30
Enable docker building in the CI

JehandadKhan:


junliume (dismissed):

Jenkinsfile
```diff
@@ -208,20 +214,23 @@ def getDockerImage(Map conf=[:])
...
+
+    def dockerImage
+    try{
         echo "Pulling down image: ${image}"
-        retimage = docker.image("${image}")
-        retimage.pull()
+        dockerImage = docker.image("${image}")
+        dockerImage.pull()
     }
     catch(Exception ex)
```
shurale-nkn, line 230: please add this to avoid catching an Interrupt signal
catch (org.jenkinsci.plugins.workflow.steps.FlowInterruptedException e){
                echo "The job was cancelled or aborted"
                throw e
            }

shurale-nkn (changes requested):

Jenkinsfile
JehandadKhan: Fixed

shurale-nkn (approved):
LGTM

junliume (approved):

PR #2033, opened 2023-03-20
Restructure find-db to match Find2.0 requirements

atamazov:
- Implements https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1987#issuecomment-1475425835.
  - `WORKAROUND_ISSUE_1987` provides compatibility with old format (for reading existing system find-db files, A/B testing etc). Let's keep it until system find-db files regenerated in the new format.
- [performance][Immediate mode] Optimize out unnecessary IsApplicable calls.
  - ⚠️ It is especially important to avoid IsApplicable of MLIR solvers
- [tests] `test_find_db` fixed to avoid false failures. Logging improved to ease triaging.
- [quality] Removes unused code (`FindDbKCacheKey` etc), improves logging etc.

### The most important changes
- e7f3ee839bf7b24f9fcfd9ccd79b87349e740c2e Change find-db key from algorithm to solver_id. Write all solvers' results info find-db. In Find API calls, emulate 1.0 results.
- e83ffbf85750442c9cdaa9ed517b9f7015f54c8b WORKAROUND_ISSUE_1987: Maintain backward compatibility with 1.0 find-db format.
- **_By-product:_** ee4365e483e5ca16dbffef40b138faa386424407 [Immediate mode] Optimize out unnecessary IsApplicable calls.

### Less important things
- 1682c7da1cb047bd0b08ecd15fb43af011a961f0 Dismiss FindDbKCacheKey::network_config and related stuff
- 85b1f9abfed16526a2c7f644394d9f6dc7dfec14 Dismiss the whole FindDbKCacheKey
- 10b69f63ab751baedfec79968522d13a1254e0b9 Removed CheckInvokerSupport (we have invokers for all conv solvers)
- ab4e60263e7c387aa4492d3d3c585425c167e244 LogFindDbItem(): remove dead code and improve.
- 34f724ebf1f9d6862cc09dd0f0b0929c47fccb23 [test_find_db] Fix to really use find-db and avoid false failures.
- 42ec951c641ec822cf85b8e4463569276673ca5d Improve error handling in conv_algo_name.cpp to avoid abortions with unclear error messages like "MIOpen Error: _Map_base::at".
- fcf4b9a035ef2aebf0a52903bb5328fbeb5a6408 [test_find_db] Log elapsed time - useful for triaging failures.

### NFC changes (no need to review, basically)
- 681ff10b11d89ccc789c42c92ea7fc06b7102500 Removed unused header
[...]

PR #2060, opened 2023-03-31
Fixing the tip of develop build failure

jerryyin:
This breakage is blocking MLIR release branching. Tip of develop is broken and would yield the compile failure as below:

> /root/MIOpen/src/ocl/rnnocl.cpp:342:44: error: no matching function for call to 'CallGemm'                                                                                                                                                                                                                                                                      const miopenStatus_t gemm_status = CallGemm(handle,                                                                                                                                                                                                                                                                                                                                        ^~~~~~~~                                                                                                                                                                                                                                                                                                     
/root/MIOpen/src/include/miopen/gemm_v2.hpp:94:16: note: candidate function not viable: no known conversion from 'std::nullptr_t' to 'miopen::GemmBackend_t' for 9th argument                                                                                                                                                                           miopenStatus_t CallGemm(const Handle& handle,                                                                                                                                                                                                                                                                                                                          ^                                                                                                                                                                                                                                                                                                                                        

> /root/MIOpen/src/ocl/rnnocl.cpp:454:44: error: no matching function for call to 'CallGemm'                                                                                                                                                                                                                                                                      const miopenStatus_t gemm_status = CallGemm(handle,                                                                                                                                                                                                                                                                                                                                        ^~~~~~~~                                                                                                                                                                                                                                                                                                     /root/MIOpen/src/include/miopen/gemm_v2.hpp:94:16: note: candidate function not viable: no known conversion from 'std::nullptr_t' to 'miopen::GemmBackend_t' for 9th argument                                                                                                                                                                           miopenStatus_t CallGemm(const Handle& handle,

Looks to me that this is caused by some conflict resolution failure.

### DIFF
--- a/src/ocl/rnnocl.cpp
+++ b/src/ocl/rnnocl.cpp
@@ -347,7 +347,6 @@ void RNNDescriptor::RNNForwardTraining_MS(Handle& handle,
                                                     wx_off,
                                                     reserveSpace,
                                                     out_offset,
-                                                    nullptr,
                                                     GemmBackend_t::miopengemm);
         if(gemm_status != miopenStatusSuccess)
             MIOPEN_THROW("GEMM execution failure");
@@ -459,7 +458,6 @@ void RNNDescriptor::RNNForwardTraining_MS(Handle& handle,
                                                     WeiBuf.get_matrix_h_off(layer),
                                                     reserveSpace,
                                                     RB_layer_save_points_off,
-                                                    nullptr,
                                                     GemmBackend_t::miopengemm);
 
         if(gemm_status != miopenStatusSuccess)


atamazov (approved):
LGTM!

atamazov:
The reason is that #2033 wasn't rebased onto tip of develop prior merging.

junliume:
@atamazov  I have to revert this change since neither #2060 nor #2061 were enough to unblock the TIP of `develop`.

FIN needs to be updated with the problem-context splitting, we are getting CI issues like:
```
[2023-03-31T20:03:18.212Z] /var/jenkins/workspace/MLLibs_MIOpen_fix_conflict_2033/fin/src/include/conv_fin.hpp:198:16: error: no matching constructor for initialization of 'miopen::ConvolutionContext'

[2023-03-31T20:03:18.212Z]     auto ctx = miopen::ConvolutionContext{problem};

[2023-03-31T20:03:18.212Z]                ^     
```

Here is how to reproduce:
```
git submodule init
git submodule sync
git submodule update fin
mkdir build; cd build
CXX=/opt/rocm/llvm/bin/clang++ cmake -DCMAKE_PREFIX_PATH=/var/jenkins/workspace/MLLibs_MIOpen_fix_conflict_2033/install -DCMAKE_INSTALL_PREFIX= -DCMAKE_BUILD_TYPE=release ..
make -j32 all
```

junliume:
> The reason is that #2033 wasn't rebased onto tip of develop prior merging.

Due to issues mentioned in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2033#issuecomment-1492813328 I reverted #2033 and think the best approach is to do #2033 again :) @atamazov

PR #2072, opened 2023-04-05
Deprecate static kernels for new asics

JehandadKhan:


src/solver/conv_asm_1x1u.cpp
```diff
@@ -377,6 +377,8 @@ bool ConvAsm1x1U::IsApplicable(const ConvolutionContext& ctx,
 {
     if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_DIRECT_ASM_1X1U{}))
         return false;
+    if(IsDeprecatedStatic(ctx))
```
shurale-nkn, line 380: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/b3c1ae6db390f55b449bb2e0247ef6e3a093290c/src/include/miopen/find_solution.hpp#L330-L334

can we do the same for IsDeprecatedStatic ?
not to change every single solver

atamazov: @shurale-nkn In some solvers we have conditions like "if device name begins with gfx9, then allow this solver to run". I think that @JehandadKhan wants to disable old non-dynamic kernels for new GPUs quickly and with 100% guarantee. Maybe some solvers do not require any modification, maybe not, but this change is an ultimate tool, like a hammer, which does the job.

Technically, this is a patch (which likely needs to be improved ;)

@JehandadKhan I can double check this and try to make it more elegant, if you wish ;)

shurale-nkn: @atamazov  I wrote my comment precisely for the purpose of " disable non-dynamic kernels for new GPUs with 100% guarantee"
If the target is to disable all bunch of static solvers, as it defined in the PR name, better to do it by checking `IsDeprecatedStatic(ctx) && !solver.IsDynamic()`
in this case we 100% sure about this PR coverage. `solver.IsDynamic()` was added a long time ago and there is no reason to think that it does not work correctly.

If target is to disable only several static solvers this approach will not work. And need to check `IsDeprecatedStatic(ctx)` from `IsApplicable`

atamazov: @shurale-nkn 
> If target is to disable only several static solvers...

Unfortunately, yes.

atamazov:
@JehandadKhan I am working on this

atamazov:
@JehandadKhan It seems like `ConvOclBwdWrW2<N>` is missed in this PR. I'll add it.

atamazov:
Superseded by #2081, can be closed.

JehandadKhan:
> Superseded by #2081, can be closed.

Closing!

PR #2088, opened 2023-04-13
[centOS/RHEL] Add bzip2 in MIOpen dependency

### LINKED ISSUE #2087: [dependency] bzip2 not found on centOS
**[Symptoms]**:
Since some DB files are growing out of GitHub limit, we introduced `bzip2` in #2037 
which works fine for Ubuntu and other Debian, but ... not RHEL or centOS since bzip2 is not pre-installed

**[Solution]**
Add `bzip2` as a dependency in `requirements.txt`:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/requirements.txt

Also since it is already in https://github.com/ROCmSoftwarePlatform/rocm-recipes/tree/master/miopen/bzip2
We may only need to add a line:
```
bzip2@1.0.8
```

junliume:
resolve https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2087

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 sqlite3@3.17 -DCMAKE_POSITION_INDEPENDENT_CODE=On
+bzip2@1.0.8
 boost@1.79 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build -DCMAKE_CXX_FLAGS=" -std=c++14 -Wno-enum-constexpr-conversion "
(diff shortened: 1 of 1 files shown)


junliume:
@pfultz2 and @JehandadKhan can confirm but this looks strange:
by this command:
`mkdir -p install_dir && cmake -P install_deps.cmake --minimum --prefix /root/MIOpen/install_dir`
then `bzip2` will be installed by default. **this PR should not be necessary**

But why release team is founding issues with #2087? 

This PR is on hold for now.

PR #2099, opened 2023-04-17
[MLIR] Address tuning failures

jerryyin:
~~I have heard from @JehandadKhan from last group meeting that MIOpen will not consume the latest release of MLIR or CK. Regardless of the context, I still bump the release pointer to make sure that from MIOpen perspective, MLIR's release commit is a good one.~~

In this PR, I need to implement a minor fix to make sure MLIR solver would not throw in middel of `isValid()` check. `miirGetKernelCount()` is invoked in middle of the `isValid()`. In MLIR end, the behavior of this API is such that when the config of the handle doesn't make sense, it will return `-1`, causing MIOpen end wrapper to throw. Because of this throw, it will break the MLIR solver tuning progress. I'm removing this throw as this API has only been used in two places. The two places are: `GetSolution()` and `IsValid()`. Since `GetSolution()` cannot run without `IsValid()` returns true, this is a safe change.

~~With this PR, please (formally) confirm that MIOpen will will not consume the latest commit hash. Once this is confirmed, I'll drop the commit hash update pointer, so that the no throw part of the code change can go into MIOpen's ROCm 5.6 release branch.~~

Note: CI has been overridden with image: `rocm/miopen:ci_mlir_release_rocm_5.6` and has full pass.

Note: Followed up with both @junliume and @JehandadKhan in the meeting that MIOpen will not consume MLIR for 5.6 release branch. This PR is on hold until the MIOpen 5.6 release branch is cut.


src/mlir_build.cpp
```diff
@@ -95,10 +95,7 @@ void MiirGenBin(const std::string& params, std::vector<char>& buffer)
 int MiirGetKernelCount(const std::string& params)
 {
     AutoMiirHandle handle(params);
-    const auto kernel_count = miirGetKernelCount(handle());
-    if(kernel_count < 1)
-        MIOPEN_THROW("miirGetKernelCount invalid count: " + std::to_string(kernel_count));
-    return kernel_count;
+    return miirGetKernelCount(handle());
```
atamazov:
```suggestion
    const auto n = miirGetKernelCount(handle());
    return n < 0 ? 0 : n;
```
This is necessary if `miirGetKernelCount` may return negative value (otherwise `IsValid` may misbehave). Or rocMLIR APL has changed in 5.6?

atamazov (changes requested):

src/mlir_build.cpp
jerryyin: Thanks Artem! More reasonable than just returning -1 as is. 

This API behavior didn't change since it come out. It has been always like this since it get exposed. The reason why I have just discovered this is due to MIOpen ctest starts to do tuning for smoke tests, and having throw in the log make the ctest to fail. That's how I realize this issue exists.

atamazov (approved):
LGTM!

src/mlir_build.cpp
atamazov: @jerryyin Well, actually I was wrong when I said that _"...(otherwise `IsValid` might not behave correctly)."_ That would have worked correctly. On the other hand, now the code looks better because the `MiirGetKernelCount` wrapper returns the mathematically correct number of kernels.

jerryyin:
@junliume I'm assuming that MIOpen will not consume the commit hash for 5.6? If that's the case, shall I remove the change from `requirements.txt`?

junliume:
@jerryyin is this one good to go now?

jerryyin:
@junliume Yes, it is still beneficial to leave mlir solver a "bug free" state :-)

junliume (approved):

Issue #2082, opened 2023-04-10
Using different handles for miopenFindConvolutionForwardAlgorithm and miopenConvolutionForward 


densamoilov:
The MIOpen documentation says that `miopenFindConvolutionForwardAlgorithm` must be called prior to calling `miopenConvolutionForward` however, it doesn't say that the handle objects used for the calls must be the same. Based on my observations MIOpen uses the handle object to carry over some information between the calls. This means that even if there are two different handle objects that are associated with the same device and stream, and one of them is used for the `miopenFindConvolutionForwardAlgorithm` and another one is used for the `miopenConvolutionForward` call the latter will still fail. 

If my understanding is correct the requirement is a significant discrepancy between MIOpen and cuDNN where `cudnnFindConvolutionForwardAlgorithm` and `cudnnConvolutionForward` can be called using different handles.

Is there a way to circumvent the requirement other than using the immediate API?

atamazov:
@densamoilov Your understanding is correct. We need to update the documentation at least.

Please indicate the importance/urgency of this problem for you (assign labels to this ticket).

Simplified, the reason is that `miopenConvolutionForward` is expecting that all necessary kernels already built, loaded and ready to use, BUT this is not so, because each handle has its own kernel cache.

Do you use different-but-same handle objects from the same thread/process? In other words is MT/MP safety a requirement?

Right away, I can recommend using the Find 2.0 API (see #1503). IIRC It finds an _optimal_ solution just like old Find API (`miopenFindConvolutionForwardAlgorithm`) but returns solver information that is _compatible_ with Immediate Mode API calls. Then you can then use Immediate Mode API to _run_ the found solution. Immediate Mode builds kernels when necessary.

The full interoperability between Find and Immediate Modes is one of our current goals (#374) and I am hoping we already reached a level suitable for you.

@DrizztDoUrden Please confirm my thoughts. Let's discuss this tomorrow.

densamoilov:
Thanks @atamazov, the Find 2.0 API is really helpful. The only thing I'd like to clarify is about interoperability between the modes. The newly introduced `miopenSolution_t` can be queried for workspace size and time but it cannot be queried for a solution ID that is required to call `miopenConvolutionForwardImmediate`. So it doesn't seem like the Find 2.0 API is fully interoperable with the Immediate API.

[...]

Issue #374, opened 2020-08-11
Find 2.0 design


daniellowell:
Let's begin the planning for Find 2.0.

Some preliminary requirements:
1. Find 2.0 should unify the perf data structure of Findv1.0 and immediate mode
2. Find 2.0 should be able to accept the uint64 solver ID

atamazov:
### Clarifications/Rationale/Plan

- [ ] (100) Find 2.0 API.
  - The main goal is interoperability between Find and Immediate modes.
  - Simplification/robustness of MIOpen code.
  - [x] (100.1) Find-db: Change structure to better suit Immediate mode
    - Why: Currently find-db "mirrors" requirements imposed by Find 1.0 API. It is able to hold only one set of metrics per algorithm; if more than one Solver is available within algorithm, then only the "fastest" metrics are stored. In other words, the fastest Solver may "hide" other fast Solvers of the same algorithm, and Immediate Mode does not return the full information to the caller. We need to store metrics from _all applicable_ Solvers.
    - This would also resolve #1987 at once.
    - 🛠️ Implemented in #2033
  - [x] (100.10) Make sure that Find 2.0 API returns `miopenConvSolution_t` (or information compatible with it) and can be used as a replacement of Find 1.0 API.
    - 🛠️ Implemented in #2104
    - Partially described at https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2082#issuecomment-1514784920
  - [ ] (100.20) Announce deprecation of Find 1.0 API.
  - [ ] (100.30) Dismiss `miopenConvAlgoPerf_t` and finally turn algorithm from a base value denoting the found convolution into an attribute of it.
  - [ ] (100.40) Remove Find 1.0 API and its rudiments (from the invoker cache etc). Announce final removal of Find 1.0 API.
- [ ] (110) Expose solver_id string <-> uint64 conversions. Write tests for those.
  - Partially described at https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2082#issuecomment-1514784920
  - Note: This plus bullet (100) allows us to implement https://github.com/AMDComputeLibraries/MLOpen/issues/426 (outdated?)

Origins:
- https://github.com/AMDComputeLibraries/MLOpen/issues/1938 "Architectural improvements", bullet (100)
- https://github.com/AMDComputeLibraries/MLOpen/issues/2093 "Find 2.0"

atamazov:
Another specific requirement (from #424):
```C++
[...]

PR #2104, opened 2023-04-20
[API] Solver id getter from a solution and solver id to algo converter

DrizztDoUrden:
Simply adds one solution getter and one value converter to the API. Allows people using Find 2.0 to filter found solutions by solver and algorithm.

This implements (100.10) from https://github.com/ROCmSoftwarePlatform/MIOpen/issues/374 and now we have a one-way interoperability between Find 2.0 and Immediate Modes. And the results of Find 2.0 can be "fully" examined now (for whatever purposes, e.g. filtering out unwanted algorithms or solvers, as said in the description of the PR).

Should resolve the immediate problem from #2082.

### DIFF
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -4985,7 +4985,7 @@ miopenStatus_t miopenGetSolutionWorkspaceSize(miopenSolution_t solution, size_t*
  */
 miopenStatus_t miopenGetSolutionTime(miopenSolution_t solution, float* time);
 
-/*! @brief Reads id of the solver referred to by the solution.
+/*! @brief Reads id of the solver referred by the solution.
  *
  * @param solution Solution to get solver id from
  * @param solverId Pointer to a location where to write the solver id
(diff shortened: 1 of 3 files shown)


atamazov (approved):
LGTM!

atamazov:
This implements (100.10) from #374 and now we have a one-way interoperability between Find 2.0 and Immediate Modes. And the results of Find 2.0 can be "fully" examined now (for whatever purposes, e.g. filtering out unwanted algorithms or solvers, as said in the description of the PR).

@DrizztDoUrden I recommend updating the description.

pfultz2 (approved):

DrizztDoUrden:
@junliume I think this can be merged when everyone you would expect a review from is satisfied - every test except test_find2_conv is not affected at all and it just adds a couple of getters to the API.

densamoilov:
@atamazov @junliume, what is the target release for this feature?

atamazov:
@densamoilov It depends on the customer needs (that's why I've askd you about urgency) and the current status of the release plan (which I am not aware of). Please expect more informative answer from @junliume

PR #2108, opened 2023-04-21
Fix MIOPEN_ENABLE_SQLITE=Off

averinevg:
This PR fixes build with `MIOPEN_ENABLE_SQLITE=Off`

Additional changes:

- Dead code has been removed
- Unused tests have been excluded from the build


averinevg:
@atamazov Please review

src/include/miopen/fusion/problem_description.hpp
```diff
@@ -70,13 +73,26 @@ struct FusionDescription : SQLiteSerializable<FusionDescription>
         }
         MIOPEN_LOG_I2(net_config.str());
     }
+
+#if !MIOPEN_ENABLE_SQLITE
+    void Serialize(std::ostream& stream) const
```
atamazov, line 79: [Recommendation]
```suggestion
    /// \todo This function will be necessary for tuning of fusions.
    void Serialize(std::ostream& stream) const
```

test/CMakeLists.txt
```diff
@@ -414,13 +414,37 @@ function(rate_added_test NAME)
...
 endfunction()
 
-foreach(TEST ${TESTS})
-    get_filename_component(BASE_NAME ${TEST} NAME_WE)
-    add_test_executable(test_${BASE_NAME} ${TEST})
+set(EXCLUDE_TESTS)
+if(NOT MIOPEN_ENABLE_SQLITE)
+    list(APPEND EXCLUDE_TESTS test_sqlite)
+endif()
+if(NOT MIOPEN_ENABLE_SQLITE OR MIOPEN_EMBED_DB)
```
atamazov: `...OR (NOT MIOPEN_EMBED_DB STREQUAL "")` ?

test/cache.cpp
```diff
@@ -30,9 +30,11 @@
 
 #include <miopen/md5.hpp>
 #include "test.hpp"
+#if MIOPEN_ENABLE_SQLITE && MIOPEN_ENABLE_SQLITE_KERN_CACHE
```
atamazov: [Recommendation] MIOPEN_ENABLE_SQLITE_KERN_CACHE cannot be On when MIOPEN_ENABLE_SQLITE=Off
```suggestion
#if MIOPEN_ENABLE_SQLITE_KERN_CACHE
```
everywhere in this file

test/sqlite_perfdb.cpp
```diff
@@ -49,7 +49,10 @@
 #include <thread>
 #include <vector>
 
-#if MIOPEN_ENABLE_SQLITE
+#if MIOPEN_EMBED_DB
+#error "MIOPEN_EMBED_DB = On"
+#endif
+
```
atamazov: [Recommendation] Please remove this.

atamazov (changes requested):

src/include/miopen/sqlite_db.hpp
```diff
@@ -27,7 +27,9 @@
 
 #include <miopen/config.h>
 
-#if MIOPEN_ENABLE_SQLITE
+#if !MIOPEN_ENABLE_SQLITE
```
DrizztDoUrden, line 30: I think it is better to produce a CMake error in this case. It would be seen before compilation.

DrizztDoUrden (approved):
LGTM with a minor suggestion

src/include/miopen/fusion/problem_description.hpp
atamazov: resolved

src/include/miopen/sqlite_db.hpp
atamazov: The idea is that this header should not be included when MIOPEN_ENABLE_SQLITE=Off. CMake can't detect such cases ;)

test/sqlite_perfdb.cpp
atamazov: resolved

test/cache.cpp
atamazov: resolved

test/CMakeLists.txt
atamazov: resolved

atamazov (approved):
LGTM. Please check CI logs if everything works as expected.

junliume (approved):

src/include/miopen/sqlite_db.hpp
JehandadKhan: It would have been much simpler, to simply wrap this header in the `#if / #endif` instead of adding the same  to every site where this header was included.

@atamazov @averinevg

test/CMakeLists.txt
```diff
@@ -414,13 +414,37 @@ function(rate_added_test NAME)
...
-foreach(TEST ${TESTS})
-    get_filename_component(BASE_NAME ${TEST} NAME_WE)
-    add_test_executable(test_${BASE_NAME} ${TEST})
+set(EXCLUDE_TESTS)
+if(NOT MIOPEN_ENABLE_SQLITE)
+    list(APPEND EXCLUDE_TESTS test_sqlite)
+endif()
+if(NOT MIOPEN_ENABLE_SQLITE OR (NOT MIOPEN_EMBED_DB STREQUAL ""))
+    list(APPEND EXCLUDE_TESTS test_sqlite_perfdb)
+endif()
```
JehandadKhan, line 423: Again we are adding test logic to the CMake file, this should just have been disabled in the test itself using the `#if / #endif`

JehandadKhan: Looking through the code, seems like the above logic existed and was removed, why ? 

@atamazov @averinevg @DrizztDoUrden

atamazov: @JehandadKhan I agree. Overlooked this possibility during review. Should be easy to do.

@averinevg In `add_test_executable`, if
```
if(NOT MIOPEN_ENABLE_SQLITE OR (NOT MIOPEN_EMBED_DB STREQUAL ""))
```
and test name is "test_sqlite_perfdb" then `set_tests_properties(${TEST_NAME} PROPERTIES DISABLED On)`. Much simpler than stuff with lists. Can you please make a PR?

atamazov: @JehandadKhan 
> Looking through the code, seems like the above logic existed and was removed, why ?

To not rebuild the binaries. This is a microscopic step towards making a package of tests.

PR #2112, opened 2023-04-26
Update fdb data to use solver keys [MI100][MI200]

cderb:
issue #1987 
MI100, MI210, and MI250: Update fdb to use solver idx, and save top 4 solvers. Update kdb to contain matching kernels.
Compressed fdb files.


atamazov:
@cderb 
> ...save top 4 solvers. 

Can you please elaborate why only 4 solvers are saved? Thanks!

JehandadKhan (changes requested):
Please see #2113

cderb:
> Can you please elaborate why only 4 solvers are saved? Thanks!

@atamazov This was to attempt to keep the size of the fdb down, when allowing all solvers to be recorded, the size of the fdb bloated a bit. 4 was selected arbitrarily, as timing seemed to slow down significantly after that point from and eyed inspection. This limit could be extended if valuable.

cderb:
Closing this in favor of #2151, to refresh the CI workspace.

PR #2116, opened 2023-04-27
Performance stage update (Resnet50)

alexandraBara:
Split Resnet50 file by data type and batchsize to parallelize the CI stage.
Added try_catch for the k_time comparison to let CI pass when a stage fails (to be removed when a stable testing node is found)


Jenkinsfile
```diff
@@ -490,8 +505,23 @@ pipeline {
             description: "Enable OpenCL backend stages")
         booleanParam(
             name: "PERF_TEST",
-            defaultValue: false,
+            defaultValue: true,
```
alexandraBara, line 511: @JehandadKhan  should i reset this to false, for subsequent stages as well?

JehandadKhan: Yes, please

JehandadKhan (dismissed):

Jenkinsfile
```diff
@@ -337,20 +338,34 @@ def RunPerfTest(Map conf=[:]){
...
-            catch (Exception err){
-                currentBuild.result = 'SUCCESS'
+            if (env.BRANCH_NAME == env.MIOPEN_GOLDEN_PERF_BRANCH || params.PERF_TEST_BRANCH_OVERRIDE){
+                if(params.PERF_TEST_OVERRIDE != '')
+                {
+                    echo "Appending MIOpenDriver cmd env vars: ${params.PERF_TEST_OVERRIDE}"
+                    sh "export LD_LIBRARY_PATH=${ld_lib} && ${env.WORKSPACE}/opt/rocm/bin/test_perf.py  --filename ${filename} --install_path ${env.WORKSPACE}/opt/rocm --override ${params.PERF_TEST_OVERRRIDE}"
+                }else
+                {
+                    sh "export LD_LIBRARY_PATH=${ld_lib} && ${env.WORKSPACE}/opt/rocm/bin/test_perf.py  --filename ${filename} --install_path ${env.WORKSPACE}/opt/rocm"
```
shurale-nkn, line 350: it seems this line, or next is redundant, next line doing the same

Jenkinsfile
```diff
@@ -337,20 +338,34 @@ def RunPerfTest(Map conf=[:]){
...
+                    currentBuild.result = 'SUCCESS'
+                }
             }
 
-            if (env.BRANCH_NAME == env.MIOPEN_GOLDEN_PERF_BRANCH){
-                archiveArtifacts artifacts: "opt/rocm/bin/perf_results/${filename}", allowEmptyArchive: true, fingerprint: true
+            archiveArtifacts artifacts: "opt/rocm/bin/perf_results/${filename}", allowEmptyArchive: true, fingerprint: true
+            try{
+              if (env.BRANCH_NAME != env.MIOPEN_GOLDEN_PERF_BRANCH){
+                  sh "${env.WORKSPACE}/opt/rocm/bin/test_perf.py --compare_results --old_results_path ${env.WORKSPACE}/opt/rocm/bin/old_results --filename ${filename}"
```
shurale-nkn: I can't understand the logic, it turns out this line is executed successfully, or the result will be ignored. So what's the point of checking it?

shurale-nkn (changes requested):

Jenkinsfile
alexandraBara: You are correct, ill remove line 348 and the else

Jenkinsfile
alexandraBara: We went with this logic for now because we don't have stable testing infrastructure in the CI. Once we get the hardware we will remove the try catch and let it fail.

JehandadKhan (dismissed):

shurale-nkn (dismissed):
line 345 in [Jenkinsfile] still not fixed, was merged by another PR.
each try catch block should have 
```
catch (org.jenkinsci.plugins.workflow.steps.FlowInterruptedException e){
        echo "The job was cancelled or aborted"
        throw e
}
```

alexandraBara:
> line 345 in [Jenkinsfile] still not fixed, was merged by another PR. each try catch block should have
> 
> ```
> catch (org.jenkinsci.plugins.workflow.steps.FlowInterruptedException e){
>         echo "The job was cancelled or aborted"
>         throw e
> }
> ```

I removed the redundant line. 
The first try-catch for the wget is for when the test branch goes through the exercise the first time and there wont be any archived files. The code will just push current results to archive and compare on the next run.
The second try-catch that results in SUCCESS is due to lack of stable testing system where one or more ktime always goes below the threshold. For now we have agreed to leave it in.

junliume:
Let's maybe disable this perf stage by default for now? I see several pipelines stuck in this stage: http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/PR-2163/3/pipeline

JehandadKhan:
@alexandraBara  Can you please take a look and suggest what are the next steps here.

BradPepersAMD:

PR #2124, opened 2023-05-04
[WA] AMD MLIR lld conflicts with conda

junliume:
There is another lld shipped under /opt/conda/ 's include and lib directory, and the conda version of LLD get prioritized over MLIR version of LLD.

AMD Pytorch team that decides to go with a deeper directory level depth.

This is to fix the error messages:
```
/root/MIOpen/install_dir/cget/build/tmp-4833060d3c1d4a7ea828cdadafa262b9/rocMLIR-rocm-5.5.0/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp:458:53: error: too many arguments to function call, expected at most 3, have 5
                        llvm::outs(), llvm::errs(), false, false)) {
                                                    ^~~~~~~~~~~~
/opt/conda/envs/py_3.8/include/lld/Common/Driver.h:28:6: note: 'link' declared here
bool link(llvm::ArrayRef<const char *> Args, bool CanExitEarly,
     ^
/root/MIOpen/install_dir/cget/build/tmp-4833060d3c1d4a7ea828cdadafa262b9/rocMLIR-rocm-5.5.0/external/llvm-project/mlir/lib/Dialect/GPU/Transforms/SerializeToHsaco.cpp:462:10: error: no member named 'CommonLinkerContext' in namespace 'lld'
    lld::CommonLinkerContext::destroy();
    ~~~~~^
2 errors generated.
make[2]: *** [external/llvm-project/llvm/tools/mlir/lib/Dialect/GPU/CMakeFiles/obj.MLIRGPUTransforms.dir/build.make:174: external/llvm-project/llvm/tools/mlir/lib/Dialect/GPU/CMakeFiles/obj.MLIRGPUTransforms.dir/Transforms/SerializeToHsaco.cpp.o] Error 1
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 sqlite3@3.17 -DCMAKE_POSITION_INDEPENDENT_CODE=On
 boost@1.79 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build -DCMAKE_CXX_FLAGS=" -std=c++14 -Wno-enum-constexpr-conversion "
 half,https://github.com/pfultz2/half/archive/1.12.0.tar.gz -X header -H sha256:0a08660b68abb176ebc2a0cdf8de46e3182a7f46c66443bb80dbfaaec98cf969 --build
-ROCmSoftwarePlatform/rocMLIR@rocm-5.5.0 -H sha256:a5f62769d28a73e60bc8d61022820f050e97c977c8f6f6275488db31512e1f42 -DBUILD_FAT_LIBROCKCOMPILER=1
+ROCmSoftwarePlatform/rocMLIR@rocm-5.5.0 -H sha256:a5f62769d28a73e60bc8d61022820f050e97c977c8f6f6275488db31512e1f42 -DBUILD_FAT_LIBROCKCOMPILER=1 -DCMAKE_IGNORE_PATH=/opt/conda/envs/py_3.8
 nlohmann/json@v3.9.1 -DJSON_MultipleHeaders=ON -DJSON_BuildTests=Off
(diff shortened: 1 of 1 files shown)


junliume:
@jerryyin do we have a more reliable solution than the WA?

jerryyin:
@junliume It is okay for the short term to unblock mainline build.

Long term wise, I really think this should be solved from pytorch. Whatever we do here is going to be a guess. Once they decide to put conda in a different path we are screwed again. (Assuming upstream pytorch is unwilling to address this) In AMD Pytorch's MIOpen installation script, they should be able to find out the line where MLIR is in `requirements.txt` and add the additional ignore path before they build MIOpen. We can follow-up with them to see what they think about it.

jerryyin:
@junliume FYI the download for image is complete and I did a second attempt on `-DCMAKE_IGNORE_PREFIX_PATH=/opt/conda` and confirmed that this isn't working.

On a second look, cmake introduced this flag earliest in cmake version 3.23. However, the docker image used the cmake version 3.22. That's the reason why we can't easily block off *all* path from that `/opt/conda` prefix. The change you have in here is the only option we have given this cmake version.

----

Having said this, I still believe this is a pytorch bug and should be fixed by them. They should attach the `-DCMAKE_IGNORE_PATH` when building any sub-component to prevent them from happening. Even better, they should use `conda/bin/activate` and `conda/bin/deactivate` to prevent their global settings to affect other software. Refer to ticket here: https://github.com/pytorch/builder/issues/1233

jerryyin (approved):

PR #2127, opened 2023-05-05
[tests][HotFix] Fix doc build issue in CI pipeline with #2053

junliume:
`make doc` is broken after#2053 is merged and changed the way doc is been built


junliume:
replaced by #2128

PR #2129, opened 2023-05-05
Using faster multi-threaded lbzip2 utility if available on the system, else default to bzip2

amd-aakash:
Using the lbunzip2 helps to untar kdb files much faster than using default utility like bunzip2.


JehandadKhan (approved):

JehandadKhan:
@junliume  Can we merge this when you get a chance?

junliume (approved):

PR #2130, opened 2023-05-05
Add Navi stages to CI

JehandadKhan:


Jenkinsfile
```diff
@@ -971,7 +975,34 @@ pipeline {
                     }
                     agent{ label rocmnode("navi21") }
                     steps{
-                        buildHipClangJobAndReboot(setup_flags: Full_test + Fp16_flags)
+                        script {
+                            try{
+                                buildHipClangJobAndReboot(setup_flags: Full_test + Fp16_flags)
+                            }
+                            catch(err){
+                                unstable(message: "${STAGE_NAME} is unstable")
```
atamazov, line 983: [Q] What does “unstable” mean? Is it so that a CI job may get the "passed" status even if the "unstable" stages are failing?

Jenkinsfile
```diff
@@ -971,7 +975,34 @@ pipeline {
...
+                        retry(2)
+                    }
+                    agent{ label rocmnode("navi32") }
+                    steps{
+                        script {
+                            try{
+                                buildHipClangJobAndReboot(setup_flags: Full_test + Fp16_flags)
+                            }
+                            catch(err){
+                                unstable(message: "${STAGE_NAME} is unstable")
```
atamazov, line 1003: [Q] Maybe Navi 32 is not unstable yet?

Jenkinsfile
```diff
@@ -1049,6 +1080,19 @@ pipeline {
...
+                    when {
+                        beforeAgent true
+                        expression { params.TARGET_NAVI32 && params.DATATYPE_FP32 }
+                    }
+                    options {
+                        retry(2)
+                    }
+                    agent{ label rocmnode("navi32") }
+                    steps{
+                        buildHipClangJobAndReboot(setup_flags: Full_test, build_cmd: Navi21_build_cmd, build_install: "true")
```
atamazov, line 1093: [Q] I suspect that `Navi21_build_cmd` (which is a W/A for a Navi21-specific SWDEV-290754) is not necessary for both Navi 21 and 32. Or I am wrong?

Jenkinsfile
```diff
@@ -458,6 +458,10 @@ pipeline {
             name: "TARGET_NAVI21",
             defaultValue: true,
             description: "")
+        booleanParam(
+            name: "TARGET_NAVI32",
+            defaultValue: true,
```
atamazov, line 463: [Q] So we have Navi32 in our CI pool? Great!

junliume: Yes!~ limit quantity so this stage is under experiment.

Jenkinsfile
junliume: If this stage fails (most likely due to instability) it will not prevent the pipeline from proceeding, and we will get a pass.

junliume (approved):

Jenkinsfile
atamazov: Then this is a workaround. I highly recommend opening a ticket and following the process as described at #1685

Jenkinsfile
atamazov: Thanks! Resolved.

Jenkinsfile
atamazov: @junliume ⬆️

JehandadKhan:
Leon has taken this task up

PR #2135, opened 2023-05-08
Use of `rand()` considered harmful

JehandadKhan:
The `rand()` function `cstdlib` is very slow due to its implementation and is not thread safe either. This PR replaces it with the C++11 version of the random number generator. 

performance uplift: TBD


dmikushin:
`rand()` uses [linear congruential generator](http://en.wikipedia.org/wiki/Linear_congruential_generator), which is faster than just about anything, and definitely a lot **faster** than the proposed C++ <random> Mersenne Twister. Here is an [example of glibc  implementation](https://github.com/bminor/glibc/blob/master/stdlib/random_r.c#L364), which is basically as fast as a single ifma. One reason `rand()` could be slow is concurrent access to it is guarded by a mutex. By replacing `rand()` with `rand_r()` and thread-private state, you should get the best performance and thread safety. If you wish to proceed with C++ <random>, consider xorwow generator instead of Mersenne Twister.

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
```diff
@@ -266,8 +265,7 @@ ConvOclDirectFwdLegacyExhaustiveSearch::SearchImpl(const ConvolutionContext& ctx
     std::vector<Tgpu> wei_sys_buf(weights_sz);
     for(size_t i = 0; i < weights_sz; i++)
     {
-        wei_sys_buf[i] = static_cast<Tgpu>((rand() * (1.0 / RAND_MAX) - 0.5) *
-                                           0.001); // NOLINT (concurrency-mt-unsafe)
+        wei_sys_buf[i] = RAN_GEN<Tgpu>(-0.5 * 0.001, 0.5 * 0.001);
```
bghimireamd, line 268:
```suggestion
        wei_sys_buf[i] = RAN_GEN<Tgpu>(-0.0005, 0.0005);
```

gonidelis:
> `rand()` uses [linear congruential generator](http://en.wikipedia.org/wiki/Linear_congruential_generator), which is faster than just about anything, and definitely a lot **faster** than the proposed C++ Mersenne Twister. Here is an [example of glibc implementation](https://github.com/bminor/glibc/blob/master/stdlib/random_r.c#L364), which is basically as fast as a single ifma. One reason `rand()` could be slow is concurrent access to it is guarded by a mutex. By replacing `rand()` with `rand_r()` and thread-private state, you should get the best performance and thread safety. If you wish to proceed with C++ , consider xorwow generator instead of Mersenne Twister.

@dmikushin Could you maybe provide any sources for that? Since we are in C++ land we may as well move to the standard `<random>` library (which I still am not seconding). If we want to take it a step further, we could use one of the newest kids on the block from the public domain (e.g. xorshift, pcg etc.). Nevertheless using `thread_local std::mt19937` would beat `rand_r()` in every aspect, no?

ps: _I will be working on that in the next couple of days, feel free to share any insight :)_

dmikushin:
> using thread_local std::mt19937 would beat rand_r() in every aspect, no?

@gonidelis Absolutely not. The complexity of RNG is identified by two aspects: the size of "state" vector (a config identifying the current RNG life moment) and the arithmetics required to transfer to the next state. Mersennse twister is significantly more complex than basic rand_r in both aspects.

I think people misunderstand the purpose of `<random>` thinking that it's a proper way of doing `rand_r()` in C++11. I'd say `rand_r` and `<random>` are rather as different as char* and std::string: you get much more functionality for quite some penalty, which won't improve anything, unless you have a clever idea how to leverage this functionality.

gonidelis:
@dmikushin seems like `rand_r()` does work better than `std::mt_19937`. Should we worry about portability (POSIX, rather than C++ standards friendly). Also what about these custom PRNGs? I know they lack literature support but speed is incomparable.

https://quick-bench.com/q/y1YxWgy8BjwW0_e0qmxZ8-w6bEs
https://godbolt.org/z/vv8T4zEd5

dmikushin:
Simpler RNGs usually have worse statistical properties. The choice should be driven by our requirements. RNG is used in MIOpen testing. In fact no statistical quality requirements are imposed by this purpose. Therefore, the fastest custom data generator should be a fit.

PR #2137, opened 2023-05-08
[GFX94X] Disable Winograd Multipass

Slimakanzer:
Winograd Multipass kernels don't support gfx940 ASICs for now because of the following:
- **Compilation errors.** Some instructions have been renamed in the new ISA.
- **Data hazards.** The shader may contain gfx940-specific data hazards that should be explicitly avoided.

This PR disables Winograd Multipass solver for gfx94x ASICs.


muralinr (approved):
LGTM

junliume (approved):

PR #2134, opened 2023-05-08
F8 related changes for naive convolution solver and tests [1/5]

xinlipn:
This PR contains new rocBLAS API changes that support F8 type

PR #2138, opened 2023-05-09
F8 supporting driver changes [2/5]

xinlipn:
sl/f8_driver is branched off sl/f8 branch. It's meant for F8 driver related changes

### DIFF
--- a/include/miopen/miopen.h
+++ b/include/miopen/miopen.h
@@ -111,6 +111,12 @@ typedef enum
     miopenStatusVersionMismatch = 10, /*!< Version mismatch of the supplied binary data argment. */
 } miopenStatus_t;
 
+typedef enum
+{
+    miopenF8RoundingModeStandard   = 0,
+    miopenF8RoundingModeStochastic = 1,
+} miopenF8RoundingMode_t;
+
 /*! @brief Get character string for an error code.
  *
  * A function which returns a NULL terminated character string of the error code.
@@ -339,7 +345,9 @@ typedef enum
         4, /*!< Pack of four 8-bit int points in NCHW_VECT_C format (Partially supported) */
     miopenBFloat16 = 5, /*!< 16-bit binary floating point (8-bit exponent, 7-bit fraction)
                            (Partially supported) */
-    miopenDouble = 6,   /*!< 64-bit floating point (Partially supported) */
+    miopenDouble  = 6,  /*!< 64-bit floating point (Partially supported) */
+    miopenFloat8  = 7,
+    miopenBFloat8 = 8
 } miopenDataType_t;
 
 /*! @ingroup tensor
@@ -573,6 +581,8 @@ typedef enum
     MIOPEN_CONVOLUTION_ATTRIB_DETERMINISTIC =
         1, /*!< Restrict MIOpen convolutions to kernels which produce numerically deterministic
               results. 0 - disabled (default), 1 - enabled >*/
+    MIOPEN_CONVOLUTION_ATTRIB_FP8_ROUNDING_MODE = 2,
+    MIOPEN_CONVOLUTION_ATTRIB_COMPUTE_TYPE      = 3,
 } miopenConvolutionAttrib_t;
 
 /** @addtogroup tensor
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -154,6 +154,27 @@ void dumpBufferToFile(const char* fileName, T* data, size_t dataNumItems)
     }
 }
 
+miopenDataType_t DataTypeFromShortString(const std::string& type)
+{
+    static const std::unordered_map<std::string, miopenDataType_t> conv_map = {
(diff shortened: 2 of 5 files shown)


include/miopen/miopen.h
```diff
@@ -339,7 +345,9 @@ typedef enum
         4, /*!< Pack of four 8-bit int points in NCHW_VECT_C format (Partially supported) */
     miopenBFloat16 = 5, /*!< 16-bit binary floating point (8-bit exponent, 7-bit fraction)
                            (Partially supported) */
-    miopenDouble = 6,   /*!< 64-bit floating point (Partially supported) */
+    miopenDouble  = 6,  /*!< 64-bit floating point (Partially supported) */
+    miopenFloat8  = 7,
+    miopenBFloat8 = 8
```
atamazov, line 350: [Recommendation] Please add documentation - explain the difference between FP8 and BF8.

include/miopen/miopen.h
```diff
@@ -573,6 +581,8 @@ typedef enum
     MIOPEN_CONVOLUTION_ATTRIB_DETERMINISTIC =
         1, /*!< Restrict MIOpen convolutions to kernels which produce numerically deterministic
               results. 0 - disabled (default), 1 - enabled >*/
+    MIOPEN_CONVOLUTION_ATTRIB_FP8_ROUNDING_MODE = 2,
+    MIOPEN_CONVOLUTION_ATTRIB_COMPUTE_TYPE      = 3,
```
atamazov, line 585: Please add doc - explain what is "compute type".

include/miopen/miopen.h
```diff
@@ -111,6 +111,12 @@ typedef enum
     miopenStatusVersionMismatch = 10, /*!< Version mismatch of the supplied binary data argment. */
 } miopenStatus_t;
 
+typedef enum
+{
+    miopenF8RoundingModeStandard   = 0,
+    miopenF8RoundingModeStochastic = 1,
```
atamazov, line 117: [Recommendation] Please add doc - explain what is standard mode and what is stochastic one.

atamazov (changes requested):
Partial review. The PR requires the https://github.com/ROCmSoftwarePlatform/MIOpen/labels/IMPACTS_API label.

xinlipn:
Current PR has dependency on this PR
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2134

include/miopen/miopen.h
xinlipn: To put in an one liner. Float8(FP8) or BFloat8(BF8) are two 8-bit floating point data types. 
* FP8: 1-4-3 (sign-exponent-mantissa), more precision, smaller range
* BF8: 1-5-2 (sign-exponent-mantissa), less precision, larger range

include/miopen/miopen.h
xinlipn: > Please add doc - explain what is "compute type".

GEMM operation is defined as D = alpha * A * B + beta * C, where A*B is dot product.

JehandadKhan: @atamazov  It represents the accumulator type for accumulation in convolution.

include/miopen/miopen.h
JehandadKhan: This represents how numbers are rounded from a higher precision type to FP8

atamazov: @JehandadKhan Thanks! I am guessing the same but what I am thinking about is adding this to the doxygen doc. Does it sound reasonable?

include/miopen/miopen.h
atamazov: @JehandadKhan @xinlipn We already have this feature in some other convolutions (e.g. FP16) and in some other primitives (e.g. FP16 pooling). For so far it is hidden, i.e. not exposed in the public API. And I am not sure if it is worth to make it public for them.

If we are going to use this attribute for F8 only, then I recommend either documenting this or (better) renaming the attribute.

include/miopen/miopen.h
atamazov: @xinlipn Thanks! What I am thinking about is adding this to the doxygen doc. Does it sound reasonable for you?

include/miopen/miopen.h
```diff
@@ -573,6 +581,8 @@ typedef enum
     MIOPEN_CONVOLUTION_ATTRIB_DETERMINISTIC =
         1, /*!< Restrict MIOpen convolutions to kernels which produce numerically deterministic
               results. 0 - disabled (default), 1 - enabled >*/
+    MIOPEN_CONVOLUTION_ATTRIB_FP8_ROUNDING_MODE = 2,
```
atamazov, line 584:
```suggestion
    MIOPEN_CONVOLUTION_ATTRIB_F8_ROUNDING_MODE = 2,
```
because this is applicable to both FP8 and BF8.

PR #2139, opened 2023-05-09
[NFC] Build MIOpen packages first in the CI 

JehandadKhan:
This allows testing branch specific features easily in the stack. Otherwise, CI needs to pass before a package can be used which is inconvenient for the development cycle.


atamazov (approved):
Sounds reasonable. LGTM.

junliume (approved):

PR #2146, opened 2023-05-11
[Find2.0][API] Invoker caching and binaries serialization in miopen::Solution

DrizztDoUrden:
This updates convolution internals to support the planned features of find 2.0 and some more as a trivial side-effect.

New features:
- Invoker caching in the `miopenSolution_t`. This allows zero lookup `miopenRunSolution` calls after either `miopenFindSolutions` or `miopenLoadSolution`.
- Solution binaries serialization. This allows users to implement all file-based caching on their side and have zero lookups from the start just by deserializing and running solutions. Binaries are not stored in the solution as is when possible. Runtime program handle wrappers are kept instead. That allows to not waste additional memory, but prevents trivially implementing the features for OpenCL, so no binary serialization happens there.
- `miopenSetFindOptionAttachBinaries` with default value of `false` that controls binary attachment to solutions. With `false` machinery behaves like before the PR + invoker caching.

This has been squashed as it was based on another PR and had 120+ commits, most of which were completely unrelated or merges.

Future work:
- Something like `miopenStatus_t miopenRemoveSolutionBinaries(miopenSolution_t)` can be added to reduce the memory footprint of a solution in case it would be held in memory long-term. This way found solutions may be saved, stripped of binaries and then used.
- Some API way to detach binaries from multiple solutions to a single external cache with constant lookup time may be implemented to reduce the total size of solutions blob if multiple solutions with same binaries are saved. A set of functions like the following may be added:
  - `miopenStatus_t miopenCreateBinariesBlob(miopenBinariesBlob_t*)`
  - `miopenStatus_t miopenDestroyBinariesBlob(miopenBinariesBlob_t)`
  - `miopenStatus_t miopenDetachSolutionBinaries(miopenBinariesBlob_t, miopenSolution_t)`
  - `miopenStatus_t miopenAttachSolutionBinaries(miopenBinariesBlob_t, miopenSolution_t)`
  - `miopenStatus_t miopensSaveBinaries(miopenBinariesBlob_t, char*, size_t*)`
  - `miopenStatus_t miopensLoadBinaries(miopenBinariesBlob_t, const char*, size_t)`

Internally it can be done as
```C++
struct BinariesBlob
{
    std::vector<Program> programs;
};
void to_json(nlohmann::json& json, const BinariesBlob& binaries);
void from_json(const nlohmann::json& json, BinariesBlob& binaries);
```
with binaries serialized from/to programs in load/save. Detaching from solution would replace program handles with indexes in the vector, removing the duplicates. Reverse for attaching.


DrizztDoUrden:
@atamazov This may be interesting for you.

include/miopen/miopen.h
```diff
@@ -4879,6 +4879,15 @@ miopenStatus_t miopenSetFindOptionPreallocatedTensor(miopenFindOptions_t options
...
                                                      void* buffer);
 
+/*! @brief Forces library to attach kernel binaries to solutions for later saving. This allows zero
+ * lookup miopenRunSolution calls after miopenLoadSolution. Default value is false.
+ *
+ * @param options    Options object to update
+ * @param attach     True means attaching, false - skipping
+ * @return           miopenStatus_t
+ */
+miopenStatus_t miopenSetFindOptionAttachBinaries(miopenFindOptions_t options, bool attach);
```
averinevg: `C`'s and `C++`'s bool types may be incompatible.

averinevg: I suggest using a mask to set/unset boolean options. This will allow the API to be extended in the future without adding new functions.
```suggestion
miopenStatus_t miopenSetFindOptionsBool(miopenFindOptions_t options, uint32_t set, uint32_t unset);
```

averinevg (dismissed):

include/miopen/miopen.h
DrizztDoUrden: Good catch! I forget that stuff from time to time.

DrizztDoUrden: I'd rather keep explicit functions for separate options, since they are unrelated. Plus, with the current approach each option can actually be extended to be a bitmask itself, if needed. For example, if we would want to add another mode of attaching binaries, we could just use a value of 2 for that.

averinevg: I'd prefer a simple but powerful API instead of a dozen functions, something like OpenCL API. You could reserve 2 or more bits for this option. Anyway, it's up to you.

pfultz2: > I'd rather keep explicit functions for separate options, since they are unrelated. Plus, with the current approach each option can actually be extended to be a bitmask itself, if needed.

Yes I think the explicit function is much better than using bitmasks.

include/miopen/miopen.h
```diff
@@ -4864,7 +4866,7 @@ miopenStatus_t miopenSetFindOptionWorkspaceLimit(miopenFindOptions_t options, si
  * @param size       Specifies the size of the buffer passed
  * @return           miopenStatus_t
  */
-miopenStatus_t
+MIOPEN_EXPORT miopenStatus_t MIOPEN_EXPORT
```
pfultz2: Why is `MIOPEN_EXPORT` added twice?

DrizztDoUrden: Typo, will fix :)

DrizztDoUrden:
I'll switch it to draft for now, I still have a couple things to fix.

junliume:
@DrizztDoUrden could you follow up on this PR?

DrizztDoUrden:
@junliume yes, sure. It just takes quite some time to merge develop and test everything

junliume:
@DrizztDoUrden the `test_find_2_conv` is failing pretty persistently now

junliume (approved):
Ping @averinevg @pfultz2 for last round of review

averinevg (approved):
LGTM

PR #2154, opened 2023-05-18
fix Tuning fails for ResNet50 

bghimireamd:
This PR is to address the "Tuning fails for ResNet50 (SWDEV-396317)". The reason it was failing was because Fused conv API does not pass user-allocated buffers during compile time. This cause the invoker to fail, since there are no buffer. 
I wrote a function (`MakeConvBiasActivFusionInvokeParams`) that allocates buffer to get rid of the issue.


junliume:
@bghimireamd there are some CI issues please take a look.

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
+    const auto out_buf  = handle.Create(conv_problem.top_sz);
+
+    const auto gfx90aaltimpl = conv_problem.conv_problem.GetConv().attribute.gfx90aFp16alt.GetFwd();
+
+    auto conv_data = std::make_unique<miopen::fusion::ConvolutionOpInvokeParam>(wei_buf.get());
+    auto bias_data = std::make_unique<miopen::fusion::BiasOpInvokeParam>(bias_buf.get());
+
+    const float activ_alpha = static_cast<double>(0.5f);
+    const float activ_beta  = static_cast<double>(0.5f);
+    const float activ_gamma = static_cast<double>(0.5f);
```
averinevg:
```suggestion
    const float activ_alpha = 0.5f;
    const float activ_beta  = 0.5f;
    const float activ_gamma = 0.5f;
```

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
     return miopenStatusSuccess;
 }
 
+void AllocateConvBiasActivFusionInvokerBufffer(const FusionContext& context,
```
averinevg:
```suggestion
void AllocateConvBiasActivFusionInvokerBuffer(const FusionContext& context,
```

src/include/miopen/fusion.hpp
```diff
@@ -254,6 +254,14 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
                                    const ActivationDescriptor& activationDesc,
                                    const TensorDescriptor& yDesc,
                                    Data_t y);
+struct FusionContext;
+struct FusionDescription;
+/// Workaround: Fused conv API does not pass user-allocated buffers during compile time,
+/// This function allocates buffer to get rid of the issue.
+void AllocateConvBiasActivFusionInvokerBufffer(const FusionContext& context,
```
averinevg:
```suggestion
void AllocateConvBiasActivFusionInvokerBuffer(const FusionContext& context,
```

averinevg (changes requested):

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
     return miopenStatusSuccess;
 }
 
+void AllocateConvBiasActivFusionInvokerBuffer(const FusionContext& context,
+                                              const FusionDescription& problem,
+                                              miopen::OperatorArgs& params,
+                                              AnyInvokeParams& fused_invoker)
```
DrizztDoUrden:
```suggestion
std::tuple<miopen::OperatorArgs, AnyInvokeParams> AllocateConvBiasActivFusionInvokerBuffer(const FusionContext& context,
                                              const FusionDescription& problem)
```

Sidenote: AnyInvokeParams is not an invoker

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
```diff
@@ -92,35 +92,13 @@ PerformanceConfigConvBiasActivAsm1x1U ConvBiasActivAsm1x1U::Search(const FusionC
...
-    auto conv_problem    = problem.GetConvProblem(0, conv::Direction::Forward);
-    conv_problem.bias    = 1;
-    conv_problem.bias_sz = static_cast<size_t>(conv_problem.GetOutChannels()) *
-                           ((conv_problem.GetOutDataType() == miopenHalf) ? 2 : 4);
-    const auto conv_ctx = context.GetConvContext(conv_problem);
-
-    if(!conv_problem.direction.IsForward())
-        MIOPEN_THROW("Only inference supported.");
-
+    AnyInvokeParams fused_invoker;
```
DrizztDoUrden: This is not an invoker

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
```diff
@@ -92,35 +92,13 @@ PerformanceConfigConvBiasActivAsm1x1U ConvBiasActivAsm1x1U::Search(const FusionC
...
+    miopen::OperatorArgs params;
     /// Workaround: Fused conv API does not pass user-allocated buffers here,
     /// but we need these buffers for search.
-    auto& handle = conv_ctx.GetStream();
-
-    const auto bias_buf = handle.Create(conv_problem.GetBiasSize());
-    const auto in_buf   = handle.Create(conv_problem.GetInSize());
-    const auto wei_buf  = handle.Create(conv_problem.GetWeightsSize());
-    const auto out_buf  = handle.Create(conv_problem.GetOutSize());
+    AllocateConvBiasActivFusionInvokerBuffer(context, problem, params, fused_invoker);
```
DrizztDoUrden: For the changes suggested above:
```suggestion
    const auto [params, invoke_params] = AllocateConvBiasActivFusionInvokerBuffer(context, problem);
```

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
```diff
@@ -92,35 +92,13 @@ PerformanceConfigConvBiasActivAsm1x1U ConvBiasActivAsm1x1U::Search(const FusionC
...
-    tensors.w                = wei_buf.get();
-    tensors.out              = out_buf.get();
-    tensors.inDesc           = conv_problem.conv_problem.GetIn();
-    tensors.wDesc            = conv_problem.conv_problem.GetWeights();
-    tensors.outDesc          = conv_problem.conv_problem.GetOut();
-    tensors.bias             = bias_buf.get();
-    const auto gfx90aaltimpl = conv_problem.conv_problem.GetConv().attribute.gfx90aFp16alt.GetFwd();
-    const auto fused_invoke_ctx = conv::FusedDataInvokeParams(tensors, nullptr, 0, gfx90aaltimpl);
-    return GenericSearch(*this, context, problem, fused_invoke_ctx);
+    return GenericSearch(*this, context, problem, fused_invoker);
```
DrizztDoUrden:
```suggestion
    return GenericSearch(*this, context, problem, invoke_params);
```

src/fusion.cpp
```diff
@@ -430,6 +474,7 @@ miopenStatus_t FusionPlanDescriptor::Compile(Handle& handle)
     auto fusion_ctx       = FusionContext{handle};
     auto fusion_problem   = FusionDescription{this};
     fusion_ctx.DetectRocm();
+    // tmp_sols is collection of all the solvers that isApplicable for the fusion_problem.
```
DrizztDoUrden: These are `ConvSolution`s, not solvers. Basically, they are instructions on how to build them + invokers.

src/fusion.cpp
```diff
@@ -430,6 +474,7 @@ miopenStatus_t FusionPlanDescriptor::Compile(Handle& handle)
...
     const auto tmp_sols = solvers.SearchForAllSolutions(
         fusion_ctx, fusion_problem, miopen::GetDb(fusion_ctx), AnyInvokeParams{});
     std::vector<miopen::solver::ConvSolution> sols;
```
DrizztDoUrden, line 525: You should filter them by adding direction to your context and checking it in `IsApplicable`, rather than manually filtering find results. This way generic searches that are known to be not needed would not happen every time.

test/gtest/cba_infer.cpp
```diff
@@ -131,6 +148,25 @@ TEST_P(ConvBiasActivInferTestHalf, ConvCKIgemmFwdBiasActivFused)
         fusePlanDesc, plan_params, conv_config, test_skipped);
 }
 
+#if MIOPEN_BACKEND_HIP
+TEST_P(ConvBiasActivInferTestFloatFusionCompileStep, ConvBiasActivAsm1x1UFloat_testCompile)
+{
+    setEnvironmentVariable("MIOPEN_FIND_ENFORCE", "3");
+    fusePlanDesc.Compile(get_handle());
+    const auto plan_params = std::make_unique<miopen::fusion::FusionInvokeParams>(
```
DrizztDoUrden, line 156: Why not just pass it around either by value/reference or as `AnyInvokeParams`? Is there any specific reason `unique_ptr`s are used everywhere for this?
This is out of scope of this PR, I think, but I am just curious, since here we are.

DrizztDoUrden (dismissed):

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
+    params.SetArg(1, std::move(bias_data));
+    params.SetArg(2, std::move(activ_data));
+
+    fused_invoker = miopen::fusion::FusionInvokeParams(params,
+                                                       conv_problem.conv_problem.GetIn(),
+                                                       in_buf.get(),
+                                                       conv_problem.conv_problem.GetOut(),
+                                                       out_buf.get(),
+                                                       gfx90aaltimpl);
+}
```
DrizztDoUrden, line 159:
```suggestion
    return std::tuple{params, invoke_params};
}
```

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
 
+void AllocateConvBiasActivFusionInvokerBuffer(const FusionContext& context,
+                                              const FusionDescription& problem,
+                                              miopen::OperatorArgs& params,
+                                              AnyInvokeParams& fused_invoker)
+{
+    auto conv_problem    = problem.GetConvProblem(0, conv::Direction::Forward);
+    conv_problem.bias    = 1;
+    conv_problem.bias_sz = static_cast<size_t>(conv_problem.n_outputs) *
+                           ((conv_problem.out_data_type == miopenHalf) ? 2 : 4);
```
averinevg:
```suggestion
                           GetTypeSize(conv_problem.out_data_type);
```

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
+
+    if(!conv_problem.direction.IsForward())
+        MIOPEN_THROW("Only inference supported.");
+
+    auto& handle = conv_ctx.GetStream();
+
+    const auto bias_buf = handle.Create(conv_problem.bias_sz);
+    const auto in_buf   = handle.Create(conv_problem.bot_sz);
+    const auto wei_buf  = handle.Create(conv_problem.weights_sz);
+    const auto out_buf  = handle.Create(conv_problem.top_sz);
```
averinevg:
```suggestion
    const auto bias_buf = handle.Create(conv_problem.GetBiasSize());
    const auto in_buf   = handle.Create(conv_problem.GetInSize());
    const auto wei_buf  = handle.Create(conv_problem.GetWeightsSize());
    const auto out_buf  = handle.Create(conv_problem.GetOutSize());
```

averinevg (changes requested):

test/gtest/cba_infer.cpp
bghimireamd: Since the scope of `plan_params` is just within `TEST_P` I agree we can get rid of `unique_ptr`. I will address this in separate PR.

src/fusion.cpp
bghimireamd: When you say "_You should filter them by adding direction to your context and checking it in IsApplicable, rather than manually filtering find results_ " similar to [this](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/solver/conv_asm_1x1u_bias_activ_fused.cpp#L251)?

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
bghimireamd: I am not sure if this works. Since params is basically vector of `unique_ptr` the same `unique_ptr` is also sent to FusionInvokeParams's constructor. When params is moved during return `params` that was passed to FusionInvokeParams gets undefined behavior. This eventually cause segfault.

src/fusion.cpp
bghimireamd: See https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2154#discussion_r1202894995

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
DrizztDoUrden: Two `unique_ptr`s should never have the same pointer owned. You should use `shared_ptr` if there are two or more owners. And `FusionInvokeParams` should then contain a copy of `OperatorArgs` rather than a reference. This would be much safer to use.
What I want to say is that `FusionInvokeParams` should have some form of explicit ownership over `OperatorArgs`. This may be alternatively conveyed by passing `OperatorArgs` around as shared/weak pointers if later changes are supposed to be reflected in invoke params while they are still owned separately.
But it may be fixed in a separate PR since it's not been implemented in this PR anyway.

src/fusion.cpp
DrizztDoUrden: Probably. I am not familiar with fusion types, but any post-find filtering would implicitly consume a lot of time due to redundant solvers find being executed at the first place. But this is out of the scope of PR as well, I guess. Just wanted to left a note here, because this can be fixed later and that would improve host performance on average.

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
```diff
@@ -88,39 +88,12 @@ bool ConvBiasActivAsm1x1U::IsValidPerformanceConfig(
...
-    auto tensors             = FusedConvDataTensors{};
-    tensors.in               = in_buf.get();
-    tensors.w                = wei_buf.get();
-    tensors.out              = out_buf.get();
-    tensors.inDesc           = conv_problem.conv_problem.GetIn();
-    tensors.wDesc            = conv_problem.conv_problem.GetWeights();
-    tensors.outDesc          = conv_problem.conv_problem.GetOut();
-    tensors.bias             = bias_buf.get();
-    const auto gfx90aaltimpl = conv_problem.conv_problem.GetConv().attribute.gfx90aFp16alt.GetFwd();
-    const auto fused_invoke_ctx = conv::FusedDataInvokeParams(tensors, nullptr, 0, gfx90aaltimpl);
```
averinevg, line 122: It seems that we don't need `conv::FusedDataInvokeParams` anymore

averinevg (changes requested):

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
bghimireamd: I do not see anyone using it anymore.

averinevg: @bghimireamd @DrizztDoUrden Let's remove it in this case

src/fusion.cpp
```diff
@@ -430,8 +475,32 @@ miopenStatus_t FusionPlanDescriptor::Compile(Handle& handle)
     auto fusion_ctx       = FusionContext{handle};
     auto fusion_problem   = FusionDescription{this};
     fusion_ctx.DetectRocm();
+    AnyInvokeParams invoke_params;
+    miopen::OperatorArgs params;
+    if(fusion_problem.fusion_plan_desc->op_map.size() >= 3 &&
```
JehandadKhan:
```suggestion
    if(<if we are tuning> && fusion_problem.fusion_plan_desc->op_map.size() >= 3 &&
```

src/fusion.cpp
```diff
@@ -430,8 +475,32 @@ miopenStatus_t FusionPlanDescriptor::Compile(Handle& handle)
...
+    {
+        // Workaround: Fused API does not pass user-allocated buffers,
+        // but we need these buffers during SearchForAllSolutions search.
+        // Since, SearchForAllSolutions invokes kernel launch.
+        AllocateConvBiasActivFusionInvokerBuffer(fusion_ctx, fusion_problem, params, invoke_params);
+        MIOPEN_LOG_I2("Done allocating buffer for conv+bias+activ fusion");
+    }
+    else
+    {
+        // handle the rest of the fusion operators cases
```
JehandadKhan, line 495: MIOPEN_THROW(miopenStatusNotImplemented);

JehandadKhan (dismissed):

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
bghimireamd: @JehandadKhan should I remove `fused_data_invoke_params.hpp` then?

src/fusion.cpp
JehandadKhan: The Fusion operators for forward and backward types are different, and should not be applicable since applicability checks the list of fusions first

src/fusion.cpp
```diff
@@ -108,6 +108,51 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
+void AllocateConvBiasActivFusionInvokerBuffer(const FusionContext& context,
+                                              const FusionDescription& problem,
+                                              miopen::OperatorArgs& params,
+                                              AnyInvokeParams& invoke_params)
+{
+    MIOPEN_LOG_I2("Calling function AllocateConvBiasActivFusionInvokerBuffer");
+    auto conv_problem = problem.GetConvProblem(0, conv::Direction::Forward);
+    conv_problem.bias = 1;
+    conv_problem.bias_sz =
+        static_cast<size_t>(conv_problem.n_outputs) * GetTypeSize(conv_problem.out_data_type);
```
averinevg:
```suggestion
        static_cast<size_t>(conv_problem.GetOutChannels()) * GetTypeSize(conv_problem.GetOutDataType());
```

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
averinevg: @bghimireamd `struct FusedConvDataTensors` can be also removed

src/fusion.cpp
```diff
@@ -422,6 +467,35 @@ static NetworkConfig GetPlanConfig(const FusionContext& fusion_ctx,
     ss << op_config.str();
     return NetworkConfig{ss.str()};
 }
```
averinevg, line 474:
```suggestion
}

```

averinevg (changes requested):

src/fusion.cpp
```diff
@@ -422,6 +467,35 @@ static NetworkConfig GetPlanConfig(const FusionContext& fusion_ctx,
...
+void InitializeBuffer(const FusionContext& fusion_ctx,
```
averinevg, line 470:
```suggestion
static void InitializeBuffer(const FusionContext& fusion_ctx,
```

src/fusion.cpp
```diff
@@ -423,15 +467,53 @@ static NetworkConfig GetPlanConfig(const FusionContext& fusion_ctx,
     return NetworkConfig{ss.str()};
 }
 
+static auto MakeFusionInvokeParams(const FusionContext& fusion_ctx,
+                                   const FusionDescription& fusion_problem,
+                                   miopen::OperatorArgs& params)
+{
+    AnyInvokeParams invoke_params;
```
averinevg: [recommendation]
```suggestion
```

src/fusion.cpp
```diff
@@ -423,15 +467,53 @@ static NetworkConfig GetPlanConfig(const FusionContext& fusion_ctx,
...
+    if(fusion_problem.fusion_plan_desc->op_map.size() == 3 &&
+       (fusion_problem.fusion_plan_desc->op_map[0]->kind() == miopenFusionOpConvForward) &&
+       (fusion_problem.fusion_plan_desc->op_map[1]->kind() == miopenFusionOpBiasForward) &&
+       (fusion_problem.fusion_plan_desc->op_map[2]->kind() == miopenFusionOpActivForward))
+    {
+        // Workaround: Fused API does not pass user-allocated buffers,
+        // but we need these buffers during SearchForAllSolutions.
+        // Since, SearchForAllSolutions invokes kernel launch and kernel launch needs these buffers.
+        invoke_params = MakeConvBiasActivFusionInvokeParams(fusion_ctx, fusion_problem, params);
```
averinevg: [recommendation]
```suggestion
        return MakeConvBiasActivFusionInvokeParams(fusion_ctx, fusion_problem, params);
```

src/fusion.cpp
```diff
@@ -423,15 +467,53 @@ static NetworkConfig GetPlanConfig(const FusionContext& fusion_ctx,
...
+        // eg: Convolution + Bias + BatchNorm + Activation,
+        //     Convolution + BatchNorm + Activation
+        //     Convolution + BatchNorm
+        //     Convolution + Activation
+        //     GEMM + Activation
+        //
+        MIOPEN_LOG_I2("Allocating buffer for given fusion operators is not supported yet.");
+        MIOPEN_THROW(miopenStatusNotImplemented);
+    }
+    return invoke_params;
```
averinevg: [recommendation]
```suggestion
```

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
     return miopenStatusSuccess;
 }
 
+auto MakeConvBiasActivFusionInvokeParams(const FusionContext& context,
```
averinevg:
```suggestion
static auto MakeConvBiasActivFusionInvokeParams(const FusionContext& context,
```

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
+                                         const FusionDescription& problem,
+                                         miopen::OperatorArgs& params)
+{
+    MIOPEN_LOG_I2("Calling function AllocateConvBiasActivFusionInvokerBuffer");
```
averinevg:
```suggestion
    MIOPEN_LOG_I2("Calling function MakeConvBiasActivFusionInvokeParams");
```

src/fusion.cpp
```diff
@@ -423,15 +467,51 @@ static NetworkConfig GetPlanConfig(const FusionContext& fusion_ctx,
...
+{
+    if(fusion_problem.fusion_plan_desc->op_map.size() == 3 &&
+       (fusion_problem.fusion_plan_desc->op_map[0]->kind() == miopenFusionOpConvForward) &&
+       (fusion_problem.fusion_plan_desc->op_map[1]->kind() == miopenFusionOpBiasForward) &&
+       (fusion_problem.fusion_plan_desc->op_map[2]->kind() == miopenFusionOpActivForward))
+    {
+        // Workaround: Fused API does not pass user-allocated buffers,
+        // but we need these buffers during SearchForAllSolutions.
+        // Since, SearchForAllSolutions invokes kernel launch and kernel launch needs these buffers.
+        MIOPEN_LOG_I2("Allocating buffer for conv+bias+activ fusion");
```
averinevg:
```suggestion
        MIOPEN_LOG_I2("Allocating buffers for conv+bias+activ fusion");
```

src/fusion.cpp
```diff
@@ -423,15 +467,51 @@ static NetworkConfig GetPlanConfig(const FusionContext& fusion_ctx,
...
     const auto solvers    = GetFusedSolvers();
     auto fusion_ctx       = FusionContext{handle};
     auto fusion_problem   = FusionDescription{this};
     fusion_ctx.DetectRocm();
+    AnyInvokeParams invoke_params;
+    miopen::OperatorArgs params;
+    const FindEnforce enforce;
+    if(enforce.IsSearch(fusion_ctx))
+        invoke_params = MakeFusionInvokeParams(fusion_ctx, fusion_problem, params);
+    // tmp_sols is collection of all the ConvSolution that isApplicable for the fusion_problem.
```
averinevg:
```suggestion
    // tmp_sols is a collection of ConvSolutions that isApplicable for the fusion_problem.
```

src/include/miopen/fusion.hpp
```diff
@@ -254,6 +254,14 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
                                    const TensorDescriptor& yDesc,
                                    Data_t y);
+struct FusionContext;
+struct FusionDescription;
+/// Workaround: Fused conv API does not pass user-allocated buffers during compile time,
+/// This function allocates buffer to get rid of the issue.
+void AllocateConvBiasActivFusionInvokerBuffer(const FusionContext& context,
+                                              const FusionDescription& problem,
+                                              miopen::OperatorArgs& params,
+                                              AnyInvokeParams& invoke_params);
```
averinevg:
```suggestion
```

src/fusion.cpp
```diff
@@ -423,15 +467,51 @@ static NetworkConfig GetPlanConfig(const FusionContext& fusion_ctx,
...
+    else
+    {
+        // handle the rest of the fusion operators cases
+        // eg: Convolution + Bias + BatchNorm + Activation,
+        //     Convolution + BatchNorm + Activation
+        //     Convolution + BatchNorm
+        //     Convolution + Activation
+        //     GEMM + Activation
+        //
+        MIOPEN_LOG_I2("Allocating buffer for given fusion operators is not supported yet.");
```
averinevg:
```suggestion
        MIOPEN_LOG_I2("Allocating buffers for given fusion operators is not supported yet.");
```

averinevg (changes requested):
Good job

src/include/miopen/fusion.hpp
bghimireamd: Good catch 👍

averinevg (dismissed):
LGTM

averinevg:
@bghimireamd Please fix the description

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
     return miopenStatusSuccess;
 }
 
+static auto MakeConvBiasActivFusionInvokeParams(const FusionContext& context,
+                                                const FusionDescription& problem,
+                                                miopen::OperatorArgs& params)
+{
+    MIOPEN_LOG_I2("Calling function MakeConvBiasActivFusionInvokeParams");
```
atamazov: All the MIOPEN_LOG_* macros print the name of the function. Add to it something useful or nothing. I highly recommend printing the buffer addresses and sizes.

Please also change the level from Info2 to Info.

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
     return miopenStatusSuccess;
 }
 
+static auto MakeConvBiasActivFusionInvokeParams(const FusionContext& context,
```
atamazov:
```suggestion
static auto AllocateBuffersAndMakeConvBiasActivFusionInvokeParams(const FusionContext& context,
```
This variant looks ugly (as it should ;)

src/fusion.cpp
```diff
@@ -108,6 +108,50 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
+
+    if(!conv_problem.direction.IsForward())
+        MIOPEN_THROW("Only inference supported.");
+
+    auto& handle = conv_ctx.GetStream();
+
+    const auto bias_buf = handle.Create(conv_problem.GetBiasSize());
+    const auto in_buf   = handle.Create(conv_problem.GetInSize());
+    const auto wei_buf  = handle.Create(conv_problem.GetWeightsSize());
+    const auto out_buf  = handle.Create(conv_problem.GetOutSize());
```
atamazov: 🔴 These auto objects of type `Allocator::ManageDataPtr` own the buffers and being destroyed at the end of `MakeConvBiasActivFusionInvokeParams`, thus freeing all the buffers with `hipFree()`. Accessing of buffers via `params` after returning from `MakeConvBiasActivFusionInvokeParams` is undefined behavior.

atamazov (changes requested):
UB requires some redesign that might lead to many changes. I'm going to wait for the fix before continuing the review of this PR.

src/fusion.cpp
bghimireamd: Good catch 👍

bghimireamd:
@atamazov fixed UB issue.

junliume:
@bghimireamd if some of the previous comments are resolved, please also mark them as "resolved" thank you!

src/fusion.cpp
```diff
@@ -108,6 +108,59 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
...
+    miopen::OperatorArgs& params)
+{
+    auto conv_problem    = problem.GetConvProblem(0, conv::Direction::Forward);
+    conv_problem.bias    = 1;
+    conv_problem.bias_sz = static_cast<size_t>(conv_problem.GetOutChannels()) *
+                           GetTypeSize(conv_problem.GetOutDataType());
+    const auto conv_ctx = context.GetConvContext(conv_problem);
+
+    if(!conv_problem.direction.IsForward())
+        MIOPEN_THROW("Only inference supported.");
```
averinevg:
```suggestion
```
We already know the direction

src/fusion.cpp
```diff
@@ -108,6 +108,59 @@ miopenStatus_t ConvBiasActivFusion(Handle& handle,
     return miopenStatusSuccess;
 }
 
+static auto AllocateBuffersAndMakeConvBiasActivFusionInvokeParams(
+    const FusionContext& context,
+    const FusionDescription& problem,
+    std::vector<Allocator::ManageDataPtr>& invoke_bufs,
+    miopen::OperatorArgs& params)
```
averinevg, line 115: [F.21: To return multiple “out” values, prefer returning a struct or tuple](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f21-to-return-multiple-out-values-prefer-returning-a-struct-or-tuple)

averinevg (changes requested):

src/fusion.cpp
atamazov: tuples must be destroyed along with their inventors.
stalin_smile_50

bghimireamd: @averinevg I will address this in separate PR.

averinevg (dismissed):

junliume (approved):
CI passed again, all minor revision suggestions will be dealt with in a separate PR (CC: @bghimireamd )

bghimireamd:
> CI passed again, all minor revision suggestions will be dealt with in a separate PR (CC: @bghimireamd )

created issues to address them 
https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2181

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
```diff
@@ -293,10 +293,12 @@ ConvOclDirectFwdLegacyExhaustiveSearch::SearchImpl(const ConvolutionContext& ctx
...
-    const auto& invoke_params = invoke_ctx.CastTo<conv::DataInvokeParams>();
-    const auto bot_ocl_ptr    = invoke_params.tensors.in;
-    const auto top_ocl_ptr    = invoke_params.tensors.out;
-    const auto wei_ocl_ptr    = invoke_params.tensors.w;
+    const auto& invoke_params = invoke_ctx.CastTo<miopen::fusion::FusionInvokeParams>();
+    const auto bot_ocl_ptr    = invoke_params.in;
+    const auto top_ocl_ptr    = invoke_params.out;
+    const auto wei_ocl_ptr =
+        dynamic_cast<miopen::fusion::ConvolutionOpInvokeParam&>(*invoke_params.op_args.params[0])
+            .weights;
```
atamazov, line 301: ⚓

atamazov, line 301: - Revert this change
- In `ConvOclDirectFwdFused::Search()`, construct `DataInvokeParams` instance from `FusionInvokeParams` instance. Pass the former to `ConvOclDirectFwd{}.Search()`.

atamazov (commented):
Proposal for fixing #2239

junliume:
> Proposal for fixing #2239

@bghimireamd @JehandadKhan  please follow up with the proposed fix above.

test/gtest/cba_infer.cpp
```diff
@@ -131,6 +148,25 @@ TEST_P(ConvBiasActivInferTestHalf, ConvCKIgemmFwdBiasActivFused)
...
+        params, input.desc, in_dev.get(), output.desc, out_dev.get(), false);
+    RunTunableSolver<miopen::solver::fusion::ConvBiasActivAsm1x1U>(
```
atamazov, line 158: This won't test tuning because `RunTunableSolver` simply takes the default performance config.

test/gtest/cba_infer.cpp
```diff
@@ -131,6 +148,25 @@ TEST_P(ConvBiasActivInferTestHalf, ConvCKIgemmFwdBiasActivFused)
         fusePlanDesc, plan_params, conv_config, test_skipped);
 }
 
+#if MIOPEN_BACKEND_HIP
+TEST_P(ConvBiasActivInferTestFloatFusionCompileStep, ConvBiasActivAsm1x1UFloat_testCompile)
+{
+    setEnvironmentVariable("MIOPEN_FIND_ENFORCE", "3");
```
atamazov, line 154: Value 3 means that tuning is not performed if the perf-db already contains the performance parameters. For reliability of the test, use `MIOPEN_FIND_ENFORCE=SEARCH_DB_UPDATE`.

The full tuning cycle may take thousands of iterations. Use `MIOPEN_DEBUG_TUNING_ITERATIONS_MAX=5` to limit tuning time in tests.

test/gtest/cba_infer.cpp
atamazov: @bghimireamd Ping

PR #2157, opened 2023-05-22
Merge kdbs by arch

cderb:
Combine kdbs across arch skews.
Reduce redundant kernel storage.


CMakeLists.txt
```diff
@@ -465,7 +465,7 @@ endif()
 function(unpack_db db_bzip2_file)
     set(KERNELS_DIR "${CMAKE_SOURCE_DIR}/src/kernels")
     STRING(REPLACE ".bz2" "" db_file "${db_bzip2_file}")
-    find_program(UNZIPPER lbunzip2 bunzip2)
+    find_program(UNZIPPER bunzip2)
```
JehandadKhan: Why is this being removed ?

junliume (approved):

src/binary_cache.cpp
```diff
@@ -123,6 +123,8 @@ KDb GetDb(const TargetProperties& target, size_t num_cu)
     boost::filesystem::path sys_path = sys_dir / (Handle::GetDbBasename(target, num_cu) + ".kdb");
     if(user_dir.empty())
         user_path = user_dir;
+    if(!boost::filesystem::exists(sys_path))
+        sys_path = sys_dir / (target.DbId() + ".kdb");
```
atamazov, line 127: Why we initialize sys_path with GetDbBasename() and if file does not exist change sys_path to target.DbId()?
What is valid .kdb filename?

junliume: @cderb ping for the above review opinion.

cderb: @atamazov sorry this slipped through. What this is doing is looking for a specific arch_skew.kdb and if it doesn't find that file goes for arch.kdb.

atamazov: @cderb Thanks, Chris. AFAICS this is not just an `arch` but `arch-with-encoded-target-features` (returned by target.DbId). When we open system KDB, we are looking for `arch-with-encoded-target-features_numCUs.kdb` first, then, if not found, repeat the same but without `_numCUs`.  The first variant (with numCUs) is no longer supplied, so this always fails.

🚧  I'll take care about cleanup of this code when time permits.

## [Informative] Projected format of "arch-with-encoded-target-features" (projected)

This is just for information. These formats are produced by `TargetProperties::InitDbId()`

### gfx906 and gfx908
We have different format for these two GPUs for the sake of compatibility with previous versions of databases, which was important during transition period.

```
target ID                target.DbId             Note
-----------------------------------------------------
gfx90X:xnack-            gfx90X_nosramecc        (1)
gfx90X:xnack+            gfx90X_nosramecc_xnack  (1)
gfx90X:sramecc-          gfx90X_nosramecc        (1)
gfx90X:sramecc+          gfx90X             
gfx90X:sramecc-:xnack-   gfx90X_nosramecc        (1)
gfx90X:sramecc-:xnack+   gfx90X_nosramecc_xnack  (1)
gfx90X:sramecc+:xnack-   gfx90X             
gfx90X:sramecc+:xnack+   gfx90X_xnack
-----------------------------------------------------
(1) Currently we do not support detection of SRAMECC state for
these targets and always assume sramecc+. Therefore
these target.DbId values do not appear in practice.
```

### Other targets
```
target ID                target.DbId             Note
-----------------------------------------------------
gfxXXX:xnack-            gfxXXX
gfxXXX:xnack+            gfxXXX_xnack
gfxXXX:sramecc-          gfxXXX
gfxXXX:sramecc+          gfxXXX_sramecc
gfxXXX:sramecc-:xnack-   gfxXXX
gfxXXX:sramecc-:xnack+   gfxXXX_xnack
gfxXXX:sramecc+:xnack-   gfxXXX_sramecc
gfxXXX:sramecc+:xnack+   gfxXXX_sramecc_xnack
```

cderb: Yes, this appears to be in error.

JehandadKhan:
Please move the documentation comment above to the developer guide in wiki.

atamazov:
@JehandadKhan I do not have rights for wiki. And prefer https://github.com/ROCm/MIOpen/labels/specification ticket, it's more flexible. Will do soon, when time permits.

JehandadKhan:
Perhaps Chris can help with that when he gets a chance.

atamazov:
@JehandadKhan @cderb BTW the wiki page about formatting is outdated (fin should be excluded).

src/binary_cache.cpp
atamazov: @cderb @JehandadKhan In fact, the state of SRAMECC is currently ignored in DB names. This is due to a W/A that we have for some bug in the old DKMS (W/A has been introduced in #832):
https://github.com/ROCm/MIOpen/blob/7b36cef67f3154dc3eff7471e335f5a7111f4b17/src/target_properties.cpp#L85-L91

As a result, KDB loading is currently working as expected.

## [Informative] Current format of "arch-with-encoded-target-features" _for all targets_ (including gfx906/908)
```
target ID                target.DbId             Note
-----------------------------------------------------
gfxXXX:xnack-            gfxXXX
gfxXXX:xnack+            gfxXXX_xnack
gfxXXX:sramecc-          gfxXXX
gfxXXX:sramecc+          gfxXXX
gfxXXX:sramecc-:xnack-   gfxXXX
gfxXXX:sramecc-:xnack+   gfxXXX_xnack
gfxXXX:sramecc+:xnack-   gfxXXX
gfxXXX:sramecc+:xnack+   gfxXXX_xnack
```

### Side note about "reported_sramecc"

This is different question, but actually we capture (possibly incorrect) "reported_sramecc" state, because we must use it when we build ASM kernels with COMGR. Otherwise correctness problems with FP16/BF16 on MI200 happened. See https://github.com/ROCm/MIOpen/issues/1257#issuecomment-972820638, bullet (3).

🟢 No immediate action is needed for now.

PR #2161, opened 2023-05-23
tiny GettingStartedFusionAPI docs link fix

gonidelis:
hyperlink wouldn't work with underscore, needs dash instead


junliume:
> hyperlink wouldn't work with underscore, needs dash instead



> hyperlink wouldn't work with underscore, needs dash instead

@gonidelis I suspect this is not the only place, Even in the same file `#compile_fusion` looks suspicious, and there are other files too. The link can be easily tested [here](https://rocmsoftwareplatform.github.io/MIOpen/doc/html/Getting_Started_FusionAPI.html#compile-the-fusion-plan)

gonidelis:
I am closing this one now that I have access rights, since it stems from my own repo.

PR #2166, opened 2023-05-24
[CI][Dockerfile] Update Dockerfile ROCm to 5.6

junliume:
- [ ] Update CI docker to ROCm 5.6.
- [ ] Bump MIOpen version to 3.0 after ROCm 5.7 branch it cut.
- [ ] Bump CK dependency version to include more functionalities.


junliume:
Issues with `cppcoreguidelines-avoid-do-while`
src/include/miopen/errors.hpp:61:5

Updated again and let's run through CI first.

junliume:
Found issue related to https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2233

cmake/EnableCompilerWarnings.cmake
```diff
@@ -96,6 +96,7 @@ else()
                 -Wno-unsafe-buffer-usage
                 -Wno-deprecated-declarations
                 -Wno-shadow-uncaptured-local
+                -Wno-return-type
```
junliume: @atamazov @JehandadKhan @DrizztDoUrden @averinevg @dmikushin 
This line is due to compiler is not parsing the following lines well:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/b73554749188f12984b3acb776bdc065ca1d20ec/src/include/miopen/errors.hpp#L60-L64

Since `MIOPEN_THROW` is used in so many different places, it's challenging to suppress them individually. 
I believe this is a false positive of the compiler being thorough but not too clever, we have `throw` in the above lines but its static check still doubts some paths do not have proper returns.

driver/conv_driver.hpp
```diff
@@ -896,7 +896,7 @@ std::vector<int> ConvDriver<Tgpu, Tref>::GetWeightTensorLengthsFromCmdLine()
         }
     }
 
-    miopenConvolutionMode_t mode;
+    miopenConvolutionMode_t mode = miopenConvolution;
```
junliume: I do not like this way of initializing variable, however, compiler complains about it risking been uninitialized and used, probably due to Line 910 below is not properly parsed and understood by the compiler.

driver/conv_driver.hpp
```diff
@@ -1833,7 +1833,7 @@ void ConvDriver<Tgpu, Tref>::GetSolutionAfterFind(
     miopenConvSolution_t& solution)
 {
     AutoMiopenWarmupMode warmupMode; // Shut logging.
-    miopenConvAlgorithm_t found_algo;
+    miopenConvAlgorithm_t found_algo = static_cast<miopenConvAlgorithm_t>(found.fwd_algo);
```
junliume: same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2166/files#r1248135803

src/convolution.cpp
```diff
@@ -290,7 +290,7 @@ ConvolutionDescriptor::GetForwardOutputTensorWithLayout(const TensorDescriptor&
         }
     }
 
-    std::size_t out_c;
+    std::size_t out_c = wei_k;
```
junliume: same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2166/files#r1248135803
Hopefully we have better way to workaround this issue? @atamazov

cmake/EnableCompilerWarnings.cmake
DrizztDoUrden: We could try wrapping its content in a function and adding [[noreturn]]. I didn't try, but theoretically should work.

driver/conv_driver.hpp
DrizztDoUrden: If you want an explicitly invalid value you can do this:
```suggestion
    miopenConvolutionMode_t mode = (miopenConvolutionMode_t )-1;
```

cmake/EnableCompilerWarnings.cmake
DrizztDoUrden: I mean an approach like this:
```C++
template <class... Params>
[[noreturn]] void MiopenThrow(const std::string& file, int line, Params&&... args)
{
	throw miopen::Exception(std::forward<Params>(args)...).SetContext(file, line);
}

#define MIOPEN_THROW(...) MiopenThrow(__FILE__, __LINE__, __VA_ARGS__)
```

src/mlir_build.cpp
```diff
@@ -53,6 +53,7 @@ class AutoMiirHandle
 
 void check_miir_error(MiirStatus status, const std::string& miir_fn_name)
 {
+    (void)miir_fn_name; // WA till C++17 [[maybe_unused]]
```
DrizztDoUrden: We already are C++17. Maybe we should use the attribute?
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/b73554749188f12984b3acb776bdc065ca1d20ec/CMakeLists.txt#L109

src/ocl/tensorocl.cpp
```diff
@@ -1896,6 +1896,10 @@ void CopyTensor(const Handle& handle,
     }
 }
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wimplicit-fallthrough"
```
DrizztDoUrden: We can use [`[[fallthrough]]`](https://en.cppreference.com/w/cpp/language/attributes/fallthrough)

src/mlir_build.cpp
atamazov: [R] Just remove or comment out the name of the unused parameter
```C++
void check_miir_error(MiirStatus status, const std::string& /*miir_fn_name*/)
```

driver/conv_driver.hpp
junliume: It will bring another error:
> error: use of old-style cast [-Werror,-Wold-style-cast]

src/sqlite_db.cpp
```diff
@@ -207,13 +207,15 @@ SQLite::result_type SQLite::Exec(const std::string& query) const
     SQLite::result_type res;
     MIOPEN_LOG_T(std::this_thread::get_id() << ":" << query);
     {
+        // NOLINTBEGIN(clang-analyzer-core.CallAndMessage)
```
junliume, line 217: @atamazov could you try to fix this? It is complaining about Line-213

src/sqlite_db.cpp
```diff
@@ -78,7 +78,14 @@ class SQLite::impl
     {
         sqlite3* ptr_tmp = nullptr;
         int rc           = 0;
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wcast-function-type-strict"
+#endif
         sqlite3_auto_extension(reinterpret_cast<void (*)(void)>(miopen_sqlite3_memvfs_init));
```
junliume, line 85: @JehandadKhan another location where it might be risky to begin with.

src/sqlite_db.cpp
atamazov: @junliume I guess you mean line 213 in the _original_ text, i.e. passing an address of `find_callback`. I highly recommend moving NOLINTBEGIN/END just aroung the guilty line.

This looks like a false positive because the context is very complex: it's like passing an address of `find_callback` to another function, which is called from lambda passed to some 3rd function and invoked by it. And analyzer incorrectly decides that `find_callback` address maybe uninialized at the moment when used.

We can try to simplify things for the analyzer. E.g. introduce some intermediate static variable in the `SQLite` class, e.g. `find_callback_address`, initialize it with the address of `find_callback` and use it in the lambda instead of `find_callback`. You may also need to capture it by-value.

I can help more when I am back.

junliume: [error.log](https://github.com/ROCmSoftwarePlatform/MIOpen/files/12147572/error.log)
Hi @atamazov thanks! It is actually complaining about this line:
```
query.c_str()
```
which is strange and also might be false alarm.

atamazov: @junliume then it is likely that analyzer is unable to track the actual lifetime of `query`. To check this hypothesis, just pass it to lambda by-value.

This is false alarm, UNLESS `sqlite3_exec` is doing something unusual, like putting the db access request into some queue, so actual execution may be postponed.

junliume:
@JehandadKhan @atamazov  @DrizztDoUrden  @averinevg 
Branch is updated and the CI has passed again, could you re-review so that we can merge this one? Thanks!

JehandadKhan (approved):

atamazov (commented):
LGTM!

src/include/miopen/errors.hpp
```diff
@@ -57,11 +57,18 @@ struct Exception : std::exception
...
+[[noreturn]] void MIOpenThrow(const std::string& file, int line, Params&&... args)
+{
+    throw miopen::Exception(std::forward<Params>(args)...).SetContext(file, line);
+}
+
+#define MIOPEN_THROW(...)                                     \
+    do                                                        \
+    {                                                         \
+        miopen::MIOpenThrow(__FILE__, __LINE__, __VA_ARGS__); \
+    } while(false);
```
atamazov, line 70: :anchor:

atamazov: @junliume sorry I missed it.

Issue #1008, opened 2021-06-24
[perf-db] None records for tunable ConvBinWinogradRxSF3x2


atamazov:
`ConvBinWinogradRxSF3x2` is tunable but there is none perf-db records for them, for gfx906/64 device at least. This leads to `test_conv_embed_db` failures when in runs on this target.

Basically, tuning for this solver should not be very important for performance. We have to decide if it is necessary. I see two options:
- Make `ConvBinWinogradRxSF3x2` non-tunable.
- Or, check which targets do not have perf-db records for `ConvBinWinogradRxSF3x2` and then tune for them.

Originated from https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1006#issue-929628160

atamazov:
I'll create a workaround for this.

atamazov:
W/A merged in #1007.

atamazov:
The issue no longer appears. W/A removed by #2179.

atamazov:
Due to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2179#issuecomment-1584011146

PR #2168, opened 2023-05-25
test_conv_embed_db (ctest -> gtest)

alexandraBara:
Moving part of the conv_2d test to gtest.
Test is automatically picked up by gtest.


test/gtest/conv_2d.hpp
```diff
@@ -0,0 +1,66 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2019 Advanced Micro Devices, Inc.
```
JehandadKhan:
```suggestion
 * Copyright (c) 2023 Advanced Micro Devices, Inc.
```

test/gtest/conv_2d_implicit_gemm.cpp
```diff
@@ -0,0 +1,241 @@
...
+#endif
+};
+
+std::vector<TestCase> GetTestCases(const std::string& precision)
+{
+    const std::vector<TestCase> test_cases = {
+        // clang-format off
+    std::make_tuple<std::vector<std::string>, std::string>(
+        {"MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2=0", "MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_V4R1=0"},
+        precision + " --disable-validation --verbose --input 128 1024 14 14 --weights 2048 1024 1 1 --pads_strides_dilations 0 0 2 2 1 1"),
```
JehandadKhan: @alexandraBara  Please turn off formatting, so that each case two lines, one for the env vars and the other for the  test case itself.

test/gtest/conv_2d_implicit_gemm.cpp
```diff
@@ -0,0 +1,241 @@
...
+    std::make_tuple<std::vector<std::string>, std::string>(
+        {"MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2=0", "MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_FWD_V4R1=0"},
```
JehandadKhan: The env vars need to be factored out into constants to reduce the amount of lines, since there are only three unique combinations

test/gtest/conv_2d_implicit_gemm.cpp
```diff
@@ -0,0 +1,241 @@
...
+    GTEST_SKIP();
+#endif
+};
+
+std::vector<TestCase> GetTestCases(const std::string& precision)
+{
+    const std::vector<TestCase> test_cases = {
+        // clang-format off
+    std::make_tuple<std::vector<std::string>, std::string>(
+        {"MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2=0", "MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_V4R1=0"},
```
shurale-nkn: these environment variables were used as WA for issue #874, we should be able to disable all this WA without changing the test code.

shurale-nkn (dismissed):

JehandadKhan (dismissed):

test/CMakeLists.txt
```diff
@@ -760,41 +760,6 @@ if(${MIOPEN_TEST_WITH_MIOPENDRIVER})
...
-# ./bin/MIOpenDriver conv -n 128 -c 1024 -H 14 -W 14 -k 2048 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 1 -t 1
-# MIOPEN_DEBUG_CONV_IMMED_FALLBACK=0
-if(MIOPEN_EMBED_DB)
-    set(MIOPEN_EMBED_TEST_ARG ${MIOPEN_TEST_FLOAT_ARG} --disable-validation --verbose)
-    # WORKAROUND for issue #874
-    set(MIOPEN_WA_ISSUE_874_F  MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_FWD_V4R1=0)
-    set(MIOPEN_WA_ISSUE_874_W  MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_V4R1=0)
-    set(MIOPEN_WA_ISSUE_874_FW MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_FWD_V4R1=0 MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_V4R1=0)
-    # WORKAROUND for issue #1008
-    set(MIOPEN_WA_ISSUE_1008 MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2=0)
```
atamazov: @alexandraBara We don't need these workarounds anymore?

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,243 @@
+#include <tuple>
+
+#include <miopen/miopen.h>
+#include <gtest/gtest.h>
+#include "conv_2d.hpp"
+#include "get_handle.hpp"
+
+using TestCase = std::tuple<std::vector<std::string>, std::string>;
+
+enum class Precision
```
atamazov: [R] Reuse `miopenDataType_t` from miopen.h

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+std::string GetFloatArg()
+{
+    static const auto tmp = std::getenv("MIOPEN_TEST_FLOAT_ARG");
+    if(tmp == nullptr)
+    {
+        return "";
+    }
+    return tmp;
+};
```
atamazov: Reuse stuff from <miopen/env.hpp>

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,243 @@
...
+{
+
+    std::vector<std::string> winograd      = {"MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2"};
+    std::vector<std::string> igemm_wrw     = {"MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2",
+                                          "MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_4R1"};
+    std::vector<std::string> igemm_fwd     = {"MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2",
+                                          "MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_FWD_4R1"};
+    std::vector<std::string> igemm_fwd_wrw = {"MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2",
+                                              "MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_FWD_4R1",
+                                              "MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_4R1"};
```
atamazov: You can add `=0` everywhere and get rid of the the `GetEnvVars` function

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,243 @@
...
+        EXPECT_FALSE(capture.find("Perf Db: record not found") != std::string::npos);
+    }
+};
+
+TEST_P(Conv2dFloat, FloatTest)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle = get_handle();
+    if(!miopen::StartsWith(handle.GetDeviceName(), "gfx906") || GetFloatArg() != "--float")
```
atamazov: The original test is enabled for all GPUs except GFX908 and GFX90A.

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,243 @@
...
+{
+    Float,
+    Half,
+    Int8,
+    BFloat16
+};
+
+std::string GetFloatArg()
+{
+    static const auto tmp = std::getenv("MIOPEN_TEST_FLOAT_ARG");
```
atamazov: This makes the test dependent on the environment variable provided by CTest. I think this is acceptable for now, but this is a workaround. What behavior we need want if test is run independently of CTest?

/cc @JehandadKhan

atamazov (changes requested):
@alexandraBara Sorry if I am too soon but I hopt you'll find this useful.

test/gtest/conv_2d_implicit_gemm.cpp
atamazov: @alexandraBara Also you can split lines manually by means of EOL comments, and keep formatting enabled:
```C++
  some text //
    you want to split //
    manually;
```

test/CMakeLists.txt
atamazov: @alexandraBara Maybe we really do not need these. Let me check.

atamazov:
@alexandraBara BTW Is it really a draft or is it ready for review?

test/gtest/conv_2d_implicit_gemm.cpp
xinlipn: > these environment variables were used as WA for issue #874, we should be able to disable all this WA without changing the test code.

@shurale-nkn , I'm working on the PR now. Do you mean MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2 and MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_V4R1 are dated and can be removed?

MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2 is also used here:
_src/solver/conv_winoRxS.cpp_

MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_V4R1 here
_src/solver/conv_hip_implicit_gemm_fwd_v4r1.cpp_

test/gtest/conv_embed_db.cpp
JehandadKhan: This was added, specifically to accomodate the current test design which relies on CMake variables to determine which precisions should and should not run. By default ( ie when no vars are set) all the precisions should run.

JehandadKhan: This was added, specifically to accomodate the current test design which relies on CMake variables to determine which precisions should and should not run. By default ( ie when no vars are set) all the precisions should run.

atamazov: @JehandadKhan @alexandraBara 
> By default ( ie when no vars are set) all the precisions should run.

But this is not implemented yet. Shall we do it right here or let's postpone?

test/gtest/conv_2d_implicit_gemm.cpp
shurale-nkn: No! This is workaround for this issue. Issue is still open. You must save the reference to this issue in your code so that it can be quickly found and deleted later. Also, the person who will fix this error should be able to quickly disable this WA in order to test MIOpen without it.

xinlipn: > Also, the person who will fix this error should be able to quickly disable this WA in order to test MIOpen without it.

@shurale-nkn , brought back GetEnvVars() which will enable/disable the WA with the env vars e.g. MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F3X2, MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_WRW_V4R1

test/gtest/conv_2d_implicit_gemm.cpp
xinlipn: > The env vars need to be factored out into constants to reduce the amount of lines, since there are only three unique combinations

@JehandadKhan , shortened the env vars

test/CMakeLists.txt
xinlipn: > @alexandraBara We don't need these workarounds anymore?

@atamazov , confimed with @shurale-nkn , still need these workarounds for now

test/gtest/conv_embed_db.cpp
xinlipn: @atamazov , GetEnvVars() takes const std::vector<std::string>& as input, and append "=0" at the end of each env var that could be found. From what I can see, the only reusable code in env.hpp is miopen::GetEnv() to replace std::getenv() here

test/gtest/conv_2d_implicit_gemm.cpp
xinlipn: @shurale-nkn , this WA has been lift. Here's the PR
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2179

Could you resolve this conversation if this answers your questions?

test/CMakeLists.txt
atamazov: @xinlipn #2179 is merged in

atamazov:
@JehandadKhan Hmm... Do we really want to move `test_conv_embed_db` (which should be run only when we embed databases) under the gtest's jurisdiction?

atamazov:
@alexandraBara Please rename this PR. It moves `test_conv_embed_db`.

test/CMakeLists.txt
```diff
@@ -760,41 +760,6 @@ if(${MIOPEN_TEST_WITH_MIOPENDRIVER})
     )
 endif()
 
-# ./bin/MIOpenDriver conv -n 128 -c 1024 -H 14 -W 14 -k 2048 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 1 -t 1
-# MIOPEN_DEBUG_CONV_IMMED_FALLBACK=0
-if(MIOPEN_EMBED_DB)
```
atamazov, line 780: @JehandadKhan [Notice] just in case - it seems that this code is working, but actually it should be `if(NOT MIOPEN_EMBED_DB STREQUAL "")` here.

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,230 @@
...
+        // clang-format on
+    };
+
+    return test_cases;
+}
+
+INSTANTIATE_TEST_SUITE_P(Conv2dGroup, Conv2dFloat, testing::Values(GetTestCases("--float")));
+INSTANTIATE_TEST_SUITE_P(Conv2dGroup, Conv2dHalf, testing::Values(GetTestCases("--half")));
+INSTANTIATE_TEST_SUITE_P(Conv2dGroup, Conv2dInt8, testing::Values(GetTestCases("--int8")));
+INSTANTIATE_TEST_SUITE_P(Conv2dGroup, Conv2dBFloat16, testing::Values(GetTestCases("--bfloat16")));
```
atamazov: Test names should reflect that the test is intended for a library configuration with embedded databases.

atamazov (changes requested):
~The PR introduces _quiet change_: It enables the test when `MIOPEN_EMBED_DB` is empty.~

test/CMakeLists.txt
xinlipn: > @JehandadKhan [Notice] just in case - it seems that this code is working, but actually it should be `if(NOT MIOPEN_EMBED_DB STREQUAL "")` here.

@atamazov , I have been thinking if this line would be necessary in this case

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/29afc6d0d4d98a2e821a420ac3613d74d51b9b9b/test/gtest/conv_embed_db.cpp#L105

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,215 @@
...
+        testing::internal::CaptureStderr();
+        test_drive<conv2d_driver>(ptrs.size(), ptrs.data());
+        auto capture = testing::internal::GetCapturedStderr();
+        EXPECT_FALSE(capture.find("Perf Db: record not found") != std::string::npos);
+    }
+};
+
+TEST_P(ConfigWithFloat, FloatTest)
+{
+#if MIOPEN_EMBED_DB
```
JehandadKhan, line 113: (anchor)

JehandadKhan:
> The PR introduces _quiet change_: It enables the test when `MIOPEN_EMBED_DB` is empty.

@atamazov  Please [see here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168/files#r1223184970) which guards this test so that it only runs when `MIOPEN_EMBED_DB` is enabled.

atamazov:
@JehandadKhan Oops, I see now. But https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168#discussion_r1222130686 remains still.

Please also look at Q at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168#issuecomment-1581458532

xinlipn:
> @JehandadKhan Oops, I see now. But [#2168 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168#discussion_r1222130686) remains still.

@atamazov , I have updated test names as below. Please note no underscore is allowed in gTest names
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2b656e53fd217526dab15e307f9e99c8dc045adb/test/gtest/conv_embed_db.cpp#L210

test/CMakeLists.txt
xinlipn: > @xinlipn #2179 is merged in

@atamazov Could you resolve this conversation if this has been resolved?

xinlipn:
> @JehandadKhan Hmm... Do we really want to move `test_conv_embed_db` (which should be run only when we embed databases) under the gtest's jurisdiction?

@atamazov , indeed this test should fail without embed database according to this line in 

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2b656e53fd217526dab15e307f9e99c8dc045adb/test/CMakeLists.txt#L522

This mechanism has been integrated here.
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2b656e53fd217526dab15e307f9e99c8dc045adb/test/gtest/conv_embed_db.cpp#L75

Could you resolve this conversation if this has been resolved?

xinlipn:
> @JehandadKhan Hmm... Do we really want to move `test_conv_embed_db` (which should be run only when we embed databases) under the gtest's jurisdiction?

@atamazov , indeed this test should fail without embed database according to this line in 

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2b656e53fd217526dab15e307f9e99c8dc045adb/test/CMakeLists.txt#LL522C1-L522C102

This mechanism has been integrated here.
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2b656e53fd217526dab15e307f9e99c8dc045adb/test/gtest/conv_embed_db.cpp#L75

Could you resolve this conversation if this has been resolved?

test/gtest/conv_embed_db.cpp
xinlipn: > Test names should reflect that the test is intended for a library configuration with embedded databases.

@atamazov  Test names have been updated. Could you resolve conversation if this has been resolved?

xinlipn:
> @JehandadKhan Hmm... Do we really want to move `test_conv_embed_db` (which should be run only when we embed databases) under the gtest's jurisdiction?

@atamazov , indeed this test should fail without embed database according to this line in

(https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2b656e53fd217526dab15e307f9e99c8dc045adb/test/CMakeLists.txt#LL522C1-L522C102)

This mechanism has been integrated here.

[MIOpen/test/gtest/conv_embed_db.cpp](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2b656e53fd217526dab15e307f9e99c8dc045adb/test/gtest/conv_embed_db.cpp#L75)

Line 75 in [2b656e5](https://github.com/ROCmSoftwarePlatform/MIOpen/commit/2b656e53fd217526dab15e307f9e99c8dc045adb)
 EXPECT_FALSE(capture.find("Perf Db: record not found") != std::string::npos); 

Could you resolve this conversation if this has been resolved?

xinlipn:
> Please also look at Q at [#2168 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168#issuecomment-1581458532)

@atamazov I have replied as above. Thanks

test/gtest/conv_embed_db.cpp
xinlipn: @atamazov , could you resolve conversation if this has been addressed?

test/CMakeLists.txt
atamazov: [Resolved]

Note that I do no have enough permissions to mark comment thread as "resolved".

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_2d.hpp
```diff
@@ -0,0 +1,66 @@
...
+        this->add(this->in_layout, "in_layout", this->generate_data({"NCHW"}));
+        this->add(this->fil_layout, "fil_layout", this->generate_data({"NCHW"}));
+        this->add(this->out_layout, "out_layout", this->generate_data({"NCHW"}));
+        this->add(this->deterministic, "deterministic", this->generate_data({false}));
+        this->add(this->tensor_vect, "tensor_vect", this->generate_data({0}));
+        this->add(this->vector_length, "vector_length", this->generate_data({1}));
+        // Only valid for int8 input and weights
+        this->add(this->output_type, "output_type", this->generate_data({"int32"}));
+        this->add(this->int8_vectorize, "int8_vectorize", this->generate_data({false}));
+    }
```
atamazov: This is partial copy of test/conv2d.cpp. Please do not make copies of existing code, refactor instead.

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,215 @@
...
+       miopen::StartsWith(handle.GetDeviceName(), "gfx103") || // Implicitly disabled by default
+       miopen::StartsWith(handle.GetDeviceName(), "gfx110") || // Implicitly disabled by default
+       GetFloatArg() != "--float")
+    {
+        GTEST_SKIP();
+    }
+    else
+    {
+        Run2dDriver(miopenFloat);
+    }
```
atamazov: [R] This code reproduces the details of implementation of test/CMakeLists.txt.

AFAIU this better reflects what the developer of the test wants (he/she wants to test on gfx906/900 only):
```suggestion
    if((handle.GetDeviceName() == "gfx906" || handle.GetDeviceName() == "gfx900") //
       && GetFloatArg() == "--float")
    {
        Run2dDriver(miopenFloat);
    }
    else
    {
        GTEST_SKIP();
    }
```

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,215 @@
...
+    GTEST_SKIP();
+#endif
+};
+
+TEST_P(ConfigWithHalf, HalfTest)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle = get_handle();
+    if(miopen::StartsWith(handle.GetDeviceName(), "gfx908") || // Explicitly disabled
```
atamazov: [R] Ditto

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,215 @@
...
+    GTEST_SKIP();
+#endif
+};
+
+TEST_P(ConfigWithInt8, Int8Test)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle = get_handle();
+    if(miopen::StartsWith(handle.GetDeviceName(), "gfx908") || // Explicitly disabled
```
atamazov: [R] Ditto

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,215 @@
...
+    GTEST_SKIP();
+#endif
+};
+
+TEST_P(ConfigWithBFloat16, BFloat16Test)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle = get_handle();
+    if(miopen::StartsWith(handle.GetDeviceName(), "gfx908") || // Explicitly disabled
```
atamazov: [R] Ditto

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,215 @@
...
+#endif
+};
+
+std::vector<std::string> GetTestCases(const std::string& precision)
+{
+#if MIOPEN_EMBED_DB
+    std::string flags = " --disable-validation --verbose ";
+#else
+    std::string flags = " ";
+#endif
```
atamazov: [R] conditional compilation is not necessary here.
```suggestion
    const std::string flags = " --disable-validation --verbose ";
```

atamazov (changes requested):
Some changes are still needed.

atamazov:
@xinlipn "[R]" means "Recommendation", "[Q]" is "Question".

test/gtest/conv_2d_implicit_gemm.cpp
atamazov: [Resolved]

test/gtest/conv_2d.hpp
atamazov: [notice] Copying the code is the worst thing that we can do from the maintenance point of view.

atamazov:
@xinlipn Please resolve conflicts (no force-pushing please) and ping me when ready for the next review round (== you think that PR is ready for merging).

test/gtest/conv_2d.hpp
xinlipn: > [notice] Copying the code is the worst thing that we can do from the maintenance point of view.

Refactored by moving common code to conv_common.hpp

test/gtest/conv_embed_db.cpp
xinlipn: Here's the original test, GFX908 and GFX90A are disabled explicitly. 
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2497e64ce5fa0ee562a9fdadc19273de86f6d618/test/CMakeLists.txt#L773

Yet, this list is not complete because of these default setting, GFX94X_DISABLED, GFX103X_DISABLED, GFX110X_DISABLED. Thus, the comments "Implicitly disabled by default".

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/2b656e53fd217526dab15e307f9e99c8dc045adb/test/CMakeLists.txt#L518

test/gtest/conv_embed_db.cpp
xinlipn: > [R] Ditto

Please see above

test/gtest/conv_embed_db.cpp
xinlipn: > [R] Ditto

Same as above

test/gtest/conv_embed_db.cpp
xinlipn: > [R] Ditto

Same as above

test/gtest/conv_2d.hpp
atamazov: This is 2D-specific and should not be moved there.

test/gtest/conv_embed_db.cpp
atamazov: The code should reflect the intent of the developer wherever possible.

test/gtest/conv_2d.hpp
xinlipn: > This is 2D-specific and should not be moved there.

Reverted and re-refactored conv2d_driver related changes. This may look out of the place since the header is in test/gtest and this file is in test/. All previous conv2d CTest cases may be refactored by using this header as we continue moving toward gTest,

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/d087a00396879dededb1bc0b13409d85060e7fc7/test/conv2d.cpp#L26

test/gtest/conv_embed_db.cpp
xinlipn: > The code should reflect the intent of the developer wherever possible.

I was a bit rushing to make that comment. Your proposed change is indeed cleaner.

xinlipn:
> @xinlipn Please resolve conflicts (no force-pushing please) and ping me when ready for the next review round (== you think that PR is ready for merging).

@atamazov , CI is still running "Performance Tests - gfx90a", which should be a residual issue and has supposedly been disabled. I believe I have addressed your comments. Could you help review? Thanks

test/gtest/conv_2d.hpp
atamazov: @xinlipn Yeah, `test/` should not depend on `test/gtest/` (Right now we are reusing code from `test/`.)

**_Please rename this header to test/conv2d.hpp._**

> All previous conv2d CTest cases may be refactored by using this header as we continue moving toward gTest,

Moving code from test/ to test/gtest/ is a separate task. Let's do that later, separately.

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_2d.hpp
```diff
@@ -0,0 +1,66 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include "../conv_common.hpp"
```
atamazov: Please add guard (proposed name: `GUARD_TEST_CONV2D_HPP`)

atamazov (changes requested):
Partial review, more info a bit later.

atamazov:
@xinlipn 
> CI is still running "Performance Tests - gfx90a", which should be a residual issue and has supposedly been disabled.

Maybe. We need a kind of approval from @junliume that this PR has passed CI tests.

atamazov:
@xinlipn BTW testing is not covered in the PR description. I means testing of the test itself, does it perform as expected.

Have you run this test with MIOPEN_LOG_LEVEL=5 and checked if the log matches your expectations?

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,196 @@
+#include <tuple>
```
atamazov, line 26: Please add license header

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include <miopen/miopen.h>
+#include <gtest/gtest.h>
+#include <miopen/miopen.h>
+#include <miopen/env.hpp>
+#include "conv_2d.hpp"
+#include "get_handle.hpp"
+
+std::string GetFloatArg()
+{
+    static const auto tmp = miopen::GetEnv("MIOPEN_TEST_FLOAT_ARG");
```
atamazov: Please reuse env.hpp from the library.

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,196 @@
...
+};
+class ConfigWithInt8 : public testing::TestWithParam<std::vector<std::string>>
+{
+};
+class ConfigWithBFloat16 : public testing::TestWithParam<std::vector<std::string>>
+{
+};
+class ConfigWithFloat : public testing::TestWithParam<std::vector<std::string>>
+{
+};
```
atamazov, line 64: Why do we need four identical classes with different names?

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,196 @@
...
+        EXPECT_FALSE(capture.find("Perf Db: record not found") != std::string::npos);
+    }
+};
+
+TEST_P(ConfigWithFloat, FloatTest)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle = get_handle();
+    if((handle.GetDeviceName() == "gfx900" || handle.GetDeviceName() == "gfx906") &&
```
atamazov: Factor out device checks to a function.
```suggestion
    if(IsTestSupportedForDevice(handle) &&
```

atamazov (changes requested):
Some final review comments.

test/gtest/conv_embed_db.cpp
xinlipn: > Why do we need four identical classes with different names?

@atamazov , there are four instantiations in gTest corresponding to different precision tests, i.e. float, half, int8, and bfloat16. 
The code is structured around the instantiations to map how ctest runs. Each instantiation would need a different class. I'm sure there are other ways of doing this. Hope this makes sense

test/gtest/conv_embed_db.cpp
xinlipn: Done

test/gtest/conv_embed_db.cpp
xinlipn: Done

test/gtest/conv_embed_db.cpp
xinlipn: Done

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,229 @@
...
+    switch(prec)
+    {
+    case miopenFloat: params = ConfigWithFloat::GetParam(); break;
+    case miopenHalf: params = ConfigWithHalf::GetParam(); break;
+    case miopenInt8: params = ConfigWithInt8::GetParam(); break;
+    case miopenBFloat16: params = ConfigWithBFloat16::GetParam(); break;
+    case miopenInt8x4:
+    case miopenInt32:
+    case miopenDouble:
+        MIOPEN_THROW(miopenStatusBadParm,
```
atamazov: Please do not use MIOPEN_THROW in tests. This one is for library.

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_2d.hpp
atamazov: Well I actually prefer macros (due to flexibility and portability) but this is fine. [Resolved]

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,229 @@
...
+    else
+        return false;
+}
+
+TEST_P(ConfigWithFloat, FloatTest)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle = get_handle();
+    const auto& envVar = miopen::GetEnv("MIOPEN_TEST_FLOAT_ARG");
```
atamazov: Please use miopen::GetStringEnv() as it implements caching. Example:
```C++
MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_FLOAT_ARG)
...
static std::string GetFloatArg()
{
    const char* p = miopen::GetStringEnv(MIOPEN_TEST_FLOAT_ARG{});
    if(p == nullptr)
    {
        // Fail the test (preferred) or return "--float" by default.
    }
    return {p};
}
```

atamazov (changes requested):

test/gtest/conv_embed_db.cpp
atamazov: just fail the test

test/gtest/conv_embed_db.cpp
atamazov: This matches the proven pattern that we use in MIOpen when we need to read a string.

atamazov:
@xinlipn 
Please note the following comments:
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168#discussion_r1211964117
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168#discussion_r1211980597

These were hidden as resolved which is not actually so. Due to that I had do duplicate some review comments. Let's hide review comments after they explicitly marked `[Resolved]`, if possible.

Please also provide an answer to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168#issuecomment-1602820706. Thank you!

xinlipn:
> @xinlipn BTW testing is not covered in the PR description. I means testing of the test itself, does it perform as expected.
> 
> Have you run this test with MIOPEN_LOG_LEVEL=5 and checked if the log matches your expectations?

Will upload

> @xinlipn BTW testing is not covered in the PR description. I means testing of the test itself, does it perform as expected.
> 
> Have you run this test with MIOPEN_LOG_LEVEL=5 and checked if the log matches your expectations?

@atamazov , CTest and gTest cover the same test cases, after having it reviewed by JD, there are a few issues, e.g. warning message that test_drive was called more than once. Have been working on them...

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+TEST_P(ConfigWithFloat, FloatTest)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle         = get_handle();
+    const char* const p_envVar = miopen::GetStringEnv(MIOPEN_TEST_FLOAT_ARG{});
+
+    if(IsTestSupportedForDevice(handle) &&
+       (p_envVar != nullptr && std::strcmp(p_envVar, "--float") == 0))
```
atamazov: [R] This can be factored out to a function, like this:
```suggestion
    if(IsTestSupportedForDevice(handle) && IsTestRunWith("--float"))
```
And of course you do not need the following definition above
```C++
const char* const p_envVar = miopen::GetStringEnv(MIOPEN_TEST_FLOAT_ARG{});
```

Proposed function:
```C++
static bool IsTestRunWith(const char* float_arg)
{
    assert(float_arg != nullptr);
    const char* const p = miopen::GetStringEnv(MIOPEN_TEST_FLOAT_ARG{})
    return p != nullptr && std::strcmp(p, float_arg) == 0;
}
```

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+TEST_P(ConfigWithHalf, HalfTest)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle         = get_handle();
+    const char* const p_envVar = miopen::GetStringEnv(MIOPEN_TEST_FLOAT_ARG{});
+
+    if(IsTestSupportedForDevice(handle) &&
+       (p_envVar != nullptr && std::strcmp(p_envVar, "--half") == 0))
```
atamazov: [R] Ditto

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+TEST_P(ConfigWithInt8, Int8Test)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle         = get_handle();
+    const char* const p_envVar = miopen::GetStringEnv(MIOPEN_TEST_FLOAT_ARG{});
+
+    if(IsTestSupportedForDevice(handle) &&
+       (p_envVar != nullptr && std::strcmp(p_envVar, "--int8") == 0))
```
atamazov: [R] Ditto

test/gtest/conv_embed_db.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+TEST_P(ConfigWithBFloat16, BFloat16Test)
+{
+#if MIOPEN_EMBED_DB
+
+    const auto& handle         = get_handle();
+    const char* const p_envVar = miopen::GetStringEnv(MIOPEN_TEST_FLOAT_ARG{});
+
+    if(IsTestSupportedForDevice(handle) &&
+       (p_envVar != nullptr && std::strcmp(p_envVar, "--bfloat16") == 0))
```
atamazov: [R] Ditto

atamazov (changes requested):
Some refactoring is highly recommended.

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
xinlipn: Done, on a side note, google test Macro ASSERT_NE(float_arg, nullptr) can only be used in a void function, otherwise there will be a compile error

"cannot initialize return object of type 'bool' with an rvalue of type 'void'"

test/gtest/conv_2d.hpp
atamazov: [Notice] Not resolved yet. Let's either fix or discuss.

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

test/gtest/conv_embed_db.cpp
atamazov: [Resolved]

atamazov:
Almost done. Two things left:
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2168#discussion_r1238661871
- resolve merge conflicts

test/gtest/conv_2d_implicit_gemm.cpp
atamazov: [Notice] W/As for #874 and #1008 finally removed by #2220

test/CMakeLists.txt
atamazov: [Notice] W/As for #874 and #1008 finally removed by #2220

test/gtest/conv_embed_db.cpp
xinlipn: > @JehandadKhan @alexandraBara
> 
> > By default ( ie when no vars are set) all the precisions should run.
> 
> But this is not implemented yet. Shall we do it right here or let's postpone?

@atamazov , right now, the gtest case is indeed controlled by environment variables as below since we are in a transit state from CTest to gTest. It appears by default, "--float" will be enabled

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/54c586f34535e50b2dd56bcd78233b896a5050b5/test/CMakeLists.txt#L203-L227
 
 My understand is as we moving completely away from CTests, we can enable all the tests simply by removing the env var controlled function, i.e. IsTestRunWith()

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/20532ed32950b6612ac29159fe14f50fcc5ad797/test/gtest/conv_embed_db.cpp#L116

xinlipn:
> @xinlipn BTW testing is not covered in the PR description. I means testing of the test itself, does it perform as expected.
> 
> Have you run this test with MIOPEN_LOG_LEVEL=5 and checked if the log matches your expectations?

@atamazov , there was some discrepancies in terms of initiated solvers for test coverage between develop branch and this branch, e.g.   Forward convolution: **ConvAsmImplicitGemmV4R1DynamicFwd_1x1** in Develop vs **ConvHipImplicitGemmV4R1Fwd** in current branch for 
Input tensor: 128, 512, 7, 7
Weights tensor: 2048, 512, 1, 1
Output tensor: 128, 2048, 7, 7

Took some time to remove the discrepancies by updated cmake flags on Develop branch. 

- Log from Develop branch:

[devleop_make_test_conv_embed_db_stdout.log](https://github.com/ROCmSoftwarePlatform/MIOpen/files/11960478/devleop_make_test_conv_embed_db_stdout.log)

- Log from current branch

[branch_ctest.log](https://github.com/ROCmSoftwarePlatform/MIOpen/files/11960480/branch_ctest.log)

Please note MIOPEN_LOG_LEVEL=5 has been enabled, but it has affected the gtest serial spew. I have been working on this, which may need a separate ticket.

junliume (approved):

PR #2175, opened 2023-05-30
[Tests] Add kernel fusion parameter change gtest.

gonidelis:
Initial effort to add kernel fusion gtest. <s>I feel like I still need to handle the destructors at `TearDown()` that's why it's Draft.</s>(destructor is never used). Tidied up the code. Should be ready now. Test passes locally.

<s>This is based on top of #2211 so it needs to be merged after it.</s> (_done_)

### DIFF
--- a/test/gtest/fusion_test.cpp
+++ b/test/gtest/fusion_test.cpp
@@ -32,6 +32,7 @@
 #include "get_handle.hpp"
 #include "verify.hpp"
 #include "cpu_conv.hpp"
+#include "cpu_bias.hpp"
 #include "solver.hpp"
 
 #if MIOPEN_BACKEND_HIP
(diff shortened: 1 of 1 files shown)


test/gtest/fusion_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+std::vector<int> strides(std::vector<int> dims)
+{
+    std::vector<int> strides_array(4);
+    int d = 1;
+    for(int i = 3; i >= 0; --i)
+    {
+        strides_array[i] = d;
+        d *= dims[i];
+    }
```
JehandadKhan: This can be converted to a `std::transform` and is likely already implemented in one of the ctors for the `TensorDescriptor`

test/gtest/fusion_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+    return strides_array;
+}
+
+void upload(float*& pd, const std::vector<float>& v)
+{
+    hipMalloc(&pd, v.size() * 4);
+    hipMemcpy(pd, &v[0], v.size() * 4, hipMemcpyHostToDevice);
+}
```
JehandadKhan: You can use the `tensor` class in the tests header. Also dont use the hip calls directly, instead use `miopen::Handle` methods.

test/gtest/fusion_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+class FusionTestApi : public ::testing::Test
+{
+protected:
+    void SetUp() override
+    {
+        // hipInit(0);
+
+        miopenCreateWithStream(&handle, (hipStream_t)(0));
```
JehandadKhan:
```suggestion
        const auto& handle = get_handle();
```

JehandadKhan (changes requested):

test/gtest/fusion_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include <gtest/gtest.h>
+#include <miopen/miopen.h>
+
+#if MIOPEN_BACKEND_HIP // equiv to hipInit(0)?
```
bghimireamd: Turning on `MIOPEN_BACKEND_HIP` does trigger a call to `hipInit(0)`  
```suggestion
#if MIOPEN_BACKEND_HIP // trigger call to hipInit(0)
```

test/gtest/fusion_test.cpp
gonidelis: done

test/gtest/fusion_test.cpp
gonidelis: Is it cogent using a handle both in `SetUp()` and inside the test case? I need the handle for the `CompileFusionPlan()` calls inside `TEST_F()`.

test/gtest/fusion_test.cpp
gonidelis: It is a separate function https://github.com/ROCmSoftwarePlatform/MIOpen/blob/68524a8223a6550115589e2bd671d378d92fe470/src/tensor.cpp#L290 but how could I use it in construction initializer since the descriptor is a member inside tensor 
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/68524a8223a6550115589e2bd671d378d92fe470/test/tensor_holder.hpp#L111 ?

test/gtest/fusion_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    tensor<T> h_bias;
+    miopen::Allocator::ManageDataPtr d_bias;
+
+    tensor<T> h_output;
+    miopen::Allocator::ManageDataPtr d_output;
+
+    tensor<T> ref_out;
+    miopen::ConvolutionDescriptor conv_desc;
+    // miopen::BiasDescriptor bias_desc; // does not exist
```
JehandadKhan: Please remove

gonidelis:
@bghimireamd @iq136boy reviews are welcomed.

gonidelis:
@JehandadKhan @bghimireamd Tests pass, feel free to review so we can merge.

JehandadKhan (approved):

JehandadKhan:
@junliume  This can be merged!

junliume (approved):

test/gtest/fusion_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+struct FusionSetArgTestFloat : FusionSetArgTest<float>
+{
+};
+
+TEST_P(FusionSetArgTestFloat, TestSetArgApiCall)
+{
+    // Original fusion_plan/args execution happens in cba_infer.cpp
+    // Original is checked independently and not sequentially, prior to FusionTestSetArgTest.
+    using cba_float = cba<float>;
```
atamazov, line 68: There is no check of the datatype under the test. As a result, this test only works with FP32, no matter what data type is specified. Please observe `MIOPEN_TEST_FLOAT_ARG` and use it in google tests, like we do in #2193. Thank you.

atamazov (commented):
The test must obey `MIOPEN_TEST_HALF/BFLOAT/INT8` etc

test/gtest/fusion_test.cpp
junliume: Hi @gonidelis could you fix in a follow up PR?

PR #2177, opened 2023-05-31
[W/A] Link dl to MIOpen driver

JehandadKhan:
Issue: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2178


driver/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ find_package(Threads REQUIRED)
 add_executable(MIOpenDriver main.cpp InputFlags.cpp)
 target_link_libraries(MIOpenDriver MIOpen)
 target_link_libraries(MIOpenDriver ${CMAKE_THREAD_LIBS_INIT})
+target_link_libraries(MIOpenDriver ${CMAKE_DL_LIBS})
```
junliume: @JehandadKhan  it is happening when linking with other executables, not just MIOpenDriver.
For example: failure when building FIN
ref: http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/jd%2Flink_dl/1/pipeline

junliume: and ${CMAKE_THREAD_LIBS_INIT} might also be needed for `-lpthread`

JehandadKhan: Then we need to add it there as well

atamazov:
@junliume The issue blocks develop. Smoke tests should be enough, so I recommend merging just after smoke tests passed.

src/CMakeLists.txt
```diff
@@ -683,7 +683,9 @@ endif()
 
 target_include_directories(MIOpen SYSTEM PUBLIC $<BUILD_INTERFACE:${HALF_INCLUDE_DIR}>)
 target_include_directories(MIOpen SYSTEM PRIVATE ${BZIP2_INCLUDE_DIR})
-target_link_libraries(MIOpen PRIVATE ${CMAKE_THREAD_LIBS_INIT} ${BZIP2_LIBRARIES} ${MIOPEN_CK_LINK_FLAGS})
+# Work around : change in rocm-cmake was causing linking error so had to add ${CMAKE_DL_LIBS}
```
atamazov:
```suggestion
# Workaround : change in rocm-cmake was causing linking error so had to add ${CMAKE_DL_LIBS} 
```
to make this W/A findable by means of a simple text search.

pfultz2:
So https://github.com/ROCmSoftwarePlatform/rocm-recipes/commit/7c09d33ac418c31cd5fe1f6e3c2605061dd367c6 should fix this issue.

junliume (approved):

PR #2187, opened 2023-06-06
Filter out fwd igemm kernel where split number is larger than input channel

carlushuang:
This can fix swdev 404056

### DIFF
--- a/src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
@@ -700,7 +700,7 @@ bool PerformanceConfigAsmImplicitGemmGTCFwdXdlopsNHWC::IsValid(
     if(!(tensor_a_thread_lengths[1] == 1 && tensor_b_thread_lengths[1] == 1))
     {
         // if both 1, indicate padded c support
-        if((c >> tunable->gemm_k_global_split == 0) ||
+        if((c >> gemm_k_global_split == 0) ||
            (((c >> gemm_k_global_split) / group) % gemm_k_per_block != 0))
             return false;
         // also, add this restriction to k, for vector write out
--- a/src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
@@ -700,7 +700,8 @@ bool PerformanceConfigAsmImplicitGemmGTCFwdXdlopsNHWC::IsValid(
     if(!(tensor_a_thread_lengths[1] == 1 && tensor_b_thread_lengths[1] == 1))
     {
         // if both 1, indicate padded c support
-        if( (c >> tunable->gemm_k_global_split == 0) || (((c >> gemm_k_global_split) / group) % gemm_k_per_block != 0))
+        if((c >> tunable->gemm_k_global_split == 0) ||
+           (((c >> gemm_k_global_split) / group) % gemm_k_per_block != 0))
             return false;
         // also, add this restriction to k, for vector write out
         if(problem.IsFp16())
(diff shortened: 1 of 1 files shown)


src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
```diff
@@ -700,7 +700,8 @@ bool PerformanceConfigAsmImplicitGemmGTCFwdXdlopsNHWC::IsValid(
     if(!(tensor_a_thread_lengths[1] == 1 && tensor_b_thread_lengths[1] == 1))
     {
         // if both 1, indicate padded c support
-        if(((c >> gemm_k_global_split) / group) % gemm_k_per_block != 0)
+        if((c >> gemm_k_global_split == 0) ||
+           (((c >> gemm_k_global_split) / group) % gemm_k_per_block != 0))
```
atamazov, line 704: Does this fix correctness issue or something else?

atamazov (changes requested):
Please add regression test. Or let me know affected targets and convolution config(s) and I'll create a patch for inclusion into this PR.

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
carlushuang: Yes. When `C` is larger than power 2 of `gemm_k_global_split`, the kernel should not be picked up. Hence if we do not have this condition, will have correctness issue.

carlushuang:
> Please add regression test. Or let me know affected targets and convolution config(s) and I'll create a patch for inclusion into this PR.

`./bin/MIOpenDriver convfp16 -n 10 -c 1 -H 128 -W 66 -k 768 -y 16 -x 16 -p 0 -q 0 -u 10 -v 10 -l 1 -j 1 -m conv -g 1 -F 1 -t 1` will trigger the failure. After this PR, this case will be correct

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
JehandadKhan: @atamazov I understand that `IsApplicable` is called after loading a Db entry, so a change to the databases is not required, can you please confirm?

junliume: Interested in whether tuning is required after this PR too. @atamazov

atamazov: @JehandadKhan @junliume 

### Perf-db.
- If applicability is narrowed, then some records become unusable (harmless junk). Re-tuning is not required unless you want to remove the junk.
- If applicability is extended, then the solver is unable to show it's full potential. This is harmless as well. Re-tuning is needed only if we need to unleash the performance for the newly enabled configs ASAP.

### Find-db
- If applicability is narrowed, then some info become unusable as well (harmless junk), because we check applicability prior using info from find-db. Re-tuning is not required unless you want to remove the junk.
  - This is true since #1987 is complete. But if system find-db is is old format (this is true for 5.5 release), then performance drop may be bigger than expected (an unwanted side effect)
- If applicability is extended, then the solver may be unable to show it's full potential if the library runs in HYBRID or DYNAMIC_HYBRID modes, or when used in Immediate Mode.

atamazov:
@carlushuang 
> > Or let me know affected targets and convolution config(s) and I'll create a patch for inclusion into this PR.

Which targets?

junliume:
> ./bin/MIOpenDriver convfp16 -n 10 -c 1 -H 128 -W 66 -k 768 -y 16 -x 16 -p 0 -q 0 -u 10 -v 10 -l 1 -j 1 -m conv -g 1 -F 1 -t 1

On MI200s at least,

On MI210 and MI250X it picked `Solution: 107/ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC` and has the issue
On MI100 it picked `Solution: 74/ConvAsmImplicitGemmGTCDynamicFwdXdlops` and works fine

junliume (approved):

atamazov:
@junliume @carlushuang well, I've asked about targets for which ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC is applicable. No problem, I can look myself.

PR #2189, opened 2023-06-07
bg/LWPMIOPEN-149 : now able to log fusion cmd

bghimireamd:
log cmd for fusion (conv + bias + activ) operations. The log will be latter used by tuna for tuning purposes. Had to re-work the branch that was [on hold](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2028).


src/fusion.cpp
```diff
@@ -44,6 +44,20 @@
 #define MIOPEN_CHECK(x)          \
     if(x != miopenStatusSuccess) \
         return x;
+
+enum class ConvDirection
```
atamazov: Please check if we really need this for fusions.

test/gtest/log_test_helper.cpp
```diff
@@ -198,3 +226,32 @@ void TestLogFun(std::function<void(const miopenTensorDescriptor_t&,
     else
         ASSERT_FALSE(isSubStr(str, sub_str)) << "str     : " << str << "str_sub : " << sub_str;
 }
+
+void TestLogCmdFusion(std::function<void(const miopenFusionPlanDescriptor_t)> const& func,
+                      std::string env_var,
+                      std::string sub_str,
+                      bool set_env)
+{
+    // start capturing std::cerr
```
atamazov: Everywhere: In comments, please explain _why_ you are doing something, not _what_ you're doing.

If the source code is so unclear that comments are needed to understand what it does, then something is wrong (and requires some refactoring at least).

atamazov (changes requested):

test/gtest/log_test_helper.cpp
bghimireamd: Agree! comment should be why and code should be what.

src/fusion.cpp
bghimireamd: Since, I am reusing `ConvArgsForMIOpenDriver` that was defined in [src/convolution_api.cpp](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/bg/LWPMIOPEN-149_re_worked/src/convolution_api.cpp), I do have to at least forward declare `enum class ConvDirection;`.

atamazov: Please see another review.

src/convolution_api.cpp
```diff
@@ -403,12 +403,12 @@ enum class ConvDirection
     WrW = 4
 };
 
-static std::string ConvArgsForMIOpenDriver(const miopenTensorDescriptor_t& xDesc,
-                                           const miopenTensorDescriptor_t& wDesc,
-                                           const miopenConvolutionDescriptor_t& convDesc,
-                                           const miopenTensorDescriptor_t& yDesc,
-                                           const ConvDirection& conv_dir,
-                                           bool is_immediate)
+std::string ConvArgsForMIOpenDriver(const miopenTensorDescriptor_t& xDesc,
```
atamazov: - Create a new module (driver_arguments.cpp/hpp?) 
- Move the definition of this function into .cpp
- Put all necessary declarations (including ConvDirection) into .hpp
- include driver_arguments.cpp here
- notice: carefully consider namespace usage (miopen::debug::impl seems adequate but this is up to you)

src/fusion.cpp
```diff
@@ -44,6 +44,15 @@
 #define MIOPEN_CHECK(x)          \
     if(x != miopenStatusSuccess) \
         return x;
+
+enum class ConvDirection;
```
atamazov: include the proposed driver_arguments.hpp instead of this.

src/fusion.cpp
```diff
@@ -158,6 +167,98 @@ static auto AllocateBuffersAndMakeConvBiasActivFusionInvokeParams(
                                               gfx90aaltimpl);
 }
 
+namespace debug {
+
+int GetFusionMode(const miopenFusionPlanDescriptor_t fusePlanDesc)
```
atamazov: Please use unnamed namespace for this function.

src/fusion.cpp
```diff
@@ -158,6 +167,98 @@ static auto AllocateBuffersAndMakeConvBiasActivFusionInvokeParams(
...
+            const miopenTensorDescriptor_t& wDesc         = &conv_op->filter_desc;
+            const miopenConvolutionDescriptor_t& convDesc = &conv_op->base_desc;
+            const miopenTensorDescriptor_t& yDesc         = &deref(fusePlanDesc).output_desc;
+
+            std::string str = "CBAInfer -F " + std::to_string(fusion_mode) + " ";
+            str += ConvArgsForMIOpenDriver(xDesc, wDesc, convDesc, yDesc, ConvDirection(1), false);
+            // cleanup ConvArgsForMIOpenDriver's string
+            str = ReplaceString(str, "-m conv", "");
+            str = ReplaceString(str, "conv", "");
+            str = ReplaceString(str, "-F 1", "");
```
atamazov: String manipulation is technically a hack (because it introduces some implicit dependencies). Please add necessary parameter(s) to `ConvArgsForMIOpenDriver()` so it will be able to handle fusions internally.

atamazov (changes requested):

src/fusion.cpp
atamazov: [Resolved]

test/gtest/log_test_helper.cpp
atamazov: [Resolved]

atamazov:
@bghimireamd Please feel free to hide ("resolve") the review threads marked with [Resolved]. I can't do that.

src/convolution_api.cpp
bghimireamd: moved declaration to src/include/miopen/driver_arguments.hpp
moved definition to src/driver_arguments.cpp. 

Question:
I get miopen::debug. But why impl? 

If we use it, we place miopen::debug::impl  around declarations inside driver_arguments.hpp and driver_arguments.cpp?

src/fusion.cpp
bghimireamd: code updated in new commit.

src/driver_arguments.cpp
```diff
@@ -0,0 +1,120 @@
+#include <miopen/driver_arguments.hpp>
```
JehandadKhan, line 26: Please add the copyright notice

src/fusion.cpp
```diff
@@ -158,6 +160,98 @@ static auto AllocateBuffersAndMakeConvBiasActivFusionInvokeParams(
...
+            const miopenTensorDescriptor_t& yDesc         = &deref(fusePlanDesc).output_desc;
+
+            std::string str = "CBAInfer -F " + std::to_string(fusion_mode) + " ";
+            str += ConvArgsForMIOpenDriver(
+                xDesc, wDesc, convDesc, yDesc, ConvDirection::Fwd, false, false);
+            MIOPEN_LOG_DRIVER_CMD(str);
+        }
+        else
+        {
+            // TODO: logging of other fused operators
```
JehandadKhan: Why leave this as TODO?

src/fusion.cpp
```diff
@@ -158,6 +160,98 @@ static auto AllocateBuffersAndMakeConvBiasActivFusionInvokeParams(
                                               gfx90aaltimpl);
 }
 
+namespace {
+
+int GetFusionMode(const miopenFusionPlanDescriptor_t fusePlanDesc)
+{
+    int fusion_mode = -1;
```
JehandadKhan: This can be moved to `driver_arguments.cpp`

JehandadKhan (dismissed):

src/driver_arguments.cpp
bghimireamd: copyright notice added.

src/fusion.cpp
bghimireamd: moved to `driver_arguments.cpp`

src/fusion.cpp
bghimireamd: added logging for batch norm.

iq136boy (dismissed):

test/gtest/log_test_helper.cpp
JehandadKhan, line 1: Perhaps we can rename the file, since the filename would become the test name with `test` prefixed to it. Therefore, I dont see why the filename should have `test` in it and also the word `helper`

bghimireamd: yes `test` in the name could be misleading. I will change the name to just `test/gtest/log.cpp`

JehandadKhan (approved):

junliume (approved):
Merging latest `develop`, prepare for merging.

iq136boy (approved):

PR #2190, opened 2023-06-07
[tests] GPU verification should be the default for the MIOpenDriver

JehandadKhan:
Currently, to use the GPU reference implementation a special variable needs to be set, however, the faster GPU verification should always be used to improve testing and developer experience.


JehandadKhan:
@atamazov

atamazov (changes requested):
Please also remove `MIOPEN_DRIVER_USE_GPU_REFERENCE=1` from tests/CMakeLists.txt

atamazov:
@JehandadKhan Please also add `MIOPEN_DRIVER_USE_GPU_REFERENCE=0` to
- `smoke_solver_ConvDirectNaiveConv_F`
- `smoke_solver_ConvDirectNaiveConv_BW`

Thanks.

junliume (dismissed):

junliume (approved):

atamazov:
@CAHEK7 @JehandadKhan @junliume FYI using GPU reference in MIOpenDriver instead of disk cache is 2-3 times slower on average. So I the `-C` option still remains important in the driver for testing (my own set of tests takes ~hour to finish with disk cache which is much better that 2-3 hours)

JehandadKhan:
@atamazov  That might be because the bulk of time is spent on populating the tensors rather than doing the convolutions, which I think is a problem we should solve independent of CPU vs GPU reference.

CAHEK7:
@JehandadKhan I won't be so sure, right now PRNG is quite fast - just a few (3-5) CPU instructions per value.
Meanwhile, loading data from the disk can be much faster than transfer it on GPU, compute everything and transfer it back.

Also I have an idea - what about data check on gpu?

PR #2205, opened 2023-06-15
Fixed a value being checked incorrectly

DrizztDoUrden:
This fixes NHWC calls returning `miopenStatusBadParam` in Find 2.0 when it would not in Find 1.0.


junliume (approved):

kahmed10:
According to the MIOpen spec for NHWC, apparently the dims are always assumed to be in NCHW order, and the strides are modified instead. So the original check would still be valid. If there are any future bugs because of this change, please back out.

DrizztDoUrden:
Yes, but the check is still performed only for backward data convolution in find 1.0. That shouldn't differ between 1.0 and 2.0, so it's fixed here.

PR #1747, opened 2022-09-12
Full blown fix for PR #1681 "Performance drops after PR 1634"

atamazov:
The effect of this PR is reduction of host-size overhead related to ConvBinWinogradRxSf2x3 and ConvBinWinogradRxSf2x3g1 for non-group Forward and Backward convolutions, which may improve the so-called Initial Iteration Time.

Achieved by making the (tunable) ConvBinWinogradRxSf2x3 non-applicable for the above configs. The "g1" solver can do the same job with the same performance, and it does not require tuning. Therefore the library does not need to access perf-db, and this saves some time.

## Details

- [find-db] Replaced ConvBinWinogradRxSf2x3 by ConvBinWinogradRxSf2x3g1 for non-group Fwd/Bwd (see #1681)
- [find-db] Replaced ConvBinWinogradRxSf2x3g1 by ConvBinWinogradRxSf2x3 for group WrW (see #1735)
  - ℹ️ Tunable f2x3 can be more efficient than non-tunable "g1" solver for all group convolutions and for non-group WrW.
- Enabled ConvBinWinogradRxSf2x3g1 for WrW to avoid perf regressions due to weird f2x3 tuning values in the system perf-db (caught for MI60)
  - Explanation and some discussion at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1735/files#r968772575
- Disabled `WORKAROUND_ISSUE_1681`
  - ⚠️ W/A is not fully removed because it is possible that "incorrect" find-db records may get back in the future, after porting of tuning results from release branch to develop (see https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1681#issuecomment-1210524707). The release branch does not contain this PR, and that is why tuning results may get problematic records again. If this happens, the issue could be quickly camouflaged by enabling the W/A again. However I would prefer fixing find-db, which is very easy with regexp.

## 🌀 Local performance & correctness testing

- Radeon VII, ROCm 5.0
- _**Databases borrowed from MI60**_
- ___All solvers except ConvBinWinogradRxSf2x3 and ConvBinWinogradRxSf2x3g1 disabled.___
- 93 known FP32 convolution configs
- Default Find mode, Normal Find mode, Immediate mode, with and without re-tuning
- HIP backend
- Develop vs this PR

  | GPU time | Wall time | Aux wall time
-- | -- | -- | --
Find API, default databases | 0% | 0% | 480%
Find API, Normal mode (1), default   databases | 0% | 0% | 380%
[...]

PR #2208, opened 2023-06-16
Op3dTensorGeneric kernel improvements

CAHEK7:
Simplified and significantly improved speed (up to 250+ times on gfx9xx and 100+ times on gfx10xx) of Op3dTensorGeneric kernel. There are few cases where it can be up to 20% slower than initial kernel, but most of them are small tensors - less than 8K elements in total.

## Local verification
- gfx90a?, rocm ??
- gfx1030, rocm 5.4

### Metrics
Average gain = (sum speedup) / (num cases).
Total time gain = sum(time old kernel) / sum(time new kernel)

B copy mode | GPU | Average gain | Total time gain | Num cases
-- | -- | -- | -- | --
1 1 1| navi21 | 70.4 | 56.0 | 3174
1 1 x| navi21 | 15.1 | 11.6 | 3174
1 x 1| navi21 | 4.3 | 1.3 | 3174
1 x x| navi21 | 20.2 | 17.5 | 3174
x 1 1| navi21 | 5.4 | 1.5 | 3174
x 1 x| navi21 | 12.8 | 9.3 | 3174
x x 1| navi21 | 4.7 | 1.3 | 3174
x x x| navi21 | 59.4 | 35.5 | 3174
1 1 1| MI210 | 126.6 | 197.3 | 2489
1 1 x| MI210 | 8.9 | 11.6 | 2489
1 x 1| MI210 | 4.5 | 2.3 | 2489
1 x x| MI210 | 11.3 | 26.6 | 2489
x 1 1| MI210 | 5.8 | 3.1 | 2489
x 1 x| MI210 | 8.2 | 18.1 | 2489
x x 1| MI210 | 2.8 | 1.5 | 2489
x x x| MI210 | 32.4 | 51.8 | 2489
1 1 1| --- | 95.1 | 125.3 | 5663
1 1 x| --- | 12.4 | 11.6 | 5663
1 x 1| --- | 4.4 | 1.6 | 5663
1 x x| --- | 16.3 | 19.8 | 5663
x 1 1| --- | 5.5 | 1.9 | 5663
x 1 x| --- | 10.8 | 11.6 | 5663
x x 1| --- | 3.9 | 1.4 | 5663
x x x| --- | 47.5 | 39.2 | 5663


CAHEK7:
[perf.xlsx](https://github.com/ROCmSoftwarePlatform/MIOpen/files/11771985/perf.xlsx)
Perf measurements

shurale-nkn (changes requested):
MIOpen no longer merging PRs with new OpenCL kernels. OCl kernels should be rewritten in HIP.

Usage of this kernel should be limited.
At navi21 allowed only at size >8k and without masks: 1x1, x11, xx1.
At MI210 allowed only at size >8k.

Based on perfDoc can 

navi21 summary perf gain:

  |   |   |   | Btensor   copy mode |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
  | 111 | 11x | 1x1 | 1xx | x11 | x1x | xx1 | xxx
avg | 39.341 | 6.964 | 2.376 | 10.347 | 2.370 | 6.158 | 2.693 | 33.827
med | 29.051 | 3.820 | 1.130 | 4.300 | 1.121 | 3.126 | 1.173 | 28.050
max | 114.574 | 54.241 | 36.840 | 67.688 | 38.028 | 47.754 | 42.621 | 103.066
min | 0.994 | 0.857 | 0.748 | 0.791 | 0.831 | 0.789 | 0.672 | 1.148

navi21 percentage of cases with perf-drop:

  |   |   |   | Btensor   copy mode |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
  | 111 | 11x | 1x1 | 1xx | x11 | x1x | xx1 | xxx
\>8k size | 0% | -7% | -8% | -5% | -7% | -4% | -5% | 0%
<8k size | 0% | 0% | -16% | 0% | -8% | 0% | -16% | 0%
total | 0% | -2% | -14% | -1% | -8% | -1% | -13% | 0%


MI210 summary perf gain:

  |   |   |   | Btensor   copy mode |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
  | 111 | 11x | 1x1 | 1xx | x11 | x1x | xx1 | xxx
avg | 85.859 | 5.919 | 3.124 | 7.241 | 3.194 | 5.531 | 1.845 | 21.626
med | 37.678 | 2.571 | 1.290 | 2.107 | 1.264 | 1.770 | 1.279 | 10.707
max | 260.639 | 77.878 | 71.348 | 52.500 | 73.897 | 36.060 | 25.053 | 68.346
min | 0.878 | 0.616 | 0.678 | 0.681 | 0.607 | 0.634 | 0.635 | 0.940


MI210 percentage of cases with perf-drop:

  |   |   |   | Btensor   copy mode |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
  | 111 | 11x | 1x1 | 1xx | x11 | x1x | xx1 | xxx
\>8k size | 0% | 0% | -1% | 0% | -1% | -1% | -5% | 0%
<8k size | -14% | -34% | -37% | -19% | -38% | -18% | -19% | 0%
total | -3% | -8% | -9% | -4% | -9% | -5% | -8% | 0%

Kirpich30000:
@JehandadKhan @junliume is "no new opencl kernels" is a strict rule or we could proceed with this PR and transfer to HIP in a separate PR? Work on this one began before the announcement.

Kirpich30000:
@CAHEK7 please provide more detailed descriptions in the top comment regarding verification and performance changes if PR could possibly affect one or another. Few examples #1927, #1747. Tables shown by @shurale-nkn  in this [comment ](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2208#pullrequestreview-1484629015) are also a good example of what would be nice to see in original description.

CAHEK7:
@Kirpich30000, thanks for the examples, I will update this pull request once we decide to proceed with opencl.

As about the limitations - for navi21 the case with full B tensor (case x x x) is always faster for packed version (+15% is the worst case) and at least has the same performance for incredible strided access (4KB strides for dimensions of 2-3 elements). Why would we ignore such kind of performance gain?  The same is for squashed B tensor (case 1 1 1) - at least it has the same performance except a few cases with 3-5% perf drop and one of the cases is outside of 8k boundary, so heuristic doesn't work.

As I mentioned in some out of github conversations - most of the small tensors performance drops can be mitigated by increasing number of active groups and decreasing number of threads per group. Since I utilize all the threads from the group, small tensors may use just a few wraps of one group and may underutilize GPU - previous version of the kernel used just one thread of the group and much more groups instead and small tensors were distributed by larger number of wraps\CUs. So we can mimic the same behavior, replace old kernel completely, significantly simplify the host code and minimize number of kernels to support (for example to rewrite kernels to HIP).

Also why would we give up on 30-40 times of performance gain totally excluding 1x1, x11 and xx1 cases?

CAHEK7:
I would like to have some clear metrics\guidelines about kernels tuning, acceptable performance drops, code quality and performance heuristics.

Looks like the main problem in this PR that is nobody knows (or at least I wasn't able to get this information from some of the participants) what is the most common cases for that kind of tensor operations and we don't have any methodology how to do performance tracking, analysis and comparison. I can create any set of tests to prove that one kernel is better than another and since all the cases are equal it's almost impossible to make a decision which kernel is better and shall we produce highly tuned kernel for each corner case or not.
We can spend a lot of time to discuss each particular case in each new PR if we don't have any strategic decision and clear agreements on that question. 

Currently I can see that we should sacrifice significant performance gains (more than 10 times faster) and code quality (we have to maintain mixed opencl and hip kernels in one function and a bunch of heuristics to choose between them and make highly tuned kernel for each special corner cases) to cover 10-15% performance drops for some marginal cases. Is it the main goal? If the answer is `yes`, then I'll try make maximum conservative heuristic to choose between kernel to avoid any performance drops. If the answer is `no` it would be nice to have any measurable guidelines how to make such decisions.

Another question, a bit out of scope be very important for me - shall we plan accordingly and, first of all, switch everything to HIP, perform all the necessary checks and measurements and only then start to improve some kernels based on a new codebase? Especially if we want to keep all the initial algorithms for each corner case. Seems like it may affect code quality and readability when we mix multiple types of kernels with multiple complex heuristics in one function.
Or at least I would be happy to know in advance that I should not modify opencl kernel, but rather to write and test a new one.

@junliume @JehandadKhan could you help me with this?

CAHEK7:
A few words about any "average" metrics from this PR - currently they all are highly biased to lower-sized tensors since tensor sizes grow in some kind of log2 scale. If I choose different datasets, for example evenly distributed tensors sizes, I will get completely different picture in terms of "average", "median" and all the percentages from this [comment](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2208#pullrequestreview-1484629015) and from my initial description.

CAHEK7:
Ok, I'll close this PR, since there is no response from the main reviewer.

NB: seems there are no perf tests for tensor operations in CI testsuite.

This PR blocks #2221 and I'll continue the work on tensor operations improvements under that issue.

PR #2215, opened 2023-06-20
update winograd convolution WTI computation

CAHEK7:
Update winograd convolution WTI computation according to the latest measurements and documentation.

The most significant changes are:
- renamed vars to match the documentation
- adjusted some the formulas
- added more precise cycle-related constants

It's an initial PR which updates only current 2x3g1 convolution Wti.
The same code will be reused for groups computation and for the common case of convolution (in separated PRs).

New and old wti computation methods comparison:
[wti-graph.xlsx](https://github.com/ROCmSoftwarePlatform/MIOpen/files/11805547/wti-graph.xlsx)

### DIFF
--- a/src/solver/conv_winoRxS.cpp
+++ b/src/solver/conv_winoRxS.cpp
@@ -464,10 +464,6 @@ public:
             !(K >= 16)
     }
     {
-        // Computations do not support negative padding.
-        // Negative padding is not applicable, so let use simple assert here.
-        assert(pad_h >= 0 && pad_w >= 0);
-
         /// \todo add G to UnifiedDescriptionConv2d
         size_t G = static_cast<size_t>(problem.GetGroupCount());
 
--- a/src/solver/conv_winoRxS.cpp
+++ b/src/solver/conv_winoRxS.cpp
@@ -410,113 +410,57 @@ ConvBinWinoRxS<Winodata, Winofilter>::Search(const ConvolutionContext& ctx,
     return GenericSearch(*this, ctx, problem, invoke_ctx);
 }
 
-template <int Winodata, int Winofilter>
 class ShaderModel : public UnifiedDescriptionConv2d
 {
-    static_assert(IS2X3 || IS3X2, "Unsupported filter configuration");
-
-    static constexpr size_t TILE_S = Winofilter;
-    static constexpr size_t TILE_R = Winofilter;
-
-    static constexpr size_t TILE_OUT_W = Winodata;
-    static constexpr size_t TILE_OUT_H = Winodata;
-
     static constexpr size_t NHW_tiles_factor = 32;
 
-    size_t G;
+    size_t Ts;
+    size_t Tr;
+
+    size_t Tow;
+    size_t Toh;
+
     size_t n_CU;
     size_t n_groups;
 
-    bool is_fp16;
-
-    bool dstride2_w;
-    bool dstride2_h;
-    bool dstride2;
-
-    bool ostride2_w;
-    bool ostride2_h;
-    bool ostride2;
-
-    bool stride1_w;
-    bool stride1_h;
-    bool stride1;
-
-    bool single_mode;
     double direct_convolution_macs;
 
     size_t S_factor;
-    size_t Sg;
-
-    size_t R_factor;
-    size_t Rg;
-
     size_t C_factor;
+
+    size_t Rg;
+    size_t Sg;
     size_t Cg;
 
-    size_t K_factor;
-    size_t Kg;
-
-    size_t OWg;
-    size_t OHg;
-
-    size_t NWH_tiles;
-    size_t NHW_tiles_g;
-
-    size_t n_works;
     size_t n_works_per_CU;
 
+    bool dstride2;
+    bool ostride2;
+
+    bool single_mode;
+
+    bool is_fp16;
     bool out_of_model_scope; // Shader model produces unreliable results.
 
+    bool is2x3() const noexcept { return Tow == 2 && Ts == 3; }
+    bool is3x2() const noexcept { return Tow == 3 && Ts == 2; }
+
 public:
-    ShaderModel(const ConvolutionContext& ctx, const ProblemDescription& problem)
+    ShaderModel(const ConvolutionContext& ctx,
+                const ProblemDescription& problem,
+                size_t Winodata,
+                size_t Winofilter)
         : UnifiedDescriptionConv2d(problem),
-        G{static_cast<size_t>(problem.GetGroupCount())},
-        n_CU{ctx.GetStream().GetMaxHardwareComputeUnits()}, /// \todo Take n_CU from PerfConfig.
-
-        n_groups{G == 1 ? n_CU : 0}, // will be recomputed for G != 1
-
-        is_fp16{problem.IsFp16()},
-
-        dstride2_w{input_stride_w == 2},
-        dstride2_h{input_stride_h == 2},
-        dstride2{dstride2_w && dstride2_h},
-
-        ostride2_w{U == 2},
-        ostride2_h{V == 2},
-        ostride2{ostride2_w && ostride2_h},
-
-        stride1_w{!(dstride2_w || ostride2_w)},
-        stride1_h{!(dstride2_h || ostride2_h)},
-        stride1{stride1_w && stride1_h},
-
-        single_mode{
-            stride1_w &&
-            (S <=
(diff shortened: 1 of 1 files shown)


src/solver/conv_winoRxS.cpp
```diff
@@ -413,115 +413,213 @@ ConvBinWinoRxS<Winodata, Winofilter>::Search(const ConvolutionContext& ctx,
 template <int Winodata, int Winofilter>
 class ShaderModel : public UnifiedDescriptionConv2d
 {
-    const size_t DATATYPE_BITS;    // S
-    const size_t n_groups;         // BQ ~compute units
-    const bool out_of_model_scope; // Shader model produces unreliable results.
+    static_assert(IS2X3 || IS3X2, "Unsupported filter configuration");
```
Kirpich30000: IS2X3 == true && IS3X2 == true is also unsupported configuration

CAHEK7: Sure, only IS2X3 or IS3X2 are supported.
IS2X3 and IS3X2 are mutual exclusive and only IS2X3 can be true or only IS3X2 can be true:
`#define IS2X3 (Winodata == 2 && Winofilter == 3)`
`#define IS3X2 (Winodata == 3 && Winofilter == 2)`
Looks like everything is good here.

CAHEK7:
There are 62 (31 fp1 and 31 fp32) cases where the new algorithm predicted low cycle count and wti returned `-1`.
[wti-disabled-cases.xlsx](https://github.com/ROCmSoftwarePlatform/MIOpen/files/11817034/wti-disabled-cases.xlsx)

src/solver/conv_winoRxS.cpp
```diff
@@ -413,115 +413,213 @@ ConvBinWinoRxS<Winodata, Winofilter>::Search(const ConvolutionContext& ctx,
...
+                                      1e6;
+
+        if(cycles_predicted <= 0.1)
             return -1.0; // Unreliable, too small work to do for the shader.
 
-        const auto N_MACS_PER_CU_PER_CLOCK = static_cast<size_t>(64) * 32 / DATATYPE_BITS;
-        const auto WTI_predicted           = direct_convolution_macs /
-                                   static_cast<double>(N_MACS_PER_CU_PER_CLOCK) /
-                                   static_cast<double>(n_groups) / GUI_predicted; // similar to BW
+        const size_t macs_per_cu_per_clock = 64 * (is_fp16 ? 2 : 1);
```
Kirpich30000, line 602: Only true for gfx10 and gfx906 and earlier. However the fix should consider usage of the predicted wti.

CAHEK7: I can add more precise calculation based on GPU architecture, but I need some sort of documentation here and we have to decide what would be in a default case.
As far as I know, for gfx11 it should be multiplied by 2.

src/solver/conv_winoRxS.cpp
```diff
@@ -413,115 +413,213 @@ ConvBinWinoRxS<Winodata, Winofilter>::Search(const ConvolutionContext& ctx,
...
+
+        const auto WTI_predicted =
+            direct_convolution_macs /
+            (static_cast<double>(macs_per_cu_per_clock * n_groups) * cycles_predicted);
```
Kirpich30000: Consider following approach
```
ideal_direct_cycles = direct_convolution_macs / n_CU / macs_per_cu_per_clock;
WTI_predicted = ideal_direct_cycles / cycles_predicted;
```

src/solver/conv_winoRxS.cpp
CAHEK7: I'll get rid of this static assert here and move it inside convolution.
Moreover we've decided to remove template parameters, so it can't be static anymore.

src/solver/conv_winoRxS.cpp
CAHEK7: In terms of current algorithm application, changing n_groups to n_CU does not change anything, since right not it is applied to only to the case where `G==1` and `n_groups==n_CU`.
So it doesn't require any additional testing for this PR.

src/solver/conv_winoRxS.cpp
CAHEK7: We decided to keep it as it is right now.

src/solver/conv_winoRxS.cpp
```diff
@@ -410,118 +410,205 @@ ConvBinWinoRxS<Winodata, Winofilter>::Search(const ConvolutionContext& ctx,
...
 #endif
-                             !(C >= 16) || //
-                             !(K >= 16))
+            !(C >= 16) || //
+            !(K >= 16)
+    }
     {
         // Computations do not support negative padding.
         // Negative padding is not applicable, so let use simple assert here.
         assert(pad_h >= 0 && pad_w >= 0);
```
Kirpich30000: I believe this assert is not needed. Only the padding parity (even or odd) affects granularity.

CAHEK7: Ok, I've just kept it from the initial version of this class. Can remove it if it seems more correct.

Basically, I don't even know why we initially put this assert here instead of moving this condition it into `out_of_scope` variable and just ignore this algorithm for negative padding and decided to crash with assertion.

Kirpich30000: We could remove it

Kirpich30000 (approved):
LGTM

junliume (approved):

src/solver/conv_winoRxS.cpp
```diff
@@ -410,118 +410,201 @@ ConvBinWinoRxS<Winodata, Winofilter>::Search(const ConvolutionContext& ctx,
...
+
+        S_loops = Sg / S_factor;
+        R_loops = Rg / Tr;
+        S_loops = Cg / C_factor;
+
+        if(G == 1)
+        {
+            n_groups       = n_CU;
+            n_works_per_CU = Ceil(n_works, n_CU);
+            return;
```
junliume, line 521: @CAHEK7 hip tidy issue on this line:
```
[2023-06-24T07:12:45.296Z] /var/jenkins/workspace/MLLibs_MIOpen_develop/src/solver/conv_winoRxS.cpp:521:13: error: 1 uninitialized field at the end of the constructor call [clang-analyzer-optin.cplusplus.UninitializedObject,-warnings-as-errors]
[2023-06-24T07:12:45.296Z]             return;
[2023-06-24T07:12:45.296Z]             ^

File: src/solver/conv_winoRxS.cpp [hunk 6]
```diff
@@ -410,118 +410,201 @@ ConvBinWinoRxS<Winodata, Winofilter>::Search(const ConvolutionContext& ctx,
...
+        if(cycles_predicted <= 0.1)
             return -1.0; // Unreliable, too small work to do for the shader.
 
-        const auto N_MACS_PER_CU_PER_CLOCK = static_cast<size_t>(64) * 32 / DATATYPE_BITS;
-        const auto WTI_predicted           = direct_convolution_macs /
-                                   static_cast<double>(N_MACS_PER_CU_PER_CLOCK) /
-                                   static_cast<double>(n_groups) / GUI_predicted; // similar to BW
+        /// \todo Adjust for different architectures, probably it should be an external source for
+        /// this metric
+        const size_t macs_per_cu_per_clock = 64 * (is_fp16 ? 2 : 1);
```
junliume, line 602: @CAHEK7 hip tidy issue with this line:
```
[2023-06-24T07:12:45.297Z] /var/jenkins/workspace/MLLibs_MIOpen_develop/src/solver/conv_winoRxS.cpp:602:46: error: performing an implicit widening conversion to type 'const size_t' (aka 'const unsigned long') of a multiplication performed in type 'int' [bugprone-implicit-widening-of-multiplication-result,-warnings-as-errors]

[2023-06-24T07:12:45.297Z]         const size_t macs_per_cu_per_clock = 64 * (is_fp16 ? 2 : 1);

[2023-06-24T07:12:45.297Z]                                              ^

junliume:
@CAHEK7 my mistake thinking this PR passed CI (the green tick). Unfortunately I have to revert this PR, due to reasons claimed above. Could you please re-work this PR to fix the issue? For new PR I will help to get CI through. Thanks!

src/solver/conv_winoRxS.cpp
CAHEK7: Addressed in #2224

src/solver/conv_winoRxS.cpp
CAHEK7: Addressed in #2224

PR #2179, opened 2023-05-31
[tests] Remove W/A for issues #874 and #1008

atamazov:
After merging this, #874 and #1008 can be closed.

PR #2037, opened 2023-03-21
[Tuning][MI100][MI200] conv1d, tensorflow, pytorch

cderb:
Tuning for tickets MSRCHA-320, SWDEV-369339, SWDEV-388250, and SWDEV-389187

Perf dbs grew too large - bzip2 compression added for perf dbs and kernel dbs
Decompression added to cmake

PR #2220, opened 2023-06-23
[tests] Remove W/A for issues #874 and #1008. [Fix] Always unzip databases.

atamazov:
Another attempt to merge #2179, which was first merged in but then reverted due to unclear reason:

> > @atamazov
> > I am curious how this PR passed the CI. Can you please attach also the passing og from this PR? I suspect it was run on gfx906/64 while the develop fails on gfx906/60, but there could be some other reason. I really want to finish those two tickets and remove the W/As.
> 
> @junliume 
> Unfortunately the cached CI history for this PR does not exist any more :( This PR is reverted in develop branch, could we try again using another PR? Thanks!

----
_Original description:_
After merging this, #874 and #1008 can be closed.


atamazov:
I am unable to find anything wrong in the library except the following.

When I build the library locally, I see lines like these in the CMake output:
```
-- HALF_INCLUDE_DIR: /opt/rocm/include
-- /usr/bin/lbunzip2 -k /root/github/MLOpen1/src/kernels/gfx906_60.db.bz2
-- /usr/bin/lbunzip2 -k /root/github/MLOpen1/src/kernels/gfx906_60.HIP.fdb.txt.bz2
-- /usr/bin/lbunzip2 -k /root/github/MLOpen1/src/kernels/gfx906_60.OpenCL.fdb.txt.bz2
...
```
This is unpacking databases from .db2 files.

However these lines are MISSING in the failing log at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2179#issuecomment-1584011146.

AFAICS this happens when unpacked (target) files already exist. I have no idea **_how these files may reside in the directory after checkout in our CI_**, but the log file confirm that they are definitely there. By default, lbunzip2 does not overwrite existing files. So it is likely that CI failed due to outdated perf-db files (where perf-configs of ConvBinWinogradRxSf3x2 are missing here and there).

This came from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2037/files#r1240092073.

I've run into issues with this kind of behavior a few times. After pulling the new version of the library, the databases remained the same and the performance testing results (which I collect often) were incorrect.

Another problem looks like this:
- (1) user does not have lbunzip installed
- (2) user runs CMake - > FAILURE (` UNZIPPER-NOTFOUND`)
- (3) user installs lbunzip
- (4) user runs CMake -> OK

Everything is fine? No! All databases are zero size. The reason is that in step (2) zero size database files were somehow created.

I consider it is necessary to resolve this problem roughly, but once and for all.

@JehandadKhan @junliume

CMakeLists.txt
```diff
@@ -472,16 +472,17 @@ function(unpack_db db_bzip2_file)
     STRING(REPLACE ".bz2" "" db_file "${db_bzip2_file}")
     find_program(UNZIPPER lbunzip2 bunzip2)
     if(EXISTS "${db_file}")
-        # message(WARNING "${db_file} already exists")
-    elseif(EXISTS "${db_bzip2_file}")
-        message(STATUS "${UNZIPPER} -k ${db_bzip2_file}")
+        message(STATUS "WARNING: ${db_file} already exists and will be overwritten")
```
atamazov, line 475: This situation is normal for local builds, but strange for CI.

Note: I use `message(STATUS "WARNING: ...)` because `message(WARNING...)` looks ugly and clutters up the screen.

CMakeLists.txt
```diff
@@ -472,16 +472,17 @@ function(unpack_db db_bzip2_file)
...
+        message(STATUS "${UNZIPPER} -k -f ${db_bzip2_file}")
         execute_process(
-            COMMAND ${UNZIPPER} -k ${db_bzip2_file}
+            COMMAND ${UNZIPPER} -k -f ${db_bzip2_file}
             WORKING_DIRECTORY ${KERNELS_DIR}
             RESULT_VARIABLE ret
             )
         if(NOT ret EQUAL "0")
-            message(WARNING "${db_file} could not be extracted, file is empty!")
+            message(FATAL_ERROR "${db_file} could not be extracted, file is empty!")
```
atamazov, line 485: FATAL because if file exists, then it must be overwritten (or we may run into db sync issues).

atamazov (commented):
for reviewers

atamazov:
@JehandadKhan @junliume This is ready for review.

atamazov:
- Proposed reviewers:
  - @JehandadKhan
  - @junliume
  - @cderb 

This PR seems important to me because it prevents building the library with outdated or empty perf-db/find-db. Building the library with wrong databases may lead to both false positives and false negatives on CI etc etc.

DrizztDoUrden (approved):

junliume:
@atamazov  somehow the CI is failing at Int8 HIP all stage for Vega20: 
[2220.log](https://github.com/ROCmSoftwarePlatform/MIOpen/files/11886728/2220.log)

atamazov:
@junliume Thanks for the feedback and for the detailed log!

This PR does not affect the failing test_tensor_trans, so I guess this is a glitch OR our tensor transform kernels may work incorrectly on MI60 (the test has failed on gfx906/64). However I will try to reproduce it on my gfx906/60, just to be sure. In the meantime, one question and one request:
- Were there failures at other stages?
- Please start another round of testing of this PR as it is.

junliume (approved):

atamazov:
@junliume Thank you!

PR #2225, opened 2023-06-27
[WORKAROUND] Disable W/A for issue #1359 starting from ROCm 5.4.3.

### LINKED ISSUE #2223: [Issue #1359] hipModuleLoad/hipModuleUnload is leaking file descriptor
(Thanks to @JehandadKhan to find the issue and to create the minimal reproduce example), creating this issue on GitHub on his behalf)
- **[Symptom]:**

During MIOpen exhaustive tuning (e.g. MIOPEN_FIND_ENFORCE=3), MIOpen needs to open a large number of modules, even though we proactively close the ones not been used (e.g. #1221), the program still fails with
```
  what():  /long_pathname_so_that_rpms_can_package_the_debug_info/src/extlibs/MLOpen/src/hipoc/hipoc_program.cpp:162: Failed creating module from file
```
This issue was previously discussed in #1359 #1360 

- **[Steps to reproduce and Findings]:**

Using the minimal reproducible example below:
[module_api.zip](https://github.com/ROCmSoftwarePlatform/MIOpen/files/11853854/module_api.zip)

_(need hip enabled docker, in this case any public docker should be able to reproduce the issue)_

*ROCm version where this issue can be reproduced*:
ROCm 5.5
*cannot* reproduce in ROCm 5.4.3
```
cd module_api
make clean
make
./hip_module_api
```
and we will see
```
Iteration: 65399
An error encountered: "shared object initialization failed" at main.hip:57
```

- **[Hypothesis]:**

The test code is very simple, essentially:
```
while(true) {
	std::cout << "Iteration: " << ++idx << std::endl;
    // std::this_thread::sleep_for(std::chrono::milliseconds(50));
    hipModule_t module;
    HIP_CHECK(hipModuleLoad(&module, module_path.u8string().c_str()));

    HIP_CHECK(hipModuleUnload(module));
}
```
however, using the following monitoring command:
```
watch -n1 'cat /proc/sys/fs/file-nr | tee -a log.log'
```
we can observe the # of FD that was left open (leaked) keep increasing till the HIP program crashes.

- **[Discussions]:**

1. is `hipModuleUnload` leaking FD?
2. the # of FD that was left open (leaked) causes crashes well before it reaches `ulimit`, so is there any HIP limitation on the # of FD?

CC: @atamazov @averinevg @DrizztDoUrden @dmikushin @CAHEK7

atamazov:
- The straightforward attempt to fix #2223 which modifies W/A for issue #1359.
  - After this PR, the WORKAROUND_ISSUE_1359 will be disabled starting with ROCm 5.4.3.
- Testing: FULL tests on Navi21 with ROCm 5.4.3 -- PASSED.

## ⚠️ Before merging this, we must ensure that SWDEV-304151 is not appearing again.
- I am expecting that some other member of MIOpen team will do that.
- [Important] Please pay special attention to PyTorch WHL

----
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/non-miopen-bug
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_high
- Proposed reviewers:
  - @JehandadKhan
  - @junliume
  - @pfultz2 
  - @DrizztDoUrden


junliume:
Recommend on hold until ROCm 5.7 branch is cut.

atamazov:
@junliume So the problem this FDs is not critical? Or we have other solutions in mind?

junliume:
> @junliume So the problem this FDs is not critical? Or we have other solutions in mind?

hipruntime/ROCr has a patch for it, we do not want to risk `hipModuleLoadData` now, and will test with a full cycle of release instead.

atamazov:
@junliume Good!

atamazov:
@junliume @JehandadKhan Maybe it's time to update this PR and test it? We are still using temporary files to pass code objects to the HIP runtime, even if the code objects were built directly in memory (without using the file system), and this may affects host-side performance.

Or does HIP runtime still need association of each device program with a FD?

junliume:
> @junliume @JehandadKhan Maybe it's time to update this PR and test it? We are still using temporary files to pass code objects to the HIP runtime, even if the code objects were built directly in memory (without using the file system), and this may affects host-side performance.
> 
> Or does HIP runtime still need association of each device program with a FD?

The FD issue in runtime keep re-appearing and re-fix for every release :( We recommended putting this check on their staging verification.

However, I agree that we should proceed with the WA removal nevertheless. Could you merge latest develop and I will kick off the CI pipeline again.

JehandadKhan:
@atamazov  Lets go ahead!

junliume (approved):

PR #2234, opened 2023-06-30
Improve GET_RAND() performance

gonidelis:
Using `rand()` is considered harmful. Replace `rand()` with `rand_r()` instead on the driver code side.

### DIFF
--- a/driver/random.hpp
+++ b/driver/random.hpp
@@ -5,19 +5,25 @@
 #include <ctime>
 #include <cstdint>
 #include <random>
-#include <chrono>
+
+std::minstd_rand& get_minstd_gen()
+{
+    static std::random_device rd;
+    static std::minstd_rand minstd_gen(rd());
+
+    return minstd_gen;
+}
 
 template <typename T>
 inline T FRAND()
 {
-    std::minstd_rand minstd_gen(std::chrono::system_clock::now().time_since_epoch().count());
-    auto d = std::generate_canonical<double, 5>(minstd_gen);
+    double d = std::generate_canonical<double, 1>(get_minstd_gen());
     return static_cast<T>(d);
 }
 
 inline int GET_RAND()
 {
-    std::minstd_rand minstd_gen(std::chrono::system_clock::now().time_since_epoch().count());
+    auto minstd_gen = get_minstd_gen();
     return minstd_gen();
 }
 
--- a/driver/random.hpp
+++ b/driver/random.hpp
@@ -2,15 +2,18 @@
 #define GUARD_RANDOM_GEN_
 
 #include <cstdlib>
+#include <ctime>
+#include <cstdint>
 
 template <typename T>
 inline T FRAND()
 {
-    double d = static_cast<double>(rand() / (static_cast<double>(RAND_MAX)));
+    std::uint32_t seed = time(nullptr);
+    double d = static_cast<double>(rand_r(&seed) / (static_cast<double>(RAND_MAX)));
     return static_cast<T>(d);
 }
 
-inline int GET_RAND() { return rand(); }
+inline int GET_RAND() {std::uint32_t seed = time(nullptr); return rand_r(&seed); }
 
 template <typename T>
 inline T RAN_GEN(T A, T B)
--- a/driver/random.hpp
+++ b/driver/random.hpp
@@ -9,11 +9,15 @@ template <typename T>
 inline T FRAND()
 {
     std::uint32_t seed = time(nullptr);
-    double d = static_cast<double>(rand_r(&seed) / (static_cast<double>(RAND_MAX)));
+    double d           = static_cast<double>(rand_r(&seed) / (static_cast<double>(RAND_MAX)));
     return static_cast<T>(d);
 }
 
-inline int GET_RAND() {std::uint32_t seed = time(nullptr); return rand_r(&seed); }
+inline int GET_RAND()
+{
+    std::uint32_t seed = time(nullptr);
+    return rand_r(&seed);
+}
 
 template <typename T>
 inline T RAN_GEN(T A, T B)
--- a/driver/random.hpp
+++ b/driver/random.hpp
@@ -5,9 +5,7 @@
 
 std::minstd_rand& get_minstd_gen()
 {
-    static std::random_device rd;
-    static std::minstd_rand minstd_gen(rd());
-
+    static thread_local std::minstd_rand minstd_gen(std::random_device{}());
     return minstd_gen;
 }
 
@@ -20,7 +18,7 @@ inline T FRAND()
 
 inline int GET_RAND()
 {
-    auto minstd_gen = get_minstd_gen();
+    decltype(auto) minstd_gen = get_minstd_gen();
     return minstd_gen();
 }
 
--- a/driver/random.hpp
+++ b/driver/random.hpp
@@ -1,9 +1,6 @@
 #ifndef GUARD_RANDOM_GEN_
 #define GUARD_RANDOM_GEN_
 
-#include <cstdlib>
-#include <ctime>
-#include <cstdint>
 #include <random>
 
 std::minstd_rand& get_minstd_gen()
--- a/driver/random.hpp
+++ b/driver/random.hpp
@@ -4,18 +4,21 @@
 #include <cstdlib>
 #include <ctime>
 #include <cstdint>
+#include <chrono>
 
 template <typename T>
 inline T FRAND()
 {
-    std::uint32_t seed = time(nullptr);
-    double d           = static_cast<double>(rand_r(&seed) / (static_cast<double>(RAND_MAX)));
+    auto const clk    = std::chrono::system_clock::now().time_since_epoch().count();
+    unsigned int seed = static_cast<unsigned int>(clk);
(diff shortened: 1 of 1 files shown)


junliume:
@gonidelis clang format issue found in CI:
https://github.com/ROCmSoftwarePlatform/MIOpen/wiki/How-to-format-code

gonidelis:
> @gonidelis clang format issue found in CI: https://github.com/ROCmSoftwarePlatform/MIOpen/wiki/How-to-format-code

Yup, I am aware of those. Something's bogus with my formater. Will figure it out.

driver/random.hpp
```diff
@@ -2,15 +2,18 @@
...
 
 #include <cstdlib>
+#include <ctime>
+#include <cstdint>
 
 template <typename T>
 inline T FRAND()
 {
-    double d = static_cast<double>(rand() / (static_cast<double>(RAND_MAX)));
+    std::uint32_t seed = time(nullptr);
```
CAHEK7: I would not recommend using `std::time()` for seed since it returns time in seconds and consecutive calls will return the same value - this mean we will have the same "random" value during each second.

I guess `std::chrono::steady_clock` or `std::chrono::high_resolution_clock` is better choice since usually it has nanosecond resolution (and usually they are the same).

driver/random.hpp
```diff
@@ -2,15 +2,18 @@
...
+    double d = static_cast<double>(rand_r(&seed) / (static_cast<double>(RAND_MAX)));
```
CAHEK7: And I would not recommend using `rand_r()` since it's kind of non-standard function.
It is a part of POSIX, not C\C++ standard and formally speaking it is deprecated.

Have you considered to use `std::generate_canonical` for [0;1) fp distribution?

CAHEK7 (changes requested):
I would suggest to use standard C++ random generators and distributions for portability and compatibility and replace seed with higher precision clock source.
Currently it doesn't produce random numbers for subsequent call during one second.

gonidelis:
> Currently it doesn't produce random numbers for subsequent call during one second.

@CAHEK7 Thanks for the review! The decision was made based on this discussion [#2134](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2135). I will make the changes regarding the seed, but which PRNG to use I guess is still up for discussion.

CAHEK7:
Sure, I'm fine with `rand_r`, since it has been discussed and decided.

But a few words about #2134 discussion - `std::mt19937` is slow indeed, but there is `std::minstd_rand` which is exactly linear congruential generator, but it is a bit slower than `rand_r` since it tries to perform modulus division (few more operations than just imul, and, add). In the other hand `std::minstd_rand` can be inlined and if at some point they will use bitmask instead of modulus, it will outperform even `rand_r`.

driver/random.hpp
gonidelis: Thanks for your suggestion. I updated the code accordingly.

driver/random.hpp
gonidelis: I am inclining towards the standards side for sure. What's the reasoning behind `std::generate_canonical` though? Why a fp in [0;1) specifically?

driver/random.hpp
```diff
@@ -2,15 +2,25 @@
...
+#include <ctime>
+#include <cstdint>
+#include <chrono>
 
 template <typename T>
 inline T FRAND()
 {
-    double d = static_cast<double>(rand() / (static_cast<double>(RAND_MAX)));
+    auto const clk    = std::chrono::system_clock::now();
+    unsigned int seed = static_cast<unsigned int>(std::chrono::system_clock::to_time_t(clk));
```
CAHEK7: `to_time_t` converts nanoseconds to seconds (
```
auto const clk = std::chrono::system_clock::now().time_since_epoch().count();
unsigned int seed = static_cast<unsigned int>(clk);
```

https://godbolt.org/z/zannnd46K

gonidelis:
Note: `std::system_clock()` can still be unreliable (e.g. on Windows) but `rand_r()` only accepts `time_t` (and aliased) types. Converting `std::high_resolution_clock` or `std::steady_clock` is like upcasting to a more coarse type. Decision needs to be made whether C rands need to be sustained.

driver/random.hpp
gonidelis: Thanks for that! I was looking for it rn.

CAHEK7 (dismissed):

CAHEK7:
LGTM
unless you want to go further and use `static thread_local` seed and don't call `::now()` for each function invocation.

gonidelis:
> LGTM
unless you want to go further and use static thread_local seed and don't call ::now() for each function invocation.

@CAHEK7  isn't `::now()` thread safe?

CAHEK7:
It is threadsafe.
The main difference is in performance - getting time may be a bit expensive, especially compared to `rand_r` function.
If we make `seed` static we call `::now` only once and then use pure `rand_r`. We won't rely on timer precision as well.

gonidelis:
> It is threadsafe.
The main difference is in performance - getting time may be a bit expensive, especially compared to rand_r function.
If we make seed static we call ::now only once and then use pure rand_r. We won't rely on timer precision as well.

Fantastic. If we don't end up turning to std generators (very likely) I will use the thread_local specifier for the seed.

gonidelis:
@JehandadKhan  Would enabling CI performance tests be relevant here? Do they include the MIOpenDriver's performance in their results?

JehandadKhan:
> @JehandadKhan Would enabling CI performance tests be relevant here? Do they include the MIOpenDriver's performance in their results?

They do indeed, however I would like to see some benchmarking of the driver itself to make sure that it works.

gonidelis:
> They do indeed, however I would like to see some benchmarking of the driver itself to make sure that it works.

On it.

gonidelis:
**Updated Changes Notes**
After doing some significant research on performance, portability, thread safety I need to clarify the following for future devs:

1. Going with a `std::random` generator improves portability given that we are going with a standards solution.
2. `minstd_rand()` is the most lightweight option we have (compared to Mersenne-Twister for example).
3. The main performance gain is coming from the fact that we are reusing the generator without initializing it every time we call `FRAND()`/`GET_RAND()`.
4. Part of not having to initialize it every time is the decision to use `std::random_device` (instead of the more lightweight `std::chrono::steady_clock::now()`) which again is only initialized once and used multiple times consecutively.
5. From benchmark results that I ran on an allocated gfx908 machine I observed more than 10% performance improvement and a significantly different hotspots profile (I can share the results, feel free to PM me).

driver/random.hpp
```diff
@@ -1,16 +1,28 @@
 #ifndef GUARD_RANDOM_GEN_
 #define GUARD_RANDOM_GEN_
 
-#include <cstdlib>
+#include <random>
+
+std::minstd_rand& get_minstd_gen()
+{
+    static std::random_device rd;
+    static std::minstd_rand minstd_gen(rd());
```
CAHEK7: `thread_local` for the sake of thread safety.

driver/random.hpp
```diff
@@ -1,16 +1,28 @@
...
 {
-    double d = static_cast<double>(rand() / (static_cast<double>(RAND_MAX)));
+    double d = std::generate_canonical<double, 1>(get_minstd_gen());
     return static_cast<T>(d);
 }
 
-inline int GET_RAND() { return rand(); }
+inline int GET_RAND()
+{
+    auto minstd_gen = get_minstd_gen();
```
CAHEK7: `auto &` since default `auto` does not inherit reference.

CAHEK7 (dismissed):

driver/random.hpp
CAHEK7: It's better to collapse two static variables into one, especially if the first one is needed just to initialize the second.

gonidelis: > It's better to collapse two static variables into one, especially if the first one is needed just to initialize the second.

as in static `std::minstd_rand minstd_gen(std::random_device());` ?

CAHEK7: Almost, something like `std::minstd_rand minstd_gen(std::random_device{}());`

driver/random.hpp
gonidelis: done!

driver/random.hpp
gonidelis: done :)

CAHEK7 (approved):
LGTM

junliume (approved):

driver/random.hpp
```diff
@@ -1,16 +1,26 @@
...
 template <typename T>
 inline T FRAND()
 {
-    double d = static_cast<double>(rand() / (static_cast<double>(RAND_MAX)));
+    double d = std::generate_canonical<double, 1>(get_minstd_gen());
     return static_cast<T>(d);
 }
 
-inline int GET_RAND() { return rand(); }
+inline int GET_RAND()
```
atamazov, line 19: This breaks the verification cache feature (`--verification_cache` or `-C`) in the convolution driver. The reason is that results of `GEN_RAN` is not reproducible anymore.

atamazov (commented):
Breaks `MIOpenDriver conv* ... -C ...`

PR #2211, opened 2023-06-16
Add convolution base test boilerplate class.

gonidelis:
Add base convolution tests header with reusable code in order to remove code duplication from various gtests on their `SetUp()` `TearDown()`.

PR #2251, opened 2023-07-10
MIOpen FP8 and BFP8 enablement

JehandadKhan:
This PR enables the F8 data type in MIOpen. 

* This means addition of the `miopenFloat8` and the `miopenBFloat8` members to the `miopenDataType` enum. 
* This PR also adds the casting type the tensor descriptor and the corresponding API to get and set it. When set the kernel will cast the fp16 input tensors to the set cast type ( either miopenFloat8 or miopenBFloat8) before performing the computation.
* Currently, no performant kernels are added for the above data types or supporting casting operations. Such kernels would be added in a subsequent PR. Currently, only naive conv kernels are available and the use of rocBlas for performing F8 convolution operations.
* Only convolution operations are supported.
* All old solvers are disabled for the new data types.


src/ocl/tensorocl.cpp
```diff
@@ -1446,12 +1446,12 @@ void SetTensor(const Handle& handle,
                                           std::multiplies<std::size_t>());
 
         std::size_t wld = 256 < wgd ? 256 : wgd;
-
-        std::string parms = "-DSUBTENSOR_OP_WITH_SCALAR=SUBTENSOR_OP_WITH_SCALAR_SET" +
-                            GetDataTypeKernelParams(dataType);
+        std::stringstream ss;
```
CAHEK7, line 1475: I'm just curious - what was the reason to replace std::string with std::stringstream? Potentially `std::stringstream` can be less efficient, more complicated, and we don't need any advanced formatting features and compatibility with stream interface since we pass `std::string` everywhere. (it's not about this particular file, it's almost everywhere)

JehandadKhan: Perhaps @atamazov  can shed more light on which is efficient, but so far the insistence of reviewers has been to prefer `std::stringstream` when constructing strings since it avoid creating many intermediate copies of strings by the `+` operator. Therefore, whenever I touch older code utilizing the `+` operator to concatenate strings I prefer to replace it with stream which can "stream in" substrings without creating copies.

include/miopen/miopen.h
```diff
@@ -690,7 +700,19 @@ MIOPEN_EXPORT miopenStatus_t miopenSetTensorDescriptor(miopenTensorDescriptor_t
                                                        const int* dimsA,
                                                        const int* stridesA);
 
-/*! @brief Get size of N-dimensional tensor
+/*! @brief Set the tensor cast type
```
JehandadKhan, line 714: [anchor]

JehandadKhan:
@averinevg  Can you help update this PR to both resolve the merged conflict and update it for the upcoming `conv::ProblemDescription` changes ?

averinevg:
> @averinevg Can you help update this PR to both resolve the merged conflict and update it for the upcoming `conv::ProblemDescription` changes ?

@JehandadKhan Sure

averinevg:
@JehandadKhan I've resolved merge conflicts and updated this PR for the upcoming `conv::ProblemDescription` changes (https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2251/commits/53c648e1abb517046a719149c646d2d00977c32e). Please take a look at this [commit](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2251/commits/c7a079a90663a8cfe28328c8307864f591aa6025), these changes are related to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2211. I would also suggest splitting this pull request into smaller ones.

junliume:
@JehandadKhan  this is high urgency item while we are having some HIP tidy issues in CI.

junliume:
@JehandadKhan some tidy issues which may need to be fixed:
```
[2023-09-05T20:16:39.003Z] /home/jenkins/workspace/MLLibs_MIOpen_jd_f8/test/gtest/conv_trans.cpp:50:12: error: enumeration values 'miopenFloat8' and 'miopenBFloat8' not explicitly handled in switch [-Werror,-Wswitch-enum]

[2023-09-05T20:16:39.003Z]     switch(prec)

[2023-09-05T20:16:39.003Z]            ^~~~

[2023-09-05T20:16:39.003Z] [ 79%] Built target test_fusion_aux

[2023-09-05T20:16:39.003Z] /home/jenkins/workspace/MLLibs_MIOpen_jd_f8/test/gtest/conv_hip_igemm_xdlops.cpp:60:12: error: enumeration values 'miopenFloat8' and 'miopenBFloat8' not explicitly handled in switch [-Werror,-Wswitch-enum]

[2023-09-05T20:16:39.003Z]     switch(prec)

[2023-09-05T20:16:39.003Z]            ^~~~
```

driver/conv_driver.hpp
```diff
@@ -380,9 +402,14 @@ class ConvDriver : public Driver
...
         // the one of fp32 because mantissa is shorter by 13 bits.
         auto tolerance = (sizeof(Tgpu) == 4 || sizeof(Tgpu) == 1) ? 1.5e-6 : 8.2e-3;
+
         // bf16 mantissa has 7 bits, by 3 bits shorter than fp16.
         if(std::is_same<Tgpu, bfloat16>::value)
             tolerance *= 8.0;
+        constexpr bool is_fp8  = std::is_same<Tgpu, float8>::value;
+        constexpr bool is_bfp8 = std::is_same<Tgpu, bfloat8>::value;
+        if(is_bfp8 || is_fp8 || TensorsCasted())
+            tolerance = 0.3f;
```
atamazov: It makes the `--tolerance` option useless with F8. Can we use the multiplier that, applied to the default tolerance, results in 0.3f?

driver/conv_driver.hpp
```diff
@@ -154,6 +154,27 @@ void dumpBufferToFile(const char* fileName, T* data, size_t dataNumItems)
     }
 }
 
+miopenDataType_t DataTypeFromShortString(const std::string& type)
```
atamazov:
```suggestion
static inline miopenDataType_t DataTypeFromShortString(const std::string& type)
```

driver/conv_driver.hpp
```diff
@@ -154,6 +154,27 @@ void dumpBufferToFile(const char* fileName, T* data, size_t dataNumItems)
...
+
+    const auto res = conv_map.find(type);
+    if(res != conv_map.end())
+    {
+        return res->second;
+    }
+    else
+    {
+        printf("Invalid compute/cast type short hand supplied\n");
+        exit(-1);
```
atamazov: In the driver we parse and validate options in the `ParseCmdLineArgs()` member function.

include/miopen/miopen.h
```diff
@@ -581,6 +589,8 @@ typedef enum
     MIOPEN_CONVOLUTION_ATTRIB_DETERMINISTIC =
         1, /*!< Restrict MIOpen convolutions to kernels which produce numerically deterministic
               results. 0 - disabled (default), 1 - enabled >*/
+    MIOPEN_CONVOLUTION_ATTRIB_FP8_ROUNDING_MODE = 2,
+    MIOPEN_CONVOLUTION_ATTRIB_COMPUTE_TYPE      = 3,
```
atamazov: The documentation of these two attributes is missing. This is simple: for example, for `MIOPEN_CONVOLUTION_ATTRIB_FP8_ROUNDING_MODE` you can simply say "A value from miopenF8RoundingMode_t". But the glue should be provided.

driver/conv_driver.hpp
```diff
@@ -1048,6 +1106,14 @@ int ConvDriver<Tgpu, Tref>::SetConvDescriptorFromCmdLineArgs()
         convDesc, spatial_dim, pads.data(), conv_strides.data(), conv_dilations.data(), mode);
 
     miopenSetConvolutionGroupCount(convDesc, group_count);
+    if(inflags.GetValueStr("compute_type") != "-1")
+    {
+        miopenDataType_t compute_type =
+            DataTypeFromShortString(inflags.GetValueStr("compute_type"));
+        // miopenSetConvolutionComputeType(convDesc, compute_type);
+        miopenSetConvolutionAttribute(
+            convDesc, MIOPEN_CONVOLUTION_ATTRIB_COMPUTE_TYPE, static_cast<int>(compute_type));
```
atamazov: This is not implemented. I highly recommend moving this incomplete functionality to another PR.

include/miopen/miopen.h
```diff
@@ -698,7 +708,19 @@ MIOPEN_EXPORT miopenStatus_t miopenSetTensorDescriptor(miopenTensorDescriptor_t
                                                        const int* dimsA,
                                                        const int* stridesA);
 
-/*! @brief Get size of N-dimensional tensor
+/*! @brief Set the tensor cast type
+ *
+ *  For tensors where the cast_type attribute is set, the tensor elements would be cast to the
+ * target type before the target operation is applied. Currently, only supported for convolution
+ * operations targeting the FP8 dataype
```
atamazov: typo

include/miopen/miopen.h
```diff
@@ -698,7 +708,19 @@ MIOPEN_EXPORT miopenStatus_t miopenSetTensorDescriptor(miopenTensorDescriptor_t
                                                        const int* dimsA,
                                                        const int* stridesA);
 
-/*! @brief Get size of N-dimensional tensor
+/*! @brief Set the tensor cast type
+ *
+ *  For tensors where the cast_type attribute is set, the tensor elements would be cast to the
```
atamazov: [Recommendation] @JehandadKhan Maybe the word "conversion" is a better fit? The "cast" is something from the C/C++ programmer's and magican's vocabulary (which is not surprising because they are almost the same except wands ;)

atamazov (changes requested):
Partial review

include/miopen/miopen.h
JehandadKhan: Fixed!

include/miopen/miopen.h
JehandadKhan: Fixed!

driver/conv_driver.hpp
JehandadKhan: Fixed!

include/miopen/miopen.h
JehandadKhan: Fixed!

driver/conv_driver.hpp
JehandadKhan: Fixed!

driver/conv_driver.hpp
JehandadKhan: Fixed!

driver/conv_driver.hpp
JehandadKhan: Fixed!

src/kernels/MIOpenUtilKernels4.cl
```diff
@@ -48,7 +48,7 @@
 #define MIOPEN_USE_INT32 0
 #endif
 
-#if MIOPEN_USE_INT8
+#if MIOPEN_USE_INT8 || MIOPEN_USE_FP8 || MIOPEN_USE_BFP8
```
junliume, line 59: @JehandadKhan `test_tensor_tran` has failed for the following issue:
```
[2023-09-09T01:57:50.415Z] MIOpen(HIP): Error [Do] 'amd_comgr_do_action(kind, handle, in.GetHandle(), out.GetHandle())' AMD_COMGR_ACTION_COMPILE_SOURCE_WITH_DEVICE_LIBS_TO_BC: ERROR (1)
[2023-09-09T01:57:50.415Z] MIOpen(HIP): Error [BuildOcl] comgr status = ERROR (1)
[2023-09-09T01:57:50.415Z] MIOpen(HIP): Warning [BuildOcl] /tmp/comgr-992752/input/MIOpenUtilKernels4.cl:51:24: error: 'MIOPEN_USE_FP8' is not defined, evaluates to 0 [-Werror,-Wundef]
[2023-09-09T01:57:50.415Z] #if MIOPEN_USE_INT8 || MIOPEN_USE_FP8 || MIOPEN_USE_BFP8
[2023-09-09T01:57:50.415Z]                        ^
[2023-09-09T01:57:50.415Z] /tmp/comgr-992752/input/MIOpenUtilKernels4.cl:51:42: error: 'MIOPEN_USE_BFP8' is not defined, evaluates to 0 [-Werror,-Wundef]
[2023-09-09T01:57:50.415Z] #if MIOPEN_USE_INT8 || MIOPEN_USE_FP8 || MIOPEN_USE_BFP8
[2023-09-09T01:57:50.415Z]                                          ^
```

src/kernels/hip_f8_impl.hpp
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    uint32_t drop_mask = (1 << (mfmt - wm)) - 1;
+    const int max_exp  = (1 << we) - (negative_zero_nan ? 1 : 2);
+    const int exp_low_cutoff =
+        (sizeof(T) == 4 ? 128 : 16) - (1 << (we - 1)) + 1 - (negative_zero_nan ? 1 : 0);
+
+    exponent -= exp_low_cutoff - 1;
+    if(exponent <= 0)
+        drop_mask = (1 << (mfmt - wm + 1 - exponent)) - 1;
```
junliume: @JehandadKhan some other tests has shown the shift may be too large in some cases: e.g.
```
runtime error: shift exponent 141 is too large for 32-bit type 'int'
```

src/check_numerics.cpp
```diff
@@ -53,30 +53,41 @@ struct CheckNumericsResult
...
-    const std::vector<size_t> vgd = {numGlobalWorkItems, size_t{1}, size_t{1}};
-    handle.AddKernel("MIOpenCheckNumerics", "", program_name, kernel_name, vld, vgd, params)(
+    const size_t threadsPerBlock = 256;
+    const size_t numBlocks       = handle.GetMaxComputeUnits() * 6;
+    const int computeStats       = (mode & CheckNumerics::ComputeStats);
+    // TODO - some constants we should get from the device:
+    std::string program_name      = "MIOpenCheckNumerics.cpp";
+    std::string kernel_name       = GetKernelName(dDesc.GetType());
+    const std::vector<size_t> vld = {size_t{threadsPerBlock}, size_t{1}, size_t{1}};
+    const std::vector<size_t> vgd = {numBlocks, size_t{1}, size_t{1}};
```
atamazov, line 88: [Q] `vgd[0]` was `handle.GetMaxComputeUnits() * 6 * 256` before, but how it is `handle.GetMaxComputeUnits() * 6`. Is it correct?

driver/conv_driver.hpp
```diff
@@ -1135,11 +1243,14 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
...
-    bool is_transform       = IsInputTensorTransform();
-    bool is_int8            = data_type == miopenInt8 || data_type == miopenInt8x4;
-    size_t in_sz            = GetTensorSize(inputTensor);
-    size_t wei_sz           = GetTensorSize(weightTensor);
-    size_t out_sz           = GetTensorSize(outputTensor);
+    bool is_transform = IsInputTensorTransform();
+    bool is_int8      = data_type == miopenInt8 || data_type == miopenInt8x4;
+    // Data generated for very low precision types follows the same constraints whether its fp8,
+    // bfp8 or even if the interim tensors are being casted
+    bool is_fp8   = data_type == miopenFloat8 || data_type == miopenBFloat8 || TensorsCasted();
```
atamazov, line 1250: [R][clarity] `is_f8` seems more clear because fp8 is miopenFloat8.

driver/conv_driver.hpp
```diff
@@ -381,9 +402,14 @@ class ConvDriver : public Driver
         // Computation error of fp16 is ~2^13 (=8192) bigger than
         // the one of fp32 because mantissa is shorter by 13 bits.
         auto tolerance = (sizeof(Tgpu) == 4 || sizeof(Tgpu) == 1) ? 1.5e-6 : 8.2e-3;
+
         // bf16 mantissa has 7 bits, by 3 bits shorter than fp16.
         if(std::is_same<Tgpu, bfloat16>::value)
             tolerance *= 8.0;
+        constexpr bool is_fp8  = std::is_same<Tgpu, float8>::value;
+        constexpr bool is_bfp8 = std::is_same<Tgpu, bfloat8>::value;
```
atamazov, line 410: [R][clarity] `is_bf8` -- seems better visually distinguishable from `is_fp8` than `is_bfp8`

driver/conv_driver.hpp
```diff
@@ -381,9 +402,14 @@ class ConvDriver : public Driver
...
+        if(is_bfp8 || is_fp8 || TensorsCasted())
```
atamazov, line 411: [Q] If we implement tensor casting for other datatypes than F8 then `GetDefaultTolerance` will return incorrect result.

Or I do not understand the tensor_casting feature correctly?

driver/conv_driver.hpp
```diff
@@ -1109,6 +1191,32 @@ void RanGenSubnormBuffer(T* buf, size_t size, int percentage)
     });
 }
 
+template <>
+float8 RanGenWeights()
+{
+    const auto tmp =
+        prng::gen_0_to_B(1.0) > 0.5 ? static_cast<float>(0.0) : static_cast<float>(1.0);
```
atamazov, line 1198: [Q] @CAHEK7 `static_cast<float>(prng::gen_0_to_B(1))` (i.e. the generator that should produce integer 0 or 1) should produce the same result, right?

@JehandadKhan BTW `static_cast<float>(0.0)` can be replaced with just `0.0f`

driver/conv_driver.hpp
```diff
@@ -1394,8 +1510,11 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
             db_host     = tensor<Tref>(miopen::deref(biasTensor));
             for(int i = 0; i < b_sz; i++)
             {
-                b.data[i] = static_cast<Tgpu>(i % 8) + prng::gen_canonical<Tgpu>();
-                db[i]     = static_cast<Tgpu>(i % 8) + prng::gen_canonical<Tgpu>();
+                b.data[i] =
+                    static_cast<Tgpu>(i % 8) +
+                    (is_fp8 ? prng::gen_A_to_B(Data_min, Data_max) : prng::gen_canonical<Tgpu>());
```
atamazov, line 1515: [R] if Data_min/max = 0/1 for non-F8 types, then 
```suggestion
                    static_cast<Tgpu>(i % 8) + prng::gen_A_to_B(Data_min, Data_max);
```
as `gen_canonical<T>()` is equivalent to `gen_A_to_B(static_cast<T>(0), static_cast<T>(1))`

@CAHEK7 right?

driver/driver.hpp
```diff
@@ -44,6 +42,12 @@ using float16 = half_float::half;
 #include <memory>
 #include <miopen/miopen.h>
 #include <miopen/bfloat16.hpp>
+using half         = half_float::half;
+using hip_bfloat16 = bfloat16;
+#include <miopen/hip_float8.hpp>
+using float16 = half_float::half;
+using float8  = miopen_f8::hip_f8<miopen_f8::hip_f8_type::fp8>;
+using bfloat8 = miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8>;
```
atamazov, line 50: [Q] Do we really need to put these type alias declarations in the middle of includes?

junliume (approved):

junliume:
@JehandadKhan thanks about the excellent work!

src/gemm_v2.cpp
```diff
@@ -68,14 +73,126 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdisabled-macro-expansion"
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    float alpha = gemm_desc.alpha;
+    float beta  = gemm_desc.beta;
+    auto flags  = FlagsForRocblasFp32Fp16Call(gemm_desc);
+    if(gemm_desc.conv_attributes.fp8rounding_mode.Get() == miopenF8RoundingModeStochastic)
+        flags = flags | rocblas_gemm_flags::rocblas_gemm_flags_stochastic_rounding;
+
+    rb_status = // cppcheck-suppress redundantInitialization
```
atamazov, line 151: [R] In order to get rid of `// cppcheck-suppress..`, simply `return rocblas_gemm_ex3(...)` directly. Create `#else` path and `return rocblas_status::rocblas_status_internal_error` there.

src/gemm_v2.cpp
```diff
@@ -401,35 +481,61 @@ miopenStatus_t CallGemm(const Handle& handle,
...
-                rocblas_datatype::rocblas_datatype_f16_r,
-                gemm_desc.ldc,
-                static_cast<rocblas_half*>(C) + c_offset,
-                rocblas_datatype::rocblas_datatype_f16_r,
-                gemm_desc.ldc,
-                rocblas_datatype::rocblas_datatype_f32_r,
-                rocblas_gemm_algo::rocblas_gemm_algo_standard,
-                0,
-                FlagsForRocblasFp32Fp16Call(gfx90a_alt_impl));
+            const auto is_gfx94x = miopen::StartsWith(handle.GetDeviceName(), "gfx94");
```
atamazov, line 484: [R] Put this at the beginning of the function
```C++
static
const auto is_ex3_api_supported = miopen::StartsWith(handle.GetDeviceName(), "gfx94")
    && USE_ROCBLAS_GEMM_EX3;
```
and use it everywhere instead of `is_gfx94x`.

include/miopen/miopen.h
```diff
@@ -698,7 +711,19 @@ MIOPEN_EXPORT miopenStatus_t miopenSetTensorDescriptor(miopenTensorDescriptor_t
                                                        const int* dimsA,
                                                        const int* stridesA);
 
-/*! @brief Get size of N-dimensional tensor
+/*! @brief Set the tensor cast type
+ *
+ *  For tensors where the cast_type attribute is set, the tensor elements would be converted to the
+ * target type before the target operation is applied. Currently, only supported for convolution
+ * operations targeting the FP8 datatype
```
atamazov, line 718: [Q] Are you sure that this is supported only for FP8 output? What about BF8?

[Q] Am I right that this is supported only for FP16 input tensors? Maybe this is worth mentioning?

[Q] Is it so that conversion (by truncation, i.e. with rounding towards zero) from FP16 (E5M10) to BF8 (E5M2) is free? Because the implementation can simply skip the low byte of FP16 elements.

atamazov (commented):
Detailed review, partial. Some recommendations and questions, no mandatory comments to resolve. More tomorrow.

atamazov:
> @JehandadKhan thanks about the excellent work!

I am second to this ))

atamazov:
@JehandadKhan @cderb Please expect many review comments, and some are going to be about redesigning of internal data structures. 🤕

atamazov:
@JehandadKhan @cderb I am going to create a PR instead of review, because I think this will give results must faster (at the cost of slightly worse knowledge sharing).

src/include/miopen/datatype.hpp
```diff
@@ -159,9 +172,15 @@ inline KernelBuildParameters GetDataTypeKBP(miopenDataType_t type)
...
         {"MIOPEN_USE_RNE_BFLOAT16", use_rne_bfloat16},
+        {"MIOPEN_FP8_IEEE_EXPONENT_BIAS", MIOPEN_FP8_IEEE_EXPONENT_BIAS},
+        {"MIOPEN_FP8_CLIPPING", MIOPEN_FP8_CLIPPING},
     };
     if(use_fp64 != 0)
         kbp.Define("MIOPEN_USE_FP64", use_fp64);
+    if(use_fp8 != 0)
+        kbp.Define("MIOPEN_USE_FP8", use_fp8);
+    if(use_bfp8 != 0)
+        kbp.Define("MIOPEN_USE_FP8", use_bfp8);
```
atamazov, line 183: Fixed in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2452/commits/95bc7175f659c262e0fca0385df518d078ebed38

driver/conv_driver.hpp
CAHEK7: @atamazov `static_cast<float>(prng::gen_canonical<int>())` is even more suitable.
But getting the whole context into account we should replace everything with:
```CPP
return static_cast<float8>(static_cast<float>(std::numeric_limits<float8>::epsilon()) * (2 * prng::gen_A_to_B(-1, 1)));
```

src/include/miopen/datatype.hpp
```diff
@@ -159,9 +172,15 @@ inline KernelBuildParameters GetDataTypeKBP(miopenDataType_t type)
         {"MIOPEN_USE_BFP16", use_bfp16},
         {"MIOPEN_USE_INT32", use_int32},
         {"MIOPEN_USE_RNE_BFLOAT16", use_rne_bfloat16},
+        {"MIOPEN_FP8_IEEE_EXPONENT_BIAS", MIOPEN_FP8_IEEE_EXPONENT_BIAS},
+        {"MIOPEN_FP8_CLIPPING", MIOPEN_FP8_CLIPPING},
```
atamazov, line 176: @JehandadKhan I guess you are aware of that this change invalidates KDB for all solvers that use `SetupFloats()`, right? I would recommend producing these macros only if `MIOPEN_USE_FP8` is defined to narrow impact at least (but now it is too late and I overlooked this initially).

/cc @CAHEK7

atamazov: @cderb Please see previous comment.

PR #2260, opened 2023-07-17
bg/LWPMIOPEN-191: Gracefully exit if search mode is not supported during fusion's compile stage.

bghimireamd:
Gracefully exit when search mode is not supported during fusion's compile state. Currently we only support (conv + bias + activ ) some fusion operation during search mode. Any other fusion operation during search mode should gracefully exit.


junliume:
CP candidate for ROCm 5.6.1

junliume (approved):

Issue #2221, opened 2023-06-23
Improve tensor operations performance


CAHEK7:
During investigation of #2208 I found out that most of the kernels for tensor operations `OP(α1 * A, α2 * B) + β *C` are highly inefficient and overengineered.

Even the simples 1D case, which is either linear by-element operation for 3 equal tensors or B can be scalar constant looks like multiple nested loops with extra index computations and so on. I could be a persistent thread model, but no. Number of groups and work per group are calculated in a strange way and GPU is heavily underloaded.
```
__kernel void Op1dTensorGeneric(...)
{
    int gid = get_group_id(0);

    global MIOPEN_TYPE* a_off = a + Aoffset;
    global MIOPEN_TYPE* b_off = b + Boffset;
    global MIOPEN_TYPE* c_off = c + Coffset;

    // num_wg: the number of workgroups should be launched
    // MAX_NUM_WG: the maximum number of workgroups actually launched
    for(; gid < num_wg; gid += MAX_NUM_WG)
    {
        int lid             = get_local_id(0);
        int o_n_gid_off     = gid % b_n;
        int bindex          = o_n_gid_off;
        MIOPEN_TYPE operand = b_off[bindex] * alpha1;
        while(lid < work_per_wg)
        {
            int o_n    = (bitmap & (1 << 0)) ? o_n_gid_off : lid % c_n;
            c_off[o_n] = MIOPEN_TENSOR_OP(a_off[o_n] * alpha0, operand) + beta * c_off[o_n];
            lid += get_local_size(0);
        }
    }
}
```

Simplifying this code can make everything more clear and supportable and even 8-10 times faster.
There is the same problem for 2d, 3d, 4d and 5d operations, though some of the 2d and 3d cases have some kind of "specialized" kernels which were invented as workarounds for such slow generic ones.

Also, 2d, 3d, 4d and 5d kernels do not support stride in the last dimension which contradicts external API and breaks transposed tensors support. Performance tuning of transposed operations must be a separated task, but for correctness it must work at least somehow.

I would really appreciate any help with testing and performance regression analysis - I want to discuss performance metrics, test cases and performance goals for this task.

CAHEK7:
It started to be more critical, since it is used in RNN layer.
Potentially it can be used for backward attention layer.

CAHEK7:
@junliume @JehandadKhan could you assign proper urgency and importance of this issue and probably plan it accordingly?

[...]

PR #2262, opened 2023-07-17
OpTensor1d improvement

CAHEK7:
1D case for #2221
Simplified kernel code, significantly improved performance, added strides support, rewritten to HIP.

Perf tests dataset
- tensor A size is evenly distributed 2620 values from 2 to 83927466 with explicit power of 2 +-1 values.
- tensor B size is either 1 or equal to tensor A size
- beta multiplier is 0 or 1

Summary for gfx1030:
beta = 0
|metric\B|1|n|
|-|-|-|
|min|1.13|0.93|
|median|88.53|49.31|
|max|147.51|99.78|
|mean|97.18|55.27|

beta = 1
|metric\B|1|n|
|-|-|-|
|min|0.96|0.86|
|median|60.18|37.58|
|max|108.54|79.99|
|mean|67.96|43.08|

All the gains less than 1 are for tiny tensors less than 128 elements with less than 2 us execution time and can be just random fluctuations. Also "average metrics" are a bit useless because of a complex nature of a process - there are a few regions of interests and probably graphs may show better picture there.
Linear scale for whole picture. Maximum perf gain can be achieved around 16M of elements - around 128-192Mb of accessed data, which almost perfectly fits to L3 of gfx1030. Then it stabilizes on some value depended on the case.

gfx1030

Log scale is for small values. Power of 2 values show better gains due to no granularity loss. Tensors less than 1024 element show almost identical performance (gain around 1).

gfx1030_logscale

Raw data:
[OpTensor1d_gfx1030.xlsx](https://github.com/ROCmSoftwarePlatform/MIOpen/files/12074336/OpTensor1d_gfx1030.xlsx)

MI200 measurement is on the way, but I don't expect any performance drops compare to old version.

NB. There are no default perf tests for any tensor operations - as far as I can see only convolutions are tested.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -439,6 +439,7 @@ if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN
         kernels/xform_out.s
         kernels/gcnAsmBNBwdTrainSpatial.s
         kernels/MIOpenTensorKernels.cl
+        kernels/MIOpenTensorKernelsHip.cpp
         kernels/MIOpenSubTensorOpWithScalarKernel.cl
         kernels/MIOpenSubTensorOpWithSubTensorKernel.cl
         kernels/MIOpenSubTensorOpWithCastTensorKernel.cl
--- a/src/kernels/MIOpenTensorKernels.cl
+++ b/src/kernels/MIOpenTensorKernels.cl
@@ -1185,67 +1185,6 @@ __kernel void Op2dTensorGeneric(global MIOPEN_TYPE* a,
 
 #endif
 
-#ifdef USE_1D_TENSOR_GENERIC
-// N
-__kernel void Op1dTensorGeneric(global MIOPEN_TYPE* a,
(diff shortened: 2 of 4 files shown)


src/CMakeLists.txt
```diff
@@ -439,6 +439,7 @@ if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN
         kernels/xform_out.s
         kernels/gcnAsmBNBwdTrainSpatial.s
         kernels/MIOpenTensorKernels.cl
+        kernels/MIOpenTensorKernelsHip.cpp
```
JehandadKhan, line 442: Perhaps not add `Hip` to the filename since the same is indicated by the extension

JehandadKhan (dismissed):
Minor nitpick

src/CMakeLists.txt
CAHEK7: I've tried, I've got a name collision. Probably it would be better to rename opencl file to something else and keep hip filename more consistent.

junliume (approved):

PR #2263, opened 2023-07-18
[FIX SW 396203] check launch kernel grid size not beyond 32bit integer

carlushuang:
Our HIP launch kernel API support launch grid size no larger than 32bit integer (4294967295), otherwise HIP runtime will through exception.
We need to check this number and set inside `isApplicable()`/`isValid()`

### DIFF
--- a/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
@@ -828,7 +828,7 @@ bool PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC::IsValid(
     std::tie(current_block_size, current_grid_size, current_splits_4G) =
         GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(problem, *this);
 
-    if(current_block_size * current_grid_size * current_splits_4G > 0xffffffff)
+    if(current_block_size * current_grid_size * current_splits_4G > 0xffffffffULL)
         return false;
 
     bool unit_conv = (x == 1) && (y == 1) && (stride_h == 1) && (stride_w == 1) &&
--- a/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
@@ -382,12 +382,28 @@ GetBwdXdlopsNHWCConfigList()
     return kernel_param_list;
 }
 
-static std::tuple<std::string, // kernel_name
-                  size_t,      // block_size
-                  size_t,      // grid_size
-                  size_t>      // splits_4G
+// clang-format off
+static const inline PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC
+GetBwdXdlopsNHWCConfigLargestTileFp32()
+{
+    return {"bwd", "nhwc", miopenFloat,  0, 1, 256,  64,  16, 32, 32,  2, 1, 1, 2, 2, 1, 0, 0, 0, 0, { 1, 4, 4, 1}, {  1,  4,  1, 64}, { 1, 4, 1, 1}, {  1,  4,  1, 64}};
+}
+static const inline PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC
+GetBwdXdlopsNHWCConfigLargestTileFp16()
+{
+    return {"bwd", "nhwc", miopenHalf,  0, 1, 256, 256,  32, 32, 32,  8, 2, 2, 2, 2, 1, 0, 0, 0, 0, { 1, 8, 4, 1}, {  1,  4,  1, 64}, { 1, 8, 1, 4}, {  1,  4,  1, 64}};
+}
+static const inline PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC
+GetBwdXdlopsNHWCConfigLargestTileBf16()
+{
+    return {"bwd", "nhwc", miopenBFloat16,  0, 1, 256, 256,  32, 32, 32,  8, 2, 2, 2, 2, 1, 0, 0, 0, 0, { 1, 8, 4, 1}, {  1,  4,  1, 64}, { 1, 8, 1, 4}, {  1,  4,  1, 64}};
+}
+// clang-format on
+
+static std::tuple<size_t, // block_size
+                  size_t, // grid_size
+                  size_t> // splits_4G
 GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(
-    const ConvolutionContext& ctx,
     const ProblemDescription& problem,
     const PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC& config)
 {
@@ -441,12 +457,11 @@ GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(
         integer_divide_ceil(gemm_n, config.gemm_n_per_block) * (1 << config.gemm_k_global_split);
     if(config.multihead != 0)
         grid_size *= num_of_gemm;
-    std::string kernel_name = config.ToKernelName(ctx);
-    return std::make_tuple(kernel_name, block_size, grid_size, splits_4G);
+    return std::make_tuple(block_size, grid_size, splits_4G);
 }
 
 void PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC::HeuristicInit(
-    const ConvolutionContext& ctx, const ProblemDescription& problem)
+    const ConvolutionContext&, const ProblemDescription& problem)
 {
     static const std::vector<std::tuple<int, int, int>> tile_list_fp32 = {
         std::make_tuple(128, 128, 16),
@@ -693,8 +708,8 @@ void PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC::HeuristicInit(
                     }
                 }
                 size_t current_grid_size;
-                std::tie(std::ignore, std::ignore, current_grid_size, std::ignore) =
-                    GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(ctx, problem, config);
+                std::tie(std::ignore, current_grid_size, std::ignore) =
+                    GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(problem, config);
                 size_t gks = ComputeLog2GemmKGlobalSplitsWith2DMerge(current_grid_size,
                                                                      1200,
                                                                      k / group,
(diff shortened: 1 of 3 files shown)


src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -382,12 +382,28 @@ GetBwdXdlopsNHWCConfigList()
     return kernel_param_list;
 }
 
-static std::tuple<std::string, // kernel_name
-                  size_t,      // block_size
-                  size_t,      // grid_size
-                  size_t>      // splits_4G
+// clang-format off
+static const inline PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC
```
junliume: HIP Tidy complains about this line (and similar lines below):
```
error: return type 'const miopen::solver::PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC' is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
```

src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp
```diff
@@ -809,6 +824,13 @@ bool PerformanceConfigAsmImplicitGemmGTCBwdXdlopsNHWC::IsValid(
     if(problem.IsFp16() && gemm_k_global_split != 0 && vector_store != 1 && splits_4G > 1)
         return false;
 
+    size_t current_block_size, current_grid_size, current_splits_4G;
+    std::tie(current_block_size, current_grid_size, current_splits_4G) =
```
CAHEK7, line 828: Since we have got C++17 available, structure binding seems to be less wordy:
```
auto [current_block_size, current_grid_size, current_splits_4G] = GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(problem, *this);
```
vs
```
size_t current_block_size, current_grid_size, current_splits_4G;
std::tie(current_block_size, current_grid_size, current_splits_4G) =  GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(problem, *this);
```
But it doesn't allow to use `std::ignore`.

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
```diff
@@ -312,12 +312,28 @@ GetFwdXdlopsNHWCConfigList()
     return kernel_param_list;
 }
 
-static std::tuple<std::string, // kernel_name
-                  size_t,      // block_size
-                  size_t,      // grid_size
-                  size_t>      // splits_4G
+// clang-format off
+static inline PerformanceConfigAsmImplicitGemmGTCFwdXdlopsNHWC
+GetFwdXdlopsNHWCConfigLargestTileFp32()
```
CAHEK7, line 317: It seems that all the fp32\fp16\bfp16 configs are almost identical, so it would be better to dynamically calculate a couple of values instead of duplicating a lot of code.

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
```diff
@@ -875,6 +897,19 @@ bool ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC::IsApplicable(
                                    miopen::GetTypeSize(problem.GetInDataType())))
         return false;
 
+    {
+        auto largest_config = problem.IsFp32()
```
CAHEK7, line 901: If `GetFwdXdlopsNHWCConfigLargestTile` would depend on data size, we can simplify this code to
```
 auto largest_config = GetFwdXdlopsNHWCConfigLargestTile(dataSize);
```

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
```diff
@@ -683,6 +698,13 @@ bool PerformanceConfigAsmImplicitGemmGTCFwdXdlopsNHWC::IsValid(
     if(problem.IsFp16() && gemm_k_global_split != 0 && vector_store != 1 && splits_4G > 1)
         return false;
 
+    size_t current_block_size, current_grid_size, current_splits_4G;
+    std::tie(current_block_size, current_grid_size, current_splits_4G) =
```
CAHEK7, line 702: Since we have got C++17 available, structure binding seems to be less wordy:
```
auto [current_block_size, current_grid_size, current_splits_4G] = GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(problem, *this);
```
vs
```
size_t current_block_size, current_grid_size, current_splits_4G;
std::tie(current_block_size, current_grid_size, current_splits_4G) =  GetImplicitGemmGtcDynamicBwdXdlopsNHWCKernel(problem, *this);
```
But it doesn't allow to use std::ignore.

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
```diff
@@ -875,6 +897,19 @@ bool ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC::IsApplicable(
                                    miopen::GetTypeSize(problem.GetInDataType())))
         return false;
 
+    {
+        auto largest_config = problem.IsFp32()
+                                  ? GetFwdXdlopsNHWCConfigLargestTileFp32()
+                                  : (problem.IsFp16() ? GetFwdXdlopsNHWCConfigLargestTileFp16()
+                                                      : GetFwdXdlopsNHWCConfigLargestTileBf16());
+        size_t current_block_size, current_grid_size, current_splits_4G;
+        std::tie(current_block_size, current_grid_size, current_splits_4G) =
```
CAHEK7, line 906: Consider using structure binding here.

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
```diff
@@ -885,14 +920,15 @@ ConvSolution ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC::GetSolution(
...
 
-    std::string kernel_name;
     size_t block_size;
     size_t grid_size;
 
     int splits_4G;
 
-    std::tie(kernel_name, block_size, grid_size, splits_4G) =
-        GetImplicitGemmGtcDynamicFwdXdlopsNHWCKernel(ctx, problem, config);
+    std::tie(block_size, grid_size, splits_4G) =
```
CAHEK7, line 928: Consider using structure binding here.

CAHEK7 (changes requested):
Structure bindings are just a suggestion but making `Get***XdlopsNHWCConfigLargestTile()` dynamically configurable through a data size can improve code quality and should be addressed.

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
carlushuang: Actually the fp32 largest tile is `256x128x16`, fp16/bf15 is `256x256x32`, there is still a difference. In case in the future need to change the largest tile size per different data type, I think it make sense to keep it as now

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
carlushuang: Like described above, not exactly depending on data size.

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
CAHEK7: `256x(512/datasize)x(64/datasize)`?

carlushuang: I think to make a formula to calculate the tile size is intuitive, because tile size (a specific tuning parameter) does not have something to do with the data size.
Which means, if in the future, fp32 can provide a larger tile size like `256x192x16`, then the calculation will have wrong result

CAHEK7: Where `datasize` is 4 for `fp32` and 2 for `(b)fp16`. I think it is better to calculate it instead of thinking is it an actuall small difference in the largest tile or a typo.
There is also a case in the future where someone want to change one of the largest tile value for all the data types and forget one of the type or miscalculate offset of the value for one of the type.

CAHEK7: Ok, if you think that it doesn't depend on datasize at all and it is easier to change it in a multiple place, I'm fine with it.

CAHEK7 (approved):

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
carlushuang: I think the "calculation" approach make sense to host side maintainers, but for kernel developers there is really no relation ship between the largest tile size and the data type, like that example in above, things may change between different data types. It will confuse the kernel developers. And it's the kernel developer's responsibility to maintain the largest tile size table

Epliz:
Shouldn't you add a test if it is meant to fix a bug?

junliume:
> Shouldn't you add a test if it is meant to fix a bug?

@carlushuang could you deal with the above review comments? For test, I suggest very large tensor and force MISA kernels.

jithunnair-amd:
> Shouldn't you add a test if it is meant to fix a bug?

I second this comment whole-heartedly :)

src/solver/conv_asm_implicit_gemm_gtc_fwd_nhwc.cpp
atamazov: [Notice] @CAHEK7 @carlushuang I am still wondering why we use tuples as return types, which is error prone (sensitive to the order) and to enforces us to use comments to keep the code understandable (like this):
```C++
static std::tuple<size_t, // block_size
                  size_t, // grid_size
                  size_t> // splits_4G
GetImplicitGemmGtcDynamicFwdXdlopsNHWCKernel(
    const ProblemDescription& problem,
    const PerformanceConfigAsmImplicitGemmGTCFwdXdlopsNHWC& config);
```
instead of structures that contain members with "normal" names.

Wrt binding -- it is better sometimes, but looks as another crutch to me.

atamazov:
@JehandadKhan @junliume 
> Our HIP launch kernel API support launch grid size no larger than 32bit integer (4294967295), otherwise HIP runtime will through exception. We need to check this number and set inside `isApplicable()`/`isValid()`

Indeed, we do not check the grid size, this is known potential problem. Perhaps we have to systematically apply this kind of check to all existing solvers. What do you think?

atamazov (commented):
LGTM!

[Note] It seems a little questionable whether it's better to explicitly keep copies of the "largest configs" (as done in this PR) or just keep an index of the "largest config". Both approaches have drawbacks.

Issue #2217, opened 2023-06-22
Remove MIOpenGEMM and MIOpenTensile code 


JehandadKhan:
Both of the above GEMM backends have been deprecated and their code sits in the library still. Perhaps its prudent to just remove it and simplify the gemm backend code.

cc @junliume @atamazov

atamazov:
@JehandadKhan Sure! I think this is https://github.com/ROCmSoftwarePlatform/MIOpen/labels/quality https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_normal

atamazov:
@JehandadKhan Please assign @averinevg @DrizztDoUrden and me as well. The one who first has free time or desire will take it into work.

DrizztDoUrden:
I'd like to do this as it will reduce my headache from fixing bazillion corner-case code paths in [kernel binary serialization in solution](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2146)

averinevg:
MIOpenGEMM and MIOpenTensile leftovers were removed in https://github.com/ROCm/MIOpen/pull/2499

PR #2265, opened 2023-07-19
Removed miopengemm and tensile

DrizztDoUrden:
Implements #2217


DrizztDoUrden:
I don't know who else to add, this is trivial even though tedious

junliume (approved):

junliume:
Ping @JehandadKhan  before merging.

DrizztDoUrden:
Ping @JehandadKhan

src/gemm_v2.cpp
```diff
@@ -177,55 +169,24 @@ static GemmBackend_t enforce_gemm_backend(miopenDataType_t data_type,
     GemmBackend_t gemm_backend_env      = GemmBackend_t::nogemmbackend;
 
     // enforce backend based on env variable
+    // I have left the commented lines here to preserve values for the enforce and hint at why are
+    // they 1 and 3
```
JehandadKhan, line 173: I was hoping we would be able to get rid of this function.

src/include/miopen/gemm_v2.hpp
```diff
@@ -38,8 +38,6 @@ enum GemmBackend_t
 {
     nogemmbackend = 0,
     rocblas       = 1,
-    miopengemm    = 2,
-    miopentensile = 3,
```
JehandadKhan, line 42: This struct is redundant now!

JehandadKhan (approved):

JehandadKhan:
@DrizztDoUrden  The Jenkinsfile and Dockerfile might need to be updated as well, however, they are minor changes.

atamazov:
@JehandadKhan @junliume 🟠 I heard that we will most likely have to use something other than rocblas for one of the new GPUs. Therefore, I highly recommend keeping the existing infrastructure that allows us to switch between different GEMM backends, so as not to reinvent it later.

src/include/miopen/gemm_v2.hpp
atamazov: @JehandadKhan maybe not, please see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2265#issuecomment-1695963969

atamazov:
@DrizztDoUrden Please notice https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2265#issuecomment-1695963969

PR #2269, opened 2023-07-21
winograd fury convolution support

CAHEK7:
Initial support of a new winograd convolution solver (codename Fury).
It has completely different interface compared to the previous winograd solver.
Right now it supports only fp16 convolution with stride 1 and works only on gfx11.

<p>
<details>
<summary>Short perf table</summary>

|W|H|c|n|k|x|y|q|p|u|v|Default Conv Algorithm|Default Time|Fury Time|Gain|
|-|-|-|-|-|-|-|-|-|-|-|-|-|-|-|
|1920|1080|16|1|16|3|3|1|1|1|1|ConvBinWinogradRxSf2x3g1|0.47|0.29|1.60|
|1920|1080|8|1|16|3|3|1|1|1|1|ConvBinWinogradRxSf2x3g1|0.29|0.18|1.58|
|960|540|16|1|16|3|3|1|1|1|1|ConvBinWinogradRxSf2x3g1|0.09|0.07|1.21|
|480|270|16|1|16|3|3|1|1|1|1|ConvBinWinogradRxSf2x3g1|0.03|0.02|1.39|
|240|135|16|1|16|3|3|1|1|1|1|ConvBinWinogradRxSf2x3g1|0.02|0.01|1.26|
|480|270|4|1|16|1|1|0|0|1|1|GemmFwd1x1_0_1|0.02|0.01|1.60|
|960|540|4|1|16|1|1|0|0|1|1|GemmFwd1x1_0_1|0.05|0.03|1.44|
|1920|1080|4|1|16|1|1|0|0|1|1|GemmFwd1x1_0_1|0.17|0.10|1.68|
|1920|1080|16|1|4|3|3|1|1|1|1|ConvBinWinogradRxSf2x3g1|0.40|0.27|1.51|
</details>
</p>

Perf summary for 9 cases:

|metric|gain|
|-|-|
|average|1.48|
|total time|1.54|
|min|1.21|
|max|1.68|


JehandadKhan:
@CAHEK7  Please provide a useful description to this PR.

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+              uint32_t(o_strides.h), // todo: check < 2^16
+              uint32_t(o_strides.c),
+              data_tensors,
+              out_tensors,
+              filter_tensors,
+              conv.N,
+              conv.K,
+              conv.C,
+              conv.R,
+              conv.S,
```
Slimakanzer: These parameters are `size_t` defined in `UnifiedDescriptionConv2d`:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/c90371bba3b81c5debe0f903a34602453c4f1c1f/src/include/miopen/problem_description.hpp#L409-L413

convert them explicitly to `uint32_t`

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+              uint32_t(f_strides.nk),
+              uint32_t(f_strides.c),
+              int32_t(f_strides.h),
+              int32_t(f_strides.w),
+              f_RS_offset,
+              uint64_t(0),
+              nullptr,
```
Slimakanzer:
```suggestion
              nullptr,
              nullptr,
```

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+              flags,
+              activation_mode,
+              uint8_t(0),
+              uint16_t(0),
+              0.0);
```
Slimakanzer: Be careful, `0.0` is `double`
```suggestion
              0.0f);
```

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+            k(conv.n_groups,
+              conv.out_Hs,
+              conv.out_We,
+              conv.d_W_window,
+              conv.d_H_window,
+              conv.d_W_clip_bot_neg,
+              conv.d_H_clip_bot_neg,
+              uint32_t(d_strides.nk),
+              uint32_t(d_strides.h), // todo: check < 2^16
+              uint32_t(d_strides.c),
```
Slimakanzer: Will the type conversion inside invocation lambda create an overhead on kernel launch? If yes, then move it outside.

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+              "\nf_strides.w=" << f_strides.w <<
+              "\nf_RS_offset=" << f_RS_offset <<
+              "\nflags=" << flags <<
+              "\nactivation_mode=" << static_cast<uint32_t>(activation_mode) <<
+              "\n");
+            // clang-format on
+
+            k(conv.n_groups,
+              conv.out_Hs,
+              conv.out_We,
```
Slimakanzer:
```suggestion
              conv.out_We,
              conv.out_Hs,
```

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+                            : primitive_params.CastTo<conv::WrWInvokeParams>().tensors.dy;
+            const auto out_tensors =
+                !is_backWrW ? primitive_params.CastTo<conv::DataInvokeParams>().tensors.out
+                            : primitive_params.CastTo<conv::WrWInvokeParams>().tensors.dw;
+
+            // clang-format off
+            MIOPEN_LOG_I2(
+              "\nn_groups=" << conv.n_groups <<
+              "\nout_Hs=" << conv.out_Hs <<
+              "\nout_We=" << conv.out_We <<
```
Slimakanzer:
```suggestion
              "\nout_We=" << conv.out_We <<
              "\nout_Hs=" << conv.out_Hs <<
```

Slimakanzer (changes requested):

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: It heavily depends on compiler - usually it just moves data from one memory location(usually on stack) to another location(on stack as well), and for this particular cast it should use just different flavors of `mov` instruction, like `mov64 rax,[mem]; mov32 [mem], eax` or `mov32 eax,[mem]; mov32 [mem], eax`.

But I'll double check if there are any potential perf problems.

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: Already fixed in local version.

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: Already fixed in local version

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: As per the document Iliya gave me, it is reserved 64bit value, it's mostly doesn't matter is it pointer or not.
Is it intended to be a pointer semantically? Or it is generic 64bit value?

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: Oh, I've missed it..

src/CMakeLists.txt
```diff
@@ -478,6 +479,7 @@ if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN
         kernels/Conv_Winograd_v30_3_1_fp32_f3x2_dilation2.s
         kernels/Conv_Winograd_v30_3_1_fp32_f3x2_stride1.s
         kernels/Conv_Winograd_v30_3_1_fp32_f3x2_stride2.s
+        kernels/Conv_Winograd_Fury_v0_0_1_fp16_fp16acc_f2x3_stride1.s
```
Slimakanzer: The released version is `v1_0_0`

src/include/miopen/solver.hpp
```diff
@@ -2450,6 +2450,79 @@ struct ConvAsmBwdWrW3x3 final : ConvTunableSolver<PerformanceConfigAsmDirect3x3W
...
+    {
+        return IsValid(ctx);
+    }
+    bool IsValid(const ConvolutionContext&) const;
+    bool operator==(const PerformanceConfigConvWinoFuryRxS& other) const;
+};
+
+template <size_t Winodata, size_t Winofilter>
+struct ConvWinoFuryRxS final : ConvTunableSolver<PerformanceConfigConvWinoFuryRxS>
+{
```
Slimakanzer: Let's make this solver non-tunable as the number of CUs for the `n_groups` parameter is expected to be optimal.

Example of a non-tunable solver:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/5dc90222b9a36fecdcbc9467ec5404a96ec4bfe9/src/include/miopen/solver.hpp#L2116-L2127

src/include/miopen/solver.hpp
```diff
@@ -2450,6 +2450,79 @@ struct ConvAsmBwdWrW3x3 final : ConvTunableSolver<PerformanceConfigAsmDirect3x3W
...
+    const std::string& SolverDbId() const override { return GetSolverDbId(); }
+
+    static const std::string& GetSolverDbId()
+    {
+        static const std::string dbId = std::string("ConvWinoFuryRxSf")
+                                            .append(std::to_string(Winodata))
+                                            .append("x")
+                                            .append(std::to_string(Winofilter));
+        return dbId;
+    }
```
Slimakanzer: Renaming is not required.

```suggestion
    const std::string& SolverDbId() const override
    {
        return GetSolverDbId<ConvWinoFuryRxS<Winodata, Winofilter>>();
    }
```

src/kernels/Conv_Winograd_Fury_v0_0_1_metadata.inc
```diff
@@ -0,0 +1,192 @@
...
+    - { .size: 2, .offset:  34, .value_kind: by_value, .name: o_H_clip_bot_neg }
+    - { .size: 4, .offset:  36, .value_kind: by_value, .name: o_N_stride }
+    - { .size: 4, .offset:  40, .value_kind: by_value, .name: o_H_stride }
+    - { .size: 4, .offset:  44, .value_kind: by_value, .name: o_K_stride }
+
+    - { .size: 8, .offset:  48, .value_kind: global_buffer, .name: data_addr,   .address_space: global, .is_const: true }
+    - { .size: 8, .offset:  56, .value_kind: global_buffer, .name: output_addr, .address_space: global, .is_const: false }
+    - { .size: 8, .offset:  64, .value_kind: global_buffer, .name: filter_addr, .address_space: global, .is_const: true }
+
+    - { .size: 4, .offset:  72, .value_kind: by_value, .name: N }
```
Slimakanzer: May cause compilation errors (see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1830)

```suggestion
    - { .size: 4, .offset:  72, .value_kind: by_value, .name: BATCH_SIZE }
```

src/kernels/Conv_Winograd_Fury_v0_0_1_metadata.inc
```diff
@@ -0,0 +1,192 @@
...
+.macro METADATA sc,wc,wg_x, kernel_name
+.amdgpu_metadata
+---
+amdhsa.version: [ 1, 0 ]
+amdhsa.kernels:
+  - .name: \kernel_name
+    .symbol: \kernel_name\().kd
+    .language: "OpenCL C"
+    .language_version: [ 1, 2 ]
+    .lds_size: 65536
```
Slimakanzer: `.lds_size` is AMDPAL metadata and can be removed
```suggestion
```

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+    MIOPEN_LOG_I("Overridden from env: " << fromEnv.ToString());
+
+    return fromEnv;
+}
+
+class ShaderModel : public UnifiedDescriptionConv2d
+{
+    static constexpr uint32_t Tw  = 4;
+    static constexpr uint32_t Toh = 2;
+    static constexpr uint32_t Tow = 2;
```
Slimakanzer: These constants depend on the Winograd algorithm (`F2x3` or `F3x2`). Currently Winograd Fury supports only F2x3, but if `ShaderModel` is supposed to be generic, then these constants should be:

```cpp
Toh = Winodata;
Tow = Winodata;
Tw = Winodata + Winofilter - 1;
```

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+    uint32_t Hs;
+    uint32_t We;
```
Slimakanzer, line 55: These names can be confusing, as from the name I'd assume it's a "scaled" **input** height and "extended" **input** width but it's "scaled" **output** height and "extended" **output** width.

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+                size_t /*Winodata*/,
+                size_t /*Winofilter*/)
+        : UnifiedDescriptionConv2d(problem),
+          Hs{Ceil<uint32_t>(out_h, Toh)},
+          We{Tow * (Ceil<uint32_t>(out_w, Tow) + Ceil(Tw, Tow) - 1)},
+
+          H{uint32_t(problem.direction.IsBackwardWrW() ? problem.GetWeightsWidth()
+                                                       : problem.GetInWidth())},
+          W{uint32_t(problem.direction.IsBackwardWrW() ? problem.GetWeightsHeight()
+                                                       : problem.GetInHeight())},
```
Slimakanzer:
```suggestion
          W{uint32_t(problem.direction.IsBackwardWrW() ? problem.GetOutWidth()
                                                       : problem.GetInWidth())},
          H{uint32_t(problem.direction.IsBackwardWrW() ? problem.GetOutHeight()
                                                       : problem.GetInHeight())},
```

src/mlo_dir_conv.cpp
```diff
@@ -142,7 +142,8 @@ static auto GetWindogradSolvers()
                                            miopen::solver::ConvMPBidirectWinograd_xdlops<3, 3>,
                                            miopen::solver::ConvMPBidirectWinograd_xdlops<4, 3>,
                                            miopen::solver::ConvMPBidirectWinograd_xdlops<5, 3>,
-                                           miopen::solver::ConvMPBidirectWinograd_xdlops<6, 3>>{};
+                                           miopen::solver::ConvMPBidirectWinograd_xdlops<6, 3>,
+                                           miopen::solver::ConvWinoFuryRxS<2, 3>>{};
```
Slimakanzer, line 147: The solver is only registered for Fwd/Bwd Data directions. Winograd Fury can be executed for the WrW direction, but with very limited cases, and the kernel is not optimized for it. Should it then be applicable for the WrW direction?

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+                   group_cnt,
+                   GetTypeSize(problem.GetWeightsDataType()));
+
+    const auto d_strides = d_buf.stride;
+    const auto o_strides = o_buf.stride;
+
+    auto f_strides       = f_buf.stride;
+    f_strides.h          = is_forward ? f_strides.h : -f_strides.h;
+    f_strides.w          = is_forward ? f_strides.w : -f_strides.w;
+    uint32_t f_RS_offset = is_forward ? 0 : ((conv.R - 1) * conv.S + (conv.S - 1));
```
Slimakanzer: Filter flip is only needed for Bwd Data direction.

```suggestion
    f_strides.h          = is_forward || is_backWrW ? f_strides.h : -f_strides.h;
    f_strides.w          = is_forward || is_backWrW ? f_strides.w : -f_strides.w;
    uint32_t f_RS_offset = is_forward ? 0 : ((conv.R - 1) * conv.S + (conv.S - 1));
```

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+    if(!ctx.rmv.IsV3())
+        return false;
+
+    const auto target = ctx.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
+
+    const auto name = ctx.GetStream().GetDeviceName();
+    if(!StartsWith(name, "gfx11"))
+        return false;
```
Slimakanzer: Navi33 is not supported due to reduced VGPR RAM size compared to Navi31.

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,569 @@
...
+    else if(problem.GetKernelStrideW() == 2 && !problem.direction.IsBackwardData())
+    {
+        kernel_name << "_stride2";
+        kernel_file << "_stride2";
+    }
+    else // if(problem.GetDilationH() == 2)
+    {
+        kernel_name << "_dilation2";
+        kernel_file << "_dilation2";
+    }
```
Slimakanzer, line 282: We don't have any short-term plans for FP32/F3x2/stride2/dilation2 support, it can be simplified.

Slimakanzer (changes requested):

src/solver/conv_wino_fury_RxS.cpp
Slimakanzer: It is reserved for a pointer to an "accumulated" buffer for future kernel versions, in this case, `nullptr` fits better.

Slimakanzer:
Please add smoke tests for the new solver. Example:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/a0c73f544d3db86511f8a521a34198cba5939ac7/test/CMakeLists.txt#L2222

src/kernels/Conv_Winograd_Fury_v0_0_1_metadata.inc
CAHEK7: Do I need to set something else to allocate LDS?

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: Ok, thanks, that part was missing in the documentation and only those constants were provided.

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: I used the same names as in the document, at least it is easy to compare the code and the documentation.
I'd prefer to get updated documentation first.

src/mlo_dir_conv.cpp
CAHEK7: I don't know. Probably.
What are the limitations for WrW direction?

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: I'd rather keep the code since it's already in and it would require less changes in the future.

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: What is the best way to check it?

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: That's exactly the difference - it uses just different flavors of mov.

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: Does it support Navi32?

src/kernels/Conv_Winograd_Fury_v0_0_1_metadata.inc
Slimakanzer: `group_segment_fixed_size` is enough

src/mlo_dir_conv.cpp
CAHEK7: I mean is there any other specific limitations rather than the normal ones?

Slimakanzer: The limitations are the same, but as only filter size <= 3 is supported by Winograd Fury, for the WrW direction it will be cases where the output tensor sizes <= 3 (in the WrW direction, we swap filter and output tensors to get the equivalent forward parameters). 

Example: for WrW case `-W 24 -H 24 -c 16 -n 1 -k 16 -x 22 -y 22 -p 0 -q 0 -u 1 -v 1` equivalent forward convolution parameters would be `-W 24 -H 24 -c 1 -n 16 -k 16 -x 3 -y 3 -p 0 -q 0 -u 1 -v 1`.

It would be interesting to parse find-db and see if there are any cases like this.

src/solver/conv_wino_fury_RxS.cpp
Slimakanzer: This kernel can be executed on Navi32 (gfx1101).

src/mlo_dir_conv.cpp
Slimakanzer: Let's disable this solver for the WrW direction for now.

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,470 @@
...
+                   group_cnt,
+                   GetTypeSize(problem.GetWeightsDataType()));
+
+    const auto d_strides = d_buf.stride;
+    const auto o_strides = o_buf.stride;
+
+    auto f_strides       = f_buf.stride;
+    f_strides.h          = is_forward || is_backWrW ? f_strides.h : -f_strides.h;
+    f_strides.w          = is_forward || is_backWrW ? f_strides.w : -f_strides.w;
+    uint32_t f_RS_offset = is_forward ? 0 : ((conv.R - 1) * conv.S + (conv.S - 1));
```
Slimakanzer:
```suggestion
    uint32_t f_RS_offset = is_forward || is_backWrW ? 0 : ((conv.R - 1) * conv.S + (conv.S - 1));
```

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,470 @@
...
+        // clang-format on
+    }
+};
+
+} // namespace
+
+template <uint32_t Winodata, uint32_t Winofilter>
+bool ConvWinoFuryRxS<Winodata, Winofilter>::IsApplicable(const ConvolutionContext& ctx,
+                                                         const ProblemDescription& problem) const
+{
```
Slimakanzer, line 170: Please check explicitly unsupported WrW direction in `IsApplicable`.

test/CMakeLists.txt
```diff
@@ -2262,6 +2262,18 @@ add_custom_test(smoke_solver_ConvBinWinogradRxSf3x2_f32 GFX103X_ENABLED GFX110X_
       ${TEST_CONV_VERBOSE_W} --input 1 20 20 20 --weights 20 20 3 3 --pads_strides_dilations 1 1 1 1 1 1 ${MIOPEN_TEST_FLAGS_ARGS}
 )
 
+# FP16 ALT attribute is disabled to enable the backward solver on MI200 for HALF.
+add_custom_test(smoke_solver_ConvWinogradFuryRxSf3x2_f16 GFX110X_ENABLED FLOAT_DISABLED HALF_ENABLED SKIP_XNACK_ON
+    COMMAND MIOPEN_DEBUG_CONVOLUTION_ATTRIB_FP16_ALT_IMPL=0
+      MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER='ConvWinoFuryRxS<2-3>' $<TARGET_FILE:test_conv2d>
+      ${TEST_CONV_VERBOSE_F} --input 1 16 16 16 --weights 16 16 3 3 --pads_strides_dilations 1 1 1 1 1 1 --trans_output_pads 1 1
+      ${MIOPEN_TEST_FLAGS_ARGS}
+    COMMAND MIOPEN_DEBUG_CONVOLUTION_ATTRIB_FP16_ALT_IMPL=0
```
Slimakanzer: FP16 ALT attribute is only needed for MI200. It can be removed from Winograd Fury tests.

test/CMakeLists.txt
```diff
@@ -2262,6 +2262,18 @@ add_custom_test(smoke_solver_ConvBinWinogradRxSf3x2_f32 GFX103X_ENABLED GFX110X_
       ${TEST_CONV_VERBOSE_W} --input 1 20 20 20 --weights 20 20 3 3 --pads_strides_dilations 1 1 1 1 1 1 ${MIOPEN_TEST_FLAGS_ARGS}
 )
 
+# FP16 ALT attribute is disabled to enable the backward solver on MI200 for HALF.
+add_custom_test(smoke_solver_ConvWinogradFuryRxSf3x2_f16 GFX110X_ENABLED FLOAT_DISABLED HALF_ENABLED SKIP_XNACK_ON
```
Slimakanzer:
```suggestion
# FP16 ALT attribute is disabled to enable the backward solver on MI200 for HALF.
add_custom_test(smoke_solver_ConvWinogradFuryRxSf2x3_f16 GFX110X_ENABLED FLOAT_DISABLED HALF_ENABLED SKIP_XNACK_ON
```

src/solver/conv_wino_fury_RxS.cpp
CAHEK7: I thought about it. Ok.

CAHEK7: Applied in the constructor

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,457 @@
...
+          Tw{Winodata + Winofilter - 1},
+          Toh{Winodata},
+          Tow{Winodata},
+          Hs{Ceil<uint32_t>(out_h, Toh)},
+          We{Tow * (Ceil<uint32_t>(out_w, Tow) + Ceil(Tw, Tow) - 1)},
+
+          W{static_cast<uint32_t>(problem.direction.IsBackwardWrW() ? problem.GetInHeight()
+                                                                    : problem.GetOutWidth())},
+          H{static_cast<uint32_t>(problem.direction.IsBackwardWrW() ? problem.GetInWidth()
+                                                                    : problem.GetOutHeight())},
```
Slimakanzer:
```suggestion
          W{static_cast<uint32_t>(problem.direction.IsBackwardWrW() ? problem.GetOutWidth()
                                                                    : problem.GetInWidth())},
          H{static_cast<uint32_t>(problem.direction.IsBackwardWrW() ? problem.GetOutHeight()
                                                                    : problem.GetInHeight())},
```

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -0,0 +1,457 @@
...
+    if(!ctx.rmv.IsV3())
+        return false;
+
+    const auto target = ctx.GetStream().GetTargetProperties();
+    if(target.Xnack() && *target.Xnack())
+        return false;
+
+    const auto name = ctx.GetStream().GetDeviceName();
+    if(!(StartsWith(name, "gfx1100") || StartsWith(name, "gfx1101")))
+        return false;
```
Slimakanzer: We do not need StartsWith when comparing for equality.

```suggestion
    if(!(name == "gfx1100" || name == "gfx1101"))
        return false;
```

Slimakanzer (changes requested):

CAHEK7:
More perf tests:
[fury-perf.xlsx](https://github.com/ROCmSoftwarePlatform/MIOpen/files/12396559/fury-perf.xlsx)

Summary:

|metric|value|
|-|-|
|minimum gain|0.58|
|median gain|1.41|
|total time gain|1.5|
|average gain|1.53|
|maximum gain|3.49|
|gain<0.95 cases|73 (6%)|
|total cases|1239|

CAHEK7:
All the 1239 additional + 18 initial (9 as in the table and 9 with swapped H and W) cases have passed the verification.

Slimakanzer (approved):
LGTM

CAHEK7:
@junliume can you take a look\merge?

JehandadKhan:
@junliume A gentle ping to merge this PR.

junliume (approved):

PR #2271, opened 2023-07-24
bg/lwpmiopen 194 : Integrate CK's batch norm inference into Non-tunable MIOpen solver. 

bghimireamd:
* Made batch norm tests modular
* Add new test for batch norm inference
* NHWC layout


JehandadKhan:
@bghimireamd Can we add a few test cases for the tuning loops ?

bghimireamd:
> Member

junliume:
@bghimireamd please resolve the conflicts. Thanks!

junliume (approved):
@JehandadKhan @atamazov final ping before merging for any reviews

test/gtest/bn.hpp
```diff
@@ -0,0 +1,86 @@
...
+{
+protected:
+    void SetUp() override
+    {
+        test_skipped                       = false;
+        std::tie(bn_config, tensor_layout) = GetParam();
+        bn_infer_test_data.SetUpImpl(bn_config, tensor_layout);
+
+        auto&& handle = get_handle();
+        miopenBatchNormalizationForwardInference(&handle,
```
JehandadKhan, line 49: It would be preferable to call the solver in the test body instead of using the public API.

test/gtest/test_operations.hpp
```diff
@@ -0,0 +1,110 @@
...
+                               dl_module.activ_gamma),
+              miopenStatusSuccess);
+}
+
+bool Skip(miopen::Handle& handle)
+{
+    const std::string arch = handle.GetDeviceName();
+    bool skip_test         = (arch != "gfx908" && arch != "gfx90a");
+
+    return skip_test;
```
JehandadKhan, line 107: I would suggest we keep these in the tests themselves so that we can document why a test is being skipped inside the test. Centralizing this logic would complicate it very quickly and then make it hard to determine why a test is skipped or failing etc.

JehandadKhan (approved):

CAHEK7:
I've missed it..
So as usual it's a candidate to #2379 and we should avoid to use `std::random_device` for random seed.
I'll take care of `std::random_device` because there are a lot of other places to fix, but I'd appreciated if somebody can apply changes from #2379 to this solver.

bghimireamd:
> I've missed it.. So as usual it's a candidate to #2379 and we should avoid to use `std::random_device` for random seed. I'll take care of `std::random_device` because there are a lot of other places to fix, but I'd appreciated if somebody can apply changes from #2379 to this solver.

I will add the changes from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2379 to this solver.

PR #2272, opened 2023-07-25
[DRAFT] enable MIOpen on Windows

apwojcik:
PR to enable compilation of MIOpen on Windows. The list of changes compared to Linux:

* Added Windows-specific code,
* By default, on Windows, the dependencies are downloaded from remote repositories and compiled during the MIOpen build
    * one-time operation during the first build - only external dependencies, the internal, such as rocMLIR, CK, etc., are built separately (llvm-project, rocMLIR, and other ROCm required components are treated as a single project on Windows)
    * options exist to use system-installed libraries instead of downloading and building them - can be turned on for all or per individual library.
* Installation of *.db files moved to ${CMAKE_CURRENT_BINAR_DIR} instead of source directory
    *  on Windows, multi-config generators run the installation of kernels in parallel for each configuration. This leads to access violations because another process is busy with a DB file, and the file is locked.
    * that is an overhead and data redundancy. Still, we cannot control it otherwise on Windows with multi-config generators (such as MSVC multi-config generator or even Ninja run as a multi-config generator).
* Added option to let switch building the tests on or off - it is on by default.
* Added default paths for global and local kernels cache on Windows.
* Added CMake command `miopen_generate_export_header()` - using `WINDOWS_EXPORT_ALL_SYMBOLS` is dangerous. It exposes the public exports, not only the API. Furthermore, more importantly, the number of symbols to export is limited to 64K per single DLL on Windows.
    * because tests require public exports, and on Windows, we want to export only API from the DLL, the `MIOpen` target was split into the `MIOpenCore` static library and the `MIOpen` dynamic library. Both tests and `MIOpen` DLL link to `MIOpenCore`.
* The file `cmake/googletest.cmake` was renamed to `cmake/google-test.cmake` - the Windows file system is case insensitive. There's an official module in CMake named `GoogleTest`. That led to configuration failures because CMake took an incorrect module during the import.
* `sequences` have been upgraded to be used with the Standard Library algorithms, functions, and containers.


averinevg:
@atamazov Please review

.gitignore
```diff
@@ -0,0 +1,7 @@
+src/kernels/*.db
+src/kernels/*.kdb
+
+# JetBrains IDEs
+.idea/
+cmake-build*/
+build*/
```
averinevg: These lines should'n be there, because developers use different IDEs and we can't add all of them

CMakeLists.txt
```diff
@@ -23,7 +23,7 @@
 # SOFTWARE.
 #
 ################################################################################
-cmake_minimum_required( VERSION 3.11 )
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
```
averinevg: The FATAL_ERROR option is accepted but ignored by CMake 2.6 and higher

CMakeLists.txt
```diff
@@ -37,23 +37,27 @@ macro(set_var_to_condition var)
...
+        set(CMAKE_BUILD_TYPE Release CACHE STRING
+            "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel.")
+    endif()
 endif()
 
 project ( MIOpen C CXX )
 
 enable_testing()
 
+find_package(Threads REQUIRED)
```
averinevg, line 69: Do we need this package for MIOpen library itself?

cmake/Findzstd.cmake
```diff
@@ -0,0 +1,65 @@
+# Try to find the zstd library
```
averinevg: Please add license header

cmake/google-test.cmake
```diff
@@ -0,0 +1,45 @@
+include(FetchContent)
```
averinevg: Please add license header

averinevg (changes requested):
Size of long integer is 4 bytes on 64-bit Windows and 8 bytes on 64-bit Linux. We have a lot of places in the code where long integers are used. You should replace all of them with exact-width integer types from stdint (or size_t in some cases) or somehow force compiler to treat them as 8 bytes integers.

test/verify.hpp
```diff
@@ -77,8 +77,11 @@ struct not_finite_fn
     template <class T>
     bool operator()(T x) const
     {
-        using std::isfinite;
-        return not isfinite(x);
+        // WIN32: The Standard Library from MSVC does implement std::isfinite()
+        // for floating-point types only - no additional overloads for integer
+        // types, which should be treated as doubles according to the specification.
+        // https://en.cppreference.com/w/cpp/numeric/math/isfinite
+        return not std::isfinite(static_cast<double>(x));
```
CAHEK7, line 89: Casting `T` to `double` may break the logic. If `T` is denormal `float` will it be denormal after conversion to `double`?
I would recommend `operator()` specializing for integer types and always return `true`.

test/tensor_vec.cpp
```diff
@@ -29,7 +29,9 @@
 #include <iterator>
 #include <limits>
 #include <memory>
+#if defined(__linux__)
 #include <sys/time.h>
```
CAHEK7: Actually, it is not required here and can be completely removed.
Seems that header has been copy-pasted across all the test_*cpp files, but it is not required in any of them.
Could you just remove it everywhere instead of enabling it on linux?

test/rnn_vanilla_common.hpp
```diff
@@ -540,8 +540,11 @@ void RNNFwdTrainCPUVerify(miopen::Handle& handle,
     {
         for(int h = 0; h < out_h; h++)
         {
-            assert(!std::isnan(rsvspace.at(prelayer_shift + bs * hy_stride + h)));
-            assert(!std::isinf(rsvspace.at(prelayer_shift + bs * hy_stride + h)));
+            assert(
+                !std::isnan(static_cast<double>(rsvspace.at(prelayer_shift + bs * hy_stride + h))));
+            assert(
+                !std::isinf(static_cast<double>(rsvspace.at(prelayer_shift + bs * hy_stride + h))));
```
CAHEK7, line 546: Probably it will work for the most cases, but it still performs the conversion and   it may wrongly report that value is infinity.

test/perfdb.cpp
```diff
@@ -989,16 +994,55 @@ class DbMultiProcessTest : public DbTest
                 if(full_set())
                     command += " --all";
 
+#if defined(WIN32)
```
CAHEK7: It would be nice to keep that code somewhere in one place. I don't insist but looks like it could be a bit better.

test/include_inliner.cpp
```diff
@@ -54,12 +54,32 @@ static int Child(const std::string& path, std::string cmd)
...
-    if(!CreateProcessA(
-           path, &cmd[0], nullptr, nullptr, false, 0, nullptr, nullptr, &info, &processInfo))
-        MIOPEN_THROW("CreateProcess error: " << GetLastError());
+    ZeroMemory(&info, sizeof(info));
+    info.cb = sizeof(info);
+    ZeroMemory(&processInfo, sizeof(processInfo));
+
+    LPSTR lpCmdLn{};
+
+#if !defined(UNICODE)
```
CAHEK7, line 66: This check is missed for all the other cases where we use CreateProcess.
That's why I think it would be better to keep that code in a one place and reuse it everywhere.

test/find_2_conv.cpp
```diff
@@ -69,7 +69,7 @@ struct Find2Test : test_driver
             "workspace_limit",
             generate_data({
                 std::numeric_limits<std::size_t>::max(),
-                0ul,
+                (uint64_t)0,
```
CAHEK7: I'm not sure about this particular project, but I would avoid c-style casts.

CAHEK7 (changes requested):

.gitignore
apwojcik: why?

apwojcik: My question is because that is precisely the purpose of .gitignore. Those files typically are huge in other projects. IDEs add a lot of pollution to the project tree. Anyway, we will be adding Visual Studio Code and Visual Studio IDE in other PRs.

CMakeLists.txt
apwojcik: Good catch, thanks!

CMakeLists.txt
apwojcik: Yes, it is linked to the MIOpen dynamic library/shared object, MIOpenDriver, and tests.

test/rnn_vanilla_common.hpp
apwojcik: According to the specification, all integer types in additional overloads are treaded as `doubles` - reference https://en.cppreference.com/w/cpp/numeric/math/isnan. You do not provide a custom overload, e.g., for denormal, so here, the integer value will be implicitly converted to `double`, and then the floating-point version is called.

test/verify.hpp
apwojcik: According to the specification, all integer types in additional overloads are treaded as doubles - reference https://en.cppreference.com/w/cpp/numeric/math/isnan. You do not provide a custom overload, e.g., for denormal, so here, the integer value will be implicitly converted to double, and then the floating-point version is called.

apwojcik: The `double` still has enough spare bits in the exponent to represent all float denormalized numbers as normalized.

.gitignore
apwojcik: Added the final version to this PR.

cmake/Findzstd.cmake
apwojcik: done

cmake/google-test.cmake
apwojcik: done

.gitignore
averinevg: > My question is because that is precisely the purpose of .gitignore.

Take a look at Linux's [.gitignore](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/.gitignore). There is nothing associated with IDEs. IDE typically creates only one directory, and this is not a big deal. You can add your favorite IDE in your own .gitignore.

Also developers usually have several build directories for convenience, so adding build directory to .gitignore makes no sense.

CMakeLists.txt
averinevg: In that case, why didn't this dependency exist before?

.gitignore
apwojcik: FYI https://github.com/github/gitignore/tree/main

averinevg: > FYI https://github.com/github/gitignore/tree/main

Does this have anything to do with our project and will it benefit?

averinevg: I mean that any action should have a specific purpose. If there is no purpose, then it's just garbage code. What is the purpose in this case?

apwojcik: We prefer to keep it as is in this PR. Thank you.

CMakeLists.txt
apwojcik: Windows requires that dependency. We build a single MIOpen.dll - that is the requirement in the project. It does not harm Linux.

.gitignore
averinevg: Did I understand you correctly that you refuse to discuss this?

CMakeLists.txt
averinevg: Then this dependency must be only for windows

apwojcik: > In that case, why didn't this dependency exist before?

Actually it existed in Linux before -> `${CMAKE_THREAD_LIBS_INIT}`
https://cmake.org/cmake/help/latest/module/FindThreads.html

.gitignore
apwojcik: > I mean that any action should have a specific purpose. If there is no purpose, then it's just garbage code. What is the goal in this case?

The purpose is the `git` command to ignore project files and directories and build artifacts. So the team working on the project is more productive, and it takes less time for them to manage code changes. It also helps to automate that process and use various tools to facilitate that, which we use in our work.

averinevg: > The purpose is the `git` command to ignore project files and directories and build artifacts. So the team working on the project is more productive, and it takes less time for them to manage code changes. It also helps to automate that process and use various tools to facilitate that, which we use in our work.

It is a bit controversial, especially part with automation. I am used to checking my changes when I commit them instead of automatically adding all changes to a new commit.

I would also like to note that these changes have nothing to do with the problem that this pull request is supposed to solve, and such kind of changes shouldn't be there to not to overcomplicate this important PR.

apwojcik: > Did I understand you correctly that you refuse to discuss this?

I am not. I just have not to noticed your question. However, I prefer not to have discussions just for the sake of it. I value your time and mine as well.

atamazov: > We prefer to keep it as is in this PR. Thank you.

@junliume this kind of feedback to review comments is not acceptable. All the review comments shall be resolved before merging, one way or another.

atamazov (changes requested):
All the review comments must be properly resolved prior merging.

test/tensor_vec.cpp
apwojcik: done

test/verify.hpp
averinevg: Let's make it clear that this is a workaround

CAHEK7: The most significant disadvantage is existent `T` to `double` conversion - I don't think that it's a performance critical code but returning `true` for integers and call native functions for other types seems a bit better from performance point of view.

Also I've just realized that there is a one potential problem - if fp exceptions are turned on, we may get one during conversion. I don't think that MIOpen checks fpenv anywhere or turns on explicit fp exceptions, but the behavior is different for `std::isfinite(static_cast<double>(std::numeric_limits<uint64_t>::max()))` and `std::isfinite(std::numeric_limits<uint64_t>::max())`. It may alter fp state even if you work with integers.

CAHEK7: I guess we can discuss this problem and decide is it critical or not and probably put a few comments that this is fine.

dmikushin:
Hi @apwojcik , thank you for working on Windows support. Having MIOpen on Windows is an important milestone, which also improves the quality of our software.

I have two general questions about this PR.

1. Why bother about MSVC compiler, if we have ROCm clang, and ROCm dependency in any case?  Getting our code to support only one compiler is highly preferred. Spending resources on MSVC is questionable, especially given that Microsoft will fade it away sooner or later, in favor of their own Clang flavor. Getting Windows build onto ROCm Clang will also resolve many issues currently being debated here in the comments.
2. This Windows port relies on the release version of rocMLIR, whereas our MIOpen docker containers are usually based on a specific rocMLIR commit.

Both questions lead to the following issue. Given that our customers will intuitively expect the same MIOpen behavior on Windows and Linux, the approach taken by this PR may lead to significant misunderstanding, a lot of extra work and communication. In this regard, @apwojcik do you agree that Windows and Linux builds should be unified further as much as possible to minimize the risks?

apwojcik:
@dmikushin I invited you to today's meeting

CAHEK7:
May I ask to not to force-push commits? It's hard to track what happened in-between and since the number of changes is a quite huge, it's hard to re-review everything from the scratch.

cmake/Findzstd.cmake
```diff
@@ -0,0 +1,91 @@
+################################################################################
+#
+# MIT License
+#
+# Copyright (c) 2017 Advanced Micro Devices, Inc.
```
averinevg:
```suggestion
# Copyright (c) 2023 Advanced Micro Devices, Inc.
```

cmake/google-test.cmake
```diff
@@ -0,0 +1,79 @@
+################################################################################
+#
+# MIT License
+#
+# Copyright (c) 2017 Advanced Micro Devices, Inc.
```
averinevg:
```suggestion
# Copyright (c) 2023 Advanced Micro Devices, Inc.
```

apwojcik:
> May I ask to not to force-push commits? It's hard to track what happened in-between and since the number of changes is a quite huge, it's hard to re-review everything from the scratch.

I am sorry for that. We are trying to pass the compilation on CI and want the branch clean without intermediate commits. Unfortunately, we do not have a local copy of your  CI to test the compilation. We will have it soon. As soon as the compilation passes, the fixes for review comments and failing tests will be added incrementally.

johnnynunez:
@saadrahim This is very close to being implemented? It would be the final step to have amd gpus in AI for windows

test/rnn_vanilla_common.hpp
atamazov, line 546: @apwojcik Why this is necessary? Is it for `half`, `bfloat16` or for integer type?

atamazov:
> apwojcik force-pushed the uif2-initial branch from 42aebc4 to 98531eb 5 days ago

@apwojcik ⚠️ **_Please do not force-push_**. This is NOT recommended due to several reasons. For example, this makes previous code reviews invisible on the "Code changes" tab (due to overwritten commit hashes). Force-pushing can only be used more or less safely if the branch has NOT yet been published for review.

atamazov:
@apwojcik @junliume 

Due to a huge size of the PR I am expecting long review time and high probability of merge conflicts. In order to avoid this and speed up the process, I highly recommend splitting the PR to several ones. Those could be:
- A PR with some "cosmetic" stuff intended to improve correctness or to get rid of compilation warnings
- Based on previous PR, another PR with some more substantial changes that are common for both Windows and Linux
- Based on previous PR, the final one, a Windows-specific PR.

All branches/PRs can be developed in parallel (git supports this), but the code review will need to be done sequentially.

atamazov:
@junliume @JehandadKhan Can we assign some ungency and value labels?

MidnightKittenCat:
Any idea as to how long this may or may not take?

Love the work you're doing for the ROCm community <3

atamazov:
@apwojcik wrt [splitting the PR/branch](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2272#issuecomment-1658690347) and [resolving conficts](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2272#issuecomment-1658656494) -- please let me know if you need some technical assistance or other help.

apwojcik:
@atamazov @johnnynunez @MidnightKittenCat I am on vacation right now. The Windows port has no hard deadline, but the plan is to do it sooner rather than later. I will keep working on the PR  (i.e., splitting it), passing CI, and resolving conflicts after I return from vacation.

atamazov:
@apwojcik Then I recommend converting this to draft. Have a good vacation time!

apwojcik:
The PR is in the process of splitting into several smaller PRs. I will submit them soon. As requested, I converted cumulative PR into draft PR. The review comment from this PR will be incorporated into respective PRs.

Foxlum:
@apwojcik I would personally like to know for the meantime how this could be built early, if possible, as I've been trying to go through the process myself and coming upon issues. (Though I'm not an expert at the build process for MIOpen) 

I am fully aware this Draft PR and subsequent split PRs could rapidly change this porting effort, but MIOpen is quite vital to quite a few AI-related projects, and almost rabidly wanted by Windows users for proper first-class support of their cards for those workflows. And while this may or may not allow those workflows in the meantime (I'm not even sure if it is functional.), it would still be invaluable to have.

johnnynunez:
@apwojcik How is this pull request going? I would love to see next September/October with ROCM 5.7 to see the 7900xtx.

MidnightKittenCat:
Any changes so far? Are we perhaps scheduled for October-November?

scarsty:
mark...

xTye:
Willing to contribute in any way I can.

johnnynunez:
When will it be merged all the PR that has been opened for windows? @apwojcik

scarsty:
As the review is too slow, I have built a MIOpen.dll with the branch of uif2-initial, and it seems work well on 7900xtx. Please see <https://github.com/ROCmSoftwarePlatform/MIOpen/discussions/2375#discussioncomment-7142335>

MidnightKittenCat:
> As the review is too slow, I have built a MIOpen.dll with the branch of uif2-initial, and it seems work well on 7900xtx. Please see #2375

This is great! what does this mean for the windows port exactly? are we close to a full release?

scarsty:
I am not an official. Everyone knows that MIOpen is one of the most important component to ROCm and Windows is the most widely used platforms for business and personal use. After the releasing of Rocm of Windows Edition, I have paid attention to this for a period of time but the review progress is worrying.

scarsty:
> > As the review is too slow, I have built a MIOpen.dll with the branch of uif2-initial, and it seems work well on 7900xtx. Please see #2375
> 
> This is great! what does this mean for the windows port exactly? are we close to a full release?

I am not an official. Everyone knows that miopen is one of (maybe we can remove "one of") the most important component to rocm and Windows is the most widely used platform for business and personal use. After the releasing the rocm of Windows Edition, I have paid attention to this for a period of time but the review progress is worrying.

NeedsMoar:
> > > As the review is too slow, I have built a MIOpen.dll with the branch of uif2-initial, and it seems work well on 7900xtx. Please see #2375
> > 
> > 
> > This is great! what does this mean for the windows port exactly? are we close to a full release?
> 
> I am not an official. Everyone knows that miopen is one of (maybe we can remove "one of") the most important component to rocm and Windows is the most widely used platform for business and personal use. After the releasing the rocm of Windows Edition, I have paid attention to this for a period of time but the review progress is worrying.

Just out of curiosity, has anyone talked to the ProRender people to find out where the MIOpen.dll they're shipping with every render plugin on Windows is coming from?  It appears to be complete as far as the API is concerned and exported functions (not something stripped down).  I think ProRender uses it run denoising mainly, I'm not sure what else.  

Glancing at the Houdini and Blender plugins, version is v2.0.5.0 / ~7MB / built 11/3/2022

scarsty:
> ProRender

that miopen.dll is using OpenCL as backend, which is not good for the new version of rocm.

scarsty:
> > > > As the review is too slow, I have built a MIOpen.dll with the branch of uif2-initial, and it seems work well on 7900xtx. Please see #2375
> > > 
> > > 
> > > This is great! what does this mean for the windows port exactly? are we close to a full release?
> > 
> > 
> > I am not an official. Everyone knows that miopen is one of (maybe we can remove "one of") the most important component to rocm and Windows is the most widely used platform for business and personal use. After the releasing the rocm of Windows Edition, I have paid attention to this for a period of time but the review progress is worrying.
> 
> Just out of curiosity, has anyone talked to the ProRender people to find out where the MIOpen.dll they're shipping with every render plugin on Windows is coming from? It appears to be complete as far as the API is concerned and exported functions (not something stripped down). I think ProRender uses it run denoising mainly, I'm not sure what else.
> 
> Glancing at the Houdini and Blender plugins, version is v2.0.5.0 / ~7MB / built 11/3/2022

that miopen.dll is using OpenCL as backend, which is not good for the new version of rocm.

NeedsMoar:
> > > > > As the review is too slow, I have built a MIOpen.dll with the branch of uif2-initial, and it seems work well on 7900xtx. Please see #2375
> > > > 
> > > > 
> > > > This is great! what does this mean for the windows port exactly? are we close to a full release?
> > > 
> > > 
> > > I am not an official. Everyone knows that miopen is one of (maybe we can remove "one of") the most important component to rocm and Windows is the most widely used platform for business and personal use. After the releasing the rocm of Windows Edition, I have paid attention to this for a period of time but the review progress is worrying.
> > 
> > 
> > Just out of curiosity, has anyone talked to the ProRender people to find out where the MIOpen.dll they're shipping with every render plugin on Windows is coming from? It appears to be complete as far as the API is concerned and exported functions (not something stripped down). I think ProRender uses it run denoising mainly, I'm not sure what else.
> > Glancing at the Houdini and Blender plugins, version is v2.0.5.0 / ~7MB / built 11/3/2022
> 
> that miopen.dll is using OpenCL as backend, which is not good for the new version of rocm.

Thanks for clearing that up.  I can picture multiple ways in which OpenCL isn't the best front-end language to describe the basic data types of tensors never mind that it doesn't seem to be worked on with the goal of at least 1M new vendor extensions per year like...  I dunno, Vulkan.  

I have both a 7900XTX and an RTX4090 in my primary workstation now so if you need tests run on code which dispatches the equivalent ops to both (I thought I read that it was a thing with ROCm & HIP + PyTorch, which is heavily dependent on this AFAIK; it certainly works with the ncnn-vulkan runners / networks but that's a whole other thing) either to compare results, benchmark, or just make sure it doesn't explode when the combination exists after there's a functional build going, feel free to ping me.

JehandadKhan:
@apwojcik Does this PR and the other windows related PRs overlap or are they exclusive?

.gitignore
Joly0: > > My question is because that is precisely the purpose of .gitignore.
> 
> Take a look at Linux's [.gitignore](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/.gitignore). There is nothing associated with IDEs. IDE typically creates only one directory, and this is not a big deal. You can add your favorite IDE in your own .gitignore.
> 
> Also developers usually have several build directories for convenience, so adding build directory to .gitignore makes no sense.

If i am allowed to give my opinion on this one: I think your statement doesnt make a lot of sense. The .gitignore file is to prevent git from adding files/folders to your git commits. So it makes absolute sense to add the various folders/files that the different IDEs add to the project tree get excluded, otherwise you would have to manually exclude them everytime. This is especially important if you have different developers with different IDEs like here, where every single developer would have to exclude every single file/folder from their commits manually every time. The .gitignore and the changes @apwojcik made to it are very useful imo.

The linux kernel does not neccesarily need such a precise .gitignore, as linux kernel development is done mostly by using tools like vim, which doesnt add this bulk to the project tree. Big ide´s like jetbrains IDEs or visual studio add this bulk and therefore would need to be excluded every time. I mean, you can use a local .gitignore, but why make it harder for everyone, if there could just be a single .gitignore, that correctly excludes everything?

And the issue here is, if there isnt a centrally managed .gitignore which excludes those IDEs and someone forgets to exclude e.g. the build folder, this folder is added to the commit and potentially could end up in the repo, adding bloat, making reviews harder or in the worst case might result in build errors for others.

So tl;dr:
The correct usage of a .gitignore is a very useful feature that git uses and it is definitly something that everyone could benefit from

averinevg: > where every single developer would have to exclude every single file/folder from their commits manually every time

I agree that this file must help developers. But this does not mean that you need to add paths to folders with private photos, your diploma project files, and other files not related to the project. Each developer should manually add files to a commit, and not automatically add junk and then try to clean it up.

Joly0: > > where every single developer would have to exclude every single file/folder from their commits manually every time
> 
> I agree that this file must help developers. But this does not mean that you need to add paths to folders with private photos, your diploma project files, and other files not related to the project. Each developer should manually add files to a commit, and not automatically add junk and then try to clean it up.

But why would i have to add paths to such things? These should not and probably will never end up in a project folder. But the folders that the different IDEs create do end up in the project folder. Its not like one moves them there manually and by will, the IDE just creates them there. Thats what the .gitignore is for. To get around this and exclude such folders, so the project is not bloated and "dependant" on a single IDE.

You are missing the point i guess. You dont write random paths in the .gitignore. You add paths there, that definitly could or will end up in the project folder, but are not part of the project. 

In Jetbrains IDEs cases, these folders ".idea/", "cmake-build*/", "build*/" will definitly end up in your project folder, if you use that IDE. Not matter if you want it or not (afaik you can configure the different IDEs not to do so, but thats a competely other thing and way out of this scope). So its not like you should add them to the .gitignore, you basically have to if you dont want to check for those files/folders everytime you make any changes/commits in your project. These files and folders will 100% guranteed end up in your project folder and you 100% dont want to have them in your public git repo

Joly0: > > where every single developer would have to exclude every single file/folder from their commits manually every time
> 
> I agree that this file must help developers. But this does not mean that you need to add paths to folders with private photos, your diploma project files, and other files not related to the project. Each developer should manually add files to a commit, and not automatically add junk and then try to clean it up.

I really think you are missing the point of that file, so i´ll try to explain it differently:
Lets imagine a project on github, a very simple hello world project. So you had the following (broken down) folder Structure:
- Hello_World 
   - helloworld.cpp

So you have the Hello_World project folder and a simple helloworld.cpp in that folder, nothing more. Everything fine. Now you open up that project in Visual Studion and now your project folder looks like this:
- Hello_World 
   - helloworld.cpp
   - .vs

Randomly a .vs folder appeared. Now you open the same project in a Jetbrains IDE and the folder structure looks like this:
- Hello_World 
   - helloworld.cpp
   - .vs
   - .idea

And there we got another folder, that has nothing to do with the project, but ended up in your project folder. You now make some changes to the helloworld.cpp file and want to commit, but git finds the .vs and the .idea folder and wants to add that to your commit, you would have to exclude that from your commit. You make further changes and further commits. You will propably at one point get really annyoed by the fact, that you have to exclude those 2 folders manually every single time you make a commit. Now the .gitignore comes in to play. You add those folders to your .gitignore and will never ever have to think about those folders in your git commits. Git will just skip them.

Now think about a bigger project with more than just one developer and they all use different IDEs. You dont want to rely on every developer to be able to always be 100% certain they have excluded those files and folders before they commit. Mistakes can happen and those files/folders will sooner or later end up in your git project.

To prevent this from ever happening, you exclude those files and folders, that you know will be created by the different IDEs in your .gitignore. Now no developer have to check first and be 100% certain, that they have correctly excluded every single file and folder that a commonly used IDE might have added to they project folder.

scarsty: Please pay more attention to review codes and cmakefiles. I can agree with @Joly0, but I never think to argue about ignore file is useful.

scarsty:
In the cmake file

```cmake
    if(WIN32)
        set(MIOPEN_USER_DB_PATH "$HOME\\\\.miopen\\\\db\\\\" CACHE STRING "Default path to user db files")
        set(MIOPEN_CACHE_DIR "$HOME\\\\.miopen\\\\cache\\\\" CACHE STRING "")
	set(MIOPEN_SYSTEM_DB_PATH "C:\\\\ProgramData\\\\MIOpen\\\\" CACHE PATH "Default path to system db files")
    else()
        set(MIOPEN_USER_DB_PATH "~/.config/miopen/" CACHE STRING "Default path of user db files")
        set(MIOPEN_CACHE_DIR "~/.cache/miopen/" CACHE STRING "")
    endif()
```
here $HOME is undefined on Windows usually, so the PATH becomes like `\\.miopen\\db`, which is treated as a net path on Windows. Then the result will be a temp path (src/db_path.cpp.in, about line 98). I think this should be repaired.

.gitignore
Joly0: Totally agree with @scarsty. There shouldnt need to be any argue about the ignore file. It is useful, it is a nice addition by @apwojcik and its smart he thought about adding it in this pr.
Please focus on reviewing the real code and bring this pr and project forward. Many are waiting for a final windows release to use their ROCm cards on Windows

averinevg: @Joly0 I think you should read @scarsty's comment one more time. I would also like to draw your attention to the fact that this particular pull request is a draft, it was an initial draft. It was decided to split all the work into several subsequent pull requests. If you have constructive suggestions for the entire work, you can present them by creating a new issue or pull request.

averinevg:
> In the cmake file

@scarsty Please take a look at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2365

averinevg:
@apwojcik Since all the work has been continued in subsequent PRs, please close this PR if you no longer need it.

apwojcik:
Closing this PR. The work continues in a separate series of PRs.

apwojcik:
Please do **not delete** the `uif2-initial` branch!

PR #2277, opened 2023-07-26
Dropout: make seed and states_num kernel arguments

tbennun:
Every time a dropout descriptor is created (even if it's with a different seed), the initialization kernel is recompiled. This PR turns the number of states to initialize and random seed into runtime parameters. There seems to be no effect on the runtime of the initialization kernel (which I'm assuming is not performance critical), but fewer JIT compilation calls.

Note: I tried setting `prng_seed` to be an `unsigned long long` parameter, but that passed a wrong value onto the kernel itself.

### DIFF
--- a/src/ocl/dropoutocl.cpp
+++ b/src/ocl/dropoutocl.cpp
@@ -132,8 +132,8 @@ void DropoutDescriptor::InitPRNGState(Handle& handle,
         MIOPEN_THROW("PRNG state size should not exceed system maximum memory allocation size.");
     }
 
-    size_t states_num = prng_stateSizeInBytes / sizeof(prngStates);
-    size_t wk_grp_num = std::min(size_t(MAX_PRNG_STATE / 256), (states_num + 255) / 256);
+    unsigned long long states_num = prng_stateSizeInBytes / sizeof(prngStates);
+    size_t wk_grp_num             = std::min(MAX_PRNG_STATE / 256ULL, (states_num + 255) / 256);
 
     std::string network_config = "initprngs-" + std::to_string(sizeof(prngStates)) + "x" +
                                  std::to_string(rng_mode) + "x" + std::to_string(wk_grp_num);
@@ -141,9 +141,7 @@ void DropoutDescriptor::InitPRNGState(Handle& handle,
     auto&& kernels = handle.GetKernels(kernel_name, network_config);
     if(!kernels.empty())
     {
-        kernels.front()(prng_states,
-                        static_cast<unsigned int>(prng_seed),
-                        static_cast<unsigned int>(states_num));
+        kernels.front()(prng_states, prng_seed, states_num);
     }
     else
     {
@@ -157,9 +155,7 @@ void DropoutDescriptor::InitPRNGState(Handle& handle,
         std::cout << "Memory allocated for PRNG states: " << stateSizeInBytes << std::endl;
 #endif
         handle.AddKernel(kernel_name, network_config, program_name, kernel_name, vld, vgd, params)(
-            prng_states,
-            static_cast<unsigned int>(prng_seed),
-            static_cast<unsigned int>(states_num));
+            prng_states, prng_seed, states_num);
 #if DROPOUT_DEBUG
         std::cout << "Succeeded in launching InitPRNGState()." << stateSizeInBytes << std::endl;
 #endif
--- a/src/ocl/dropoutocl.cpp
+++ b/src/ocl/dropoutocl.cpp
@@ -135,15 +135,14 @@ void DropoutDescriptor::InitPRNGState(Handle& handle,
     size_t states_num = prng_stateSizeInBytes / sizeof(prngStates);
     size_t wk_grp_num = std::min(size_t(MAX_PRNG_STATE / 256), (states_num + 255) / 256);
 
-    std::string network_config = "initprngs-" +
-                                 std::to_string(sizeof(prngStates)) + "x" +
-                                 std::to_string(rng_mode) + "x" +
-                                 std::to_string(wk_grp_num);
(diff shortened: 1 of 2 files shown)


junliume:
@atamazov could you help to take a look?

JehandadKhan (dismissed):

tbennun:
@JehandadKhan sorry, called `clang-format` on the code now.

JehandadKhan:
@tbennun  The CL file needs to formatted as well.

tbennun:
@JehandadKhan Done now. I thought I had formatted both.

src/ocl/dropoutocl.cpp
```diff
@@ -135,15 +135,15 @@ void DropoutDescriptor::InitPRNGState(Handle& handle,
     size_t states_num = prng_stateSizeInBytes / sizeof(prngStates);
     size_t wk_grp_num = std::min(size_t(MAX_PRNG_STATE / 256), (states_num + 255) / 256);
 
-    std::string network_config = "initprngs-" + std::to_string(states_num) + "x" +
-                                 std::to_string(sizeof(prngStates)) + "x" +
-                                 std::to_string(rng_mode) + "x" + std::to_string(prng_seed) + "x" +
-                                 std::to_string(wk_grp_num);
+    std::string network_config = "initprngs-" + std::to_string(sizeof(prngStates)) + "x" +
```
DrizztDoUrden, line 138: Please use `sdt::ostringstream` instead of concatenations if there are more than two strings. It's both faster and easier to read.

src/ocl/dropoutocl.cpp
```diff
@@ -135,15 +135,15 @@ void DropoutDescriptor::InitPRNGState(Handle& handle,
...
+                                 std::to_string(rng_mode) + "x" + std::to_string(wk_grp_num);
 
     auto&& kernels = handle.GetKernels(kernel_name, network_config);
     if(!kernels.empty())
     {
-        kernels.front()(prng_states);
+        kernels.front()(prng_states,
```
DrizztDoUrden: Maybe `prng_seed` and `states_num` should be declared as `unsigned`? That would make calls single lines and remove the casts.

DrizztDoUrden (changes requested):

junliume:
@tbennun could you resolve the review opinions above? We will try to get this merged.

src/ocl/dropoutocl.cpp
tbennun: Thank you. I would do this, but the rest of the file (as well as other files such as `softmaxocl.cpp`) use this style of code. Would it be ok to not match the style of the rest of the code?

src/ocl/dropoutocl.cpp
tbennun: Fixed (through a different way). Changing prng_seed to be `unsigned` would change the MIOpen C API (in which `miopenSetDropoutDescriptor` uses `unsigned long long`). I'd rather pass in an unsigned long long value, which I now do.

tbennun:
@junliume I now addressed the two comments, see above and new commit.

src/ocl/dropoutocl.cpp
DrizztDoUrden: Yes, otherwise it would never really change.

src/ocl/dropoutocl.cpp
DrizztDoUrden: I have missed that. That's good now.

DrizztDoUrden (approved):

junliume (approved):

junliume:
@atamazov could you comment on the user's question?
> The private_seg_size of 3204 seems to be caused by the precomputed XORWOW matrices, which are placed in OpenCL constant memory and, upon compilation, may be stored in every thread's memory. It might be useful to store one or both matrices in global memory to avoid the private_seg_size pollution.

CC: @zjing14

atamazov:
@junliume I need some clue to the code as I am not confident with dropout. What are the symbolic names of the "the precomputed XORWOW matrices"?

/cc @zjing14

tbennun:
> @junliume I need some clue to the code as I am not confident with dropout. What are the symbolic names of the "the precomputed XORWOW matrices"?
> 
> /cc @zjing14

@atamazov This refers to the two included files in the Dropout kernel, which initialize large arrays in constant memory:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/kernels/precalc_xorwow_skipahead_sequence_matrices_kernel.h
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/kernels/precalc_xorwow_skipahead_matrices_kernel.h

atamazov:
@tbennun @junliume AFAICS the parts of these matrices are copied to the private memory for a purpose, i.e. each workitem _modifies_ its local copy. If we want to put those into global memory, then we need to allocate some workspace (enough to hold data from matrices) and then manually manage it, which doesn't seem easy to me.

But why we may need to save space in private segment? Are we hitting the limit?

/cc @zjing14

junliume:
@atamazov yes we are hitting a limit on constant memory, with this fix it "gives a reduction in initialization time of a Transformer model from 10:22 minutes to 31 seconds." We should take a look at the OCL code to see if we can work on heap memory instead.

CAHEK7:
PRNG state requires some workspace in any case, but it supposed to be tiny, but it should be double-checked.
The tables, as far as I know, are required to speedup PRNG initialization and warmup (it may require huge numbers of PRNG invocation) and if that's the case - we can hold the tables on CPU side, choose particular entry for the seed and offset, upload it to the state and save a lot of constant\workspace memory.

But to be sure we need to re-investigate our GPU PRNG implementation.

atamazov:
@CAHEK7 Thanks for feedback. Now we at least better understand what to begin with.

/cc @JehandadKhan @junliume

PR #2278, opened 2023-07-27
Properly failing FindSoluions on exceptions

DrizztDoUrden:
This does two things:
- Exhaustive search adds default perf config to the search space when the virtual container is empty. This is done to report the solver as failed when exiting the search. Another solver would be attempted by find rather than going on with an untested perf config.
- When an error is catched from search-getsolution pair we return an empty `ConvSolution` with `miopenStatusInternalError` rather than `GetSolution(..., solver.GetDefaultPerfConfig(...))`. This way rest of find knows that this solver is bugged and we cannot use it for this netconfig.

This way when something like this happens again, user would be informed, but the result of operation should still be correct unless it's a GPU segfault in the solver.

To reiterate: completely empty generic search container indicates one of two things:
- a recoverable bug in the solver as it shows itself as applicable, but cannot produce a valid perfconfig. This is not fine and we should fix it, but should be manageable unless segaults.
- an incorrect usage of API. We check what we can check, but if the API is internal (including calls from inside library) it is hard to have both performance and lack of unchecked parameters.

### DIFF
--- a/src/include/miopen/find_solution.hpp
+++ b/src/include/miopen/find_solution.hpp
@@ -126,6 +126,7 @@ auto FindSolutionImpl(rank<1>,
             catch(const miopen::Exception& ex)
             {
                 MIOPEN_LOG_E("Search failed for: " << s.SolverDbId() << ": " << ex.what());
+                return ConvSolution(miopenStatusInternalError);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


DrizztDoUrden:
@atamazov you may be interested in this

DrizztDoUrden:
I don't understand failure logs in fin build test. Does it fail because of "error: 0" lines?

shurale-nkn:
> I don't understand failure logs in fin build test. Does it fail because of "error: 0" lines?

no, more likely on the name of the branch in the logs

DrizztDoUrden:
Oh, that was epic fail, lol.

Issue #2270, opened 2023-07-21
Issue with #2266 and fix within mainline


junliume:
**[Issue]**
**I explicitly asked for and am responsible for the implementation of #2266**.

After some discussion we decided to find the root cause instead of hiding it.

Let's see how to fix it properly. Thanks!

**[Content]** 
more on   https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2266

> I think we all agree that the problem is caused by a broken solver

No, after talking to @shurale-nkn I think that it is caused by incorrect usage of our internal APIs from outside of MIOpen.

> this change should stay as an ugly but reasonable "hotfix"

This is not a fix for two reasons.
1. There is no error in GenericSearch. It correctly reports that it's been provided incorrect data for whatever reason. As a method of error reporting, a log line is printed and an exception is thrown. The exception is handled by find and the solver is skipped as invalid for this case.
2. This doesn't fix error. It propagates error further to `GetSolution`. Is returning a default-constructed performance config when solver says no config is valid considered a valid practice? I don't think so. The only way to know that such practice is acceptable is to somehow fetch info from the solver itself. As of now it just assumes that this way of handling is correct (which it is likely not).

@junliume My advice is to revert the PR and to find what causes the error. I think that it is incorrect usage by MIGraphX, but I am not sure. Maybe, there is some undocumented UB in our internals (which is not good), but that's kind of why they are internals and not public API. I am 100% sure that the only change that may be made to GenericSearch from the report is to add
```C++
if (n_runs_total == 0)
    MIOPEN_THROW(...);
```
at the begfining of the function. But the only effect it would have is skipping 0 config space size warnings before the error.

_Originally posted by @DrizztDoUrden in https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2266#issuecomment-1643903706_

junliume:
@DrizztDoUrden @atamazov @shurale-nkn @dmikushin 
Indeed it seems there are consequences if we tolerate `n_runs_total == 0` without throwing errors:
Now the run fails with:
```
[2023-07-23T11:45:17.673Z] MIOpen(HIP): Error [Compile] 'hiprtcCompileProgram(prog.get(), c_options.size(), c_options.data())' static_kernel_gridwise_convolution_forward_implicit_gemm_v4r4_xdlops_nchw_kcyx_nkhw.cpp: HIPRTC_ERROR_COMPILATION (6)
[...]

PR #2279, opened 2023-07-27
Properly failing FindSoluions on exceptions

DrizztDoUrden:
Recreated #2278
Removed the word failed from the branch name so file paths would not fail the tests.

This does two things:

Exhaustive search adds default perf config to the search space when the virtual container is empty. This is done to report the solver as failed when exiting the search. Another solver would be attempted by find rather than going on with an untested perf config.
When an error is catched from search-getsolution pair we return an empty ConvSolution with miopenStatusInternalError rather than GetSolution(..., solver.GetDefaultPerfConfig(...)). This way rest of find knows that this solver is bugged and we cannot use it for this netconfig.
This way when something like this happens again, user would be informed, but the result of operation should still be correct unless it's a GPU segfault in the solver.

To reiterate: completely empty generic search container indicates one of two things:

a recoverable bug in the solver as it shows itself as applicable, but cannot produce a valid perfconfig. This is not fine and we should fix it, but should be manageable unless segaults.
an incorrect usage of API. We check what we can check, but if the API is internal (including calls from inside library) it is hard to have both performance and lack of unchecked parameters.

### DIFF
--- a/src/include/miopen/generic_search.hpp
+++ b/src/include/miopen/generic_search.hpp
@@ -373,7 +373,7 @@ auto GenericSearch(const Solver s,
     auto& profile_h = context.GetStream();
     const AutoEnableProfiling enableProfiling{profile_h};
 
-    auto tmp_all_configs = GetAllConfigs(s, context, problem);
+    auto tmp_all_configs = GetAllConfigs(s, context, problem); 
     // For random access
     std::vector<PerformanceConfig> all_configs;
     std::copy(tmp_all_configs.begin(), tmp_all_configs.end(), std::back_inserter(all_configs));
--- a/src/include/miopen/generic_search.hpp
+++ b/src/include/miopen/generic_search.hpp
@@ -384,6 +384,9 @@ auto GenericSearch(const Solver s,
     const std::size_t n_runs_total = std::min(all_configs.size(), GetTuningIterationsMax());
     all_configs.resize(n_runs_total);
 
+    if(all_configs.empty())
+        all_configs.emplace_back(s.GetDefaultPerformanceConfig(context, problem));
+
     bool is_passed  = false; // left false only if all iterations failed.
     float best_time = std::numeric_limits<float>::max();
     size_t n_failed = 0;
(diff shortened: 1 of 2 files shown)


junliume (approved):

atamazov:
This partially resolves #2270 -- now incorrect Solvers would not lead to a hard failure of the library. Fixing the guilty solver, `ConvHipImplicitGemmForwardV4R4Xdlops`, is another work.

src/include/miopen/generic_search.hpp
```diff
@@ -381,9 +381,26 @@ auto GenericSearch(const Solver s,
...
+    if(all_configs.empty())
+    {
+        const auto default_config = s.GetDefaultPerformanceConfig(context, problem);
+
+        if(default_config.IsValid(context, problem))
+        {
+            all_configs.emplace_back(default_config);
+            n_runs_total += 1;
+        }
+        else
```
atamazov, line 396: @DrizztDoUrden @shurale-nkn @JehandadKhan This change is not necessary to resolve #2270 and, unfortunately, seems a bit harmful to me. Why:

If the Solver produces none performanceConfigs for tuning (container is empty), then something is definitely wrong and needs to be fixed (e.g. applicability of the solver needs to be corrected or so). If we manually add one default config, then we just **_camouflage the problem_** and waste resources. Because as a result of tuning, the library will write the default performance config into the perf-db. 

If we simply fail tuning and keep perf-db intact, the solver will use the default config later anyway. This way the library will be protected against the "empty container" problem while keeping the database free of junk.

I am going to fix this soon.

### Some details

The container is expected to hold all the potentially usable performanceConfigs, including the default one. Therefore the default performance config is expected to reside in the container (either primary of spare). If this is not so, then the Solver must be fixed. In other words, the current contract does not allow situations when container is empty while the default performance config is valid.

If we would like to change the contract and allow Solvers to produce empty containers (but allow the default performance config to be valid -- which I do not like), then we should seriously consider splitting the Solver to two: оne is tunable, another one is not. That is necessary to optimize out unnecessary database accesses.

atamazov: See "Restrictions" at https://github.com/ROCmSoftwarePlatform/MIOpen/issues/866#issuecomment-822585406

atamazov: [won't fix] see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2296#issuecomment-1680731004

PR #2283, opened 2023-07-28
fix ASM iGEMM applicability with datalayout

junliume:
https://ontrack-internal.amd.com/browse/SWDEV-404404
https://ontrack-internal.amd.com/browse/SWDEV-404537

Somhow in Find 2.0 ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC is picked when exaustive search is enforced, even though the datalayout is not NHWC


carlushuang:
asm igemm solver do support NCHW like in https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ef907d6cb406f9859a7bcfbc0858c924f7225be5/src/solver/conv_asm_implicit_gemm_gtc_bwd_nhwc.cpp#L1002-L1010

It will call transpose kernel then launch the original NHWC kernel.

Is any specific MIOpenDriver cmd for this crash?

junliume:
Put this PR on hold.

PR #2287, opened 2023-08-01
Dont add librt to the MIOpen's interface

pfultz2:
librt is not used in miopen's public header so it should only be linked privately and for the tests. 

This causes an issue when using miopen on ubuntu 22.04, because it is named librt.so.1 instead of librt.so. We really shouldn't ever use absolute paths in the target interface since libraries can be located in other locations.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -836,7 +836,7 @@ if(NOT WIN32 AND NOT APPLE)
     find_library(LIBRT rt)
     if(LIBRT)
         MESSAGE(STATUS "Librt: " ${LIBRT})
-        target_link_libraries(MIOpen PUBLIC ${LIBRT})
+        target_internal_library(MIOpen ${LIBRT})
     endif()
 endif()
 


junliume:
@pfultz2 it's outdated but the PUBLIC link is explicitly changed in https://github.com/ROCmSoftwarePlatform/MIOpen/commit/ec3fdbe6b907f2b6599fc8de366d41a67eeaaf09 to resolve a problem. Is there a way to track what was the linking problem?

pfultz2:
> it's outdated but the PUBLIC link is explicitly changed in https://github.com/ROCmSoftwarePlatform/MIOpen/commit/ec3fdbe6b907f2b6599fc8de366d41a67eeaaf09 to resolve a problem.

My assumption is that it was made PUBLIC for the tests, which use miopen's internal headers. Thats why I used `target_internal_library` so it will still link for the tests.

However, miopen's public header doesn't include Boost.Interprocess(nor any headers from librt), so users of miopen shouldn't need to link with librt explicitly. The runtime loader will automatically load librt with miopen automatically. 

For the static library, cmake will propagate the dependency, which means it would still use the absolute path. Maybe it would be better to use `-lrt` instead and let the linker find the library.

Either way, this at least fixes the issue for shared library case, and we could try to fix static library case in the future.

junliume (approved):
Ping @atamazov and @JehandadKhan  to reach consensus.

atamazov:
@junliume @pfultz2 @JehandadKhan I totally agree with Paul.

atamazov (approved):
LGTM

PR #2288, opened 2023-08-01
fix missing DetectRocm() in FindSolutionsImpl(), where options = nullptr

carlushuang:
This will fix [SWDEV-404537](https://ontrack-internal.amd.com/browse/SWDEV-404537)

We people use `miopenFindSolutoins()` API with `options = nullptr`, will have GPU crash if `MIOPEN_FIND_ENFORCE=4`

The root cause is there is no workspace allocated while tuning, since when call `GetWorkspace()` every asm solver will return not applicable, but when launching the kernel they will return applicable, and run the kernel with nullptr workspace.
This PR fix this issue.

image

### DIFF
--- a/src/problem.cpp
+++ b/src/problem.cpp
@@ -258,7 +258,9 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
     }
     else
     {
-        const auto workspace_max = conv_desc.GetWorkSpaceSize({&handle}, conv_problem);
+        auto tmp_ctx = ExecutionContext{&handle};
+        tmp_ctx.DetectRocm();
+        const auto workspace_max = conv_desc.GetWorkSpaceSize(tmp_ctx, conv_problem);
         workspace_size           = std::min(options.workspace_limit, workspace_max);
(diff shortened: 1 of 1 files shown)


JehandadKhan (approved):

junliume (approved):

atamazov (approved):
Good as a hot fix!

@carlushuang Thanks for this finding. Can you propose some regression test (or just describe what needs to be done and in what order)?

junliume:
> Good as a hot fix!
> 
> @carlushuang Thanks for this finding. Can you propose some regression test (or just describe what needs to be done and in what order)?

@atamazov  there is no good way of reproducing it in MIOpenDriver, actually it is not found when running migrtaphx-driver. Maybe there is a way to reproduce in Find 2.0 path @DrizztDoUrden ?

carlushuang:
Agreed with Jun. `MIOpenDriver` has different code path to utilize internal APIs, in this case `miopenFindSolutoins(options=nullptr)` can only fire the bug.
@atamazov I guess just test the API level with options = nullptr and not nullptr?

PR #1803, opened 2022-10-04
[CK API] 2d forward convolution composable kernel integration (multi-datatype)

iq136boy:
Integrated 2d forward convolution composable kernels into MIOpen.

PR #1692, opened 2022-08-11
Composable kernels integration & Small Fix in GEMM Solver

JehandadKhan:

PR #2292, opened 2023-08-01
[tests] test_conv_hip_igemm_xdlops to gTest

xinlipn:
Convert ctest test_conv_hip_igemm_xdlops to gTest


junliume:
@xinlipn please resolve conflicts and this PR has passed CI once.

xinlipn:
> @xinlipn please resolve conflicts and this PR has passed CI once.

@junliume , Fixed conflicts

test/gtest/conv_hip_igemm_xdlops.cpp
```diff
@@ -0,0 +1,168 @@
...
+    case miopenHalf:
+    case miopenBFloat16:
+    case miopenFloat:
+    case miopenInt8x4:
+    case miopenInt32:
+    case miopenDouble:
+        FAIL() << "miopenHalf, miopenBFloat16, miopenFloat, miopenInt8x4, miopenInt32, "
+                  "miopenDouble data "
+                  "type not supported by "
+                  "conv_hip_igemm_mlir_xdlops test";
```
atamazov: [Recommendation] At the time of adding the ConvHipImplicitGemmFwdXdlops solver (in #1692), only the INT8 type was supported, and the initial test set was of INT8 type only. Then there was #1803 which adds support for HALF and FLOAT, but that PR is missing the support of these new datatypes in tests. That is why the test lacks HALF and FLOAT support.

There is another problem with this test: it allows all solvers to work and therefore there there is no guarantee that namely the `ConvHipImplicitGemmFwdXdlops` is being tested. If there is a problem with it (i.e. it is disabled), then any other solver (i.e. Gemm) can do the job and test would succeed.

@xinlipn @JehandadKhan What about resolving there problems at once? Or let's address these in a separate PR?

test/gtest/CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ function(add_gtest TEST_NAME)
       target_link_libraries(test_${TEST_NAME} gtest_main MIOpen ${Boost_LIBRARIES} hip::host $<BUILD_INTERFACE:roc::rocblas>)
     endif()
     # Enable CMake to discover the test binary
-    gtest_discover_tests(test_${TEST_NAME} PROPERTIES ENVIRONMENT "MIOPEN_USER_DB_PATH=${CMAKE_CURRENT_BINARY_DIR};MIOPEN_TEST_ALL=${MIOPEN_TEST_ALL};MIOPEN_TEST_MLIR=${MIOPEN_TEST_MLIR}")
+    gtest_discover_tests(test_${TEST_NAME} PROPERTIES ENVIRONMENT "MIOPEN_USER_DB_PATH=${CMAKE_CURRENT_BINARY_DIR};MIOPEN_TEST_ALL=${MIOPEN_TEST_ALL};MIOPEN_TEST_COMPOSABLEKERNEL=${MIOPEN_TEST_COMPOSABLEKERNEL}")
```
atamazov: Please do not remove the MLIR thing

test/gtest/conv_hip_igemm_xdlops.cpp
atamazov:
```suggestion
                  "test_conv_hip_igemm_xdlops test";
```

atamazov (changes requested):
Almost good.

test/gtest/conv_hip_igemm_xdlops.cpp
xinlipn: > [Recommendation] At the time of adding the ConvHipImplicitGemmFwdXdlops solver (in #1692), only the INT8 type was supported, and the initial test set was of INT8 type only. Then there was #1803 which adds support for HALF and FLOAT, but that PR is missing the support of these new datatypes in tests. That is why the test lacks HALF and FLOAT support.
> 
> There is another problem with this test: it allows all solvers to work and therefore there there is no guarantee that namely the `ConvHipImplicitGemmFwdXdlops` is being tested. If there is a problem with it (i.e. it is disabled), then any other solver (i.e. Gemm) can do the job and test would succeed.
> 
> @xinlipn @JehandadKhan What about resolving there problems at once? Or let's address these in a separate PR?

@atamazov , thanks for bringing up these points. Does it make sense to open another PR? The reason is 1) Please correct me if I was wrong, it seems ConvHipImplicitGemmFwdXdlops is a different test, we can update with the datatypes you mentioned in its own gTest 2) in case it's the same test as here, this PR can be used as 1:1 map from the original cTest and any other changes can be put in a separate PR to avoid confusion.

xinlipn: @atamazov Done

test/gtest/CMakeLists.txt
xinlipn: @atamazov Done

atamazov: [Resolved]

test/gtest/conv_hip_igemm_xdlops.cpp
atamazov: [Resolved]

atamazov: @xinlipn 
> 1) Please correct me if I was wrong, it seems ConvHipImplicitGemmFwdXdlops is a different test

This `test_conv_hip_igemm_xdlops` (which is being refactored in this PR) was introduced together with the `ConvHipImplicitGemmFwdXdlops` solver, in #1692 (which was created by @JehandadKhan). Therefore I suspect that this test is intended to test specifically `ConvHipImplicitGemmFwdXdlops`. If this is so, then I recommend adding support for other datatypes, and limiting it to the `ConvHipImplicitGemmFwdXdlops` only.

Please discuss with @JehandadKhan in order to clarify things and to decide what needs to be done. My comment is only recommendation.

> 2) in case it's the same test as here, this PR can be used as 1:1 map from the original cTest and any other changes can be put in a separate PR to avoid confusion.

No objections!

atamazov (approved):
LGTM. I am assuming that https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2292#discussion_r1286416434 will be followed up here or in the separate PR.

JehandadKhan:
> LGTM. I am assuming that [#2292 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2292#discussion_r1286416434) will be followed up here or in the separate PR.

I think that makes sense, @xinlipn  can address the comment in a follow up PR.

xinlipn:
Tested on gfx908, fpadmin@rocm-framework-11.amd.com 

$printenv | grep MIOPEN
MIOPEN_ENABLE_LOGGING_CMD=1
MIOPEN_TEST_ALL=1
MIOPEN_TEST_COMPOSABLEKERNEL=TRUE
MIOPEN_TEST_FLOAT_ARG=--int8
MIOPEN_ENABLE_LOGGING=1
MIOPEN_LOG_LEVEL=5

gtest passed, log is as below:

[gtest_test_conv_hip_igemm_xdlops.log](https://github.com/ROCmSoftwarePlatform/MIOpen/files/12471880/gtest_test_conv_hip_igemm_xdlops.log)

junliume (approved):

PR #2299, opened 2023-08-04
fix test after removal of mlir from gfx908 kdb

bghimireamd:
* Currently we have set `MIOPEN_FIND_ENFORCE=3` in our test. Setting value `3` means that tuning is not performed if the perf-db already contains the performance parameters. For reliability of the test, use MIOPEN_FIND_ENFORCE=SEARCH_DB_UPDATE.

* The full tuning cycle may take thousands of iterations. We use `MIOPEN_DEBUG_TUNING_ITERATIONS_MAX=5` to limit tuning time in tests.
* This PR is also to fix the current failing develop branch. 

Reference : https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2154#discussion_r1252350092

### DIFF
--- a/test/gtest/cba_infer.cpp
+++ b/test/gtest/cba_infer.cpp
@@ -151,7 +151,8 @@ TEST_P(ConvBiasActivInferTestHalf, ConvCKIgemmFwdBiasActivFused)
 #if MIOPEN_BACKEND_HIP
 TEST_P(ConvBiasActivInferTestFloatFusionCompileStep, ConvBiasActivAsm1x1UFloat_testCompile)
 {
-    setEnvironmentVariable("MIOPEN_FIND_ENFORCE", "3");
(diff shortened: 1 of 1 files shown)


atamazov (approved):
LGTM!

junliume (approved):

junliume:
@bghimireamd : one question: #2293 passed CI check as shown in  http://micimaster.amd.com/blue/organizations/jenkins/MLLibs%2FMIOpen/detail/cderb%2Fgold18_patch/3/pipeline 
> This PR is also to fix the current failing develop branch.

So what is the problem?

atamazov:
@junliume Currently, the test does not guarantee that it actually tests anything. https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2154#discussion_r1252350092 explains why.

PR #2303, opened 2023-08-07
[NFC] Replace miopen::ProblemDescription with conv::ProblemDescription, part 3

averinevg:
* Make miopen::ProblemDescription derived from conv::ProblemDescription
* miopen::ProblemDescription: remove getters that exist in the parent class
* conv::ProblemDescription: use proper data type of return value for getters that use data from TensorDescriptor
* conv::ProblemDescription: add trailing underscores to getters with changed data type in order to get the attention of developers who are not familiar with these changes (will be removed after ROCm 6.0)
* Remove unused code


averinevg:
@atamazov Please review

atamazov:
@averinevg 
> conv::ProblemDescription: add trailing underscores to getters with changed data type in order to get the attention of developers who are not familiar with these changes (will be removed after ROCm 6.0)

Are you going to remove those underscores after MIOpen 6.0 RC is cut off?

src/include/miopen/tensor.hpp
```diff
@@ -285,7 +285,7 @@ struct TensorDescriptor : miopenTensorDescriptor
 };
 
 template <class TElement>
-constexpr auto GetNCDHW(int spatial_dims, const std::vector<TElement>& data)
+constexpr auto GetNCDHW(unsigned spatial_dims, const std::vector<TElement>& data)
```
atamazov, line 288: [Q] why this is necessary?

src/include/miopen/conv/problem_description.hpp
```diff
@@ -190,17 +190,17 @@ struct ProblemDescription : ProblemDescriptionBase
...
+    unsigned GetBatchSize_() const { return GetInBatchSize_(); } // alias of GetInBatchSize_()
+    unsigned GetInChannels_() const { return GetC5(GetSpatialDims(), in.GetLengths()); }
+    unsigned GetInDepth_() const { return GetD5(GetSpatialDims(), in.GetLengths()); }
+    unsigned GetInHeight_() const { return GetH5(GetSpatialDims(), in.GetLengths()); }
+    unsigned GetInWidth_() const { return GetW5(GetSpatialDims(), in.GetLengths()); }
+    unsigned GetInBatchStride_() const { return GetN5(GetSpatialDims(), in.GetStrides()); }
+    unsigned GetInChannelStride_() const { return GetC5(GetSpatialDims(), in.GetStrides()); }
+    unsigned GetInStrideD_() const { return GetD5(GetSpatialDims(), in.GetStrides()); }
+    unsigned GetInStrideH_() const { return GetH5(GetSpatialDims(), in.GetStrides()); }
+    unsigned GetInStrideW_() const { return GetW5(GetSpatialDims(), in.GetStrides()); }
```
atamazov, line 203: [Q] I am not sure this is a good change. Let's discuss.

atamazov (commented):
Something to discuss.

averinevg:
> @averinevg
> 
> > conv::ProblemDescription: add trailing underscores to getters with changed data type in order to get the attention of developers who are not familiar with these changes (will be removed after ROCm 6.0)
> 
> Are you going to remove those underscores after MIOpen 6.0 RC is cut off?

Yes, this is a temporary change

src/include/miopen/tensor.hpp
averinevg: This variable is always positive and using a signed data type can be misleading

atamazov: That is, you just caught a not quite high-quality code and decided to fix it, right? 😄

atamazov:
WRT https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2303#discussion_r1286316295. After ~2 hours discussion we decided to continue as is and then switch to size_t in a dedicated PR.

src/solver/conv_MP_bidirectional_winograd.cpp
```diff
@@ -89,17 +89,17 @@ MIOPEN_DECLARE_ENV_VAR(MIOPEN_CONV_PRECISE_ROCBLAS_TIMING)
...
-    const auto N         = (problem).GetBatchSize();               \
-    const int K          = (problem).GetOutChannels() / group_cnt; \
-    const int C          = (problem).GetInChannels() / group_cnt;  \
-    const auto R         = (problem).GetWeightsHeight();           \
-    const auto S         = (problem).GetWeightsWidth();            \
-    const auto H         = (problem).GetInHeight();                \
-    const auto W         = (problem).GetInWidth();                 \
-    const auto out_H     = (problem).GetOutHeight();               \
-    const auto out_W     = (problem).GetOutWidth();
+#define DEFINE_SHADER_ALIASES(problem)                              \
```
DrizztDoUrden, line 92: An idea for future work to ease solver developers life:
Similar code is verbose and present in most of our solvers. We should add some method like `CommonlyUsedConvParameters GetCommonlyUsedConvParameters` with `CommonlyUsedConvParameters` being some type which is compatible to structured bindings so it would be possible to either extract them as `const auto [grop_cnt, N, K, ...] = problem.GetCommonlyUsedConvParameters()` or just as `const auto params = GetCommonlyUsedConvParameters()` to use as `params.group_cnt` and such.

DrizztDoUrden (dismissed):
LGTM

src/include/miopen/tensor.hpp
averinevg: Yes, you are right :smile:. I decided to change this too, because it is related to my changes

src/solver/conv_MP_bidirectional_winograd.cpp
averinevg: Perhaps we should also refactor other "helpers" that get data from `ProblemDescription`. I assume that they partially overlap and can be generalized.

junliume (dismissed):
LGTM but I am waiting for the good opportunity to merge. Too many files changed means higher chances for other PRs to have conflicts.

DrizztDoUrden (dismissed):

JehandadKhan (approved):

JehandadKhan:
@junliume  We can merge this when you get a chance

junliume (approved):

src/include/miopen/conv/problem_description.hpp
amberhassaan: I would also like to know why we went to `unsigned` from `size_t` and why the names have an `_` at the end of getters.

averinevg: @amberhassaan You can find details here https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2303#issue-1839780618 and here https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2303#issuecomment-1672048340

In most places in the library `miopen::ProblemDescription` is used, its getters returned the `int` type, the getters of `conv::ProblemDescription` returned the `size_t` type. For now, it is enough for us to have `int`. `unsigned` is used because we don't need a sign.

src/include/miopen/problem_description.hpp
```diff
@@ -38,6 +38,8 @@
 #include <cstdint>
 #include <string>
 
+#define FIN_OLD_PROBLEM_DESCRIPTION_COMPAT 1
```
amberhassaan, line 41: Is this flag documented somewhere?

averinevg: No, it is supposed to be used with MIFin for backward compatibility and will be removed soon.

Issue #2301, opened 2023-08-07
Post-merge issues of #2125


atamazov:
- [x] 🔴 https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2125#discussion_r1285961294 - the "deterministic" attribute is not observed.
- [x] 🟡 https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2125#discussion_r1286284003 - computations that depend on problem should be moved from Invoker to InvokerFactory.

----
[Attribution] @junliume @JehandadKhan
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/correctness
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_high
- Proposed assignees: @iq136boy

JehandadKhan:
@amberhassaan  Can you please take a look at this issue ? IIUC this should be a two liner fix.

Thanks @atamazov  for catching this.

amberhassaan:
Will put out a fix later today.

atamazov:
One more bullet (yellow) added.

atamazov:
~The topmost comment updated with unresolved comments from #2305 and a new ticket.~

atamazov:
via #2305

PR #2286, opened 2023-07-31
3D group forward convolution solver

iq136boy:
Enable the 3d group forward convolution support in miopen using the composable kernel.

PR #2305, opened 2023-08-07
Post-merge fixups: Replace environment variable check with problem config check and reduce lambda capture for Invoker obj

amberhassaan:
addressing comments from #2301 . 
1. Replace env variable based check with a check based on `conv_problem`. 
2. Move InvokerFactory creation into a separate function with careful attention to what is captured by `Invoker` lambda as the latter is cached for reuse and we want it to be fast and minimal in size.

### DIFF
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -4561,7 +4561,7 @@ struct PerformanceConfigHipImplicitGemmFwdXdlops
     PerformanceConfigHipImplicitGemmFwdXdlops() : PerformanceConfigHipImplicitGemmFwdXdlops(0, "")
     {
     }
-    PerformanceConfigHipImplicitGemmFwdXdlops(bool)
+    explicit PerformanceConfigHipImplicitGemmFwdXdlops(bool)
         : PerformanceConfigHipImplicitGemmFwdXdlops(0, "")
     {
     }
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -238,7 +238,7 @@ MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem,
     InvokerFactory ret =
         [ck_args = std::move(ck_args), data_type, config_idx](const std::vector<Kernel>& kernels) {
             std::ignore = kernels;
-            return [ck_args = std::move(ck_args), data_type, config_idx](
+            return [ck_args, data_type, config_idx](
                        const Handle& handle, const AnyInvokeParams& primitive_parameters) {
                 switch(data_type)
                 {
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -226,38 +226,40 @@ void RunCKSolution(const Handle& handle,
     }
 }
 
-InvokerFactory MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem, 
-    const PerformanceConfigHipImplicitGemmFwdXdlops& config) {
+InvokerFactory
+MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem,
+                                       const PerformanceConfigHipImplicitGemmFwdXdlops& config)
+{
 
-  auto ck_args = CKArgs{problem};
-  miopenDataType_t data_type = problem.conv_problem.GetInDataType();
-  auto config_idx = config.index;
+    auto ck_args               = CKArgs{problem};
+    miopenDataType_t data_type = problem.conv_problem.GetInDataType();
+    auto config_idx            = config.index;
 
-
-  InvokerFactory ret = [ck_args, data_type, config_idx] (const std::vector<Kernel>& kernels) {
-        std::ignore = kernels;
-        return [ck_args, data_type, config_idx] (const Handle& handle, const AnyInvokeParams& primitive_parameters) {
-            switch(data_type)
-            {
-            case miopenInt8:
-                RunCKSolution<int8_t>(handle, primitive_parameters, ck_args, config_idx);
-                break;
-            case miopenHalf:
-                RunCKSolution<ck::half_t>(handle, primitive_parameters, ck_args, config_idx);
-                break;
-            case miopenFloat:
-                RunCKSolution<float>(handle, primitive_parameters, ck_args, config_idx);
-                break;
-            case miopenInt32:
-            case miopenInt8x4:
-            case miopenBFloat16:
-            case miopenDouble: break;
-            }
+    InvokerFactory ret =
+        [ck_args = std::move(ck_args), data_type, config_idx](const std::vector<Kernel>& kernels) {
+            std::ignore = kernels;
+            return [ck_args = std::move(ck_args), data_type, config_idx](
+                       const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+                switch(data_type)
+                {
+                case miopenInt8:
+                    RunCKSolution<int8_t>(handle, primitive_parameters, ck_args, config_idx);
+                    break;
+                case miopenHalf:
+                    RunCKSolution<ck::half_t>(handle, primitive_parameters, ck_args, config_idx);
+                    break;
+                case miopenFloat:
+                    RunCKSolution<float>(handle, primitive_parameters, ck_args, config_idx);
+                    break;
+                case miopenInt32:
+                case miopenInt8x4:
+                case miopenBFloat16:
+                case miopenDouble: break;
+                }
+            };
         };
-    };
 
-  return ret;
-
+    return ret;
 }
 
 } // namespace
@@ -411,10 +413,7 @@ ConvSolution ConvHipImplicitGemmFwdXdlops::GetSolution(
     return {};
 #else
     ConvSolution result;
-    result.invoker_factory = MakeInvokerFactoryHipImplGemmFwdXdlops(
-        problem, 
-        config.index
-        );
+    result.invoker_factory = MakeInvokerFactoryHipImplGemmFwdXdlops(problem, config.index);
 
     return result;
 #endif
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -238,8 +238,8 @@ MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem,
     InvokerFactory ret =
         [ck_args = std::move(ck_args), data_type, config_idx](const std::vector<Kernel>& kernels) {
             std::ignore = kernels;
-            return [ck_args, data_type, config_idx](
-                       const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            return [ck_args, data_type, config_idx](const Handle& handle,
+                                                    const AnyInvokeParams& primitive_parameters) {
                 switch(data_type)
                 {
                 case miopenInt8:
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -236,15 +236,14 @@ void RunCKSolution(const Handle& handle,
 template <typename DataType>
 InvokerFactory MakeInvokerFactoryHelper(CKArgs ck_args, size_t config_idx)
 {
-    return std::move([ck_args = std::move(ck_args),
-                      config_idx](const std::vector<Kernel>& kernels) mutable {
+    return [ck_args = std::move(ck_args), config_idx](const std::vector<Kernel>& kernels) mutable {
         std::ignore = kernels;
 
-        return std::move([ck_args = std::move(ck_args), config_idx](
-                             const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+        return [ck_args = std::move(ck_args),
+                config_idx](const Handle& handle, const AnyInvokeParams& primitive_parameters) {
             RunCKSolution<DataType>(handle, primitive_parameters, ck_args, config_idx);
-        });
-    });
+        };
+    };
 }
 
 InvokerFactory MakeInvokerFactoryThrowError()
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -413,7 +413,7 @@ ConvSolution ConvHipImplicitGemmFwdXdlops::GetSolution(
     return {};
 #else
     ConvSolution result;
-    result.invoker_factory = MakeInvokerFactoryHipImplGemmFwdXdlops(problem, config.index);
+    result.invoker_factory = MakeInvokerFactoryHipImplGemmFwdXdlops(problem, config);
 
     return result;
 #endif
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -246,16 +246,6 @@ InvokerFactory MakeInvokerFactoryHelper(CKArgs ck_args, size_t config_idx)
     };
 }
 
-InvokerFactory MakeInvokerFactoryThrowError()
-{
-    return [](const std::vector<Kernel>& kernels) {
-        std::ignore = kernels;
-        MIOPEN_THROW(miopenStatusNotImplemented,
-                     "Convolution operation not implemented for this data type");
-        return [](const Handle&, const AnyInvokeParams&) {};
-    };
-}
-
 InvokerFactory
 MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem,
                                        const PerformanceConfigHipImplicitGemmFwdXdlops& config)
@@ -272,7 +262,10 @@ MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem,
     case miopenInt32:
     case miopenInt8x4:
     case miopenBFloat16:
-    case miopenDouble: return MakeInvokerFactoryThrowError();
+    case miopenDouble:
+    default:
+        MIOPEN_THROW(miopenStatusInternalError,
+                     "ConvHipImplicitGemmFwdXdlops operation not implemented for this data type");
     }
 }
 
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -80,6 +80,13 @@ struct CKArgs
         rPadding = {ProblemInterpreter::GetAdjustedInputRightPadH(problem),
                     ProblemInterpreter::GetAdjustedInputRightPadW(problem)};
     }
+
+    CKArgs(const CKArgs&) = default;
+    CKArgs(CKArgs&&)      = default;
+    CKArgs& operator=(const CKArgs&) = default;
+    CKArgs& operator=(CKArgs&&) = default;
+    ~CKArgs()                   = default;
+
     int N;
     int K;
     int C;
@@ -226,40 +233,48 @@ void RunCKSolution(const Handle& handle,
     }
 }
 
+template <typename DataType>
+InvokerFactory MakeInvokerFactoryHelper(CKArgs ck_args, size_t config_idx)
+{
+    return std::move([ck_args = std::move(ck_args),
+                      config_idx](const std::vector<Kernel>& kernels) mutable {
+        std::ignore = kernels;
+
+        return std::move([ck_args = std::move(ck_args), config_idx](
+                             const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            RunCKSolution<DataType>(handle, primitive_parameters, ck_args, config_idx);
+        });
+    });
+}
+
+InvokerFactory MakeInvokerFactoryThrowError()
+{
+    return [](const std::vector<Kernel>& kernels) {
+        std::ignore = kernels;
+        MIOPEN_THROW(miopenStatusNotImplemented,
+                     "Convolution operation not implemented for this data type");
+        return [](const Handle&, const AnyInvokeParams&) {};
+    };
+}
+
 InvokerFactory
 MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem,
                                        const PerformanceConfigHipImplicitGemmFwdXdlops& config)
 {
 
-    auto ck_args               = CKArgs{problem};
-    miopenDataType_t data_type = problem.conv_problem.GetInDataType();
-    auto config_idx            = config.index;
+    auto ck_args          = CKArgs{problem};
+    const auto config_idx = config.index;
 
-    InvokerFactory ret =
-        [ck_args = std::move(ck_args), data_type, config_idx](const std::vector<Kernel>& kernels) {
-            std::ignore = kernels;
-            return [ck_args, data_type, config_idx](const Handle& handle,
-                                                    const AnyInvokeParams& primitive_parameters) {
-                switch(data_type)
-                {
-                case miopenInt8:
-                    RunCKSolution<int8_t>(handle, primitive_parameters, ck_args, config_idx);
-                    break;
-                case miopenHalf:
-                    RunCKSolution<ck::half_t>(handle, primitive_parameters, ck_args, config_idx);
-                    break;
-                case miopenFloat:
-                    RunCKSolution<float>(handle, primitive_parameters, ck_args, config_idx);
-                    break;
-                case miopenInt32:
-                case miopenInt8x4:
-                case miopenBFloat16:
-                case miopenDouble: break;
-                }
-            };
-        };
-
-    return ret;
+    switch(problem.conv_problem.GetInDataType())
+    {
+    case miopenInt8: return MakeInvokerFactoryHelper<int8_t>(std::move(ck_args), config_idx);
+    case miopenHalf: return MakeInvokerFactoryHelper<ck::half_t>(std::move(ck_args), config_idx);
+    case miopenFloat: return MakeInvokerFactoryHelper<float>(std::move(ck_args), config_idx);
+    case miopenInt32:
+    case miopenInt8x4:
+    case miopenBFloat16:
+    case miopenDouble: return MakeInvokerFactoryThrowError();
+    }
 }
 
 } // namespace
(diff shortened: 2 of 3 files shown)


iq136boy:
@amberhassaan The modification failed the CI pipeline. Please fix the issue.

iq136boy:
@atamazov @DrizztDoUrden Please confirm if this is the code refactoring that you are expecting.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +226,40 @@ void RunCKSolution(const Handle& handle,
...
 
+InvokerFactory MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem, 
+    const PerformanceConfigHipImplicitGemmFwdXdlops& config) {
+
+  auto ck_args = CKArgs{problem};
+  miopenDataType_t data_type = problem.conv_problem.GetInDataType();
+  auto config_idx = config.index;
+
+
+  InvokerFactory ret = [ck_args, data_type, config_idx] (const std::vector<Kernel>& kernels) {
```
atamazov: [R] To simplify the code.
```suggestion
  InvokerFactory ret = [=] (const std::vector<Kernel>& kernels) {
```

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +226,40 @@ void RunCKSolution(const Handle& handle,
...
+        std::ignore = kernels;
+        return [ck_args, data_type, config_idx] (const Handle& handle, const AnyInvokeParams& primitive_parameters) {
```
atamazov: [R]
```suggestion
        return [=] (const Handle& handle, const AnyInvokeParams& primitive_parameters) {
```

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -188,12 +188,11 @@ namespace {
...
                    const AnyInvokeParams& primitive_parameters,
-                   const ProblemDescription& problem,
-                   const PerformanceConfigHipImplicitGemmFwdXdlops& config)
+                   const CKArgs& ck_args,
+                   size_t config_idx)
 {
-    const auto args      = CKArgs{problem};
     const auto conv_ptrs = DeviceOpPtrs<DataType>::GetInstances();
-    auto& conv_ptr       = conv_ptrs.at(config.index);
+    auto& conv_ptr       = conv_ptrs.at(config_idx);
```
atamazov, line 202: I do not know about the complexity of `DeviceOpPtrs<DataType>::GetInstances()`. Even if we assume that it is very fast right now, it belongs to CK and its complexity may quietly change in the future. I highly recommend moving computation of `conv_ptr` into the factory and capturing it by-value in the invoker.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +226,40 @@ void RunCKSolution(const Handle& handle,
     }
 }
 
+InvokerFactory MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem, 
+    const PerformanceConfigHipImplicitGemmFwdXdlops& config) {
+
+  auto ck_args = CKArgs{problem};
```
atamazov: [Notice] 👍

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +226,40 @@ void RunCKSolution(const Handle& handle,
...
+  miopenDataType_t data_type = problem.conv_problem.GetInDataType();
+  auto config_idx = config.index;
+
+
+  InvokerFactory ret = [ck_args, data_type, config_idx] (const std::vector<Kernel>& kernels) {
+        std::ignore = kernels;
+        return [ck_args, data_type, config_idx] (const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            switch(data_type)
```
atamazov: Overhead of switch is small, but I recommend moving it out of Invoker -- to GetSolution or to InvokerFactory. That would make the final Invoker object smaller at least.

atamazov (changes requested):
Much better but there is still something to optimize out from the Invoker.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
DrizztDoUrden: I don't think it would be worth the effort. There would be a whole bunch of copy-paste lambdas for each case with one line difference.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +226,40 @@ void RunCKSolution(const Handle& handle,
...
+            case miopenHalf:
+                RunCKSolution<ck::half_t>(handle, primitive_parameters, ck_args, config_idx);
+                break;
+            case miopenFloat:
+                RunCKSolution<float>(handle, primitive_parameters, ck_args, config_idx);
+                break;
+            case miopenInt32:
+            case miopenInt8x4:
+            case miopenBFloat16:
+            case miopenDouble: break;
```
DrizztDoUrden: This should be `MIOPEN_THROW("some description like unsopported data type")` as it is unlikely that noop is expected behaviour. I understand that invoker is not expected to be created in such case in the first place, but it's better to future-proof it with a transparent error message and option to select another solver as recovery during find stage.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
DrizztDoUrden:
```suggestion
  InvokerFactory ret = [ck_args = std::move(ck_args), data_type, config_idx] (const std::vector<Kernel>& kernels) {
```

DrizztDoUrden (changes requested):

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
DrizztDoUrden: Alternatively, the factory can be implemented as a class/structure with constructor and `operator ()`. We don't really use that pattern anywhere, but `std::function` should accept any functor with a valid signature. That is kind of what lambdas compile to anyway.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
DrizztDoUrden: Also, it's better to move the switch outside of the invoker as Artem has suggested. That would allow to enter recovery mode on throw rather than fail the find entirely.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
CAHEK7: It depends on how explicitly we want to capture the values. Defaulted `by value` capture may lead to the problem if someone blindly capture a vector or any other heavy-copied class. It's just available in the lambda due to defaulted behavior.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: @atamazov : See @CAHEK7 's comment. I want to make the capture explicit because `[=]` leads to programmer errors where a large object may be silently captured by value.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Good suggestion. I didn't do this because of my lack of familiarity with CK but let me look into the lifetime of `conv_ptr` and perhaps I can capture that instead.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Good suggestion on MIOPEN_THROW. @JehandadKhan : Do we have such an API or error reporting mechanism?

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Good catch on the `move` optimization.

atamazov: @CAHEK7 Very good observation! [Resolved]

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: Well, I would not identify this case as an _error_, but this is definitely a _waste of resources_ that may have some performance impact.
[Resolved]

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: @amberhassaan If you mean MIOPEN_THROW then yes. We use it to report fatal errors (which is what this case is).

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: ...or better to the solver which can MIOPEN_LOG_E and then return error code in the solution for the unhandled cases (instead of MIOPEN_THROW in the factory)

@DrizztDoUrden what is your opinion?

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: After discussion with @JehandadKhan , I concluded that we shouldn't throw an exception here as the `IsApplicable` method should implement the required checks.

atamazov: @amberhassaan If you 100% sure that IsApplicable implements the necessary checks (**_and will remain in sync with GetSolution in the future_**), then MIOPEN_THROW is just a dead code. Otherwise, please add it (or put assert at least). Pls decide yourself. [Resolved]

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: added `move` capture in 23d8931

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: @atamazov , @DrizztDoUrden : I'm posting here a combined reply to a couple of comments regarding moving `DeviceOpPtrs::GetInstances()` and moving `switch` out of the `Invoker` lambda. 

Here's why I decided not to do either:
1. The `data_type` in `ProblemDescription` is a dynamic piece of information while `ck::GetInstances()` expects a static type. Which is why we need `switch`. 
2. If I were to move the `switch` out , I'd be creating and returning a different `Invoker` lambda for each data type. Looking at the `invoker_cache.hpp` code, it seems that we don't use data type to disambiguate between Invokers of different data types for the same problem, so returning different Invokers may lead to a bug where an Invoker of one data type overwrites the cached entry of the Invoker of a different data type.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Happy to have a discussion about how to improve `InvokerFactory` pattern but IMHO it's beyond the scope of this PR.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: @amberhassaan 
> we don't use data type to disambiguate between Invokers of different data types for the same problem

We use data type: https://github.com/ROCmSoftwarePlatform/MIOpen/blob/bc8891758d15d708cc3c186b9263e9a739796e3e/src/conv/problem_description.cpp#L128

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: @amberhassaan ~It's sad, but this is exactly what the 2nd bullet of #2301 is about ;)~

atamazov: @amberhassaan Oops, how I see what you mean. Well, what @DrizztDoUrden is trying to tell, is that InvokerFactory can be implemented as "a class/structure with constructor and operator()" -- which may be thought as just another, more explicit way of implementing a lambda. The Invoker/Factory machinery and its use patterns remain the same.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: @amberhassaan We create a separate InvokerFactory for each individual Problem, this is a design choice. But kernel objects can be shared between different Invokers.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: Removed [R] ("recommendation") tag as this change is necessary to compute `conv_ptr` (https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2305#discussion_r1290087188)

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
DrizztDoUrden: > The data_type in ProblemDescription is a dynamic piece of information

That is not so in the context of `GetSolution` and produced invoker. The problem description is the specific problem we expect to get invoker for. Its content (including data types) is used to produce a key for `InvokerCache`. Everything which can change is stored in `*InvokeParams`. Technically speaking there are data types in tensor descriptors in invoke params, but we really do not expect every (or, I suppose, any) single kernel binary to be able to support multiple data type combinations.

At the moment we do not expect tensor descriptors to be changed as well. The main reason they are there is to allow invokers to do simple calculations while not catching inputs/results and to do fast validation.

In general, it's prefferable to have any non-trivial preparations for the invoker in the factory. (not GetSolution or invoker) It allows us to skip them if they are not needed. I don't think it would have noticeable impact here, but I am not very familiar with CK.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +226,42 @@ void RunCKSolution(const Handle& handle,
...
+{
+
+    auto ck_args               = CKArgs{problem};
+    miopenDataType_t data_type = problem.conv_problem.GetInDataType();
+    auto config_idx            = config.index;
+
+    InvokerFactory ret =
+        [ck_args = std::move(ck_args), data_type, config_idx](const std::vector<Kernel>& kernels) {
+            std::ignore = kernels;
+            return [ck_args, data_type, config_idx](const Handle& handle,
```
DrizztDoUrden: Factory is not expected to be executed multiple times. So in fact, you can do this:

```suggestion
        [ck_args = std::move(ck_args), data_type, config_idx](const std::vector<Kernel>& kernels) mutable {
            std::ignore = kernels;
            return [ck_args = std::move(ck_args), data_type, config_idx](const Handle& handle,
```

DrizztDoUrden (changes requested):

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Thank you both for clarification. I have moved the switch statement out of the lambda and that cleans up the error logic for unsupported types as well because now `MIOPEN_THROW` is not in the invoker lambda. See 9034f3e. 

Now, AFA moving the ck pointers out of the invoker lambda, such as `conv_ptr` and `invoker_ptr` , that did not work because `std::function` in the definition of `Invoker` expects the lambda to by copyable, while `conv_ptr` is only `moveable`. Please see (and try compiling) 3900614 (or branch amber/lambda-with-move) if you wish to see the compilation error.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: `switch` moved out in 9034f3e but `conv_ptr` and `invoker_ptr` cannot be moved out due to them being `non-copyable`. See the comment above (https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2305#discussion_r1294994063)

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Thank you for the suggestion. That was the piece I was missing yesterday and found out about it when dealing with `non-copyable` objects from CK (see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2305#discussion_r1294994063). 
The change above is implemented in the latest commit 9034f3ea

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +233,50 @@ void RunCKSolution(const Handle& handle,
...
+{
+    return std::move([ck_args = std::move(ck_args),
+                      config_idx](const std::vector<Kernel>& kernels) mutable {
+        std::ignore = kernels;
+
+        return std::move([ck_args = std::move(ck_args), config_idx](
+                             const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            RunCKSolution<DataType>(handle, primitive_parameters, ck_args, config_idx);
+        });
+    });
```
CAHEK7: [R]
`moving` something returned by value disables RVO.
But as far as it's supposed to be called only once, it shouldn't be a problem.

amberhassaan: Good catch. That `move` needs to be removed. I was trying other stuff with trying to make moving work with `ck`'s `unique_ptrs`.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +233,49 @@ void RunCKSolution(const Handle& handle,
...
+
+    switch(problem.conv_problem.GetInDataType())
+    {
+    case miopenInt8: return MakeInvokerFactoryHelper<int8_t>(std::move(ck_args), config_idx);
+    case miopenHalf: return MakeInvokerFactoryHelper<ck::half_t>(std::move(ck_args), config_idx);
+    case miopenFloat: return MakeInvokerFactoryHelper<float>(std::move(ck_args), config_idx);
+    case miopenInt32:
+    case miopenInt8x4:
+    case miopenBFloat16:
+    case miopenDouble: return MakeInvokerFactoryThrowError();
```
DrizztDoUrden:
```suggestion
    case miopenDouble: MIOPEN_THROW(miopenStatusInternalError, "ConvHipImplicitGemmFwdXdlops not implemented for this data type");
```

You should just fail rather than failing after compilation.
Plus this would mean an internal error in IsApplicable or usage inside library (even though it is recoverable) rather than some not supported parameters configuration provided by user.

DrizztDoUrden (changes requested):

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Implemented in 448628e

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Implemented in f4d0831

amberhassaan:
@atamazov , @DrizztDoUrden : Can I get an approval if you guys are happy with the current state? Thanks.

JehandadKhan (approved):

CAHEK7 (approved):

DrizztDoUrden (approved):

atamazov:
I can't review right now, but please go ahead with merge when all tests pass. If anything, we can sort it out later.

iq136boy (approved):

junliume (approved):

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: [Q] @amberhassaan Please discuss with CK team why they made `conv_ptr` non-copyable (unique_ptr?) It should be well-grounded because this prevents us from the host code optimizations.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +233,42 @@ void RunCKSolution(const Handle& handle,
...
+
+        return [ck_args = std::move(ck_args),
+                config_idx](const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            RunCKSolution<DataType>(handle, primitive_parameters, ck_args, config_idx);
+        };
+    };
+}
+
+InvokerFactory
+MakeInvokerFactoryHipImplGemmFwdXdlops(const ProblemDescription& problem,
```
atamazov, line 250: [R] I don't see the point in introducing this function because it's only called once.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -227,6 +233,42 @@ void RunCKSolution(const Handle& handle,
     }
 }
 
+template <typename DataType>
+InvokerFactory MakeInvokerFactoryHelper(CKArgs ck_args, size_t config_idx)
```
atamazov, line 237: [R] Rename to `MakeInvokerFactory`

src/include/miopen/solver.hpp
```diff
@@ -4561,7 +4561,7 @@ struct PerformanceConfigHipImplicitGemmFwdXdlops
     PerformanceConfigHipImplicitGemmFwdXdlops() : PerformanceConfigHipImplicitGemmFwdXdlops(0, "")
     {
     }
-    PerformanceConfigHipImplicitGemmFwdXdlops(bool)
+    explicit PerformanceConfigHipImplicitGemmFwdXdlops(bool)
```
atamazov, line 4564: [Q] Why `explicit` is necessary?

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: [Q] @amberhassaan 
> _@atamazov wrote:_
> I do not know about the complexity of `DeviceOpPtrs<DataType>::GetInstances()`. Even if we assume that it is very fast right now, it belongs to CK and its complexity may quietly change in the future.

Can you discuss this topic with CK team? I am asking because if `DeviceOpPtrs<DataType>::GetInstances()` is very fast and it is guaranteed that it will remain very fast in the future, then perhaps there is no point in this optimization.

atamazov: [Q] @DrizztDoUrden 
> _@amberhassaan wrote_:
> `std::function` in the definition of Invoker expects the lambda to by copyable

Can you please check if it is possible to avoid this?

atamazov (commented):
Some [R]ecommendations and [Q]uestions.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: [Resolved]

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: I had a discussion with the CK team. The are large objects and that's why they are wrapped in `unique_ptr`. What I'm going to do in a follow up PR (coming soon) is that I will cache the `unique_ptr` as a `shared_ptr` in the `invoker` lambda and then it becomes copyable and `std::function` can accept it.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: Agree. Also see my comment above about `shared_ptr` (https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2305#discussion_r1303583199).

src/include/miopen/solver.hpp
amberhassaan: Any custom defined constructor that takes a single parameter must be marked `explicit` to avoid automatic conversion. For example, without `explicit`, if I pass `bool` var to a `fun(const PerformanceConfigXXX& obj)`, `C++` will silently create an object along the way instead of throwing an error. `explicit` causes the compiler to throw an error.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
atamazov: @amberhassaan 
> The are large objects and that's why they are wrapped in unique_ptr

If an object can be copied correctly, then disabling copying may be considered as a kind of over-design (unless the objects are really huge). BTW, maybe CK guys can replace unique_ptr with shared_ptr (which can be copied)?

amberhassaan: Yes, my plan is to convert the `unique_ptr` to `shared_ptr` in our code. `CK` doesn't need to change their return type yet.

src/include/miopen/solver.hpp
atamazov: [Resolved]

atamazov:
@amberhassaan It seems we have now reached a consensus. Would you like me to create a post-merge ticket to track unresolved comments?

amberhassaan:
> @amberhassaan It seems we have now reached a consensus. Would you like me to create a post-merge ticket to track unresolved comments?

Sure, actually you can edit https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2336

atamazov:
@amberhassaan Sure! #2301 and #2336 updated.

amberhassaan:
@atamazov : I have put up #2338 to address comments for this PR and @CAHEK7 will take over the follow up work for #2286 in a similar fashion. I will help review the code as much as I can but I will be focusing on some other work.

atamazov:
@amberhassaan Thanks for the clarification. #2301, #2336, #2338 updated accordingly.

> @CAHEK7 will take over the follow up work for https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2286 in a similar fashion.

Do you mean post-merge fixes of the code merged from #2286?

amberhassaan:
Yes, that's correct.

Issue #2080, opened 2023-04-08
Deprecate some old static solvers for new ASICs


atamazov:
- [x] **(1) Some old static solvers should be allowed only for the following targets:**
  - gfx900/906 (Vega10/20, MI50/60, Radeon VII etc)
  - gfx908/90a (MI100, MI200)
  - gfx103X (Navi 2X)
  - ℹ️ Note that gfx110X (Navi 3X) is NOT included
  - Env.var `MIOPEN_DEBUG_ENABLE_DEPRECATED_SOLVERS` can be used to override, i.e. enable old static solvers for all targets.
  - ✔️ Implemented in #2081

- **(2) The list of solvers:**
  - ConvAsm1x1U
  - ConvAsm1x1UV2
  - ConvAsm3x3U
  - ConvAsm5x10u2v2b1
  - ConvAsm5x10u2v2f1
  - ConvAsm7x7c3h224w224k64u2v2p3q3f1
  - ConvAsmBwdWrW1x1
  - ConvAsmBwdWrW3x3
  - ConvCkIgemmFwdV6r1DlopsNchw
  - ConvHipImplicitGemmBwdDataV1R1
  - ConvHipImplicitGemmBwdDataV1R1Xdlops
  - ConvHipImplicitGemmBwdDataV4R1
  - ConvHipImplicitGemmBwdDataV4R1Xdlops
  - ConvHipImplicitGemmForwardV4R4Xdlops
  - ConvHipImplicitGemmForwardV4R4Xdlops_Padded_Gemm
  - ConvHipImplicitGemmForwardV4R5Xdlops
  - ConvHipImplicitGemmV4R1Fwd
  - ConvHipImplicitGemmV4R4Fwd
  - ConvHipImplicitGemmV4R4WrW
  - ConvHipImplicitGemmWrwV4R4Xdlops
  - ConvHipImplicitGemmWrwV4R4Xdlops_Padded_Gemm
  - ConvOclBwdWrW1x1
  - ConvOclBwdWrW2<N>
  - ConvOclBwdWrW2NonTunable
  - ConvOclBwdWrW53
  - ConvOclDirectFwd
  - ConvOclDirectFwd11x11
  - ConvOclDirectFwd1x1
  - ConvOclDirectFwdGen

- [ ] **(3) The solvers listed at (2) are to be removed when support for the targets listed at (1) is discontinued.**

[...]

PR #2306, opened 2023-08-07
[HotFix] Disable HIP iGEMM V6R1 DLOps Kernels

junliume:
The actual issue:

These commands will get stuck while executing on MI200s:
```
root@hpe-sjc2-43:/opt/rocm# ./bin/MIOpenDriver conv -n 256 -c 256 -H 2 -W 2 -k 512 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 1 -t 1
MIOpenDriver conv -n 256 -c 256 -H 2 -W 2 -k 512 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 1 -t 1
^C


root@hpe-sjc2-43:/opt/rocm# ./bin/MIOpenDriver conv -n 512 -c 256 -H 2 -W 2 -k 512 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 1 -t 1
MIOpenDriver conv -n 512 -c 256 -H 2 -W 2 -k 512 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 1 -t 1
Memory access fault by GPU node-8 (Agent handle: 0x2430240) on address 0x7f71c7e18000. Reason: Unknown.
Aborted (core dumped)
```

But for any other batch sizes it will pass:
```
root@hpe-sjc2-43:/opt/rocm# ./bin/MIOpenDriver conv -n 522 -c 256 -H 2 -W 2 -k 512 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 1 -t 1
MIOpenDriver conv -n 522 -c 256 -H 2 -W 2 -k 512 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 1 -t 1
MIOpen Forward Conv. Algorithm: 5, Solution: 107/ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC
GPU Kernel Time Forward Conv. Elapsed: 0.043342 ms (average)
stats: fwd-conv1x1u2, 522, 256, 1, 1, 1, 1, 512,  136839168, 2662400, 1069056, 3157, 86, 0.043342
Forward Convolution Verifies OK on CPU reference (3.71208e-08)
```

It is found that in these cases, `MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW=0` will let the configurations pass.

HIP iGEMM V6R1 DLOps Kernels are not using XDLOps and should be less efficient than ASM iGEMM XDLOps kernels.


src/solver/conv_ck_igemm_fwd_v6r1_dlops_nchw.cpp
```diff
@@ -83,7 +85,7 @@ bool PerformanceConvCkIgemmFwdV6r1DlopsNchw::IsValid(const ProblemDescription& p
 bool ConvCkIgemmFwdV6r1DlopsNchw::IsApplicable(const ConvolutionContext& ctx,
                                                const ProblemDescription& problem) const
 {
-    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW{}))
+    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW{}) || WORKAROUND_SWDEV_411729)
```
atamazov: Let's allow overriding the W/A with env var
```suggestion
#if WORKAROUND_SWDEV_411729
    if(!miopen::IsEnabled(MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW{}))
#endif
    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW{}))
#endif
```

atamazov (changes requested):

atamazov:
@junliume Shall we open a ticket and assign an engineer to investigate the problem and find root reason. I can do that for you, just let me know.

junliume:
> @junliume Shall we open a ticket and assign an engineer to investigate the problem and find root reason. I can do that for you, just let me know.

@atamazov  thank you! and only if you have some time then. This kernel is not using XDLOps and digging into the root cause might be time consuming, I would suggest take a look while not spending too much time on it.

atamazov:
@junliume 

> ...
> But for any other batch sizes it will pass:
> 
> ```
> ...
> MIOpen Forward Conv. Algorithm: 5, Solution: 107/ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC
> ```

As you can see, in this example MIOpen selects another solver, ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC (I guess because of HIP iGEMM V6R1 DLOps is not applicable) and that if why the test is passing.

junliume:
@atamazov Yes! For some reason, when the mini batch size is in the form `n == 2^m`, the find algorithm thinks the HIP iGEMM kernels are faster, otherwise the ASM iGEMM will kick in.
I assume actually in all cases ASM should kick in instead.

atamazov:
@junliume 
> thank you! and only if you have some time then. 

I mean only opening a ticket and proposing some assignment ;) Because I do not have MI200 on hand. BTW `ConvCkIgemmFwdV6r1DlopsNchw` is in the list of deprecated solvers (#2080), so we can simply disable it ~for MI200~ as a W/A and forget about it...? @JehandadKhan Do you agree?

junliume:
> @junliume
> 
> > thank you! and only if you have some time then.
> 
> I mean only opening a ticket and proposing some assignment ;) Because I do not have MI200 on hand. BTW `ConvCkIgemmFwdV6r1DlopsNchw` is in the list of deprecated solvers, so we can simply disable it for MI200 as a W/A and forget about it...? @JehandadKhan Do you agree?

@atamazov actually, we can reproduce this issue on Navi systems, since the kernel itself does not utilize XDLOps.

atamazov:
@junliume Please add
```bash
# MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW=1 is due to WORKAROUND_SWDEV_411729
...
MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW=1
```
to `smoke_solver_ConvCkIgemmFwdV6r1DlopsNchw` in test\CMakeLists.txt or it will fail.

atamazov:
@junliume 
> actually, we can reproduce this issue on Navi systems, since the kernel itself does not utilize XDLOps.

Oh, I see. Let's check that the reason of the issue is NOT the recent compiler changes and then we can forget about it (well... er... open some low-urgency ticket ;). Otherwise everything is in danger and it seems reasonable to escalate the issue to the compiler team.

test/CMakeLists.txt
```diff
@@ -1756,7 +1756,8 @@ COMMAND ${IMPLICITGEMM_TESTING_ENV} MIOPEN_LOG_LEVEL=5 $<TARGET_FILE:test_conv2d
 
 set(CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW_ENV
```
atamazov, line 1758: [Recommendation] Can you please add comment from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2306#issuecomment-1668655231 to explain why we use env var to enable the solver.

test/CMakeLists.txt
```diff
@@ -2065,7 +2066,7 @@ add_custom_test(smoke_solver_ConvAsmImplicitGemmGTCDynamicFwdDlopsNCHWC GFX900_D
 add_custom_test(smoke_solver_ConvCkIgemmFwdV6r1DlopsNchw GFX103X_ENABLED HALF_ENABLED TEST_TUNING
     COMMAND MIOPEN_FIND_ENFORCE=SEARCH_DB_UPDATE MIOPEN_DEBUG_TUNING_ITERATIONS_MAX=2
       MIOPEN_DEBUG_CONVOLUTION_ATTRIB_FP16_ALT_IMPL=0
-      MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvCkIgemmFwdV6r1DlopsNchw $<TARGET_FILE:test_conv2d>
+      MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvCkIgemmFwdV6r1DlopsNchw MIOPEN_DEBUG_CONV_CK_IGEMM_FWD_V6R1_DLOPS_NCHW=1 $<TARGET_FILE:test_conv2d>
```
atamazov, line 2071: Ditto.

atamazov (approved):
LGTM

test/CMakeLists.txt
junliume: Done. Next time I will do them all at once instead of separate commits ;-P

zjing14 (approved):

PR #2312, opened 2023-08-09
Add ability to log to roctx.  Currently just for functions

mwootton:
Add the ability to log function calls to roctx.  This allows attached profilers to see which MIOpen calls are being called by application and which kernels are being invoked by MIOpen.
Enabled via the MIOPEN_ENABLE_LOGGING_ROCTX env var.

### DIFF
--- a/src/include/miopen/logger.hpp
+++ b/src/include/miopen/logger.hpp
@@ -253,9 +253,9 @@ inline const void* LogObjImpl(const void* x) { return x; }
 template <class T, typename std::enable_if<(std::is_pointer<T>{}), int>::type = 0>
 std::ostream& LogParam(std::ostream& os, std::string name, const T& x, bool indent = true)
 {
-    if (indent)
+    if(indent)
         os << '\t';
-    os <<  name << " = ";
+    os << name << " = ";
     if(x == nullptr)
         os << "nullptr";
     else
@@ -266,16 +266,17 @@ std::ostream& LogParam(std::ostream& os, std::string name, const T& x, bool inde
 template <class T, typename std::enable_if<(not std::is_pointer<T>{}), int>::type = 0>
 std::ostream& LogParam(std::ostream& os, std::string name, const T& x, bool indent = true)
 {
-    if (indent)
+    if(indent)
         os << '\t';
     os << name << " = " << get_object(x);
     return os;
 }
 
 template <class T>
-std::ostream& LogParam(std::ostream& os, std::string name, const std::vector<T>& vec, bool indent = true)
+std::ostream&
+LogParam(std::ostream& os, std::string name, const std::vector<T>& vec, bool indent = true)
 {
-    if (indent)
+    if(indent)
         os << '\t';
     os << name << " = { ";
     for(auto& val : vec)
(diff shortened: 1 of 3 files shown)


src/include/miopen/logger.hpp
```diff
@@ -383,6 +408,31 @@ class LogScopeTime
 #define MIOPEN_LOG_SCOPE_TIME
 #endif
 
+class LogScopeRoctx
+{
+public:
+    LogScopeRoctx() {};
```
CAHEK7:
```suggestion
    LogScopeRoctx() = default;
```

src/include/miopen/logger.hpp
```diff
@@ -383,6 +408,31 @@ class LogScopeTime
...
+    LogScopeRoctx(const std::string &name)
```
CAHEK7:
```suggestion
    explicit LogScopeRoctx(const std::string &name)
```
But it seems that kind of constructor is not required at all. The usecase is to construct empty object at the begging of the scope and then assign the name if tracing enabled. There is no way to construct it together with the name. All other constructors should be explicitly deleted or properly implemented if we remove this constructor.

src/include/miopen/logger.hpp
```diff
@@ -300,7 +318,14 @@ std::ostream& LogParam(std::ostream& os, std::string name, const std::vector<T>&
             miopen_log_func_ss << miopen::LoggingPrefix() << "}" << std::endl;          \
             std::cerr << miopen_log_func_ss.str();                                      \
         }                                                                               \
-    while(false)
+        if(miopen::IsLoggingToRoctx())                                                  \
+        {                                                                               \
+            std::ostringstream miopen_log_func_ss;                                      \
+            miopen_log_func_ss << "s_api = "<< __FUNCTION__ << " | ";                   \
+            MIOPEN_PP_EACH_ARGS(MIOPEN_LOG_FUNCTION_EACH_ROCTX, __VA_ARGS__)            \
+            logtx.logRange(miopen_log_func_ss.str().c_str());                           \
```
CAHEK7: `logRange` gets `std::string` as parameter and calls `c_str` internally. In this case it constructs `std::string` just for no reason.

```suggestion
            logtx.logRange(miopen_log_func_ss.str());                                   \
```

src/include/miopen/logger.hpp
```diff
@@ -300,7 +318,14 @@ std::ostream& LogParam(std::ostream& os, std::string name, const std::vector<T>&
             miopen_log_func_ss << miopen::LoggingPrefix() << "}" << std::endl;          \
             std::cerr << miopen_log_func_ss.str();                                      \
         }                                                                               \
-    while(false)
+        if(miopen::IsLoggingToRoctx())                                                  \
+        {                                                                               \
+            std::ostringstream miopen_log_func_ss;                                      \
```
CAHEK7, line 325: Just a suggestion. Shouldn't the function name combining be unified with `if(miopen::IsLoggingFunctionCalls())`?
I mean all that `__FUNCTION__` \ `__PRETTY_FUNCTION__` and so on.
It's just an idea and you can just resolve it without any comments or changes.

CAHEK7 (dismissed):

junliume:
@mwootton please follow https://github.com/ROCmSoftwarePlatform/MIOpen/wiki/How-to-format-code to format the code too. Thanks!

junliume:
@mwootton could you please follow up on this PR? Thanks!

junliume:
@mwootton should we consider moving this out of ROCm 6.0 target?

src/include/miopen/logger.hpp
mwootton: Yeah.  The function name and formatting are very different.  The arguments into LogParam are different.  The way it is there are only those two outer guard checks.  I like that better than having guard checks on each place the delimiter and args are added.  And I mostly left the original code in place line for line, less chance of me breaking it. :)

src/include/miopen/logger.hpp
mwootton: Opps, I think I changed the arg type there.  Good catch.

src/include/miopen/logger.hpp
mwootton: This is the autolocker/scoped-marker pattern.  We SHOULD have passed the scope name into the constructor .  See the LogScopeTime class right above.  The "empty constructor/pass the name in later" is the hack that lets up keep the marker class in the stack-frame that called the logging macro.

Just because we are using the hack 100% of the time doesn't mean we shouldn't have the proper constructor. :)

CAHEK7:
I guess it's better to continue in #2391

PR #2317, opened 2023-08-10
SWDEV-416566 - Set composablekernel as a package dependency for miopen-hip

raramakr:
miopen-hip package doesn't have a dependency to composablekernel package Added composablekernel-dev/devel as a package dependency if composablekernel is enabled

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -466,6 +466,12 @@ elseif(MIOPEN_BACKEND STREQUAL "OpenCL")
     set(CPACK_RPM_PACKAGE_CONFLICTS miopen-hip)
 endif()
 
+if(MIOPEN_USE_COMPOSABLEKERNEL)
+    # Add composablekernel-dev/devel as package dependency
+    set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, composablekernel-dev")
+    set(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, composablekernel-devel")
+endif()
+
 # Unpack DB files
 function(unpack_db db_bzip2_file)
     set(KERNELS_DIR "${CMAKE_SOURCE_DIR}/src/kernels")


JehandadKhan (changes requested):
Since MIOpen builds CK using a specific commit hash and links statically to it, this should not be merged.

raramakr:
Since MIOpen links statically with CK, package dependency is not required

pramenku:
Since MIOpen builds CK using a specific commit hash and links statically to it, this should not be merged.
>> This is not correct for ROCM CI

IN ROCM CI, we have split the package and we are not using specific commit hash as kept in MIOpen.
So, the package is generated out of ROCM CI or what we are publicly releasing is not going as said in above from ROCm5.7

May be DEV CI is doing but we need to go with ROCM CI for QA/public release

@raramakr @junliume @JehandadKhan 

Thanks.

pramenku:
CK will follow the BRANCHING steps so it will have below branches for releases 

https://github.com/ROCmSoftwarePlatform/composable_kernel/tree/release/rocm-rel-5.7

For Mainline

https://github.com/ROCmSoftwarePlatform/composable_kernel/tree/amd-master

Goal is to do branching for CK like we are doing for MIOpen.

cc:  @amd-aakash

pramenku:
I am re-opening this pull request

@raramakr  @JehandadKhan  we need to merge

pramenku:
Adding Test: tuna_net

-- Configuring done

CMake Error at src/CMakeLists.txt:692 (target_link_libraries):

  Target "MIOpen" links to:

 

    composable_kernel::device_operations

 

  but the target was not found.  Possible reasons include:

 

    * There is a typo in the target name.

    * A find_package call is missing for an IMPORTED target.

    * An ALIAS target is missing.

 

 

-- Generating done

CMake Generate step failed.  Build files cannot be regenerated correctly.

make: *** [Makefile:8469: cmake_check_build_system] Error 1

root@ixt-hq-52:~/MLOpen/build_hip#

we are doing "make check" to build miopen ut and there it's needs CK otherwise test compilation will fail


so either miopen install automatic ck which makes user life easy 

or

user need to specifically install to build the test

raramakr:
Testing of miopen-hip passed without any dependencies to package CK.
Since statically linked no need of linkage. closing the PR

PR #2319, opened 2023-08-11
[driver] Fix prng seed by default

### LINKED ISSUE #2313: [driver][convolutions] The verification cache feature is broken

----
- Proposed assignees:
  - @gonidelis 
  - @CAHEK7 
  - @atamazov

CAHEK7:
Proposed solution for #2313

Probably new env variable should be documented somewhere.

@atamazov could you take a look?

### DIFF
--- a/driver/random.hpp
+++ b/driver/random.hpp
@@ -2,8 +2,8 @@
 #define GUARD_RANDOM_GEN_
 
 #include <miopen/env.hpp>
-#include <miopen/logger.hpp>
-
+#include <iostream>
+#include <cstdlib>
 #include <random>
 
 std::minstd_rand& get_minstd_gen()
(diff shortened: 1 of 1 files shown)


driver/random.hpp
```diff
@@ -1,11 +1,23 @@
...
 std::minstd_rand& get_minstd_gen()
 {
-    static thread_local std::minstd_rand minstd_gen(std::random_device{}());
+    static thread_local std::minstd_rand minstd_gen{[]() {
+        auto external_seed = miopen::EnvvarValue("MIOPEN_DRIVER_PRNG_SEED", 100500);
+
+        auto seed = external_seed == 0
+                        ? std::random_device{}()
+                        : static_cast<std::random_device::result_type>(external_seed);
+        MIOPEN_LOG_I("Random seed: " << seed);
```
atamazov: Please go not use MIOPEN_LOG macros in the driver. Just output to `cout`.

atamazov (changes requested):
Almost good!

What if we'll use environment variable (e.g. `MIOPEN_DRIVER_PRNG_LEGACY=1/0` etc) to fall back to the old code? Developers who want to use old cached data (like me) would appreciate that. I suspect that regeneration of huge amount of data may take dozen of hours.

CAHEK7:
Added MIOPEN_USE_LEGACY_PRNG and MIOPEN_DRIVER_PRNG_SEED env variables.
If MIOPEN_DRIVER_PRNG_SEED == 0, it means random seed - can be useful for smoke testing.

dmikushin:
Please take note that std::random_device [takes some costs](https://stackoverflow.com/a/39301751/4063520) and side-effects associated with random state creation taken seriously. Almost definitely it should be enough to produce a time()-based seed.

PR #2295, opened 2023-08-02
[NFC] Moved DetectRocm calls to constructor of ExecutionContext

DrizztDoUrden:
Implements #2289

PR #2323, opened 2023-08-14
[NFC] Replace long integers with int64_t and size_t for better compatibility with Windows

averinevg:
* Replace `long` integers with `int64_t` and `size_t` for better compatibility with Windows
* Replace `uint` with `unsigned int`
* Remove `C` style casting in the changed files


averinevg:
@atamazov Please review

driver/conv_driver.hpp
```diff
@@ -1564,7 +1564,7 @@ void ConvDriver<Tgpu, Tref>::PrintForwardTime(const float kernel_total_time,
         std::tie(out_n, out_c, out_h, out_w) =
             miopen::tien<4>(miopen::deref(outputTensor).GetLengths());
 
-        size_t flopCnt = 2L * in_n * in_c * wei_h * wei_w * out_c * out_h * out_w / group_count;
+        size_t flopCnt = 2LL * in_n * in_c * wei_h * wei_w * out_c * out_h * out_w / group_count;
```
atamazov: It should be ULL or cast to std::size_t. Ditto other similar places in this file.

src/ocl/dropoutocl.cpp
```diff
@@ -290,25 +290,25 @@ void DropoutDescriptor::DropoutForward(const Handle& handle,
         kernels.front()(pstates,
                         dropout,
                         amp_scale,
-                        int(in_len[1]),
-                        int(in_len[2]),
-                        int(in_len[3]),
-                        int(in_len[4]),
+                        static_cast<int>(in_len[1]),
```
atamazov, line 293: [Notice]
Well, formally the portable heterogeneous program should use the following scalar data types for passing parameters to kernels:
- For OpenCL kernels: cl_int for passing 32-bit integers, cl_uint for passing 32-bit unsigned integers etc (see https://registry.khronos.org/OpenCL/sdk/1.0/docs/man/xhtml/scalarDataTypes.html
- For HIP kernels: use native C++ types (int for int etc)
- For ASM kernels: I would vote for using types with explicit, platform-independent bitness like uint32_t etc.

Here we use HIP convention for passing parameters to OCL kernels. There is no harm, I just would like to let you know.

driver/conv_driver.hpp
CAHEK7: Is it better to use `ULL`? We've already cast to `size_t` which is unsigned.
The same applies to the rest `LL` changes.

But for the other places `static_cast<size_t>` used for literals with `UL`.

src/ocl/dropoutocl.cpp
CAHEK7, line 293: It's mostly a question, but why do you prefer cast instead of constructor call?
Is it to be more explicit?

I'm trying to adapt current coding conventions and is it mean that we should always use `static_cast<int>()` instead of `int()`?

The only serious drawback of the constructor call is that you can't say `unsigned int(val)`.

src/ocl/dropoutocl.cpp
```diff
@@ -290,25 +290,25 @@ void DropoutDescriptor::DropoutForward(const Handle& handle,
...
+                        static_cast<int>(in_str[0]),
+                        static_cast<int>(in_str[1]),
+                        static_cast<int>(in_str[2]),
+                        static_cast<int>(in_str[3]),
                         reserveSpace,
-                        uint(total_work),
-                        uint(in_offset),
-                        uint(out_offset),
-                        uint(rsvsp_offset));
+                        static_cast<unsigned>(total_work),
```
CAHEK7, line 308: I know `int` is implicit, but shouldn't it be explicit `unsigned int`?

src/solver/conv_ocl_dir2D_bwdWrW_53.cpp
```diff
@@ -166,7 +166,7 @@ static inline miopenStatus_t ComputeInputParams(
 
     // As each width chunk starts to get split,
     // it should include complete kernel filter in horizontal span.
-    const uint filter_adjustment = problem.GetWeightsWidth() - 1;
+    const unsigned filter_adjustment = problem.GetWeightsWidth() - 1;
```
CAHEK7, line 169: Implicit `int`?

CAHEK7 (commented):
All of the comments are mostly cosmetic changes and I just want to clarify it for myself and no insist to change anything.

UPD: I've read PR description one more time and now it's now a cosmetic request. I see a mix-up c-style cast and constructor call and description says `uint` -> `unsigned int` replacement, but the actual replacement is `uint` -> `unsigned`.

dmikushin:
We want to migrate to fixed-size type system, however [integer literals](https://en.cppreference.com/w/cpp/language/integer_literal) such as LL belong to the variable-size type system. AFAIK, there are no literals for fixed-size types, so wherever we have things like `2LL` it may impose the same side effects that we try to avoid.

src/ocl/dropoutocl.cpp
CAHEK7: Meanwhile `int(val)` is not a c-style cast, it's a constructor call. C-style cast is `(int)val`.

src/ocl/tensorocl.cpp
```diff
@@ -214,21 +214,21 @@ void OpTensor3d(const Handle& handle,
...
                        miopen_alpha1,
                        miopen_beta,
-                       long(Aoffset),
-                       long(Boffset),
-                       long(Coffset),
-                       long(total_work),
-                       long(total_work2),
-                       int(!float_equal(miopen_beta, 0.0)),
-                       int(blens[1] == 1));
+                       static_cast<int64_t>(Aoffset),
```
atamazov, line 225: [Notice] Formally, here we should use `cl_long` (OCL way), or `long long` (HIP way). `int64_t` would also work, of course.

src/solver/conv_multipass_wino3x3WrW.cpp
```diff
@@ -496,9 +496,9 @@ bool ConvWinograd3x3MultipassWrW<WinoDataH, WinoFilterH, WinoDataW, WinoFilterW>
 
     // clang-format off
     {
-        const long input_line_size = 4L * problem.GetInWidth();
-        const long input_feature_map_size = input_line_size * problem.GetInHeight();
-        const long input_stack_size = input_feature_map_size * problem.GetInChannels();
+        const int64_t input_line_size = 4LL * problem.GetInWidth();
+        const int64_t input_feature_map_size = input_line_size * problem.GetInHeight();
+        const int64_t input_stack_size = input_feature_map_size * problem.GetInChannels();
```
atamazov: [R] ~I think we use `long long` as 64-bit portable integer in the host code.~
See https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2323#issuecomment-1678091548

src/include/miopen/hipoc_kernel.hpp
```diff
@@ -145,12 +145,12 @@ struct HIPOCKernelInvoke
         memcpy(hip_args, &(any_args[0].buffer[0]), any_args[0].size());
         //        copy_arg(any_args[0], hip_args, 0);
 
-        for(unsigned long idx = 1; idx < any_args.size(); idx++)
+        for(size_t idx = 1; idx < any_args.size(); idx++)
```
atamazov: [R] I am not sure if plain `size_t` is portable for C++. I think it is better to use `std::size_t` (at least where you insert it in this PR). Up to you.

atamazov (changes requested):

atamazov:
@averinevg Some Qs to clarify things:
> - Replace long integers with int64_t and size_t for better compatibility with Windows

WRT `long`. It is 4 bytes with MSVC/Windows and 8 bytes with gcc/clang/Linux, which means that `long` is non-portable for us. But `long long` seems portable between these two platforms, right? Then why we use `int64_t` in this PR?

Please take into account that if we assume that `long long` is non-robust, then it is wise to assume that `LL/ULL` suffixes are non-robust as well. Then we have to use `static_cast<uint64_t>(42)` instead of 42ULL, right? Alternatively, we can use [user-defined literals](https://en.cppreference.com/w/cpp/language/user_literal), e.g: 
```C++
constexpr int64_t operator""_I64(unsigned long long v) { return v; }
constexpr int32_t operator""_I32(unsigned long long v) { return v; }
constexpr uint64_t operator""_U64(unsigned long long v) { return v; }
...
  const auto i64 = 1_I64;
  const auto i32 = 2_I32;
  const auto u64 = 0xffffffffffffffff_U64;
  std::cout << "i64 " << sizeof(i64) << ' ' << i64 << std::endl;
  std::cout << "i32 " << sizeof(i32) << ' ' << i32 << std::endl;
  std::cout << "u64 " << sizeof(u64) << ' ' << u64 << std::endl;
```
```bash
i64 8 1
i32 4 2
u64 8 18446744073709551615
```
However I am not sure if it is possible to produce negative numbers with user-defined literals.

> - Replace uint with unsigned int

Am I correct that the motivation is that `uint` is a kind of non-standard gcc extension?

> - Replace long with long long for numbers

[Notice] I think you mean (integer) literals, not numbers ;)

driver/conv_driver.hpp
atamazov: This review comment is duplicated at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2323#discussion_r1293542099. I recommend marking this one "resolved".

src/ocl/dropoutocl.cpp
atamazov: @CAHEK7 It _looks_ like a ctor, but please don't let C++ to fool you -- it is still a C-style cast, pls see https://en.cppreference.com/w/cpp/language/explicit_cast. static_casts are more flexibile, readable, smell stronger ;) and have much less ambiguous semantics. [R] mark this "resolved"

src/ocl/dropoutocl.cpp
atamazov: @CAHEK7 No, `unsigned` and `unsigned int` are equivalent in Standard C++. [R] mark this "resolved"

CAHEK7: That's why I've told "implicit".

driver/conv_driver.hpp
CAHEK7: Almost. But for other literals `static_cast` to explicit type has been used.
There are two problems - `ULL` vs `LL` and `static_cast` vs `LL`. The comment covers only the first part.

CAHEK7: I guess you've meant the last comment - it covers the both cases.

averinevg:
> We want to migrate to fixed-size type system, however [integer literals](https://en.cppreference.com/w/cpp/language/integer_literal) such as LL belong to the variable-size type system. AFAIK, there are no literals for fixed-size types, so wherever we have things like `2LL` it may impose the same side effects that we try to avoid.

@dmikushin No, it's not about fixed-size types. The main goal of this pull request is to remove the long type. Its size in 64-bit Linux and 64-bit Windows is different.

DrizztDoUrden:
> Then why we use int64_t in this PR?

`long long` is the same, but unlike int64_t it is not guaranteed to be. If the calculation depends on type size it's better to use a type with explicit size unless we are talking pointers. This way we guarantee that we won't have to change them again in future unless we change the limits of computations.

DrizztDoUrden:
> AFAIK, there are no literals for fixed-size types

Yes, but they are trivial to define. Like this:
```C++
// Nowdays the check is mostly redundant, but it's better to do that to get VERY upset later.
static_assert(sizeof(unsigned long long) >= 8);

consteval std::int_least8_t operator "" _int8(unsigned long long v) { return static_cast<std::int_least8_t>(v); }
consteval std::uint_least8_t operator "" _uint8(unsigned long long v) { return static_cast<std::uint_least8_t>(v); }
consteval std::int_least16_t operator "" _int16(unsigned long long v) { return static_cast<std::int_least16_t>(v); }
consteval std::uint_least16_t operator "" _uint16(unsigned long long v) { return static_cast<std::uint_least16_t>(v); }
consteval std::int_least32_t operator "" _int16(unsigned long long v) { return static_cast<std::int_least32_t>(v); }
consteval std::uint_least32_t operator "" _uint16(unsigned long long v) { return static_cast<std::uint_least32_t>(v); }
consteval std::int_least64_t operator "" _int16(unsigned long long v) { return static_cast<std::int_least64_t>(v); }
consteval std::uint_least64_t operator "" _uint16(unsigned long long v) { return static_cast<std::uint_least64_t>(v); }
```

P.S.: throwing checks can be added to each operator. They would be caught at build time as functions are consteval.

atamazov:
@averinevg In addition to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2323#issuecomment-1678091548. We can also add _SZT user-defined literal for using in `size_t` expressions like this: `auto sizetVar = 2_SZT * intVar ...`

src/include/miopen/hipoc_kernel.hpp
DrizztDoUrden: Taken from https://open-std.org/jtc1/sc22/wg21/docs/wp/html/nov97-2/lib-intro.html
> 17.4.3.1.4  Types                                   [lib.extern.types]
> 1 For  each  type  T  from  the Standard C library,26) the types ::T and
>  std::T are reserved to the implementation and, when defined, ::T shall
>  be identical to std::T.

So it's not a problem to use size_t, but it's best to completely avoid including or at least using stuff defined c headers when it's possible, including size_t. So my recomendation is to use std::size_t everywhere outside of C API.

DrizztDoUrden:
Oh, I have missed that Artem has already posted the operators, sorry :)

averinevg:
> We can also add _SZT user-defined literal ...

> Yes, but they are trivial to define. Like this: ...

@atamazov @DrizztDoUrden I'm not a big fan of such helpers, `static_cast` suits me just fine. The machine code is the same for both cases. We will be able to use literal suffix for size_t with `C++23`, let's wait a bit :)

DrizztDoUrden:
Main problem with static_cast is that it wouldn't produce a build time error. But since we are not introducing this problem that  does not have to be done now.

DrizztDoUrden (dismissed):
LGTM for the exception of several places where `LL` literal is used for `size_t` value calculation.

averinevg:
> Am I correct that the motivation is that `uint` is a kind of non-standard gcc extension?

@atamazov `uint` is not a standard data type, I think it is better to use `unsigned int` instead in the host code.

> [Notice] I think you mean (integer) literals, not numbers ;)

@atamazov Correct :)

test/gtest/group_conv3d_fwd.cpp
```diff
@@ -59,7 +59,6 @@ void SolverFwd(const miopen::TensorDescriptor& inputDesc,
     auto ctx = miopen::ConvolutionContext{};
 
     ctx.SetStream(&handle);
-    ctx.DetectRocm();
```
CAHEK7, line 62: It's a bit unrelated..

averinevg: Yes, but CI was broken by merging https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2295 and https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2286

CAHEK7: Thanks, it may explain why one of my PR has failed the tests.

atamazov:
> LGTM for the exception of several places where `LL` literal is used for `size_t` value calculation.

@averinevg Yep, please resolve https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2323#discussion_r1293542099 and go ahead ;)

CAHEK7:
I have no more questions.

DrizztDoUrden (approved):

CAHEK7 (approved):

junliume:
Ping @atamazov and @JehandadKhan before merging

JehandadKhan (approved):

junliume (approved):

PR #2324, opened 2023-08-14
[HotFix] Update requirements.txt MLIR ignore PATH for Python conda LLD

junliume:
Thanks to @jerryyin who provided the following information:

Non-MIOpen issue but we have to patch it in one way or another:

> We should drive Pytorch to fix their environment setting according to https://github.com/pytorch/builder/issues/1233. They can alter their path all the time and add their environment variable as highest priority to link with, and there's nothing MLIR can do about it.

So there could be only two solutions:
1. PyTorch fix the env setting (not happening in near term;
2. MIOpen switching the `CMAKE_IGNORE_PATH` every time PyTorch docker updates Python version.

For the second solution:
1. an alternative to `CMAKE_IGNORE_PATH` is `CMAKE_IGNORE_PREFIX_PATH` such as `-DCMAKE_IGNORE_PREFIX_PATH=/opt/conda`
2. However, the above requires using CMAKE version `3.22` which it not by default provided by Ubuntu 20.04.

Update: the urgency level is not blocker. If MLIR is built before PyTorch then it should be fine. However, if one tried to build MLIR within a PyTorch docker then we will meet the problem.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 sqlite3@3.17 -DCMAKE_POSITION_INDEPENDENT_CODE=On
 boost@1.79 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build -DCMAKE_CXX_FLAGS=" -std=c++14 -Wno-enum-constexpr-conversion "
 ROCmSoftwarePlatform/half@4f19ce3e56f3d3a17cf69f9db4ff3722f7445b0d --build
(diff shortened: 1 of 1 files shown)


jerryyin (dismissed):
Looks good. I'd actually also add `-DCMAKE_IGNORE_PREFIX_PATH=/opt/conda`. I hope adding this will at least reduce some of the pain if this build happens in an OS newer than ubuntu 20.04. In 20.04, it will act as a no-op.

jerryyin (approved):

PR #2325, opened 2023-08-15
Fix ConvCkIgemmFwdV6r1DlopsNchw solver to reflect that its not dynamic

JehandadKhan:
`ConvCkIgemmFwdV6r1DlopsNchw` is not dynamic, however the class definition marks it as dynamic. This PR updates the definition so that `ConvCkIgemmFwdV6r1DlopsNchw ` is no longer marked as Dynamic.

### DIFF
--- a/src/include/miopen/solver.hpp
+++ b/src/include/miopen/solver.hpp
@@ -2954,7 +2954,7 @@ struct ConvCkIgemmFwdV6r1DlopsNchw final : ConvTunableSolver<PerformanceConvCkIg
     bool IsApplicable(const ConvolutionContext&, const ProblemDescription&) const override;
     size_t GetWorkspaceSize(const ConvolutionContext&, const ProblemDescription&) const override;
     bool MayNeedWorkspace() const override { return true; }
-    bool IsDynamic() const override { return true; }
+    bool IsDynamic() const override { return false; }
     PerformanceConvCkIgemmFwdV6r1DlopsNchw
     GetDefaultPerformanceConfig(const ConvolutionContext&,
                                 const ProblemDescription&) const override;


JehandadKhan:
@junliume  Please review the urgency label.

JehandadKhan:
@cderb  Can you please check if this solver is marked Dynamic in Tuna

cderb:
> @cderb Can you please check if this solver is marked Dynamic in Tuna

Marked as non-dynamic.

junliume:
@carlushuang mentioned that this solver might indeed be dynamic

carlushuang:
This solver will launch 2 kernels. 1st kernel calculate all the descriptors, and store into a workspace. second kernel will have 4 input, A/B/C pointer, and pointer to workspace. So if view it as a whole, this is actually dynamic kernel

JehandadKhan:
@carlushuang Thanks for the clarification, the compilation of the first kernel for each and every new conv problem means that the compiler will launch and we would have to wait a few seconds before the kernel can launch. I suggest we still mark it as static.

carlushuang:
@JehandadKhan sure, this make sense. simliar to what compiler will do.

junliume (approved):

atamazov:
> @carlushuang 
>> This solver will launch 2 kernels. 1st kernel calculate all the descriptors, and store into a workspace. second kernel will have 4 input, A/B/C pointer, and pointer to workspace. So if view it as a whole, this is actually dynamic kernel
>
> @JehandadKhan 
> the compilation of the first kernel for each and every new conv problem means that the compiler will launch and we would have to wait a few seconds before the kernel can launch. I suggest we still mark it as static.

AFAICS from the source code, the most of convolution parameters are passed to the kernels at run time (not at compile time), therefore the kernels are dynamic. The problem is that tuning config may be different for each problem. But different tuning config means different compilation options, and that is why binary cache miss happens. In other words, it is the tuning which makes the Solver non-dynamic. This is known potential drawback of tuning.

I am writing about this so that we all have a common understanding of this problem. Please do not hesitate to ask if something is unclear or if you not agree.

PR #2326, opened 2023-08-15
disable MLIR and static solvers

JehandadKhan:


atamazov:
@JehandadKhan 
- Do we want to disable deprecated solvers for ALL targets or only for MI200?
- If we disable (but still keep) deprecated solvers (see #2080), then we have to either add `MIOPEN_DEBUG_ENABLE_DEPRECATED_SOLVERS=1` to the relevant solver-specific tests (preferred) or disable that tests.

Or this PR is not intended for merging (and thus should be considered as a draft)?

JehandadKhan:
> Or this PR is not intended for merging (and thus should be considered as a draft)?

Converted to draft!

PR #2327, opened 2023-08-15
[driver] [test] PRNG: allow custom\fixed seed, maintain backward compatibility. Update test PRNG.

### LINKED ISSUE #2313: [driver][convolutions] The verification cache feature is broken

CAHEK7:
@atamazov 

Proposed solution for #2313

I've noticed that we use duplicated prng for the tests and it seems it has to be synchronized with driver's one, since there we had the reasons to update it for the driver.

The details about prng update for the driver described in #2135 and #2234

Has to be merged after #2319 or instead of it.


driver/random.hpp
```diff
@@ -1,32 +1,46 @@
...
+#include <miopen/env.hpp>
+#include <iostream>
+#include <cstdlib>
 #include <random>
 
 std::minstd_rand& get_minstd_gen()
 {
-    static thread_local std::minstd_rand minstd_gen(std::random_device{}());
+    static thread_local std::minstd_rand minstd_gen{[]() {
+        auto external_seed = miopen::EnvvarValue("MIOPEN_DRIVER_PRNG_SEED", 100500);
```
atamazov:
```suggestion
        static const auto external_seed = miopen::Value("MIOPEN_DRIVER_PRNG_SEED", 100500);
```
Actually it seems worth moving internal functions (like `miopen::EnvvarValue()` and `miopen::IsEnvvarValueEnabled()`) into `miopen::env::impl` namespace, and enlose the public stuff into `miopen::env` but I always do not have time for that ;)

driver/random.hpp
```diff
@@ -1,32 +1,46 @@
...
 }
 
-template <typename T>
-inline T FRAND()
+bool use_legacy_prng()
 {
-    double d = std::generate_canonical<double, 1>(get_minstd_gen());
-    return static_cast<T>(d);
+    static bool legacy_prng = miopen::IsEnvvarValueEnabled("MIOPEN_USE_LEGACY_PRNG");
+    return legacy_prng;
```
atamazov:
```suggestion
    return miopen::IsEnabled("MIOPEN_DRIVER_PRNG_LEGACY"); // Allows reusing of old verification cache.
```
Let's unify env var name.

test/random.hpp
```diff
@@ -26,31 +26,52 @@
...
 
+#include <miopen/env.hpp>
+#include <iostream>
 #include <cstdlib>
+#include <random>
+
+std::minstd_rand& get_minstd_gen()
+{
+    static thread_local std::minstd_rand minstd_gen{[]() {
+        auto external_seed = miopen::EnvvarValue("MIOPEN_DRIVER_PRNG_SEED", 100500);
```
atamazov:
```suggestion
        static const auto external_seed = miopen::Value("MIOPEN_TEST_PRNG_SEED", 100500);
```

driver/random.hpp
```diff
@@ -1,32 +1,46 @@
...
 std::minstd_rand& get_minstd_gen()
 {
-    static thread_local std::minstd_rand minstd_gen(std::random_device{}());
+    static thread_local std::minstd_rand minstd_gen{[]() {
+        auto external_seed = miopen::EnvvarValue("MIOPEN_DRIVER_PRNG_SEED", 100500);
+
+        auto seed = external_seed == 0
+                        ? std::random_device{}()
+                        : static_cast<std::random_device::result_type>(external_seed);
+        std::cout << "Random seed: " << seed << "\n";
```
atamazov: [R] The message gives an impression that that the seed is random, which is not always so. Maybe "PRNG seed"? or "Seed of random generator"?

test/random.hpp
```diff
@@ -26,31 +26,52 @@
...
+                        ? std::random_device{}()
+                        : static_cast<std::random_device::result_type>(external_seed);
+        std::cout << "Random seed: " << seed << "\n";
+        return seed;
+    }()};
+
+    return minstd_gen;
+}
+
+bool use_legacy_prng()
```
atamazov: I think that tests do not have features that may require legacy PRNG. Let's remove it.

atamazov (changes requested):
Good initiative! But some polishing seems to still be needed.

atamazov:
@CAHEK7 
> Has to be merged after https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2319 or instead of it.

Let's use this one instead (and close #2319 right away). Please also copy description from there into this PR.

test/random.hpp
CAHEK7: It's not required, it is inside lambda and called only once during `static thread_local std::minstd_rand minstd_gen` initialization.

atamazov:
@CAHEK7 Maybe it is worth renaming PR to something like "[driver] PRNG: allow custom seed, maintain backward compatibility. [test] Update PRNG."

driver/random.hpp
CAHEK7: It is supposed to be inlined (probably I should add `inline` keyword). Moreover, it is already exposed to everyone, since it resides in a header.

driver/random.hpp
CAHEK7: But it's actually random sometimes) Ok, I'll change it.

CAHEK7:
Ok, let's keep only this PR

driver/random.hpp
CAHEK7: Meantime that construction
```
static thread_local std::minstd_rand minstd_gen {[]()[]{static const auto external_seed = miopen::Value()}()};
```
will trigger static member initialization 3 times just for no reasons. I insist to keep my version.

atamazov: First time, yes. Then it will not even call any function. But all right, the only thing I want to see here is `miopen::Value`, other stuff is up to you.

test/random.hpp
atamazov: Please use `Value()` and `MIOPEN_TEST_PRNG_SEED`; the rest is up to you.

CAHEK7: Already done

atamazov (approved):
LGTM

driver/random.hpp
```diff
@@ -1,32 +1,49 @@
...
 #define GUARD_RANDOM_GEN_
 
+#include <miopen/env.hpp>
+#include <iostream>
+#include <cstdlib>
 #include <random>
 
-std::minstd_rand& get_minstd_gen()
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DRIVER_PRNG_SEED)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DRIVER_PRNG_LEGACY)
```
JehandadKhan:
```suggestion
MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_DRIVER_PRNG_SEED)
MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_DRIVER_PRNG_LEGACY)
```

JehandadKhan (changes requested):

driver/random.hpp
```diff
@@ -1,32 +1,49 @@
...
 
-std::minstd_rand& get_minstd_gen()
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_DRIVER_PRNG_SEED)
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_DRIVER_PRNG_LEGACY)
+
+inline std::minstd_rand& get_minstd_gen()
 {
-    static thread_local std::minstd_rand minstd_gen(std::random_device{}());
+    static thread_local std::minstd_rand minstd_gen{[]() {
+        auto external_seed = miopen::Value(MIOPEN_DRIVER_PRNG_SEED{}, 100500);
```
CAHEK7:
```suggestion
        auto external_seed = miopen::Value(MIOPEN_DEBUG_DRIVER_PRNG_SEED{}, 100500);
```

junliume:
@CAHEK7 the "Fp16 Hip gfx90a" stage fails pretty consistently, could you check?

CAHEK7:
@junliume looks like the problem is in the test\algorithm itself - it is numerically unstable and heavily depends on input data.
It even forces prng seed:
https://github.com/ROCmSoftwarePlatform/MIOpen/blame/5c6d9743ffb22ffc9671aac719c8e46cf882487e/test/bn_spatial_test.cpp#L1145
and there is a comment which points to the problems in the algorithm:
https://github.com/ROCmSoftwarePlatform/MIOpen/blame/5c6d9743ffb22ffc9671aac719c8e46cf882487e/test/bn_spatial_test.cpp#L1127

As I can see, all of the fp16 bn_spatial tests have the same problems.

atamazov:
@CAHEK7 Most likely you are right, but if we want to merge this PR, then we have to resolve (or workaround) the issue somehow. Any proposal?

> https://github.com/ROCmSoftwarePlatform/MIOpen/blame/5c6d9743ffb22ffc9671aac719c8e46cf882487e/test/bn_spatial_test.cpp#L1145

FYI this is a different story, but this W/A needs to be fixed (generalized at first and formalized as a workaround).

CAHEK7:
It requires a bit more effort than expected - there are a few places where the code relies on a global glibc random state.

CAHEK7:
@atamazov could you check that your cache now works fine with `MIOPEN_DEBUG_DRIVER_PRNG_LEGACY=1`? Random seed set to 0 by default, so you don't need any additional env vars.
Also I'm not happy with current solution, it's mostly like a workaround with a few quite weak assumptions:
 - all the data initialized only once at the beginning 
 - nobody call PRND during run iterations
 - nobody relies on a global application wide prng which can be reset from any place

The first one has already been broken in fp16 bn_spatial tests - it generated 2 sequences, one for gpu and one for cpu for verification and both must match.

I'll create a ticker to re-design prng facility to be more reliable and predictable, but I still have doubts that we have to keep all the legacy options forever.

atamazov:
@CAHEK7 https://github.com/ROCmSoftwarePlatform/MIOpen/labels/ON_HOLD

CAHEK7:
Legacy support dropped. New PR #2337

PR #2330, opened 2023-08-16
Test to verify sync between various MIOpen DBs 

JehandadKhan:
This PR adds a new test to MIOpen which does the following:

* Parse all the fdb keys to construct `conv::ProblemDescription`
* Lookup Perf DB entries for tunable solvers to verify they exist and are valid
* Combine the `ProblemDescription` and perf db entry to look up KDB object
* If possible build the kernel and populate local binary cache
* Report various failures at critical stages in the above process
* Provide maximum debugging information to track down the root cause and fix it

This test will also be disabled by default since the time it takes to run this test is too long to be added to our CI. However, this test must be run (either using the CI or manually ) to verify the integrity of DBs whenever changes are made to a Find, Perf or Kernel DB.


src/include/miopen/find_solution.hpp
```diff
@@ -77,36 +77,37 @@ auto FindSolutionImpl(rank<1>,
         {
             using PerformanceConfig = decltype(s.GetDefaultPerformanceConfig(context, problem));
             PerformanceConfig config{};
-            if(db.Load(problem, s.SolverDbId(), config))
```
JehandadKhan, line 80: We want the passed in string perf config to take precedence over the one in the data base. If the entry exists in the database then the passed in config will not be used, which defeats the purpose of the function argument.

src/include/miopen/readonlyramdb.hpp
```diff
@@ -87,7 +87,7 @@ class ReadonlyRamDb
         return record->GetValues(id, value);
     }
 
-private:
```
JehandadKhan, line 90: The internal map object is useful in iterating over the class, I dont see any reason to keep it private. I am open to suggestions though.

test/gtest/db_sync.cpp
```diff
@@ -0,0 +1,606 @@
...
+        else
+        {
+            out_channels = std::stoi(attrs[0]);
+            out_h        = std::stoi(attrs[1]);
+            out_w        = std::stoi(attrs[2]);
+            in_h         = std::stoi(attrs[5]);
+            in_w         = std::stoi(attrs[6]);
+            in_channels  = std::stoi(attrs[4]);
+        }
+#endif
```
JehandadKhan: To my understanding the input/output in the fdb key are switched for forward vs backwards directions. However, in this case I discovered doing so would construct invalid Problem Descriptions. @DrizztDoUrden can you help me understand what is going on here ?

test/gtest/db_sync.cpp
```diff
@@ -0,0 +1,606 @@
...
+            out_channels = std::stoi(attrs[0]);
+            out_d        = std::stoi(attrs[1]);
+            out_h        = std::stoi(attrs[2]);
+            out_w        = std::stoi(attrs[3]);
+            in_channels  = std::stoi(attrs[5]);
+            in_d         = std::stoi(attrs[6]);
+            in_h         = std::stoi(attrs[7]);
+            in_w         = std::stoi(attrs[8]);
+        }
+#endif
```
JehandadKhan: Same as above but for 3D

test/gtest/db_sync.cpp
```diff
@@ -0,0 +1,606 @@
...
+                    EXPECT_TRUE(found) << "KDB entry not found for fdb-key:" << kinder.first << " Solver: " << id.ToString() << " filename:" << program_file << " compile_args:" << compile_options;
+                    checked_kdbs.emplace(std::make_pair(KDBKey{program_file, compile_options}, found));
+                    BuildKernel(kern.kernel_file, kern.comp_options);
+                }
+            }
+        }
+    }
+}
+
+TEST(DbSync, StaticFDBSync)
```
JehandadKhan: Need to disable it by default

test/gtest/db_sync.cpp
```diff
@@ -0,0 +1,606 @@
...
+            else
+                EXPECT_TRUE(pdb_vals.find(val.solver_id) ==
+                       pdb_vals.end())  << "Non-Tunable solver found in PDB" << solv.GetSolverDbId() ;
+            ++fdb_idx;
+            // If a solver used to be tunable and is no longer such, the pdb
+            // entries should be removed to reclaim space in the db
+
+            // It is possible that some perf db entries are not accessed in the map,
+            // in which case we should remove them since they are taking up
+            // space but are not useful.
```
JehandadKhan:
```suggestion

```

src/include/miopen/readonlyramdb.hpp
DrizztDoUrden: I'd rather add a getter. Or it would also be editable from outside.

test/gtest/db_sync.cpp
```diff
@@ -0,0 +1,606 @@
...
+        return miopenInt8;
+    else if(data_type == "INT8x4")
+        return miopenInt8x4;
+    else if(data_type == "INT32")
+        return miopenInt32;
+    else if(data_type == "BF16")
+        return miopenBFloat16;
+    else if(data_type == "FP64")
+        return miopenDouble;
+    throw std::runtime_error("Invalid data type in find db key");
```
DrizztDoUrden, line 83: I think test should use either miopen or std exceptions in these functions. Probably, std ones: we don't really need hip error codes in this test which is the main difference.

test/gtest/db_sync.cpp
DrizztDoUrden: They are switched at construction, not at serialization, so they should be parsed as is, like now.

test/gtest/db_sync.cpp
```diff
@@ -0,0 +1,606 @@
...
+            break;
+        else if(rc == SQLITE_ERROR || rc == SQLITE_MISUSE)
+            MIOPEN_THROW(miopenStatusInternalError, sql.ErrorMessage());
+    }
+}
+
+void CheckKDBObjects(const boost::filesystem::path& filename,
+                     const std::string& kernel_name,
+                     const std::string& kernel_args,
+                     bool& found)
```
DrizztDoUrden, line 266: Why not just return bool?

test/gtest/db_sync.cpp
```diff
@@ -0,0 +1,606 @@
...
+    }
+
+};
+
+template<>
+struct std::hash<KDBKey>
+{
+    std::size_t operator()(const KDBKey& k) const
+    {
+        using std::size_t;
```
DrizztDoUrden: I think this can be removed.

DrizztDoUrden (changes requested):

test/gtest/db_sync.cpp
JehandadKhan: So in this particular case, its fine? I will make the rest uniform.

test/gtest/db_sync.cpp
JehandadKhan: GTest Macros return void, therefore the function has to return void as well.

test/gtest/db_sync.cpp
JehandadKhan: removed

junliume:
@JehandadKhan These new tests still fails in CI:
```
[2023-08-17T23:57:01.412Z] The following tests FAILED:

[2023-08-17T23:57:01.412Z] 	296 - DBSync.KDBTargetID (Failed)

[2023-08-17T23:57:01.412Z] 	298 - DbSync.StaticFDBSync (Failed)
```

test/gtest/db_sync.cpp
DrizztDoUrden: Yes, sure.

test/gtest/db_sync.cpp
DrizztDoUrden: Oh, sorry :)

DrizztDoUrden (dismissed):

src/include/miopen/readonlyramdb.hpp
JehandadKhan: Done

test/gtest/db_sync.cpp
JehandadKhan: Done

JehandadKhan:
@junliume  This can be merged. Not a functional change.

junliume (approved):

src/hip/handlehip.cpp
```diff
@@ -484,18 +484,36 @@ Program Handle::LoadProgram(const std::string& program_name,
...
                                     this->GetMaxComputeUnits(),
                                     program_name,
                                     params,
                                     is_kernel_str);
     if(hsaco.empty())
+    {
+        const auto arch_target_id = miopen::SplitDelim(arch_name, ':');
+        if(arch_target_id.size() > 1)
+        {
+            // The target name has target ID in there, fall back on the generic code object
```
atamazov, line 504: @JehandadKhan @junliume `TargetProperties` is intended to hold values of target properties: name (string, the name of the gfx architecture) and `optional<bool>` attributes like xnack, sramecc etc. Therefore `GetTargetProperties().Name()` do not contian any compiler-speicific info, including semicolons etc (unlike the `LcOptionTargetStrings` class which resides in hip_build_utils.hpp).

Therefore the change in `Handle::LoadProgram()` is useless and can be reverted.

PR #2331, opened 2023-08-18
Perf Tests include setting Find Mode to 1

JehandadKhan:
The Perf test script sets the MIOpen find mode to 1 which increases the time perf tests take but would also result in incorrect performance numbers being logged.


DrizztDoUrden (approved):

atamazov:
@JehandadKhan 👍 Very important finding, Jehandad!

PR #2333, opened 2023-08-19
[CI] Update CI docker and suppress warnings

junliume:


junliume:
@JehandadKhan @atamazov
We have seen many issues with this update and some of them look like:
```
[2023-08-20T06:24:00.398Z] /long_pathname_so_that_rpms_can_package_the_debug_info/src/out/ubuntu-20.04/20.04/build/hip-on-rocclr/hipamd/src/hiprtc/hip_rtc_gen/hipRTC_header.h:1034:49: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
[2023-08-20T06:24:00.398Z]       __hip_assert(false && "invalid tile size");
[2023-08-20T06:24:00.398Z]                                                 ^
```

another issue:
```
[2023-08-20T06:24:00.885Z] MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, HIP version 5.7.23302, MIOpen version 3.00.0.a9f822391
[2023-08-20T06:24:00.885Z] MIOpen(HIP): Info [PrintVersionImpl] COMgr v.2.5.0, USE_HIP_PCH: 1
[2023-08-20T06:24:00.885Z] /home/jenkins/workspace/MLLibs_MIOpen_ci_rocm57/src/include/miopen/hipoc_kernel.hpp:61:36: runtime error: constructor call on misaligned address 0x7fff9a5b698c for type 'const void *', which requires 8 byte alignment
[2023-08-20T06:24:00.885Z] 0x7fff9a5b698c: note: pointer points here
```
which might be sanitizer checks related. by making this change
```
static const int second_index = (sizeof(T) + padding + 7) & ~7;
```
is *not* helping unfortunately.

atamazov:
@junliume BTW I see that submodule `fin` updated from `ebf9b355da09e10cb77e4b163a02e57be954095e` to `b2f3f4db3c3d7dd757e6d9e68719a780d8114dfa`. Diff seems empty. Please revert this change.

cderb:
> I see that submodule `fin` updated from `ebf9b355da09e10cb77e4b163a02e57be954095e` to `b2f3f4db3c3d7dd757e6d9e68719a780d8114dfa`. Diff seems empty. Please revert this change.

This is ok @atamazov. The branch update returns Fin to the tip of fin develop after it was branched for PR #2295.

cderb:
>  We have seen many issues with this update and some of them look like:

@junliume I've been addressing many of the rocm 5.7 related make check issues in an MIOpen-internal branch, I'll send you.

atamazov:
@junliume 
> @JehandadKhan @atamazov We have seen many issues with this update and some of them look like:
> 
> ```
> [2023-08-20T06:24:00.398Z] /long_pathname_so_that_rpms_can_package_the_debug_info/src/out/ubuntu-20.04/20.04/build/hip-on-rocclr/hipamd/src/hiprtc/hip_rtc_gen/hipRTC_header.h:1034:49: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
> [2023-08-20T06:24:00.398Z]       __hip_assert(false && "invalid tile size");
> [2023-08-20T06:24:00.398Z]                                                 ^
> ```

This seems like a a problem in the implementation of the precompiled HIP headers. _I do not know what `__hip_assert` is,_ but I suspect that it should not expand to nothing  when `NDEBUG` is defined. If I am right, then the simplest way to fix this in the HIP compiler/library is to define that macro like this state-of-the-art thing:
```C++
#ifndef NDEBUG
#define __hip_assert(ex) ... // normal definition of assert
#else
#define __hip_assert(ex) (void(0))
#endif
```
The code like `(void(0));` should not produce any warning. Alternatively, hipRTC_header.h should observe NDEBUG at line 1034 and do not use `__hip_assert` when NDEBUG is defined.

Again, AFAICS this happens in the PCH, not in our code. I recommend informing the guys who are responsible for PCH, then silence the warning in MIOpen permanently (I do not think it's worth a full-featured "workaround" stuff) and forgetting about this.

JehandadKhan:
> @junliume
> 
> > @JehandadKhan @atamazov We have seen many issues with this update and some of them look like:
> > ```
> > [2023-08-20T06:24:00.398Z] /long_pathname_so_that_rpms_can_package_the_debug_info/src/out/ubuntu-20.04/20.04/build/hip-on-rocclr/hipamd/src/hiprtc/hip_rtc_gen/hipRTC_header.h:1034:49: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
> > [2023-08-20T06:24:00.398Z]       __hip_assert(false && "invalid tile size");
> > [2023-08-20T06:24:00.398Z]                                                 ^
> > ```
> 
> This seems like a a problem in the implementation of the precompiled HIP headers. _I do not know what `__hip_assert` is,_ but I suspect that it should not expand to nothing when `NDEBUG` is defined. If I am right, then the simplest way to fix this in the HIP compiler/library is to define that macro like this state-of-the-art thing:
> 
> ```c++
> #ifndef NDEBUG
> #define __hip_assert(ex) ... // normal definition of assert
> #else
> #define __hip_assert(ex) (void(0))
> #endif
> ```
> 
> The code like `(void(0));` should not produce any warning. Alternatively, hipRTC_header.h should observe NDEBUG at line 1034 and do not use `__hip_assert` when NDEBUG is defined.
> 
> Again, AFAICS this happens in the PCH, not in our code. I recommend informing the guys who are responsible for PCH, then silence the warning in MIOpen permanently (I do not think it's worth a full-featured "workaround" stuff) and forgetting about this.

@atamazov  I agree, this should be fixed in the runtime headers

junliume:
> @junliume BTW I see that submodule `fin` updated from `ebf9b355da09e10cb77e4b163a02e57be954095e` to `b2f3f4db3c3d7dd757e6d9e68719a780d8114dfa`. Diff seems empty. Please revert this change.

This is to keep track of FIN's `develop` branch. I see this happens several times:
1. When updating MIOpen, FIN commit hash is updated to a "branch" commit hash,
2. When that branch commit hash is committed to FIN, there will be no difference between its `develop` branch and the one used to commit in MIOpen.

I think we need a standard procedure to update FIN :) notably, Clang formatting will change one file in FIN and I wanted to make that change a while ago.

src/include/miopen/hipoc_kernel.hpp
```diff
@@ -55,12 +55,9 @@ struct KernelArgsPair
     static const int padding      = (alignment - (sizeof(T) % alignment)) % alignment;
     static const int second_index = sizeof(T) + padding;
     KernelArgsPair(T x, U y)
-#if defined(__clang__)
-        __attribute__((no_sanitize("undefined")))
-#endif
     {
         new(buffer) T(x); // NOLINT (clang-analyzer-cplusplus.PlacementNew)
-        new(buffer + second_index) U(y);
+        alignas(U) new(buffer + second_index) U(y);
```
atamazov, line 60:
```
hipoc_kernel.hpp:63:9: error: 'alignas' attribute cannot be applied to a statement
        alignas(U) new(buffer + second_index) U(y);
```

atamazov: @junliume This looks strange ;) And indeed,
```
hipoc_kernel.hpp:63:9: error: 'alignas' attribute cannot be applied to a statement
        alignas(U) new(buffer + second_index) U(y);
```

src/include/miopen/hipoc_kernel.hpp
```diff
@@ -56,11 +56,10 @@ struct KernelArgsPair
     static const int second_index = sizeof(T) + padding;
     KernelArgsPair(T x, U y)
     {
-
         new(buffer) T(x); // NOLINT (clang-analyzer-cplusplus.PlacementNew)
         new(buffer + second_index) U(y);
     }
-    char buffer[second_index + sizeof(U)] = {};
+    alignas(U) char buffer[second_index + sizeof(U)] = {};
```
atamazov, line 62: This may change the layout of kernel arguments in memory. Without alignas(U), the required `buffer` alignment is 1 (i.e., no alignment is required). ~Therefore, the alignment of KernelArgsPair instances is also 1, and several such instances reside in memory without any gaps.~

With alignas(U), the padding required for KernelArgsPair is alginof(U)~, which may lead to gaps between instances of KernelArgsPair.~

If you see kernel failures, then please revert this change.

junliume: @atamazov I do see lots of failures on `gfx906` with
```
handlehip.cpp:80: Failed getting available memory: invalid argument
```
However, it looks more likely a runtime issue since it cannot be reproduced on other newer ASICs. But I will revert this change and try again.

atamazov: @junliume This is not related and I see the same on navi21.

Reverting this change won't resolve the issue with getting available memory.

atamazov: @junliume It seems like this change is indeed correct and should be kept.

atamazov:
@junliume I am working on this. Please expect update soon.

src/include/miopen/hipoc_kernel.hpp
junliume: @atamazov can you verify it happens on Navi21?
For Vega nodes some have problems but some other nodes do not.

atamazov: @junliume Of course, this happens on Navi21, as I've reported a while ago at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2307#issuecomment-1675977376 where you can find the dirty hacks for this. I am working on more or less regular W/A which should be suitable for merging into develop.

atamazov:
@junliume  I think that https://github.com/atamazov/MIOpen/blob/ci_rocm57_ata1 is ready for merging here. 
- Fixed potential UB in KernelArgs detected by sanitizer.
- `WORKAROUND_FAULTY_HIPMEMGETINFO_VEGA_NAVI2X` introduced.
  - If hipMemGetInfo fails, issues warning message and returns "fake" 16 GiB value. For now, this allows for the continued use of gfx906 and Nav21 nodes in development and testing.
  - W/A is GPU-agnostic because we do not know the GPU type in the place where the W/A is applied.
  - W/A may be considered as potentially dangerous because it always camouflages hipMemGetInfo failures. However, we're safe as long as hipMemGetInfo works fine (as it should on all officially supported GPUs, as the HIP runtime must be well tested before use with MIOpen).
- [CI] Fix: Applied WORKAROUND_ISSUE_1148 to a 'Fp16 Hip All gfx1030' stage.
  - This is expected to prevent random failures on Navi21 nodes. My Navi21 system fails quite consistently (but randomly ;) without this W/A.
- [CI] Quality: Removed unnecessary params.TARGET_NAVI21 from AnyGPU stages.
- Merged recent changes from ci_rocm57, ~_except_ b59403dd0eea0fa36fc6f1e778434104555a0d41~
- Merged recent changes from develop, conflicts resolved.

## 🌀 Testing

Tested on Navi21 and gfx906/60 (Radeon VII) systems. Note that my Navi21 system has the "hipMemGetInfo" issue, but my Radeon VII doesn't. Tried the followed tests on both systems:
- Debug build with sanitizers, Smoke FP32 test - PASSED
- Release build, Full FP16 test - PASSED

atamazov:
@junliume b59403dd0eea0fa36fc6f1e778434104555a0d41 is merged into https://github.com/atamazov/MIOpen/blob/ci_rocm57_ata1.

junliume:
> * WORKAROUND_FAULTY_HIPMEMGETINFO_VEGA_NAVI2X

@atamazov  I am not reproducing the hipGetMemoryInfo() error with 5.7 RC3 docker (at least so far), but indeed many CI nodes are not as stable now, and I have asked these base OS to be updated.

atamazov:
@junliume Are you going to merge https://github.com/atamazov/MIOpen/blob/ci_rocm57_ata1 into this branch? If yes, then I will double-check if hipMemGetInfo problem is resolved in RC3 and narrow (or remove) the W/A.

junliume:
> @junliume Are you going to merge https://github.com/atamazov/MIOpen/blob/ci_rocm57_ata1 into this branch? If yes, then I will double-check if hipMemGetInfo problem is resolved in RC3 and narrow (or remove) the W/A.

I wish to and need to figure out how to sync from `ci_rocm57_ata1` ...

**Update**: done.

atamazov:
I see, let me check if merge finished cleanly...

atamazov:
@junliume 🆗

atamazov:
@junliume hipMemGetInfo problem is NOT resolved in 5.7 RC3. This PR can be merged as is.

atamazov:
@junliume @JehandadKhan [Notice] Interesting detail: 
Both RC1 and RC3 report HIP version 5.7.31921 in `hip_version.h` which is not good. However, CMake (`hip-config-version.cmake`) reports different numbers:
- RC1: 5.7.23302
- RC3: 5.7.23341

"Thanks" to some problems with `hip_version.h` in some old versions of ROCm we use CMake machinery as a workaround, and the library is able to distinguish between RC1 and RC3 (even if this is not necessary yet).

junliume:
Ping @JehandadKhan and @atamazov for review after CI has passed.

JehandadKhan (approved):

PR #2337, opened 2023-08-24
[driver] [test] PRNG: allow custom\fixed seed

### LINKED ISSUE #2313: [driver][convolutions] The verification cache feature is broken

CAHEK7:
https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_blocker because we've already got randomly failed cases.
<details>
<summary>For example (GFX1100 required)</summary>

```
MIOPEN_DEBUG_DRIVER_PRNG_SEED=1372743782 ./bin/MIOpenDriver convfp16 -W 1195 -H 16 -c 1 -n 16 -k 1 -x 1 -y 1 --pad_w 12 --pad_h 5 -F 1 -t 1 -i 1 -V 1 -S 139
```

</details>

Significant PRNG re-design.

Highlights:
- switch `std::linear_congruential_engine` parameters to glibc version (speedup)
- simplify canonical number generation (speedup)
- simplify uniform distribution generation (speedup)
- provide new primitives to generate random numbers (code quality, speedup)
- remove lots of copy-pasted generators and replaced them with new primitives (code quality)
- "hide" direct access to PRNG (code quality)
- fix verification cache, at least partially
- drop legacy generator support (code quality)
- unify test and driver generators (code quality) - I see no reasons to split them:
initially they were dependent through global `rand()` state and through including files from `test` to `driver`
most probably they will suffer from the same problems, so will be updated simultaneously and I can hardly imagine separated updates. But I'm not quite sure about this unification, probably I've missed something.

I did some simple performance measurements for `AllocateBuffersAndCopy` for convolution and got following results (compare to the current develop):

`AllocateBuffersAndCopy`:
|case|old (ms)|new (ms)|gain|
|-|-|-|-|
|-N 128|1253.08|936.23|1.34|
|-N 1|43.57|47.39|1.09|

Depending on the case allocation may take from 60% to 10% of total execution time and thus total improvement may not be so significant.

Total time:
|case|old (ms)|new (ms)|gain|
|-|-|-|-|
|-N 128 -V 1|3056|3328|1.09|
|-N 1 -V 1|1709|1726|1.01|
|-N 128 -V 0|1423|1756|1.23|
|-N 1 -V 0|464|476|1.03|

<details>
<summary>The other driver parameters</summary>

```
./bin/MIOpenDriver conv --spatial_dim 2 --pad_h 0 --pad_w 0 --pad_d 0 --conv_stride_h 1 --conv_stride_w 1 --conv_stride_d 0 --dilation_h 1 --dilation_w 1 --dilation_d 0 --group_count 1 --mode conv --pad_mode default --trans_output_pad_h 0 --trans_output_pad_w 0 --trans_output_pad_d 0 --out_layout NCHW --in_d 1 --in_h 210 --in_w 210 --fil_d 1 --fil_h 3 --fil_w 3 --in_channels 16 --out_channels 16 --forw 1 -i 10 -t 1 -w 1 -s 84
```

</details>

I haven't tried it on the other cases or algorithms, but the overall trend seems to be: bigger tensors and no verification (perf runs) - more speedup.

@atamazov please join.

Closes #2313


driver/conv_driver.hpp
```diff
@@ -1423,14 +1381,25 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
...
             db_dev->ToGPU(q, db.data());
         }
+    }
 
-        if(!weiRead)
+    if(!dataRead)
+    {
+        for(int i = 0; i < in_sz; i++)
+        {
+            auto val = prng::gen_0_to_B(Data_scale);
```
atamazov, line 1416: Please comment why we generate numbers even if we don't intend to use them.

CAHEK7: Done

JehandadKhan:
@CAHEK7  Test `test_rnn_vanilla_dropout ` fails FP16 on MI100.

CAHEK7:
@JehandadKhan should be fixed now.
May I ask you to restart the tests if they failed due to infrastructure reasons?

CAHEK7:
Recently for some reasons the tests often failed on a git-checkout stage.

junliume:
@CAHEK7 I restarted full test :) Let's see if it can run through, I will ping if I see some persistent issues.

CAHEK7:
Changing PRNG triggered reduction error again:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ac3dedb5d4ba6a941c9c480d2c161efa3eb37061/test/reduce_test.cpp#L770

junliume:
This list of tests is becoming unstable with this update on major platforms:
test_conv_igemm_dynamic_xdlops_nhwc_wrw_nchw
test_conv2d
test_conv_for_implicit_gemm
test_immed_conv2d

test/reduce_test.cpp
```diff
@@ -850,12 +850,11 @@ struct reduce_driver : test_driver
...
-        // This method is slower due to the use of GET_RAND(), it is usually used for manual
-        // testing
         auto gen_value_3 = [&](auto... is) {
-            auto rand_upper   = tensor_elem_gen_integer{max_value}(is...);
-            auto sign_value   = tensor_elem_gen_checkboard_sign{}(is...);
-            double rand_ratio = static_cast<double>(GET_RAND() / (static_cast<double>(RAND_MAX)));
+            auto rand_upper = tensor_elem_gen_integer{max_value}(is...);
+            auto sign_value = tensor_elem_gen_checkboard_sign{}(is...);
+            auto rand_ratio = prng::gen_A_to_B(
+                0.1, 1.); // limit range due to numeric errors, see WORKAROUND_GPU_NUMERIC_ERROR
```
atamazov, line 840: Where can I see this W/A?

CAHEK7: It's a define in this file, which disables the test completely.
The problem is still existing and accidentally it was just hidden for some time.

atamazov: @CAHEK7 Sorry but I am unable to see how value the of rand_ratio is related to WORKAROUND_GPU_NUMERIC_ERROR.

Also it seems to me that the range of rand_ratio was not [0.1 ... 1.0] before this change. Can you please check or justify the change?

CAHEK7: @atamazov 
This test was disabled completely by `WORKAROUND_GPU_NUMERIC_ERROR` due to numerical instability. For some reasons the problem has gone but it was not fixed since it is so sensitive to the input data. From the other hand loss of precision is a quite common problem for reduction of big amount of data, when tiny values started to be ignored.
So there are:
1) reduction without loss of precision mitigation - it can be ok or not, depends on the API precision specification or guarantees
2) the test generates the data which is out of scope of the algorithm - it's not ok, because even algorithms with loss of precision mitigation can be broken by specially prepared data
3) the test uses higher precision algorithm for verification - it's not ok, because loss of precision can't be mitigated by epsilon comparison

I've chosen 2 and excluded small values from the input data.
There is also 2.5 option when limited range can be applied to that specific test like it was done by `WORKAROUND_GPU_NUMERIC_ERROR` but do not disable the test.

We can either get an API precision specification and fix reduction algorithm, and\or we can use the same precision for verification, but this PR is not going to do any of this.

atamazov: @CAHEK7 
> This test was disabled completely by WORKAROUND_GPU_NUMERIC_ERROR due to numerical instability

To clarify: we do not know the reason of the issue (and I do not remember how it manifested itself... likely a validation failure, but I am not sure about instability). I tend to blame the compiler for GPU.

> the test uses higher precision algorithm for verification - it's not ok, because loss of precision can't be mitigated by epsilon comparison

Let's discuss this 1:1.

> the test generates the data which is out of scope of the algorithm - it's not ok, because even algorithms with loss of precision mitigation can be broken by specially prepared data

This is important observation. So the main reason of this change is putting the input data back into the valid scope of the algorithm, right? Very good finding.

I recommend removing the comment about WORKAROUND_GPU_NUMERIC_ERROR from the source code, as we can neither prove nor disprove that it relates to the input range. Thanks.

test/driver.hpp
```diff
@@ -1147,14 +1143,11 @@ void run_config(std::vector<std::string>& config,
     {
         config_driver.disabled_cache = true;
     }
-
+    const auto fixed_seed = prng::details::get_prng()();
     for(int j = 0; j < test_repeat_count; j++)
     {
-        run_data(config_data_args.begin(), config_data_args.end(), [&] {
-            std::srand(65521);
```
shurale-nkn, line 1154: #.2

test/driver.hpp
```diff
@@ -916,9 +916,7 @@ struct test_driver
             }
             else
             {
-                std::srand(65521);
                 static_cast<Derived*>(this)->run();
-                std::srand(65521);
```
shurale-nkn, line 921: #.2 leads cache invalidation and test coverage modification.
Some test parameters are generated at runtime based on already set parameters plus number from generator. Therefore, it is necessary to guarantee a similar state of the generator before each argument generation.

test/tensor_holder.hpp
```diff
@@ -216,7 +217,8 @@ struct tensor
                                     });
         seed ^= data.size();
         seed ^= desc.GetLengths().size();
-        std::srand(seed);
```
shurale-nkn, line 219: #.1 tensor_holder change affects cache state.

shurale-nkn (changes requested):
**tensor_holder** was invented to simplify control of tensors and associated data buffers.
Values in associated buffer are tied to the tensor configuration to make the buffer initialization position independent and free from the last state of the generator.
Changing of this statement unnecessarily complicates debugging and increases the number of reasons for cache invalidation.

For the same reason we have set rand calls(srand(0) ...) before buffer initialization. To have stable values in buffers that does not depend on other functions that could use same generator.

Also, the state of the generator, at start of testing of an arbitrary configuration, should not depend on the number of testing configurations and their order. configuration run both as part of the test list and separately from it should have same values in the buffers.

test/tensor_holder.hpp
CAHEK7: In that case I would suggest to have multiple independent generators, but it looks like it's too big change.

test/driver.hpp
CAHEK7: done

test/driver.hpp
CAHEK7: done

test/tensor_holder.hpp
CAHEK7: done

driver/random.hpp
```diff
@@ -1,32 +1,93 @@
...
-    decltype(auto) minstd_gen = get_minstd_gen();
-    return minstd_gen();
+    if constexpr(std::is_floating_point_v<T>) // native fp
+    {
+        return gen_canonical<T>() * B;
+    }
+    else if constexpr(std::is_integral_v<T>)
+    {
+        // can only generate 27bit range, so it may not be suitable
+        // for huge 64 bit ranges, but we do not expect such ranges
```
atamazov, line 82: [R] static_assert here if sizeof(T) > 4?

test/conv_common.hpp
```diff
@@ -2255,20 +2225,19 @@ struct conv_driver : test_driver
...
-                    auto data_type    = input.desc.GetType();
-                    std::size_t v_max = is_int8 ? 16 : (data_type == miopenHalf) ? 4 : 16;
-
-                    return gen_float ? scalar_gen_random_float{-1, 1}()
-                                     : scalar_gen_random_integer{1, v_max}() *
-                                           tensor_elem_gen_checkboard_sign{}(is...);
+                    auto data_type = input.desc.GetType();
+                    int v_max      = is_int8 ? 16 : (data_type == miopenHalf) ? 4 : 16;
+                    return gen_float
+                               ? prng::gen_A_to_B(-1., 1.)
```
atamazov: [R] Do not omit "0" after dot in floating literals.

driver/random.hpp
```diff
@@ -1,32 +1,93 @@
...
+    }
+    else
+    {
+        return static_cast<T>(gen_canonical<float>());
+    }
 }
 
-inline int GET_RAND()
+template <typename T>
+inline T gen_0_to_B(T B)
```
atamazov, line 73: [Q][quality] This function is very sensitive to the type of B. For example, `static_cast<half>gen_0_to_B(4.)` and `static_cast<half>gen_0_to_B(4)` would yield totally different results. I am not sure this is good.

Maybe it’s worth making a couple of functions `gen_...` and `gen_integer_...` and removing conditional statements from the function body at once?

atamazov (commented):
Some questions and recommendations, no objections.

driver/random.hpp
CAHEK7: Yes, but not exactly. It actually depends on B. If B < 2^27 everything is fine even for 64 bit type, but I'd prefer to do not perform runtime check for it.

driver/random.hpp
CAHEK7: Probably. It will be more explicit.
But we can't get rid of conditional statement since we can't generate half and bfloat directly.
Or we can?

CAHEK7: And it will require runtime checks in the callee code.
For example:
```
template <Tgpu> run()
{
    Tgpu range = is_half ? Tgpu(4) : Tgpu(16);
    auto val = gen_0_to_B(range);
}
```

atamazov: @CAHEK7 we can provide explicit specializations for each (or for some) of the supported types, thus reducing the need in conditionals, right?

driver/random.hpp
atamazov: [resolved]

driver/random.hpp
CAHEK7: `[u]int.._t`, `size_t`, `float`, `double`, `half`, `bfloat`. For me it seems that 2 conditions is better to understand and reduce code duplication.
`if constexpr` is a modern way to do more flexible specialization and guaranteed to be performed at compile time.

atamazov: @CAHEK7 I am not against using modern stuff ;) But what about the original problem? Does it seem reasonable to split into `gen_...` and `gen_integer_...` or not? I will agree with everything you suggest ;)

test/conv_common.hpp
atamazov: Still no "0" after dot in the new code ;)

atamazov: @CAHEK7 ...but marked resolved ;)

driver/random.hpp
CAHEK7: It doesn't, because it breaks high-level templates.
Instead of passing templated parameters down to the call tree, each place must check the type and explicitly call appropriate function.
Like:
```
template <Tgpu> run()
{
    Tgpu range = is_half ? Tgpu(4) : Tgpu(16);
    auto val = gen_0_to_B(range);
}
```
and
```
template <Tgpu> run()
{
    Tgpu range = is_half ? Tgpu(4) : Tgpu(16);
    auto val;
    if constexpr (std::is_integer_v<Tgpu>)
        val = gen_integer_0_to_B(range);
    else
        val = gen_real_0_to_B(range);
}
```
And that must be done everywhere in the codebase instead of inside the lowest level function.

junliume (dismissed):
@atamazov  @JehandadKhan  last ping before merging. I would hope this PR will resolve some random failure issues, not introducing them. :)

shurale-nkn (dismissed):
LGTM

driver/random.hpp
atamazov: [Resolved]

atamazov:
LGTM!

dmikushin (approved):

junliume (approved):

Issue #2336, opened 2023-08-24
Cache CK convolution kernel objects as shared_ptr in Invoker lambda


amberhassaan:
`CK` has a `DeviceOpFactory::GetInstances()` API used by MIOpen Solvers that returns a `vector<unique_ptr<T>`, where `T` is an abstraction for a kernel. `unique_ptr<T>` makes it uncacheable inside MIOpen's `Invoker` lambda thus preventing reuse and forcing multiple calls to `GetInstances()` especially at the time of invoking a kernel. 

I will put up a PR that promotes `unique_ptr<T>` to `shared_ptr<T>` thus allowing us to cacheable inside MIOpen's `Invoker`.

atamazov:
[Notice] IIUC this ticket is related to all CK solvers where this optimization is possible (see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2305#issuecomment-1693677130)

PR #2338, opened 2023-08-25
Post Merg Fixes for 2305: Cache CK conv_ptr as a shared_ptr. Clean up by moving common code into functions

amberhassaan:
Move `ck::DeviceOpFactory::GetInstances()` call out of `Invoker` lambda and cache the `DeviceOp` pointer as a `shared_ptr` in the `Invoker` lambda. 

Cleanup by moving repeated code into common functions.

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -90,34 +90,30 @@ struct CKArgs
     template <typename ConvPtr>
     auto MakeArgPtr(const ConvPtr& conv_ptr, ConstData_t in, ConstData_t w, Data_t out) const
     {
-
-        return conv_ptr->MakeArgumentPointer(
-            const_cast<Data_t>(in), // NOLINT (cppcoreguidelines-pro-type-const-cast)
-            const_cast<Data_t>(w),  // NOLINT (cppcoreguidelines-pro-type-const-cast)
-            out,
-            this->N,
-            this->K,
-            this->C,
-            this->input,
-            this->filter,
-            this->output,
-            this->strides,
-            this->dilation,
-            this->lPadding,
-            this->rPadding,
-            {},
-            {},
-            {});
+        return conv_ptr->MakeArgumentPointer(in,
+                                             w,
+                                             out,
+                                             this->N,
+                                             this->K,
+                                             this->C,
+                                             this->input,
+                                             this->filter,
+                                             this->output,
+                                             this->strides,
+                                             this->dilation,
+                                             this->lPadding,
+                                             this->rPadding,
+                                             {},
+                                             {},
+                                             {});
     }
 
     template <typename ConvPtr>
     auto MakeArgPtr(const ConvPtr& conv_ptr, const ConvDataTensors& tensors) const
     {
-
         return MakeArgPtr(conv_ptr, tensors.in, tensors.w, tensors.out);
     }
 
-    // TODO(Amber): make method const
     template <typename ConvPtr>
     bool IsSupportedBy(const ConvPtr& conv_ptr) const
     {
@@ -173,7 +169,7 @@ bool ConvHipImplicitGemmFwdXdlops::CheckCKApplicability(const ProblemDescription
     const auto conv_ptrs = DeviceOpPtrs<DataType>::GetInstances();
     assert(!conv_ptrs.empty());
     const auto args = CKArgs{problem};
-    if(!std::all_of(args.strides.begin(), args.strides.end(), [&](auto x) { return x == 1; }))
+    if(!std::all_of(args.strides.begin(), args.strides.end(), [](auto x) { return x == 1; }))
         return false;
     for(int i = 0; i < conv_ptrs.size(); i++)
     {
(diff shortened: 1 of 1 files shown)


atamazov:
Comments to be resolved:
- [ ] 🟡 [#2336](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2305#discussion_r1290087188) - avoid invocation of `DeviceOpPtrs<DataType>::GetInstances()` in Invoker
- [ ] ⚪ [Style] Refactor `MakeInvokerFactory*` functions
  - [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2305#discussion_r1302805709
  - [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2305#discussion_r1302806115

amberhassaan:
@atamazov @DrizztDoUrden : this PR is ready for you to review. Thank you.

junliume:
@amberhassaan there are some CI issues:
```
[2023-08-29T16:12:57.841Z] /home/jenkins/workspace/ibs_MIOpen_amber_post-fixes-2305/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp:383:33: error: no member named 'GetInDataType' in 'miopen::ProblemDescription::(unnamed struct at /home/jenkins/workspace/ibs_MIOpen_amber_post-fixes-2305/src/include/miopen/problem_description.hpp:91:5)'

[2023-08-29T16:12:57.841Z]     switch(problem.conv_problem.GetInDataType())

[2023-08-29T16:12:57.841Z]            ~~~~~~~~~~~~~~~~~~~~ ^
```

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -234,39 +206,24 @@ void RunCKSolution(const Handle& handle,
 }
 
 template <typename DataType>
-InvokerFactory MakeInvokerFactoryHelper(CKArgs ck_args, size_t config_idx)
+InvokerFactory MakeInvokerFactoryHipImplGemmFwdXdlops(CKArgs ck_args, size_t config_idx)
 {
-    return [ck_args = std::move(ck_args), config_idx](const std::vector<Kernel>& kernels) mutable {
-        std::ignore = kernels;
+    auto conv_ptrs = DeviceOpPtrs<DataType>::GetInstances();
+    auto conv_ptr  = std::move(conv_ptrs.at(config_idx));
```
DrizztDoUrden: Are you sure you should move here? That invalidates the object in the map, while not removing it from there.

DrizztDoUrden (changes requested):

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
CAHEK7: 1) It must be moved since it's a `std::unique_ptr`
2) In a prefect world `conv_ptrs` should not contain any "invalid" pointers
3) but we hold an index in this array, so we can't even remove any pointers from it at all
4) and it seems it doesn't matter. `DeviceOpPtrs<DataType>::GetInstances();` returns vector by value, so `conv_ptrs` will be destroyed anyway and it's not critical that we have one invalid entry.

But this particular construction is still error-prone. It's better to get rid of `conv_ptrs` at all.

CAHEK7: Done

amberhassaan:
> @amberhassaan there are some CI issues:
> 
Seems like @CAHEK7  took care of it. Thank you both.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: What @CAHEK7 said: i.e., we own the vector so we can and must do `move` because it's a `move-only` object. 
I have noticed that other solvers `search` through the `conv_ptrs` vector by string name, so this logic will most likely change. I think it is currently broken, right @iq136boy? (Please correct me if I'm wrong).

CAHEK7:
I'll do some minor clean-up as well and apply everything to backward convolution to whatever else convolution is using CK.

DrizztDoUrden (approved):

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -108,29 +142,16 @@ void PerformanceConfigHipImplicitGemmFwdXdlops::Init(const ProblemDescription& p
...
+        if(args.IsSupportedBy(conv_ptrs[i]))
         {
             this->kernel_id = conv_ptrs[i]->GetTypeString();
             break;
         }
         ++this->index;
     }
+    // TODO(Amber): This logic is broken and different from other
+    // PerformanceConfigHipImplicitsGemm* classes. Other classes keep a vector of
+    // strings called 'valid_kernels'
```
CAHEK7, line 154: It is indeed broken, since `valid_kernels` is used to iterate over performance configs.
I'll take care.

junliume (approved):

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
junliume: Thanks and probably in another PR :)

PR #2343, opened 2023-08-30
RNN padding and SeqTensorDescriptor

shurale-nkn:


src/rnn_api.cpp
```diff
@@ -503,6 +579,206 @@ static void LogCmdRNN(const miopenTensorDescriptor_t* xDesc,
...
+                        w,
+                        weightSpaceSize,
+                        workSpace,
+                        workSpaceNumBytes,
+                        reserveSpace,
+                        reserveSpaceNumBytes);
+
+    // bfloat16 not supported for rnn operation
+    if(miopen::deref(rnnDesc).dataType == miopenBFloat16 ||
+       miopen::deref(rnnDesc).dataType == miopenBFloat16)
```
CAHEK7: Conditions are identical

src/rnn_api.cpp
```diff
@@ -36,6 +36,62 @@ enum RNNDir_t
...
+miopenSetRNNDataSeqTensorDescriptor(miopenSeqTensorDescriptor_t seqTensorDesc,
+                                    miopenDataType_t dataType,
+                                    miopenRNNBaseLayout_t layout,
+                                    int maxSequenceLen,
+                                    int batchSize,
+                                    int vectorSize,
+                                    const int* sequenceLenArray,
+                                    void* paddingMarker)
+{
+    (void)paddingMarker;
```
CAHEK7: If we support C++17, shouldn't we use attribute `[[maybe_unused]]` instead of C-style cast to `void`?
```suggestion
                                    [[maybe_unused]] void* paddingMarker)
{
```

src/rnn_api.cpp
```diff
@@ -36,6 +36,62 @@ enum RNNDir_t
...
+extern "C" miopenStatus_t
+miopenGetRNNDataSeqTensorDescriptor(miopenSeqTensorDescriptor_t seqTensorDesc,
+                                    miopenDataType_t* dataType,
+                                    miopenRNNBaseLayout_t* layout,
+                                    int* maxSequenceLen,
+                                    int* batchSize,
+                                    int* vectorSize,
+                                    int sequenceLenArrayLimit,
+                                    int* sequenceLenArray,
+                                    void* paddingMarker)
```
CAHEK7, line 74: `[[maybe_unused]]`?

src/rnn_api.cpp
```diff
@@ -36,6 +36,62 @@ enum RNNDir_t
...
+            miopen::tien<3>(miopen::deref(seqTensorDesc).GetLengths());
+
+        if(sequenceLenArray != nullptr)
+        {
+            size_t copy_sz =
+                std::min(static_cast<size_t>(sequenceLenArrayLimit),
+                         miopen::deref(seqTensorDesc).GetSequenceLengthsVector().size());
+            const auto seq_lens_begin =
+                miopen::deref(seqTensorDesc).GetSequenceLengthsVector().begin();
+            std::copy(seq_lens_begin, seq_lens_begin + copy_sz, sequenceLenArray);
```
CAHEK7: Seems `std::copy_n` is better.
```suggestion
            std::copy_n(seq_lens_begin, copy_sz, sequenceLenArray);
```

src/rnn_api.cpp
```diff
@@ -503,6 +579,206 @@ static void LogCmdRNN(const miopenTensorDescriptor_t* xDesc,
...
+                        dx,
+                        w,
+                        workSpace,
+                        workSpaceNumBytes,
+                        reserveSpace,
+                        reserveSpaceNumBytes);
+
+    // bfloat16 not supported for rnn operation
+    if(miopen::deref(rnnDesc).dataType == miopenBFloat16 ||
+       miopen::deref(rnnDesc).dataType == miopenBFloat16)
```
CAHEK7: Conditions are identical

src/rnn_api.cpp
```diff
@@ -503,6 +579,206 @@ static void LogCmdRNN(const miopenTensorDescriptor_t* xDesc,
...
+                        dw,
+                        weightSpaceSize,
+                        workSpace,
+                        workSpaceNumBytes,
+                        reserveSpace,
+                        reserveSpaceNumBytes);
+
+    // bfloat16 not supported for rnn operation
+    if(miopen::deref(rnnDesc).dataType == miopenBFloat16 ||
+       miopen::deref(rnnDesc).dataType == miopenBFloat16)
```
CAHEK7: Conditions are identical

src/seq_tensor.cpp
```diff
@@ -0,0 +1,501 @@
...
+                                      : *std::max_element(sequence_len.begin(), sequence_len.end());
+}
+
+std::size_t SeqTensorDescriptor::GetNonEmptySeqCount() const
+{
+    return (all_sequences_equal_to_max || padded_seq_layout) ? lens[0]
+           : samples_in_descending_order
+               ? std::distance(sequence_len.begin(),
+                               std::lower_bound(sequence_len.begin(), sequence_len.end(), 0))
+               : std::count_if(sequence_len.begin(), sequence_len.end(), [&](auto const& val) {
```
CAHEK7: What's the point to capture anything into this lambda?

src/include/miopen/tensor_ops.hpp
```diff
@@ -75,7 +75,13 @@ GetConsistentFlattenedTensorDescriptors(const TDescriptors&... real_descriptor_p
     for(std::size_t itensor = 0; itensor < NTensor; ++itensor)
         is_all_packed &= real_descriptors[itensor]->IsPacked();
 
-    if(is_all_packed)
+    bool is_all_same_strided        = true;
+    const auto& real_desc_0_strides = real_descriptors[0]->GetStrides();
+    for(std::size_t itensor = 0; itensor < NTensor; ++itensor)
+        if(real_desc_0_strides != real_descriptors[itensor]->GetStrides())
+            is_all_same_strided = false;
```
CAHEK7: 1) Strides with index 0 has been compared to itself
2) Early exit

```suggestion
    const auto& real_desc_0_strides = real_descriptors[0]->GetStrides();
    for(std::size_t itensor = 1; itensor < NTensor; ++itensor)
        if(real_desc_0_strides != real_descriptors[itensor]->GetStrides())
        {
            is_all_same_strided = false;
            break;
        }
```

src/rnn.cpp
```diff
@@ -897,80 +985,595 @@ void RNNDescriptor::GetLayerBiasOffset(const int layer,
...
-            }
-            left_id = i;
+void RNNDescriptor::SeqTensorToTensorDescArray(const SeqTensorDescriptor& desc,
+                                               std::vector<miopen::TensorDescriptor>& td,
+                                               std::vector<miopenTensorDescriptor_t>& ptd)
+{
+    if(!desc.IsPacked())
+        MIOPEN_THROW(miopenStatusInternalError, "Only packed SeqTensorDescriptor supported.");
+
+    const std::vector<size_t> bs = desc.GetBatchesPerSequence();
```
CAHEK7, line 1079: Should it be copied?
```suggestion
    const std::vector<size_t>& bs = desc.GetBatchesPerSequence();
```

src/rnn.cpp
```diff
@@ -897,80 +985,595 @@ void RNNDescriptor::GetLayerBiasOffset(const int layer,
...
+    const size_t vector_size     = desc.GetLengths()[2];
+    const auto data_type         = desc.GetType();
+
+    std::transform(bs.begin(), bs.end(), std::back_inserter(td), [&](size_t batch_size) {
```
CAHEK7: Why `vector_size` and `data_type` have been captured by ref? They are both kind of basic types and capturing by value is much more efficient.
Also I'd prefer explicit capturing, but it's up to you.

src/rnn.cpp
```diff
@@ -897,80 +985,595 @@ void RNNDescriptor::GetLayerBiasOffset(const int layer,
...
+            padded_sequences};
+}
 
-                src_offset += packed_desc.GetElementSpace();
-                dst_offset += WA_padded_ElementSpace;
-            }
-            left_id = i;
+void RNNDescriptor::SeqTensorToTensorDescArray(const SeqTensorDescriptor& desc,
+                                               std::vector<miopen::TensorDescriptor>& td,
+                                               std::vector<miopenTensorDescriptor_t>& ptd)
```
CAHEK7, line 1074: Isn't it better to preliminary reserve `td`/`ptd` to `bs.size()`?

CAHEK7 (dismissed):
The one serious question is about identical conditions in the `rnn_api`.
The others are mostly style fixed and are not so critical, but it would be nice to address them too.

src/rnn.cpp
shurale-nkn: this is not a direct GeterFunction. This should take calss field Seq(perBatch) and convert it to BatchesPerSequence

junliume (approved):

PR #2347, opened 2023-08-31
cmake: unpack kernels into a build directory

apwojcik:
Unpacking kernels (and other db files) into a source directory is causing read-write file access issues for multi-configuration generators. One of the examples is  Visual Studio solution files, where each build type's configuration process is performed simultaneously. That results in errors such as "cannot open a file used by another process" - file locked.


CMakeLists.txt
```diff
@@ -434,92 +434,69 @@ elseif(MIOPEN_BACKEND STREQUAL "OpenCL")
...
-            execute_process(
-                COMMAND touch ${db_file}
-                WORKING_DIRECTORY ${KERNELS_DIR}
-                )
-        endif()
-    endif()
+    get_filename_component(__fname ${db_bzip2_file} NAME_WLE)
+    add_custom_command(OUTPUT ${KERNELS_BINARY_DIR}/${__fname}
+                       COMMAND ${UNZIPPER} -k ${db_bzip2_file} > ${KERNELS_BINARY_DIR}/${__fname})
+    string(REPLACE "." "_" __tname ${__fname})
```
JehandadKhan, line 490: If we unzip in the build directory we should unzip in the same relative path to the `so` file so that MIOpen can find the file at runtime.

JehandadKhan (dismissed):

CMakeLists.txt
JehandadKhan: i.e. extract to `<build_dir>/share/miopen/db/`

apwojcik: Done. I modified `KERNELS_BINARY_DIR`.

CMakeLists.txt
```diff
@@ -441,92 +442,75 @@ elseif(MIOPEN_BACKEND STREQUAL "OpenCL")
     set(CPACK_RPM_PACKAGE_CONFLICTS miopen-hip)
 endif()
 
-# Unpack DB files
-function(unpack_db db_bzip2_file)
-    set(KERNELS_DIR "${CMAKE_SOURCE_DIR}/src/kernels")
-    STRING(REPLACE ".bz2" "" db_file "${db_bzip2_file}")
-    find_program(UNZIPPER NAMES lbunzip2 bunzip2)
-    if(EXISTS "${db_file}")
-        message(STATUS "WARNING: ${db_file} already exists and will be overwritten")
```
atamazov, line 475: [Q] @JehandadKhan Do we no longer need this warning?

apwojcik: The warnings are unnecessary - the unzipping has changed, and now it is CMake responsibility to regenerate (override) a file if a source has changed. The files are unpacked into the build directory instead of a source directory (as it was previously). I explained in the PR description why.

CMakeLists.txt
```diff
@@ -443,92 +444,75 @@ elseif(MIOPEN_BACKEND STREQUAL "OpenCL")
...
-endforeach()
+set(KDB_BZ2_FILES gfx90a.kdb.bz2 gfx1030.kdb.bz2 gfx908.kdb.bz2 gfx906.kdb.bz2 gfx900.kdb.bz2)
+set(COMPONENT_LST gfx90akdb gfx1030kdb gfx908kdb gfx906kdb gfx900kdb)
 
-if(LFS_MISSING_FILES)
-    string(REPLACE ";" "; " LFS_MISSING_FILES "${LFS_MISSING_FILES}")
-    message(WARNING "GIT LFS Files not pulled down, skipped: ${LFS_MISSING_FILES}")
-    set(MIOPEN_NO_LFS_PULLED TRUE CACHE INTERNAL "")
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.17)
+    foreach(__file __component IN ZIP_LISTS KDB_BZ2_FILES COMPONENT_LST)
```
junliume, line 528: @apwojcik This part of code needs to be updated. In the case where user has not pulled down all the LFS files:
previous:
```
CMake Warning at CMakeLists.txt:543 (message):
  GIT LFS Files not pulled down, skipped: gfx90a.kdb.bz2; gfx1030.kdb.bz2;
  gfx908.kdb.bz2; gfx906.kdb.bz2; gfx900.kdb.bz2
```
Now with this change it emits error for each file which is clogging the console output:
```
Make Warning at CMakeLists.txt:501 (message):
  GIT LFS file not pulled down, skipping: gfx90a.kdb
  CMakeLists.txt:520 (install_kdb)


CMake Warning at CMakeLists.txt:501 (message):
  GIT LFS file not pulled down, skipping: gfx1030.kdb
  CMakeLists.txt:520 (install_kdb)
```

CMakeLists.txt
```diff
@@ -408,8 +408,9 @@ endif()
 set(MIOPEN_GPU_SYNC Off CACHE BOOL "")
 if(BUILD_DEV)
     set(MIOPEN_BUILD_DEV 1)
-    set(MIOPEN_SYSTEM_DB_PATH "${CMAKE_SOURCE_DIR}/src/kernels" CACHE PATH "Default path of system db files")
-    set(MIOPEN_USER_DB_PATH "${CMAKE_SOURCE_DIR}/src/kernels" CACHE PATH "Default path of user db files")
+    set(KERNELS_BINARY_DIR ${PROJECT_BINARY_DIR}/share/miopen/db)
+    set(MIOPEN_SYSTEM_DB_PATH "${KERNELS_BINARY_DIR}" CACHE PATH "Default path of system db files")
```
junliume, line 425: @apwojcik this seems to have messed with `GetSystemDbPath()` at
https://github.com/ROCm/MIOpen/blob/63e5ff36e6e0eb86254f6e834887d47027eb9a8f/src/conv/heuristics/ai_heuristics.cpp#L67
which is defined here:
https://github.com/ROCm/MIOpen/blob/63e5ff36e6e0eb86254f6e834887d47027eb9a8f/src/db_path.cpp.in#L66-L85

The error:
```
src/conv/heuristics/ai_heuristics.cpp:39: Unable to load file: /home/jenkins/workspace/MLLIBS_MIOpen_unpacking_kernels/build/share/miopen/db/gfx90a_metadata.tn.model
```

CAHEK7:
It looks like ai_heuristic test is failed because of missing model:
```
[2023-12-22T14:33:11.786Z] build/bin/test_find_2_conv --float --all --limit 2 --disable-verification-cache 
[2023-12-22T14:33:11.786Z] PRNG seed: 12345678
[2023-12-22T14:33:11.786Z] Testing miopenFindSolutions...
[2023-12-22T14:33:11.786Z] MIOpen Error: src/conv/heuristics/ai_heuristics.cpp:39: Unable to load file: build/share/miopen/db/gfx90a_metadata.tn.model
[2023-12-22T14:33:11.786Z] CMake Error at test_test_find_2_conv.cmake:7 (message):
[2023-12-22T14:33:11.786Z]   Test failed
```
It relies on SYSTEM_DB_PATH, but do not install models properly.
@Dmantri98 @msaudulhassan could you help?

junliume:
@apwojcik could you update this PR based on review opinions above? Thanks!

apwojcik:
@junliume, I have updated it according to the suggestions. The models are not missing anymore. Do you have some other options that I still need to look into? The last comment I see is from @CAHEK7. That is the comment I followed to fix the PR. The previous comments I have incorporated the feedback from them.

junliume (approved):

junliume:
> @junliume, I have updated it according to the suggestions. The models are not missing anymore. Do you have some other options that I still need to look into? The last comment I see is from @CAHEK7. That is the comment I followed to fix the PR. The previous comments I have incorporated the feedback from them.

Hi @apwojcik could you also fix the issue mentioned here: https://github.com/ROCm/MIOpen/pull/2347#discussion_r1435993373 ? For users who does not `git-lfs pull` all large files, the cmake output warning message for each of the missing files.

CMakeLists.txt
JehandadKhan: @atamazov Should I mark the comment resolved?

atamazov: @JehandadKhan @apwojcik Sure [resolved]

CMakeLists.txt
apwojcik: Done

apwojcik:
> 
> Hi @apwojcik could you also fix the issue mentioned here: [#2347 (comment)](https://github.com/ROCm/MIOpen/pull/2347#discussion_r1435993373) ? For users who does not `git-lfs pull` all large files, the cmake output warning message for each of the missing files.

Done

johnnynunez:
Nice @apwojcik 🎉

CMakeLists.txt
```diff
@@ -399,19 +399,31 @@ option( MIOPEN_DEBUG_FIND_DB_CACHING "Use system find-db caching" ON)
...
 option(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY "Build with file/folder reorg with backward compatibility enabled" OFF)
 
-set( MIOPEN_INSTALL_DIR miopen)
-set( DATA_INSTALL_DIR ${CMAKE_INSTALL_DATAROOTDIR}/miopen )
+if(WIN32)
+    set( DATA_INSTALL_DIR bin )
+    set( DATABASE_INSTALL_DIR ${DATA_INSTALL_DIR})
+else()
+    set( MIOPEN_INSTALL_DIR miopen)
+    set( DATA_INSTALL_DIR ${CMAKE_INSTALL_DATADIR}/${MIOPEN_INSTALL_DIR} )
```
atamazov, line 407: Is it correct to change from `CMAKE_INSTALL_DATAROOTDIR` to `CMAKE_INSTALL_DATADIR` here?

apwojcik: Yes, according to CMake documentation of GNUInstallDir
`DATADIR` inherits the default from `DATAROOTDIR` which is `share` on UN*X systems.
https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

apwojcik: Both are read-only values, so we cannot change them in the CMakeLists.txt files.

atamazov: @apwojcik Okay, thanks.

apwojcik: Sorry, my bad. The variables are supposed to hold the names of the directories for read-only files. They can be changed in CMakeLists.txt. The rest is proper. Both variables behave according to the GNU Coding Standard - https://www.gnu.org/prep/standards/html_node/Directory-Variables.html.

PR #2348, opened 2023-08-31
cmake: generate export header function

apwojcik:
The threats model on Windows for UIF requires exporting only public API from dynamically linked libraries. Using `EXPORT_ALL_SYMBOLS`, CMake will export all public symbols from compilation units, including API, which is the threat model violation.

Moreover, the number of symbols allowed for exporting per single DLL on Windows is limited to 64K. Again, using the `EXPORT_ALL_SYMBOLS` property, we lose control over what and how it is exported.

The function `miopen_generate_export_header()` follows the agreed convention with MIGraphX and can be migrated into `rocm_generate_export_header()` and, in the future, be part of the `rocm-cmake` project.


averinevg (commented):
Wouldn't it be easier to just uncomment [this line](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/CMakeLists.txt#L813)?

src/tensor.cpp
```diff
@@ -501,7 +501,6 @@ void from_json(const nlohmann::json& j, TensorDescriptor& descriptor)
 } // namespace miopen
 
 // TODO(paul): Remove
```
JehandadKhan: The comment can go as well!

JehandadKhan (dismissed):
Minor nitpick.

junliume (approved):

PR #2349, opened 2023-08-31
cmake: cleanup outdated code

apwojcik:


junliume (approved):
LGTM but @atamazov could you give it a quick look?

src/CMakeLists.txt
```diff
@@ -667,23 +661,13 @@ clang_tidy_check(MIOpen)
...
-    set(PASS_ARGS debug optimized)
-    set(DEPS)
-    foreach(DEP ${ARGN})
-        if(DEP IN_LIST PASS_ARGS)
-            list(APPEND DEPS ${DEP})
-        else()
-            list(APPEND DEPS $<BUILD_INTERFACE:${DEP}>)
-        endif()
-    endforeach()
-    target_link_libraries(${TARGET} INTERFACE ${DEPS})
```
atamazov, line 704: @pfultz2 Do we no longer need this patch for "debug" and "optimized"? I am not quite confident with this stuff.

atamazov (commented):
I have no objections, but one mandatory question -- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2349/files#r1320859644

src/CMakeLists.txt
junliume: Ping @pfultz2 and @JehandadKhan

junliume: Ping @pfultz2 and @JehandadKhan for review

pfultz2: No, it isnt used anymore. I think this was used before we could use boost cmake targets.

pfultz2 (approved):

PR #2350, opened 2023-08-31
cmake: fix cmake/googletest.cmake on Windows

apwojcik:
Windows file system is case insensitive. Both file names `GoogleTest.cmake` and `googletest.cmake` refer to the same file according to the file system.


test/gtest/CMakeLists.txt
```diff
@@ -29,24 +23,25 @@ function(add_gtest TEST_NAME)
...
+add_gtest(conv_api)
+add_gtest(conv_api_strided_tensors)
+add_gtest(log_test)
+add_gtest(log_test_neg)
+add_gtest(na_infer)
+add_gtest(solver_convasm3x3u)
+
+if(NOT MIOPEN_BACKEND_OPENCL)
+  add_gtest(dumpTensorTest)
+endif()
```
JehandadKhan: I don't follow why the loop was removed, do you suggest for us to add each test explicitly ?

JehandadKhan (dismissed):

test/gtest/CMakeLists.txt
apwojcik: Yes. The `file(GLOBE...)` command was designed for CMake in _script mode_ (`*.cmake` files). Please read more about it here: https://cmake.org/cmake/help/latest/manual/cmake-language.7.html. The author's intention was not to use it in _build system generation mode_ - in this case. Please see the note for `GLOBE` here: https://cmake.org/cmake/help/latest/command/file.html#glob.
I will revert this change if you insist on returning the `file(GLOB...)` and the loop.

apwojcik: CMake scrip mode is not equivalent to `include(...)` files - that still counts as build system generation mode even when `.cmake` is used. CMake script mode is invoked with `cmake -P <script.cmake>` and does not configure anything and generate a build system.

averinevg: @apwojcik This is not a big problem in this case. I think these changes should be reverted

junliume: @apwojcik could you follow up with the review opinion above? Thanks!

apwojcik: done

junliume (approved):

PR #2351, opened 2023-08-31
cmake: make building tests optional

apwojcik:
The module `CTest` automatically creates the `BUILD_TESTING` option that we use to select whether to enable testing support. The option is `ON` by default. Refer to https://cmake.org/cmake/help/latest/module/CTest.html for details.

Moreover, explicitly calling `enable_testing()` is unnecessary because the command is automatically invoked when the `CTest` module is included. Refer to https://cmake.org/cmake/help/latest/command/enable_testing.html for details.


atamazov:
@apwojcik So this is a kind of refactoring that should not lead to any functional changes?

apwojcik:
> @apwojcik So this is a kind of refactoring that should not lead to any functional changes?

Correct. The only change is when PR is merged, one may decide not to build tests while building MIOpen. For instance, that is when we build MIOpen as a dependency for MIGraphX in Windows CI.

JehandadKhan (approved):

junliume (approved):

junliume (dismissed):
```
CMake Error at test/gtest/CMakeLists.txt:1 (include):

  include could not find requested file:



    googletest
```

junliume (approved):

PR #2354, opened 2023-08-31
Fix parameter truncation for CK solvers

Dmantri98:
When running tuning on CK solvers their respective parameter strings were being truncated. This is due to the fact CK stores their kernel parameters in string format which is different from other MIOpen solvers, which store the parameters in PerfConfig structs. 
This means that how serialization/deserialization is done for non-ck solvers does not properly extend to CK solver's. For non-ck solvers a deserialized string such as "8,9,10,2" of parameters would be read into a struct, where each digit would be parsed into it's own respective integer, that is a member of said struct. Whereas for CK solvers a parameter string such as "8,9,10,2" would be stored as string member of the PerfConfig struct and would need to have slightly differing implementations of serialized/deserialized (which in this case was solved by just templating) since they would not use comma's as a delimiter.

### DIFF
--- a/src/include/miopen/performance_config.hpp
+++ b/src/include/miopen/performance_config.hpp
@@ -68,6 +68,20 @@ struct PerfConfigBase : PerfConfig
     }
 };
 
+template <class Derived>
+struct PerfConfigBaseCK : PerfConfig
+{
+    void Serialize(std::ostream& stream) const final
+    {
+        serialize::SerDes<'\0'>::Serialize(static_cast<const Derived&>(*this), stream);
+    }
+
+    bool Deserialize(const std::string& s) final
+    {
+        return serialize::SerDes<'\0'>::Deserialize(static_cast<Derived&>(*this), s);
+    }
+};
+
 } // namespace solver
 } // namespace miopen
 
--- a/src/include/miopen/performance_config.hpp
+++ b/src/include/miopen/performance_config.hpp
@@ -73,12 +73,18 @@ struct PerfConfigBaseCK : PerfConfig
 {
     void Serialize(std::ostream& stream) const final
     {
-        serialize::SerDes<'\0'>::Serialize(static_cast<const Derived&>(*this), stream);
+        const Derived& self = static_cast<const Derived&>(*this);
+        stream.write(self.kernel_id.c_str(), self.kernel_id.length());
     }
 
     bool Deserialize(const std::string& s) final
     {
-        return serialize::SerDes<'\0'>::Deserialize(static_cast<Derived&>(*this), s);
+        Derived& self = static_cast<Derived&>(*this);
+        std::stringstream ss;
+        ss.str(s);
+        if(!std::getline(ss, self.kernel_id))
+            return false;
+        return true;
     }
 };
 
--- a/src/include/miopen/performance_config.hpp
+++ b/src/include/miopen/performance_config.hpp
@@ -79,11 +79,8 @@ struct PerfConfigBaseCK : PerfConfig
 
     bool Deserialize(const std::string& s) final
     {
-        Derived& self = static_cast<Derived&>(*this);
-        std::stringstream ss;
-        ss.str(s);
-        if(!std::getline(ss, self.kernel_id))
-            return false;
+        Derived& self  = static_cast<Derived&>(*this);
+        self.kernel_id = s;
         return true;
     }
 };
--- a/src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
@@ -160,7 +160,9 @@ void PerformanceConfigHipImplicitGemm3DGroupFwdXdlops::Init(const ProblemDescrip
                                                               {});
         if(conv_ptrs[i]->IsSupportedArgument(argument_ptr.get()))
         {
-            valid_kernels.push_back(conv_ptrs[i]->GetTypeString());
+            std::string temp = conv_ptrs[i]->GetTypeString();
+            temp.erase(std::remove(temp.begin(), temp.end(), ' '), temp.end());
+            valid_kernels.push_back(temp);
         }
     }
     assert(!valid_kernels.empty());
--- a/src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
@@ -160,9 +160,7 @@ void PerformanceConfigHipImplicitGemm3DGroupFwdXdlops::Init(const ProblemDescrip
                                                               {});
         if(conv_ptrs[i]->IsSupportedArgument(argument_ptr.get()))
         {
-            std::string temp = conv_ptrs[i]->GetTypeString();
-            temp.erase(std::remove(temp.begin(), temp.end(), ' '), temp.end());
-            valid_kernels.push_back(temp);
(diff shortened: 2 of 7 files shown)


src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
```diff
@@ -160,7 +160,9 @@ void PerformanceConfigHipImplicitGemm3DGroupFwdXdlops::Init(const ProblemDescrip
                                                               {});
         if(conv_ptrs[i]->IsSupportedArgument(argument_ptr.get()))
         {
-            valid_kernels.push_back(conv_ptrs[i]->GetTypeString());
+            std::string temp = conv_ptrs[i]->GetTypeString();
+            temp.erase(std::remove(temp.begin(), temp.end(), ' '), temp.end());
+            valid_kernels.push_back(temp);
```
CAHEK7: Minor. Avoid extra copy.
```suggestion
            valid_kernels.push_back(std::move(temp));
```

CAHEK7 (dismissed):
I'm not requesting the changes, since I'm going to modify this code anyway and make it the same like in the fwd case.
It's better to merge it before my changes.

CAHEK7:
@Dmantri98 could you check what's the problem with CI? This PR will have merge conflict with my current task, and it would be nice to have this finalized at least in terms of CI and tests.

Dmantri98:
> @Dmantri98 could you check what's the problem with CI? This PR will have merge conflict with my current task, and it would be nice to have this finalized at least in terms of CI and tests.

@CAHEK7  
Jehandad was talking about how you were also getting similar issues when running tuning on the CK solvers. Can you point me to your PR that fixes these issues so I can take a look at them?

CAHEK7:
It's not a similar issue, it is refactoring of ck kernels. #2379

junliume:
@Dmantri98 could you resolve conflicts? Thanks!

Dmantri98:
@CAHEK7 
Instead of actually modifying the kernel_id string and taking out the spaces I have just modified the apply function, where the spaces were causing truncation in the CK configs since the ">>" operator stops reading at first whitespace character. I've changed it to where if result is of type string that all of string s will be read into result without stopping at any whitespace.

CAHEK7 (dismissed):
From my point of view it looks good, but I'm not very familiar with all the internals of config save\restore functionality and can miss something.
Probably @atamazov or @DrizztDoUrden have something to say?

atamazov:
@CAHEK7 @Dmantri98 @DrizztDoUrden Let's talk about this.

DrizztDoUrden (dismissed):

Dmantri98:
@atamazov I see Vasillii (@DrizztDoUrden) has just approved these changes, were there any other reservations or doubts that you had?

src/include/miopen/performance_config.hpp
```diff
@@ -68,6 +68,20 @@ struct PerfConfigBase : PerfConfig
     }
 };
 
+template <class Derived>
+struct PerfConfigBaseCK : PerfConfig
+{
+    void Serialize(std::ostream& stream) const final
+    {
+        serialize::SerDes<'\0'>::Serialize(static_cast<const Derived&>(*this), stream);
+    }
```
atamazov: Serialization and de-serialization of a single string should be trivial.

All the CK solvers use `kernel_id` of the `std::string` type to hold value. Let's serialize directly from (deserialize directly into) the `kernel_id` member of Derived. The implementation won't require any parsing or involving `operator>>` or `operator<<`.

Please also remove all `Visit` template methods from classes inherited from PerConfigBaseCK - they will become unused.

----
Alternative solution:

As you can see `SerDes` is intended for handling of CSV things (which may include strings but without spaces). You can create an alternative for `SerDes` -- say, `SerDesSingleString`, -- which should be intended for S/D of a single string. This would allow to use arbitrary names for the (de)serializable class members.

I prefer the 1st, simpler approach.

src/include/miopen/performance_config.hpp
```diff
@@ -68,6 +68,20 @@ struct PerfConfigBase : PerfConfig
...
+
+    bool Deserialize(const std::string& s) final
```
atamazov, line 80: See above

atamazov (commented):
🟡 This breaks support for (de)serialization of CSV stuff that contains strings (e.g. a mix of numbers with textual identifiers). However this feature is not used for now.

🟡 Another drawback is that the serialization of a single string looks too complex ;)

🟢 I am NOT against merging this as is if the patch is urgent, provided that the review comments will be resolved after the merge in a separate PR (in other words, the relevant ticket should be opened and assigned right after the merge).

src/include/miopen/performance_config.hpp
Dmantri98: The first solution makes the most sense to me, didn't realize that just eliminating the abstraction for deserialization/serialization would lead to such a simple solution. Thanks.

src/include/miopen/performance_config.hpp
```diff
@@ -68,6 +68,26 @@ struct PerfConfigBase : PerfConfig
...
+        stream.write(self.kernel_id.c_str(), self.kernel_id.length());
+    }
+
+    bool Deserialize(const std::string& s) final
+    {
+        Derived& self = static_cast<Derived&>(*this);
+        std::stringstream ss;
+        ss.str(s);
+        if(!std::getline(ss, self.kernel_id))
+            return false;
```
CAHEK7:
```suggestion
        self.kernel_id = s;
```

CAHEK7 (approved):

junliume (approved):

src/include/miopen/performance_config.hpp
```diff
@@ -68,6 +68,23 @@ struct PerfConfigBase : PerfConfig
     }
 };
 
+template <class Derived>
+struct PerfConfigBaseCK : PerfConfig
+{
+    void Serialize(std::ostream& stream) const final
+    {
+        const Derived& self = static_cast<const Derived&>(*this);
+        stream.write(self.kernel_id.c_str(), self.kernel_id.length());
```
atamazov, line 77: Looks like an overkill. I think this should work:
```C++
stream << static_cast<const Derived&>(*this).kernel_id;
```

src/include/miopen/performance_config.hpp
```diff
@@ -68,6 +68,23 @@ struct PerfConfigBase : PerfConfig
...
+    }
+
+    bool Deserialize(const std::string& s) final
+    {
+        Derived& self  = static_cast<Derived&>(*this);
+        self.kernel_id = s;
```
atamazov, line 83: [R] The `Derived& self...` stuff looks cryptic to me. I recommend more clear definition, like
```C++
auto instance = static_cast<Derived&>(*this)
instance.kernel_id = ...
```
or just inlining the cast.

atamazov (commented):
@Dmantri98 Let's polish the implementation.

src/include/miopen/performance_config.hpp
amberhassaan: I prefer the code the way it is when compared to using `auto` (as too much of auto hurts readability). 
In reality, this case does indicate a broken interface. I discussed with @CAHEK7 on how we should have this base class `PerfConfigBaseCK` and move the members like `kernel_id` up to this common base instead of have 4 or 5 derived classes, each with the same member `kernel_id`.

amberhassaan: So in short, I recommend moving `kernel_id` up to this base class. I or @CAHEK7 can take it up as he's already dealt with the derived classes recently.

CAHEK7: CK base class is a kind of tricky solution and it's still in progress.

PR #2355, opened 2023-08-31
cmake: use imported target for threads library instead of variable

apwojcik:
The imported target carries all the information required to link a target against the Threads library correctly. However, the variable `${CMAKE_THREAD_LIBS_INIT}` only provides the thread library. Since thread support is a Windows built-in feature (unlike Linux thread support, which is provided as a separate library), using an imported target is preferable.


JehandadKhan (approved):
This approach is certainly preferable.

junliume (approved):

PR #2357, opened 2023-08-31
cmake: a few fixes for multi-config generators

apwojcik:


JehandadKhan (approved):

junliume (approved):

cmake/EnableCompilerWarnings.cmake
```diff
@@ -26,89 +26,104 @@
...
-            -Wendif-labels
-            -Wformat
-            -Winit-self
-            -Wreturn-type
-            -Wsequence-point
-            # Shadow is broken on gcc when using lambdas
-            # -Wshadow
-            -Wswitch
-            -Wtrigraphs
-            -Wundef
```
atamazov, line 64: `-Wno-undef` must be reverted. We do not allow usage of undefined macros.

CMakeLists.txt
```diff
@@ -37,10 +37,20 @@ macro(set_var_to_condition var)
     endif()
 endmacro()
 
+get_property(_GENERATOR_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
```
atamazov: [R] Normally we use the "MIOPEN" prefix in such cases, don't we?
```suggestion
get_property(MIOPEN_GENERATOR_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
```

src/sqlite/CMakeLists.txt
```diff
@@ -29,13 +29,11 @@ add_library(
...
-# using Clang
-set_source_files_properties(memvfs.cpp PROPERTIES COMPILE_FLAGS "-Wno-everything ")
-elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-# using GCC
-set_source_files_properties(memvfs.cpp PROPERTIES COMPILE_FLAGS "-w ")
-endif()
+
+target_compile_options(sqlite_memvfs PRIVATE
+        $<$<CXX_COMPILER_ID:Clang>:-Wno-everything>
+        $<$<CXX_COMPILER_ID:GNU>:-w>)
```
atamazov, line 35: 👍

atamazov (changes requested):

cmake/EnableCompilerWarnings.cmake
junliume: @apwojcik could you follow up with this one? Currently, we only disable it for third party like `GTest`

CMakeLists.txt
junliume: @apwojcik if you agree we can make this change quickly :)

cmake/EnableCompilerWarnings.cmake
apwojcik: done

CMakeLists.txt
```diff
@@ -254,8 +264,7 @@ endif()
 
 message(STATUS "Hip compiler flags: ${HIP_COMPILER_FLAGS}")
 
-add_definitions("-DHIP_COMPILER_FLAGS=${HIP_COMPILER_FLAGS}")
-
+add_compile_definitions($<$<COMPILE_LANGUAGE:CXX>:HIP_COMPILER_FLAGS="${HIP_COMPILER_FLAGS}">)
```
atamazov, line 267: @apwojcik 🐛 These extra double ticks break HIP kernel builds on offline path (`-DMIOPEN_USE_COMGR=Off -DMIOPEN_USE_HIPRTC=Off`), and, possibly, on COMgr path (`-DMIOPEN_USE_HIPRTC=Off`), thus affecting Tuna. Fix TBD soon.

/cc @JehandadKhan @junliume

cmake/EnableCompilerWarnings.cmake
```diff
@@ -26,89 +26,104 @@
...
+
+set(__gnu_cxx_compile_options
+    -Wno-missing-field-initializers
+)
+
+add_compile_options(
+    "$<$<CXX_COMPILER_ID:MSVC>:${__msvc_cxx_compile_options}>"
+    "$<$<CXX_COMPILER_ID:Clang>:${__default_cxx_compile_options};${__clang_cxx_compile_options}>"
+    "$<$<CXX_COMPILER_ID:GNU>:${__default_cxx_compile_options};${__gnu_cxx_compile_options}>"
+)
```
atamazov, line 124: [Recommendation] @apwojcik This should work, but please double-check correctness of this.

atamazov (commented):
Breaks offline HIP builds.

CMakeLists.txt
atamazov: Fixed in #2544

PR #2363, opened 2023-08-31
half: fix compiling with HIP SDK 5.7+ on Windows

apwojcik:


CAHEK7 (dismissed):

CAHEK7:
LGTM besides this comment https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2364#discussion_r1316389831

apwojcik:
The case with comgr is different from here. We are not using `rbuild` for dependencies on Windows. Instead, we use CMake's native support, `FetchContent` and `ExternalProject` (for non-CMake projects). However, the `rbuild` tool may still be used to build the project and bare CMake - no need for additional steps in both cases. The upcoming PRs (not yet submitted) will explain why we chose that way.

JehandadKhan (approved):

JehandadKhan:
@apwojcik I would like for us to unify the build system, if CMake can support both platforms, then that would be my choice.

junliume (approved):

PR #2364, opened 2023-08-31
comgr: fix compiling with HIP SDK 5.5+ on Windows

apwojcik:


src/comgr.cpp
```diff
@@ -38,7 +38,7 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/stringutils.hpp>
 
-#if HIP_PACKAGE_VERSION_FLAT >= 5004000000ULL
+#if !defined(_WIN32) && (HIP_PACKAGE_VERSION_FLAT >= 5004000000ULL)
```
atamazov, line 41: Why in Windows the location of amd_comgr.h does not correspond to the one in Linux? Is it a bug?

atamazov (changes requested):

src/comgr.cpp
apwojcik: It is not a bug. We must build comgr from source code on Windows. We're talking with the respective people from HIP SDK not to build the comgr on our own. We prefer they provide the package instead. But that discussion has just started, and it is pending.

JehandadKhan (approved):

junliume (approved):

PR #2367, opened 2023-08-31
[HotFix] Fix issue with MIOPEN_THROW when expanded

junliume:
There is a problem with `MIOPEN_THROW` MACRO:

```
#define MIOPEN_THROW(__VA_ARGS__...) do { miopen::MIOpenThrow(__FILE__, __LINE__, __VA_ARGS__); } while(false);
```
Expands to:
```
do { miopen::MIOpenThrow("MIOpen/src/seq_tensor.cpp", 213, miopenStatusBadParm, "Lengths and padding number dimensions must be equal"); } while(false);
```
and when it is followed with `else` we will see the error:
```
MIOpen/src/seq_tensor.cpp:213:9: error: expected expression
        else
        ^
```

I am not sure how it get passed the CI but it is showing up on standalone desktops

### DIFF
--- a/src/seq_tensor.cpp
+++ b/src/seq_tensor.cpp
@@ -208,8 +208,10 @@ SeqTensorDescriptor::SeqTensorDescriptor(miopenDataType_t t,
     else
     {
         if(padding_in.size() != dims)
+        {
             MIOPEN_THROW(miopenStatusBadParm,
                          "Lengths and padding number dimensions must be equal");
+        }
         else
             padds = padding_in;
     }


muralinr (approved):
LGTM

junliume:
@atamazov @shurale-nkn @muralinr I think we were "lucky" none of the previous use cases met this scenario
```
if (...)
    MIOPEN_THROW();
else
    do_something
```
If one tried to wrap the MACRO with `{}` it won't help because it will expand to
```
if (...)
    {do {something;} while(false);};
else
    do_something
```
and the compiler is still complaining about
```
error: expected expression
```

atamazov:
@junliume The proper fix is removing ";" from the end of the definition of MIOPEN_THROW. The possibility of removal of the ending semicolon is the only reason why the `do...while(false)` wrapper is used. Can you please do that?

atamazov:
@junliume This is where the issue has been introduced: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2166/files#r1314316746

atamazov:
@junliume I recommend reverting this and merging a proper fix. I can create a PR quicky, if you wish, just let me know.

junliume:
> @junliume I recommend reverting this and merging a proper fix. I can create a PR quicky, if you wish, just let me know.

@atamazov yes it would be great to fix this issue if you could create a PR :) Thanks!

pfultz2:
Why not just remove the do-while loop? I dont think its necessary.

atamazov:
@pfultz2 Without it, the macro is expanded to a void expression that may be accidentally used in wrong contexts.

atamazov:
some superficial comments https://stackoverflow.com/questions/923822/whats-the-use-of-do-while0-when-we-define-a-macro

atamazov:
@junliume #2376 is ready))

PR #2369, opened 2023-09-04
Fix perf config search and serialization/deserialization

CAHEK7:
Related to PR #2338
Since `kernel_id` is more reliable and persistent entity for CK kernels, and it's exactly a value which can be serialized and deserialized, I've brough back the logic based on `kernel_id` instead of `index`.

@atamazov please take a look.

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -143,12 +143,11 @@ void PerformanceConfigHipImplicitGemmFwdXdlops::Init(const ProblemDescription& p
     for(size_t idx = 0; idx < conv_ptrs.size(); ++idx)
     {
         if(args.IsSupportedBy(conv_ptrs[idx]))
-            valid_kernels.emplace_back(std::move(static_cast<int>(idx)),
-                                       std::move(conv_ptrs[idx]->GetTypeString()));
+            valid_kernels.emplace_back(std::move(conv_ptrs[idx]->GetTypeString()));
     }
     assert(!valid_kernels.empty());
     index     = 0;
-    kernel_id = valid_kernels[0].second;
+    kernel_id = valid_kernels[0];
 }
 
 template <typename DataType>
@@ -237,7 +236,7 @@ bool PerformanceConfigHipImplicitGemmFwdXdlops::SetNextValue(const ProblemDescri
     if((index + 1) < valid_kernels.size())
     {
         ++index;
-        kernel_id = valid_kernels[index].second;
+        kernel_id = valid_kernels[index];
         return true;
     }
     else
--- a/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
@@ -93,16 +93,16 @@ struct CKArgs
         return conv_ptr->MakeArgumentPointer(in,
                                              w,
                                              out,
-                                             this->N,
-                                             this->K,
-                                             this->C,
-                                             this->input,
-                                             this->filter,
-                                             this->output,
-                                             this->strides,
-                                             this->dilation,
-                                             this->lPadding,
-                                             this->rPadding,
+                                             N,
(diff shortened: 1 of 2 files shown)


src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -156,8 +156,8 @@ bool PerformanceConfigHipImplicitGemmFwdXdlops::CheckIsSupportCKArgs(
 {
     const auto conv_ptrs = DeviceOpPtrs<DataType>::GetInstances();
     auto ptr_idx         = std::find_if(
-        conv_ptrs.begin(), conv_ptrs.end(), [&kernel_id = this->kernel_id](const auto& ptr) {
-            return ptr->GetTypeString() == kernel_id;
+        conv_ptrs.begin(), conv_ptrs.end(), [&kernl_id = this->kernel_id](const auto& ptr) {
```
atamazov: # 🤣

atamazov:
@CAHEK7 If this is a bug, then there should be a regression test included with the fix. Or maybe this is not a bug? ;)

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
CAHEK7: And it depends on installed ROCM, for ROCM 5.6 it does not trigger any warnings.

CAHEK7:
What is the best way to check that saved\restored perf config does the same thing?

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
amberhassaan: good use of `std::find_if`. Please see if you can repilcate this for other solvers, or better yet, consolidate it into some base-class that all such solvers can use. Thanks.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -352,13 +358,21 @@ ConvSolution ConvHipImplicitGemmFwdXdlops::GetSolution(
     return {};
 #else
 
-    auto InitInvokerFactory = [&problem, idx = config.index](auto DataTypeVal) {
+    auto InitInvokerFactory = [&problem, &kernel_id = config.kernel_id](auto DataTypeVal) {
         using DataType = decltype(DataTypeVal);
 
+        auto conv_ptrs = DeviceOpPtrs<DataType>::GetInstances();
+        auto ptr_idx =
+            std::find_if(conv_ptrs.begin(), conv_ptrs.end(), [&kernel_id](const auto& ptr) {
```
amberhassaan: this is also a reason we should consolidate this code in a common function such as `FindConvPtrByID`.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
```diff
@@ -352,13 +358,21 @@ ConvSolution ConvHipImplicitGemmFwdXdlops::GetSolution(
...
+                return ptr->GetTypeString() == kernel_id;
+            });
+
+        if(ptr_idx == conv_ptrs.end())
```
amberhassaan: [minor] [stylistic] : rename to `ptr_it` for `iterator` since that's what it.

src/solver/conv_hip_implicit_gemm_fwd_xdlops.cpp
CAHEK7: It may reduce the number of parameters passed in `find_if`, but `find_if` is the only peace of code which can be moved into separated function.

CAHEK7:
@amberhassaan can you take a look at the latest version?
@junliume this PR is a bit critical, since the current develop branch may have inconsistent perf tuning behavior for CK implicit gemm forward operation.

amberhassaan (approved):
I think this is good. For future work, we should promote `FindConvPtrByID` in a header (stand-alone function or a base-class method) that other `CK` solvers can call. Thanks.

junliume (approved):

PR #2370, opened 2023-09-04
FIX:single element copy and  workspace size bugs

shurale-nkn:
single element copy and RNN wrong workspace size bugs
Found by extra testing


junliume (approved):

junliume:
@shurale-nkn it would be great to add the test into CI suit, however considering the urgency of this patch I will merge it first.

PR #2376, opened 2023-09-05
[HOTFIX] Restore safe implementation of MIOPEN_THROW

atamazov:
Reverts #2367 and fixes issue introduced in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2166/files#r1314316746 that leads to https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2367#issue-1876440217 and possibly more (because currently each occurence of `MIOPEN_TROW(...);` expands to two statements instead of one).

----
- Proposed reviewers:
  - @muralinr 
  - @junliume
  - @shurale-nkn

### DIFF
--- a/src/seq_tensor.cpp
+++ b/src/seq_tensor.cpp
@@ -208,10 +208,8 @@ SeqTensorDescriptor::SeqTensorDescriptor(miopenDataType_t t,
     else
     {
         if(padding_in.size() != dims)
-        {
             MIOPEN_THROW(miopenStatusBadParm,
                          "Lengths and padding number dimensions must be equal");
-        }
         else
             padds = padding_in;
     }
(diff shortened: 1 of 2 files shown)


src/seq_tensor.cpp
```diff
@@ -208,10 +208,8 @@ SeqTensorDescriptor::SeqTensorDescriptor(miopenDataType_t t,
     else
     {
         if(padding_in.size() != dims)
-        {
```
junliume, line 211: @atamazov well it should never hurt to have an extra pair of {}  ;-P

atamazov: @junliume Yes except two things:
- These may give a false impression that we need braces because of the macro (which is not so)
- They waste screen space (which I personally value very high even with my 2560x1600 display).

But I do not insist on these things (all the more they eventually sneak into our code). Just let me know and I'll bring'em back.

junliume (approved):

PR #2379, opened 2023-09-06
Ck kernels invocation refactoring

### LINKED ISSUE #2332: Post merge fixes for #2286: Simplify Invoker lambda
Simplify the `Invoker` lambda (returned by `ConvHipImplicitGemm3DGroupFwdXdlops`) by moving as much of the computation  outside of the lambda as possible.

### LINKED ISSUE #2336: Cache CK convolution kernel objects as shared_ptr in Invoker lambda

CAHEK7:
Simplified and unified CK kernel invocation.
Applied patches from #2338, #2369, #2305, #2336 to all other CK kernels and unified the code.

@atamazov


CAHEK7:
@iq136boy can you take a look since it affects your PRs?
@amberhassaan that's exactly the change which brings similar CK invocation code into one place.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -0,0 +1,115 @@
...
+    if(ptr_iter == conv_ptrs.end())
+        MIOPEN_THROW("PerformanceConfig kernel '" + kernel_id + "' does not exist");
+
+    ConvSolution result;
+    result.invoker_factory =
+        [ck_args     = CKArgsType{problem},
+         sh_conv_ptr = std::shared_ptr{std::move(*ptr_iter)}](const std::vector<Kernel>&) mutable {
+            return [ck_args = std::move(ck_args), sh_conv_ptr = std::move(sh_conv_ptr)](
+                       const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+                const auto& data_ctx = primitive_parameters.CastTo<conv::DataInvokeParams>();
```
iq136boy: This DataInvokeParams only works for fwd and backward data. For backward weight, we need WrWInvokeParams. I would suggest to either template it or check the convolution direction to decide which Invokeparams to be called. Sth like:
if(problem.direction.IsBackwardWrW())
   data_ctx = primitive_parameters.CastTo<conv::WrWInvokeParams>();
else
   data_ctx = primitive_parameters.CastTo<conv::DataInvokeParams>();

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -0,0 +1,115 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#pragma once
+
+#include <miopen/conv/data_invoke_params.hpp>
```
iq136boy, line 29: Need add the include for wrw invoke params
#include <miopen/conv/wrw_invoke_params.hpp>

iq136boy (dismissed):

src/include/miopen/solver.hpp
```diff
@@ -4666,17 +4666,18 @@ struct ConvHipImplicitGemmFwdXdlops final
...
     PerformanceConfigHipImplicitGemmBwdXdlops(int idx, std::string kernl_id)
         : index(idx), kernel_id(kernl_id)
     {
     }
-    PerformanceConfigHipImplicitGemmBwdXdlops() : PerformanceConfigHipImplicitGemmBwdXdlops(0, "")
-    {
-    }
-    PerformanceConfigHipImplicitGemmBwdXdlops(bool)
+
+    PerformanceConfigHipImplicitGemmBwdXdlops() = default;
```
amberhassaan, line 4678: I think we can combine the `PerformanceConfigHipImplicit...` classes into one. This will allow some of the utility functions in `implicitgemm_ck_util.hpp` to be moved in base class.

CAHEK7:
CI is broken for a while. It can't even checkout the repo.

amberhassaan (approved):
Looks good. Nice consolidation and improvement. I'd recommend introducing a base performance config class for these implicit-gemm solvers, and that may remove some more repeated code.

src/include/miopen/solver.hpp
CAHEK7: I am thinking about it, but the main challenge is not to expose `CKArgs` and `DeviceOpPtrs` to the global scope, but have the base class fully specified.

Current solution is a kind of compromise.

iq136boy (approved):

JehandadKhan:
@junliume  This can be merged as well

junliume (approved):

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
```diff
@@ -459,25 +322,37 @@ bool ConvHipImplicitGemm3DGroupFwdXdlops::IsApplicable(const ConvolutionContext&
...
-    return {};
-#else
-    ConvSolution result;
-    result.invoker_factory = conv::MakeCK3DGroupFwdInvokerFactory(problem, config);
-    return result;
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    switch(problem.GetInDataType())
+    {
+    case miopenInt8:
+        return InitInvokerFactory<DeviceOpGFwdPtrs<int8_t>, CKArgs, conv::DataInvokeParams>(
```
atamazov, line 338: [Q] Maybe still `MakeInvokerFactory`? It not only initializes the object, but also creates it, right?

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
```diff
@@ -459,25 +322,37 @@ bool ConvHipImplicitGemm3DGroupFwdXdlops::IsApplicable(const ConvolutionContext&
...
+    case miopenFloat:
+        return InitInvokerFactory<DeviceOpGFwdPtrs<float>, CKArgs, conv::DataInvokeParams>(
+            problem, config.kernel_id);
+    case miopenInt32:
+    case miopenInt8x4:
+    case miopenBFloat16:
+    case miopenDouble:
+    default:
+        MIOPEN_THROW(miopenStatusInternalError,
+                     "ConvHipImplicitGemmFwdXdlops operation not implemented for this data type");
```
atamazov, line 352: ConvHipImplicitGemm3DGroupFwdXdlops

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
```diff
@@ -459,25 +322,37 @@ bool ConvHipImplicitGemm3DGroupFwdXdlops::IsApplicable(const ConvolutionContext&
...
+    }
 #endif
+    return {};
```
atamazov, line 355:
```C++
return {miopenStatusNotInitialized};
```

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+template <typename DeviceOpType, typename CKArgsType, typename CastType>
+ConvSolution InitInvokerFactory(const ProblemDescription& problem, const std::string& kernel_id)
+{
+    auto conv_ptrs = DeviceOpType::GetInstances();
+    auto ptr_iter  = FindConvPtrByID(conv_ptrs, kernel_id);
+
+    if(ptr_iter == conv_ptrs.end())
+        MIOPEN_THROW("PerformanceConfig kernel '" + kernel_id + "' does not exist");
```
atamazov, line 90: It is much better to return ConvSolution(miopenStatusInvalidValue) when this happens. Why: This won't break Find operation, i.e. would improve stability of releases at the users' sites. In all other cases (i.e. attempt to run a convolution) returning a solution with error code would lead to MIOPEN_THROW somewhere in the upper layers of the library.

Implementation: Maybe the most robust approach is throwing a custom exception and catching it in GetSolution (and then MIOPEN_LOG_E there). You can find example in comgr.cpp (which handles compilation errors).

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -91,171 +133,37 @@ struct CKArgsBwd
...
-            break;
-        }
-    }
-    assert(i != conv_ptrs.size());
-    auto& conv_ptr      = conv_ptrs.at(i);
-    auto& data_ctx      = primitive_parameters.CastTo<conv::DataInvokeParams>();
-    const auto& tensors = data_ctx.tensors;
-    auto argument_ptr   = conv_ptr->MakeArgumentPointer(
-        static_cast<void*>(tensors.out),
-        const_cast<void*>( // NOLINT (cppcoreguidelines-pro-type-const-cast)
```
atamazov, line 222: [Notice] No more `const_cast`. Good! 👍

src/include/miopen/solver.hpp
```diff
@@ -4666,17 +4666,18 @@ struct ConvHipImplicitGemmFwdXdlops final
 struct PerformanceConfigHipImplicitGemmBwdXdlops
     : PerfConfigBase<PerformanceConfigHipImplicitGemmBwdXdlops>
 {
-    int index;
-    std::string kernel_id;
+    int index             = 0;
+    std::string kernel_id = "";
     std::vector<std::string> valid_kernels;
+
```
atamazov, line 4672: [Q] AFAICS we use `kernel_id` as PerformanceConfig value, and we still need `index` and `valid_kernels` for traversing the ComputedContainer. Maybe we can get rid of `kernel_id` and store only `index` in the database? The positive side effect is smalled perf-db footprint.

Or this would make PerformanceConfig incompatible between CK versions because `valid_kernels` may become different?

atamazov (commented):
Part 1

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
atamazov: [resolved]

atamazov:
@CAHEK7 What are your plans about the unresolved post-merge review comments?

CAHEK7:
Thanks for mentioning me, I missed all the comments.

I'm going to address all of them when #2385 and #2386 are merged.

src/include/miopen/solver.hpp
CAHEK7: I think we can't since we do not have any guarantees that the order of the kernels is persistent. And it may break backward compatibility as well.

src/include/miopen/solver/implicitgemm_ck_util.hpp
CAHEK7: I've just kept initial functionality as it is. But I agree with this comment and I guess it should be a separated issue - how to handle bad perf configs gracefully. Issue #2426 created.

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
CAHEK7: Will wait for https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2385 and https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2386.

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
CAHEK7: Good point. But who has made `miopenStatusSuccess` for `ConvSolution` by default? (mostly rhetorical question, it doesn't matter who, but it must be fixed).

`ConvSolution(miopenStatus_t status_ = miopenStatusSuccess)`

Issue #2427 created.

src/include/miopen/solver.hpp
atamazov: [Recommendation]
@CAHEK7 Please put C++ comment that summarizes this review thread, i.e. explains why we need string `kernel_id` as well as `index` and `valid_kernels` in the CK's PerformanceConfigs.

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
CAHEK7: I'll keep this code since it is unified for all CK, MLIR and GEMM based kernel and it must be fixed as a part of #2427.
Unless #2427 is fixed this code will be there.

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
atamazov: [resolved]

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
atamazov: [resolved]

PR #2384, opened 2023-09-08
Add dnorm test convolution driver code

muralinr:
Adding dnorm test convolution driver code

### DIFF
--- a/driver/random.hpp
+++ b/driver/random.hpp
@@ -34,53 +34,25 @@ inline T RAN_GEN(T A, T B)
     return r;
 }
 
-template <typename T>
-inline T RAN_SUBNORM(int fraction_msb_bit = 0, int fraction_lsb_bit = 0)
+template <typename T, bool Signed = false>
+inline T RAN_SUBNORM()
 {
-    (void)fraction_msb_bit;
-    (void)fraction_lsb_bit;
-    return static_cast<T>(0);
+    T denorm_val = static_cast<T>(0);
+    if constexpr(std::is_same_v<T, float> || std::is_same_v<T, half_float::half>)
+    {
+        using BitType = std::conditional_t<sizeof(T) == 2, uint16_t, uint32_t>;
+        static_assert(sizeof(T) == sizeof(BitType));
+
+        // -1 because ::digits counts the first implicit digit
+        static constexpr auto mantissa_bits = std::numeric_limits<T>::digits - 1;
+        static constexpr auto mantissa_mask = (1 << mantissa_bits) - 1;
+
+        BitType denorm_bits = GET_RAND() & mantissa_mask;
+        denorm_bits |= Signed ? ((GET_RAND() % 2) << (sizeof(T) * 8 - 1)) : 0;
+
+        // the proper way to do a type punning
+        std::memcpy(&denorm_val, &denorm_bits, sizeof(T));
+    }
+    return denorm_val;
 }
-
-#define SUBNORM_INITIALIZE_FRACTION(msb, lsb, msb_default, lsb_default) \
-    do                                                                  \
-    {                                                                   \
-        if((msb) == 0 && (lsb) == 0)                                    \
-        {                                                               \
-            (msb) = (msb_default);                                      \
-            (lsb) = (lsb_default);                                      \
-        }                                                               \
-        assert((msb) <= (msb_default) && (msb) >= (lsb_default));       \
-        assert((lsb) <= (msb_default) && (lsb) >= (lsb_default));       \
-        assert((lsb) <= (msb));                                         \
-    } while(0)
-
-#define SUBNORM_RANDOM_FRACTION(result, msb, lsb)            \
-    do                                                       \
-    {                                                        \
-        int _bits = (msb) - (lsb);                           \
-        if(_bits == 0)                                       \
-            (result) = 0;                                    \
-        else                                                 \
-            (result) = (GET_RAND() % (1 << _bits)) << (lsb); \
-    } while(0)
-
-template <>
-inline float RAN_SUBNORM(int fraction_msb_bit, int fraction_lsb_bit)
-{
-    uint32_t float32_value;
-    SUBNORM_INITIALIZE_FRACTION(fraction_msb_bit, fraction_lsb_bit, 22, 0);
-    SUBNORM_RANDOM_FRACTION(float32_value, fraction_msb_bit, fraction_lsb_bit);
-    return *(reinterpret_cast<float*>(&float32_value));
-}
-
-template <>
-inline half_float::half RAN_SUBNORM(int fraction_msb_bit, int fraction_lsb_bit)
-{
-    uint16_t float16_value;
-    SUBNORM_INITIALIZE_FRACTION(fraction_msb_bit, fraction_lsb_bit, 9, 0);
-    SUBNORM_RANDOM_FRACTION(float16_value, fraction_msb_bit, fraction_lsb_bit);
-    return *(reinterpret_cast<half_float::half*>(&float16_value));
-}
-
 #endif // GUARD_RANDOM_GEN_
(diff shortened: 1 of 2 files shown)


driver/random.hpp
```diff
@@ -29,4 +38,53 @@ inline T RAN_GEN(T A, T B)
...
+
+#define SUBNORM_RANDOM_FRACTION(result, msb, lsb)        \
+    do                                                   \
+    {                                                    \
+        int _bits = (msb) - (lsb);                       \
+        if(_bits == 0)                                   \
+            (result) = 0;                                \
+        else                                             \
+            (result) = (rand() % (1 << _bits)) << (lsb); \
+    } while(0)
```
CAHEK7: I don't think `defines` is a good idea.
Can you consider to use normal functions\templates and probably `std::bitset` for bitwise operations?

driver/random.hpp
```diff
@@ -29,4 +38,53 @@ inline T RAN_GEN(T A, T B)
...
+
+template <>
+inline float RAN_SUBNORM(int fraction_msb_bit, int fraction_lsb_bit)
+{
+    uint32_t float32_value;
+    SUBNORM_INITIALIZE_FRACTION(fraction_msb_bit, fraction_lsb_bit, 22, 0);
+    SUBNORM_RANDOM_FRACTION(float32_value, fraction_msb_bit, fraction_lsb_bit);
+    return *(reinterpret_cast<float*>(&float32_value));
```
CAHEK7: I know that's a common way to do suck trick, but the true way is `std::memcpy`.
It's an intrinsic and does not contradict the standard.
C++20 has `std::bit_cast` for it.

driver/random.hpp
```diff
@@ -29,4 +38,53 @@ inline T RAN_GEN(T A, T B)
     return r;
 }
 
+template <typename T>
+inline T RAN_SUBNORM(int fraction_msb_bit = 0, int fraction_lsb_bit = 0)
+{
+    (void)fraction_msb_bit;
+    (void)fraction_lsb_bit;
```
CAHEK7: I'd rather be using `[[maybe_unused]]` attribute, but it has its own drawbacks.

driver/random.hpp
```diff
@@ -29,4 +38,53 @@ inline T RAN_GEN(T A, T B)
...
+            (result) = 0;                                \
+        else                                             \
+            (result) = (rand() % (1 << _bits)) << (lsb); \
+    } while(0)
+
+template <>
+inline float RAN_SUBNORM(int fraction_msb_bit, int fraction_lsb_bit)
+{
+    uint32_t float32_value;
+    SUBNORM_INITIALIZE_FRACTION(fraction_msb_bit, fraction_lsb_bit, 22, 0);
```
CAHEK7: I'm not quite sure that we need `SUBNORM_INITIALIZE_FRACTION` at all, especially as a define.

driver/random.hpp
```diff
@@ -29,4 +38,53 @@ inline T RAN_GEN(T A, T B)
     return r;
 }
 
+template <typename T>
+inline T RAN_SUBNORM(int fraction_msb_bit = 0, int fraction_lsb_bit = 0)
```
CAHEK7: Defaulted parameters may lead to misuse of the function.
Integer `T` must be forbidden.
I'd moved `fraction_msb_bit` and `fraction_lsb_bit` into template parameters or even removed them and made that bit ranges depended on the types.

driver/random.hpp
```diff
@@ -2,6 +2,15 @@
 #define GUARD_RANDOM_GEN_
 
 #include <random>
+#include <cstdlib>
+#include <cassert>
+#include <cstdint>
```
CAHEK7: Is it required for half?

driver/random.hpp
```diff
@@ -2,6 +2,15 @@
...
+#if HIP_PACKAGE_VERSION_FLAT >= 5006000000ULL
+#include <half/half.hpp>
+#else
+#include <half.hpp>
+#endif
+using float16 = half_float::half;
```
CAHEK7: It leads to leak `float16` type into the global scope.

driver/random.hpp
```diff
@@ -2,6 +2,15 @@
...
 
 #include <random>
+#include <cstdlib>
+#include <cassert>
+#include <cstdint>
+#if HIP_PACKAGE_VERSION_FLAT >= 5006000000ULL
+#include <half/half.hpp>
+#else
+#include <half.hpp>
+#endif
```
CAHEK7, line 9: Is it possible to avoid such includes in this header?

driver/random.hpp
```diff
@@ -29,4 +38,53 @@ inline T RAN_GEN(T A, T B)
...
+    } while(0)
+
+#define SUBNORM_RANDOM_FRACTION(result, msb, lsb)        \
+    do                                                   \
+    {                                                    \
+        int _bits = (msb) - (lsb);                       \
+        if(_bits == 0)                                   \
+            (result) = 0;                                \
+        else                                             \
+            (result) = (rand() % (1 << _bits)) << (lsb); \
```
CAHEK7: Probably `rand` shouldn't be used here since we've got ridded of it recently. It's slow and not thread safe.
Take a look into this file for the proper random number generator.

CAHEK7 (changes requested):

junliume:
@muralinr could you follow up with the review opinions? Most of them seem to be relatively easy to patch. Thanks!

muralinr:
> @muralinr could you follow up with the review opinions? Most of them seem to be relatively easy to patch. Thanks!

Yes, I discussed with @CAHEK7. I am making code changes and having issues with my github commit currently.  I will push these changes later today evening or by tomorrow morning.

muralinr:
@CAHEK7 I made code changes as discussed with you. You can enhance it later by convert it to templates, define conversions , and other like using library functions enhancements

CAHEK7:
> @CAHEK7 I made code changes as discussed with you. You can enhance it later by convert it to templates, define conversions , and other like using library functions enhancements

Ok, but I don't like the practice to make not yet done solutions and let the others fix it right away after the merge.
The main reason that I agree is I'm working on overall improvements of random number generation facilities and infrastructure, I'm in a context of this task and most probably I'll change this code as well.
But in common case I'd preferred to have everything fixed prior merge. Otherwise it requires twice more resources and twice more time of twice more developers.

N.B. there are some merge conflicts to be resolved.

JehandadKhan:
@muralinr  I would assume that we add this as a separate test so that it runs in our CI, adding it to the driver would not ensure this verification in MIOpen dev cycle.

muralinr:
> @muralinr I would assume that we add this as a separate test so that it runs in our CI, adding it to the driver would not ensure this verification in MIOpen dev cycle.

This is one time test to test hardware DENORM functionality.  We don't need to run this test in our CI testing. This test is used only during bring up time once for testing DENORM h/w functionality.

CAHEK7:
I'm just wondering do you need negative denormal values? (it's just a question for further improvements)

muralinr:
> I'm just wondering do you need negative denormal values? (it's just a question for further improvements)
@CAHEK7 We have denorm rms test design document here from Carlos.  We will discuss further enhancements on top of this document. 
https://amdcloud-my.sharepoint.com/:w:/g/personal/carhuang_amd_com/EQoRL_iEpjhJqzRW4tMks4gBweAhTuRZpKlc1caXd0v2Qw?e=roM4w3&CID=A9B3E1F6-1D3B-46A0-8D8B-EC75D644EDDE&wdLOR=c67D49826-63B3-4C9E-8DA5-54E6BE0B34D5

CAHEK7 (approved):

CAHEK7:
@muralinr the document mentioned sign bit, so I added an ability to generate negative denormal values but turned it off by default. Feel free to use it.

junliume (approved):

PR #2334, opened 2023-08-20
Reference kernel for 3D convolution for non-packed tensors

amberhassaan:
Current reference kernels used for verifying convolution (aka naive direct convolution kernels) work only with packed tensors. This PR enhances the state of code  by creating variants of these kernels that work with non-packed tensors by taking custom stride specification as kernel parameters.

PR #2394, opened 2023-09-18
Test non-packed inputs with naive reference convolution kernels

amberhassaan:
This PR follows up #2334 where I fix the testing to code to allow testing non-packed tensors for convolution reference kernel. 
Changes included:
* Modify tensor generation for reference kernel test to expand the strides by multiplying strides with an integer.
* Changes in call stack to allow non-strided inputs to be accepted (some checks disallow non-strided inputs to begin with).


driver/random.hpp
```diff
@@ -91,6 +91,7 @@ inline T gen_0_to_B(T B)
 template <typename T>
 inline T gen_A_to_B(T A, T B)
 {
+    assert(B >= A);
```
CAHEK7:
```suggestion
    assert(B > A);
```
I guess B must not be equal to A, otherwise we may get integer division by 0.

But interesting fact - while B < A does not make much sense, it will work fine for signed integers, since `X%(Y)==X%(-Y) `, but won't work for unsigned integers (at all) and floats (will behave strangely).

CAHEK7 (changes requested):

driver/random.hpp
amberhassaan: good point. I was thinking about that myself. Initially I though B was inclusive to the range, but now it's clear that it's exclusive.

test/gpu_reference_kernel.cpp
```diff
@@ -310,6 +312,29 @@ static std::string miopen_type_to_string(miopenDataType_t type)
...
+        std::cout << name << ": [";
+        for(const auto& v : vec)
+        {
+            std::cout << v << ", ";
+        }
+        std::cout << "]\n";
+    };
+
+    pvec("orig strides", strides);
+    auto c = prng::gen_A_to_B(1, 3);
```
CAHEK7: Is there a plan to get non-strided tensors?

CAHEK7 (dismissed):

test/gpu_reference_kernel.cpp
amberhassaan: These are some initial tests. I have another variant, where I multiply a prefix of strides with a random constant instead of all strides. Need to discuss this with @JehandadKhan and others.

test/gpu_reference_kernel.cpp
```diff
@@ -310,6 +312,29 @@ static std::string miopen_type_to_string(miopenDataType_t type)
...
+    // int c = 2;
+    for(auto& v : strides)
+    {
+        v = v * c;
```
junliume, line 316: @amberhassaan HIP tidy issue:
```
[2023-09-18T16:25:31.036Z] /var/jenkins/workspace/Open_amber_tests-non-packed-conv/test/gpu_reference_kernel.cpp:333:11: warning: Consider using std::transform algorithm instead of a raw loop. [useStlAlgorithm]
[2023-09-18T16:25:31.036Z]         v = v * c;
[2023-09-18T16:25:31.036Z]           ^
```

CAHEK7: Thats exactly I was thinking about but decided to keep it!
Didn't know that clang-tidy does that kind of analysis.

iq136boy (commented):
@amberhassaan did you test the reference kernel when g>1?

amberhassaan:
> @amberhassaan did you test the reference kernel when g>1?

Hi @iq136boy , yes, I believe there are tests that provide  a value of `g` that is `> 1`.

test/gpu_reference_kernel.cpp
amberhassaan: Tbh, I take issue with this suggestion from `hip-tidy`. The original code is more readable and `std::transform` will obscure it. I would make such a change for a large vector input but that will never happen here. So, I'll silence `hip-tidy` with a comment.

junliume: @amberhassaan okay but please suppress it only for this snippet of code instead of globally. :)

amberhassaan: Yes, that was always my intention. This is a small silly loop that might itself change by the end of this PR.

junliume:
@amberhassaan could you check the "target" branch of this PR?

amberhassaan:
> @amberhassaan could you check the "target" branch of this PR?

Hi @junliume , This is based off of PR #2334 and I have set it as the merge target because I wanted to show the actual diff over 2334 to the reviewers. Once #2334 merges, I'll switch the target to `develop` (or you can do that too).

atamazov:
@amberhassaan 
> Hi @junliume , This is based off of PR #2334 and I have set it as the merge target because I wanted to show the actual diff over 2334 to the reviewers. Once #2334 merges, I'll switch the target to `develop` (or you can do that too).

👍 I think this is proper approach. But please mark this as a draft to avoid accidental merging into the wrong branch.

src/ocl/convolutionocl.cpp
```diff
@@ -416,11 +416,12 @@ void ConvolutionDescriptor::ConvolutionForward(Handle& handle,
                                                size_t workSpaceSize) const
 {
     MIOPEN_LOG_I("algo = " << algo << ", workspace = " << workSpaceSize);
-
+#if 0
```
atamazov: Why this hack is necessary? Of course I know that [F8 support needs some improvement](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2251#issuecomment-1728457592).

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -134,12 +134,8 @@ ConvSolution ConvDirectNaiveConvBwd::GetSolution(const ExecutionContext& ctx,
     kernel.l_wk.push_back(1);
     kernel.l_wk.push_back(1);
 
-    const auto is_f8 = [&]() {
-        if(kernel.kernel_file == "fp8_naive_conv.cpp")
-            return true;
-        else
-            return false;
-    }();
+    const auto is_f8 = (kernel.kernel_file == "fp8_naive_conv.cpp");
```
atamazov, line 137: [Notice] This is good, but I recommend avoiding making any changes that beyond main PR topic in the future.

test/gpu_reference_kernel.cpp
```diff
@@ -341,11 +367,11 @@ struct gpu_reference_conv_2d : gpu_reference_kernel_base
...
             int c_per_group = c / g;
-            int k_per_group = k / g;
+            // int k_per_group = k / g;
 
-            int in_sz  = g * n * c_per_group * hi * wi;
-            int wei_sz = g * k_per_group * c_per_group * fy * fx;
-            int out_sz = g * n * k_per_group * ho * wo;
+            // int in_sz  = g * n * c_per_group * hi * wi;
+            // int wei_sz = g * k_per_group * c_per_group * fy * fx;
+            // int out_sz = g * n * k_per_group * ho * wo;
```
atamazov: Please remove this or explain in comments why this is necessary to keep.

test/gpu_reference_kernel.cpp
```diff
@@ -447,6 +482,13 @@ struct gpu_reference_conv_2d : gpu_reference_kernel_base
                                  wei.data.data(),
                                  sizeof(TRef) * wei_sz,
                                  hipMemcpyHostToDevice) == hipSuccess);
+                // TODO(Amber): copy output before computation because output may
```
atamazov: Usual advice about doxygen formatting and github username. Also I do not recommend copying comments. But you can use the `\anchor` tag like
```C++
/// \anchor test_naive_gpu_kernel_why_copy_output_before_computation
/// \todo ...
/// ...
```
and then simply
```C++
/// \ref test_naive_gpu_kernel_why_copy_output_before_computation
```

test/gpu_reference_kernel.cpp
```diff
@@ -447,6 +482,13 @@ struct gpu_reference_conv_2d : gpu_reference_kernel_base
...
+                // be not be packed, and convolution may update only a subset of
+                // indices
+                EXPECT(hipMemcpy(out_dev,
```
atamazov, line 439: hipMemcpy should not be used with OpenCL backend.

BTW We no longer support OCL backend so you can simply keep this as is or remove the whole OpenCL block of code (I would prefer the former due to [this](https://github.com/ROCmSoftwarePlatform/MIOpen/discussions/1687#discussioncomment-7124363))

test/gpu_reference_kernel.cpp
```diff
@@ -719,11 +774,11 @@ struct gpu_reference_conv_3d : gpu_reference_kernel_base
             int wo          = conv_out_size(wi, px, dx, fx, sx);
             int do_         = conv_out_size(di, pz, dz, fz, sz);
             int c_per_group = c / g;
-            int k_per_group = k / g;
+            // int k_per_group = k / g;
```
atamazov: same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2394#discussion_r1337264777

test/gpu_reference_kernel.cpp
```diff
@@ -518,9 +560,15 @@ struct gpu_reference_conv_2d : gpu_reference_kernel_base
                                                nullptr);
                 EXPECT(status == CL_SUCCESS);
 #elif MIOPEN_BACKEND_HIP
+                // TODO(Amber): copy output before computation because output may
```
atamazov: [Q] Can you please explain a bit more why this TODO is necessary? Are you sure that the word "indices" is adequate in this context?

test/gpu_reference_kernel.cpp
```diff
@@ -600,9 +648,16 @@ struct gpu_reference_conv_2d : gpu_reference_kernel_base
                 EXPECT(hipMemcpy(
                            in_dev, in.data.data(), sizeof(TRef) * in_sz, hipMemcpyHostToDevice) ==
                        hipSuccess);
+                // TODO(Amber): copy output before computation because output may
```
atamazov: same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2394#discussion_r1337270907

test/gpu_reference_kernel.cpp
```diff
@@ -821,6 +885,13 @@ struct gpu_reference_conv_3d : gpu_reference_kernel_base
...
```
atamazov: same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2394#discussion_r1337270907

test/gpu_reference_kernel.cpp
```diff
@@ -980,9 +1057,16 @@ struct gpu_reference_conv_3d : gpu_reference_kernel_base
...
```
atamazov: same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2394#discussion_r1337270907

src/ocl/convolutionocl.cpp
amberhassaan: This is necessary to allow accepting non-strided tensors. In fact, we should remove this check all together.

src/solver/conv_direct_naive_conv_bwd.cpp
amberhassaan: Agree. I just couldn't resist making this change :).

test/gpu_reference_kernel.cpp
amberhassaan: Sorry, the hyperlink to `this` above is broken. I don't think I have any `hipMemCpy` outside of `#elif MIOPEN_BACKEND_HIP`, so OpenCL backend is unaffected. I could perhaps add the corresponding `clEnqueWriteBuffer` to OpenCL backend if we care to maintain it. If not, I prefer deleting all `OpenCL` code from this file.

test/gpu_reference_kernel.cpp
amberhassaan: So output is non-packed, i.e. larger than what's needed for convolution and may have data that is not touched by convolution operation. In order to verify the output, GPU buffer should have the same data in the `untouched` parts as the CPU, which is achieved by copying the output to the GPU before the convolution is invoked.

atamazov: @amberhassaan Have we made sure that untouched areas (gaps) in the reference buffer are zeroed? The random data may contain INF or NAN, etc., and this will spoil the comparison results.

atamazov: @amberhassaan 
> So output is non-packed, i.e. larger than what's needed for convolution and may have data that is not touched by convolution operation. In order to verify the output, GPU buffer should have the same data in the untouched parts as the CPU, which is achieved by copying the output to the GPU before the convolution is invoked.

[Recommendation] Put this explanation into C++ comments as it seems much more clear to me ;)

In fact I would prefer zeroing both reference and output buffers prior computations, and _only if_ the output tensor is non-packed. There is some MIOpen kernel that can do this job quickly. Is it possible to do?

src/ocl/convolutionocl.cpp
atamazov: [resolved]

test/gpu_reference_kernel.cpp
atamazov: [resolved]

src/solver/conv_direct_naive_conv_bwd.cpp
atamazov: [resolved]

test/gpu_reference_kernel.cpp
atamazov: [resolved]

test/gpu_reference_kernel.cpp
atamazov: @amberhassaan URL fixed, sorry.

atamazov: @amberhassaan I prefer keeping this as is, but the final decision is up to you.

test/gpu_reference_kernel.cpp
atamazov: @amberhassaan Alternatively, we can implement a tensor-layout-aware verification. This is more robust solution, but implementation seems to be much more complicated.

test/gpu_reference_kernel.cpp
atamazov: [resolved]

test/gpu_reference_kernel.cpp
atamazov: [resolved]

test/gpu_reference_kernel.cpp
atamazov: [resolved]

test/gpu_reference_kernel.cpp
atamazov: [resolved]

@amberhassaan Please hide all [resolved] comments (I can't do that), thanks.

test/gpu_reference_kernel.cpp
CAHEK7: Tensor-layout-aware may help to compare only changed parts, at least as I understood it. But we also want to check that unchanged parts are actually unchanged, and it means that zero-filled tensors are not so good and fp-comparison is not a good idea too. Ideally it must compare the untouched parts as a binary data and doesn't care about all of those `INF`s and `NAN`s.
N.B. as far as I know current random functions do not generate `INF`s and `NAN`s, but there are some direct `NAN` initialiations.

amberhassaan: @atamazov , @CAHEK7, good suggestions. I find layout-aware verification a bit much given the deadlines for this work. I see two ways forward (and I prefer 1)
1. Fill the output tensor with random integers (that's how we generate inputs). Our verification (against cpu code) compares all data elements (including those not touched by convolution). 
2. Zero-out the entire tensor, which is not as robust as @CAHEK7 pointed out because we're relying on zero being a special value.

test/gpu_reference_kernel.cpp
amberhassaan: Discussed this with Jehandad and we decided that it's OK to remove OpenCL code. I can do that in this PR @atamazov if you're OK with that or leave it as a follow up action.

atamazov: @amberhassaan Removing looks good -- as I said, the final decision is up to you.

test/gpu_reference_kernel.cpp
atamazov: @CAHEK7 
> But we also want to check that unchanged parts are actually unchanged

Why? AFAIU the gaps in the real life may contain any garbade.

atamazov: @amberhassaan 
> because we're relying on zero being a special value.

Zero is and will forever be a special value in all IEEE-754 compliant FP numbers and their extensions, as well in all two's complement integers. Are we going to support other number formats in the visible future?

src/ocl/convolutionocl.cpp
```diff
@@ -402,6 +377,84 @@ static void ConvForwardCheckNumerics(const Handle& handle,
     }
 }
 
+void ConvolutionDescriptor::ValidateConvTensors(const ConvTensors& tensors) const
```
atamazov: [Recommendation] Let's avoid tautology in the new code.
```suggestion
void ConvolutionDescriptor::ValidateTensors(const ConvTensors& tensors) const
```
BTW maybe it is better to keep this as a freestanding function and pass group_count as a parameter?

test/gpu_reference_kernel.cpp
atamazov: @amberhassaan ~If you copy reference output buffer into device output buffer before convolution, then verification will pass even if convolution did nothing.~ [irrelevant]

src/ocl/convolutionocl.cpp
```diff
@@ -417,11 +470,6 @@ void ConvolutionDescriptor::ConvolutionForward(Handle& handle,
 {
     MIOPEN_LOG_I("algo = " << algo << ", workspace = " << workSpaceSize);
 
-    if(!(xDesc.IsPacked() && wDesc.IsPacked() && yDesc.IsPacked()))
-    {
-        MIOPEN_THROW(miopenStatusNotImplemented, "Only fully packed tensors are supported");
-    }
-
```
averinevg, line 423: This check must be added to all convolution solvers that do not support non-packed tensors

src/include/miopen/convolution.hpp
```diff
@@ -36,6 +36,7 @@
 #include <miopen/names.hpp>
 #include <miopen/invoke_params.hpp>
 #include <miopen/invoker.hpp>
+#include <miopen/conv/tensors.hpp>
```
averinevg, line 39: [Recommendation] Since this header was added here, it can be removed from src/ocl/convolutionocl.cpp

Or you could use a forward declaration here and not add this header, which is preferable in my opinion. The more includes, the longer the compilation time

test/gpu_reference_kernel.cpp
atamazov: @amberhassaan Aha, now I see. Well, I still think that zeroing is better, but I am Ok with randomizing gaps as well.
[resolved]

atamazov (commented):
One non-blocking comment

src/ocl/convolutionocl.cpp
```diff
@@ -402,6 +377,84 @@ static void ConvForwardCheckNumerics(const Handle& handle,
     }
 }
 
+void ConvolutionDescriptor::ValidateConvTensors(const ConvTensors& tensors) const
+{
+
+    // Group stride in current TensorDescriptor is implicit. When invoking kernels,
+    // we need to add the group dimension G and compute its stride. We want the stride
+    // left of C to be a multiple of group count G. e.g. for NCHW, the stride for N
+    // should be a multiple of G so that we can compute the strides for NGCHW
```
averinevg: Here the functionality overlaps with ValidateGroupCount(). We also have Problem::ValidateGroupCount(). It would be nice to sort this out.

averinevg (changes requested):

test/gtest/conv_api_strided_tensors.cpp
```diff
@@ -139,7 +139,6 @@ class ConvStridedTensors : public ::testing::Test
     std::vector<float> h_output;
 };
 
-// This test should be replaced when strided tensors are fully implemented
```
averinevg, line 142: FYI This test was needed to verify that the library returns an error if non-packed tensors are used for convolutions.

amberhassaan: Are you OK with this change or should we remove the test altogether?

averinevg: This test is no longer needed and can be removed. We need to make a test for non-packed tensors instead.

test/gpu_reference_kernel.cpp
amberhassaan: Will remove (next commit once CI completes).

src/ocl/convolutionocl.cpp
amberhassaan: @averinevg : What's the best way to go about finding these solvers? How about I update the test to not apply the check if the algorithms are "naive" or "ck" based?

averinevg: @amberhassaan You can find them all in `src/include/miopen/solver.hpp`. It is better to implement this check inside solvers' `IsApplicable()` because this is exactly the place where such checks should be made.

averinevg: A general function for checking can be added to `conv::ProblemDescription`

amberhassaan: Got it. However, due to shortage of time, I'd like an easier solution. It seems like we have many many solvers for convolution and updating all of them will take time.
@JehandadKhan , @junliume : thoughts? context: we need to disable the check above to allow non-packed tensors in convolution, but then the check is needed for all the solvers that don't support non-packed tensors, and those are like 99% of them.

test/gtest/conv_api_strided_tensors.cpp
amberhassaan: True. We don't have a public API test for non-packed tensors. I have modified reference convolution tests to generate non-packed strides in this PR and #2418 will add tests for 3D convolution, but those target specific solvers. In general, it seems we lack public API tests (or is it my misconception?). A lot of tests seem to use internal APIs.

src/ocl/convolutionocl.cpp
JehandadKhan: In `conv::ProblemDescription` create a new helper function which checks if all the tensor ( input. output, weights) are packed. Then in each solver's `IsApplicable` call it to enforce the condition. While I agree it is cumbersome but in the current implementation, that is how all of this is implemented. For example, when a new layout/data type is added, we have go through all solver and make sure the old solvers respect the new feature.

amberhassaan: ok, I agree with all of that. I just want to postpone this change to after our deadline.

junliume:
@CAHEK7 @averinevg @DrizztDoUrden @atamazov @JehandadKhan @iq136boy :
CI has passed, last call for any blocking reviews before merging.

src/ocl/convolutionocl.cpp
amberhassaan: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2444 created to handle this. 

Proposal: We remove `Problem::ValidateGroupCount()` and then combine `ValidateGroupCount()` in `convolutionocl` with `conv::ProblemDescription::ValidateTensors()` (currently `ConvDescriptor::ValidateTensors()` and use it everywhere. This will allow getting `N` and `C` lengths and strides easily from accessors in `conv::ProblemDescrioption`. 

Alternate Proposal: Combine all `ValidateGroupCount` with `ConvDescriptor::ValidateTensors()`. Downside: this requires logic to extract `N` and `C` lengths and strides.

src/include/miopen/convolution.hpp
amberhassaan: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2444 created to handle this in case @junliume wants to merge this PR as is (due to urgency).

src/ocl/convolutionocl.cpp
amberhassaan: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2444 created to handle this as it's a more involved change. @junliume @JehandadKhan

src/ocl/convolutionocl.cpp
amberhassaan: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2444 created to handle this in case we decide to merge the PR as is: @junliume

amberhassaan:
@junliume : https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2444 created to track the unresolved stuff.

CAHEK7 (dismissed):
Seems fine, but conflicting with #2418 and I'm not sure which one has to be merged first.

amberhassaan:
> Seems fine, but conflicting with #2418 and I'm not sure which one has to be merged first.

#2418 was based on top of #2394 (this PR), so this should merge first.

src/ocl/convolutionocl.cpp
```diff
@@ -402,6 +377,84 @@ static void ConvForwardCheckNumerics(const Handle& handle,
...
+            g_stride_index = 0; // stride index for N;
+        }
+        else if(l == miopenTensorNHWC || l == miopenTensorNDHWC)
+        {
+            // stride index for W. Normally this would be 2nd-last stride but we store
+            // strides in NCHW order for some weird reason.
+            g_stride_index = td.GetStrides().size() - 1;
+        }
+
+        if(g_stride_index != 1)
```
averinevg:
```suggestion
        if(g_stride_index != -1)
```

src/ocl/convolutionocl.cpp
```diff
@@ -402,6 +377,84 @@ static void ConvForwardCheckNumerics(const Handle& handle,
...
+        {
+            g_stride_index = 0; // stride index for N;
+        }
+        else if(l == miopenTensorNHWC || l == miopenTensorNDHWC)
+        {
+            // stride index for W. Normally this would be 2nd-last stride but we store
+            // strides in NCHW order for some weird reason.
+            g_stride_index = td.GetStrides().size() - 1;
+        }
+
```
averinevg, line 404: Let's add exception for other layouts

averinevg (dismissed):

test/gtest/conv_api_strided_tensors.cpp
averinevg: Let's comment out this test for now and create an issue to handle the correct test for non-packed tensors using the same cpp

src/include/miopen/convolution.hpp
amberhassaan: Done in latest commit.

src/ocl/convolutionocl.cpp
amberhassaan: Done in latest commit.

junliume (approved):

PR #2396, opened 2023-09-18
[MI300] add CI test stages

cderb:
Updates make test for MI300.
Add MI300 stages to Jenkinsfile.


CAHEK7 (dismissed):
Seems good

junliume:
@cderb "All nodes of label ‘rocmtest&&miopen&&(gfx94X)’ are offline" :(

bghimireamd:
LGTM. 

PS: MI300 is already supported in [batch norm NHWC solvers](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/solver/batchnorm/forward_inference_ck.cpp) since we use[ is_ck_supported_hardware](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/solver/ck_utility_common.hpp#L49)  in its IsApplicable.

JehandadKhan (approved):

src/include/miopen/sqlite_db.hpp
```diff
@@ -230,8 +230,7 @@ class SQLiteBase
             dbInvalid = true;
             return;
         }
-
-        if(!is_system && !filename.empty())
```
atamazov, line 234: [Q] Why this is necessary?

src/kernel_warnings.cpp
```diff
@@ -68,6 +68,7 @@ static std::vector<std::string> HipKernelWarnings()
         "-Wno-double-promotion",
         "-Wno-exit-time-destructors",
         "-Wno-extra-semi",
+        "-Wno-extra-semi-stmt",
```
atamazov, line 71: [Q] This looks like a workaround of some other problem (albeit non-critical). Can we resolve it right here?

test/CMakeLists.txt
```diff
@@ -729,7 +730,7 @@ if(${MIOPEN_TEST_WITH_MIOPENDRIVER})
         COMMAND MIOPEN_FIND_MODE=1 MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvDirectNaiveConvFwd $<TARGET_FILE:MIOpenDriver> ${MIOPENDRIVER_MODE_CONV} --forw 1 --in_layout NCHW --out_layout NCHW --fil_layout NCHW -n 256 -c 1024 -H 14 -W 14 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1
     )
 
-    add_custom_test(test_miopendriver_regression_float_half_gfx10 SKIP_UNLESS_ALL GFX900_DISABLED GFX906_DISABLED GFX908_DISABLED GFX90A_DISABLED GFX103X_ENABLED HALF_ENABLED
+    add_custom_test(test_miopendriver_regression_float_half_gfx10 SKIP_UNLESS_ALL GFX900_DISABLED GFX906_DISABLED GFX908_DISABLED GFX90A_DISABLED GFX94X_ENABLED GFX103X_ENABLED HALF_ENABLED
```
atamazov: This test is for gfx10, please revert this change.

atamazov (changes requested):
Almost good!

src/include/miopen/sqlite_db.hpp
cderb: Not a necessary change, but the removed boolean is always true do to the previous if statement.

test/CMakeLists.txt
cderb: Will do.

src/kernel_warnings.cpp
cderb: I'll need to regain access to a gfx94x machine before I can reassess.

src/include/miopen/sqlite_db.hpp
atamazov: [resolved]

test/CMakeLists.txt
atamazov: [resolved]

src/kernel_warnings.cpp
atamazov: @cderb [Recommendation] Fix kernel build warnings right here and revert this change.

----
As usual, "recommendations" are up to the author discretion. The author can actually resolve or just mark as resolved -- no explanation needed.

atamazov (approved):
LGTM!

junliume (approved):

test/CMakeLists.txt
```diff
@@ -117,6 +117,7 @@ if(NOT (MIOPEN_TEST_GFX900 OR MIOPEN_TEST_GFX906 OR MIOPEN_TEST_GFX908 OR MIOPEN
             OUTPUT_VARIABLE ROCMINFO_OUTPUT
             RESULT_VARIABLE ROCMINFO_EXIT_STATUS
         )
+        message(STATUS "${ROCMINFO_OUTPUT}")
```
atamazov, line 120: @cderb This clutters cmake output. Why this is necessary? If something gets wrong you can manually run rocminfo and inspect its output. In CI logs we have `clinfo` output; we can replace `clinfo` with `rocminfo` there if necessary. We can print rocminfo output at highel verbosity level. Finally we can only do that if something gets wrong.

Variants of resolving this comment:
- Replace `clinfo` with `rocminfo` in Jenkinsfile
- Replace STATUS with VERBOSE (`cmake --log-level=VERBOSE ...` will show it)
- (preferred) Only log rocminfo output if MIOPEN_TEST_GPU_DETECTION_FAILED is TRUE.

Maybe I'll do that myself in my next PR,

atamazov: Resolved by #2490

PR #2397, opened 2023-09-19
[tests] remove direct std::random_device usage.

### LINKED ISSUE #2392: [tests] remove direct std::random_device usage for prng seed.
There are few places where `std::random_device` passed directly to random number generators as a seed.
The drawback of this solution described here https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2373#discussion_r1320558139
and in the following PR #2337 

The main points are - random seed should be logged and there should be a way to specify custom random seed.
All the required facilities are provided in #2337

CAHEK7:
For further development please use functions and random generator from `driver/random.hpp` or `test/random.hpp`.
#2392 contains more detailed description why.

Closes #2392.


bghimireamd:
LGTM

test/gtest/log.cpp
```diff
@@ -29,8 +29,8 @@
 
 #include <miopen/config.h>
 #include <miopen/fusion_plan.hpp>
-#include <stdlib.h>
-#include <random>
+#include <cstdlib>
+#include "random.hpp"
```
amberhassaan: minor: perhaps we should name (and move) our "random.hpp" to "miopen/random.hpp" or if that's not feasible then "miopen_random.hpp" to avoid confusion with `<random>` from STL.

amberhassaan (approved):
Looks good. Left one minor comment (non-blocking).

test/gtest/log.cpp
atamazov: @amberhassaan
> perhaps we should name (and move) our "random.hpp" to "miopen/random.hpp" 

The library does not use random.hpp (while tests and driver use different versions of this file). Therefore, this is not possible and not needed.

> or if that's not feasible then "miopen_random.hpp" to avoid confusion with <random> from STL.

But I do not see any confusion. For example, double quotes imply that the included file is not from the standard library.

test/bn_peract_test.cpp
```diff
@@ -233,11 +233,11 @@ struct verify_forward_train_bn_per_activation
             runMean = tensor<U>{rs_n_batch, rs_channels, rs_height, rs_width};
             runVar  = tensor<U>{rs_n_batch, rs_channels, rs_height, rs_width};
 
-            const U Data_scale = static_cast<U>(0.001);
+            const double Data_scale = 0.001;
```
atamazov, line 236: [Notice] This seems fine so far, but what about cases when U is not one of the natively supported types (e.g. BF16 or FP8)? Just curious.

atamazov (approved):
LGTM

CAHEK7:
Had to force-push the last commit since accidentally updated `fin` submodule. It shouldn't affect any previous history.

And had to update ck batchnorm test threshold since it's incredibly sensitive to the input data. I did lots of experiments and even swapping the order of sign and value generation may break the test while random values parameters like mean value, variance and distribution have not been changed noticeably.

Moreover those tests are currently broken even in develop branch, but they failed due to some HIP runtime issues.

CAHEK7:
@atamazov @amberhassaan @bghimireamd could you re-review the changes since I did some significant updates.

test/bn_3d_peract_test.cpp
```diff
@@ -108,11 +108,11 @@ struct verify_forward_train_3d_bn_per_activation
             runMean = tensor<U>{rs_n_batch, rs_channels, rs_depth, rs_height, rs_width};
             runVar  = tensor<U>{rs_n_batch, rs_channels, rs_depth, rs_height, rs_width};
 
-            const U Data_scale = static_cast<U>(0.001);
+            const double Data_scale = 0.001;
```
amberhassaan, line 111: why is this `double` and not `U`? Just wondering.

test/gtest/log.cpp
amberhassaan: You're right `"random.hpp"` should convey that it's local. However, we are not consistent in our repository and we often use `<>` for local MIOpen includes.

amberhassaan (approved):
Looks good. Have one question that needs clarification.

test/bn_3d_peract_test.cpp
CAHEK7: Potentially computing in a higher precision domain may give slightly better distribution. It may not be so critical since distribution calculation itself heavily depends on rounding and other factors.
for example distribution for range 1000 and scale 0.001 casted to float at the end (double case, blue) or during calculations (U-case, orange):
chart

Another point - since x86 does not support `half` (and other low-precision ML types) natively, all the calculations in that domain may not be so efficient and thus it's much faster to calculate in a native domain and convert the result at the end.

amberhassaan: Makes sense. Although, I don't understand the plot. What am I looking at?

CAHEK7: It's a probability density plot.
We generate "discreet" random values in range [-0.999; 0.999] and count how many numbers come into the bins [-0.999; 0.998), [0.998; 0.997) ... [0; 0.001) ... [0.999; 1). So for the uniform distribution it must be linear.
By "discreet" I mean we generate true discreet values in range  [-999; 999] and multiply them by 0.001 so mathematically they are discreet with 0.001 step, but due to finite precision of fp values they may have a bit different nature and thus we can get some spikes because rounding and loss of precision shifted some values to the different bins.

amberhassaan: OK, I see. It is strange how you see variation in bucket density only on left side. I would imagine this would happen throughout.

CAHEK7: That strange behavior is not so critical for the "real" numbers. For example the sequence 0.9936, 0.9944, 9.9956, 0.99964, 0.99976 generates distribution like 0, 2, 0, 2, 1. It's just not so perfectly numerically uniform. It is still discreet - we don't have any values in the middle, it still uniform since each distinctive value has uniform probability, but numerically the difference between two neighbor values is fluctuating.

CAHEK7:
@junliume I fixed merge conflicts and all the tests passed.

test/gtest/log.cpp
pfultz2: The quoted form should only be used for headers that are relative to the including file, and everything else should use the angle brackets. In general, MIOpen follows this rule. This is also the guideline from C++ Core Guidelines [SF.12](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform).

amberhassaan: Didn't know that we were following that core guidelines. Some projects exclusively use `<>` for system includes and `""` for project includes.

atamazov: @CAHEK7 Please resolve https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2397#discussion_r1368869376

atamazov: @CAHEK7 "../random.hpp" here is expected to comply the rule.

junliume (approved):

test/gtest/log.cpp
CAHEK7: @atamazov @pfultz2 done.
But what about the others includes? Mixing different paradigms of including headers may be confusing.
```CPP
#include "log.hpp"
#include "tensor_util.hpp"
#include "get_handle.hpp"
...
#include "../random.hpp"
```
Shall everything be fixed uniformly across the project, because it's a very common case where we include files from the project using quotes and implying relative includes.

Almost all the files in `gtests` directory and some of the files in the `test` directory should be fixed.

atamazov:
@CAHEK7 
> Had to force-push the last commit since accidentally updated fin submodule. It shouldn't affect any previous history.

You can simply revert fin changes and commit this into your development branch.

atamazov (approved):
LGTM.

JehandadKhan:
@junliume  Can we please get this merged in?

junliume (approved):

PR #2399, opened 2023-09-19
SWDEV-414487 : enable 2d conv + bias + activ CK kernel for MI300

bghimireamd:
* In PR I am enabling fused CK kernel (conv + bias + activ) on MI300.


junliume (approved):
LGTM simple enough to be blessed, and test infra needs to be ready though ...

PR #2400, opened 2023-09-19
[tests] remove std::rand usage

### LINKED ISSUE #2393: [tests] remove rand() function usage
As described in #2234 function `rand()` shouldn't be used for various reasons.
It was replaced inside a common place of random number generators and helpers, but there are few places where it is used directly.

Direct `rand()` calls should be replaces to the proper functions from #2337 PR.

CAHEK7:
PR removes a few places where we used `std::rand`. More details in #2393

Closes #2393


src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
```diff
@@ -274,8 +276,8 @@ ConvOclDirectFwdLegacyExhaustiveSearch::SearchImpl(const ConvolutionContext& ctx
     std::vector<Tgpu> wei_sys_buf(weights_sz);
     for(size_t i = 0; i < weights_sz; i++)
     {
-        wei_sys_buf[i] = static_cast<Tgpu>((rand() * (1.0 / RAND_MAX) - 0.5) *
-                                           0.001); // NOLINT (concurrency-mt-unsafe)
+        wei_sys_buf[i] =
+            static_cast<Tgpu>((std::generate_canonical<double, 11>(prng) - 0.5) * 0.001);
```
JehandadKhan, line 280: Why not use the `miopen::prng` variant ?

JehandadKhan: Why not use `miopen::prng` variant ?

JehandadKhan (changes requested):

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
CAHEK7: I'm not sure about including driver headers into solver. If it's ok, then `miopen::prng` is the choise.

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
```diff
@@ -253,14 +255,14 @@ ConvOclDirectFwdLegacyExhaustiveSearch::SearchImpl(const ExecutionContext& ctx,
 
 #if MIOPEN_ALLOC_BUFFERS
```
atamazov: This `MIOPEN_ALLOC_BUFFERS` is a leftover of functionality removed some time ago. I highly recommend removing all this stuff. This way, the `rand()` calls can simply be removed.

atamazov: @JehandadKhan @CAHEK7 ☝️

src/solver/conv_ocl_dir2Dfwd_exhaustive_search.cpp
JehandadKhan: I see your point, perhaps move the header to MIOpen and then use it both in the solver and the driver

JehandadKhan (approved):

CAHEK7:
@junliume seem this one can be merged

junliume (approved):

PR #2320, opened 2023-08-11
Add support for layernorm primitive

seungmanhan:
- Added basic layernorm operation and kernel
- Added driver test and gtest for layernorm
- When make tidy was executed, an error occurred in 324~340 of src/hippoc/hipoc_program.cpp, which is an unmodified file
- cppcheck didn't show any problems
- The parts I worked on were guarded by a MIOPEN_BETA_API macro(MIOPEN_BETA_API needs to be declared for testing)
- This PR is related to [this](https://github.com/orgs/ROCmSoftwarePlatform/projects/34/views/1?pane=issue&itemId=33502153) and is aimed at 1st pipe-cleaning PR
<details><summary>Simple perf test</summary>
<p>

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">

</head>

<body link="#0563C1" vlink="#954F72">

op_name | dtype | size | direction | rocm | modnn | rocm / modnn | MIOpen HIP | MIOpen HIP/rocm
-- | -- | -- | -- | -- | -- | -- | -- | --
layernorm | float16 | [4, 8, 8] | fwd | 10,496 | 20,320 | 0.52 | 6,489 | 1.62
layernorm | float16 | [4, 8, 8] | bwd | 17,632 | 37,744 | 0.47 | 11,165 | 1.58
layernorm | float16 | [16, 64, 256] | fwd | 11,376 | 31,232 | 0.36 | 10,044 | 1.13
layernorm | float16 | [16, 64, 256] | bwd | 25,712 | 287,583 | 0.09 | 32,960 | 0.78
layernorm | float16 | [32, 256, 1024] | fwd | 97,056 | 40,192 | 2.41 | 87,325 | 1.11
layernorm | float16 | [32, 256, 1024] | bwd | 197,071 | 265,423 | 0.74 | 567,417 | 0.35
layernorm | float32 | [4, 8, 8] | fwd | 11,328 | 19,616 | 0.58 | 6,169 | 1.84
layernorm | float32 | [4, 8, 8] | bwd | 15,488 | 25,264 | 0.61 | 11,307 | 1.37
[...]

PR #2405, opened 2023-09-20
rocblas: disable Beta API on Windows for HIP < 5.7

apwojcik:
The official HIP SDK release for Windows is 5.5.1. The Beta API is not available in that release. The API will be available from 5.7 and above on Windows.


src/gemm_v2.cpp
```diff
@@ -117,11 +121,14 @@ static inline rocblas_datatype rocBlasComputeType(const miopen::GemmDescriptor&
 
 auto rocBlasDataType(miopenDataType_t data_type)
 {
+#if !defined(_WIN32) || (HIP_PACKAGE_VERSION_FLAT >= 5007000000ULL)
```
JehandadKhan:
```suggestion
#if ROCBLAS_BETA_FEATURES_API
```

JehandadKhan (changes requested):

junliume:
@apwojcik could you follow up with the change request above?

apwojcik:
I did, and the change was made. I am waiting for @JehandadKhan to review it again.

src/gemm_v2.cpp
```diff
@@ -36,10 +36,14 @@
 #endif
 
 #if MIOPEN_USE_ROCBLAS
+#if defined(_WIN32) && (HIP_PACKAGE_VERSION_FLAT < 5007000000ULL)
```
JehandadKhan: @apwojcik  IIUC this means the `ROCBLA_BETA_FATURES_API` would always be enabled on Linux, regardless of the HIP version, which is not what we want ?

JehandadKhan (changes requested):

src/gemm_v2.cpp
JehandadKhan:
```suggestion
#if defined(_WIN32) || (HIP_PACKAGE_VERSION_FLAT < 5007000000ULL)
```

JehandadKhan: @apwojcik  Would you agree? Perhaps I am missing something.

JehandadKhan (dismissed):

src/gemm_v2.cpp
apwojcik: There was no check before and on Linux Beta Feature API was always on. I made it optional on Windows and available only for HIP SDK 5.7, but not for HIP SDK 5.5 (which is officialy released on Windows and there is no newer available).

apwojcik: I can change as you suggesting so all using older HIP SDK will compile on Linux, too.

apwojcik: @JehandadKhan do you want me to change that for Linux, too? By default, Linux installs the latest available ROCm - currently 5.7

src/gemm_v2.cpp
```diff
@@ -114,11 +118,14 @@ static inline rocblas_datatype rocBlasComputeType(const miopen::GemmDescriptor&
 
 auto rocBlasDataType(miopenDataType_t data_type)
 {
+#if ROCBLAS_BETA_FEATURE_API
```
junliume: @apwojcik minor CI issues since we treat warnings as errors:
```
/var/jenkins/workspace/MLLibs_MIOpen_roblas_beta_api/src/gemm_v2.cpp:121:5: error: 'ROCBLAS_BETA_FEATURE_API' is not defined, evaluates to 0 [-Werror,-Wundef]

#if ROCBLAS_BETA_FEATURE_API

    ^
```

atamazov: Typo.

atamazov (changes requested):

atamazov:
@apwojcik @JehandadKhan @junliume There are some important questions that need to be answered before this PR is merged in.

Some considerations first.

AFAICS `ROCBLAS_BETA_FEATURES_API` exposes some "beta" API entities in the roclblas interface. In rocblas 3.1.0 it includes:

- `rocblas_gemm_ex_get_solutions()`
- `rocblas_gemm_ex_get_solutions_by_type()`
- `rocblas_gemm_batched_ex_get_solutions()`
- `rocblas_gemm_batched_ex_get_solutions_by_type()`
- `rocblas_gemm_strided_batched_ex_get_solutions()`
- `rocblas_gemm_ex3()`

However the list of "beta" entities is not permanent. In some future versions or rocblas the beta API stuff may change. For example, some of the beta stuff may be moved to the normal rocblas API. Some new entities may appear under the ROCBLAS_BETA_FEATURES_API guard. Some existing beta things may be changed or removed etc. 

**_When we define ROCBLAS_BETA_FEATURES_API as 1 (prior inclusion of `rocblas.h`), the only meaning of that is that we would like to get access to all the beta features of rocblas._** Defining `ROCBLAS_BETA_FEATURES_API` prior inclusion of `rocblas.h` should be allowed with any rocblas version and can't lead to build errors.

However the list of beta features is not known and the only way to find them out is examining the rocblas version. For example, please see how we define and use `USE_ROCBLAS_GEMM_EX3` in PR #2480.

> The official HIP SDK release for Windows is 5.5.1. The Beta API is not available in that release. The API will be available from 5.7 and above on Windows.

What is the exact problem this PR is intended to resolve?
- If some "beta" feature is not available on Windows yet, then we need to examine rocblas version and disable the usage of that feature based on that. Usage of HIP version is not Ok because developers and QA may install some different rocblas on top of specific HIP SDK version.
- If rocblas versioning on Windows and Linux is different, then we need to take that into account. 

----
[Note] As we can see, this “ROCBLAS_BETA_FEATURES_API” is almost useless in practice. Its only function is to hide beta functions from the user. That is why I suggested not simply reusing similar approach in MIOpen, but [extend it with features that indicate the presence of a particular beta function and even report its version number](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1316419544).

atamazov:
@apwojcik BTW please check if #2480 resolves the problem on Windows.

atamazov (changes requested):
No go. Please see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2405#issuecomment-1787866776. It is likely that #2480 supersedes this PR (but I an not sure because I do not know what exactly is the problem with Windows).

src/include/miopen/handle.hpp
```diff
@@ -51,10 +51,14 @@
...
 
 #if MIOPEN_USE_ROCBLAS
+#if defined(_WIN32) && (HIP_PACKAGE_VERSION_FLAT < 5007000000ULL)
+#define ROCBLAS_BETA_FEATURES_API 0
+#else
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunused-macros"
 #define ROCBLAS_BETA_FEATURES_API 1
 #pragma clang diagnostic pop
+#endif
```
atamazov: [Side note] @JehandadKhan BTW is seems like handle.hpp does not use any beta API features from rocblas. Shall we simply remove ROCBLAS_BETA_FEATURES_API here?

JehandadKhan: Yes, please

atamazov: @JehandadKhan Oh, no. We have to define ROCBLAS_BETA_FEATURES_API in each header where rocblas.h is included. Because the order of inclusion of headers is not guaranteed, and rocblas.h in included only once, at the first appearance!

atamazov: @JehandadKhan ...and, as gemm_v2.cpp includes handle.hpp, removal of ROCBLAS_BETA_FEATURES_API from handle.hpp would lead to that rocblas.h is included (from handle.hpp) without beta stuff.

atamazov: @JehandadKhan 
> We have to define ROCBLAS_BETA_FEATURES_API in each header where rocblas.h is included

I've quickly inspected the code and found that rocblas.h is included only in handle.hpp and gemm_v2.cpp, so we are fine.

[end of the side thread]

apwojcik:
> @apwojcik BTW please check if #2480 resolves the problem on Windows.

It does partially resolve. However, for HIP SDK 5.5.1 on Windows, we still need to switch the beta API off since there is no `rocblas_computetype` available. It is available starting from HIP SDK 5.7

atamazov:
@apwojcik 
> It does partially resolve. However, for HIP SDK 5.5.1 on Windows, we still need to switch the beta API off since there is no `rocblas_computetype` available. It is available starting from HIP SDK 5.7

Can you please let me know which rocblas version is included into HIP SDK 5.5.1 and into HIP SDK 5.7? It should be 2.47 and 3.0 (I guess).

atamazov:
@apwojcik Please read https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2405#issuecomment-1787866776 and do not hesitate to ask if something left unclear. It explains why this PR should not be merged in. Sorry.

As far as I understand, Wnidows HIP SDK 5.5.1 contains rocblas 2.47 that do not support GEMM_EX3 feature, while rocblass 2.47 in ROCm supports it. Let me update #2480.

atamazov:
@apwojcik As #2480 is merged in, then let's make the necessary changes right here:
1. Revert everything in this branch
2. Update this branch from the fresh `develop`
3. In src/gemm_v2.cpp, replace
```C++
#define USE_ROCBLAS_GEMM_EX3 ((MIOPEN_ROCBLAS_VERSION_FLAT >= 2047000) && ROCBLAS_BETA_FEATURES_API)
```
with
```C++
#ifdef _WIN32
#define USE_ROCBLAS_GEMM_EX3 ((MIOPEN_ROCBLAS_VERSION_FLAT >= 3000000) && ROCBLAS_BETA_FEATURES_API)
#else
#define USE_ROCBLAS_GEMM_EX3 ((MIOPEN_ROCBLAS_VERSION_FLAT >= 2047000) && ROCBLAS_BETA_FEATURES_API)
#endif
```
~Profit!~ That should work for you.

apwojcik:
@atamazov it's done

atamazov (approved):
LGTM!

atamazov:
@apwojcik 
> force-pushed the roblas_beta_api branch from 76b249a to d3a4ed2

No problem in this PR. But just in case if you are not aware that force-pushing is not welcome here:

<details><summary><h3>⚠️ Please do not force-push (click to see why)</h3></summary><hr>

This is NOT recommended due to several reasons. For example, this makes previous code reviews invisible on the "Code changes" tab (due to overwritten commit hashes). Force-pushing can only be used more or less safely if the branch has NOT yet been published for review.

In order to resolve merge conflicts it is recommended to _merge `develop`_ into your branch. _Rebasing on top of `develop`_ leads to repeating merge conflicts (because **_we squash branches_** prior merging into `develop`) and therefore not recommended.

If develop has been substantially updated since branch creation, then it is recommended to update from `develop` even if there are no merge conflicts. This is to ensure that recent changes to `develop` are actually compatible with your changes. Otherwise, `develop` may become broken after the merge.
<hr></details>

atamazov:
@apwojcik BTW does this PR resolve the issue for you? Thanks!

apwojcik:
> @apwojcik BTW does this PR resolve the issue for you? Thanks!

Yes, it does. Thank you for your effort.

CAHEK7:
@junliume shall we update to the lates develop? It seems the change is trivial and affects only windows build.

CAHEK7 (approved):

junliume (approved):

PR #2409, opened 2023-09-22
Replace test_conv_igemm_dynamic_xdlops_bwd  with gtest

xinlipn:
Will add a few more tests into this same PR


test/gtest/conv_igemm_dynamic_xdlops_nhwc_bf16.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    std::vector<std::string> env_nhwc_wrw = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC"};
+
+    std::string v             = " --verbose";
+    std::string dis_bk_data   = " --disable-backward-data";
+    std::string dis_bk_wei    = " --disable-backward-weights";
+    std::string dis_fwd       = " --disable-forward";
+    std::string dis_vali      = " --disable-validation";
```
CAHEK7: Why do we need this?

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    std::vector<std::string> env_wrw_xdlops = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlops"};
+
+    std::string v           = " --verbose";
+    std::string dis_bk_data = " --disable-backward-data";
+    std::string dis_bk_wei  = " --disable-backward-weights";
+    std::string dis_fwd     = " --disable-forward";
+    std::string dis_vali    = " --disable-validation";
```
CAHEK7: Why do we need this?

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,224 @@
...
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicFwdXdlops"};
+
+    std::vector<std::string> env_wrw_xdlops = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlops"};
+
+    std::string v           = " --verbose";
+    std::string dis_bk_data = " --disable-backward-data";
+    std::string dis_bk_wei  = " --disable-backward-weights";
+    std::string dis_fwd     = " --disable-forward";
```
CAHEK7: Can we make it `std::string_view` or at least `const`?

test/gtest/conv_igemm_dynamic_xdlops_float.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+
+    std::vector<std::string> env_bwd = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicBwdXdlops"};
+
+    std::string v           = " --verbose";
+    std::string dis_bk_data = " --disable-backward-data";
+    std::string dis_bk_wei  = " --disable-backward-weights";
+    std::string dis_fwd     = " --disable-forward";
```
CAHEK7: Can we make it `std::string_view` or at least `const`?

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
```diff
@@ -0,0 +1,163 @@
...
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicFwdXdlops"};
+
+    std::vector<std::string> env_wrw = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlops"};
+
+    std::string v           = " --verbose";
+    std::string dis_bk_data = " --disable-backward-data";
+    std::string dis_bk_wei  = " --disable-backward-weights";
+    std::string dis_fwd     = " --disable-forward";
```
CAHEK7: Can we make it `std::string_view` or at least `const`?

test/gtest/conv_igemm_dynamic_xdlops_nhwc_bf16.cpp
```diff
@@ -0,0 +1,220 @@
...
+    std::string in_nhwc       = " --in_layout NHWC";
+    std::string fil_nhwc      = " --fil_layout NHWC";
+    std::string out_nhwc      = " --out_layout NHWC";
+    std::string args_nhwc_wrw = dis_fwd + dis_bk_data + in_nhwc + fil_nhwc + out_nhwc;
```
CAHEK7: It looks like the other directions and the other tests may get benefits if all the common option would be combined like this wrw case.
For example:
```cpp
const std::string args_nhwc_fwd = dis_bk_data + dis_bk_wei + in_nhwc + fil_nhwc + out_nhwc;
const std::string args_nhwc_bwd = dis_fwd + dis_bk_wei + in_nhwc + fil_nhwc + out_nhwc;
```

It's just a suggestion since I've noticed that's already done for wrw case. Probably it's better to keep it as it is because it's more convenient to modify the options individually.

CAHEK7:
I put some comments and suggestions, but all of them are not so critical.

test/gtest/conv_igemm_dynamic_xdlops.cpp
xinlipn: > Why do we need this?

@CAHEK7 Thanks for the comments. This is to keep the strings the same across different gTests. I guess we can consolidate them

test/gtest/conv_igemm_dynamic_xdlops.cpp
xinlipn: > Can we make it `std::string_view` or at least `const`?

@CAHEK7 , `std::string_view` or at least `const` is a good idea :)

CAHEK7:
There is also a discussion regarding to the way of including headers: https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2397#discussion_r1369276573

xinlipn:
> There is also a discussion regarding to the way of including headers: [#2397 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2397#discussion_r1369276573)

@CAHEK7 cleaned up and re-exam'd header based on the discussion in 2397

#include <miopen/miopen.h> // A file that is not locally relative, included from another library; use the <> form
#include <gtest/gtest.h> // A file that is not locally relative, included from another library; use the <> form
#include "../conv2d.hpp"  // A file locally relative to foo.cpp in the same project, use the "" form

test/gtest/conv_igemm_dynamic_xdlops.cpp
xinlipn: > Can we make it `std::string_view` or at least `const`?

std::string_view would give an error. Updated with const

/MIOpen/test/gtest/conv_igemm_dynamic_xdlops.cpp:180:40: error: invalid operands to binary expression ('const std::string' (aka 'const basic_string<char>') and 'std::string_view' (aka 'basic_string_view<char>'))
    TestCase{env_bwd_xdlops, precision + v + " --input  400  256 1 1 --weights 1024  256  1 1 --pads_strides_dilations 0 0 1 1 1 1" + dis_fwd + dis_bk_wei},

test/gtest/conv_igemm_dynamic_xdlops_float.cpp
xinlipn: > Can we make it `std::string_view` or at least `const`?

Done

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
xinlipn: > Can we make it `std::string_view` or at least `const`?

Updated with const, please see comment above

test/gtest/conv_igemm_dynamic_xdlops_nhwc_bf16.cpp
xinlipn: > Why do we need this?

This is address below

test/gtest/conv_igemm_dynamic_xdlops.cpp
CAHEK7: Yes, `std::string_view` is immutable and does not support altering operations.
And it seems `std::string` only supports '.append' method to add `stding_view`s.
So either it should be rewritten to adapt this behavior or we can use `const std::string`.

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
xinlipn: @CAHEK7 , could you "Resolve conversation" if the changes look ok?

test/gtest/conv_igemm_dynamic_xdlops.cpp
xinlipn: > So either it should be rewritten to adapt this behavior or we can use `const std::string`.

@CAHEK7 , I had updated with const std::string.

CAHEK7:
Is there anyone else who's familiar with the test infrastructure?
I understand the intension to minimize the work and transfer strings from CMAKE to gtest's C++. But also I'd prefer to minimize the code footprint and copy-pasted parts.
So from my point of view there is a room for improvement like putting all the variable parts of command line strings into separated vectors, combining all the constant parts into one string and constructing all the test cases dynamically instead of copy-pasting whole lines. Partially this work has already done for some of the tests (https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2409#discussion_r1362176428) and I wish it expanded to all the tests and cases.

I can't explicitly approve or disapprove this PR, because that's only my vision and I'm not the main user of the tests and probably you feel like it's more convenient and obvious way to add the tests cases.
That's why I'm asking some else to take a look into this PR.

xinlipn:
Previous test_conv_igemm_dynamic_xdlops_float failed on MI200. It should have been fixed, as shown in the log below on MI200 node, 10.7.47.119

[test_conv_igemm_dynamic_xdlops_float.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/13221174/test_conv_igemm_dynamic_xdlops_float.txt)

xinlipn:
> Is there anyone else who's familiar with the test infrastructure? I understand the intension to minimize the work and transfer strings from CMAKE to gtest's C++. But also I'd prefer to minimize the code footprint and copy-pasted parts. So from my point of view there is a room for improvement like putting all the variable parts of command line strings into separated vectors, combining all the constant parts into one string and constructing all the test cases dynamically instead of copy-pasting whole lines. Partially this work has already done for some of the tests ([#2409 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2409#discussion_r1362176428)) and I wish it expanded to all the tests and cases.
> 
> I can't explicitly approve or disapprove this PR, because that's only my vision and I'm not the main user of the tests and probably you feel like it's more convenient and obvious way to add the tests cases. That's why I'm asking some else to take a look into this PR.



> Is there anyone else who's familiar with the test infrastructure? I understand the intension to minimize the work and transfer strings from CMAKE to gtest's C++. But also I'd prefer to minimize the code footprint and copy-pasted parts. So from my point of view there is a room for improvement like putting all the variable parts of command line strings into separated vectors, combining all the constant parts into one string and constructing all the test cases dynamically instead of copy-pasting whole lines. Partially this work has already done for some of the tests ([#2409 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2409#discussion_r1362176428)) and I wish it expanded to all the tests and cases.
> 
> I can't explicitly approve or disapprove this PR, because that's only my vision and I'm not the main user of the tests and probably you feel like it's more convenient and obvious way to add the tests cases. That's why I'm asking some else to take a look into this PR.

@CAHEK7 , thanks for the input. I agree with your comments. Some gTests have been refactored with your idea, I'll list one here after find one example. Because of the number of tests, we'd like to have all converted first with current infrastructure to wrap this task up then update them with newer infra if that makes sense

test/CMakeLists.txt
```diff
@@ -29,9 +29,9 @@ cmake_policy(SET CMP0057 NEW)
 set(test_perf.py ${CMAKE_INSTALL_BINDIR}/test_perf.py)
 
 if( NOT ENABLE_ASAN_PACKAGING )
-  install(FILES  test_perf.py
-      PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
-      DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(FILES  test_perf.py
+    PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
+    DESTINATION ${CMAKE_INSTALL_BINDIR})
```
atamazov: Please revert whitespace-only changes everywhere.

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,221 @@
...
+        test_drive<conv2d_driver>(ptrs.size(), ptrs.data());
+        auto capture = testing::internal::GetCapturedStderr();
+        std::cout << capture;
+    }
+};
+
+bool IsTestSupportedForDevice(const miopen::Handle& handle)
+{
+    std::string devName = handle.GetDeviceName();
+    if(devName == "gfx908" || miopen::StartsWith(devName, "gfx94"))
```
atamazov: ~The comment about WORKAROUND_ISSUE_1187 is silently removed which does not comply with #1685.~

test/CMakeLists.txt
```diff
@@ -1236,305 +1217,6 @@ add_custom_test(test_conv_igemm_dynamic_xdlops_bwd SKIP_UNLESS_ALL HALF_ENABLED
...
 #)
 
-# gfx90a is disabled due to WORKAROUND_ISSUE_1187
-add_custom_test(test_conv_igemm_dynamic_xdlops_bwd_float SKIP_UNLESS_ALL HALF_DISABLED FLOAT_ENABLED GFX90A_DISABLED GFX900_DISABLED GFX906_DISABLED SKIP_XNACK_ON
-    ENVIRONMENT ${DYNAMIC_IMPLICITGEMM_BWD_ENVS_XDLOPS}
-    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --verbose --input  4  512 128 128 --weights 12  512  1 1 --pads_strides_dilations 0 0 1 1 1 1 --disable-forward --disable-backward-weights
-)
-
-# gfx90a is disabled due to WORKAROUND_ISSUE_1187
-# Be careful to add testings for (x=1, y=1, c % 8 != 0) due to WORKAROUND_SWDEV_306318
```
atamazov, line 1060: Please do not quietly remove this informative comment.

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+using TestCase = std::tuple<std::vector<std::string>, std::string>;
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_ALL)
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_GPU_XNACK_ENABLED)
+
+static bool SkipTest(void)
+{
+    return miopen::IsEnabled(MIOPEN_TEST_GPU_XNACK_ENABLED{}) ||
```
atamazov: The new test does not comply with #1843

atamazov (changes requested):

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: Well, actually the original comment (in CmakeLists.txt) was incorrect.

These solvers are enabled for gfx908 target only (so the test should fail with gfx94x).
```suggestion
    if(devName == "gfx908")
```

test/gtest/conv_igemm_dynamic_xdlops_float.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+using TestCase = std::tuple<std::vector<std::string>, std::string>;
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_ALL)
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_GPU_XNACK_ENABLED)
+
+static bool SkipTest(void)
+{
+    return miopen::IsEnabled(MIOPEN_TEST_GPU_XNACK_ENABLED{}) ||
```
atamazov: ~The test should be enable for `float` datatype only (like the original; I do not know why)~

This test is originated from some incorrect regression test, see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/1068/files#r696038412. Let's enable it for both half and float: move this config to test/gtest/conv_igemm_dynamic_xdlops.cpp and get rid of this file.

atamazov (changes requested):

atamazov:
@xinlipn 

> Previous test_conv_igemm_dynamic_xdlops_float failed on MI200. It should have been fixed, as shown in the log below on MI200 node, 10.7.47.119
> 
> [test_conv_igemm_dynamic_xdlops_float.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/13221174/test_conv_igemm_dynamic_xdlops_float.txt)

This test should not pas on MI200 because it is targeted to the ConvAsmImplicitGemmGTCDynamicBwdXdlops solver, which is not applicable for this GPU. So something is wrong. And indeed, in the test:

```C++
std::vector<std::string> env_bwd = {
        "MIOPEN_FIND_MODE=normal",
        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicBwdXdlops"};
...
TestCase{env_bwd, ... };
```

In the log:
```bash
...
MIOpen(HIP): Info [FindConvBwdDataAlgorithm] BWD Chosen Algorithm: ConvHipImplicitGemmBwdDataV1R1 , 0, 0.140159
```
🔴 THIS IS WRONG. It seems like this `env_bwd... TestCase{env_bwd, ...}` stuff is not working as expected at all. I am wondering if the gtests that leverage this technique ever worked (and **_if developers of tests carefully check logs_**).

atamazov:
@JehandadKhan @junliume FYI my previous comment. Sorry.

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
```diff
@@ -0,0 +1,160 @@
...
+        test_drive<conv2d_driver>(ptrs.size(), ptrs.data());
+        auto capture = testing::internal::GetCapturedStderr();
+        std::cout << capture;
+    }
+};
+
+bool IsTestSupportedForDevice(const miopen::Handle& handle)
+{
+    std::string devName = handle.GetDeviceName();
+    if(devName == "gfx908" || devName == "gfx90a" || miopen::StartsWith(devName, "gfx94"))
```
atamazov: Only gfx908

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
```diff
@@ -0,0 +1,160 @@
...
+std::vector<TestCase> GetTestCases(const std::string& precision)
+{
+
+    std::vector<std::string> env_fwd = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicFwdXdlops"};
+
+    std::vector<std::string> env_wrw = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlops"};
```
atamazov: This seems not working as expected, see https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2409#issuecomment-1788093364

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+using TestCase = std::tuple<std::vector<std::string>, std::string>;
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_ALL)
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_GPU_XNACK_ENABLED)
+
+static bool SkipTest(void)
+{
+    return miopen::IsEnabled(MIOPEN_TEST_GPU_XNACK_ENABLED{}) ||
```
atamazov: This should comply with #1843

test/gtest/conv_igemm_dynamic_xdlops_nhwc_bf16.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+using TestCase = std::tuple<std::vector<std::string>, std::string>;
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_ALL)
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_GPU_XNACK_ENABLED)
+
+static bool SkipTest(void)
+{
+    return miopen::IsEnabled(MIOPEN_TEST_GPU_XNACK_ENABLED{}) ||
```
atamazov: This should comply with https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1843

test/gtest/conv_igemm_dynamic_xdlops_nhwc_nchw.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+using TestCase = std::tuple<std::vector<std::string>, std::string>;
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_ALL)
+
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_TEST_GPU_XNACK_ENABLED)
+
+static bool SkipTest(void)
+{
+    return miopen::IsEnabled(MIOPEN_TEST_GPU_XNACK_ENABLED{}) ||
```
atamazov: This should comply with https://github.com/ROCmSoftwarePlatform/MIOpen/issues/1843

test/gtest/conv_igemm_dynamic_xdlops_nhwc_nchw.cpp
```diff
@@ -0,0 +1,316 @@
...
+        test_drive<conv2d_driver>(ptrs.size(), ptrs.data());
+        auto capture = testing::internal::GetCapturedStderr();
+        std::cout << capture;
+    }
+};
+
+bool IsTestSupportedForDevice(const miopen::Handle& handle)
+{
+    std::string devName = handle.GetDeviceName();
+    if(devName == "gfx908" || devName == "gfx90a" || miopen::StartsWith(devName, "gfx94"))
```
atamazov, line 124: [Notice] 👍

atamazov (changes requested):

atamazov:
@xinlipn 
> Will add a few more tests into this same PR

Can you please open a new one for the new tests? Adding more tests here might confuse me as a reviewer. Thanks.

atamazov:
@JehandadKhan @xinlipn @CAHEK7 @junliume 
The likely reason of the problem https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2409#issuecomment-1788093364 is use of putenv. It flaws ([see here why](https://stackoverflow.com/questions/5873029/questions-about-putenv-and-setenv)).

test/CMakeLists.txt
atamazov, line 1: All the comments of the form
```
# gfx90a is disabled due to WORKAROUND_ISSUE_1187
```
in this file are incorrect and should be removed. @xinlipn I would appreciate if you do this right in this PR. Thanks.

xinlipn:
> @xinlipn
> 
> > Previous test_conv_igemm_dynamic_xdlops_float failed on MI200. It should have been fixed, as shown in the log below on MI200 node, 10.7.47.119
> > [test_conv_igemm_dynamic_xdlops_float.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/13221174/test_conv_igemm_dynamic_xdlops_float.txt)
> 
> This test should not pas on MI200 because it is targeted to the ConvAsmImplicitGemmGTCDynamicBwdXdlops solver, which is not applicable for this GPU. So something is wrong. And indeed, in the test:
> 
> ```c++
> std::vector<std::string> env_bwd = {
>         "MIOPEN_FIND_MODE=normal",
>         "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicBwdXdlops"};
> ...
> TestCase{env_bwd, ... };
> ```
> 
> In the log:
> 
> ```shell
> MIOpen(HIP): Info [GetFindModeValueImpl] MIOPEN_FIND_MODE = DYNAMIC_HYBRID(5)
> ...
> MIOpen(HIP): Info [FindConvBwdDataAlgorithm] BWD Chosen Algorithm: ConvHipImplicitGemmBwdDataV1R1 , 0, 0.140159
> ```
> 
> 🔴 THIS IS WRONG. It seems like this `env_bwd... TestCase{env_bwd, ...}` stuff is not working as expected at all. I am wondering if the gtests that leverage this technique ever worked (and **_if developers of tests carefully check logs_**).

@atamazov converting to death test for each test case is one possible way of fixing this issue. That way env var will be reset for each test case

atamazov:
@xinlipn The "death test" (i.e. aborting the test right after the first failure) won't help because the reason is that environment setting is not working at all. putenv should be replaced by setenv.

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,221 @@
...
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include <tuple>
+#include <miopen/miopen.h>
+#include <gtest/gtest.h>
+#include "../conv2d.hpp"
+#include "get_handle.hpp"
+
+using TestCase = std::tuple<std::vector<std::string>, std::string>;
```
atamazov:
```suggestion
using TestCase = std::string;
```

The reason is that the library reads each environment variable only once (and caches theirs values in static objects). This is important for speed as environment accesses are slow.

Therefore, once the first TestCase has executed, there is no guarantee that further changes of environment variables will be seen by the library. Thus, it does not make sense to include a different set of environment variables in each TestCase. The environment should be set only once per each gtest executable.

UPDATE: It is actually possible to set different environment in each TEST (and gtest executable may contain several tests) by converting the test to "death tests" (https://github.com/google/googletest/blob/main/docs/advanced.md#death-tests) but I am not sure if it's worth doing provided the increased complexity and known caveats of death tests.

@junliume @JehandadKhan We need to inspect existing gtests and, possibly, make similar changes there. Please let me know if you wish me to open a ticket and/or make this work myself (however the latter may be difficult for me if updated tests start to fail on MI100/200/300)

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,221 @@
...
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicBwdXdlops"};
+
+    std::vector<std::string> env_fwd_xdlops = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicFwdXdlops"};
+
+    std::vector<std::string> env_wrw_xdlops = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlops"};
```
atamazov: Set in the environment once:
```
MIOPEN_FIND_MODE=normal
MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicBwdXdlops;ConvAsmImplicitGemmGTCDynamicFwdXdlops;ConvAsmImplicitGemmGTCDynamicWrwXdlops
```
 Please see my previous comment.

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
atamazov: Please see my previous comment (set env once, use `MIOPEN_DEBUG_FIND_ONLY_SOLVER` with semicolon to specify more than one solver)

test/gtest/conv_igemm_dynamic_xdlops_nhwc_bf16.cpp
```diff
@@ -0,0 +1,217 @@
...
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC"};
+
+    std::vector<std::string> env_nhwc_bwd = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC"};
+
+    std::vector<std::string> env_nhwc_wrw = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC"};
```
atamazov: Please see my prev comment.

test/gtest/conv_igemm_dynamic_xdlops_nhwc_nchw.cpp
```diff
@@ -0,0 +1,316 @@
...
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC"};
+
+    std::vector<std::string> env_nhwc_bwd = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC"};
+
+    std::vector<std::string> env_nhwc_wrw = {
+        "MIOPEN_FIND_MODE=normal",
+        "MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC"};
```
atamazov: Ditto.

atamazov (changes requested):
It needs to be taken into account that env variables are cached in the library.

atamazov:
@xinlipn Note that conversion of _each TestCase_ to "death test" allows to re-read env-vars, but this is not supported by googletest framework AFAICS.

https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2409#discussion_r1379307515 updated.

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,221 @@
...
+        GetArgs(test_value, tokens);
+        std::vector<const char*> ptrs;
+
+        std::transform(tokens.begin(),
+                       tokens.end(),
+                       std::back_inserter(ptrs),
+                       [](const std::string& str) { return str.data(); });
+
+        testing::internal::CaptureStderr();
+        test_drive<conv2d_driver>(ptrs.size(), ptrs.data());
```
cderb: test_drive expects the first char* in the argv to be the program name, but it looks like there is no program name in the test cases, or inserted at any later point. This may dump the first argument.

test/gtest/conv_igemm_dynamic_xdlops.cpp
xinlipn: @atamazov , I have tried the death test, with environment variable set to static const in the APIs in src/include/miopen/env.hpp, which can't be changed on the fly. Thus death test doesn't quite meet our need to update the environment variables.

This PR is to address this issue:
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2514

atamazov: @xinlipn I do not like "death test" approach as well. But there are some other means to run each test as in a separate process. And #2514 is not free. Please see https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2521#issuecomment-1819467431 for more info.

atamazov: @xinlipn 
> I have tried the death test,

Thanks for evaluating it.

test/CMakeLists.txt
xinlipn: @atamazov a new PR is created with the "# gfx90a is disabled due to WORKAROUND_ISSUE_1187" removed. Please kindly help review. Thanks
https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2551

atamazov: [resolved]
@xinlipn Thank you, #2551 approved.

test/CMakeLists.txt
xinlipn: @atamazov  seems something wrong with the text editor. Reverted white space changes.

test/CMakeLists.txt
xinlipn: > Please do not quietly remove this informative comment.

@atamazov , added back both to CMakeLists.txt and conv_igemm_dynamic_xdlops.cpp

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ab4ab6948211b6513c982d264e27052977494f70/test/CMakeLists.txt#L1217

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ab4ab6948211b6513c982d264e27052977494f70/test/gtest/conv_igemm_dynamic_xdlops.cpp#L174

test/CMakeLists.txt
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [resolved]

test/CMakeLists.txt
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: Unresolved yet

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops_half.cpp
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops_nhwc_bf16.cpp
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops_nhwc_nchw.cpp
atamazov: [resolved]

atamazov:
@xinlipn `putenv` should not be used. Please use new machinery from #2514

test/gtest/conv_igemm_dynamic_xdlops_float.cpp
xinlipn: @atamazov , done

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,212 @@
...
+ *
+ *******************************************************************************/
+#include <tuple>
+#include <miopen/miopen.h>
+#include <gtest/gtest.h>
+#include "../conv2d.hpp"
+#include "get_handle.hpp"
+
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_ALL)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_GPU_XNACK_ENABLED)
```
cderb: Perhaps this env should be passed in the gtest CMakeLists?

cderb (approved):

test/gtest/conv_igemm_dynamic_xdlops.cpp
xinlipn: > Perhaps this env should be passed in the gtest CMakeLists?

hi @cderb , I have updated gtest/CMakeLists.txt with MIOPEN_TEST_GPU_XNACK_ENABLED. Thanks

CAHEK7:
:cry:
I believe we can go through all the permanent merge conflicts and finally merge all of the gtest PRs.
Later we must revisit everything one more time.

junliume:
> 😢 I believe we can go through all the permanent merge conflicts and finally merge all of the gtest PRs. Later we must revisit everything one more time.

@CAHEK7 Sorry for the constant conflicts :( Now the purpose of merging "[gtest] Combine gtests into single binary. (#2599)" is to avoid such possible conflicts in the future. I see there are issues with #2599 while hopefully we can fix them instead of backing it out.

junliume (approved):

test/gtest/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ function(add_gtest TEST_NAME TEST_CPP)
   if(NOT WIN32) # TODO: cannot run on Windows due to missing DLL dependencies
     # Enable CMake to discover the test binary
     # Extend GTest DISCOVERY_TIMEOUT to 5 mins
-    gtest_discover_tests(${TEST_NAME} DISCOVERY_TIMEOUT 300 PROPERTIES ENVIRONMENT "MIOPEN_USER_DB_PATH=${CMAKE_CURRENT_BINARY_DIR};MIOPEN_TEST_FLOAT_ARG=${MIOPEN_TEST_FLOAT_ARG};MIOPEN_TEST_ALL=${MIOPEN_TEST_ALL};MIOPEN_TEST_MLIR=${MIOPEN_TEST_MLIR};MIOPEN_TEST_COMPOSABLEKERNEL=${MIOPEN_TEST_COMPOSABLEKERNEL};CODECOV_TEST=${CODECOV_TEST};MIOPEN_TEST_DBSYNC=${MIOPEN_TEST_DBSYNC}")
+    gtest_discover_tests(${TEST_NAME} DISCOVERY_TIMEOUT 300 PROPERTIES ENVIRONMENT "MIOPEN_USER_DB_PATH=${CMAKE_CURRENT_BINARY_DIR};MIOPEN_TEST_FLOAT_ARG=${MIOPEN_TEST_FLOAT_ARG};MIOPEN_TEST_ALL=${MIOPEN_TEST_ALL};MIOPEN_TEST_MLIR=${MIOPEN_TEST_MLIR};MIOPEN_TEST_COMPOSABLEKERNEL=${MIOPEN_TEST_COMPOSABLEKERNEL};CODECOV_TEST=${CODECOV_TEST};MIOPEN_TEST_DBSYNC=${MIOPEN_TEST_DBSYNC};MIOPEN_FIND_MODE=${MIOPEN_FIND_MODE};MIOPEN_DEBUG_FIND_ONLY_SOLVER=${MIOPEN_DEBUG_FIND_ONLY_SOLVER};MIOPEN_TEST_GPU_XNACK_ENABLED=${MIOPEN_TEST_GPU_XNACK_ENABLED}")
```
atamazov: This must be reverted. The test-specific and hardware-specific parameters should not be passed from CMake. See also #2646.

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,217 @@
...
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_FIND_MODE)
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_DEBUG_FIND_ONLY_SOLVER)
+
+namespace conv_igemm_dynamic_xdlops {
+
+static bool SkipTest(void)
+{
+    return miopen::IsEnabled(ENV(MIOPEN_TEST_GPU_XNACK_ENABLED)) ||
```
atamazov: This variable should not be used, see #2646.

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,217 @@
...
+    return test_cases;
+}
+
+} // namespace conv_igemm_dynamic_xdlops
+using namespace conv_igemm_dynamic_xdlops;
+
+TEST_P(Conv2dFloat, FloatTest_Conv_Igemm_Dynamic_dlops)
+{
+    const auto& handle = get_handle();
+    if(IsTestSupportedForDevice(handle) && !SkipTest())
```
atamazov: Please check also the value of the `MIOPEN_TEST_FLOAT_ARG`.

Also the test should not be skipped when run standalone (i.e. when both `MIOPEN_TEST_ALL` and `MIOPEN_TEST_FLOAT_ARG` are not set).

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,217 @@
...
+    else
+    {
+        GTEST_SKIP();
+    }
+};
+
+TEST_P(Conv2dHalf, HalfTest_conv_igemm_dynamic_xdlops)
+{
+    const auto& handle = get_handle();
+    if(IsTestSupportedForDevice(handle) && !SkipTest())
```
atamazov: See previous comment

test/CMakeLists.txt
```diff
@@ -1184,24 +1184,7 @@ set(DYNAMIC_IMPLICITGEMM_XDLOPS_NHWC_BWD_ENVS
...
-    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --verbose --input  2  256 12 18 --weights 256  256  3 3 --pads_strides_dilations 1 1 1 1 1 1 --disable-forward --disable-backward-weights
-    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --verbose --input  400  256 7 7 --weights 1024  256  7 7 --pads_strides_dilations 0 0 1 1 1 1 --disable-forward --disable-backward-weights
-    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --verbose --input  400  256 1 1 --weights 1024  256  1 1 --pads_strides_dilations 0 0 1 1 1 1 --disable-forward --disable-backward-weights
-    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --verbose --input  8  16 5 5 --weights 8  16  2 2 --pads_strides_dilations 0 0 1 1 1 1 --disable-forward --disable-backward-weights
-# ho=wo=1 stride=2
-    COMMAND $<TARGET_FILE:test_conv2d> ${MIOPEN_TEST_FLOAT_ARG} --verbose --input  256 2048 2 2 --weights 1024  2048  1 1 --pads_strides_dilations 0 0 2 2 1 1 --disable-forward --disable-backward-weights
-)
-
+# gfx90a is disabled due to WORKAROUND_ISSUE_1187
```
atamazov: Why is this comment necessary?

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,217 @@
...
+                       [](const std::string& str) { return str.data(); });
+
+        testing::internal::CaptureStderr();
+        test_drive<conv2d_driver>(ptrs.size(), ptrs.data());
+        auto capture = testing::internal::GetCapturedStderr();
+        std::cout << capture;
+    }
+};
+
+bool IsTestSupportedForDevice(const miopen::Handle& handle)
```
atamazov: [tip] Find out and check the XNACK target feature in this function.

atamazov (changes requested):
The same comments apply to other new gtest sources.

test/gtest/CMakeLists.txt
CAHEK7: That's kind of important thing. It looks like we've missed few important architectural decisions and discussions and have already done some hardly maintainable things.
This env-var overusing is one of it - there is no common approach and no documentation and still no way to perfectly fit the same functionality from ctest.

xinlipn: > This must be reverted. The test-specific and hardware-specific parameters should not be passed from CMake. See also #2646.

@atamazov @CAHEK7 , my understanding is we first keep the environment variables combining with gTest for now to wrap up gTest conversion then removing the environment variables afterwards.

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: resolved here, but other .cpp in this PR require the same check instead of environment variable

test/CMakeLists.txt
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,226 @@
...
+}
+
+} // namespace conv_igemm_dynamic_xdlops
+using namespace conv_igemm_dynamic_xdlops;
+
+TEST_P(Conv2dFloat, FloatTest_Conv_Igemm_Dynamic_dlops)
+{
+    const auto& handle = get_handle();
+    if(IsTestSupportedForDevice(handle) && !SkipTest() &&
+       miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG)) == "--float")
```
atamazov: This prevents running the test in standalone mode. 
```suggestion
    if(IsTestSupportedForDevice(handle) && !SkipTest("--float"))
```
Also see next comment.

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+namespace conv_igemm_dynamic_xdlops {
+
+static bool SkipTest(void)
+{
+    return !(miopen::IsUnset(ENV(MIOPEN_TEST_ALL)) &&
+             miopen::IsUnset(ENV(MIOPEN_TEST_FLOAT_ARG))) ||
+           miopen::IsEnabled(ENV(MIOPEN_TEST_GPU_XNACK_ENABLED)) ||
+           miopen::IsDisabled(ENV(MIOPEN_TEST_ALL));
+}
```
atamazov: - The logic is a bit incorrect.
- XNACK env var should not be used
- "wanted" datatype should be analyzed here
```suggestion
static bool SkipTest(const std::string& float_arg)
{
    if(miopen::IsUnset(ENV(MIOPEN_TEST_ALL)))
        return false;
    if(miopen::IsEnabled(ENV(MIOPEN_TEST_ALL)))
        if(miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG)) == float_arg)
            return false;
    return true;
}
```

atamazov (changes requested):
## 🔴 The same comments apply to other .cpp in this PR

test/gtest/CMakeLists.txt
atamazov: not yet resolved

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [unresolved, deprecated by the next review, can be marked as resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [unresolved, deprecated by the next review, can be marked as resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [unresolved, deprecated by the next review, can be marked as resolved]

junliume:
@xinlipn  @JehandadKhan  could you help to resolve the conflicts? This PR is almost ready. Thanks!

atamazov:
@junliume 
> This PR is almost ready. 

Maybe I am wrong, but unfortunately this is not exactly so. These must be resolved prior merging:
- https://github.com/ROCm/MIOpen/pull/2409#discussion_r1438907931
- https://github.com/ROCm/MIOpen/pull/2409#pullrequestreview-1868366932

test/gtest/CMakeLists.txt
xinlipn: Updated XNACK query with API for all .CPP files

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -0,0 +1,225 @@
...
+        if(miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG)) == float_arg)
+            return false;
+    return true;
+}
+
+void SetupEnvVar(void)
+{
+    miopen::UpdateEnvVar(ENV(MIOPEN_FIND_MODE), std::string("normal"));
+    miopen::UpdateEnvVar(
+        ENV(MIOPEN_DEBUG_FIND_ONLY_SOLVER),
```
atamazov: These two variables should be unset after test complete. Otherwise next tests will misbehave, most likely.

**_In all tests, please._**

atamazov (changes requested):

test/gtest/CMakeLists.txt
atamazov: [quality] Please revert this change, it is unnecessary.

atamazov (changes requested):

test/gtest/CMakeLists.txt
atamazov: [can be marked as resolved] Continued in other review thread.

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [resolved]

test/gtest/conv_igemm_dynamic_xdlops.cpp
atamazov: [resolved]
- This is bigger problem, #2749 opened.

atamazov:
🟡 This may be merged, but I would prefer fixing https://github.com/ROCm/MIOpen/pull/2409#discussion_r1486277895 first.

junliume:
@xinlipn almost ready but please resolve the conflicts. Thanks!

xinlipn:
> @xinlipn almost ready but please resolve the conflicts. Thanks!

@junliume , checked in changes for fixing conflicts.

test/gtest/conv_igemm_dynamic_xdlops.cpp
```diff
@@ -53,74 +53,82 @@ auto GetTestCases()
...
+    std::pair{env_xdlops, cmd_v + "  --input  64   64 56 56 --weights 256  64  1 1 --pads_strides_dilations 0 0 1 1 1 1" + dis_fwd + dis_bk_data},
+    std::pair{env_xdlops, cmd_v + "  --input  64  224 17 17 --weights 224  224  1 7 --pads_strides_dilations 0 3 1 1 1 1" + dis_fwd + dis_bk_data},
+    std::pair{env_xdlops, cmd_v + "  --input  128  128 35 35 --weights 256  128  3 3 --pads_strides_dilations 1 1 1 1 1 1" + dis_fwd + dis_bk_data},
+    std::pair{env_xdlops, cmd_v + "  --input  128  128 64 64 --weights 256  128  3 3 --pads_strides_dilations 1 1 2 2 1 1" + dis_fwd + dis_bk_data},
+    std::pair{env_xdlops, cmd_v + "  --input  128  768 17 17 --weights 256  768  3 3 --pads_strides_dilations 1 1 1 1 2 2" + dis_fwd + dis_bk_data},
+    std::pair{env_xdlops, cmd_v + "  --input  3  256 28 28 --weights 80  256  1 1 --pads_strides_dilations 0 0 1 1 1 1" + dis_fwd + dis_bk_data},
+    std::pair{env_xdlops, cmd_v + "  --input  2  256 12 18 --weights 256  256  3 3 --pads_strides_dilations 1 1 1 1 1 1" + dis_fwd + dis_bk_data},
+    std::pair{env_xdlops, cmd_v + "  --input  4  512 128 128 --weights 12  512  1 1 --pads_strides_dilations 0 0 1 1 1 1" + dis_fwd + dis_bk_data},
```
atamazov, line 95: 🐛 This should be enabled for HALF && Backward Data at least (see https://github.com/ROCm/MIOpen/pull/1068/files/c2233e741717e2f25782e096262272c98975c6d4..7a8a5a6461196d3453c883bb0ca53f65e3bfbf83#r696038412 for more info).

PR #2413, opened 2023-09-25
Add typecast to config key

cderb:
Add typecast value to config key. This prevents aliasing for fp16 and fp16 typecast f8 configs, etc.
Add typecast as optional arg to fdb_key


DrizztDoUrden (dismissed):

atamazov:
@cderb Thanks. This is one thing missing in #2251 I wanted to fix.

src/conv/problem_description.cpp
```diff
@@ -176,6 +198,17 @@ void ProblemDescription::Serialize(std::ostream& stream) const
...
+        if(GetInCastType() || GetWeightsCastType() || GetOutCastType())
+        {
+            std::tuple<miopenDataType_t, std::optional<miopenDataType_t>> in_cast =
+                std::make_tuple(GetInDataType(), GetInCastType());
+            std::tuple<miopenDataType_t, std::optional<miopenDataType_t>> weights_cast =
+                std::make_tuple(GetWeightsDataType(), GetWeightsCastType());
+            std::tuple<miopenDataType_t, std::optional<miopenDataType_t>> out_cast =
+                std::make_tuple(GetOutDataType(), GetOutCastType());
+            optional << "c" << EncodeDataTypesForKey(in_cast, weights_cast, out_cast);
+        }
```
atamazov: [R] Save space in db key and optimize code. Do not print casting value when casting is not actually necessary. Get rid of ~crutches~ tuples ;)
```suggestion
        if(const auto ct = GetInCastType())
            optional << "_ci" << GetDataTypeName(*ct);
        if(const auto ct = GetWeightsCastType())
            optional << "_cw" << GetDataTypeName(*ct);
        if(const auto ct = GetOutCastType())
            optional << "_co" << GetDataTypeName(*ct);
```

src/conv/problem_description.cpp
```diff
@@ -127,6 +139,16 @@ void ProblemDescription::BuildConfKey(std::string& conf_key) const
         ss << 'x' << GetOutLayout();
     }
     ss << 'x' << EncodeDataTypesForKey(GetInDataType(), GetWeightsDataType(), GetOutDataType());
+    if(GetInCastType() || GetWeightsCastType() || GetOutCastType())
```
atamazov: Similar to previous comment

src/conv/problem_description.cpp
```diff
@@ -176,6 +198,17 @@ void ProblemDescription::Serialize(std::ostream& stream) const
         // Group count > 1 identifies Group/Depthwise modes.
         if(GetGroupCount() != 1)
             optional << 'g' << GetGroupCount();
```
atamazov: Readability of db key: Remove `'_' <<` from line 216 and:
```suggestion
            optional << '_g' << GetGroupCount();
```

src/conv/problem_description.cpp
```diff
@@ -44,6 +44,18 @@ EncodeDataTypesForKey(miopenDataType_t in, miopenDataType_t weights, miopenDataT
...
+                      std::tuple<miopenDataType_t, std::optional<miopenDataType_t>> weights,
+                      std::tuple<miopenDataType_t, std::optional<miopenDataType_t>> out)
+{
+    miopenDataType_t cin, cweights, cout;
+    cin      = std::get<1>(in) ? std::get<1>(in).value() : std::get<0>(in);
+    cweights = std::get<1>(weights) ? std::get<1>(weights).value() : std::get<0>(weights);
+    cout     = std::get<1>(out) ? std::get<1>(out).value() : std::get<0>(out);
+    return EncodeDataTypesForKey(cin, cweights, cout);
+}
+
```
atamazov: [R]
```suggestion
std::string
EncodeDataTypesForKey(std::tuple<miopenDataType_t, std::optional<miopenDataType_t>> in,
                      std::tuple<miopenDataType_t, std::optional<miopenDataType_t>> weights,
                      std::tuple<miopenDataType_t, std::optional<miopenDataType_t>> out)
{
    miopenDataType_t cin, cweights, cout;
    cin      = std::get<1>(in) ? std::get<1>(in).value() : std::get<0>(in);
    cweights = std::get<1>(weights) ? std::get<1>(weights).value() : std::get<0>(weights);
    cout     = std::get<1>(out) ? std::get<1>(out).value() : std::get<0>(out);
    return EncodeDataTypesForKey(cin, cweights, cout);
}

```

atamazov (changes requested):

atamazov:
@cderb Also I highly recommend changing "BFP8" to "BF8" in `GetDataTypeName()` at once.

atamazov:
[Attribution] @junliume @JehandadKhan The lib may select wrong Invoker without this PR.

atamazov:
@JehandadKhan @cderb 🔴 It seems like we have to update the SQL queries with casting attributes as well:
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/6d539ee81321121570606e4ef62e6d072775bbd8/src/include/miopen/conv/problem_description.hpp#L437-L444

This is not urgent for now, because existing solvers that can perform tensor casting are not tunable, but the value is potentially very high.

src/conv/problem_description.cpp
JehandadKhan: This makes sense, we should minimize the changes we make to DBs

JehandadKhan:
> @JehandadKhan @cderb 🔴 It seems like we have to update the SQL queries with casting attributes as well:
> 
> https://github.com/ROCmSoftwarePlatform/MIOpen/blob/6d539ee81321121570606e4ef62e6d072775bbd8/src/include/miopen/conv/problem_description.hpp#L437-L444
> 
> This is not urgent for now, because existing solvers that can perform tensor casting are not tunable, but the value is potentially very high.

That makes sense, however, I would rather change the SQLite DB to use the problem key directly instead of the columns stuff we have right now. Perhaps, in a separate PR we can change that and unify this logic.

atamazov:
@JehandadKhan Shall I open a ticket for SQLite DB update?

JehandadKhan:
> @JehandadKhan Shall I open a ticket for SQLite DB update?

Always a good idea!

src/conv/problem_description.cpp
cderb: reduced cast format to 
x\<ciF16>\<cwF16>\<coF16>
where each <> is optional, labels remain in case not all 3 are present
the phrase with leading x isn't included if there are no casts

src/conv/problem_description.cpp
cderb: updated

src/conv/problem_description.cpp
cderb: removed

junliume (approved):

junliume:
@atamazov @DrizztDoUrden @JehandadKhan  last round call of pressing reviews before merging :) Anything not mentioned as blocking will need to go into post merge issues.

src/conv/problem_description.cpp
atamazov: [resolved]

src/conv/problem_description.cpp
atamazov: [resolved]

src/conv/problem_description.cpp
atamazov: [resolved]

atamazov (commented):
LGTM

PR #2414, opened 2023-09-25
[Bug Fixes] miopen_rocblas_gemm_ex3 call - invoker cache extra elements - conv direct naive input cast

cderb:
miopen_rocblas_gemm_ex3 call would always throw error 
invoker cache adding extra elements
conv direct naive yielding incorrect input cast for kernel arg


junliume:
@cderb bug fixes automatically gets blocker priority :) Meanwhile, could you update the description a bit? If issue is already created, please link to those issues too. Thanks!

cderb:
@junliume I think these wouldn't have issues yet
the rocblas gemm call came after some of the f8 changes (recent error)
the invoker cache issue is a double insert, which is just extra computation
conv direct naive cast would only be apparent when casting and input + ouput types are different

src/invoker_cache.cpp
```diff
@@ -104,8 +104,11 @@ void InvokerCache::Register(const Key& key, const Invoker& invoker)
     auto it = invokers.find(key.first);
     if(it != invokers.end())
         it->second.invokers.insert({key.second, invoker});
-    auto& item = invokers.insert({key.first, Item{}}).first->second;
-    item.invokers.insert({key.second, invoker});
+    else
```
atamazov, line 107: This must be approved by @DrizztDoUrden before merging.

atamazov:
@junliume 
> bug fixes automatically gets blocker priority :)

Well, actually no, for ex. consider a bug that is very difficult to fix or reproduce and has never been mentioned by our users or CI.

DrizztDoUrden (approved):
LGTM
Invoker cache bug is a great catch! That stuff was done in a bit of a hassle, and it's very nice to see people actually taking a look inside.

src/invoker_cache.cpp
```diff
@@ -104,8 +104,11 @@ void InvokerCache::Register(const Key& key, const Invoker& invoker)
...
+    {
+        auto& item = invokers.insert({key.first, Item{}}).first->second;
```
atamazov, line 109: Note: `insert` does nothing if the container already contains an element with an equivalent key, so this is not a bug and the library does not spend any extra time except the check.

atamazov (approved):
LGTM

junliume (approved):

PR #2417, opened 2023-09-26
[HotFix] zstd dependency on multi Linux distributes

junliume:
Observation:
On `RHEL` and `SLES`, we observe the following issues:
```
CMake Error at src/CMakeLists.txt:687 (target_link_libraries):
  Target "MIOpen" links to:

    zstd::libzstd_static

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```


junliume:
@apwojcik https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2362 is reverted in mainline, and this PR is an attempt to fix the issue for any Linux platforms aside from Ubuntu.
CC: @JehandadKhan @atamazov

atamazov:
@junliume 
> @apwojcik #2362 is reverted in mainline, and this PR is an attempt to fix the issue for any Linux platforms aside from Ubuntu. 

AFAIU making changes directly in `amd-master` breaks smoothness of promotions. I am afraid this PR won't merge into mainline cleanly. In order to fix the problem, I guess, we need to
- cherry-pick c2c8b41e0d7e494437d1cc35ee90a1291a959225 back into develop (thus reverting #2362 in develop), then
- restore #2362 in develop by reverting commit from previous bullet, then
- apply this fix (just merge fresh develop in this PR and it is ready)

Or maybe it is possible to revert c2c8b41e0d7e494437d1cc35ee90a1291a959225 right in mainline just prior merging from `amd-develop` that contains commit from this PR, but I am not sure.

WRT this PR I do not have any objections.

/cc @JehandadKhan

JehandadKhan (approved):

apwojcik:
Have you tried to install the `libzstd-devel-static` package for `RHEL` and `SLES`?

apwojcik (approved):

junliume:
> libzstd-devel-static

@apwojcik which one should we install? On the other hand, the change made in this PR seems to pass cmake stage, does it meet your original requirement?
We reverted in mainline to unblock daily builds, we can fix the issue properly in develop and let it propagate to mainline.
```
bash-4.4# yum install -y libzstd-static
Updating Subscription Management repositories.
Unable to read consumer identity
Subscription Manager is operating in container mode.

This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 12:40:02 ago on Tue Sep 26 05:52:34 2023.
No match for argument: libzstd-static
Error: Unable to find a match: libzstd-static


bash-4.4# yum install -y libzstd-devel-static
Updating Subscription Management repositories.
Unable to read consumer identity
Subscription Manager is operating in container mode.

This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 12:40:10 ago on Tue Sep 26 05:52:34 2023.
No match for argument: libzstd-devel-static
Error: Unable to find a match: libzstd-devel-static

```

PR #2418, opened 2023-09-26
3D forward convolution solver with non-packed input tensors

iq136boy:
Note: This PR builds on top of #2394, which itself is based on top of #2334, so the diff set appears bigger than what it is. 

Enable 3D forward convolution solver with non-packed input tensors using CK kernels.
The "make check" tests passed on mi200 development machine.  
PR current blocked by MIOpen CI.


junliume:
@iq136boy please resolve the conflict by merging the latest `develop` into your feature branch. Thanks!

iq136boy:
> @iq136boy please resolve the conflict by merging the latest `develop` into your feature branch. Thanks!

@junliume conflict resolved. Let me merge the latest develop branch.

CAHEK7:
It looks good, but it seems it has #2394 and #2334 included. Should it be merged sequentially?

iq136boy:
> It looks good, but it seems it has #2394 and #2334 included. Should it be merged sequentially?

Not necessarily. I have merged PR [#2334](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2334). So this PR can be merged before #2334 if it pass the MIOpen CI to unblock the rocm6.0 p1 gating.

test/gtest/nonpack_conv3d_fwd.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include "conv3d_test_case.hpp"
+
+struct NonPackTestCase : Conv3DTestCase
+{
+    size_t s0;
+    size_t s1;
+    size_t s2;
+    size_t s3;
+    size_t s4;
+    std::vector<size_t> GetInputStrides() { return {s0, s1, s2, s3, s4}; }
```
amberhassaan: @iq136boy : to be more thorough, you should also pass strides for `Weight` and `Output` tensors. One way to generate correct strides is to multiply `packed` strides with a small constant. E.g., you have packed dims as `{N, C, D, H, W}`, you can produce non-packed strides as `{2 * C* D* H * W, 2 * D * H * W, 2 * H * W, 2 * W, 2 * 1}`. 

You can also take it one step further and multiply a prefix of strides by a constant, e.g., (for the case above):
`{2 * C* D* H * W, 2 * D * H * W, 2 * H * W,  W, 1}` is a valid non-packed strides value where only a prefix of strides are non-packed and remainder is packed.

src/solver/conv_hip_implicit_gemm_3d_fwd_nonpack_xdlops.cpp
```diff
@@ -0,0 +1,373 @@
+/*******************************************************************************
```
amberhassaan: @iq136boy : Instead of adding a new solver for non-packed case, why not modify the existing 3d grouped solvers for `fwd,bwd,wrw` to use `strides` like you're doing below. That way we get all 3 for both packed and non-packed cases. CC: @JehandadKhan, do you agree?

JehandadKhan: Yeah I agree, with the new solver just for non-packed tensors would result in a lot of duplicate code. More importantly, the underlying kernels are the same since they dont have any assumptions no the tensor strides.

test/gtest/nonpack_conv3d_fwd.hpp
iq136boy: @amberhassaan I agreed that we should also put the strides for weight and output. I also agree that there are more ways that the non-packed tensors can be generated. However, the user should give the strides of nonpacked tensors anyway. What we do need is just to check if the tensor strides are valid or not.

src/solver/conv_hip_implicit_gemm_3d_fwd_nonpack_xdlops.cpp
iq136boy: > @iq136boy : Instead of adding a new solver for non-packed case, why not modify the existing 3d grouped solvers for `fwd,bwd,wrw` to use `strides` like you're doing below. That way we get all 3 for both packed and non-packed cases. CC: @JehandadKhan, do you agree?

@amberhassaan I agree with it. Currently working on it.

test/gtest/nonpack_conv3d_fwd.hpp
amberhassaan: Yes, but these are tests you are writing and we need to test if convolution works for non-packed inputs. If we don't test a feature then we can't offer a guarantee that it works.

iq136boy: > Yes, but these are tests you are writing and we need to test if convolution works for non-packed inputs. If we don't test a feature then we can't offer a guarantee that it works.

Oh, I see. You mean add more test cases. That make sense.

junliume:
@CAHEK7 @averinevg @DrizztDoUrden @atamazov @JehandadKhan @amberhassaan  :
CI has passed, last call for any blocking reviews before merging.

CAHEK7 (dismissed):
Seems fine, but conflicting with #2394 and I'm not sure which one has to be merged first)
Looks like this one since it includes both.

iq136boy:
> Seems fine, but conflicting with #2394 and I'm not sure which one has to be merged first) Looks like this one since it includes both.

@CAHEK7 Yes. This one will merge first because it include the the content of #2394

iq136boy:
@junliume Please merge this PR before #2394.

amberhassaan:
> @junliume Please merge this PR before #2394.

@iq136boy : that's the wrong order of merging. This PR was based on top of #2394, so we merge #2394 first and then this.

atamazov:
@junliume @iq136boy 
> Note: This PR builds on top of #2394, which itself is based on top of #2334, so the diff set appears bigger than what it is.

>> @junliume Please merge this PR before #2394.

I do not understand why. According to the PR description this should be merged AFTER #2394.

Moreover, the fix for the critical problems of #2334 (listed at #2440) must be merged before #2394.

Proposed order of actions:
- PR with fixes for critical problems from [#2440](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2440) created, reviewed, tested and merged in develop
- develop shall be merged into [#2394](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2394), then the latter to be finally reviewed, tested etc then merged into develop
- develop shall be merged here, then this PR is ready for final review etc.

iq136boy:
> > @junliume Please merge this PR before #2394.
> 
> @iq136boy : that's the wrong order of merging. This PR was based on top of #2394, so we merge #2394 first and then this.

@amberhassaan Yes. If we have enough time, the order should be #2394 first and then this one. But this PR is a p1 gating issue for rocm6.0 and today is the last day of branching date if I am correct. And the content of this PR should go in mainline. What I suggest and am doing now is that I create a PR that merge latest #2394 and let that PR run through MIOpen CI separately. If that PR run through the CI without any issue before the branching cutting date, then we can merge 2394 first. Otherwise, I still suggest that this PR should be merged first to unblock the mainline gating status. 
@junliume @JehandadKhan what do you think?

iq136boy:
@atamazov Yes. The procedure that you mentioned is correct if we do not have a urgent hard deadline today. However, the MIOpen CI issue blocks the PRs for a couple days and caused some delays. So I proposed a solution [here](https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2418#issuecomment-1755696196). Generally, I would like to merge 2394 into this PR and let it run through MIOpen CI separately. If something happened, then this PR should be a backup plan to unblock the rocm6.0 p1 gating.

iq136boy:
@junliume @amberhassaan @CAHEK7 Well, it seems there is NO conflict between this PR and #2394. I merged #2394 with this PR and there is NO merge conflict. Also, I tested it on MI200 machine development machine locally and all tests passed after the merge. So we are good to merge the two PRs no matter the orders.

atamazov:
@iq136boy I understand your motives. However, I advise to never cut corners unless there is an **_explicit and specific request from management_**. From personal experience, I see that cutting corners to save time often leads to problems that then require more time, efforts, and stress to resolve than doing things the right way. 🤝

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
```diff
@@ -314,7 +320,7 @@ bool ConvHipImplicitGemm3DGroupFwdXdlops::IsApplicable(
     if(!problem.IsLayoutNHWC())
         return false;
     const std::string& arch = ctx.GetStream().GetDeviceName();
-    if(!(arch == "gfx908" || arch == "gfx90a"))
+    if(miopen::StartsWith(arch, "gfx11") || miopen::StartsWith(arch, "gfx10"))
```
JehandadKhan: I would suggest we convert this to a white list, so, list the arches that CK supports.

bghimireamd: something similar to https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/include/miopen/solver/ck_utility_common.hpp#L49

amberhassaan: Good suggestion and should be used in all CK-based solvers.

iq136boy: > I would suggest we convert this to a white list, so, list the arches that CK supports.

Great idea. I have applied the change to all ck based solvers and currently doing the local test before push the changes.

src/include/miopen/solver/ck_utility_common.hpp
```diff
@@ -63,6 +63,16 @@ static inline bool is_ck_supported_hardware(const Handle& handle)
...
 }
 
+static inline bool is_conv_ck_supported_hardware(const std::string& device_name, bool is_wrw)
+{
+    auto res_wrw = StartsWith(device_name, "gfx908") || StartsWith(device_name, "gfx90a") ||
+                   StartsWith(device_name, "gfx940") || StartsWith(device_name, "gfx941") ||
+                   StartsWith(device_name, "gfx942");
+    return is_wrw ? res_wrw
+                  : (res_wrw || StartsWith(device_name, "gfx900") ||
+                     StartsWith(device_name, "gfx906"));
```
bghimireamd, line 73: This is just a recommendation : place `StartsWith(device_name, "gfx900") || StartsWith(device_name, "gfx906")` in sperate bool function. That way we don't need to pass `is_wrw`.

junliume (approved):

test/gtest/nonpack_conv3d_fwd.hpp
amberhassaan: @iq136boy : please create an issue for this item. I can take it up. Basically, I want to generate non-packed strides for all 3 of input, output and weights. CC: @JehandadKhan

PR #2421, opened 2023-09-27
TunaNet Integration: MI250x

msaudulhassan:
Enable AI Based Heuristics for Immediate Mode Fallback on MI250x GPUs.


src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -145,6 +145,47 @@ class Model
     virtual std::vector<float> ToFeatures(const ProblemDescription& problem) const = 0;
 };
 
+class Gfx90aModel : public Model
```
CAHEK7: Just a minor question - should it be `final`?
I guess yes and I guess all of the other models derived from `Model` should be final as well.

src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -145,6 +145,47 @@ class Model
...
+{
+public:
+    Gfx90aModel() : Model("gfx90a") {}
+    bool IsProblemSupported(const ProblemDescription& problem,
+                            const ConvolutionContext& ctx) const override
```
CAHEK7:
```suggestion
    bool IsProblemSupported(const ProblemDescription&,
                            const ConvolutionContext&) const override
```
I guess HIP Tidy complains about unused parameters.

src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -255,7 +296,12 @@ class Gfx908Model : public Model
     }
 };
 
-std::unique_ptr<Model> GetModel(const std::string&) { return std::make_unique<Gfx908Model>(); }
+std::unique_ptr<Model> GetModel(const std::string& device) {
+    if (device == "gfx90a")
+        return std::make_unique<Gfx90aModel>();
+    else
+        return std::make_unique<Gfx908Model>();
```
CAHEK7: What if `device` is not a `gfx90a` and not a `gfx908`? Should it be handled somehow?

src/conv/heuristics/ai_heuristics.cpp
msaudulhassan: Good point. I'll mark it final.

src/conv/heuristics/ai_heuristics.cpp
msaudulhassan: They are going to be used, I just haven't added the code yet. This PR is in draft.

src/conv/heuristics/ai_heuristics.cpp
msaudulhassan: I tried to maintain the previous behavior, but on going through the code again, I think it was a bug. I have fixed it to return a nullpointer if the device is not gfx90a or gfx908

src/conv/heuristics/ai_heuristics.cpp
msaudulhassan: @CAHEK7 Done

src/conv/heuristics/ai_heuristics.cpp
msaudulhassan: @CAHEK7 Done. I now return a nullptr if the device is not supported

src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -255,7 +255,108 @@ class Gfx908Model : public Model
...
+        {
+            auto solver_id = solver::Id{solver_name.second};
+            auto solver    = solver_id.GetSolver();
+            if(solver.IsApplicable(ctx, problem))
+            {
+                applicable_solvers++;
+                break;
+            }
+        }
+        if(applicable_solvers == 0)
```
CAHEK7, line 310: Probably it's a bit wordy, but from a high-level point of view if show that we want to find a sovler which `IsApplicable`.

```suggestion
        if(std::find_if(metadata.solver_map.begin(),
                        metadata.solver_map.end(),
                        [&ctx, &problem](const auto& solver_name)
                        {
                            return solver::Id{solver_name.second}.GetSolver().IsApplicable(ctx, problem);
                        }) == metadata.solver_map.end())
```

CAHEK7 (approved):

Dmantri98 (approved):

junliume (approved):
Ping @JehandadKhan for last round of review

src/conv/heuristics/ai_heuristics.cpp
msaudulhassan: @CAHEK7 I have revered this back after discussing with JD. If we have a model specifically trained for a given GPU, we use it. If we don't, we use the model for MI100. The assumption is that the fastest solver remains the same no matter the GPU (this is actually not necessarily the case, but we still want to have a heuristic even if we do not have a specially designed model).

PR #2424, opened 2023-09-30
upgrade cmake so that MIOpen docker can compile Composable Kernel

bghimireamd:
* Currently if you build Composable Kernel (CK) in MIOpen's docker it fails to compile. This is because the CMake version of MIOpen's docker is old. This PR will upgrade the CMake in Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -105,8 +105,9 @@ RUN ccache -s
 # Install doc requirements
 ADD docs/.sphinx/requirements.txt /doc-requirements.txt
 RUN pip3 install -r /doc-requirements.txt
-# Composable Kernel requires upgraded cmake
-RUN pip3 install --upgrade cmake
+
+# Composable Kernel requires this version cmake
+RUN pip3 install --upgrade cmake==3.27.5
 
 # Use parallel job to accelerate tensile build
 # Workaround for Tensile with TargetID feature
--- a/Dockerfile
+++ b/Dockerfile
@@ -105,6 +105,8 @@ RUN ccache -s
 # Install doc requirements
 ADD docs/.sphinx/requirements.txt /doc-requirements.txt
 RUN pip3 install -r /doc-requirements.txt
+# Composable Kernel requires upgraded cmake
+RUN pip3 install --upgrade cmake
 
 # Use parallel job to accelerate tensile build
 # Workaround for Tensile with TargetID feature


Dockerfile
```diff
@@ -105,6 +105,8 @@ RUN ccache -s
 # Install doc requirements
 ADD docs/.sphinx/requirements.txt /doc-requirements.txt
 RUN pip3 install -r /doc-requirements.txt
+# Composable Kernel requires upgraded cmake
+RUN pip3 install --upgrade cmake
```
pfultz2: You should pin the cmake version, so the builds are reproducible.

bghimireamd: Good catch. Now cmake version is pinned to 3.27.5.

illsilin (approved):

Dockerfile
junliume: Hi @bghimireamd I assume you are referring to issue resolved by this PR https://github.com/ROCmSoftwarePlatform/composable_kernel/pull/940

```
cmake_minimum_required(VERSION 3.14)
if(POLICY CMP0140)
  # policies CMP0140 not known to CMake until 3.25
  cmake_policy(SET CMP0140 NEW)
endif()
```
So with this fix, actually upgrading cmake version will not be necessary.

bghimireamd: @junliume 

MIOpen docker   : `rocm/miopen:ci_686727` 
MIOpen docker's cmake version : `3.16.3`
CK  `develop`'s commit hash   : `bd09b5c5389c9f7f51459e5a38d3d7253b0c8dc0`
When building CK on MIOpen docker I still get linking **error** :/

Below I have new rebuilt docker, new cmake but CK's commit hash is same.

MIOpen docker   : `rocm/miopen:ci_ca3793` 
MIOpen docker's cmake version : `3.27.5`
CK  `develop`'s commit hash   : `bd09b5c5389c9f7f51459e5a38d3d7253b0c8dc0`
When building CK on MIOpen docker **it works** :)

illsilin: Have you rebuilt the docker? Did the cmake version get upgraded? Can you share the link to the logs?

bghimireamd: @illsilin I updated my comment. Sorry for confusion. When we rebuilt the docker with this PR which updates the CMake version to 3.27.5 CK builds works. Its just with old version cmake even though with latest CK the build fails.

illsilin: ah, ok. false alarm.

junliume (approved):

PR #2425, opened 2023-09-30
regression: do not use file system symbolic/hard links

apwojcik:
Windows does not fully support symbolic links. It has some support but is not on by default and requires extra steps to enable it. Please do not commit symbolic links. Otherwise, you are breaking Windows.


JehandadKhan (approved):

junliume:
Updated feature branch with the latest `develop`, will watch the last round of CI before merging.

atamazov:
> Windows does not fully support symbolic links. It has some support but is not on by default and requires extra steps to enable it. Please do not commit symbolic links. Otherwise, you are breaking Windows.

It also breaks some cross-development scenarios (e.g. development on windows via samba on linux machine)

junliume (approved):

PR #2429, opened 2023-10-02
Update CK-based 2d/3d  convolution solvers to support  nchw/ncdhw layout

amberhassaan:
Using transpose kernel to transform NCDHW to NDHWC on the fly and then invoke CK kernels to do the convolution and then transform back. 
Group convolution solvers for both 2D and 3D types supported. Fwd, Bwd data and Bwd weights (Wrw) passes supported.


src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
```diff
@@ -311,7 +311,7 @@ bool ConvHipImplicitGemm3DGroupFwdXdlops::IsApplicable(
         return false;
     if(!problem.Is3d())
         return false;
-    if(!problem.IsLayoutNHWC())
+    if(!(problem.IsLayoutNHWC() || problem.IsLayoutDefault()))
```
CAHEK7, line 346: Does it mean that we do not support default layout?

amberhassaan: Yes, CK supports only NHWC type layouts. The irony :).

amberhassaan:
@junliume , @JehandadKhan : 3D solvers are working with NCDHW layout. As I was developing the feature, I realized 2D case can be supported with little extra effort, so I went ahead and did that too. I have modified  existing grouped-conv tests to start exercising this feature. Tests are not comprehensive, so I'll see if I can add more values for testing.

junliume:
@amberhassaan thanks! Some CI issues which we observed:
```
/var/jenkins/workspace/s_MIOpen_amber_layout-xform-conv/test/gpu_nchw_nhwc_transpose.cpp:29:10: fatal error: 'miopen/util_sol.hpp' file not found

#include <miopen/util_sol.hpp>

         ^~~~~~~~~~~~~~~~~~~~~
```

test/conv_common.hpp
```diff
@@ -2631,3 +2631,27 @@ struct conv_bias_driver : test_driver
...
+                               const miopen::TensorDescriptor& y)
+{
+
+    constexpr size_t alignment = 256u;
+
+    auto align = [&](size_t sz) {
+        auto ret = (sz + alignment - 1) & ~(alignment - 1);
+        assert(ret >= sz);
+        return ret;
+    };
```
CAHEK7: [Minor]
```suggestion
    constexpr size_t alignment = 256u;

    auto align = [](size_t sz) {
        static_assert(alignment > 0);
        constexpr size_t alignment_mask = alignment - 1;
        static_assert(alignment & alignment_mask  == 0);
        return (sz + alignment_mask ) & ~(alignment_mask );
    };
```
Runtime assertion is not required if `alignment` is not zero (which does not make sense in any way).
But `alignment` must be a power of 2, so few static assertion added.

test/conv_common.hpp
```diff
@@ -2631,3 +2631,27 @@ struct conv_bias_driver : test_driver
...
+
+    auto w_sz = align(x.GetNumBytes()) + align(w.GetNumBytes()) + align(y.GetNumBytes());
+    std::vector<std::uint8_t> w_cpu(w_sz);
+
+    auto w_gpu = handle.Write(w_cpu);
+    assert(w_gpu.get());
+
+    return std::make_pair(std::move(w_gpu), w_cpu.size());
```
CAHEK7: [Minor]
```suggestion
    auto res = std::pair(handle.Write(w_cpu), w_cpu.size());
    assert(res.first.get());

    return res;
```
A bit simplified version and it uses RVO instead of moving.

test/gtest/group_conv3d_fwd.cpp
```diff
@@ -32,7 +32,8 @@
 #include "get_handle.hpp"
 #include "group_conv3d_fwd.hpp"
 
-struct ConvFwdSolverTest3D : ConvFwdSolverTest<half_float::half>
+// struct ConvFwdSolverTest3D : ConvFwdSolverTest<half_float::half>
+struct ConvFwdSolverTest3D : ConvFwdSolverTest<float>
```
CAHEK7: [question]
```suggestion
struct ConvFwdSolverTest3D : ConvFwdSolverTest<float>
```
Do we need this comment?

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -157,5 +160,406 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
     return result;
 }
 
+struct TransposeInstance
```
CAHEK7: Can it be a `class` and do not expose all of the internals outside?

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -157,5 +160,406 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+
+private:
+    void Run(const Handle& handle,
+             const std::vector<Kernel>& kernels,
+             Data_t out_ptr,
+             ConstData_t in_ptr)
+    {
+        assert(out_ptr);
+        assert(in_ptr);
+        assert(kernels.size() > index);
```
CAHEK7: It triggers `signed vs unsigned comparison` warning.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -157,5 +160,406 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
     return result;
 }
 
+struct TransposeInstance
+{
+    size_t tensor_sz = 0;
+    std::vector<OpKernelArg> kern_args{};
+    int index         = -1;
```
CAHEK7: This one is important.
```suggestion
    size_t index         = std::numeric_limits<size_t>::max();
```
It has `unsigned` nature and will be properly compared with `unsigned` `kernels.size()`.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -157,5 +160,406 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+    size_t buf_offset = 0;
+    shared<Data_t> buf_handle{};
+
+    template <typename TransSolnType>
+    TransposeInstance(const TransSolnType& trans_sol, int idx, const MultiBufferWorkspaceTraits& wt)
+    {
+        tensor_sz = trans_sol.GetOutputTensorSize();
+        kern_args = trans_sol.GetKernelArg();
+        index     = idx;
+        assert(index >= 0);
```
CAHEK7: [Minor]
This assertion is redundant.
`wt` performs all the necessary checks for itself and later `index` checked for `kernels.size()`.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -157,5 +160,406 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+    using TrInstMaker =
+        std::conditional_t<ND == 2,
+                           TransposeInstanceMaker2D<TransposeSolutionDefault2Nhwc,
+                                                    TransposeSolutionDefault2Nhwc,
+                                                    TransposeSolutionNhwc2Default>,
+                           TransposeInstanceMaker3D<TransposeSolutionDefault2Ndhwc,
+                                                    TransposeSolutionDefault2Ndhwc,
+                                                    TransposeSolutionNdhwc2Default>>;
+
+    TrInstMaker tr_inst_maker{};
```
CAHEK7: This instance is not required. You can pass `TrInstMaker` type directly to `InitInvokerFactoryNCHW`.
Later you may create an "instance" in-place inside the function:
```CPP
auto [_tr_inst_in, _tr_inst_wei, _tr_inst_out] = TpostInstMaker()(result, ctx, problem, ck_args);
```
or you can make `operator()` `static` (and rename it) and call it directly from the type
```CPP
auto [_tr_inst_in, _tr_inst_wei, _tr_inst_out] = TpostInstMaker::make(result, ctx, problem, ck_args);
```

CAHEK7 (changes requested):
Most of the changes are not so critical except `singend` vs `unsigned` comparison.

But I'd redesigned all the stuff related to `TransposeInstance` and `TransposeInstanceMaker*D` classes. They are kind of too complex in terms of usage and still have a lot of duplicated code.
But I don't have exact solution or exact suggestions right now.

Also I'd like to talk about all of the assertions. I think some of them should be removed and some of the should be replaced with `MIOPEN_THROW`. `assert` is not a very reliable functionality.

src/include/miopen/batched_transpose_sol.hpp
```diff
@@ -66,6 +66,64 @@ struct BatchedTransposeSolution
...
+{
+    TransposeSolutionDefault2Nhwc(const ExecutionContext& ctx_,
+                                  miopenDataType_t data_type_,
+                                  uint32_t n_,
+                                  uint32_t c_,
+                                  uint32_t h_,
+                                  uint32_t w_)
+        : BatchedTransposeSolution(ctx_, data_type_, n_, c_, h_ * w_)
+    {
+        assert(size_t(h_ * w_) == (size_t(h_) * size_t(w_)));
```
JehandadKhan: Please replace the `assert`s with `MIOPEN_THROW`

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
```diff
@@ -64,15 +64,24 @@ struct CKArgs
...
-        K        = ProblemInterpreter::GetOutputChannelK(problem);
-        C        = ProblemInterpreter::GetInputChannelC(problem);
-        input    = {ProblemInterpreter::GetInputHeightHi(problem),
-                 ProblemInterpreter::GetInputWidthWi(problem)};
-        output   = {ProblemInterpreter::GetOutputHeightHo(problem),
-                  ProblemInterpreter::GetOutputWidthWo(problem)};
-        filter   = {ProblemInterpreter::GetFilterHeightY(problem),
-                  ProblemInterpreter::GetFilterWidthX(problem)};
+        G = ProblemInterpreter::GetGroupCountG(problem);
+        assert(G == 1);
```
JehandadKhan: If this is the assumption, why do we need to divide `K` and `C` below?

amberhassaan: I copy pasted this code from grouped solver. Eventually I'd like to combine these CKArgs classes as much as I can.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: I don't like this. Each object (tr_inst_maker, pre_tpose_fn, post_tpose_fn) is passed as a callable object and I don't want to impose a restriction that the function is static or that the object has default constructor.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -157,5 +160,406 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+        constexpr size_t buf_alignment = 256;
+        MultiBufferWorkspaceTraits wt({tr_in.GetOutputTensorSize(),
+                                       tr_wei.GetOutputTensorSize(),
+                                       tr_out.GetOutputTensorSize()},
+                                      buf_alignment);
+
+        return std::make_tuple(TransposeInstance(tr_in, 0, wt),
+                               TransposeInstance(tr_wei, 1, wt),
+                               TransposeInstance(tr_out, 2, wt));
+    }
```
CAHEK7: A little bit of C++
You can add such function
```CPP
template <typename I, typename W, typename O>
std::tuple<TransposeInstance, TransposeInstance, TransposeInstance> tuple_cvt(ConvSolution& result, std::tuple<I, W, O>&& t)
{
    return std::apply([&result](auto&&... args){
        result.construction_params.insert(result.construction_params.end(), {args.GetKernelInfo()...});

        constexpr size_t buf_alignment = 256;
        MultiBufferWorkspaceTraits wt({args.GetOutputTensorSize()...}, buf_alignment);
        size_t n = 0;
        return std::tuple{TransposeInstance(args, n++, wt)...};
    }, t);
}
```
and replace this operator with following code both for 2d and 3d cases (adjusting ck_args sequence accordingly)

```suggestion
    template <typename CKArgsType>
    std::tuple<TransposeInstance, TransposeInstance, TransposeInstance>
    operator()(ConvSolution& result,
               const ExecutionContext& ctx,
               const ProblemDescription& problem,
               const CKArgsType& ck_args) const
    {
        return tuple_cvt(result,
                  InTransSolType(ctx, problem.GetInDataType(), ck_args.N, ck_args.C1, ck_args.Hi, ck_args.Wi),
                  WeiTransSolType(ctx, problem.GetWeightsDataType(), ck_args.K1, ck_args.C, ck_args.Y, ck_args.X),
                  OutTransSolType(ctx, problem.GetOutDataType(), ck_args.N, ck_args.K1, ck_args.Ho, ck_args.Wo));
    }
```

src/include/miopen/batched_transpose_sol.hpp
CAHEK7: I guess not all the asserts, some of them may not be required at all.
That's exactly what I wanted to discuss.

src/include/miopen/solver/implicitgemm_ck_util.hpp
CAHEK7: Ok, It's not so important. I just wanted to avoid passing object at all since we are working with pure types.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
CAHEK7: This assert shall not be required if everything is fine inside `isApplicable` function.

Or do we have to check everything everywhere?

src/include/miopen/solver/implicitgemm_ck_util.hpp
CAHEK7: You can even get rid of `TransposeInstanceMaker2D` and `TransposeInstanceMaker3D` structs, replace it with modified `tuple_cvt` and localize everything inside `ConvSolution InitInvokerFactoryFwdNCHW` function with `if constexpr`. 

It can minimize lots of entities and dependencies, make the code and the meanings more localized and reduce amount of code as well.

test/conv_common.hpp
CAHEK7: And if we can get rid of assertion, we can do `return std::pair(handle.Write(w_cpu), w_cpu.size());` which is guaranteed to be RVO.

src/include/miopen/batched_transpose_sol.hpp
amberhassaan: Do we have an `MIOPEN_ASSERT` that works in Release build as well? If not, I'd like to introduce it for these situations.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: using an `int` here simplifies things. `-1` is a sentinel value. The check later on against `kernel.size()` serves another purpose as well, which is that the vector being passed is of the right size.

src/include/miopen/batched_transpose_sol.hpp
amberhassaan: Also, I don't agree with this opposition to `asserts`. They're free. I'm open to converting some of them to `MIOPEN_THROW` but removing them doesn't help us.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: We can add a `static_cast` here but I'd like to keep the `int` for other reasons.

src/include/miopen/batched_transpose_sol.hpp
CAHEK7:
```CPP
if (size_t(h_ * w_) != (size_t(h_) * size_t(w_)))
    MIOPEN_THROW(...)
```
That sort of problem must be properly detected and reported.

Actually asserts are not free, not safe and almost useless due to Release\Debug builds.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: Many good suggestions. Let me see how I can simplify the code. My end goals are to keep the code generic and simple. So if it's not the most minimal code, I'm OK with it. I don't want to replace more lines with less lines of more complex code. That doesn't serve C++ codebases in the long-term as people of different skill-levels join the project, and most complex code means difficulty to understand the code. 
CC: @JehandadKhan.

src/include/miopen/batched_transpose_sol.hpp
amberhassaan: I don't see how `asserts` are not free. Please help me understand. They are dead code in release builds and the compiler should be able to optimize them out completely.

atamazov:
@amberhassaan @CAHEK7 @JehandadKhan Currently the library employs the following policy w.r.t asserts and MIOPEN_THROW:
- Use `MIOPEN_THROW` in error cases that may happen in real time.
- Use `assert` for errors that should never happen.
- Do not use asserts that work in the release builds.

[Example] If `IsApplicable()` filters out configs that have SOME_CONDITION, then SOME_CONDITION can't happen in `GetSolution()`, so `MIOPEN_THROW` should not be used in the latter method. But it is fine to use `assert(SOME_CONDITION)`. This can be done for documentation purposes, or to detect some theoretically possible elusive errors (so an engineer who stuck into some problem can begin with reproducing the issue with Debug build) and so on.

If asserts in this PR comply to the current policy, then I recommend closing the respective review comments. Specifically this may apply to:
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2429#discussion_r1352630727
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2429#discussion_r1352991781
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2429#discussion_r1355475754
- https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2429#discussion_r1355507141

If you think the current policy should be changed, then please open a discussion targeted to development of the updated policy. Once it is ready, we can apply it everywhere (including reviews).

fin
atamazov, line 1: [Q] Does this PR require changes in fin?

src/include/miopen/batched_transpose_sol.hpp
atamazov: @JehandadKhan @amberhassaan I think that these checks must be put in all the Solvers that leverage `TransposeSolutionNhwc2Default` or `TransposeSolutionDefault2Nhwc`. Specifically, theirs `IsApplicable()` must return `false` when the necessary conditions are not met.

Maybe it is better to implement something like `IsApplicable` in `TransposeSolution*` and use it in the Solvers. Because we do not want the implementation details of `TransposeSolution*` to sneak into other parts of code.

Also it may worth making `IsApplicable` also in `BatchedTransposeSolution` (if it has its own implementation limits) and reusing it in `TransposeSolution*`.

WRT the asserts here I do not care.

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: @amberhassaan What are the reasons? Unfortunately, I don't see them.

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: @amberhassaan I am second to @CAHEK7. Let's avoid mixing signed and unsigned types wherever possible -- _that_ simplifies things IMHO. Otherwise the future code readers must always have in mind types of operands and implicit conversion rules (instead of focusing on more important things).

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -157,5 +160,406 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+    TransposeInstance(const TransSolnType& trans_sol, int idx, const MultiBufferWorkspaceTraits& wt)
+    {
+        tensor_sz = trans_sol.GetOutputTensorSize();
+        kern_args = trans_sol.GetKernelArg();
+        index     = idx;
+        assert(index >= 0);
+        buf_offset = wt.GetOffset(index);
+    }
+
+    void AllocBuffer(const Handle& handle, Data_t workSpace)
```
atamazov: [R] "Alloc" implies "Free" but we do not have it. Maybe "GetSubBuffer" or just "SubBuffer"?

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -157,5 +160,406 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+        assert(kernels.size() > index);
+
+        kern_args[0] = out_ptr;
+        kern_args[1] = in_ptr;
+
+        handle.Run(kernels[index])(kern_args);
+        if(handle.IsProfilingEnabled())
+        {
+            handle.AccumKernelTime(handle.GetKernelTime());
+        }
```
atamazov: IIRC This won't work because the legacy internal API for accumulation of elapsed kernel times is cumbersome. Running the kernel overwrites the value stored in the handle by `AccumKernelTime()`, i.e. when you run the 2nd kernel, the elapsed time of the 1st kernel is lost.

## Approach 1: Use variable for accumulation at the invoker level

Invoker:
```C++
float accumulated = 0;
... // Run the 1st kernel, then:
if(handle.IsProfilingEnabled())
    accumulated += handle.GetKernelTime();
...
// A function can be used to run the kernel, but it may not be known
// whether the kernel will be actually run in the called function.
// If so, then ResetKernelTime() before calling it:
handle.ResetKernelTime();
function_that_may_run_a_kernel(...); // Possibly runs the Nth kernel, then:
if(handle.IsProfilingEnabled())
    accumulated += handle.GetKernelTime();
...
// Store result in handle.
if(handle.IsProfilingEnabled())
{
    handle.ResetKernelTime();
    handle.AccumKernelTime(accumulated);
}
```

The above approach is a legacy one, but is currently used throughout the library.

## Approach 2: Use accumulator in handle.

Or you can `ResetKernelTime()` once somewhere in the Invoker, before running the 1st kernel, and then do like this:
```C++
auto save = handle.IsProfilingEnabled() ? 0.0f : handle.GetKernelTime(); // save previously accumulated time
handle.Run(kernels[index])(kern_args);
if(handle.IsProfilingEnabled())
    handle.AccumKernelTime(save); // add prev accumulated time to the execution time of kernels[index].
```
This approach allows using the value stored in the handle as an accumulator.

atamazov (changes requested):

fin
atamazov: [resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
CAHEK7: A bit simplified, but more copy-pasted version.
```CPP
template <typename I, typename W, typename O>
std::tuple<TransposeInstance, TransposeInstance, TransposeInstance> tuple_cvt(ConvSolution& result, I&& i, W&& w, O&& o)
{
        result.construction_params.insert(result.construction_params.end(), {i.GetKernelInfo(), w.GetKernelInfo(), o.GetKernelInfo()});

        constexpr size_t buf_alignment = 256;
        MultiBufferWorkspaceTraits wt({i.GetOutputTensorSize(), w.GetOutputTensorSize(), o.GetOutputTensorSize()}, buf_alignment);
        return {TransposeInstance(i, 0, wt), TransposeInstance(w, 1, wt), TransposeInstance(o, 2, wt)};
}
```

amberhassaan:
@atamazov , @CAHEK7 , @JehandadKhan : please see a redesign of the core layout transformation logic in #2461. I personally like the new code a lot. 
I will address the minor comments around signed/unsigned comparison, asserts-vs-throw etc. here after or in parallel with merging #2461.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: fixed in #2461

atamazov:
@amberhassaan Whenever possible, please avoid situations when a review comment is created in one PR but resolved in another PR. If this is unavoidable, then please please create a review  comment in the second PR and put a link to it into the first PR (as an answer to the original review comment). Thanks.

CAHEK7 (approved):

CAHEK7 (approved):

src/solver/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
```diff
@@ -341,28 +360,29 @@ ConvSolution ConvHipImplicitGemm3DGroupBwdXdlops::GetSolution(
...
-#endif
+    return MakeSolutionGroupConvImplicitGemmXdlops(
+        problem,
+        [&](auto data_type_val) {
+            using T = decltype(data_type_val);
+            /// \todo This call should be InitInvokerFactoryBwdNCHW but due to the
+            /// silliness of "in tensor is out and out is in" in backward pass,
+            /// InitInvokerFactoryFwdNCHW works correct while Bwd call causes wrong
+            /// output -- amberhassaan
+            return InitInvokerFactoryFwdNCHW<3,
```
iq136boy: I agree with the confusion about the tensors in bwd. However, if you switched the input and output tensors, then the bwd should work properly.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: fixed in latest code.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: fixed in latest code.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: fixed in latest code.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: fixed in redesign.

src/include/miopen/batched_transpose_sol.hpp
amberhassaan: fixed in latest code.

src/solver/conv_hip_implicit_gemm_bwd_data_xdlops.cpp
amberhassaan: keeping for now in anticipation that these classes will be combined in the future.

src/solver/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
amberhassaan: Calling the `InitInvokerFactoryFwd...` has the exact same effect.

iq136boy (approved):

src/solver/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
CAHEK7: @iq136boy @amberhassaan as I recall there is a "unified 2d convolution problem" which automatically swaps all the tensors and dimensions for fwd\bwd\wrw passes. But I didn't did deeper and probably it doesn't solve all the problems.

amberhassaan: Good point. I'm planning on doing something similar here, where the `CKArgs` classes are combined into one and some compile time decision making to swap pointers for `bwd` pass.

test/conv_common.hpp
```diff
@@ -2622,3 +2622,27 @@ struct conv_bias_driver : test_driver
         verify(verify_backwards_bias<T>{output, bias});
     }
 };
+
+inline std::pair<miopen::Allocator::ManageDataPtr, size_t>
+AllocateConvTransposeWorkspace(miopen::Handle& handle,
```
atamazov: @amberhassaan What is the purpose of this function?

junliume:
@amberhassaan when you have time could you resolve the conflicts?

amberhassaan:
> @amberhassaan when you have time could you resolve the conflicts?

@junliume : this PR needs more debugging, so I'll do that when I get back. The conflicts are due to my fixes from other PRs, but there are other CI failures that need resolved for `bwd` and `wrw` convolutions. Some would be related to the fixes I made, but more time is needed to debug other ones.

amberhassaan:
@CAHEK7 , @iq136boy , @DrizztDoUrden , @atamazov , @JehandadKhan : Please review the PR again as it was reviewed a long time ago. CI has passed once and should pass again (unless I introduced a bug in last 2 commits).

iq136boy (approved):

JehandadKhan:
@CAHEK7  Can you please take some time to review this PR.

JehandadKhan:
@junliume  I have asked @CAHEK7  to review this PR again, once that happens we can merge this.

src/include/miopen/errors.hpp
```diff
@@ -69,6 +69,18 @@ template <class... Params>
         miopen::MIOpenThrow(__FILE__, __LINE__, __VA_ARGS__); \
     } while(false)
 
+#define MIOPEN_THROW_IF(condition, msg)                                                \
```
atamazov, line 72: 👍

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -28,6 +28,11 @@
 
 #include <miopen/conv/data_invoke_params.hpp>
 #include <miopen/conv/wrw_invoke_params.hpp>
+#include <miopen/batched_transpose_sol.hpp>
+
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
```
atamazov: [Recommendation]
```suggestion
#if MIOPEN_USE_COMPOSABLEKERNEL
```

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+        assert(kernels.size() > kern_idx);
+
+        kern_args[0] = out_ptr;
+        kern_args[1] = in_ptr;
+
+        handle.Run(kernels[kern_idx])(kern_args);
+        if(handle.IsProfilingEnabled())
+        {
+            handle.AccumKernelTime(handle.GetKernelTime());
+        }
```
atamazov: [assumption] AFAICS you are using Approach 1 from https://github.com/ROCm/MIOpen/pull/2429#discussion_r1360795493

What is incorrect: This code stores 2*kernel_time in the handle. Kernel time is stored in the handle automatically, you do not need to accumulate it there manually.

```suggestion
```

See https://github.com/ROCm/MIOpen/pull/2429#discussion_r1360795493 for more info.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+                                 const std::vector<Kernel>& kernels) mutable {
+        return [kernels,
+                ck_args             = std::move(ck_args),
+                sh_conv_ptr         = std::move(sh_conv_ptr),
+                input1_tr_inst      = std::move(input1_tr_inst),
+                input2_tr_inst      = std::move(input2_tr_inst),
+                output_tr_inst      = std::move(output_tr_inst),
+                output_init_tr_inst = std::move(output_init_tr_inst)](
+                   const Handle& handle, const AnyInvokeParams& primitive_parameters) mutable {
+            handle.ResetKernelTime();
```
atamazov, line 583: [Recommendation] This is not needed.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+
+    return std::make_tuple(
+        TransposeInstanceTagged{input1_solver, 0, wt, 0, Input1TposeOp::CONV_OP_TAG},
+        TransposeInstanceTagged{input2_solver, 1, wt, 1, Input2TposeOp::CONV_OP_TAG},
+        TransposeInstanceTagged{output_solver, 2, wt, 2, OutputTposeOp::CONV_OP_TAG},
+        TransposeInstanceTagged{output_init_solver, 3, wt, 2, OutputTposeOp::CONV_OP_TAG});
+}
+
+template <typename V>
+void DebugPrintVec(const char* name, const V& vec)
```
atamazov, line 467: [Recommendation] Enclose the whole debugging-only block of code into
```C++
#if 0 /// This code should not be used in the production.
...
#endif
```
Why: It can be easy for developers to forget to remove calls to `DebugPrintCKArgPtrs()` or `DebugPrintConvTensors()` in the future.

atamazov (changes requested):
Partial review

atamazov:
@amberhassaan Sorry for delaying the review.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+            const_cast<Data_t>(tensors.y)  // NOLINT (cppcoreguidelines-pro-type-const-cast)
+        };
+
+        return data_ptrs[GetConvOperandTagAsInt()];
+    }
+};
+
+/*
+template <typename CKArgsType, typename... TransposeOps, size_t... indices>
+auto MakeTaggedTransposeInstancesHelper(ConvSolution& result,
```
atamazov: [Recommendation] Either remove this or enclose in `#if 0...#endif` and add a comment that explains why we need to keep this.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+
+    auto output_init_solver = OutputInitOp{}.MakeTransposeSolver(ctx, problem, ck_args);
+
+    result.construction_params.insert(result.construction_params.end(),
+                                      {input1_solver.GetKernelInfo(),
+                                       input2_solver.GetKernelInfo(),
+                                       output_solver.GetKernelInfo(),
+                                       output_init_solver.GetKernelInfo()});
+
+    constexpr size_t buf_alignment = 256ull;
```
atamazov: [Recommendation] Get rid of the magic number. It is used also in `GetWorkspaceSizeLayoutTransformConv()` below and maybe somewhere else.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+            {
+                // this is backward pass, swap back input and output
+                std::swap(conv_tensors.x, conv_tensors.y);
+                std::swap(conv_tensors.xDesc, conv_tensors.yDesc);
+            }
+
+            float tot_time = 0;
+
+            input1_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
+            tot_time += handle.GetKernelTime();
```
atamazov:
```suggestion
            if(handle.IsProfilingEnabled())
                tot_time += handle.GetKernelTime();
```

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+
+            input2_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
+            tot_time += handle.GetKernelTime();
```
atamazov: Ditto

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+
+            output_init_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
+            tot_time += handle.GetKernelTime();
```
atamazov: Ditto

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+            auto invoker_ptr  = sh_conv_ptr->MakeInvokerPointer();
+            auto argument_ptr = ck_args.MakeArgPtr(sh_conv_ptr,
+                                                   tr_ptrs[0]->GetBufferPtr(),
+                                                   tr_ptrs[1]->GetBufferPtr(),
+                                                   tr_ptrs[2]->GetBufferPtr());
+            tot_time += invoker_ptr->Run(argument_ptr.get(),
+                                         {handle.GetStream(), handle.IsProfilingEnabled()});
+
+            output_tr_inst.ConvertTo(handle, kernels, conv_tensors);
+            tot_time += handle.GetKernelTime();
```
atamazov: Ditto.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,634 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+}
+
+template <typename InvokerFactoryMakerNCHW, typename InvokerFactoryMakerNHWC>
+ConvSolution
+MakeSolutionGroupConvImplicitGemmXdlops(const miopen::conv::ProblemDescription& problem,
+                                        InvokerFactoryMakerNCHW&& invoker_factory_maker_ncdhw,
+                                        InvokerFactoryMakerNHWC&& invoker_factory_maker_ndhwc)
+{
+
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
```
atamazov: [R]
```suggestion
#if MIOPEN_USE_COMPOSABLEKERNEL
```

atamazov (changes requested):
Very good design, but we still need to fix GPU time measurements.

CAHEK7 (approved):

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: done.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: I've wrapped it in `#ifndef NDEBUG` so that the code still compiles in debug mode but is disabled for release builds.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: forgot to remove this. Will do now.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: done.

amberhassaan:
@atamazov : I have addressed all comments. Please review the handling of kernel time. Thanks.

junliume:
> @atamazov : I have addressed all comments. Please review the handling of kernel time. Thanks.

@amberhassaan some CI tests starts to fail with the latest changes, we should take a look into them.

amberhassaan:
> > @atamazov : I have addressed all comments. Please review the handling of kernel time. Thanks.
> 
> @amberhassaan some CI tests starts to fail with the latest changes, we should take a look into them.

@junliume : yeah Jun, that's what happens if you run CI again on working code ;)

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved]

atamazov:
@amberhassaan Now I see that you would like to use Approach 2 from https://github.com/ROCm/MIOpen/pull/2429#discussion_r1360795493. However, it looks like some fixes are still needed. Reviewing now.

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved] dropped, see other review

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved] dropped, see other review

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved] dropped, see other review

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved] dropped, see other review

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved] dropped, see other review

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved] dropped, see other review

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,584 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+        assert(kernels.size() > kern_idx);
+
+        kern_args[0] = out_ptr;
+        kern_args[1] = in_ptr;
+
+        handle.Run(kernels[kern_idx])(kern_args);
+        if(handle.IsProfilingEnabled())
+        {
+            handle.AccumKernelTime(handle.GetKernelTime());
+        }
```
atamazov: You need to save previously accumulated time (because `handle.Run()` overwrites it):
```suggestion
        auto save = handle.IsProfilingEnabled() ? 0.0f : handle.GetKernelTime();
        handle.Run(kernels[kern_idx])(kern_args);
        if(handle.IsProfilingEnabled())
        {
            handle.AccumKernelTime(save);
        }
```

atamazov (changes requested):
Almost ready

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: So, this has been quite confusing. Partially because I wasn't paying attention to how kernel time was being handled via `Run()`. My question is:
Can we do better and avoid this confusion altogether? For example, why not have `Run()` automatically accumulate kernel times, so an invoker looks like: 
```
handle.ResetKernelTime();
handle.Run();
handle.Run();
...
```
Accumulation happens automatically via `elapsed_time_handler()`  here: https://github.com/ROCm/MIOpen/blob/5c7617c9241272cec6b6ae64fab18a59182f1b46/src/hip/handlehip.cpp#L206 , where all we need to change is update to accumulate. This will require keeping a second counter for `accumulated_time` in addition to `profiling_result`.

atamazov: @amberhassaan 
> So, this has been quite confusing.

+1

> Can we do better and avoid this confusion altogether?

Sure. I've wanted to do this since the first day I saw this ugly API. The one only needs free time and will to redesign this stuff and perform a lot of testing.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,585 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+                                  static_cast<uint32_t>(ck_args.Wo)};
+            }
+            else
+            {
+                return SolverType{ctx,
+                                  problem.GetOutDataType(),
+                                  static_cast<uint32_t>(ck_args.N),
+                                  static_cast<uint32_t>(ck_args.K1),
+                                  static_cast<uint32_t>(ck_args.Ho),
+                                  static_cast<uint32_t>(ck_args.Wo)};
```
atamazov, line 280: ⚓ casting to uint32_t

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
```diff
@@ -323,8 +323,6 @@ bool ConvHipImplicitGemmGroupWrwXdlops::IsApplicable(
         return false;
     if(miopen::IsEnabled(ENV(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC)))
         return false;
-    if(problem.HasAtLeastOne64BitTensor())
-        return false;
```
atamazov: I am afraid we still need this because of https://github.com/ROCm/MIOpen/pull/2429/files#r1481674855

src/solver/conv_hip_implicit_gemm_grouped_bwd_xdlops.cpp
```diff
@@ -327,8 +327,6 @@ bool ConvHipImplicitGemmGroupBwdXdlops::IsApplicable(
         return false;
     if(miopen::IsEnabled(ENV(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC)))
         return false;
-    if(problem.HasAtLeastOne64BitTensor())
-        return false;
```
atamazov: Ditto

atamazov (commented):
Important Q

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
amberhassaan: It got renamed in `develop` and I started getting errors. I left the check with the new name `AllTensorsDimsFitIntoInt`.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: yes, this is a limitation with the current transpose solver and kernel. We can add a TODO or open an issue for this.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: I'll be motivated some day. A frustrated user is a motivated user :).

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: If this is out of scope of this PR, then let's open an issue. Otherwise it is better to fix right here, but I am not insisting.

[Note] `HasAtLeastOne64BitTensor()` has just been renamed to `AllTensorsDimsFitIntoInt()` in #2731, but the constraint is a bit relaxed here (`AllTensorsDimsFitIntoUint32()`?)

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [offtopic] @amberhassaan Haha, exactly! Wait... so that means I'm very motivated? ;))

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
atamazov: @amberhassaan Now I see.
[resolved]

src/solver/conv_hip_implicit_gemm_grouped_bwd_xdlops.cpp
atamazov: [resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved]

atamazov (approved):
🟢 LGTM!

junliume (approved):

junliume:
@amberhassaan  and all, thank you for making this big step in MIOpen happen !

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: https://github.com/ROCm/MIOpen/issues/2735 created.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: #2736 created.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -165,5 +172,585 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
+            input2_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
+
+            output_init_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
+
+            /// \todo: Fix NHWC Wrw invokers to also issue a zero-out kernel. Will
+            /// need SetTensor() to properly zero out non-packed tensors
+            if(output_tr_inst.GetConvOperandTag() == internal::ConvOperandTag::Weights)
+            {
+                output_tr_inst.ZeroOutBuffer();
+            }
```
atamazov, line 620: ⚓

PR #2430, opened 2023-10-02
Add MIOPEN_BETA_API defines around f8

cderb:


include/miopen/miopen.h
```diff
@@ -355,8 +357,10 @@ typedef enum
     miopenBFloat16 = 5, /*!< 16-bit binary floating point (8-bit exponent, 7-bit fraction)
                            (Partially supported) */
     miopenDouble  = 6,  /*!< 64-bit floating point (Partially supported) */
+#ifdef MIOPEN_BETA_API
     miopenFloat8  = 7,
     miopenBFloat8 = 8
+#endif
```
JehandadKhan, line 366: Lets add an #else clause which resreves the enum IDs ( 7 and 8) so that we dont break backwards compatability

include/miopen/miopen.h
```diff
@@ -601,11 +605,13 @@ typedef enum
     MIOPEN_CONVOLUTION_ATTRIB_DETERMINISTIC =
         1, /*!< Restrict MIOpen convolutions to kernels which produce numerically deterministic
               results. 0 - disabled (default), 1 - enabled >*/
+#ifdef MIOPEN_BETA_API
```
JehandadKhan, line 611: Same as above

JehandadKhan (changes requested):

JehandadKhan (dismissed):

junliume (dismissed):

JehandadKhan (dismissed):
Discussed in a meeting, this approach is not ideal since it would require a recompile to enable MIOpen beta features.

CMakeLists.txt
```diff
@@ -68,6 +68,9 @@ include(ROCMCreatePackage)
 include(CheckCXXCompilerFlag)
 include(ROCMHeaderWrapper)
 
+# Build librarie with beta apis
```
JehandadKhan:
```suggestion
# Build library with Beta APIs
```

JehandadKhan (dismissed):

include/miopen/miopen.h
```diff
@@ -354,9 +356,14 @@ typedef enum
         4, /*!< Pack of four 8-bit int points in NCHW_VECT_C format (Partially supported) */
     miopenBFloat16 = 5, /*!< 16-bit binary floating point (8-bit exponent, 7-bit fraction)
                            (Partially supported) */
-    miopenDouble  = 6,  /*!< 64-bit floating point (Partially supported) */
+    miopenDouble = 6,   /*!< 64-bit floating point (Partially supported) */
+#ifdef MIOPEN_BETA_API_FP8
     miopenFloat8  = 7,
-    miopenBFloat8 = 8
+    miopenBFloat8 = 8,
+#else
```
atamazov: I think the library won't compile without MIOPEN_BETA_API_FP8 because `miopenFloat8` and `miopenBfloat8` are used in many places.

atamazov (changes requested):

atamazov:
@cderb The working approach is implicitly described at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#pullrequestreview-1629537076 (which is currently being ignored for some unknown reason).

In short, the experimental feature should be unconditionally enabled in all MIOpen sources. But when the user includes miopen.h, the experimental API should be disabled by default.

include/miopen/miopen.h
```diff
@@ -112,11 +112,13 @@ typedef enum
     miopenStatusVersionMismatch = 10, /*!< Version mismatch of the supplied binary data argment. */
 } miopenStatus_t;
 
+#ifdef MIOPEN_BETA_API_FP8
```
atamazov: @cderb Right now we already have `MIOPEN_BETA_API` which controls LayerNorm. Adding `MIOPEN_BETA_API_FP8` for controlling F8 convolutions seems inconsistent.

Let's either control F8 convolutions with the same `MIOPEN_BETA_API` or do the following:
- Rename existing `MIOPEN_BETA_API` to `MIOPEN_BETA_API_LAYERNORM`
- Rename `MIOPEN_BETA_API_FP8` to `MIOPEN_BETA_API_CONV_F8`. Why:
  - `CONV` because this only controls convolutions
  - `F8` because this controls both FP8 and BF8 datatypes.

Note: Related review thread (unresolved for now) can be found at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1316419544

/cc @JehandadKhan

atamazov (changes requested):

include/miopen/miopen.h
cderb: @atamazov I had meant to use simply MIOPEN_BETA_API and default the CMakeLists.txt to defining this variable thusly unconditionally enabling for all sources, and disabling in miopen.h by default. However the implementation of layernorm appears to rely on MIOPEN_BETA_API remaining undefined, because there are compilation errors in driver/layernorm_driver.hpp pending an undefined RAN_GEN function.

include/miopen/miopen.h
cderb: Just so, MIOpen wouldn't compile without MIOPEN_BETA_API_FP8. The compilation variable was added to the CMakeLIsts.txt so that miopen would always build with it. This would however prevent the use of these enum values by a simple inclusion of the header without the defines argument.
But perhaps it would be better form to leave the enumerations without the guards and only keep them around the functions?

include/miopen/miopen.h
cderb: fixed this

include/miopen/miopen.h
atamazov: [resolved]

include/miopen/miopen.h
atamazov: [resolved]

CMakeLists.txt
```diff
@@ -68,6 +68,9 @@ include(ROCMCreatePackage)
 include(CheckCXXCompilerFlag)
 include(ROCMHeaderWrapper)
 
+# Build library with Beta APIs
+add_definitions("-DMIOPEN_BETA_API=1")
+
```
atamazov, line 73: [Recommendation] Remove all conditionals like 
```C++
#ifdef MIOPEN_BETA_API
...
#endif
```
from the library sources EXCEPT miopen.h.

But this can be done later.

atamazov (approved):
LGTM

junliume (approved):

PR #2460, opened 2023-10-17
Fix the return code on the workspace API

JehandadKhan:


amberhassaan (approved):
Looks good.

umangyadav (approved):
Thanks !

src/convolution_api.cpp
```diff
@@ -380,14 +380,12 @@ miopenConvolutionForwardGetWorkSpaceSize(miopenHandle_t handle,
 {
 
     MIOPEN_LOG_FUNCTION(handle, wDesc, xDesc, convDesc, yDesc, workSpaceSize);
-    miopen::try_([&] {
+    return miopen::try_([&] {
```
umangyadav, line 383: This may start breaking models though in MIGraphX. For example 2DUNet. 
cc : @pfultz2

junliume (approved):

PR #2464, opened 2023-10-18
bg/fix_ck_guard_in_bn : fix CK guard around bn

bghimireamd:
* move function definition inside CK guard
* fix CK guard around batch norm


junliume (approved):
Thanks! This unblock FIN build issues.

PR #2471, opened 2023-10-20
Add a check for packed tensors for convolution solvers

amberhassaan:
Some solvers allow non-packed tensor inputs while others don't. This PR adds a check to the ones that don't support.


junliume (approved):

src/solver/conv_winoRxS.cpp
```diff
@@ -622,6 +622,10 @@ static bool IsApplicableBase(const ExecutionContext& ctx, const ProblemDescripti
 {
     if(!problem.Is2d())
         return false;
+    if(problem.HasNonPackedTensors())
+    {
+        return false;
+    }
```
CAHEK7: As far as I know that solver works fine with non-packet tensors. @Kirpich30000 @Slimakanzer, am I correct?

src/solver/conv_wino_fury_RxS.cpp
```diff
@@ -171,6 +171,10 @@ bool ConvWinoFuryRxS<Winodata, Winofilter>::IsApplicable(const ExecutionContext&
     if(!problem.Is2d())
         return false;
 
+    if(problem.HasNonPackedTensors())
+    {
+        return false;
+    }
```
CAHEK7: This solver works fine with strides too.

CAHEK7 (changes requested):
Some of the solvers should not be disabled, but I don't know about the others.

Let's discuss shall we disable all the solvers in this commit and re-enable some of them later or we'll spend some time and disable only those which do not support strides.

I vote for the second option - disable only some of the solvers which do not support strides.

iq136boy:
> Some of the solvers should not be disabled, but I don't know about the others.
> 
> Let's discuss shall we disable all the solvers in this commit and re-enable some of them later or we'll spend some time and disable only those which do not support strides.
> 
> I vote for the second option - disable only some of the solvers which do not support strides.

@CAHEK7 Do we have a list of solvers that support non-packed tensors? I would recommend to disable solvers that we are not sure if they support non-packed tensors or not. And only enable the solvers if they passed the non-packed tensors tests.

src/solver/conv_direct_naive_conv_fwd.cpp
```diff
@@ -53,6 +53,10 @@ bool ConvDirectNaiveConvFwd::IsApplicable(const ExecutionContext& ctx,
     if(!problem.direction.IsForward())
         return false;
 
+    if(problem.HasNonPackedTensors())
+    {
+        return false;
+    }
```
iq136boy: The naive conv fwd solver should support non-packed tensors, right?

CAHEK7:
For sure I know only those two Winograd solvers. I'm not familiar with the others, but probably someone else is more familiar and can confirm it quickly.

Otherwise the solvers can be verified: if no stride information passed into kernel - that for sure does not support non-packed tensors, else do a simple direct test. But this may require a lot of time and probably should be a separate task.

src/solver/conv_direct_naive_conv_fwd.cpp
amberhassaan: correct.

amberhassaan:
@CAHEK7 , @iq136boy : I have removed the checks from  `CK 3D grouped convolution` and `naive convolution` solvers, but for the rest, I'm not sure of the support myself.

CAHEK7 (approved):
Probably we need to create some issue to check the other solvers can they be applicable for non-packed tensors or not and re-enable them.

CAHEK7:
> @CAHEK7 , @iq136boy : I have removed the checks from `CK 3D grouped convolution` and `naive convolution` solvers, but for the rest, I'm not sure of the support myself.

Could you re-enable two winograd solvers as well?

iq136boy (approved):

src/solver/conv_winoRxS.cpp
Kirpich30000: Winograd supports arbitrary strides for N, C, H, K, R dimensions in number of elements (fp16 or fp32). W on both input and output, and S on filters should be dense.

Kirpich30000: Non dense W and R is possible, but likely will greatly affect performance. Is there real interest in functional (but not performance) support for this?

CAHEK7: It still can be faster than naive convolution.

amberhassaan:
@CAHEK7 , @iq136boy , @junliume , @JehandadKhan : https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2478 created to develop a test that tests every solver against non-packed inputs.

src/include/miopen/fusion/utils.hpp
```diff
@@ -84,6 +84,10 @@ inline bool WinoCommonIsApplicable(const FusionContext& context, const FusionDes
         return false;
     if(!conv_problem.IsFp32())
         return false;
+    if(conv_problem.HasNonPackedTensors())
+    {
+        return false;
+    }
```
averinevg: [Recommendation] Curly braces here and in other places may be omitted. The code in this case will look better when you need to check multiple conditions.

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
```diff
@@ -243,6 +243,10 @@ bool ConvBiasActivAsm1x1U::IsApplicable(const FusionContext& context,
     const auto conv_problem = problem.GetConvProblem(0, conv::Direction::Forward);
     const auto conv_ctx     = context.GetConvContext(conv_problem);
 
+    if(conv_problem.HasNonPackedTensors())
+    {
+        return false;
+    }
```
averinevg: This condition is checked in `ConvAsm1x1U::IsApplicable()`

src/solver/conv_ck_igemm_fwd_bias_activ_fused.cpp
```diff
@@ -411,15 +411,18 @@ bool ConvCKIgemmFwdBiasActivFused::IsApplicable(const FusionContext& ctx,
         return false;
     const auto conv_problem = fdesc_problem.GetConvProblem(0, conv::Direction::Forward);
 
-    if(conv_problem.IsTensorsCasted())
-        return false;
```
averinevg: [Recommendation] It's better not to move code that hasn't changed (here and in other places too)

src/include/miopen/conv/problem_description.hpp
```diff
@@ -367,6 +367,17 @@ struct ProblemDescription : ProblemDescriptionBase
...
+    {
+        return !(in.IsPacked() && weights.IsPacked() && out.IsPacked());
+    }
+
+    bool HasMixedDataTypes() const
+    {
+        return !(GetInDataType() == GetWeightsDataType() &&
+                 GetWeightsDataType() == GetOutDataType());
+    }
+
```
averinevg, line 380: :+1:

src/include/miopen/conv/problem_description.hpp
```diff
@@ -367,6 +367,17 @@ struct ProblemDescription : ProblemDescriptionBase
     bool IsNCHWc_NCHWc() const;
     bool IsNCHWc_CHWNc() const;
 
+    bool HasNonPackedTensors() const
```
atamazov, line 370: [Recommendation] Formally speaking `HasNonPackedTensors` is incorrect when only one tensor is non-packed. I would say that `HasAtLeastOneNonPackedTensor` is fully correct, but I am not going to recommend this ;)

From the other hand, when we say "non-packed convolution" we imply that at least one tensor is non-packed, right? So maybe simply
```suggestion
    bool NonPacked() const
```
is better. Up to you; this is just a matter of taste.

src/include/miopen/fusion/utils.hpp
atamazov: No `{ }` around single statement (`return false` in this case) everywhere, please.

amberhassaan: I can remove them. My personal preference is to always have `{}` and have the starting `{` on the same line.

src/solver/conv_ck_igemm_fwd_bias_activ_fused.cpp
amberhassaan: So I moved these to order the checks by a guess of cost. Expensive checks should come after cheaper checks.

JehandadKhan (approved):

src/solver/conv_ck_igemm_fwd_bias_activ_fused.cpp
```diff
@@ -411,15 +411,18 @@ bool ConvCKIgemmFwdBiasActivFused::IsApplicable(const FusionContext& ctx,
         return false;
     const auto conv_problem = fdesc_problem.GetConvProblem(0, conv::Direction::Forward);
 
-    if(conv_problem.IsTensorsCasted())
```
atamazov: Please revert all code moves. Thanks.

atamazov (changes requested):

atamazov:
@amberhassaan 

> @CAHEK7 Do we have a list of solvers that support non-packed tensors? I would recommend to disable solvers that we are not sure if they support non-packed tensors or not. And only enable the solvers if they passed the non-packed tensors tests.

Please assume that all old solvers do not support non-packed tensors. Even if Winograd seems supporting them, enabling the feature require investigations and testing.

junliume:
@amberhassaan could you resolve the last round of review opinions above? We will let CI run once and then merge. Thanks!

src/solver/conv_winoRxS.cpp
amberhassaan: adding back the check to WinoGrad solvers at @atamazov 's recommendation

src/solver/conv_wino_fury_RxS.cpp
amberhassaan: adding back the check to WinoGrad solvers at @atamazov 's recommendation

src/include/miopen/conv/problem_description.hpp
amberhassaan: @atamazov : favoring the current name as it is more descriptive.

src/solver/conv_ck_igemm_fwd_bias_activ_fused.cpp
amberhassaan: done in latest commit.

amberhassaan:
> @amberhassaan could you resolve the last round of review opinions above? We will let CI run once and then merge. Thanks!

@junliume : I have addressed all the comments. Please merge if CI passes.

src/solver/conv_asm_7x7c3h224w224k64u2v2p3q3f1.cpp
```diff
@@ -65,17 +68,11 @@ bool ConvAsm7x7c3h224w224k64u2v2p3q3f1::IsApplicable(const ExecutionContext& ctx
 #endif
     if(!(name == "gfx800" || name == "gfx802" || name == "gfx803" || name == "gfx804" ||
          name == "gfx900" || name == "gfx904" || name == "gfx906" || name == "gfx908"))
-    {
         return false;
-    }
```
averinevg, line 70: Curly braces are needed here. There was some misunderstanding here, I meant that curly braces are not required where the condition is short enough and the returned statement occupies one line. My comment related primarily to the new code. I should have explained this in more detail.

src/solver/conv_asm_1x1u_stride2.cpp
```diff
@@ -505,13 +507,9 @@ bool ConvAsm1x1UV2::IsApplicable(const ExecutionContext& ctx,
 
     const std::string name = ctx.GetStream().GetDeviceName();
     if(name.find("gfx8") == std::string::npos && name.find("gfx9") == std::string::npos)
-    {
         return false;
-    }
```
averinevg, line 510: Ditto

src/solver/conv_asm_dir_BwdWrW1x1.cpp
```diff
@@ -493,13 +494,9 @@ bool ConvAsmBwdWrW1x1::IsApplicable(const ExecutionContext& ctx,
 
     const std::string name = ctx.GetStream().GetDeviceName();
     if(name.find("gfx8") == std::string::npos && name.find("gfx9") == std::string::npos)
-    {
         return false;
-    }
```
averinevg, line 498: Ditto

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -44,15 +44,14 @@ bool ConvDirectNaiveConvBwd::IsApplicable(const ExecutionContext& ctx,
     if(!ConvDirectNaiveConvIsApplicableByKernelType(ctx, problem))
         return false;
 
+    if(!problem.direction.IsBackwardData())
+        return false;
```
averinevg, line 48: Existing code has been moved

src/solver/conv_asm_7x7c3h224w224k64u2v2p3q3f1.cpp
amberhassaan: OK, this is quite arbitrary and I find it hard to agree with. The language allows two choices: (1) no braces when the conditional has a single statement in the body and that has nothing to do with the length of the condition. That can be a source of bugs in absence of formatting, so I use `{}` all the time (choice #2). What you're asking is an arbitrary style choice and I find it inconsistent. Should we define how long is a "long condition"?

src/solver/conv_direct_naive_conv_bwd.cpp
amberhassaan: I will revert it but the existing order of checks itself is something to discuss as it seems arbitrary and based on what the programmer wrote at the time or copy-pasted from somewhere else.

src/solver/conv_asm_7x7c3h224w224k64u2v2p3q3f1.cpp
CAHEK7: I think it must be formalized and forced by clang-format.
If it's not forced by clang-format, we are free to do whatever we want - and most probably we want to argue what's the best codestyle ever.
@atamazov @averinevg please add appropriate check for it and we automatically will follow. I see no point to check codestyle on review.

CAHEK7:
Let's keep all the curly braced in this PR like they are, add appropriate clang-format rule, apply it and forget about this problem forever. I see no point to spend any time to review codestyle if it can be checked and applied automatically.

src/solver/conv_direct_naive_conv_bwd.cpp
averinevg: I agree that the current order of checking conditions may not be optimal, but here we need to do a little research and evaluate the expected benefit from this. It's better to do this in a separate pull request

src/solver/conv_asm_7x7c3h224w224k64u2v2p3q3f1.cpp
averinevg: I'll see if I can set up a clang format for this

amberhassaan: I'm of the opinion that we should always use `{}` and modify `.clang-format` to put the starting `{` on the same line as `if` so that code height remains somewhat small.

On the general topic of style, I've opened up https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2482 for discussion on adopting Google Style Guide.

JehandadKhan:
> Let's keep all the curly braced in this PR like they are, add appropriate clang-format rule, apply it and forget about this problem forever. I see no point to spend any time to review codestyle if it can be checked and applied automatically.

I agree with this, formatting issues should never be part of a PR discussion since these checks can be automated.

JehandadKhan:
@junliume  This PR can be merged.

src/include/miopen/conv/problem_description.hpp
atamazov: [resolved]

src/solver/conv_asm_7x7c3h224w224k64u2v2p3q3f1.cpp
atamazov: @CAHEK7 
> If it's not forced by clang-format, we are free to do whatever we want

NO. If there is no enforcing tool, then code review comes into play. That's one of the purposes of code reviews.

The reviewers are spending time, and **_good authors value this and learn from the reviews_**. And professional engineers are able to adapt his/her own tastes to whatever practices used in the communities (they joined).

@amberhassaan 
Is it fine that a small code fix you were asked to make here transformed to a separate ticket, discussion thread here and additional work for Evgenii?

atamazov (commented):
LGTM!

src/solver/conv_asm_7x7c3h224w224k64u2v2p3q3f1.cpp
CAHEK7: @atamazov NO for you NO)
The problem happened because in the first comment it was recommended to remove the brackets for a single statement conditions (which was done) and later for some of the cases it was recommended to return the brackets back if the condition is long enough. It is too fuzzy and it's a matter of taste, not a rule.

clang-format can handle curly braced case in one way or another. Probably it handles it not ideally, but in uniform and automated way, which simplify everything. So, it must be used, moreover, if the ideal rules can't be formalized and each reviewer may have his own flavor of ideal solution.
And as you've already said - we must adapt our tastes to make the life easier.

Just to mention, I'm talking about the case which can be handled by clang-format.
If it can't handle the case with includes (quotes vs brackets), so I don't ask for it, but @pfultz2 provided a reference how to properly fix it.
If it can't handle naming convention, I don't ask for it, I'm asking for some sort of small notes like "`WeUseCamelCase` for types and classes and `we_use_underscores` for variables". It would help a lot for newcomers or if the code is mess.
Since it's an opensource project and available for everyone, it would be nice to have such a document.

And as I've already said - we want to argue about the codestyle, it happens over and over again.

amberhassaan: @atamazov : no, it's not fine that we are discussing style issues that would've been easily handled if we had a style document. We are creating unnecessary back and forth for minor matters. We need to have a `developer's guide` document in the repository that documents all the nuances of things like `Invoker & InvokerFactory` and how `in is out and out is in` in backward convolution and style issues like include files and variable/function naming etc. Hence the discussion in #2483

Issue #2472, opened 2023-10-21
[test] test_na_train error: 'MIO_BN_GFX110X' is not defined


junliume:
@DrizztDoUrden After #2388 the following test starts to fail in CI pipelines.

It is interesting:
- Why it happens now?
- It seems that from the command line `MIO_BN_GFX103X` is defined while `MIO_BN_GFX110X` is not. Maybe we missed it?

```
          Start   39: test_na_train
  38/1533 Test   #39: test_na_train ..................................................................................................................................................................................................................................................................................................................................................................***Failed    1.34 sec
PRNG seed: 12345678
MIOpen(HIP): Info [get_device_name] Raw device name: gfx906:sramecc+:xnack-
MIOpen(HIP): Info [Handle] stream: 0x2456b20, device_id: 3
MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, HIP version 5.7.23365, MIOpen version 3.00.0.5798751b2
MIOpen(HIP): Info [Measure] RamDb::Prefetch time: 0.105626 ms
MIOpen(HIP): Info [TryLoad] Find-db regenerating.
MIOpen(HIP): Warning [hip_mem_get_info_wrapper] hipMemGetInfo error, status: 1
MIOpen(HIP): Warning [hip_mem_get_info_wrapper] hipMemGetInfo error, status: 1
MIOpen(HIP): Warning [hip_mem_get_info_wrapper] hipMemGetInfo error, status: 1
MIOpen(HIP): Warning [hip_mem_get_info_wrapper] hipMemGetInfo error, status: 1
MIOpen(HIP): Warning [hip_mem_get_info_wrapper] hipMemGetInfo error, status: 1
MIOpen(HIP): Warning [hip_mem_get_info_wrapper] hipMemGetInfo error, status: 1
MIOpen(HIP): Warning [hip_mem_get_info_wrapper] hipMemGetInfo error, status: 1
MIOpen(HIP): Info [AllocateBuffersAndMakeFusionInvokeParams] in addr: 0x7f873bc06000, size: 802816
MIOpen(HIP): Warning [hip_mem_get_info_wrapper] hipMemGetInfo error, status: 1
MIOpen(HIP): Info [AllocateBuffersAndMakeFusionInvokeParams] out addr: 0x7f873bcca000, size: 802816
[...]

PR #2473, opened 2023-10-22
[HotFix] Missing MIO_BN_GFX110X building kernels

junliume:
https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2472


muralinr (approved):
LGTM. This is passing target name flag, shouldn't cause any side effects.

PR #2477, opened 2023-10-23
[HOTFIX] disable bwd_gemm solver to fix 2474

carlushuang:


src/solver/gemm_bwd.cpp
```diff
@@ -43,6 +43,7 @@
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_CONV_PRECISE_ROCBLAS_TIMING)
 
 #define WORKAROUND_MIOPENGEMM_ISSUE_59 1
+#define WORKAROUND_MIOPENGEMM_ISSUE_2474 1
```
atamazov:
```suggestion
#define WORKAROUND_ISSUE_2474 1
```
We do not use MIOpenGemm (GEMM backend for OpenCL) anymore.

atamazov (changes requested):
- It should be possible to cancel the W/A by explicit setting `MIOPEN_DEBUG_CONV_GEMM=1`

atamazov:
@carlushuang @junleemoreh @JehandadKhan 
Generally, this W/A seems too wide to be merged in, unless there is some absolute need. The promotion of MIOpen into Staging (amd-develop) should not be performed until W/A is removed.

I recommend redesigning:
- Introduce the individual env controls for GEMM solvers.
- The W/A should change the defaults of some of them (those that belong to backward suspects) from "enabled" to "disabled".
- Explicit enabling of the env controls should re-enable the suspects.

carlushuang:
Hi @atamazov , this PR unblock the fix for issue: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2459, this is why I made a quick hotfix for the bwd_gemm solver

src/solver/gemm_bwd.cpp
carlushuang: done in https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2477/commits/fa5a598adc02d6a381d50ceb80fd41d7cf68c9b7

atamazov:
@carlushuang 
> Hi @atamazov , this PR unblock the fix for issue: #2459, this is why I made a quick hotfix for the bwd_gemm solver

Do you see the problem on clean develop?
- If yes, then this PR is very important because the issue blocks the entire development pipeline.
- If not, then could it be so that #2476 somehow breaks GEMM?

Regardless of the importance of this W/A the following remains valid:
- **_The promotion of MIOpen into Staging (amd-develop) should not be performed until this W/A is removed or narrowed down._**

(Just for awareness of everyone! 🤝 )

/cc @junliume @JehandadKhan

src/solver/gemm_bwd.cpp
atamazov: [resolved]

carlushuang:
This PR is identified as specific docker related issue, hence I will close this PR

Issue #2403, opened 2023-09-20
Resolve post-merge review comments of PR #2320


atamazov:
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_high https://github.com/ROCmSoftwarePlatform/MIOpen/labels/request_for_comments Finally decide how to enable experimental API features
  - [ ] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1316419544
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_blocker Let's enable and properly test the experimental feature everywhere.
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1326549529 
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327664870
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327665023
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327667003
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327674330
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327674436
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327676122
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327678271
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327798211
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327798755
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327801851
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327817140
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_normal Easy to resolve:
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1316416030
  - [x] https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2320#discussion_r1327669307
[...]

PR #2463, opened 2023-10-18
Update dependencies and Dockerfile ROCm versions

junliume:
Some old dependencies have security risks and we are asked to update them if we can

PR #2481, opened 2023-10-25
Integrate CK's layer norm into MIOpen solver

seungmanhan:
- Applied issue #2403 
- Integrated CK's layer norm into MIOpen solver
- Added test for CK' layer norm


junliume:
@seungmanhan some issues in building MIOpen (because we enabled warning as errors):
```
In file included from /var/jenkins/workspace/bs_MIOpen_integrate_CK_layernorm/src/normalization/problem_description.cpp:27:

/var/jenkins/workspace/bs_MIOpen_integrate_CK_layernorm/src/include/miopen/normalization/problem_description.hpp:90:19: error: 'MakeNetworkConfig' overrides a member function but is not marked 'override' [-Werror,-Wsuggest-override]

    NetworkConfig MakeNetworkConfig() const;

                  ^

/var/jenkins/workspace/bs_MIOpen_integrate_CK_layernorm/src/include/miopen/problem_description_base.hpp:61:41: note: overridden virtual function is here

    [[nodiscard]] virtual NetworkConfig MakeNetworkConfig() const = 0;

                                        ^

**Review by CAHEK7 — DISMISSED** [review]
May I ask to double check the latest versions of CK kernel using and apply changes from https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2379
It has a few simplifications, but it may not be perfectly suitable for layernorm solvers. Feel free to ask any questions.

File: src/include/miopen/normalization/problem_description.hpp [hunk 1]
```diff
@@ -0,0 +1,109 @@
...
+#include <miopen/tensor.hpp>
+
+#include <cassert>
+#include <string>
+
+namespace miopen {
+
+struct NetworkConfig;
+
+namespace normalization {
```
averinevg:
```suggestion
namespace norm {
```

averinevg: I don't insist on renaming, it's just a suggestion, but the namespaces for other layers are either shorthand words or the words themselves are short enough. The same applies to paths.

driver/layernorm_driver.hpp
```diff
@@ -24,7 +24,6 @@
  *
  *******************************************************************************/
 #include <miopen/miopen.h>
-#ifdef MIOPEN_BETA_API
```
atamazov, line 27: Please 
```C++
#define MIOPEN_BETA_API 1
```
prior any inclusion of miopen.h in our source code. This is to make beta declarations/definitions visible in the library sources.

driver/driver.hpp
```diff
@@ -150,11 +150,7 @@ inline void PadBufferSize(size_t& sz, int datatype_sz)
     printf("Supported Base Arguments: conv[fp16|int8|bfp16|fp8|bfp8], CBAInfer[fp16], "
            "pool[fp16], lrn[fp16], "
            "activ[fp16], softmax[fp16], bnorm[fp16], rnn[fp16], gemm, ctc, dropout[fp16], "
-           "tensorop[fp16], reduce[fp16,fp64]"
-#ifdef MIOPEN_BETA_API
-           ", layernorm[bf16, fp16, fp32]"
-#endif
```
atamazov, line 156: Please see my previous comment and keep `", layernorm[bf16, fp16, fp32]"`, just remove the conditional.

driver/driver.hpp
```diff
@@ -175,11 +171,8 @@ inline std::string ParseBaseArg(int argc, char* argv[])
        arg != "bnormfp16" && arg != "rnn" && arg != "rnnfp16" && arg != "rnn_seq" &&
        arg != "rnn_seqfp16" && arg != "gemm" /*&& arg != "gemmfp16"*/ && arg != "ctc" &&
        arg != "dropout" && arg != "dropoutfp16" && arg != "tensorop" && arg != "tensoropfp16" &&
-       arg != "reduce" && arg != "reducefp16" && arg != "reducefp64" &&
-#ifdef MIOPEN_BETA_API
-       arg != "layernorm" && arg != "layernormfp16" && arg != "layernormbfp16" &&
-#endif
```
atamazov, line 181: Same.

src/include/miopen/layernorm.hpp
```diff
@@ -24,7 +24,6 @@
  *
  *******************************************************************************/
 #include <miopen/miopen.h>
-#ifdef MIOPEN_BETA_API
```
atamazov, line 27: Please `#define MIOPEN_BETA_API 1` right above inclusion of miopen.h.

src/include/miopen/layernorm.hpp
```diff
@@ -35,7 +34,7 @@ namespace miopen {
 struct Handle;
 struct TensorDescriptor;
 
-miopenStatus_t LayerNormForward(const Handle& handle,
```
atamazov, line 38: Is it really necessary?

src/kernels/MIOpenLayerNorm.cpp
```diff
@@ -109,16 +102,15 @@ extern "C" __global__ void LayernormFwdContiguous(const FLOAT* __restrict__ x,
...
-        uint64_t idx = gid * inner_size + i;
+        size_t idx = gid * inner_size + i;
 
         FLOAT_ACCUM pweight;
         FLOAT_ACCUM pbias;
 
-        pweight = mode ? CVT_FLOAT2ACCUM(1) : CVT_FLOAT2ACCUM(weight[i]);
-        pbias   = mode ? CVT_FLOAT2ACCUM(0) : CVT_FLOAT2ACCUM(bias[i]);
+        pweight = mode ? CVT_FLOAT2ACCUM(weight[i]) : CVT_FP32_2ACCUM(1.0f);
+        pbias   = mode ? CVT_FLOAT2ACCUM(bias[i]) : static_cast<FLOAT>(0);
```
atamazov, line 119: Change of behavior. Bugfix?

test/gtest/layernorm.hpp
```diff
@@ -24,7 +24,6 @@
  *
  *******************************************************************************/
 #include <miopen/miopen.h>
-#ifdef MIOPEN_BETA_API
```
atamazov, line 27: Please `#define MIOPEN_BETA_API 1` right above inclusion of miopen.h.

src/solver/norm/forward_layernorm2d_ck.cpp
```diff
@@ -0,0 +1,257 @@
...
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+#include <miopen/solver/ck_utility_common.hpp>
+#include <ck/library/tensor_operation_instance/gpu/normalization.hpp>
+#endif
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_CK_LN)
+
+namespace miopen {
+namespace solver {
+namespace norm {
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
```
atamazov: [Side comment] @JehandadKhan I am still wondering why we need `MIOPEN_BACKEND_HIP` in such constructs. I guess that `MIOPEN_USE_COMPOSABLEKERNEL` is 1 only with HIP backend, right?

src/solver/norm/forward_layernorm2d_ck.cpp
```diff
@@ -0,0 +1,257 @@
...
+ *******************************************************************************/
+
+#include <miopen/norm/solvers.hpp>
+#include <miopen/norm/invoke_params.hpp>
+#include <miopen/layernorm.hpp>
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+#include <miopen/solver/ck_utility_common.hpp>
+#include <ck/library/tensor_operation_instance/gpu/normalization.hpp>
+#endif
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_CONV_CK_LN)
```
atamazov: Env vars are "global" in the execution environment. Let's encode solver name in the solver-specific vars. 
```suggestion
MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_LAYERNORM2DCKFORWARD_CONV_CK_LN)
```

atamazov (changes requested):
Partial review

src/kernels/MIOpenLayerNorm.cpp
seungmanhan: Yes, It was working incorrectly previously.

src/include/miopen/layernorm.hpp
seungmanhan: This is a change to call ExecutePrimitive function.

driver/layernorm_driver.hpp
seungmanhan: What you mean is leave MIOPEN_BETA_API as is and just define MIOPEN_BETA_API before miopen.h?
I understood that MIOPEN_BETA_API should only exist in miopen.h. https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2430#discussion_r1352807948

atamazov: @seungmanhan 
> What you mean is leave MIOPEN_BETA_API as is and just define MIOPEN_BETA_API before miopen.h?

No. The conditional (`#ifdef MIOPEN_BETA_API`) is correctly removed in this PR. But the PR misses definition of MIOPEN_BETA_API before `#include <miopen/miopen.h>`.

> https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2430#discussion_r1352807948

That comment says that _conditionals_ should be removed. But we need _definitions_ of MIOPEN_BETA_API because:

>> This is to make beta declarations/definitions \<from miopen.h\> visible in the library sources.

The idea is to unconditionally include all the beta functionality into the library and tests, so
- the users who want to use beta features do not need to rebuild the library
- the beta functionality is always tested

src/include/miopen/layernorm.hpp
atamazov: @seungmanhan Thanks [resolved]

src/kernels/MIOpenLayerNorm.cpp
atamazov: @seungmanhan Thanks [resolved]

driver/driver.hpp
seungmanhan: In this case layernorm[bfp16, fp16] still remains and only the condition has been deleted. Please explain if your intention is different.

driver/layernorm_driver.hpp
seungmanhan: fixed this

test/gtest/layernorm.hpp
seungmanhan: fixed this

src/solver/norm/forward_layernorm2d_ck.cpp
seungmanhan: fixed this

src/include/miopen/layernorm.hpp
seungmanhan: fixed this

src/solver/norm/forward_layernorm2d_ck.cpp
JehandadKhan: I agree, both of these are redundant.

JehandadKhan:
@atamazov  and @CAHEK7  Please take another look at the fixes.

src/solver/norm/forward_layernorm2d_ck.cpp
seungmanhan: fixed this

src/solver/norm/forward_layernorm2d_ck.cpp
atamazov: [resolved]

src/solver/norm/forward_layernorm2d_ck.cpp
atamazov: [resolved]

test/gtest/layernorm.hpp
atamazov: [resolved]

src/include/miopen/layernorm.hpp
atamazov: [resolved]

driver/layernorm_driver.hpp
atamazov: [resolved]

driver/driver.hpp
atamazov: @seungmanhan Oops, wrong review comment [resolved]

driver/driver.hpp
atamazov: @seungmanhan Wrong review comment [resolved]

atamazov:
@JehandadKhan Yeah, as soon as time permits.

junliume (approved):

driver/layernorm_driver.hpp
```diff
@@ -23,8 +23,8 @@
  * SOFTWARE.
  *
  *******************************************************************************/
+#define MIOPEN_BETA_API 1
```
JehandadKhan: This is not necessary since the directive is added to CMake.

src/include/miopen/layernorm.hpp
```diff
@@ -23,8 +23,8 @@
  * SOFTWARE.
  *
  *******************************************************************************/
+#define MIOPEN_BETA_API 1
```
JehandadKhan: This should not be required since its defined in CMake.

src/norm/problem_description.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    std::ostringstream ss;
+
+    ss << "dtype" << dtype;
+    ss << "normalized_dim" << normalized_dim;
+    ss << "grid_size" << grid_size;
+    ss << "outer_size" << outer_size;
+    ss << "inner_size" << inner_size;
+    ss << "mode" << mode;
+    ss << "epsilon" << epsilon;
```
JehandadKhan: Please be conservative about the entries here, since a change here would mean that a new entry would trickle accross MIOpen in various caches. For example, `epsilon` is a floating point number ? Similarly, `*_size` numbers etc would have a massive space.

JehandadKhan: From the kernel source code, I dont understand why the grid size needs to be part of the key?

JehandadKhan (changes requested):
Generally looks good, I dont see any updates to documentation, perhaps general answers to the following questions would help:

* What is LayerNorm operation ( A link to an external page should suffice)
* What types of layer norms are supported ( 2D/3D, generally height, width, depth etc) 
* What is mode

A question around the problem config key, that I would like an answer to.

seungmanhan:
> Generally looks good, I dont see any updates to documentation, perhaps general answers to the following questions would help:
> 
> * What is LayerNorm operation ( A link to an external page should suffice)
> * What types of layer norms are supported ( 2D/3D, generally height, width, depth etc)
> * What is mode
> 
> A question around the problem config key, that I would like an answer to.

I never received any explanation about doc work when working with layernorm. Where can I write documentation?

driver/layernorm_driver.hpp
seungmanhan: This part is a modification requested by @atamazov. https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2481#discussion_r1379293866

src/norm/problem_description.cpp
seungmanhan: Anything other than those related to kernel call has been removed.

junliume:
> > Generally looks good, I dont see any updates to documentation, perhaps general answers to the following questions would help:
> > 
> > * What is LayerNorm operation ( A link to an external page should suffice)
> > * What types of layer norms are supported ( 2D/3D, generally height, width, depth etc)
> > * What is mode
> > 
> > A question around the problem config key, that I would like an answer to.
> 
> I never received any explanation about doc work when working with layernorm. Where can I write documentation?

@seungmanhan Doc updates should be included here: https://github.com/ROCmSoftwarePlatform/MIOpen/tree/develop/docs

junliume:
@zjing14 @JehandadKhan : could you check why the user adopted the updated CK format and the CI is failing?
```
/var/jenkins/workspace/bs_MIOpen_integrate_CK_layernorm/src/solver/norm/forward_layernorm2d_ck.cpp:117:44: error: too many arguments to function call, expected 14, have 16

                                           rstd,

                                           ^~~~~
```

src/include/miopen/layernorm.hpp
atamazov: @JehandadKhan Right! Thanks for informing. We do not need this in library sources and in tests.

driver/layernorm_driver.hpp
atamazov: @seungmanhan We actually do not need this, please look at https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2481#discussion_r1383825898. My comment about necessity of this definition is correct, but we do not need to change source code for that ;)

src/include/miopen/layernorm.hpp
atamazov: @seungmanhan Same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2481#discussion_r1390435324

test/gtest/layernorm.hpp
atamazov: @seungmanhan Same as https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2481#discussion_r1390435324

driver/layernorm_driver.hpp
atamazov: @seungmanhan Actually this is not needed, I was not aware of CMakeLists.txt changes. See

atamazov:
> > > I dont see any updates to documentation, perhaps general answers to the following questions would help:
> > > 
> > > * What is LayerNorm operation ( A link to an external page should suffice)
> > > * What types of layer norms are supported ( 2D/3D, generally height, width, depth etc)
> > > * What is mode
> 
> @seungmanhan Doc updates should be included here: https://github.com/ROCmSoftwarePlatform/MIOpen/tree/develop/docs

@seungmanhan Some more details. IIRC you need to:
- Update https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/docs/apireference.rst to include layernorm documentation into the table of contents
- Create /docs/layernorm.rst (see convolution.rst for example).
- miopen.h where the actual documentation resides.

@JehandadKhan @junliume If this is too much work, then perhaps it would be acceptable to edit the documentation in a separate PR?

atamazov:
@junliume 
> @zjing14 @JehandadKhan : could you check why the user adopted the updated CK format and the CI is failing?
> 
> ```
> /var/jenkins/workspace/bs_MIOpen_integrate_CK_layernorm/src/solver/norm/forward_layernorm2d_ck.cpp:117:44: error: too many arguments to function call, expected 14, have 16
>                                            rstd,
>                                            ^~~~~
> ```

@seungmanhan This PR is 26 commits behind develop, including #2501 (may be important for doc updates), #2479 (CK hash update) and #2463. I recommend merging develop into `integrate_CK_layernorm`, updating your development environment with newer CK etc. This should make any CI problems visible on your development machine. Good luck ;)

src/include/miopen/norm/problem_description.hpp
```diff
@@ -0,0 +1,109 @@
...
+    bool IsRank2Dim1() const { return (xDesc.GetLengths().size() == 2) && (normalized_dim == 1); }
+    bool IsRank4Dim1() const { return (xDesc.GetLengths().size() == 4) && (normalized_dim == 1); }
+    bool IsLargeSize() const
+    {
+        auto dims = xDesc.GetLengths();
+
+        size_t outer_size = 1;
+        for(size_t i = 0; i < normalized_dim; i++)
+        {
+            outer_size *= dims[i];
```
atamazov, line 123: [Q] Layernorm does not support non-packed tensors so far, right?

src/include/miopen/norm/problem_description.hpp
```diff
@@ -0,0 +1,109 @@
...
+    const TensorDescriptor& GetBiasDesc() const { return biasDesc; }
+    const TensorDescriptor& GetYDesc() const { return yDesc; }
+    const TensorDescriptor& GetMeanDesc() const { return meanDesc; }
+    const TensorDescriptor& GetRstdDesc() const { return rstdDesc; }
+    float GetEpsilon() const { return epsilon; }
+    int32_t GetNormalizedDim() const { return normalized_dim; }
+
+    bool IsRank2Dim1() const { return (xDesc.GetLengths().size() == 2) && (normalized_dim == 1); }
+    bool IsRank4Dim1() const { return (xDesc.GetLengths().size() == 4) && (normalized_dim == 1); }
+    bool IsLargeSize() const
```
atamazov, line 116: [Recommendation] AFAICS this is not a problem-specific attribute, but a helper function to support some solver-specific check. I highly recommend removing it from the problem.

src/include/miopen/norm/problem_description.hpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+        auto dims = xDesc.GetLengths();
+
+        size_t outer_size = 1;
+        for(size_t i = 0; i < normalized_dim; i++)
+        {
+            outer_size *= dims[i];
+        }
+
+        return (outer_size > 32);
```
atamazov, line 126: [Q] Suspiciously small limit. Do you have 32 bits in mind or limit is really only 32 pixels?

src/solver/norm/forward_layernorm2d_ck.cpp
```diff
@@ -0,0 +1,272 @@
...
+bool Layernorm2DCKForward::IsApplicable(
+    [[maybe_unused]] const ExecutionContext& context,
+    [[maybe_unused]] const miopen::norm::ProblemDescription& problem) const
+{
+#if MIOPEN_USE_COMPOSABLEKERNEL
+    if(miopen::IsDisabled(MIOPEN_DEBUG_LAYERNORM2DCKFORWARD_CONV_CK_LN{}))
+        return false;
+    if(!problem.IsRank2Dim1())
+        return false;
+    if(!problem.IsLargeSize())
```
atamazov, line 229: And IIRC there is some member functions in tensor that may satisfy the solver needs. IIRC You can ask tensor about its properties:
- Is it packed (and then return false if it is not)
- What is its flattened size (and return false if it exceeds some solver-specific limit)

atamazov (changes requested):
Partial review.

src/norm/problem_description.cpp
```diff
@@ -0,0 +1,66 @@
...
+    size_t i          = 0;
+    for(; i < normalized_dim; i++)
+    {
+        outer_size *= dims[i];
+    }
+
+    for(; i < dims.size(); i++)
+    {
+        inner_size *= dims[i];
+    }
```
atamazov: [Resommendation] This looks less fragile to me:
```suggestion
    for(size_t i = 0ULL; i < dims.size(); ++i)
    {
        if (i < normalized_dim)
            outer_size *= dims[i];
        else
            inner_size *= dims[i];
    }
```
[Q] If you have time, can you please also shed some light to the notions of "outer" and "inner" sizes? These seem quite problem-specific and I am curious.

src/include/miopen/norm/problem_description.hpp
seungmanhan: you're right. Only cases with better performance than ROCm pytorch are being added to PR, so only packed cases will be supported.

src/include/miopen/norm/problem_description.hpp
seungmanhan: I misunderstood the problem description. fixed it.

src/include/miopen/norm/problem_description.hpp
seungmanhan: This conditional statement was added because the compoasble kernel performs better when outer_size is larger than 32. In fact, it works flawlessly even at larger sizes.

src/solver/norm/forward_layernorm2d_ck.cpp
seungmanhan: I added a few more conditions.

src/norm/problem_description.cpp
seungmanhan: fixed it.

driver/layernorm_driver.hpp
seungmanhan: fixed it.

src/include/miopen/layernorm.hpp
seungmanhan: fixed it.

seungmanhan:
> @zjing14 @JehandadKhan : could you check why the user adopted the updated CK format and the CI is failing?
> 
> ```
> /var/jenkins/workspace/bs_MIOpen_integrate_CK_layernorm/src/solver/norm/forward_layernorm2d_ck.cpp:117:44: error: too many arguments to function call, expected 14, have 16
> 
>                                            rstd,
> 
>                                            ^~~~~
> ```

It was resolved after merging develop

src/norm/problem_description.cpp
seungmanhan: The outer size is used to determine the global work size and is calculated as the total tensor size/inner size.
The inner size is the size of the dimension where normlize is performed.

src/include/miopen/norm/problem_description.hpp
atamazov: @seungmanhan So this is a kind of heuristics intended to optimize layernorm performance, right? If so, then please add a comment like `// Performance optimization heuristics: make the CK solver non-applicable because non-CK kernel is expected to be faster`.

JehandadKhan (approved):
LGTM.

The documentation will follow up in a separate PR as suggested by @atamazov

JehandadKhan:
@junliume  we can go ahead and merge this.

src/include/miopen/norm/problem_description.hpp
atamazov: @seungmanhan 
> I highly recommend removing it from the problem.

Have you decided to keep this in the problem description? Or would like to resolve this comment in a subsequent PR?

src/include/miopen/norm/problem_description.hpp
atamazov: Would you like to resolve this in the subsequent PR?

[Notice] Please don't be afraid of ugly code: sometimes we have to write it. Ugly code is a sign that the design needs to be improved. Therefore, it should be in sight, not hidden under the carpet ;)

src/solver/norm/forward_layernorm2d_ck.cpp
atamazov: [resolved]

src/norm/problem_description.cpp
atamazov: [resolved] thanks.

atamazov (approved):
No objections!

test/gtest/layernorm.cpp
```diff
@@ -24,15 +24,29 @@
...
+    }
+    return tmp.front();
+}
+
+struct LayerNormTestFloat : LayerNormTest<float>
 {
 };
 
-TEST_P(LayerNormSolverTestFloat, LayerNormTestFw){};
+TEST_P(LayerNormTestFloat, LayerNormTestFw)
```
atamazov, line 42: We need to skip this test on unsupported GPUs.

docs/layernorm.rst
```diff
@@ -0,0 +1,17 @@
+
+Layernorm Layer
+===================
+
+The layernorm  types and functions.
```
atamazov, line 5: @seungmanhan Please mention that this is experimental functionality and the user need to define MIOPEN_BETA_API before inclusion of miopen.h to enable it.

src/include/miopen/norm/problem_description.hpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    bool IsAllPacked() const
+    {
+        if(!(xDesc.IsPacked() && weightDesc.IsPacked() && biasDesc.IsPacked() && yDesc.IsPacked() &&
+             meanDesc.IsPacked() && rstdDesc.IsPacked()))
+        {
+            MIOPEN_THROW(miopenStatusBadParm, "LayerNormForward: Unpacked tensors not supported.");
+        }
+        return true;
```
atamazov, line 113: @seungmanhan Missed it. IsApplicable should not throw. In all primitives.

PR #2486, opened 2023-10-26
Find 2.0 fusion

DrizztDoUrden:
This adds core fusion functionality to find 2.0. Currently only Convolution+activation fusions are implemented, but the rest are easy to add in separate PRs.

Current progress:
- development done
- test added to validate correctness

Removal of some code with duplicated functionality is not done in this PR, so removed lines are relatively low. In future it is possible to replace most of fusion internals with simple find 2.0 calls as they map extremely well.
The test is basically a copy of another fusion test with calls replaced to find 2.0 versions.


junliume:
@DrizztDoUrden 
```
/var/jenkins/workspace/Libs_MIOpen_ddu_find2_fused_find/src/problem.cpp:416:29: error: no viable conversion from 'const miopen::Problem' to 'ProblemContainer'

        solution.SetProblem(*this);

                            ^~~~~
```

atamazov:
@DrizztDoUrden Please fix, update from recent develop and let me know: I'll run performance tests.

atamazov:
@DrizztDoUrden Should I begin testing?

DrizztDoUrden:
Yes, I think it should pass now, you can try.

averinevg (approved):
LGTM

DrizztDoUrden:
@JehandadKhan ping

test/gtest/cba_find2_infer.cpp
```diff
@@ -0,0 +1,184 @@
...
+{
+    RunTunableSolver<miopen::solver::fusion::ConvCKIgemmFwdBiasActivFused>(
+        fused_problem, invoke_params, conv_config, test_skipped);
+}
+
+#if MIOPEN_BACKEND_HIP
+TEST_P(ConvBiasActivFind2InferTestFloatFusionFind, ConvBiasActivFind2Float_testFind)
+{
+    setEnvironmentVariable("MIOPEN_FIND_ENFORCE", "SEARCH_DB_UPDATE");
+    setEnvironmentVariable("MIOPEN_DEBUG_TUNING_ITERATIONS_MAX", "5");
```
JehandadKhan: Can this be enforced by using the context instead? We are working on making our tests packageable, which would require merging all test targets into, resulting in a single binary, which would break this logic since env vars cannot be set and unset in the same binary.

test/gtest/cba_find2_infer.cpp
```diff
@@ -0,0 +1,184 @@
...
+template <typename Solver, typename TestCase>
+void RunSolver(miopen::FusedProblem& problem,
+               const miopen::AnyInvokeParams& invoke_ctx,
+               const TestCase& conv_config,
+               bool& test_skipped)
+{
+    auto& handle = get_handle();
+    Solver solv{};
+    const auto plan           = problem.AsFusionPlan();
+    const auto fusion_problem = miopen::FusionDescription{&plan};
```
JehandadKhan:
```suggestion
    const auto fusion_desc = miopen::FusionDescription{&plan};
```

test/gtest/cba_find2.hpp
```diff
@@ -0,0 +1,187 @@
...
+            std::make_shared<miopen::ConvForwardOpDescriptor>(cfsb::conv_desc, cfsb::weights.desc);
+        auto biasOp  = std::make_shared<miopen::BiasFusionOpDescriptor>(bias.desc);
+        auto activOp = std::make_shared<miopen::ActivFwdFusionOpDescriptor>(activ_desc.GetMode());
+        EXPECT_EQ(fusePlanDesc.AddOp(convOp), miopenStatusSuccess);
+        convOp->SetArgs(params, &alpha, &beta, cfsb::wei_dev.get());
+        EXPECT_EQ(fusePlanDesc.AddOp(biasOp), miopenStatusSuccess);
+        biasOp->SetArgs(params, &alpha, &beta, bias_dev.get());
+        EXPECT_EQ(fusePlanDesc.AddOp(activOp), miopenStatusSuccess);
+        activOp->SetArgs(params, &alpha, &beta, activ_alpha, activ_beta, activ_gamma);
+         */
```
JehandadKhan:
```suggestion
```

test/gtest/cba_find2.hpp
```diff
@@ -0,0 +1,187 @@
...
+struct ConvBiasActivInferFind2Test
+    : public ::testing::TestWithParam<
+          std::tuple<miopenActivationMode_t, ConvTestCase, miopenTensorLayout_t>>,
+      ConvFwdSolverTestBase<T>
+{
+protected:
+    ConvTestCase conv_config;
+    miopen::ActivationDescriptor activ_desc;
+    tensor<T> bias;
+    miopen::Allocator::ManageDataPtr bias_dev;
```
JehandadKhan, line 55: Would it help if this were a `tensor<T>` instead so that it has a descriptor attached to it.

test/gtest/cba_find2_infer.cpp
JehandadKhan: Or perhaps use the equivalent Find 2.0 construct

test/gtest/cba_find2.hpp
```diff
@@ -0,0 +1,187 @@
...
+        EXPECT_EQ(fusePlanDesc.AddOp(biasOp), miopenStatusSuccess);
+        biasOp->SetArgs(params, &alpha, &beta, bias_dev.get());
+        EXPECT_EQ(fusePlanDesc.AddOp(activOp), miopenStatusSuccess);
+        activOp->SetArgs(params, &alpha, &beta, activ_alpha, activ_beta, activ_gamma);
+         */
+    }
+
+    void TearDown() override
+    {
+        if(test_skipped || checks_ran)
```
JehandadKhan: If the purpose of `checks_ran` is to not recompute the host side result, we can still skip the CPU computation but the verification should still happen.

JehandadKhan (dismissed):
Minor comments and questions

test/gtest/cba_find2_infer.cpp
DrizztDoUrden: Yes. This is 1 to 1 copy of another test with API swapped. I will think on how to set it through the internals and implement today.

test/gtest/cba_find2.hpp
DrizztDoUrden: You are right, will remove it. I have added a check to skip repeating cpu computation, but forgot to remove this one.

test/gtest/cba_find2.hpp
DrizztDoUrden: This is 1 to 1 copy from another test. I can change it, but the purpose was to introduce the least amount of changes possible outside of the PR scope. I can change both tests if you want.

test/gtest/cba_find2_infer.cpp
DrizztDoUrden: done

DrizztDoUrden:
Ping @JehandadKhan

junliume (approved):

PR #2489, opened 2023-10-27
[clang-tidy] Use config file for clang-tidy configuration

averinevg:
Now we configure `clang-tidy` using cmake. We could use the native config to configure the analyzer more flexibly. We currently have a number of warnings disabled; using the config, we could adjust them for better code analysis. Suppressed warnings (almost all) have been moved to the config; it is more convenient when all the settings are in one place.

list of changes:

- Configuration from the single `.clang-tidy` has been added
- `cert-dcl37-c` and `cert-dcl51-cpp` have been removed
- `readability-braces-around-statements`, `google-readability-braces-around-statements` and `hicpp-braces-around-statements` have been adjusted


CAHEK7 (approved):
Thanks a lot!

src/reducetensor_api.cpp
```diff
@@ -39,16 +39,21 @@ static void LogCmdRedux(const miopen::ReduceTensorDescriptor reduceTensorDesc,
...
+        case miopenHalf: ss << "reducefp16"; break;
+        case miopenFloat: ss << "reducefp"; break;
+        case miopenInt32: ss << "reduceint"; break;
+        case miopenInt8: ss << "reduceint8"; break;
+        case miopenBFloat16: ss << "reducebfp16"; break;
+        case miopenDouble: ss << "reducefp64"; break;
+#ifdef MIOPEN_BETA_API
+        case miopenFloat8: ss << "reducefp8"; break;
+        case miopenBFloat8: ss << "reducebfp8"; break;
+#endif
```
atamazov: MIOPEN_BETA_API should not affect implementation, only API.

atamazov (changes requested):

DrizztDoUrden (approved):
lgtm

CMakeLists.txt
```diff
@@ -559,8 +559,6 @@ enable_clang_tidy(
         -bugprone-macro-parentheses
         # too many narrowing conversions in our code
         -bugprone-narrowing-conversions
-        -cert-dcl37-c
```
JehandadKhan, line 555: @averinevg  Can we move all these to the `.clang-tidy` file ?

averinevg: Done

averinevg:
CI tests passed, please re-review

junliume (approved):

.clang-tidy
```diff
@@ -1,3 +1,192 @@
...
+# TODO Code Quality WORKAROUND ROCm 5.1 update
+#   -cert-err33-c
+#   -google-readability-casting
+#   -hicpp-use-emplace
+#   -modernize-use-emplace
+#   -performance-unnecessary-copy-initialization
+#   -readability-container-data-pointer
+#
+# TODO Code Quality WORKAROUND ROCm 5.3 && Ubuntu 22.04 && C++17 && cppcheck 2.9 update
+#   -bugprone-use-after-move
```
atamazov, line 37: [Notice] We are loosing some flexibility. For example, the comments about tidy checks cannot be placed nearby the checks anymore

CMakeLists.txt
```diff
@@ -543,161 +543,19 @@ if(CMAKE_CXX_COMPILER MATCHES ".*clang\\+\\+")
...
 elseif(MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPNOGPU")
     set(MIOPEN_TIDY_ERRORS ALL)
-
 endif()
 
 include(ClangTidy)
 enable_clang_tidy(
     CHECKS
-        *
-        -abseil-*
```
atamazov, line 545: [Notice] Before this change, is was potentially possible to switch off and on some checks depending on HIP version. Well, this possibility is kept but will require a little bit more work.

addkernels/addkernels.cpp
```diff
@@ -261,19 +267,33 @@ int main(int argsn, char** args)
...
             guard = args[++i];
+        }
         else if(arg == "n" || arg == "no-recurse")
+        {
             recurse = false;
+        }
         else if(arg == "m" || arg == "mark-includes")
+        {
             mark_includes = true;
+        }
```
atamazov, line 288: [Notice] I do not like such changes (but I am not objecting).

atamazov (approved):
No objections, just some notes to to attract some attention.

atamazov:
@averinevg Please explain the **_main goal_** of this PR in the description, for clarity. Thanks.

CMakeLists.txt
averinevg: This possibility still exists, it can be done in the same way as before

.clang-tidy
averinevg: Yes, you are right. This is difficult to do using YAML. But here at least it is possible to use comments, unlike JSON.

averinevg:
> @averinevg Please explain the **_main goal_** of this PR in the description, for clarity. Thanks.

The description has been updated

atamazov:
@averinevg Thanks! So with `.clang-tidy` we can more precisely adjust the analyzer (and improve the quality of its work), and that is the main goal, right?

I am asking because "convenience" is subjective, and "all the settings are in one place" is often less flexible (and thus may be considered as questionable/subjective as well).

averinevg:
> @averinevg Thanks! So with `.clang-tidy` we can more precisely adjust the analyzer (and improve the quality of its work), and that is the main goal, right?

Yes, you are right

PR #2493, opened 2023-10-31
Tests for RNN seq API

shurale-nkn:
Fixed several bugs in RNN:
- RNN kernel: failed if when the output buffer was dirty.
- RNN TensorBaseLayoutConverter: avoid zero size copy
- SeqTensorDescriptor: a wider range of values ( value in the seqLens now can be equal to zero) 

Created test for new API:
- all 3 layout configurations
- padding symbol
- seqTensor

Updated verification:
- added ability to compare empty containers
- removed memory leaks from dropout verification


atamazov:
@shurale-nkn Huge PR ;) Any review recommendations/hints?

shurale-nkn:
@atamazov 
> @shurale-nkn Huge PR ;) Any review recommendations/hints?

Recommendations
To ignore  test/cpu_lstm.hpp, because it was renamed to test/cpu_rnn.hpp.
To ignore first 4540 lines of  test/cpu_rnn.hpp, because this is old CPU implementation of GRU,LSTM,RNN from multiple files merged into a single header file. The rest 500 lines are APIs to call these implementations in a unified way. 

If you subtract these 4500 new lines and 1500 deleted lines, you are left with only 2500 unique lines of new code for review.

cpu_rnn_packed_ref - reference implementation only for packed layout
cpu_rnn_universal_ref - reference implementation for all 3 supported layouts (wraper for cpu_rnn_packed_ref). (converts tensor to packed layout and uses cpu_rnn_packed_ref for calculations)

atamazov:
@shurale-nkn 
> only 2500 unique lines of new code for review

😄 👍

test/driver.hpp
```diff
@@ -678,7 +678,8 @@ struct test_driver
                     std::cout << "Non finite number found in gpu at " << gpu_nan_idx << ": "
                               << out_gpu[gpu_nan_idx] << std::endl;
             }
-            else if(miopen::range_zero(out_cpu) and miopen::range_zero(out_gpu))
+            else if(miopen::range_zero(out_cpu) and miopen::range_zero(out_gpu) and
+                    (miopen::range_distance(out_cpu) != 0))
```
atamazov, line 692: 👍

src/rnn/rnn_util.cpp
```diff
@@ -311,6 +311,10 @@ void RNNTensorBaseLayoutConverter::ChangeTensorGPUDataPadding(
         const std::vector<size_t> packed_stride =
             get_packed_stride(copy_size, tensor_desc.GetLayoutVector());
 
+        // Nothing to copy, avoiding error with zero lens in TensorDescriptor
+        if(!std::all_of(copy_size.cbegin(), copy_size.cend(), [](size_t x) { return x > 0; }))
+            continue;
```
atamazov, line 322: `std::all_of` returns true if the range is empty, https://en.cppreference.com/w/cpp/algorithm/all_any_none_of

src/seq_tensor.cpp
```diff
@@ -421,22 +421,31 @@ std::vector<size_t> SeqTensorDescriptor::GetBatchesPerSequence() const
...
     else
     {
+        batches.reserve(sequence_len[0]);
         auto block_begin = sequence_len.rbegin();
-        auto sample_ptr  = sequence_len.rbegin();
-        auto batch_size  = sequence_len.size();
 
-        batches.insert(batches.end(), *block_begin, batch_size);
+        while(block_begin != sequence_len.rend() && *block_begin == 0)
+            block_begin++;
```
atamazov: [Notice] Prefer pre-increment wherever possible.

test/rnn_seq_api.cpp
```diff
@@ -0,0 +1,176 @@
...
+        this->add(this->nohx, "nohx", this->generate_data(modes));
+        this->add(this->nocx, "nocx", this->generate_data(modes));
+        this->add(this->nohy, "nohy", this->generate_data(modes));
+        this->add(this->nocy, "nocy", this->generate_data(modes));
+    }
+
+    void run()
+    {
+        if(this->nohx && this->biasMode == 1)
+            return;
```
atamazov, line 166: [Q] Is it fine that it returns silently?

test/rnn_seq_api.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+        if(this->type == miopenFloat)
+            rnn_seq_driver<T>::run();
```
atamazov, line 171: [Q] Is it fine to quietly skip running the test with `--half` etc.?

atamazov (changes requested):
Partial review

test/rnn_seq_api.hpp
```diff
@@ -0,0 +1,1663 @@
...
+
+        if(badtensor >= 0)
+        {
+            if(badtensor < 3)
+            {
+                std::cout << "FWD Train LSTM: " << std::endl;
+                switch(badtensor)
+                {
+                case(0):
+                    std::cout << "Output tensor output failed verification." << std::endl;
```
atamazov: [Q] Does this message mean the test has _Really Failed_, or is it additional information? I guess you remember that our tests grep for `FAILED`.

test/rnn_seq_api.hpp
```diff
@@ -0,0 +1,1663 @@
...
+            inputMode,
+            reserveSpace,
+            workSpace,
+            nohx);
+
+        return {std::move(dwei_data)};
+    }
+
+private:
+    bool checkSeqTensor(const miopen::SeqTensorDescriptor& desc) const
```
atamazov, line 619: [note] This function is used in debug builds only. I recommend running clang-tidy locally.

test/rnn_seq_api.hpp
```diff
@@ -0,0 +1,1663 @@
...
+};
+
+//****************************************************
+// RNN TRAIN
+//****************************************************
+
+template <class T>
+struct rnn_ref
+{
+    struct FWDResObj
```
atamazov: [highly recommended] Can we use CamelCase in the new code? FwdSomething, BwdSomething, WrwSomething...

test/rnn_seq_api.hpp
```diff
@@ -0,0 +1,1663 @@
...
+    {
+        std::vector<T> y;
+        std::vector<T> hy;
+        std::vector<T> cy;
+        FWDResObj(std::vector<T> fwd_y,
```
atamazov: [Q] Does ResObj stand for "result object"? Can we try to find better name? Basically, an instance of every class is an object, so this "Obj" looks redundant in the class name. But maybe it means something special?

test/rnn_seq_api.hpp
```diff
@@ -0,0 +1,1663 @@
...
+        check_GPU_mem_limit(
+            handle, rnnDesc, input, output, hx, cx, weights.size() * sizeof(T), statesSizeInBytes);
+
+        fill_buffers(input, dy, hx, cx, dhy, dcy, weights);
+
+        // avoid BWD unexpected fails
+        if(inVecLen == 1 && hiddenSize == 13 && seqLength == 1 && batchSize == 1)
+        {
+            tolerance = 110;
+        }
```
atamazov: [Q] This looks like a W/A for some problem. Does it worth a ticket?

atamazov (changes requested):
Another partial/superficial review

test/rnn_seq_api.hpp
atamazov: [resolved] AFAICS CI has passed.

src/rnn/rnn_util.cpp
shurale-nkn: Do you want to add a check to see if the container is empty? Because in the current format it always contains 3 elements (created at lines 307-309 ).

test/rnn_seq_api.cpp
shurale-nkn: The current implementation is incorrect for this case. The library and tests think that everything is fine, but it is not, it will be fixed later so not to increase this PR. The old tests never used nohx=true.

test/rnn_seq_api.cpp
shurale-nkn: Yes, because this test(lstm_MS_solver) is for a single path in LSTM. And this path is not used in half, because it is not so efficient for it.

test/rnn_seq_api.hpp
shurale-nkn: Yes, this is Really Failed if verbose flag is false and additional information if it's true.
I agree better to fix this here and in old RNN tests

test/rnn_seq_api.hpp
shurale-nkn: Ok. Also FwData, BwData, BwFilter of BwWeights

test/rnn_seq_api.hpp
shurale-nkn: Yes, of course you can, I'm out of ideas.
This is just a set of local classes whose task is to make verification objects more explicit and simplify interaction with them.

shurale-nkn: ResData, ResBuffers

test/rnn_seq_api.hpp
shurale-nkn: No this is WA for our data generator and our comparing function. At this combination too much information accumulated in single element and his value was around float::eps(), 
std::sqrt(square_difference) / (std::sqrt(n) * mag) => std::sqrt(square_diff(cpu,gpu)) / (1 * max(cpu,gpu))
so if we have single element in container this is now relative difference of single value, but we still using eps as threshold.
as example:
CPU = 5.960e-06, GPU = 5.9606e-06 -> std::sqrt(square_diff(cpu,gpu)) =6.0e-10; 
6.0e-10/(1 * 5.9606e-06)=1.0e4 .
And 1.0e4 bigger than eps*80 so this test failed

src/rnn/rnn_util.cpp
atamazov: [resolved] @shurale-nkn Now I see.

test/rnn_seq_api.cpp
atamazov: [resolved] @shurale-nkn Good, thanks.

test/rnn_seq_api.cpp
atamazov: @shurale-nkn If the test is not intended for half, and this path is not implemented, then maybe the test should explicitly fail when used with half?

test/rnn_seq_api.hpp
atamazov: [Notice] In convolutions, we use { Fwd, Bwd, Wrw } or { Forward, BackwardData, BackwardWeights }. To me, the shortest,  three-letter version is visually distinguishable best of all. Just FYI.

test/rnn_seq_api.hpp
atamazov: [Recommendation] Maybe just "FwdResult" or "FwdData""? So the fully qualified name would look like `rnn_ref::FwdResult` or `rnn_ref::FwdData` which seems clear enough. Up to you.

[resolved]

test/rnn_seq_api.hpp
atamazov: [Recommendation] I would add an explanatory comment (the explanation above seems perfect). Maybe it's even worth changing the tolerance from 80 to 160 (regardless of the rnn parameters) and removing the conditional. But I am not sure, especially about the latter. Up to you.

[resolved]

test/rnn_seq_api.cpp
shurale-nkn: @atamazov 
Parsing of all arguments, including the data type, takes place inside the test_drive class. 

lstm_MS_solver is a subtest for test [rnn_seq_api.cpp]. if [rnn_seq_api.cpp ] was called it always runs two subtests: rnn_seq_driver and lstm_MS_solver. This allows you not to store logic outside of the executable file in cmake. And skipping unused configurations happens relatively quickly and at no additional cost.

atamazov: @shurale-nkn 
> Parsing of all arguments, including the data type, takes place inside the test_drive class.

Of course.

> lstm_MS_solver is a subtest for test...

Normally we generate failures (throw, assert etc) on attempts to use unimplemented functionality. This case is clearly an exception which is worth documenting in comments.

shurale-nkn: @atamazov 
Oh, I see what you mean.
Let's just say it won't lead to an error, we can run the test (and rnn API) with these arguments, but then the library will choose another code path (another 'solution') that has already been tested by rnn_seq_driver class . Therefore, in order to save CI time and not to test same code twice, I skip configurations not related to 'lstm_MS_solver' code path.

data_type is the only parameter that is processed not using "struct argument" but directly in test_drive_impl_1, therefore it can only be passed at startup via cmd and cannot be generated as batch_size and others.

And yes, rnn_seq_api have some cheeks for passed parameters, but all these checks are in the general class 'rnn_seq_api_test_driver'. Limitation described in the file 'rnn_seq_api.cpp ' is only used to optimize the size of the test coverage.

test/rnn_seq_api.hpp
atamazov: @shurale-nkn "failed verification" -> "report". I am really sorry, but I still do not understand what's the meaning of the message

test/rnn_seq_api.hpp
atamazov: [resolved]

test/rnn_seq_api.hpp
atamazov: Let's cope with this later. [resolved]

atamazov (approved):
🟢 LGTM

shurale-nkn:
@junliume ping

shurale-nkn:
@junliume  This is blocker for RNN, LSTM, GRU PRs

junliume (approved):

Issue #2496, opened 2023-11-01
[Fp16] MIOpen integration or layout transpose issue with FP16_INF and FP16_MAX


junliume:
**[Observations]**
On gfx90a platform, with the input 
[input.txt](https://github.com/ROCmSoftwarePlatform/MIOpen/files/13224035/input.txt)
```
# /opt/rocm/bin/MIOpenDriver convfp16 -n 5 -c 1 -H 3 -W 3 -k 3 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 1 -V 0 --dump_output 1 --in_data input.txt.bin --dout_data dout.txt.bin
MIOpenDriver convfp16 -n 5 -c 1 -H 3 -W 3 -k 3 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 1 -V 0 --dump_output 1 --in_data input.txt.bin --dout_data dout.txt.bin
Read data from input file input.txt.bin
Could not open file dout.txt.bin for reading
PRNG seed: 12345678
Wrote output to file dump_in.bin
Wrote output to file dump_dout.bin
MIOpen Backward Weights Conv. Algorithm: 5, Solution: 110/ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC
GPU Kernel Time Backward Weights Conv. Elapsed: 0.024996 ms (average)
stats: name, n, c, ho, wo, x, y, k, flopCnt, bytesRead, bytesWritten, GFLOPs, GB/s, timeMs
stats: bwdw-conv1x1u1, 5, 1, 3, 3, 1, 1, 3,  270, 0, 0, 0, 0, 0.024996
Wrote output to file dump_bwd_dwei_gpu.bin

```
The IGEMM kernel gave output:

> [65504. 65504. 65504.]

But the expected output (from direct convolution) should be:

> [inf, inf, inf]

**[Analysis]**

- 65504 is FP16_MAX
- It is determined that the iGEMM kernel actually generates `INF`
- @JehandadKhan @atamazov Now it is suspected that during data-layout transpose or MIOpen integration, the INF is reinterpreted as MAX?

@carlushuang  could you help to comment on this issue if anything is missing? Thanks!

carlushuang:
hotfix is here:
[...]

PR #2497, opened 2023-11-01
workaround SWDEV_423713 by disabling the unit test case in wrw solver

carlushuang:
#2496


atamazov (approved):
Technically, this is good!

junliume (approved):

PR #2498, opened 2023-11-01
Add checks on workspace params

amberhassaan:
Throw exception if workspace pointer and size are not in agreement, e.g., if the pointer is null, size should be zero and vice versa.


atamazov:
@amberhassaan Why this is necessary? Perhaps you have encountered some use case where such a check could have helped if it existed?

src/ocl/convolutionocl.cpp
```diff
@@ -260,6 +270,7 @@ void ConvolutionDescriptor::FindConvFwdAlgorithm(Handle& handle,
                                                  size_t workSpaceSize,
                                                  bool exhaustiveSearch) const
 {
+    ValidateWorkspace(workSpace, workSpaceSize);
```
atamazov: Let's move this after logging everywhere.

amberhassaan:
> @amberhassaan Why this is necessary? Perhaps you have encountered some use case where such a check could have helped if it existed?

Yes, I'm still debugging it but I have run into similar situations in the course of debugging `test_conv3d` for #2429. It seems that such a scenario is possible in `test/conv_common.h` , e.g., this [line](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/4e37509aae8eec891637bab05de8a4210f998a78/test/conv_common.hpp#L843), where we reset the pointer but not the size.

atamazov (changes requested):

atamazov:
@amberhassaan 

> Yes, I'm still debugging it but I have run into similar situations in the course of debugging `test_conv3d` for #2429. It seems that such a scenario is possible in `test/conv_common.h` , e.g., this [line](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/4e37509aae8eec891637bab05de8a4210f998a78/test/conv_common.hpp#L843), where we reset the pointer but not the size.

Yeah, we should pass perf.memory to `ConvolutionForward` at https://github.com/ROCmSoftwarePlatform/MIOpen/blob/4e37509aae8eec891637bab05de8a4210f998a78/test/conv_common.hpp#L861

atamazov:
@amberhassaan However I do not see how this could lead to any problem in this specific case. We are passing enough workspace for the solver reported by `FindConvFwdAlgorithm`. Even if we pass ~incorrect~ bigger size, the solver won't try to use anything beyond it's workspace requirements. Please keep me informed, thanks.

atamazov:
@amberhassaan WRT this PR -- the check is incomplete anyway. Actually we can implement a full blown check: store the required workspace size in each invoker instance and check it there. However, when we considered whether it is worth doing such a check (a while ago), we decided not to. After all, it is impossible to make similar checks for tensors. So does it worth messing with workspace in each invoker?

Checking implemented in this PR is much simpler but still looks helpful. The question is whether we should always have it enabled. IIRC the library API does not formally require consistency of workspace size and ptr when a solver does not need any workspace. If a solver does not require workspace, the library ignores ws size and ptr, so passing any junk is legal. Shall we make this check under some debugging option (env var or only activate it in the debug builds), what do you think?

I am afraid of regressions with frameworks during QA testing.

/cc @JehandadKhan

amberhassaan:
> IIRC the library API does not formally require consistency of workspace size and ptr when a solver does not need any workspace. If a solver does not require workspace, the library ignores ws size and ptr, so passing any junk is legal. Shall we make this check under some debugging option (env var or only activate it in the debug builds), what do you think?

I'm OK with making it a `debug` mode only option, i.e. `assert()` statements. Having this check has helped me debug #2429 and uncover a possible bug with `Find API` when it comes to the workspace size.

CAHEK7:
Even if some solvers do not require a workspace at all they may work as a good checker as well, since when you run exhaustive search for the solver, it may choose either one with a workspace or another one without it, and thus we may pass garbage to the solver with workspace or bypass this check by the solver without.

From the other hand - if we don't have any consistency requirements for the workspace configuration, we can check only explicitly bad cases - either the size is less than required or the pointer is not null if the workspace is required. The other cases can be just logged as warnings or even info messages (and probably fail the tests in debug mode, but that's optional)

CAHEK7 (commented):
Technically it seem fine, but probably it should be extended to everything which requires workspace. (I mean not only convolutions)
And we should make an agreement what kind of checks are strict and failed the process and what kind of check are informational only.

JehandadKhan:
@atamazov  

> so passing any junk is legal. 

I am not sure letting libraries pass in Junk is a good idea, it can lead to inconsistent expectation from the caller. 

> Shall we make this check under some debugging option (env var or only activate it in the debug builds)

If we are not happy with the current checks and are concerned that it might have performance implications, then I would suggest we use asserts since they are checked as part of the nightly builds. An environment variable means we would need to 

* Document it 
* Have a test which ensure consistent behavior when it is enabled and when it is disabled.

src/ocl/convolutionocl.cpp
amberhassaan: done

amberhassaan:
@JehandadKhan @atamazov : I have converted the checks to `asserts` and moved the check after the `LOG` line.

atamazov:
@JehandadKhan 
> If we are not happy with the current checks and are concerned that it might have performance implications

What I really fear is the crashes of existing networks that would immediately lead to P0 tickets.

/cc @junliume

atamazov:
@amberhassaan
> I have converted the checks to asserts and moved the check after the LOG line.

Do you mean that PR is ready for the next review round?

amberhassaan:
> @amberhassaan
> 
> > I have converted the checks to asserts and moved the check after the LOG line.
> 
> Do you mean that PR is ready for the next review round?

Yes, ready for re-review.

atamazov (approved):
LGTM!

atamazov:
@JehandadKhan @amberhassaan 
> I would suggest we use asserts since they are checked as part of the nightly builds. An environment variable means we would need to... Document it

💡 BTW we can control this check by MIOPEN_CHECK_NUMERICS=0x20. Only one line of the documentation needs to be added. The advantage is that we'll be able to enable checks without rebuilding of the library (which may be especially useful when we take the 2rd party docker images for investigation). And our clients and QA will be able to do the same. Just an idea.

CAHEK7 (approved):

JehandadKhan (approved):

JehandadKhan:
@junliume  We can merge this when you get a chance.

junliume (approved):

PR #2506, opened 2023-11-03
Fix for end of line issue on Windows

saeid-rostami:
When reading from a text file on Windows line ending is CRLF, unlike on Linux whish is LF.
The code did not properly handle the line ending on Windows due to discrepancy in the interpretation of line endings between the tellg() function and the C++ runtime.
tellg() function considers CRLF (Carriage Return + Line Feed) as two characters, but when filling the buffer,  C++ runtime interprets CRLF as a single character. As a result, the buffer contained incorrect data at the end of the buffer.
To address this issue, files are opened in binary mode to ensure that line ending is LF in both Windows and Linux.

### DIFF
--- a/src/db.cpp
+++ b/src/db.cpp
@@ -149,7 +149,7 @@ boost::optional<DbRecord> PlainTextDb::FindRecordUnsafe(const std::string& key,
 
     MIOPEN_LOG_I2("Looking for key " << key << " in file " << filename);
 
-    std::ifstream file(filename);
+    std::ifstream file(filename, std::ios::binary);
 
     if(!file)
     {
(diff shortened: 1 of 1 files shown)


atamazov:
@DrizztDoUrden Can you please look at this.

DrizztDoUrden (approved):

atamazov:
@saeid-rostami I am not sure that if using [`uif2-initial`](https://github.com/ROCmSoftwarePlatform/MIOpen/tree/uif2-initial) as a baseline is a good idea. You may with to rebase this change on top of `develop`. That would allow merging your fix into the main development branch of MIOpen sooner than `uif2-initial` is merged in. But this is up to you, of course.

saeid-rostami:
> 

@atamazov , I'll send another PR for develop branch. Thanks.

junliume:
@saeid-rostami with #2515 up, should we close this one? if this is needed for `uif2-initial` branch which is also understandable, then let us know.

atamazov:
@junliume `uif2-initial` is only 7 commits behind develop and can be easily updated from there, I guess ;)

saeid-rostami:
@junliume , Yes we can close this one. Already sent another one to develop branch.

Issue #2492, opened 2023-10-30
[PyTorch][Winograd] Winograd kernel been selected has caused issue with test_Conv2d_naive_groups_cuda_float16


junliume:
**[Summary]**

Winograd kernels are by design aiming performance by sacrificing numerical accuracy.
However, in this case for very small and non-practical case, selecting `winograd` kernels have caused `test_Conv2d_naive_groups_cuda_float16` to fail.

Question:

- `test_Conv2d_naive_groups_cuda_float16` has keyword `naive` in it, does it expect naive implementations to begin with?
- @Kirpich30000 should `winograd` kernels have issues with such cases? i.e. `-H 6 -W 6 -k 2 `

```
MIOpenDriver convfp16 -n 2 -c 2 -H 6 -W 6 -k 2 -y 3 -x 3 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 -S 0
Forward Conv solutions available: 2
- id: 84 algo: 3, time: 10 ms, ws: 0, name: ConvBinWinogradRxSf2x3g1
- id: 107 algo: 5, time: 20 ms, ws: 1280, name: ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC
MIOpen Forward Conv. Algorithm: 3, Solution: 84/ConvBinWinogradRxSf2x3g1
GPU Kernel Time Forward Conv. Elapsed: 0.015378 ms (average)
stats: name, n, c, ho, wo, x, y, k, flopCnt, bytesRead, bytesWritten, GFLOPs, GB/s, timeMs
stats: fwd-conv3x3u1, 2, 2, 4, 4, 3, 3, 2,  2304, 360, 128, 0, 0, 0.015378
Forward Convolution Verifies OK on CPU reference (0.000340009)
```

**[Observation and Steps to reproduce]:**

To Reproduce:

PYTORCH_TEST_WITH_ROCM=1 python3 nn/test_convolution.py --use-pytest --verbose -k test_Conv2d_naive_groups_cuda_float16
Docker Images:
```
ROCM 5.6: rocm/pytorch:rocm5.6_ubuntu20.04_py3.8_pytorch_2.0.1
PyTorch Installed at /var/lib/jenkins/pytorch/test
[...]

PR #2507, opened 2023-11-04
Workaround for issue #2492 - disable ConvBinWinoRxS when granularity loss is huge

atamazov:
This is expected to resolve issue #2492, at least for some time (see https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2492#issuecomment-1793570143, https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2492#issuecomment-1793574432 for more info).

- The W/A disables ConvBinWinoRxS when granularity loss is > 0.995 (performance drops 200 times)
- The W/A can be switched OFF by setting `MIOPEN_DEBUG_WORKAROUND_ISSUE_2493=0` in the env.

----
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/workaround
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/urgency_high
- https://github.com/ROCmSoftwarePlatform/MIOpen/labels/non-miopen-bug
- Proposed reviewers:
  - @JehandadKhan
  - @Kirpich30000 
  - @wenchenvincent
  - @CAHEK7

### DIFF
--- a/src/solver/conv_winoRxS.cpp
+++ b/src/solver/conv_winoRxS.cpp
@@ -52,6 +52,20 @@
 // we will keep ConvBinWinoRxS<2,3> for group convolutions only.
 #define WORKAROUND_ISSUE_1681 0
 
+/// \anchor disable_winograd_with_huge_granularity_loss
+/// This is to exclude Winograd from PyTorch correctness tests.
+/// These tests are written in incorrect manner.
+/// They are very sensitive to the values in the output tensors,
+/// but do not provide inputs that guarantee minimal deviation
+/// (a numerical difference between kernel output and theoretically
+/// "ideal" output). This is especially important for Winograd.
+/// As a result, the PyTorch correctness tests often produce false
+/// failures when Winograd algorithm is used. As PyTorch correctness
+/// tests use very small convolutions and Winograd algorithm is
+/// ineffective with such small configs due to huge granularity loss,
+/// we can disable Winograd without any performance implications.
+#define WORKAROUND_ISSUE_2493 1
+
 MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_AMD_WINOGRAD_RXS_F2X3)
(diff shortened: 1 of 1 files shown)


junliume:
@atamazov  @Kirpich30000 
I am worried about performance impact on this change:
```
 disable ConvBinWinoRxS when granularity loss is > 0.995 (performance drops 200 times)
```
especially if that could be the expected behavior of Winograd kernels in certain cases.

atamazov:
@junliume "200x performance drop" are cases when Winograd is unable to utilize GPU efficiently, and it's performance is expected to be so low that it makes no sense to use it.

src/solver/conv_winoRxS.cpp
```diff
@@ -656,7 +679,15 @@ static bool IsApplicableBase(const ExecutionContext& ctx, const ProblemDescripti
         && problem.GetBias() == 0
         && problem.GetInLayout() == "NCHW"))
         return false;
-    // clang-format on
+        // clang-format on
+
+#if WORKAROUND_ISSUE_2493
+    if(!miopen::IsDisabled(MIOPEN_DEBUG_WORKAROUND_ISSUE_2493{}))
+    {
+        if(ShaderModel(ctx, problem, Winodata, Winofilter).GetGranularityLoss() > 0.995)
```
JehandadKhan, line 687: Perhaps we can move the constant to a define or a `constexpr` variable so that the limit is documented ?

JehandadKhan (approved):
One minor nitpick, otherwise looks fine.

junliume (approved):

src/solver/conv_winoRxS.cpp
atamazov: Resolved in PR #2510, commit f186fafede484eff08277fa2392413720df3916c.

atamazov:
@junliume @JehandadKhan Thanks but we were too fast ;) Please look at the 2nd part, PR #2510.

Issue #2502, opened 2023-11-03
[rocTracer] Protect dependency on roctx64


junliume:
Introduced by https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2391

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/19c6046d1a1d2ac7123fbeaf7775dd7a5749bfd3/src/CMakeLists.txt#L849

Maybe the line is better with
```
    find_library(LIBTX64 roctx64)
    if(LIBTX64)
        MESSAGE(STATUS "Libtx64: " ${LIBTX64})
        target_internal_library(MIOpen ${LIBTX64})
    endif()
```

There might be better design when rocTracer is not installed, I think we should protect the case somehow.

Other related issues: the **NEEDED** below needs to be explicitly set as dependencies for MIOpen package
```
➜  build git:(develop) ✗ objdump -x lib/libMIOpen.so | grep -A 15 "Dynamic Section"
Dynamic Section:
  RUNPATH              /opt/rocm/lib:/opt/rocm-5.7.0/lib::/opt/rocm/lib
  NEEDED               libamd_comgr.so.2
  NEEDED               librocblas.so.3
  NEEDED               libroctx64.so.4
  NEEDED               libamdhip64.so.5
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
  NEEDED               ld-linux-x86-64.so.2
  SONAME               libMIOpen.so.1
  RELA                 0x00000000002710c0
  RELASZ               0x00000000006364f8
  RELAENT              0x0000000000000018
  RELACOUNT            0x000000000003c97b
```

junliume:
@atamazov  @JehandadKhan could you take a look?

PR #2508, opened 2023-11-07
Fix miopen package dependency roctracer etc

junliume:
https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2502

Hence the issue:
```
➜  build git:(develop) ✗ objdump -x lib/libMIOpen.so | grep -A 15 "Dynamic Section"
Dynamic Section:
  RUNPATH              /opt/rocm/lib:/opt/rocm-5.7.0/lib::/opt/rocm/lib
  NEEDED               libamd_comgr.so.2 <=== amd_comgr
  NEEDED               librocblas.so.3 <=== rocBLAS
  NEEDED               libroctx64.so.4 <=== rocTracer
  NEEDED               libamdhip64.so.5
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
  NEEDED               ld-linux-x86-64.so.2
  SONAME               libMIOpen.so.1
  RELA                 0x00000000002710c0
  RELASZ               0x00000000006364f8
  RELAENT              0x0000000000000018
  RELACOUNT            0x000000000003c97b

```

Before the fix:
```
➜  build git:(develop) ✗ sudo dpkg -I ./miopen-hip_3.00.0-c1e631d8a_amd64.deb | grep Depends
 Depends: hip-rocclr, rocblas, rocm-core
```
After the fix:
```
➜  build git:(dep_roctracer) ✗ sudo dpkg -I ./miopen-hip_3.00.0-1a5c0dd44_amd64.deb | grep Depends
 Depends: hip-rocclr, amd_comgr, roctracer, rocblas, rocm-core
```

```

### CODE REVIEW

File: CMakeLists.txt [hunk 1]
```diff
@@ -418,7 +420,10 @@ else()
 endif()
 set(MIOPEN_SYSTEM_FIND_DB_SUFFIX "${MIOPEN_BACKEND}" CACHE PATH "Filename suffix for the system find-db files")
 
-set(MIOPEN_PACKAGE_REQS "hip-rocclr")
+# PR-2391 Add the ability to log function calls to roctx.
+# This allows attached profilers to see which MIOpen calls are being called by application and which kernels are being invoked by MIOpen.
+# Enabled via the MIOPEN_ENABLE_LOGGING_ROCTX env var.
+set(MIOPEN_PACKAGE_REQS "${MIOPEN_PACKAGE_REQS}, roctracer")
```
apwojcik, line 442: Please exclude it from Windows. The `roctracer` is unavailable.

junliume: Updated

junliume:
@averinevg @atamazov @JehandadKhan Ping for review

CMakeLists.txt
```diff
@@ -418,7 +420,12 @@ else()
 endif()
 set(MIOPEN_SYSTEM_FIND_DB_SUFFIX "${MIOPEN_BACKEND}" CACHE PATH "Filename suffix for the system find-db files")
 
-set(MIOPEN_PACKAGE_REQS "hip-rocclr")
+# PR-2391 Add the ability to log function calls to roctx.
+# This allows attached profilers to see which MIOpen calls are being called by application and which kernels are being invoked by MIOpen.
+# Enabled via the MIOPEN_ENABLE_LOGGING_ROCTX env var.
+if(NOT WIN32)
+    set(MIOPEN_PACKAGE_REQS "${MIOPEN_PACKAGE_REQS}, roctracer")
```
atamazov: [Question] @junliume This ~makes~ means that roctracer and roctx64 are mandatory. Is it possible that these libraries are not available on the target system (for example due to outdated ROCm, which we would still like to support)?

atamazov (approved):
No objections!

CMakeLists.txt
junliume: @atamazov thanks and I have the same concern. Dependency on `dev` tools should be "optional" IMHO. However, https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2502 originated from a UIF docker issue where only during runtime the missing dependency is reported. i.e. I think we should not build two sets of packages, one depends on rocTracer while the other does not.
Meanwhile, I think we can and should put a version check here, like
```
if(NOT WIN32 && MIOPEN_VERSION >= 3.0) 
```
**update 1**: second thought, this version check is not necessary since the changes are part of MIOpen itself.
**update 2**: still thinking ;-P I will update this PR later today

atamazov: @junliume We can add CMake flag or option and pass it into our sources via config.txt (like MIOPEN_USE_COMGR). We can use it as a workaround if libs are missing.

junliume:
@atamazov PR updated :)
Now one missing functionality: if MIOpen is built without rocTracer, but the user tries to use `MIOPEN_ENABLE_LOGGING_ROCTX` env var: we should emit a warning once instead of ignoring it silently, and we should minimize its cost. 
I need some advice from you and others how to do that properly.

atamazov:
@junliume BTW just run into this with ROCm 5.1 ;)
```
[ 58%] Linking CXX shared library ../lib/libMIOpen.so
ld.lld: error: unable to find library -lroctx64
```

junliume:
> @junliume BTW just run into this with ROCm 5.1 ;)
> 
> ```
> [ 58%] Linking CXX shared library ../lib/libMIOpen.so
> ld.lld: error: unable to find library -lroctx64
> ```

This PR is trying to resolve this exact issue :) Now updated with more changes.

atamazov:
> @atamazov 
> I need some advice from you and others how to do that properly.

The simplest thing: Instead of warning we should FAIL the build if the library needs roctracer libs but unable to find them. The user will then be forced to explicitly set `-DMIOPEN_USE_ROCTRACER=Off` and will not expect it to work (including `MIOPEN_ENABLE_LOGGING_ROCTX`).

BTW failing the build if something is required but not found is a mandatory requirement.

src/include/miopen/logger.hpp
```diff
@@ -39,8 +39,10 @@
 #include <miopen/config.h>
 
 #ifndef _WIN32
```
atamazov: this _WIN32 check is no longer needed

src/include/miopen/logger.hpp
```diff
@@ -331,6 +336,22 @@ LogParam(std::ostream& os, std::string name, const std::vector<T>& vec, bool ind
...
+                               << std::endl;                                            \
+            std::cerr << miopen_log_func_ss.str();                                      \
+            MIOPEN_PP_EACH_ARGS(MIOPEN_LOG_FUNCTION_EACH, __VA_ARGS__)                  \
+            std::ostringstream().swap(miopen_log_func_ss);                              \
+            miopen_log_func_ss << miopen::LoggingPrefix() << "}" << std::endl;          \
+            std::cerr << miopen_log_func_ss.str();                                      \
+        }                                                                               \
+    while(false)
+#endif
+#else
```
atamazov: Can you try this apptoach:
```C++
#if MIOPEN_USE_ROCTRACER
#define MIOPEN_LOG_ROCTX_DEFINE_OBJECT miopen::LogScopeRoctx logtx;
#define MIOPEN_LOG_ROCTX_DO_LOGGING(...)                                                \
        if(miopen::IsLoggingToRoctx())                                                  \
        {                                                                               \
            std::ostringstream miopen_log_func_ss;                                      \
            miopen_log_func_ss << "s_api = " << __FUNCTION__ << " | ";                  \
            MIOPEN_PP_EACH_ARGS(MIOPEN_LOG_FUNCTION_EACH_ROCTX, __VA_ARGS__)            \
            logtx.logRange(miopen_log_func_ss.str());                                   \
        }
#else
#define MIOPEN_LOG_ROCTX_DEFINE_OBJECT
#define MIOPEN_LOG_ROCTX_DO_LOGGING(...)
#endif

#define MIOPEN_LOG_FUNCTION(...)                                                        \
    MIOPEN_LOG_ROCTX_DEFINE_OBJECT                                                      \
    do                                                                                  \
    {                                                                                   \
        if(miopen::IsLoggingFunctionCalls())                                            \
        {                                                                               \
            std::ostringstream miopen_log_func_ss;                                      \
            miopen_log_func_ss << miopen::LoggingPrefix() << __PRETTY_FUNCTION__ << "{" \
                               << std::endl;                                            \
            std::cerr << miopen_log_func_ss.str();                                      \
            MIOPEN_PP_EACH_ARGS(MIOPEN_LOG_FUNCTION_EACH, __VA_ARGS__)                  \
            std::ostringstream().swap(miopen_log_func_ss);                              \
            miopen_log_func_ss << miopen::LoggingPrefix() << "}" << std::endl;          \
            std::cerr << miopen_log_func_ss.str();                                      \
        }                                                                               \
        MIOPEN_LOG_ROCTX_DO_LOGGING(__VA_ARGS__)                                        \
    } while(false)
#else
#define MIOPEN_LOG_FUNCTION(...)
#endif
```
Cumbersome, but no copy-paste.

src/include/miopen/logger.hpp
```diff
@@ -413,6 +434,7 @@ class LogScopeTime
 #endif
 
 #ifndef _WIN32
```
atamazov: _WIN32

test/gtest/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
-find_package(GTest CONFIG REQUIRED)
+find_package(GTest CONFIG)
 if(GTest_FOUND)
   MESSAGE(STATUS "GoogleTest: " ${GTest})
-  enable_testing()
   include(GoogleTest)
+else()
+  message(WARNING "GoogleTest is not found! Skip GTests ... try install GTest or MIOpen dependencies.")
```
atamazov: I think we can introduce something like `MIOPEN_USE_GTEST`, make it enabled by default and fail if Gtest is not found. Then the user can explicitly disable or install GTest.

atamazov (changes requested):

src/include/miopen/logger.hpp
atamazov: EDIT: `MIOPEN_LOG_ROCTX_DO_LOGGING` -> `MIOPEN_LOG_ROCTX_DO_LOGGING(...)`

test/gtest/CMakeLists.txt
junliume: `GTest`, like `rocTracer` are development tools instead of hard dependencies of the library. How about we deal with them in separate PR? I will adopt the other review options in this PR.

atamazov: If GTest is broken, then CI will silently omit many tests.

JehandadKhan: I suggest we let the default `MIOpen` and `MIOpenDriver` target build without the GTest dependency, however, if tests are being built then GTest must be present otherwise compile will fail.

atamazov: +1

CAHEK7 (approved):

test/gtest/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
-find_package(GTest CONFIG REQUIRED)
+find_package(GTest CONFIG)
 if(GTest_FOUND)
   MESSAGE(STATUS "GoogleTest: " ${GTest})
-  enable_testing()
   include(GoogleTest)
+else()
+  message(WARNING "GoogleTest is not found! Skip GTests ... try install GTest or preferably follow instructions to install MIOpen dependencies.")
+  return()
```
CAHEK7: Is this change related to the PR?

junliume: Not directly, a byproduct of the review I believe.
Update: updated from `develop` and make GTest a required package

PR #2515, opened 2023-11-08
fix for end of line issue on Windows

saeid-rostami:
When reading from a text file on Windows line ending is CRLF, unlike on Linux whish is LF.
The code did not properly handle the line ending on Windows due to discrepancy in the interpretation of line endings between the tellg() function and the C++ runtime.
tellg() function considers CRLF (Carriage Return + Line Feed) as two characters, but when filling the buffer, C++ runtime interprets CRLF as a single character. As a result, the buffer contained incorrect data at the end of the buffer.
To address this issue, files are opened in binary mode to ensure that line ending is LF in both Windows and Linux.

### DIFF
--- a/src/db.cpp
+++ b/src/db.cpp
@@ -149,7 +149,7 @@ boost::optional<DbRecord> PlainTextDb::FindRecordUnsafe(const std::string& key,
 
     MIOPEN_LOG_I2("Looking for key " << key << " in file " << filename);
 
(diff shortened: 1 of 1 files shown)


saeid-rostami:
@atamazov

DrizztDoUrden (approved):

junliume (approved):

PR #2517, opened 2023-11-09
Fused solver for Fwd Convolution with Residual add, Bias add and then activation function

amberhassaan:
Will use recently added CK kernel instances for this fused operation. 

Initial CK PR: https://github.com/ROCm/composable_kernel/pull/1006
Bias Fixup CK PR: https://github.com/ROCm/composable_kernel/pull/1081


src/solver.cpp
```diff
@@ -585,6 +585,11 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
              Primitive::Fusion,
              fusion::ConvCKIgemmFwdBiasActivFused{}.SolverDbId(),
              miopenConvolutionAlgoImplicitGEMM);
+    Register(registry,
```
atamazov: New solvers must be added to the end of the list.

atamazov (changes requested):
🔴 Partial review

src/solver.cpp
amberhassaan: Done.

junliume:
@atamazov  re-request review before merging this PR

atamazov:
@DrizztDoUrden @CAHEK7 Please review ASAP, if possible.

src/fusion.cpp
```diff
@@ -512,12 +548,24 @@ miopenStatus_t ConvForwardOpDescriptor::GetOutputDesc(TensorDescriptor& output_d
         [&]() { output_desc = base_desc.GetForwardOutputTensor(input_desc, filter_desc); });
 }
 
+/*
```
bghimireamd, line 551: dead code?

src/include/miopen/fusion.hpp
```diff
@@ -215,6 +225,7 @@ struct ConvForwardOpDescriptor : FusionOpDescriptor
           conv_compiler_options(""){};
     miopenStatus_t GetOutputDesc(TensorDescriptor& output_desc) const override;
     miopenStatus_t SetArgs(OperatorArgs& args, const void* alpha, const void* beta, ConstData_t w);
+    // miopenStatus_t SetArgs(OperatorArgs& args, float alpha, float beta, ConstData_t w);
```
bghimireamd, line 228: dead code?

src/solver/conv_ck_igemm_fwd_bias_res_add_activ_fused.cpp
```diff
@@ -0,0 +1,481 @@
...
+                          data_ctx.in,
+                          conv_param.weights,
+                          data_ctx.out,
+                          z_param.tensor_ptr,
+                          bias_param.bdata,
+                          conv_param.alpha,
+                          z_param.alpha);
+    }
+
+#if 0
```
bghimireamd, line 192: dead code?

JehandadKhan (approved):

src/fusion.cpp
junliume: @amberhassaan maybe kept for debugging purposes?

junliume:
Ping @bghimireamd @DrizztDoUrden @CAHEK7 for review before merging.

junliume (approved):

src/solver/conv_ck_igemm_fwd_bias_res_add_activ_fused.cpp
```diff
@@ -0,0 +1,481 @@
...
+        ck::tensor_operation::element_wise::PassThrough,
+        ck::tensor_operation::element_wise::
+            ScaleAddScaleAddRelu>>; // end DeviceOperationInstanceFactory
+
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+namespace {
+
+struct CKArgs
+{
+    CKArgs(const ProblemDescription& problem)
```
junliume, line 75: @amberhassaan @JehandadKhan @atamazov weird how this get passed CI without 
either
`CKArgs(const miopen::conv::ProblemDescription& problem)`
or 
`using ProblemDescription = miopen::conv::ProblemDescription;`
at the top of file

atamazov:
@junliume @JehandadKhan The build with `-DMIOPEN_USE_COMPOSABLEKERNEL=Off` is now broken in develop.

src/solver/conv_ck_igemm_fwd_bias_res_add_activ_fused.cpp
averinevg: After merging https://github.com/ROCm/MIOpen/pull/2503, the code previously needed for compatibility with Fin was removed. It was intended that only Fin would use it, but it turned out that it was used in other places too. I suspect this was because work on this pull request had started before the `ProblemDescription` was changed.

src/fusion.cpp
amberhassaan: yes, this should be removed. @iq136boy commented it out when I wasn't paying attention :). 
In reality, we need to deduce the pointer type from `ConvolutionDescriptor` and not assume `float` all the time, but for now CK seems to support `float` only for these scaling factors.

src/include/miopen/fusion.hpp
amberhassaan: ditto. Will remove in a future PR.

amberhassaan:
@junliume : thank you for taking this PR past the finish line.

atamazov:
@amberhassaan @junliume 🔴 Navi21, ROCm 6.0, develop 2b9d9cdbb4b18a5bcec5603b6dfc9f3a7b58b6c7, **_"No supported fusion solvers found"_**, make check fails:
```bash
 833/1799 Test  #835: ConvFwdBiasActivAPI/ConvFwdBiasResAddActivTest.ConvFusedAPI/(5,  G:1 N:1 C:4 D:14 H:11 W:1 k:4 z:3 y:3 x:3 pad_z:1 pad_y:1 pad_x:1 stride_z:1 stride_y:1 stride_x:1 dilation_z:1 dilation_y:1 dilation_x:1 conv_mode:0, 1, 1, 8) ........................................................................................................................................................................................***Failed    3.48 sec
Running main() from /opt/rocm/cget/build/tmp-c4f1a42bcee536579a1248d25169f834201170c187fc6d80e8c510a7d28f6f48/googletest-1.14.0/googletest/src/gtest_main.cc
Note: Google Test filter = ConvFwdBiasActivAPI/ConvFwdBiasResAddActivTest.ConvFusedAPI/0
[==========] Running 1 test from 1 test suite.
[----------] 1 test from ConvFwdBiasActivAPI/ConvFwdBiasResAddActivTest
[ RUN      ] ConvFwdBiasActivAPI/ConvFwdBiasResAddActivTest.ConvFusedAPI/0
PRNG seed: 12345678
MIOpen(HIP): Info [Handle] stream: 0x2654620, device_id: 0
MIOpen(HIP): Info [AmdRocmMetadataVersionDetect] ROCm MD version AMDHSA_COv3, HIP version 6.0.23494, MIOpen version 3.1.0.2b9d9cdbb
MIOpen(HIP): Info [Measure] RamDb::Prefetch time: 0.037761 ms
MIOpen(HIP): Info [TryLoad] Find-db regenerating.
MIOpen(HIP): Info [AllocateBuffersAndMakeFusionInvokeParams] bias addr: 0x7f1f09a05000, size: 8
MIOpen(HIP): Info [AllocateBuffersAndMakeFusionInvokeParams] weight addr: 0x7f1f09a06000, size: 864
MIOpen(HIP): Info [AllocateBuffersAndMakeFusionInvokeParams] in addr: 0x7f1f09a08000, size: 1232
MIOpen(HIP): Info [AllocateBuffersAndMakeFusionInvokeParams] out addr: 0x7f1f09a09000, size: 1232
MIOpen(HIP): Info [GetPerfDbPathFile] Found exact perf database file
MIOpen(HIP): Info [Compile] No supported fusion solvers found
...
/root/github/MLOpen1/test/gtest/fused_conv_bias_res_add_activ.cpp:185: Failure
Expected equality of these values:
  status
    Which is: 8
  miopenStatusSuccess
    Which is: 0

MIOpen(HIP): Info [PrintVersion] HIPRTC v.9.0
/root/github/MLOpen1/test/gtest/fused_conv_bias_res_add_activ.cpp:130: Failure
Value of: error < threshold
  Actual: false
Expected: true
Error beyond tolerance Error:nan,  Threshold: 0.078125

[  FAILED  ] ConvFwdBiasActivAPI/ConvFwdBiasResAddActivTest.ConvFusedAPI/0, where GetParam() = (5,  G:1 N:1 C:4 D:14 H:11 W:1 k:4 z:3 y:3 x:3 pad_z:1 pad_y:1 pad_x:1 stride_z:1 stride_y:1 stride_x:1 dilation_z:1 dilation_y:1 dilation_x:1 conv_mode:0, 1, 1, 8) (3367 ms)
```

junliume:
@amberhassaan @atamazov @JehandadKhan I think this test should not have been enabled on gfx103x or gfx110x platforms.
https://github.com/ROCm/composable_kernel/pull/1081/files#diff-2ac9aab19c7443bbb77ad3e540bff2e78dd604943cc9327d3bc90fbe5e7a83ea uses features not yet available on these platforms.

However,
https://github.com/ROCm/MIOpen/blob/ae2682c2befa29c1c5704bd2f89817e817d913db/src/solver/conv_ck_igemm_fwd_bias_res_add_activ_fused.cpp#L421-L422
should have returned that solver is not applicable.

We should skip the test is detected inapplicable, like
```
    if(!solv.IsApplicable(ctx, problem))
    {
        test_skipped = true;
        GTEST_SKIP() << solv.SolverDbId()
                     << "ConvHipImplicitGemm3DGroupBwdXdlops Not Applicable for this problem"
                     << conv_config;
    }
```

src/solver/conv_ck_igemm_fwd_bias_res_add_activ_fused.cpp
```diff
@@ -0,0 +1,481 @@
...
+        return false;
+    if(conv_problem.HasNonPackedTensors())
+        return false;
+    if(conv_problem.HasMixedDataTypes())
+        return false;
+    if(!(conv_problem.Is2d() || conv_problem.Is3d()))
+        return false;
+    if(!conv_problem.IsLayoutNHWC())
+        return false;
+    if(!ck_utility::is_ck_whitelist(ctx.GetStream().GetDeviceName()))
```
junliume, line 420: @amberhassaan maybe we should add here?
```
    if(miopen::StartsWith(arch, "gfx11") || miopen::StartsWith(arch, "gfx10"))
        return false;
```

atamazov:
@junliume 
> We should skip the test is detected inapplicable, like...

Basically, not. Otherwise the test won't detect the cases when solver's `IsApplicable()` is broken due to some reason. All the stuff like 
```C++
if(!solv.IsApplicable(ctx, problem))
    GTEST_SKIP();
```
should be removed from the tests, unless clearly explained why it is absolutely necessary.

atamazov:
[Recommendation] @junliume can you please move the discussion to #2633? I can't as I do not have enough privileges.

PR #2407, opened 2023-09-21
Update the SyncDB tests to use multi-threading

JehandadKhan:
The SyncDB tests were previously serial and needed a long time to run. With the updates, they can finish in less than 10 mins per GPU arch making it realistic to add them to the CI.

PR #2518, opened 2023-11-09
[gfx94x] Enable and enhance gfx94x nodes in CI

cderb:


Jenkinsfile
```diff
@@ -477,6 +477,10 @@ pipeline {
             name: "TARGET_GFX94X",
             defaultValue: true,
             description: "")
+        booleanParam(
+            name: "TARGET_GFX942",
```
atamazov: But this target should be covered by the existing `TARGET_GFX94X` parameter, right?

cderb: That's right, we can do without this variable.

cderb:
Shook out errors in the CI machine.
Limited enablement of MI300 CI to dbsync test due to limited hardware.

Jenkinsfile
```diff
@@ -997,6 +1001,25 @@ pipeline {
                                                         needs_gpu:false, needs_reboot:false, build_install: "true")
                     }
                 }
+                stage('dbsync gfx942') {
```
alexandraBara:
```suggestion
                stage('Dbsync gfx94X') {
```
or something similar where the first letter is upper case to maintain consistency.

alexandraBara (approved):

Jenkinsfile
```diff
@@ -479,6 +479,10 @@ pipeline {
             name: "TARGET_GFX94X",
             defaultValue: false,
             description: "")
+        booleanParam(
+            name: "TARGET_GFX94X_DBSYNC",
```
atamazov: [Recommendation] @cderb AFAICS this is workaround so I recommend:
- moving it to the end of the list of params
- renaming this to `WORKAROUND__TARGET_GFX94X__ALWAYS_ENABLE_EVEN_IF_SHORTAGE_OF_HW`, so it becomes cleaner that:
  - This is W/A for some temporary problem
  - This var enables a subset of tests on gfx94x even if all other tests are disabled for this target due to a lack of HW.

Jenkinsfile
```diff
@@ -692,7 +696,7 @@ pipeline {
                     }
                     agent{ label rocmnode("gfx94X") }
                     steps{
-                        buildHipClangJobAndReboot(build_type: 'debug', make_targets: Smoke_targets)
+                        buildHipClangJobAndReboot(build_type: 'debug', make_targets: Smoke_targets, needs_reboot:false)
```
atamazov, line 690: [Question] @cderb Can you please clarify why reboot is disabled for MI300 here and at some other places?

Jenkinsfile
```diff
@@ -961,7 +965,7 @@ pipeline {
                 Navi21_build_cmd = "LLVM_PATH=/opt/rocm/llvm CTEST_PARALLEL_LEVEL=2 MIOPEN_CONV_PRECISE_ROCBLAS_TIMING=0 MIOPEN_LOG_LEVEL=5 make -j\$(nproc) check"
             }
             parallel{
-                stage('dbsync gfx908') {
+                stage('Dbsync gfx908') {
```
atamazov, line 959: [R] We need to add "Dbsync" to the list of available TestSets at line 417. We should have done this a while ago, in PR #2407

Jenkinsfile
atamazov: I highly recommend enabling also `'Fp32 Hip Debug gfx94X'` stage even is there is a lack of HW.

atamazov (changes requested):
[Recommendation] @cderb I would also appreciate if we resolve 
- https://github.com/ROCm/MIOpen/pull/2407#pullrequestreview-1720670375
- https://github.com/ROCm/MIOpen/pull/2407#pullrequestreview-1792980066

Jenkinsfile
atamazov: btw this justifies renaming requested at https://github.com/ROCm/MIOpen/pull/2518#discussion_r1567442010

junliume:
@cderb could you follow up on this PR?

Jenkinsfile
cderb: This is to avoid disabling the machine, as sometimes the MI300 node isn't reconnecting after reboot.

cderb:
Resolved test node directory permission issue.

cderb:
@atamazov @junliume this is ready.

junliume (approved):

PR #627, opened 2020-11-26
[COMGR] Fixes for HIP PCH

atamazov:
Resolves issues related to SWDEV-257056:

- Fixed: Observed `MIOPEN_DONT_USE_HIP_RUNTIME_HEADERS` in `handle_test`.
- Fixed: Observed `MIOPEN_DONT_USE_HIP_RUNTIME_HEADERS` in reduction_operator.hpp (fixes `reduce_test`).
- Fixed: Observed `MIOPEN_DONT_USE_HIP_RUNTIME_HEADERS` in ConvAsmImplicitGemmV4R1DynamicWrw.
- Fixed: `handle_test` for the case when HIP kernels disabled.
- Fixed: Incorrect `#include` in kernel source.
- Added: Workaround in `handle_test` for macros missing from HIP PCH in ROCm, 3.10 RC3
  - Macro: `WORKAROUND_SWDEV_257056_PCH_MISSING_MACROS`.
  - The workaround is enabled by default.
  - It will be disabled automatically as soon as ROCm version exceeds 3.10.x
- Added feature: Introduced env var control `MIOPEN_DEBUG_COMGR_HIP_PCH_ENFORCE` that allows overriding of HIP PCH feature auto-detection.
- Added: Workaround for incorrectly reported support for HIP PCH in ROCm 3.9
  - Macro: `WORKAROUND_SWDEV_257056_PCH_INCORRECTLY_REPORTED`.
  - Fixes #624.

PR #2519, opened 2023-11-11
Bump MIOpen version to 3.1.0 and update CI docker

junliume:


requirements.txt
```diff
@@ -1,6 +1,5 @@
 sqlite3@3.43.2 -DCMAKE_POSITION_INDEPENDENT_CODE=On
 boost@1.83 -DCMAKE_POSITION_INDEPENDENT_CODE=On --build -DCMAKE_CXX_FLAGS=" -std=c++14 -Wno-enum-constexpr-conversion -Wno-deprecated-builtins -Wno-deprecated-declarations "
-ROCmSoftwarePlatform/half@10abd99e7815f0ca5d892f58dd7d15a23b7cf92c --build
```
atamazov, line 4: [Q] Where we'll take half after this?

junliume: How about `apt-get install half` ? :)

Now if we use `amdgpu-install -y --usecase=rocm ` as part of the dockerfile script, both `half` and `composablekernel` will be pre-installed. Then if we try to install it via building dependencies there will be errors.

atamazov: @junliume I recommend keeping this, but commenting this out with `#` (and maybe putting some short explanation). This may be useful for clients who use previous version of ROCm which does not install `half` by default.

averinevg (approved):

atamazov:
@junliume Thanks!

CMakeLists.txt
```diff
@@ -104,7 +104,7 @@ if(NOT WIN32 AND NOT APPLE)
     set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -s")
 endif()
 
-rocm_setup_version(VERSION 3.00.0)
+rocm_setup_version(VERSION 3.01.0)
```
JehandadKhan:
```suggestion
rocm_setup_version(VERSION 3.1.0)
```

JehandadKhan (dismissed):

junliume:
@atamazov @JehandadKhan there is a known issue being tracked by `hipRTC` and `runtime`:
```
MIOpen(HIP): Error [Compile] 'hiprtcCompileProgram(prog.get(), c_options.size(), c_options.data())' MIOpenCheckNumerics.cpp: HIPRTC_ERROR_COMPILATION (6)
MIOpen(HIP): Error [BuildHip] HIPRTC status = HIPRTC_ERROR_COMPILATION (6), source file: MIOpenCheckNumerics.cpp
MIOpen(HIP): Warning [BuildHip] In file included from /tmp/comgr-c4d73c/input/MIOpenCheckNumerics.cpp:73:
In file included from /../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/limits:42:
/../lib/gcc/x86_64-linux-gnu/11/../../../../include/x86_64-linux-gnu/c++/11/bits/c++config.h:281:28: error: typedef redefinition with different types ('long' vs 'long long')
  typedef __PTRDIFF_TYPE__      ptrdiff_t;
                                ^
/long_pathname_so_that_rpms_can_package_the_debug_info/src/external/clr/hipamd/include/hip/amd_detail/amd_hip_runtime.h:99:19: note: previous definition is here
typedef long long ptrdiff_t;
                  ^
```
We know that the header file conflict has been an issue. Do we have clear list of green-lighted header files? Sure we may work around it (we are only using `std::numeric_limits` in this case @JehandadKhan ), but I'm afraid we will keep seeing similar issues later, if we do not have the blacklist(whitelist) of standard header files from hipRTC/COMgr.

atamazov:
@junliume The recipe for the fix can be found at https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2049#issuecomment-1484242502, more details can be found at #627. The fix should be applied here: https://github.com/ROCmSoftwarePlatform/MIOpen/blame/956825adbf4b86aa75e790ecdcf72f26c80f3744/src/kernels/MIOpenCheckNumerics.cpp#L73

> Do we have clear list of green-lighted header files?

I do not remember if HIPRTC developers provided us with it; most likely, not. I recommend looking into SWDEV-257056 and tickets linked to it. IIRC there are none system headers that we may safely include when we use HIPRTC. One of the reasons is that the library should NOT require installation of any development packages (e.g. `libstdc++-12-dev`) to work on the target system. 

/cc @JehandadKhan

junliume:
> @junliume The recipe for the fix can be found at [#2049 (comment)](https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2049#issuecomment-1484242502), more details can be found at #627. The fix should be applied here: https://github.com/ROCmSoftwarePlatform/MIOpen/blame/956825adbf4b86aa75e790ecdcf72f26c80f3744/src/kernels/MIOpenCheckNumerics.cpp#L73
> 
> > Do we have clear list of green-lighted header files?
> 
> I do not remember if HIPRTC developers provided us with it; most likely, not. I recommend looking into SWDEV-257056 and tickets linked to it. IIRC there are none system headers that we may safely include when we use HIPRTC. One of the reasons is that the library should NOT require installation of any development packages (e.g. `libstdc++-12-dev`) to work on the target system.
> 
> /cc @JehandadKhan

@atamazov I cannot bypass `<limit>` since `std::numeric_limits` is used here, unless we define it by ourselves ;-P

atamazov:
@junliume Please try it, it might work because the precompiled headers may contain the necessary definitions.

If not, then indeed we have to provide the necessary stuff (and only it) in the library. If this is the case, please let me know and I will help.

junliume:
> @junliume Please try it, it might work because the precompiled headers may contain the necessary definitions.
> 
> If not, then indeed we have to provide the necessary stuff (and only it) in the library. If this is the case, please let me know and I will help.

@atamazov  I'm afraid it might be `float8` related?

```
MIOpen(HIP): Warning [BuildHip] In file included from /tmp/comgr-28ca3a/input/MIOpenCheckNumerics.cpp:75:
/tmp/comgr-28ca3a/include/hip_float8.hpp:266:38: error: no member named 'numeric_limits' in namespace 'std'
           (fabs(rhs - *this) < std::numeric_limits<hip_f8<T>>::epsilon()))
                                ~~~~~^
```

atamazov:
@junliume Yes, it is f8 related.
1. **The use of epsilon there (`operator==()` for F8) is highly questionable**. If we want to allow 1ULP difference, then the values being compared needs to be normalized to 1.0. But allowing 1ULP conflicts with operators `>` and `<`, i.e. currently operator `a==b` is not equal to `!(a < b || a > b)`. @JehandadKhan 

2. **Is it so that the compilation error happens only with F8 specializations of numeric_limits' members?** If so, then it seems like the PCH is outdated and does not include the F8 definitions. We can provide a W/A but at the same time request updating HIPRTC with fresh PCH.

src/rnn_api.cpp
```diff
@@ -529,7 +529,7 @@ static void LogCmdRNN(const miopenTensorDescriptor_t* xDesc,
                       const int seqLength,
                       const RNNDir_t dir)
 {
-    if(miopen::IsLoggingCmd())
+    if(miopen::IsLoggingCmd() && seqLength > 0)
```
atamazov, line 526: [Notice] line 549 `miopen::deref(xDesc[seqLength - 1])` is UB when `seqLength == 0`. Shall we MIOPEN_THROW in the callers (`miopenRNNForwardTraining()` etc)?

/cc @shurale-nkn

atamazov (approved):
LGTM

src/rnn_api.cpp
shurale-nkn: This is internal function for API logging. If user passed seqLength == 0 to API this is UB at any RNN_API function because this is empty zero sized tensor.

test/handle_test.cpp
```diff
@@ -28,6 +28,10 @@
 /// \todo Create dedicated ticket and rename macro.
 #define WORKAROUND_SWDEV_257056_PCH_MISSING_MACROS 1
 
+// https://gerrit-git.amd.com/c/compute/ec/clr/+/972441
+#define WORKAROUND_ISSUE_2600 \
+    (HIP_PACKAGE_VERSION_FLAT > 5007023384ULL && HIP_PACKAGE_VERSION_FLAT <= 6000023494ULL)
```
atamazov, line 33: @junliume Are we sure that HIPRTC fix will arrive to HIP runtime right after 6.0.23494?

src/comgr.cpp
```diff
@@ -1293,7 +1293,7 @@ void BuildHip(const std::string& name,
             miopen::SplitSpaceSeparated(options, miopen::comgr::compiler::lc::GetOptionsNoSplit());
         compiler::lc::RemoveOptionsUnwanted(opts);
         opts.push_back("-DWORKAROUND_ISSUE_HIPRTC_TRUE_TYPE"); // Workaround for SWDEV-308073
-        opts.push_back("-D__HIP_PLATFORM_HCC__=1");            // Workaround?
+        opts.push_back("-D__HIP_PLATFORM_AMD__=1");            // Workaround?
```
atamazov:
```suggestion
#if HIP_PACKAGE_VERSION_FLAT < 6000023494ULL
        opts.push_back("-D__HIP_PLATFORM_HCC__=1");            // Workaround?
#endif
```

src/composable_kernel/external/rocm/include/bfloat16_dev.hpp
```diff
@@ -30,7 +30,7 @@
 extern "C" {
 #endif
 
-#ifdef __HIP_PLATFORM_HCC__
+#ifdef __HIP_PLATFORM_AMD__
```
atamazov, line 33: We must check offline compilation with ROCm 5.7. If 5.7 offline compiler does not define `__HIP_PLATFORM_AMD__`, then we'll have backward compatibility problem.

atamazov (changes requested):

src/composable_kernel/external/rocm/include/bfloat16_dev.hpp
atamazov: According to the documentation, this should work down to ROCm 5.0.0 at least (as it defines `__HIP_PLATFORM_AMD__`).

junliume: Yes I checked with compiler folks before making these changes :)

test/handle_test.cpp
junliume: Yes that's the info provided to me, the patch from hipRTC has merged but not yet in 6.0

JehandadKhan (approved):

test/handle_test.cpp
atamazov: [Recommendation] Okay. Anyway, we'll see the failure and adjust the upper bound if necessary. But let's prepare for potential 5.7.x, i.e. do not disable this test for it, what do you think?
```suggestion
    (HIP_PACKAGE_VERSION_FLAT > 6000000000ULL && HIP_PACKAGE_VERSION_FLAT <= 6000023494ULL)
```

src/composable_kernel/external/rocm/include/bfloat16_dev.hpp
atamazov: [resolved]

test/handle_test.cpp
atamazov: I'll do this in my `fix-hiprtc-60`. [resolved]

JehandadKhan (approved):

atamazov:
@JehandadKhan This breaks build with `-DMIOPEN_USE_COMPOSABLEKERNEL=Off`

PR #2523, opened 2023-11-14
AI Based Parameter Prediction Model for conv_hip_igemm_group_fwd_xdlops Solver

Dmantri98:
This model predicts kernel_id's (kernel parameters) for the conv_hip_implicit_gemm_grouped_fwd_xdlops solver. These are the gpu kernel runtimes of the model's predicted parameters compared to the default parameters (in milliseconds):

Model  GPU Kernel Time Forward Conv. Elapsed 
mean:  1.0351979585798816 ms
std:  3.697059412553356 ms

Default  GPU Kernel Time Forward Conv. Elapsed
mean:  3.5112720892245406 ms
std:  11.53792551329461 ms

The model's average runtime in microseconds:
Time taken by Model
mean:  868.8673310495172
std:  55.71458305871032


src/solver/conv_asm_1x1u.cpp
```diff
@@ -367,37 +367,42 @@ bool PerformanceConfigConvAsm1x1U::IsValidImpl(const ProblemDescription& problem
...
+    if(miopen::IsDisabled(MIOPEN_DEBUG_CONV_DIRECT_ASM_1X1U_AI_HEUR{}))
         return false;
     if(ctx.GetStream().GetDeviceName() != "gfx908")
         return false;
     if(problem.GetKernelStrideH() != 1)
         return false;
     return true;
+#else
+    return false;
+#endif
```
CAHEK7: Does it mean that this solver does not have any default parameters or something like this and should be completely disabled without AI heuristic?

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -189,23 +193,143 @@ bool ConvHipImplicitGemmGroupFwdXdlops::CheckCKApplicability(
...
+    if(idx >= 5)
+        idx += 2; // skip MPerXDL and NPerXDL as they are constant
+    std::copy_if(
+        heuristic_indexes.begin(),
+        heuristic_indexes.end(),
+        std::back_inserter(new_heuristic_indexes),
+        [&](int heuristic_index) { return heuristic_kernels[heuristic_index][idx] == value; });
+    if(new_heuristic_indexes.empty())
+        return false;
+    heuristic_indexes = new_heuristic_indexes;
```
CAHEK7: It seems we don't need a temporary vector for `heuristic_indexes`.
Using `std::remove_if` and following `erase` if `remove_if` returned the last iterator is not equal to `begin()` or do nothing otherwise.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -189,23 +193,143 @@ bool ConvHipImplicitGemmGroupFwdXdlops::CheckCKApplicability(
...
+}
+
+template <typename DataType>
+void PerformanceConfigHipImplicitGemmGroupFwdXdlops::RunParameterPredictionModel(
+    const ExecutionContext& ctx, const ProblemDescription& problem)
+{
+    valid_kernels = FillValidKernelsIDs<DeviceOpGFwdPtrs<DataType>, CKArgs>(
+        problem); // filter valid_kernel ID's
+    InitHeuristicKernelIDs();
+    static const std::string& arch  = ctx.GetStream().GetDeviceName();
```
CAHEK7, line 274: Is it for caching purposes?

src/solver/conv_asm_1x1u.cpp
Dmantri98: This just means that if the model is not applicable (or if the cmake variable is not set) heuristic_init will fall back to static heuristic to setup it's parameters.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
Dmantri98: The static const is because the name of the architecture will not change during a run of MIOpen, so calling GetStream().GetDeviceName() is redundant as the architecture should remain the same across all instances of PerformanceConfigHipImplicitGemmGroupFwdXdlops.

src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -509,11 +515,16 @@ std::shared_ptr<Model> GetModel(const std::string& arch, const std::string& solv
...
-    auto model             = GetModel(arch, solver);
-    int dim                = std::sqrt(features.size());
-    fdeep::tensors context = model->Encode(features, dim);
+    auto model = GetModel(arch, solver);
+    int dim    = 0;
+    if(transform_features)
+        dim = std::sqrt(features.size());
+    else
+        dim = features.size();
+    fdeep::tensors context = model->Encode(features, dim, transform_features);
```
msaudulhassan, line 520: can we make `dim` a part of the `Model` object? Since `dim` is known apriori, maybe we can just add it to the metadata for the model.

src/solver/conv_asm_1x1u.cpp
msaudulhassan: @JehandadKhan do you think KernelTuningNet trained for one GPU might also work well for the others (like TunaNet does)? I don't think this would be the case here because the the mapping from kernel parameters to kernel runtime is a very unstable one, but I was wrong about TunaNet, too :)

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
msaudulhassan: To follow up on this, `ai::tuning::GetModel` checks for whether the architecture has changed. 

https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ffbc4280e2b309574e41241ea8527cd1277c8e48/src/conv/heuristics/ai_heuristics.cpp#L495-L513

But since `arch` is declared `static`, that check is broken, right?

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -189,23 +193,143 @@ bool ConvHipImplicitGemmGroupFwdXdlops::CheckCKApplicability(
...
+        index     = 0;
+        kernel_id = valid_kernels[index];
+    }
+}
+
+bool PerformanceConfigHipImplicitGemmGroupFwdXdlops::IsModelApplicable(
+    const ExecutionContext& ctx, const ProblemDescription& problem) const
+{
+    if(ctx.GetStream().GetDeviceName() != "gfx90a")
+        return false;
```
msaudulhassan, line 294: We validate the device name on each run. Can we just do this check once, and store the result in a static variable? Since you have a check later on that makes sure that MIOpen throws an error if the device changes

msaudulhassan (commented):
Some minor comments.

BTW, the statistics you have provided in the description, could you also provide:
1. how many problems you tested?
2. were those problems taken from FindDB?

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
Dmantri98: @msaudulhassan This was a check suggested by @DrizztDoUrden I believe. prevArch should only be initialized the first time and if there are subsequent calls to GetModel for architectures that do not match the initial architecture that was passed in we would throw an error. However I didn't understand exactly why this was necessary as I thought this would be impossible since a run of MIOpen can only run on a single gpu instance/architecture.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
Dmantri98: When we have a gfx908, gfx940 models sure. Otherwise this would try and find a .ktn file that would not exist. for example if this was run on gfx908 it would try to find gfx908_ConvHipIgemmGroupFwdXdlops_decoder.ktn.model file even though it doesn't exist atm.

Dmantri98:
> Some minor comments.
> 
> BTW, the statistics you have provided in the description, could you also provide:
> 
> 1. how many problems you tested?
> 2. were those problems taken from FindDB?

1) Around 12.5K problems
2) Taken from tuning, I'm not sure if our current FindDB in MIOpen takes into account the configs I ran within my session. But then again these problems were gotten from all of NHWC configs I could find in the conv_config table, so these entries may line up with what is currently in MIOpen's FindDB.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
msaudulhassan: @Dmantri98 Yes, I too am unclear as to when such a case could come into play, but I wonder if we can handle it instead of throwing an error. What if we make `models` a 2D dictionary, indexed first by `arch` and then by `solver`?

msaudulhassan: Another point: in this function
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ffbc4280e2b309574e41241ea8527cd1277c8e48/src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp#L266-L289
on line 
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ffbc4280e2b309574e41241ea8527cd1277c8e48/src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp#L273
you are defining arch as `static`, and then on line
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ffbc4280e2b309574e41241ea8527cd1277c8e48/src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp#L276
`arch` is passed to `ModelSetParams()`. On line
https://github.com/ROCmSoftwarePlatform/MIOpen/blob/ffbc4280e2b309574e41241ea8527cd1277c8e48/src/conv/heuristics/ai_heuristics.cpp#L521
`arch` gets passed to `GetModel()`.  So, `GetModel()` is getting a cached arch, which means `prevArch` would _always_ equal `arch`

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -33,9 +33,12 @@
 #include <miopen/solver/problem_description_interpreter.hpp>
 #if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
 #include <ck/library/tensor_operation_instance/gpu/grouped_convolution_forward.hpp>
+#include <miopen/conv/heuristics/ai_heuristics.hpp>
+#include <nlohmann/json_fwd.hpp>
```
averinevg: Probably an unnecessary header

src/include/miopen/solver.hpp
```diff
@@ -370,12 +370,8 @@ struct PerformanceConfigConvAsm1x1U : PerfConfigBase<PerformanceConfigConvAsm1x1
...
     void StaticHeuristic(const miopen::conv::ProblemDescription& problem);
     void HeuristicInit(const ExecutionContext&, const miopen::conv::ProblemDescription&);
-#if MIOPEN_ENABLE_AI_KERNEL_TUNING
-    void RunParmeterPredictionModel(const ExecutionContext&,
-                                    const miopen::conv::ProblemDescription&,
-                                    bool& valid);
-    bool ModelApplyToken(int index, int value, const miopen::conv::ProblemDescription&);
-#endif
+    bool IsModelApplicable(const ExecutionContext& ctx,
+                           const miopen::conv::ProblemDescription& problem) const;
```
averinevg, line 374:
```suggestion
#if MIOPEN_ENABLE_AI_KERNEL_TUNING
    bool IsModelApplicable(const ExecutionContext& ctx,
                           const miopen::conv::ProblemDescription& problem) const;
#endif
```

src/include/miopen/solver.hpp
```diff
@@ -4497,12 +4497,21 @@ struct PerformanceConfigHipImplicitGemmGroupFwdXdlops
     }
     bool IsValid(const miopen::conv::ProblemDescription&) const;
     bool operator==(const PerformanceConfigHipImplicitGemmGroupFwdXdlops& other) const;
+    bool IsModelApplicable(const ExecutionContext& ctx,
+                           const miopen::conv::ProblemDescription& problem) const;
```
averinevg, line 4504:
```suggestion
#if MIOPEN_ENABLE_AI_KERNEL_TUNING
    bool IsModelApplicable(const ExecutionContext& ctx,
                           const miopen::conv::ProblemDescription& problem) const;
#endif
```

src/include/miopen/solver.hpp
```diff
@@ -4497,12 +4497,21 @@ struct PerformanceConfigHipImplicitGemmGroupFwdXdlops
...
 
 private:
+    std::vector<int> heuristic_indexes;
+    std::vector<std::vector<std::string>> heuristic_kernels;
```
averinevg, line 4509:
```suggestion
#if MIOPEN_ENABLE_AI_KERNEL_TUNING
    std::vector<int> heuristic_indexes;
    std::vector<std::vector<std::string>> heuristic_kernels;
#endif
```

src/include/miopen/solver.hpp
```diff
@@ -4497,12 +4497,21 @@ struct PerformanceConfigHipImplicitGemmGroupFwdXdlops
...
     template <typename DataType>
     void Init(const miopen::conv::ProblemDescription&);
     template <typename DataType>
     bool CheckIsSupportCKArgs(const miopen::conv::ProblemDescription&) const;
+    template <typename DataType>
+    void RunParameterPredictionModel(const ExecutionContext& ctx,
+                                     const miopen::conv::ProblemDescription& problem);
+    void InitHeuristicKernelIDs();
+    bool ModelApplyToken(int idx, std::string value);
```
averinevg:
```suggestion
#if MIOPEN_ENABLE_AI_KERNEL_TUNING
    template <typename DataType>
    void RunParameterPredictionModel(const ExecutionContext& ctx,
                                     const miopen::conv::ProblemDescription& problem);
    void InitHeuristicKernelIDs();
    bool ModelApplyToken(int idx, std::string value);
#endif
```

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -189,23 +193,142 @@ bool ConvHipImplicitGemmGroupFwdXdlops::CheckCKApplicability(
 }
 #endif
 
+std::vector<std::string> GetKernelAsTokens(const std::string& kernel)
```
averinevg:
```suggestion
static std::vector<std::string> GetKernelAsTokens(const std::string& kernel)
```
Or you can use unnamed namespace here

averinevg (changes requested):

src/include/miopen/solver.hpp
JehandadKhan: I believe the above condition would be handled inside the method, such that if `MIOPEN_ENABLE_AI_KERNEL_TUNING` is not set, it will return false?

src/include/miopen/solver.hpp
JehandadKhan: Same as before

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
Dmantri98: @msaudulhassan I will remove this check in the GetModel() function as it seems redundant.

src/conv/heuristics/ai_heuristics.cpp
Dmantri98: This would be fine but I don't see the point as we have to hard-code the feature vector anyway so if we know the size of the feature vector we can just extract ```dim``` from it.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
Dmantri98: json_fwd.hpp is indeed unnecessary, it seems to be an artifact of how we previously had KernelTuningNet integrated into MIOpen. ai_heuristics.hpp is necessary as it contains the declaration for ModelSetParams() function that is called from these two solvers.

src/include/miopen/solver.hpp
averinevg: I think this is implemented here without `#ifdef` to remove some `#ifdef`s in the code of the solver itself. But `RunParameterPredictionModel()` that is called after checking the condition is protected by `#ifdef` (as it should be), and there will be a compiler error if `MIOPEN_ENABLE_AI_KERNEL_TUNING` is not set.

CAHEK7 (approved):

Dmantri98:
@DrizztDoUrden Can you please review this PR?

src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -450,9 +451,14 @@ class Model
...
+    fdeep::tensors Encode(const std::vector<float>& features, std::size_t dim, bool transform) const
     {
-        fdeep::tensor input_tensor = fdeep::tensor(fdeep::tensor_shape(dim, dim), features);
+        if(transform)
+        {
+            fdeep::tensor input_tensor = fdeep::tensor(fdeep::tensor_shape(dim, dim), features);
+            return encoder.predict({input_tensor});
+        }
+        fdeep::tensor input_tensor = fdeep::tensor(fdeep::tensor_shape(dim, 1), features);
         return encoder.predict({input_tensor});
```
DrizztDoUrden:
```suggestion
        const auto tensor_shape_depth = transform ? dim : 1;
        fdeep::tensor input_tensor = fdeep::tensor(fdeep::tensor_shape(dim, tensor_shape_depth), features);
        return encoder.predict({input_tensor});
```

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -214,7 +340,17 @@ bool PerformanceConfigHipImplicitGemmGroupFwdXdlops::SetNextValue(const ProblemD
...
+        switch(problem.GetInDataType())
+        {
+        case miopenHalf: Init<ck::half_t>(problem); break;
+        case miopenFloat: Init<float>(problem); break;
+        case miopenInt8: Init<int8_t>(problem); break;
+        case miopenInt32:
+        case miopenBFloat16:
+        case miopenFloat8:
+        case miopenBFloat8:
+        case miopenDouble: break;
```
DrizztDoUrden, line 353: Shouldn't this throw if we haven't initialized with anything?

Dmantri98: The assert on line 354 should throw an error.

src/include/miopen/solver.hpp
Dmantri98: @averinevg I have gotten rid of the compiler error, the #ifdef is now placed in a more appropriate position.

src/solver/conv_asm_1x1u.cpp
```diff
@@ -482,22 +471,30 @@ void PerformanceConfigConvAsm1x1U::StaticHeuristic(const ProblemDescription& pro
...
     {
-        bool valid = false;
-        RunParmeterPredictionModel(ctx, problem, valid);
-        if(valid)
+
+        if(RunParameterPredictionModel(ctx, problem))
             return;
     }
-#else
-    std::ignore = ctx;
```
averinevg, line 500: This line should probably be returned

averinevg (changes requested):

src/solver/conv_asm_1x1u.cpp
Dmantri98: I've added [[maybe_unused]] to the ctx function parameter declaration.

averinevg (approved):

JehandadKhan:
@junliume  Can we merge this when convenient ?

junliume (approved):

PR #2524, opened 2023-11-15
Standardize workspace abstraction

amberhassaan:
Tests lack a standardized way of passing workspace to various MIOpen calls. Often we create and track 3 things: 
- a host vector
- a size variable
- a device pointer (created in multiple ways, e..g, handle.Write() or hipMemcpy). 

This PR introduces a standard abstraction class for this purpose and fixes some subtle bugs in `test/conv_common.hpp`.


test/workspace.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+    auto ptr() const { return gpu_buf_.ptr(); }
+
+    auto ptr() { return gpu_buf_.ptr(); }
+
+    void AdjustToSize()
+    {
+        if(sz_ != 0)
+        {
+            gpu_buf_ = GPUBuffer(sz_);
```
CAHEK7: I'm not sure is it critical or not, but:
1) it doubles the memory requirements in case of reallocation, because the old and the new buffers must co-exist till the end of the function.
2) in a common case we don't want to reallocate on shrink, but for test code it seems good

test/workspace.hpp
```diff
@@ -0,0 +1,160 @@
...
+    {
+        using T = typename V::value_type;
+        if(vec.size() * sizeof(T) != size())
+        {
+            std::abort();
+        }
+        auto s = hipMemcpy(&vec[0], ptr(), size(), hipMemcpyDeviceToHost);
+        if(s != hipSuccess)
+        {
+            abort();
```
CAHEK7: [minor]
mixing `std::abort` and `abort`.

[not so minor, but still minor]
probably print some meaningful info?

test/workspace.hpp
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    template <typename V>
+    V Read() const
+    {
+        using T         = typename V::value_type;
+        size_t num_elem = size() / sizeof(T);
+        V ret(num_elem);
+        ReadTo(ret);
```
CAHEK7, line 154: I guess `num_elem` must not be a `floorInt(size(), sizeof(T))`, but rather `ceilInt(size(), sizeof(T))` or at least there should be check that `size()` can be divided by `sizeof(T)` equally.

Currently the biggest problem is the error reporting happens too late and it may require some extra time to understand why the vector has less elements than required. And since we do not control any types and can reuse workspace for any types, it may happen.

test/workspace.hpp
```diff
@@ -0,0 +1,160 @@
...
+            return *this;
+        }
+
+    private:
+        void* buf_ = nullptr;
+    };
+
+public:
+    Workspace() = default;
+
```
CAHEK7: Why don't we have `explicit Workspace(size_t byte_size)` constructor?

test/workspace.hpp
```diff
@@ -0,0 +1,160 @@
...
+            gpu_buf_ = GPUBuffer();
+        }
+    }
+
+    template <typename V>
+    void Write(const V& vec)
+    {
+        using T = typename V::value_type;
+        resize(vec.size() * sizeof(T));
+        auto s = hipMemcpy(this->ptr(), &vec[0], size(), hipMemcpyHostToDevice);
```
CAHEK7: [minor]
`&vec[0]` -> `.data()`? Either std::vector (which is implied here) or `std::array` supports `.data()` and we can't use plain arrays since they don't' have `.size()` and `::value_type`

test/workspace.hpp
```diff
@@ -0,0 +1,160 @@
...
+        GPUBuffer(GPUBuffer&& that) noexcept : buf_(that.buf_)
+        {
+            that.buf_ = nullptr; // take over ownership
+        }
+
+        GPUBuffer& operator=(GPUBuffer&& that) noexcept
+        {
+            std::swap(this->buf_, that.buf_);
+            return *this;
+        }
```
CAHEK7: This is the second place of doubling memory requirement on re-allocation.
I guess the most common case is reallocation from an empty state, but nevertheless.

test/workspace.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+    void resize(size_t sz_in_bytes)
+    {
+        sz_ = sz_in_bytes;
+        AdjustToSize();
+    }
+
+    // for use in miopen .*GetWorkSpaceSize() methods where a pointer to size_t is
+    // passed to capture the size. Must call AdjustToSize() after calling such a method
+    size_t* SizePtr() { return &sz_; }
```
CAHEK7: Since this class is for the testing purposes, that sort of resize can be checked and reported.
Using `sz_` and `old_sz_` may help to detect unexpected `sz_` update.
Both members can be synchronized on `AdjustToSize` and checked to be equal on each other operation.

CAHEK7 (changes requested):
I did not check any other files except `workspace.hpp`, but let's finalize it first.

atamazov:
@amberhassaan Do we have a ticket that this PR is intended to resolve?

atamazov:
@amberhassaan @JehandadKhan @junliume @DrizztDoUrden I would like to explain why we deliberately did not pay much attention to checking the size of the workspace.

All MIOpen operations have two or more buffers with tensors as parameters. _Some_ operations require a workspace buffer. For example, convolutions get three tensors and an optional workspace.

Thus, potential problems with the size of tensor buffers may arise _more often_ than problems with the workspace buffer.

However, the library knows about tensor buffers only their starting addresses and has no way to check the _true_ size of the buffers. And we don't do that. We only check for nullptr.

And, if problems with tensors can potentially occur more often than with workspaces, then what's the point of bothering with workspace size checks? Moreover, the library cannot check the _true_ size of the workspace buffer at all, because there is no guarantee that the user application provides its correct value. Therefore, the check does not 100% guarantee anything, but wastes CPU cycles. Therefore, we deliberately do not pay much attention to the size of the workspace.

Given that the library does not need workspace size as a parameter -- it is used only for checks, -- I would remove it from the library API altogether. But, unfortunately, this is not possible for the reasons of backward compatibility.

That is why we implemented (and I suggest leaving) only nullptr checks in the release builds.

amberhassaan:
> @amberhassaan @JehandadKhan @junliume @DrizztDoUrden I would like to explain why we deliberately did not pay much attention to checking the size of the workspace.
> 
> All MIOpen operations have two or more buffers with tensors as parameters. _Some_ operations require a workspace buffer. For example, convolutions get three tensors and an optional workspace.
> 
> Thus, potential problems with the size of tensor buffers may arise _more often_ than problems with the workspace buffer.
> 
> However, the library knows about tensor buffers only their starting addresses and has no way to check the _true_ size of the buffers. And we don't do that. We only check for nullptr.
> 
> And, if problems with tensors can potentially occur more often than with workspaces, then what's the point of bothering with workspace size checks? Moreover, the library cannot check the _true_ size of the workspace buffer at all, because there is no guarantee that the user application provides its correct value. Therefore, the check does not 100% guarantee anything, but wastes CPU cycles. Therefore, we deliberately do not pay much attention to the size of the workspace.
> 
> Given that the library does not need workspace size as a parameter -- it is used only for checks, -- I would remove it from the library API altogether. But, unfortunately, this is not possible for the reasons of backward compatibility.
> 
> That is why we implemented (and I suggest leaving) only nullptr checks in the release builds.

@atamazov : Your argument is weak. You're saying, since we can't check A, we shouldn't check B. What is the wisdom in that? We should check whatever we reasonably and cheaply can. 

@JehandadKhan , @junliume : I need these fixes to help my PR that implements NCHW convolution, pass CI. I noticed that CI was failing because I introduced a requirement of workspace for NCHW (due to on-demand transformations) and I started seeing failures elsewhere in our tests where workspace pointer and size were being passed inconsistently. I'd like to remove such quirks from our code so that we don't keep running into them.

CAHEK7 (approved):
Seems good, but I'd rather stick to `std::byte` vector type for all the operations, since workspace by the meaning is a raw memory buffer without any clue about its internal structure or type or anything else. Moreover, it can be internally reused as any sort of data and types.
It's not quite good to read or write any other types except bytes. It helps to avoid the issues like #2529 where algorithm requires a workspace for `half`s but in the test `std::vector<float>` is used and it wrongly calculate its internal size and allocated less memory than required.

JehandadKhan:
@atamazov I am not sure I follow, while you are right we cant check the size of a raw pointer, we do make assumptions about its size, for example when the user asks for the required size of workspace, we assume that the supplied pointer by the client application is of a certain size. In which case we know the expected size of the buffer and can filter out solvers based on this information. 

Moreover, I dont see anything wrong with creating an abstraction for an entity which we use frequently and can cause correctness issues simply because the pointer and the size were not in sync.

amberhassaan:
> Seems good, but I'd rather stick to `std::byte` vector type for all the operations, since workspace by the meaning is a raw memory buffer without any clue about its internal structure or type or anything else. Moreover, it can be internally reused as any sort of data and types. It's not quite good to read or write any other types except bytes. It helps to avoid the issues like #2529 where algorithm requires a workspace for `half`s but in the test `std::vector<float>` is used and it wrongly calculate its internal size and allocated less memory than required.

I'd like to allow writing any `vector<T>` into the workspace for convenience's sake but also because it makes sense, e.g. if you're working with `float` data vs `half` vs `int8_t`. Workspace size will always be in `bytes`, so I don't see an issue here. AFA things like #2529 go, I'll get to the bottom of that as I suspect there is another reason behind the failure I'm seeing. Either the algorithm is not reporting the correct workspace size or that test is incorrect.

CAHEK7:
> it makes sense, e.g. if you're working with `float` data vs `half` vs `int8_t`

I don't think so.
The point is that you have no idea what kind of data will reside in workspace. A type of input and output tensors does not matter at all. At can be anything, it depends on algorithm and only algorithm knows (and when you perform a typical run with find, you don't know even algorithm itself).
That's it. You can't write anything else except zero or garbage and you can read only garbage from it.

atamazov:
@JehandadKhan 
> Moreover, I dont see anything wrong with creating an abstraction for an entity which we use frequently and can cause correctness issues simply because the pointer and the size were not in sync.

Me too. My primary goal was to explain the reasons why some time ago the decision was made not to bother checking the size of the workspace.

atamazov:
@amberhassaan 
> Your argument is weak. You're saying, since we can't check A, we shouldn't check B. What is the wisdom in that? ...

I absolutely agree with you. A weak mind produces weak arguments that contain no wisdom.

amberhassaan:
> > it makes sense, e.g. if you're working with `float` data vs `half` vs `int8_t`
> 
> I don't think so. The point is that you have no idea what kind of data will reside in workspace. A type of input and output tensors does not matter at all. At can be anything, it depends on algorithm and only algorithm knows (and when you perform a typical run with find, you don't know even algorithm itself). That's it. You can't write anything else except zero or garbage and you can read only garbage from it.

@CAHEK7 : I can agree with this much that we shouldn't need to `Read` or `Write` the workspace. I don't know yet as to why the tests are doing that (e.g. some places we use cpu workspace to verify gpu workspace). I think we need to go back to the tests that need `Read/Write` and see if we can remove this need.

amberhassaan:
@JehandadKhan , @junliume : this PR is ready to merge. Needed for #2429, which is targeted for ROCm 6.1.

junliume:
> @JehandadKhan , @junliume : this PR is ready to merge. Needed for #2429, which is targeted for ROCm 6.1.

@amberhassaan Let me test on some legacy nodes, to prevent similar issues from last time: https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2539

JehandadKhan (approved):

DrizztDoUrden (approved):

atamazov (approved):
🟢 LGTM!

atamazov:
@amberhassaan Can you please shed some light on why this PR is a release blocker?

amberhassaan:
> @amberhassaan Can you please shed some light on why this PR is a release blocker?

I need it for #2429 which is targeted for 6.1.

junliume:
Watching CI and will merge after all passes

junliume (approved):

PR #2526, opened 2023-11-15
Use split CK libraries.

illsilin:
This change will replace the single static libdevice_operations.a library with multiple files, split by operation type, e.g., device_gemm_operations, device_conv_operations, etc.


atamazov:
@illsilin Could you please describe little bit why this is necessary? Thank you!

junliume:
> @illsilin Could you please describe little bit why this is necessary? Thank you!

While @illsilin is on vacation, this is necessary since CK is splitting the binaries into smaller ones, because they are exceeding the size limitations.

CMakeLists.txt
```diff
@@ -260,7 +260,7 @@ add_definitions("-DHIP_COMPILER_FLAGS=${HIP_COMPILER_FLAGS}")
 # HIP
 if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "HIPNOGPU")
     if(MIOPEN_USE_COMPOSABLEKERNEL)
-        find_package(composable_kernel 1.0.0 COMPONENTS device_operations)
+        find_package(composable_kernel 1.0.0 COMPONENTS device_other_operations device_gemm_operations device_conv_operations device_contraction_operations device_reduction_operations)
```
iq136boy, line 269: @illsilin One concern: is there any dependency among libraries or they are totally exclusive?  Does CK also support separate build for instances by operation type? For example, when I was integrating convolution kernels from CK into MIOpen, I only want to build the convolution instances and link 'device_conv_operations' library.

junliume: In addition to @iq136boy comment above, here is the ref: https://github.com/ROCmSoftwarePlatform/composable_kernel/issues/1053

atamazov (approved):
No objections!

CMakeLists.txt
illsilin: Yes, the library will be split in just a few blocks by most common operations: gemms, convolutions, reductions, multi-head-attention, contractions, and everything else. So if your application only requires convolutions, you won't have to link all the libraries, just pick the convolutions one. If you also need softmax or layernorm, just add the "device_other_operations". 
This is nothing new, intel mkl libraries, for example, have been organized in a similar way.

CAHEK7 (approved):

junliume (approved):
This is needed for the next staging

PR #2531, opened 2023-11-16
Forward, backward data and backward weight convolution solver with fp8/bfp8 compute datatype.

iq136boy:
Enable the support of forward, backward data and backward weight convolution solver with fp16->fp8/bfp8 compute type in MIOpen. 
The original input, weight and output tensors are fp16 datatype. Then the tensors are casted to the proper compute type to do the convolution operation utilizing CK kernels. More specific, the following fp16->fp8/bfp8 convolutions are enabled as requested from framework and the current kernels support of CK teams:
Forward: input fp16->fp8, filter fp16->fp8, output fp32->fp16.
Backward data: filter fp16->fp8, output fp16->bf8, input fp32->fp16.
Backward weight: input fp16->fp8, output fp16->bf8, filter fp32->fp16.


atamazov:
@iq136boy 
> the following fp16->fp8/bfp8 convolutions are enabled as requested from framework...
> ...
> Backward data: filter fp16->fp8, output fp16->bf8, input fp32->fp16.
> Backward weight: input fp16->fp8, output fp16->bf8, filter fp32->fp16.

Do they really want to reduce _backwards pass_ precision down to f8 types? For speed?

iq136boy:
> Do they really want to reduce _backwards pass_ precision down to f8 types? For speed?

@atamazov Yes. I also have the same concern before starting this task and has talked with Jehandad about it. Both forward and backward are requested by framework. They asked CK team to add the kernels and then asked us to integrated these kernels into MIOpen.

src/mlo_dir_conv.cpp
```diff
@@ -163,8 +163,12 @@ static auto GetImplicitGemmWrWSolvers()
         miopen::solver::conv::ConvMlirIgemmWrW,
         miopen::solver::conv::ConvAsmImplicitGemmGTCDynamicWrwXdlops,
 #if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+<<<<<<< HEAD
```
junliume: @iq136boy  These are merge conflicts and should be resolved :)

junliume (approved):

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+ *******************************************************************************/
+
+#include <vector>
+#include <cstdint>
+
+#include <miopen/solver.hpp>
+#include <miopen/generic_search.hpp>
+#include <miopen/conv/data_invoke_params.hpp>
+#include <miopen/solver/problem_description_interpreter.hpp>
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
```
atamazov: [R] Everywhere
```suggestion
#if MIOPEN_USE_COMPOSABLEKERNEL
```

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <vector>
+#include <cstdint>
```
atamazov: Everywhere: standard headers should be included after all other headers.

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+#include <miopen/solver.hpp>
+#include <miopen/generic_search.hpp>
+#include <miopen/conv/data_invoke_params.hpp>
+#include <miopen/solver/problem_description_interpreter.hpp>
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+#include <miopen/solver/ck_utility_common.hpp>
+#include <ck/library/tensor_operation_instance/gpu/grouped_convolution_backward_data.hpp>
+#endif
+#include <miopen/solver/implicitgemm_ck_util.hpp>
+MIOPEN_DECLARE_ENV_VAR(MIOPEN_DEBUG_F16F8F16_CONV_IMPLICIT_GEMM_HIP_BWD_XDLOPS)
```
atamazov: include env.hpp

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
atamazov: include config.h

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+        return false;
+    if(miopen::IsEnabled(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC{}))
+        return false;
+    if(problem.HasMixedDataTypes())
+        return false;
+    if(!problem.IsTensorsCasted())
+        return false;
+    if(!problem.IsDirectionBackwardData())
+        return false;
+    if(!problem.Is3d())
```
atamazov: [Q] Is it really only applicable for 3D?

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+        return false;
+    if(!problem.IsLayoutNHWC())
+        return false;
+    if(!problem.IsFp16())
+        return false;
+    if(!ck_utility::is_ck_whitelist(ctx.GetStream().GetDeviceName()))
+        return false;
+    const auto& yDesc = problem.GetOut();
```
atamazov: 🐛 Output for Bwd is dxDesc, dyDesc is input.
```suggestion
    const auto& yDesc = problem.GetIn();
```

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+bool PerformanceConfigHipImplicitGemmF16F8F16BwdXdlops::IsValidValue() const
+{
+    return index < valid_kernels.size();
+}
+
+bool PerformanceConfigHipImplicitGemmF16F8F16BwdXdlops::IsValid(
+    [[maybe_unused]] const ProblemDescription& problem) const
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    const auto& y_cast_type = problem.GetOut().GetCastType();
```
atamazov: 🐛 
```suggestion
    const auto& y_cast_type = problem.GetIn().GetCastType();
```

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+#endif
+
+void PerformanceConfigHipImplicitGemmF16F8F16BwdXdlops::HeuristicInit(
+    [[maybe_unused]] const ProblemDescription& problem)
+{
+    index     = 0;
+    kernel_id = "";
+
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    const auto& y_cast_type = problem.GetOut().GetCastType();
```
atamazov: :bug: Ditto

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+        return false;
+    if(!ck_utility::is_ck_whitelist(ctx.GetStream().GetDeviceName()))
+        return false;
+    const auto& yDesc = problem.GetOut();
+    const auto& wDesc = problem.GetWeights();
+    if(yDesc.GetCastType() && wDesc.GetCastType())
+    {
+        const auto y_cast_type = yDesc.GetCastType();
+        const auto w_cast_type = wDesc.GetCastType();
+        if(y_cast_type == miopenBFloat8 && w_cast_type == miopenFloat8)
```
atamazov: [R] Starting from C++17, std::optional may be directly compared with value, and this is enough:
```suggestion
    if(yDesc.GetCastType() == miopenBFloat8 && wDesc.GetCastType() == miopenFloat8)
```
Or, if you would like to use C++14 style, then compare like `*y_cast_type == miopenBFloat8`.

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+        {
+            return CheckCKApplicability<ck::half_t, ck::bf8_t, ck::f8_t>(problem);
+        }
+    }
+    else
+    {
+        MIOPEN_LOG_I("Both the output and weights tensors need to be casted");
```
atamazov: Please delete this or change to the TRACE level

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+        return false;
+    if(!problem.IsLayoutNHWC())
+        return false;
+    if(!problem.IsFp16())
+        return false;
+    if(!ck_utility::is_ck_whitelist(ctx.GetStream().GetDeviceName()))
+        return false;
+    const auto& yDesc = problem.GetOut();
+    const auto& wDesc = problem.GetWeights();
+    if(yDesc.GetCastType() && wDesc.GetCastType())
```
atamazov: [Q] Why don't we check casting for dxDesc? PR description says "fp32->fp16". Or frameworks do not require us to compute to fp32 and then cast output to fp16, i.e. they only ask us to cast inputs from 16 to 8 bits? And this fp32->fp16 casting is simply inherent to the implementation?

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
```diff
@@ -0,0 +1,368 @@
...
+    [[maybe_unused]] const PerformanceConfigHipImplicitGemmF16F8F16BwdXdlops& config) const
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    const auto& yDesc = problem.GetOut();
+    const auto& wDesc = problem.GetWeights();
+    if(yDesc.GetCastType() && wDesc.GetCastType())
+    {
+        const auto y_cast_type = yDesc.GetCastType();
+        const auto w_cast_type = wDesc.GetCastType();
+        if(y_cast_type == miopenBFloat8 && w_cast_type == miopenFloat8)
```
atamazov: We do not need to check this condition -- it is checked in IsApplicable.

src/solver/conv_hip_implicit_gemm_f16f8f16_fwd_xdlops.cpp
atamazov, line 1: [Notice] Most likely, comments from conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp apply here.

src/solver/conv_hip_implicit_gemm_f16f8f16_wrw_xdlops.cpp
atamazov, line 1: [Notice] Most likely, comments from conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp apply here.

atamazov (changes requested):

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
iq136boy: > [Q] Is it really only applicable for 3D?

It is applicable for both 2d and 3d. I will fix it.

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
iq136boy: > Output for Bwd is dxDesc, dyDesc is input.

Yeap. A bit confusing here. I will rename it.

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
iq136boy: Fixed it.

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
iq136boy: > [Q] Why don't we check casting for dxDesc? PR description says "fp32->fp16". Or frameworks do not require us to compute to fp32 and then cast output to fp16, i.e. they only ask us to cast inputs from 16 to 8 bits? And this fp32->fp16 casting is simply inherent to the implementation?

We do not need to check from MIOpen side. The [backward data API](https://github.com/ROCmSoftwarePlatform/composable_kernel/blob/develop/include/ck/tensor_operation/gpu/device/device_grouped_conv_bwd_data_multiple_d.hpp) that CK provide to us only requests the compute type for output and weights, the input type casting is handled by the CK kernel.

atamazov: [Recommendation] @iq136boy Which means that fp32->fp16 casting of _output_ (dx tensor) is a detail of implementation, right? (Ditto other solvers.) This is fine so far. However it may worth explaining this in the description of the PR.

What I see:
- The kernels convert FP16 inputs to FP8 or BF8 internally. This improves performance at the cost of accuracy. The conversion is a parameter of convolution and expressed in terms of "tensor casting" of inputs.
- The kernels perform intermediate computations in FP32 domain and convert output to FP16. This is fairly common in MIOpen and does not require additional convolutions parameters.

I recommend updating the description of this PR with the explanation shown above.

src/solver/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
atamazov: Unresolved.

PR #2528, opened 2023-11-15
[Tests] Limit layernorm CK test applicability

junliume:
#2527

Issue #2527, opened 2023-11-15
[Tests] LayerNormTestSet platform applicability issue after #2481


junliume:
After #2481 some CI stages start to fail when the tests are performed on legacy platforms:

For example, on `gfx906`:
```
The following tests FAILED:
	941 - LayerNormTestSet/LayerNormTestFloat.LayerNormTestFw/ N:64 C:4 D:0 H:4 W:256 dim:1 eps:1e-05 LayerNorm_mode:0 (Failed)
	943 - LayerNormTestSet/LayerNormTestFloat.LayerNormTestFw/ N:64 C:4 D:0 H:4 W:256 dim:1 eps:1e-05 LayerNorm_mode:1 (Failed)
	945 - LayerNormTestSet/LayerNormTestFloat.LayerNormTestFw/ N:64 C:0 D:0 H:0 W:256 dim:1 eps:1e-05 LayerNorm_mode:0 (Failed)
	947 - LayerNormTestSet/LayerNormTestFloat.LayerNormTestFw/ N:64 C:0 D:0 H:0 W:256 dim:1 eps:1e-05 LayerNorm_mode:1 (Failed)
```

The error message:
```
unknown file: Failure
C++ exception with description "HIP runtime error: invalid device function. /opt/rocm/cget/build/tmp-6a54a0c103a88df9ce804b69df353620d496a4a26d329eb22384908711114873/composable_kernel-f30e59754205db83eea8e4ea42caf421eafb70e0/include/ck/host_utility/hip_check_error.hpp: 15in function: hip_check_error" thrown in TearDown().
```

We should **protect** and be **explicit** about test applicabilities specific to platforms, i.e. only run these tests on applicable ASICs.

iq136boy:
@junliume this could be done by setting the platforms/device list in the TEST_P function in gtest. Sth like the following: 
    const auto& handle = get_handle();
    if(!((miopen::StartsWith(handle.GetDeviceName(), "gfx908") ||
        miopen::StartsWith(handle.GetDeviceName(), "gfx90a")))    // set the device list for the gtest
    {
        GTEST_SKIP();
    }

atamazov:
@junliume 
> After https://github.com/ROCmSoftwarePlatform/MIOpen/pull/2481 some CI stages start to fail

❓ How is this possible if PR has passed our CI?

If this is really so, then our testing procedure (defined in Jenkinsfile) has serious flaws that must be fixed.

[...]

PR #2535, opened 2023-11-17
[Tests] split teardown to runtest and verify in layernorm gtest

seungmanhan:
When used teardown as main part, gtest was skipped after the kernel was called.
Changed to not use teardown by split teardown to runtest and verify.
Follow-up #2528
Related to #2527


seungmanhan:
I can't choose @atamazov as a reviewers. @junliume Please add him.

junliume (approved):

atamazov (commented):
Looks good.

PR #2536, opened 2023-11-17
2d CK  group convolution backward and weight update solvers

amberhassaan:
This PR provides CK-based Backward-Data and Backward-Weight update solvers for 2D Group Convolution. This completes the solver picture for 2D group convolutions (except for f8, which has its own solvers). 

Also included in this PR is a generic design for the GTest Fixture class that handles 2D and 3D (Fwd,Bwd,Wrw) passes at the same time, so testing code can reuse this class and related macros. A future PR will use this class for 3D grouped conv gtests (arriving shortly).


amberhassaan:
@JehandadKhan , @junliume , I am closing this PR in favor of #2663 which contains the change-set of this PR plus some refactoring improvements to 3D convolution tests.

PR #2545, opened 2023-11-21
[gtest] conversion for lstm

cderb:
Depends on #2514


atamazov:
@cderb Please convert this to a draft.

xinlipn:
LGTM

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,148 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include "lstm.hpp"
+#include "get_handle.hpp"
+#include <miopen/env.hpp>
+#include <gtest/gtest.h>
+#include <boost/algorithm/string.hpp>
```
xinlipn: HI @cderb, is this booster header still needed? Otherwise, LGTM

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,147 @@
...
+    {flags + " --batch-size 16 --seq-len 150 --vector-len 256 --hidden-size 256" + commonFlags},
+    {flags + " --batch-size 32 --seq-len 150 --vector-len 256 --hidden-size 256" + commonFlags},
+    {flags + " --batch-size 64 --seq-len 150 --vector-len 256 --hidden-size 256" + commonFlags}
+        // clang-format on
+    };
+
+    return test_cases;
+}
+
+INSTANTIATE_TEST_SUITE_P(ConvTrans, ConfigWithFloat, testing::Values(GetTestCases()));
```
shurale-nkn: ConvTrans in lstm??????

shurale-nkn (dismissed):
Blocked. this PR reduces the test coverage. General LSTM test was removed, LSTM not tested in half data type, removed testing in smoke stage.

required to provide a log with list of tested configuration before and after this PR.

@JehandadKhan @junliume @atamazov  please double check that other similar PRs about gtests have not reduced the test coverage of our CI.

test/gtest/deepbench_lstm.cpp
atamazov: This should be renamed to DeepbenchLstm

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,147 @@
...
+ *******************************************************************************/
+
+#include "lstm.hpp"
+#include "get_handle.hpp"
+#include <miopen/env.hpp>
+#include <gtest/gtest.h>
+
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_DEEPBENCH)
+
+static bool SkipTest(void) { return miopen::IsDisabled(ENV(MIOPEN_TEST_DEEPBENCH)); }
```
atamazov:
```suggestion
static bool SkipTest(void) { return !miopen::IsEnabled(ENV(MIOPEN_TEST_DEEPBENCH)); }
```
This must be explicitly enabled

Or, better, pass `MIOPEN_TEST_DEEPBENCH` from CTest to google tests.

atamazov:
@shurale-nkn 
> LSTM not tested in half data type

The original test does not check half. But let's add this.

UPDATE: No need.

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,156 @@
...
+    {flags + commonFlags + " -dir-mode 1 --no-cy --no-dcx"},
+    {flags + commonFlags + " -dir-mode 0 --no-hx --no-dhy --no-cx --no-dcy --no-hy --no-dhx --no-cy --no-dcx"},
+    {flags + commonFlags + " -dir-mode 1 --no-hx --no-dhy --no-cx --no-dcy --no-hy --no-dhx --no-cy --no-dcx"}
+        // clang-format on
+    };
+
+    return test_cases;
+}
+
+INSTANTIATE_TEST_SUITE_P(ConvTrans, ConfigWithFloat, testing::Values(GetTestCases()));
```
atamazov: LstmExtra

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,156 @@
...
+std::vector<std::string> GetTestCases(void)
+{
+    std::string flags       = "test_lstm --verbose ";
+    std::string commonFlags = "--batch-size 32 --seq-len 3 --batch-seq 32 32 32 --vector-len 128 "
+                              "--hidden-size 128 --num-layers 1 --in-mode 0 --bias-mode 0";
+
+    const std::vector<std::string> test_cases = {
+        // clang-format off
+    {flags + commonFlags + " -dir-mode 0 --no-hx"},
+    {flags + commonFlags + " -dir-mode 0 --no-dhy"},
```
atamazov: Let's avoid disabling clang-format, use EOL comments to control formatting
```suggestion
    const std::vector<std::string> test_cases = { //
        {flags + commonFlags + " -dir-mode 0 --no-hx"}, //
        {flags + commonFlags + " -dir-mode 0 --no-dhy"}, //
```
etc

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+bool IsTestSupportedForDevice(const miopen::Handle& handle)
+{
+    std::string devName = handle.GetDeviceName();
+    if(devName == "gfx900" || devName == "gfx906" || devName == "gfx908" || devName == "gfx90a" ||
+       miopen::StartsWith(devName, "gfx94") || miopen::StartsWith(devName, "gfx103") ||
+       miopen::StartsWith(devName, "gfx110"))
+        return true;
+    else
+        return false;
```
atamazov: The test should work for all devices, you can simply return true.

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+bool IsTestSupportedForDevice(const miopen::Handle& handle)
+{
+    std::string devName = handle.GetDeviceName();
+    if(devName == "gfx900" || devName == "gfx906" || devName == "gfx908" || devName == "gfx90a" ||
+       miopen::StartsWith(devName, "gfx94") || miopen::StartsWith(devName, "gfx103") ||
+       miopen::StartsWith(devName, "gfx110"))
+        return true;
+    else
+        return false;
```
atamazov:
```suggestion
        return true;
```

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,156 @@
...
+};
+
+void Run2dDriver(miopenDataType_t prec)
+{
+
+    std::vector<std::string> params;
+    switch(prec)
+    {
+    case miopenFloat: params = ConfigWithFloat::GetParam(); break;
+    case miopenHalf:
```
atamazov: [R] Can we add testing with `half``?

test/gtest/deepbench_lstm.cpp
atamazov: The test should be skipped if FLOAT testing is not requested, see #1843

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+#include "lstm.hpp"
+#include "get_handle.hpp"
+#include <miopen/env.hpp>
+#include <gtest/gtest.h>
+#include <boost/algorithm/string.hpp>
+
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_ALL)
+
+static bool SkipTest(void) { return miopen::IsDisabled(ENV(MIOPEN_TEST_ALL)); }
```
atamazov: The test should be run only if FLOAT ~or HALF~ testing is requested, see #1843

atamazov (changes requested):
...

atamazov:
@shurale-nkn 
> removed testing in smoke stage.

The original test is not enabled for smoke testing.

shurale-nkn:
> @shurale-nkn
> 
> > LSTM not tested in half data type
> 
> The original test does not check half. But let's add this.



> @shurale-nkn
> 
> > removed testing in smoke stage.
> 
> The original test is not enabled for smoke testing.

@atamazov 
nonono the lstm_extra was not tested in half and smoke, but the main LSTM based on tests from cpp file was. Current PR deleted this test(main lstm), so now LSTM is gone from half and smoke stages.

test/lstm.hpp
atamazov, line 1: This renaming (`test/lstm.cpp → test/lstm.hpp`) implicitly deletes the test_lstm executable. Please move the common code to the new header, but do not delete lstm.cpp.

atamazov (changes requested):
....

test/lstm.hpp
shurale-nkn: yes, this is it. But I do not insist on choosing a solution.

atamazov: @cderb ⚠️ This kind of change (cpp->hpp) silently removes thousands of test cases (launched by CTest using the `test_lstm --all` command). Please look at the following code fragment, notice lines 382 and 436: https://github.com/ROCm/MIOpen/blob/dfaff6aa6077b19e34e7b15c0a74fb055bb6256d/test/CMakeLists.txt#L382-L437

atamazov: @cderb How to move "--all" tests from CTest to GTest is another story.

For example, we can skip `add_test_command` for the executable in `add_test_executable` and create another google test. That test should feed `--all` (but in this case we have to take also other parameters from `add_test_executable`, e.g. `--limit`, `--xnack`, `--tolerance`, and properly pass these too).

But we still need the executable for development purposes -- it provides us with capability to manually run test configs, reproduce issues etc.

/cc @JehandadKhan @shurale-nkn

test/gtest/lstm_extra.cpp
atamazov: [resolved] Not needed

test/lstm.hpp
cderb: I'll add back in the c main. Though I think the eventual goal is to convert the tests referenced by "--all" into gtest.

test/gtest/deepbench_lstm.cpp
cderb: removed

test/gtest/deepbench_lstm.cpp
cderb: renamed

test/gtest/deepbench_lstm.cpp
cderb: converted env check

test/gtest/lstm_extra.cpp
cderb: renamed

test/gtest/lstm_extra.cpp
cderb: enabled clang format

test/gtest/deepbench_lstm.cpp
cderb: reduced to true

test/gtest/lstm_extra.cpp
cderb: added half

test/gtest/deepbench_lstm.cpp
cderb: added FLOAT flag check

atamazov: @cderb Can you please also pass `MIOPEN_TEST_DEEPBENCH`? This should be done in test/gtest/CMakeLists.txt.

test/gtest/lstm_extra.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/lstm_extra.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
```diff
@@ -108,40 +106,67 @@ TEST_P(ConfigWithFloat, FloatTest)
...
+        {flags + " --batch-size 64 --seq-len 150 --vector-len 256 --hidden-size 256" +
+         commonFlags}};
 
     return test_cases;
 }
 
-INSTANTIATE_TEST_SUITE_P(ConvTrans, ConfigWithFloat, testing::Values(GetTestCases()));
+INSTANTIATE_TEST_SUITE_P(DeepbenchLstm, ConfigWithFloat, testing::Values(GetTestCases("--float")));
+
+INSTANTIATE_TEST_SUITE_P(DeepbenchLstm, ConfigWithHalf, testing::Values(GetTestCases("--half")));
```
atamazov: [Notice] HALF is not required for deepbench but I am not against this.

/cc @shurale-nkn

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/lstm_extra.cpp
```diff
@@ -109,48 +107,62 @@ TEST_P(ConfigWithFloat, FloatTest)
...
+        {flags + commonFlags +
+         " -dir-mode 1 --no-hx --no-dhy --no-cx --no-dcy --no-hy --no-dhx --no-cy --no-dcx"}};
 
     return test_cases;
 }
 
-INSTANTIATE_TEST_SUITE_P(ConvTrans, ConfigWithFloat, testing::Values(GetTestCases()));
+INSTANTIATE_TEST_SUITE_P(LstmExtra, ConfigWithFloat, testing::Values(GetTestCases("--float")));
+
+INSTANTIATE_TEST_SUITE_P(LstmExtra, ConfigWithHalf, testing::Values(GetTestCases("--half")));
```
atamazov: [Notice] HALF is not necessary here, but again, I am not against this. Sorry for confusion

/cc @shurale-nkn

test/gtest/deepbench_lstm.cpp
```diff
@@ -108,40 +106,67 @@ TEST_P(ConfigWithFloat, FloatTest)
...
-    {flags + " --batch-size 8 --seq-len 50 --vector-len 1536 --hidden-size 1536" + commonFlags},
-    {flags + " --batch-size 16 --seq-len 50 --vector-len 1536 --hidden-size 1536" + commonFlags},
-    {flags + " --batch-size 32 --seq-len 50 --vector-len 1536 --hidden-size 1536" + commonFlags},
-    {flags + " --batch-size 16 --seq-len 150 --vector-len 256 --hidden-size 256" + commonFlags},
-    {flags + " --batch-size 32 --seq-len 150 --vector-len 256 --hidden-size 256" + commonFlags},
-    {flags + " --batch-size 64 --seq-len 150 --vector-len 256 --hidden-size 256" + commonFlags}
-        // clang-format on
-    };
+        {flags + " --batch-size 16 --seq-len 25 --vector-len 512 --hidden-size 512" + commonFlags},
+        {flags + " --batch-size 32 --seq-len 25 --vector-len 512 --hidden-size 512" + commonFlags},
```
atamazov: [Notice] @cderb Well, this looks much worse ;( I am not against reverting this (if you wish too).

atamazov (approved):
LGTM!

test/lstm.hpp
atamazov: [resolved]

atamazov: @cderb @JehandadKhan 
> Though I think the eventual goal is to convert the tests referenced by "--all" into gtest.

Sure. I can create a sample implementation of this based on principles described at https://github.com/ROCm/MIOpen/pull/2545#discussion_r1423229457. Please let me know if you need this sample.

test/gtest/lstm_extra.cpp
atamazov: [resolved] moved to another thread

test/gtest/lstm_extra.cpp
atamazov: see https://github.com/ROCm/MIOpen/pull/2545#discussion_r1423862205

test/gtest/lstm_extra.cpp
atamazov: [resolved]

atamazov:
@cderb Oh, no, can you please also resolve https://github.com/ROCm/MIOpen/pull/2545#discussion_r1423851989 prior merging? Thanks!

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
cderb: Consider it a bonus feature.

test/gtest/deepbench_lstm.cpp
cderb: I'll take a look.

cderb: added

bghimireamd:
@shurale-nkn both float and half are now being [tested ](https://github.com/ROCm/MIOpen/pull/2545/files#diff-702e3215bbb7733fb3ca27aa294ed6c13541a937766aed7e60ae0510b5fbccd6R48)

junliume (approved):

junliume:
@shurale-nkn  last ping before merging.

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+class ConfigWithHalf : public testing::TestWithParam<std::vector<TestCase>>
+{
+};
+
+static bool IsTestSupportedForDevice()
+{
+    using namespace miopen::debug;
+    using e_mask = enabled<Gpu::gfx94X, Gpu::gfx103X, Gpu::gfx110X>;
+    using d_mask = disabled<Gpu::gfx900, Gpu::gfx906, Gpu::gfx908, Gpu::gfx90A>;
```
shurale-nkn: @junliume @atamazov 
 is it ok? we want to remove this test from CI? 
At Jenkins "deepbench_lstm and extra" skipped at every stage. 

as I remember this is default settings https://github.com/ROCm/MIOpen/blob/0efa608226ea94358543f6f0070cd04e61de034c/test/CMakeLists.txt#L524

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+class ConfigWithHalf : public testing::TestWithParam<std::vector<TestCase>>
+{
+};
+
+bool IsTestSupportedForDevice()
+{
+    using namespace miopen::debug;
+    using e_mask = enabled<Gpu::gfx94X, Gpu::gfx103X, Gpu::gfx110X>;
+    using d_mask = disabled<Gpu::gfx900, Gpu::gfx906, Gpu::gfx908, Gpu::gfx90A>;
```
shurale-nkn: same question.

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,193 @@
...
+        },
+        env_vars);
+}
+
+auto GetTestCases(std::string precision)
+{
+    const auto env = std::tuple{
+        std::pair{ENV(MIOPEN_TEST_DEEPBENCH), std::string_view("ON")},
+        std::pair{ENV(MIOPEN_TEST_FLOAT),
+                  precision == "--float" ? std::string_view("ON") : std::string_view("OFF")},
```
shurale-nkn: How this should work? What is the use case?
I can't understand this code design, why do you need an environment variables?

shurale-nkn (dismissed):
I don't understand why it is necessary to always copy a function
_`Run2dDriver(miopenDataType_t prec)
`_ It always contains the same algorithm. 

`INSTANTIATE_TEST_SUITE_P `- will define `ConfigWithType `class contend, In other words list of tests. 
`TEST_P` will define code of test which will take this tests as `vector< pair<env, string>>` .  This code always calls the _`Run2dDrive`r_ function with `miopenDataType_t `argument, just to get list of argument's from right class, but this is not needed because you can send it as _`Run2dDriver(float, GetParam())`_ argument (as recommended in GoogleTest guide).

test/gtest/deepbench_lstm.cpp
atamazov: @shurale-nkn See #1843. CTest is passing env vars to gtests:
https://github.com/ROCm/MIOpen/blob/25a399f9127a67567a027c900dc1da1c6564f61d/test/gtest/CMakeLists.txt#L33

test/gtest/deepbench_lstm.cpp
atamazov: @shurale-nkn Good finding.

@cderb The original code does not disable the test for gfx900/906/908/90a. d_mask should be empty (`Gpu::Default`, see also https://github.com/ROCm/MIOpen/pull/2605/files#r1436162312)

test/gtest/deepbench_lstm.cpp
atamazov: @shurale-nkn But in this specific case this code is incorrect.

@cderb The original test does not **_set_** the environment so this one should do the same. What we need to do is **_check_** the environment. I'll cover this in my review.

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,193 @@
...
+        std::pair{ENV(MIOPEN_TEST_HALF),
+                  precision == "--half" ? std::string_view("ON") : std::string_view("OFF")}};
```
atamazov: This is not needed.

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    std::string flags = "test_lstm --verbose " + precision;
```
atamazov, line 49: The original test is only enabled for FP32, which is the default for test_lstm, so we can either pass `--float` or nothing.

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+#include "lstm.hpp"
+#include "get_handle.hpp"
+#include <miopen/env.hpp>
+#include <gtest/gtest_common.hpp>
+#include <gtest/gtest.h>
+
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_DEEPBENCH)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_FLOAT)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_HALF)
```
atamazov: These two aren't needed.

test/gtest/deepbench_lstm.cpp
shurale-nkn: @atamazov current implementation will rewrite any passed variable.

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+namespace deepbench_lstm {
+using EnvType = std::tuple<std::pair<miopen::env::MIOPEN_TEST_DEEPBENCH, std::string_view>,
+                           std::pair<miopen::env::MIOPEN_TEST_FLOAT, std::string_view>,
+                           std::pair<miopen::env::MIOPEN_TEST_HALF, std::string_view>>;
+
+static bool Skip(miopenDataType_t prec)
+{
+    bool flag = !miopen::IsEnabled(ENV(MIOPEN_TEST_DEEPBENCH));
```
atamazov: This function should allow the test to run
- if `MIOPEN_TEST_FLOAT_ARG` is equal "--float" or empty
- and if `MIOPEN_TEST_DEEPBENCH` is enabled.

Just like the original test.

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,204 @@
...
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_HALF)
+
+namespace lstm_extra {
+using EnvType = std::tuple<std::pair<miopen::env::MIOPEN_TEST_ALL, std::string_view>,
+                           std::pair<miopen::env::MIOPEN_TEST_FLOAT, std::string_view>,
+                           std::pair<miopen::env::MIOPEN_TEST_HALF, std::string_view>>;
+
+bool Skip(miopenDataType_t prec)
+{
+    bool flag = !miopen::IsEnabled(ENV(MIOPEN_TEST_ALL));
```
atamazov: This function should allow the test to run
- if `MIOPEN_TEST_FLOAT_ARG` is equal "--float" or empty
- and if `MIOPEN_TEST_ALL` is not disabled.

Under CTest control, the test would behave just like the original test.

If run without parameters (i.e. with clean environment), the test would work as well (which is what we want in the future).

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include "lstm.hpp"
+#include "get_handle.hpp"
+#include <miopen/env.hpp>
+#include <gtest/gtest_common.hpp>
+#include <gtest/gtest.h>
+#include <boost/algorithm/string.hpp>
+
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_ALL)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_FLOAT)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_HALF)
```
atamazov: Not needed
```suggestion
```

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,204 @@
...
+}
+
+auto GetTestCases(std::string precision)
+{
+    const auto env = std::tuple{
+        std::pair{ENV(MIOPEN_TEST_ALL), std::string_view("ON")},
+        std::pair{ENV(MIOPEN_TEST_FLOAT),
+                  precision == "--float" ? std::string_view("ON") : std::string_view("OFF")},
+        std::pair{ENV(MIOPEN_TEST_HALF),
+                  precision == "--half" ? std::string_view("ON") : std::string_view("OFF")}};
```
atamazov: This is not needed
```suggestion
```

test/gtest/lstm_extra.cpp
atamazov: Yeah, `disabled` should match the original code (Gpu::none).

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    std::string flags       = "test_lstm --verbose " + precision;
```
atamazov, line 49: Always pass "--float" or nothing.

[Warning] If you want to pass "--float" explicitly, then please check if options are parsed correctly by the test. I remember some quirks related to the ordering of options.

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,204 @@
...
+    case miopenBFloat8:
+    case miopenInt8:
+    case miopenBFloat16:
+    case miopenInt32:
+    case miopenDouble:
+        FAIL() << "miopenInt8, miopenBFloat16, miopenInt32, "
+                  "miopenDouble, miopenFloat8, miopenBFloat8 "
+                  "data types not supported by "
+                  "lstm_extra test";
+
```
atamazov: Not needed.
```suggestion
```

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,204 @@
...
+} // namespace lstm_extra
+using namespace lstm_extra;
+
+TEST_P(ConfigWithFloat, FloatTest_lstm_extra) { Run2dDriver(miopenFloat); };
+
+TEST_P(ConfigWithHalf, HalfTest_lstm_extra) { Run2dDriver(miopenHalf); };
+
+INSTANTIATE_TEST_SUITE_P(LstmExtra, ConfigWithFloat, testing::Values(GetTestCases("--float")));
+
+INSTANTIATE_TEST_SUITE_P(LstmExtra, ConfigWithHalf, testing::Values(GetTestCases("--half")));
```
atamazov: Let's concentrate on functionally correct reworking and get rid of HALF testing in this PR.

atamazov (changes requested):

test/gtest/deepbench_lstm.cpp
atamazov: @shurale-nkn Yeah it should be fixed

cderb: There appears to have been some changes while I was away, I'll be making some corrections shortly.

test/gtest/deepbench_lstm.cpp
cderb: will be commenting out the half tests

test/gtest/lstm_extra.cpp
xinlipn: @atamazov , just to clarify this function looks redundant to me since the env var wouldn't be needed. 

https://github.com/ROCm/MIOpen/blob/3a32d798353462670c003ae0d68fbd78261b4874/test/gtest/lstm_extra.cpp#L39C1-L55C2

But the above mentioned line should stay to get "params" except for this line since half precision is removed, right? Thanks

https://github.com/ROCm/MIOpen/blob/3a32d798353462670c003ae0d68fbd78261b4874/test/gtest/lstm_extra.cpp#L140

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved] not necessary here

test/gtest/lstm_extra.cpp
atamazov: [resolved] moved to another review thread

test/gtest/lstm_extra.cpp
atamazov: [resolved]

test/gtest/lstm_extra.cpp
atamazov: [resolved]

test/gtest/lstm_extra.cpp
atamazov: [resolved]

test/gtest/lstm_extra.cpp
atamazov: [resolved]

atamazov:
I'll take another review round tomorrow.

test/gtest/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ function(add_gtest TEST_NAME TEST_CPP)
   if(NOT WIN32) # TODO: cannot run on Windows due to missing DLL dependencies
     # Enable CMake to discover the test binary
     # Extend GTest DISCOVERY_TIMEOUT to 5 mins
-    gtest_discover_tests(${TEST_NAME} DISCOVERY_TIMEOUT 300 PROPERTIES ENVIRONMENT "MIOPEN_USER_DB_PATH=${CMAKE_CURRENT_BINARY_DIR};MIOPEN_TEST_FLOAT_ARG=${MIOPEN_TEST_FLOAT_ARG};MIOPEN_TEST_ALL=${MIOPEN_TEST_ALL};MIOPEN_TEST_MLIR=${MIOPEN_TEST_MLIR};MIOPEN_TEST_COMPOSABLEKERNEL=${MIOPEN_TEST_COMPOSABLEKERNEL};CODECOV_TEST=${CODECOV_TEST};MIOPEN_TEST_DBSYNC=${MIOPEN_TEST_DBSYNC}")
+    gtest_discover_tests(${TEST_NAME} DISCOVERY_TIMEOUT 300 PROPERTIES ENVIRONMENT "MIOPEN_USER_DB_PATH=${CMAKE_CURRENT_BINARY_DIR};MIOPEN_TEST_FLOAT_ARG=${MIOPEN_TEST_FLOAT_ARG};MIOPEN_TEST_ALL=${MIOPEN_TEST_ALL};MIOPEN_TEST_MLIR=${MIOPEN_TEST_MLIR};MIOPEN_TEST_COMPOSABLEKERNEL=${MIOPEN_TEST_COMPOSABLEKERNEL};CODECOV_TEST=${CODECOV_TEST};MIOPEN_TEST_DBSYNC=${MIOPEN_TEST_DBSYNC};MIOPEN_TEST_DEEPBENCH=${MIOPEN_TEST_DEEPBENCH};MIOPEN_TEST_FLOAT=${MIOPEN_TEST_FLOAT};MIOPEN_TEST_HALF=${MIOPEN_TEST_HALF}")
```
atamazov:
```suggestion
    gtest_discover_tests(${TEST_NAME} DISCOVERY_TIMEOUT 300 PROPERTIES ENVIRONMENT "MIOPEN_USER_DB_PATH=${CMAKE_CURRENT_BINARY_DIR};MIOPEN_TEST_FLOAT_ARG=${MIOPEN_TEST_FLOAT_ARG};MIOPEN_TEST_ALL=${MIOPEN_TEST_ALL};MIOPEN_TEST_MLIR=${MIOPEN_TEST_MLIR};MIOPEN_TEST_COMPOSABLEKERNEL=${MIOPEN_TEST_COMPOSABLEKERNEL};CODECOV_TEST=${CODECOV_TEST};MIOPEN_TEST_DBSYNC=${MIOPEN_TEST_DBSYNC};MIOPEN_TEST_DEEPBENCH=${MIOPEN_TEST_DEEPBENCH};
```

test/gtest/deepbench_lstm.cpp
atamazov, line 1: Generally, this is deja vu. Please follow design of test/gtest/deepbench_rnn.cpp from #2550

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,166 @@
...
+void Run2dDriver(miopenDataType_t prec)
+{
+    if(!IsTestSupportedForDevice())
+    {
+        GTEST_SKIP();
+    }
+    if(Skip(prec))
+    {
+        GTEST_SKIP();
+    }
```
atamazov:
```suggestion
    if(!miopen::IsEnabled(ENV(MIOPEN_TEST_DEEPBENCH)))
        GTEST_SKIP();

```

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,166 @@
...
+    case miopenBFloat16:
+    case miopenInt32:
+    case miopenDouble:
+        FAIL() << "miopenInt8, miopenBFloat16, miopenInt32, "
+                  "miopenDouble, miopenFloat8, miopenBFloat8 "
+                  "data types not supported by "
+                  "deepbench_lstm test";
+
+    default: params = ConfigWithFloat::GetParam();
+    }
```
atamazov:
```suggestion
    std::vector<std::string> params = ConfigWithFloat::GetParam();
```

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+};
+
+static bool IsTestSupportedForDevice()
+{
+    using namespace miopen::debug;
+    using e_mask = enabled<Gpu::gfx94X, Gpu::gfx103X, Gpu::gfx110X, Gpu::gfx908>;
+    using d_mask = disabled<Gpu::Default>;
+    return miopen::debug::IsTestSupportedForDevice<d_mask, e_mask>();
+}
```
atamazov: We do not need this.
```suggestion
```

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,166 @@
...
+    case miopenFloat8:
+    case miopenBFloat8:
+    case miopenInt8:
+    case miopenBFloat16:
+    case miopenInt32:
+    case miopenDouble:
+    default: MIOPEN_THROW("Unsupported datatype");
+    }
+    return true;
+}
```
atamazov: Not needed.
```suggestion
```

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+static bool IsTestSupportedForDevice()
+{
+    using namespace miopen::debug;
+    using e_mask = enabled<Gpu::gfx94X, Gpu::gfx103X, Gpu::gfx110X, Gpu::gfx908>;
+    using d_mask = disabled<Gpu::Default>;
+    return miopen::debug::IsTestSupportedForDevice<d_mask, e_mask>();
+}
+
+void Run2dDriver(miopenDataType_t prec)
```
atamazov:
```suggestion
void Run2dDriverFloat()
```

test/gtest/deepbench_lstm.cpp
```diff
@@ -0,0 +1,166 @@
...
+        test_drive<lstm_driver>(ptrs.size(), ptrs.data());
+        auto capture = testing::internal::GetCapturedStderr();
+        std::cout << capture;
+    }
+};
+} // namespace deepbench_lstm
+
+using namespace deepbench_lstm;
+
+TEST_P(ConfigWithFloat, FloatTest_deepbench_lstm) { Run2dDriver(miopenFloat); };
```
atamazov:
```suggestion
TEST_P(ConfigWithFloat, FloatTest_deepbench_lstm) { Run2dDriverFloat(); };
```

atamazov (changes requested):
deepbench_lstm.cpp: overdesign

test/gtest/deepbench_lstm.cpp
atamazov: Let's add more types (and parametrize the function) when necessary.

atamazov:
More reviews tomorrow.

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+    using namespace miopen::debug;
+    using e_mask = enabled<Gpu::gfx94X, Gpu::gfx103X, Gpu::gfx110X>;
+    using d_mask = disabled<Gpu::Default>;
+    return ::IsTestSupportedForDevMask<d_mask, e_mask>();
+}
+
+void Run2dDriver(miopenDataType_t prec)
+{
+    if(!IsTestSupportedForDevice())
```
atamazov:
```suggestion
    if(!(IsTestSupportedForDevice() //
        && (miopen::IsUnset(ENV(MIOPEN_TEST_ALL)) // standalone run
            || (miopen::IsEnabled(ENV(MIOPEN_TEST_ALL)) // or --float full tests enabled
                && miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG)) == "--float"))))
```
the test should be run if supported by device AND (standalone run OR full tests enabled with "--float").

test/gtest/lstm_extra.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+using TestCase = decltype(GetTestCases({}))::value_type;
+
+class ConfigWithFloat : public testing::TestWithParam<std::vector<TestCase>>
+{
+};
+
+class ConfigWithHalf : public testing::TestWithParam<std::vector<TestCase>>
+{
+};
```
atamazov: [Recommendation] Remove unused code.
```suggestion
```

atamazov (changes requested):
Almost good.

test/gtest/lstm_extra.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/lstm.hpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/lstm_extra.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/lstm_extra.cpp
atamazov: [resolved]

test/gtest/CMakeLists.txt
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

test/gtest/deepbench_lstm.cpp
atamazov: [resolved]

junliume:
@cderb could you help to resolve the conflicts? This PR is almost ready. Thanks!

test/gtest/conv_trans.cpp
```diff
@@ -43,7 +43,7 @@ void GetArgs(const std::string& param, std::vector<std::string>& tokens)
         tokens.push_back(*begin++);
 }
 
-class ConfigWithFloat : public testing::TestWithParam<std::vector<std::string>>
+class ConfigWithFloat_conv_trans : public testing::TestWithParam<std::vector<std::string>>
```
atamazov: [Question] Why we need `ConfigWithFloat_conv_trans`? We are within the `conv_trans` namespace already. Ditto other similar places. Or maybe this is somehow related to GTest?

atamazov (approved):
🟢 I think this can be merged.

junliume (approved):

test/gtest/conv_trans.cpp
atamazov: @xinlipn What about

> Ditto other similar places.

e.g. `ConfigWithFloat_deepbench_lstm`, `ConfigWithFloat_lstm_extra`?

xinlipn: @atamazov , thanks for picking them up. ConfigWithFloat would pass locally, but I recall CI complained about duplicated names, thus the change. 

First reverted back to ConfigWithFloat  in conv_trans.cpp for test. Now reverted ConfigWithFloat  in all files.

junliume:
@xinlipn some issues:
```
[2024-02-12T17:02:43.682Z] [ ERROR ] /opt/rocm/cget/build/tmp-c4f1a42bcee536579a1248d25169f834201170c187fc6d80e8c510a7d28f6f48/googletest-1.14.0/googletest/src/gtest.cc:2788:: /home/jenkins/workspace/MLLIBS_MIOpen_cderb_gtest/test/gtest/deepbench_lstm.cpp:110: Attempted redefinition of test suite ConfigWithFloat.
```

atamazov:
@junliume @xinlipn that's terrific

xinlipn:
> @xinlipn some issues:
> 
> ```
> [2024-02-12T17:02:43.682Z] [ ERROR ] /opt/rocm/cget/build/tmp-c4f1a42bcee536579a1248d25169f834201170c187fc6d80e8c510a7d28f6f48/googletest-1.14.0/googletest/src/gtest.cc:2788:: /home/jenkins/workspace/MLLIBS_MIOpen_cderb_gtest/test/gtest/deepbench_lstm.cpp:110: Attempted redefinition of test suite ConfigWithFloat.
> ```

This is caused by changing back the test suite names from ConfigWithFloat_conv_trans, ConfigWithFloat_deepbench_lstm, ConfigWithFloat_lstm_extra, to ConfigWithFloat to address @atamazov 's comments. This is why ConfigWithFloat  was renamed before. I'll change them back.

PR #2557, opened 2023-11-28
[test] fix gemm driver dataType initialization

### LINKED ISSUE #2505: MIOpenDriver gemm is broken and MIOpenDriver gemmfp16 is disabled without a reason
Hi, 

While testing build result of MIOpen 5.7.1 ebuild, I noticed that `MIOpenDriver gemm` fails in https://github.com/ROCmSoftwarePlatform/MIOpen/blob/rocm-5.7.1/src/gemm_v2.cpp#L672 after falling through switch above it with uninitialized gemm_desc.dataType.

Additionally I noticed that GemmDriver also works fine with float16. It looks like `gemmfp16` flag was commented out since initial commit.

Here is the fix I've checked https://github.com/AngryLoki/MIOpen/commit/717861b3863ff5ea7a87bcec10c88beca6ae9e74
Feel free to apply it as is or partially up to your decision.

CAHEK7:
The fix properly initializes gemm dataType and also makes driver more dataType friendly, but does not enable `fp16\bfp16` since verification fails. It seems enabling half/bfloat16 types is not so trivial as just create a `Driver` with proper type.

Closes #2505

CC: @atamazov @AngryLoki @junliume @JehandadKhan

### DIFF
--- a/driver/gemm_driver.hpp
+++ b/driver/gemm_driver.hpp
@@ -93,10 +93,12 @@ void callCpuGemmStridedBatched(bool isColMajor,
                                         : a_offset + strideA * bi + lda * mi + ki;
                     int bindex = transB ? b_offset + strideB * bi + ldb * ni + ki
                                         : b_offset + strideB * bi + ldb * ki + ni;
-                    y += a_ptr[aindex] * b_ptr[bindex];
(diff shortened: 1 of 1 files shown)


CAHEK7:
Used wrong branch for PR.

PR #2558, opened 2023-11-28
[Test] fix gemm driver dataType initialization

CAHEK7:
The fix properly initializes gemm dataType and also makes driver more dataType friendly, but does not enable fp16\bfp16 since verification fails. It seems enabling half/bfloat16 types is not so trivial as just create a Driver with proper type.

Related to https://github.com/ROCmSoftwarePlatform/MIOpen/issues/2505

CC: @atamazov @AngryLoki @junliume @JehandadKhan


atamazov:
@CAHEK7 special thinks for assigning proper attributes for this PR! 🤝

driver/gemm_driver.hpp
```diff
@@ -93,10 +93,12 @@ void callCpuGemmStridedBatched(bool isColMajor,
                                         : a_offset + strideA * bi + lda * mi + ki;
                     int bindex = transB ? b_offset + strideB * bi + ldb * ni + ki
                                         : b_offset + strideB * bi + ldb * ki + ni;
-                    y += a_ptr[aindex] * b_ptr[bindex];
+                    y += static_cast<double>(a_ptr[aindex]) * static_cast<double>(b_ptr[bindex]);
```
atamazov, line 96: 👍

atamazov (approved):

atamazov:
@CAHEK7 
> It seems enabling half/bfloat16 types is not so trivial as just create a Driver with proper type.

Yeah, for validation of fp16 primitives we use double reference. And basically I do not see any harm in using double reference for all datatypes.

But what about https://github.com/AngryLoki/MIOpen/commit/717861b3863ff5ea7a87bcec10c88beca6ae9e74, does it work fine together with this PR?

CAHEK7:
It is already a `double` reference and I did few changes to make it more precise. And it is exactly https://github.com/AngryLoki/MIOpen/commit/717861b3863ff5ea7a87bcec10c88beca6ae9e74 just without fp16 support since verification failed locally.

CAHEK7:
@junliume @JehandadKhan

JehandadKhan (approved):

AngryLoki:
> But what about [AngryLoki@717861b](https://github.com/AngryLoki/MIOpen/commit/717861b3863ff5ea7a87bcec10c88beca6ae9e74), does it work fine together with this PR?

Agreed with cast to double. Ignore my fork, it is not attached anywhere. Patch for Gentoo https://github.com/gentoo/gentoo/blob/718b88b72a3c75fcdd959bf3dadcaa20a1904b6c/sci-libs/miopen/files/miopen-5.7.1-fix-miopendriver-gemm.patch still enables fp16, as LLVM/Clang is patched there.

atamazov:
@AngryLoki @JehandadKhan @junleemoreh This resolves the "_MIOpenDriver gemm is broken_" part of #2505. I think this is ready to merge.

junliume (approved):

PR #2559, opened 2023-11-28
fix compilation without CK on Windows

apwojcik:
The porting of Composable Kernels to Windows is pending. The PR fixes the compilation when `MIOPEN_USE_COMPOSABLEKERNEL` is `OFF`.

### DIFF
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -608,13 +608,13 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
                        ++id,
                        conv::ConvHipImplicitGemm3DGroupBwdXdlops{},
                        miopenConvolutionAlgoImplicitGEMM);
-#if MIOPEN_USE_COMPOSABLEKERNEL					   
+#if MIOPEN_USE_COMPOSABLEKERNEL
     Register(registry, ++id, Primitive::Batchnorm, batchnorm::BnCKFwdInference{}.SolverDbId());
     Register(registry, ++id, Primitive::Batchnorm, batchnorm::BnCKBwdBackward{}.SolverDbId());
     Register(registry, ++id, Primitive::Batchnorm, batchnorm::BnCKFwdTraining{}.SolverDbId());
     Register(registry, ++id, Primitive::Normalization, norm::Layernorm2DCKForward{}.SolverDbId());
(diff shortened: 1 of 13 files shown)


CAHEK7 (changes requested):
I don't think it should be solved in this manner (unless it's a super urgent case).

As far as I know, all the algorithms should report that they are not applicable if there is no CK support provided, but must be compiled even with that limited functionality - just to report that they are not supported.

The macro must not go outside of the kernels\algorithms and pollute a global scope (and CMakeList as well).

Moreover, recently I've tried to compile everything without CK support, and it worked fine.
Could you provide what exactly happened on windows while you are trying to build the project with ` -DMIOPEN_USE_COMPOSABLEKERNEL=Off `?

Btw, https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/src/solver/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp is not excluded from a build in your PR, but it uses CK, it means it "works" and it doesn't break windows build.
So all the build problems should be fixed in the same way.

src/solver.cpp
```diff
@@ -580,11 +582,13 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
     Register(registry, ++id, Primitive::Fusion, fusion::BnFwdInferActivationFused{}.SolverDbId());
     Register(registry, ++id, Primitive::Fusion, fusion::BnFwdTrgActivationFused{}.SolverDbId());
     Register(registry, ++id, Primitive::Fusion, fusion::BnBwdTrgActivationFused{}.SolverDbId());
+#if MIOPEN_USE_COMPOSABLEKERNEL
     Register(registry,
              ++id,
              Primitive::Fusion,
              fusion::ConvCKIgemmFwdBiasActivFused{}.SolverDbId(),
              miopenConvolutionAlgoImplicitGEMM);
+#endif
```
CAHEK7, line 591: This definitely breaks stable monotonic `id` which must be compatible across all of the MIOpen versions.

apwojcik:
@CAHEK7 I turned off the changes from the PR, and it does compile without them. That is precisely what you said. I am closing the PR. It must have been early in the porting process that we hit the build brake because of CK, which propagated. Thanks!

PR #2563, opened 2023-11-30
[MHA][tests] Cpu multi head attention

bghimireamd:
First attempt for cpu implementation of multi-head attention. Only inference is done in this PR.


test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,414 @@
...
+
+namespace test {
+namespace cpu {
+
+template <typename T>
+auto squareRoot(T value) -> decltype(std::sqrt(value))
+{
+    static_assert(std::is_arithmetic<T>::value, "squareRoot only supports arithmetic types");
+    return std::sqrt(value);
+}
```
CAHEK7: Seems it just returns `std::sqrt`, is there any purpose for this?

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,414 @@
...
+
+template <typename T>
+void BrodcastParMul(tensor<T> A_mat, tensor<T> B_mat, tensor<T>& ret_value)
+{
+    // std::mutex coutMutex;
+    size_t A_col = A_mat.desc.GetLengths()[3];
+    ret_value.par_for_each([&](size_t b_id, size_t h_id, size_t s_id, size_t d_id) {
+        T sum(0);
```
CAHEK7: I would recommend to use `double` as accumulator.

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,414 @@
...
+}
+
+template <typename T>
+void ParMul(tensor<T> A_mat, tensor<T> B_mat, tensor<T>& ret_value, size_t key_dim = 1)
+{
+    std::mutex coutMutex;
+    size_t A_col = A_mat.desc.GetLengths()[3];
+    ret_value.par_for_each([&](size_t b_id, size_t h_id, size_t s_id, size_t d_id) {
+        T sum(0);
+        std::lock_guard<std::mutex> guard(coutMutex);
```
CAHEK7: If `ret_value.par_for_each` access to unique elements of `ret_value` then mutex is not needed, especially before the loop.
Currently it serializes almost all the computation.

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,414 @@
...
+template <typename T>
+void ParMulAttention(tensor<T> A_mat, tensor<T> B_mat, tensor<T>& ret_value, size_t d_k)
+{
+    std::mutex coutMutex;
+    // sequence length
+    size_t A_col = A_mat.desc.GetLengths()[3];
+    ret_value.par_for_each([&](size_t b_id, size_t h_id, size_t s_id, size_t d_id) {
+        T sum(0);
+        size_t dk_id = d_id % d_k;
+        std::lock_guard<std::mutex> guard(coutMutex);
```
CAHEK7: See `void ParMul` comment.

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,414 @@
...
+    size_t head   = tensor_val.desc.GetLengths()[1];
+    size_t sq_len = tensor_val.desc.GetLengths()[2];
+
+    m_tensor = tensor<T>{batch, head, sq_len, 1};
+
+    for(int b = 0; b < batch; ++b)
+    {
+        for(int h = 0; h < head; ++h)
+        {
+            T max_score_in_row;
```
CAHEK7: The value can be assigned to `m_tensor` directly, without this variable.

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,414 @@
...
+    size_t head   = tensor_val.desc.GetLengths()[1];
+    size_t sq_len = tensor_val.desc.GetLengths()[2];
+
+    zinv_tensor = tensor<T>{batch, head, sq_len, 1};
+
+    for(int b = 0; b < batch; ++b)
+    {
+        for(int h = 0; h < head; ++h)
+        {
+            T sum(0);
```
CAHEK7: The value can be assigned to `zinv_tensor` directly, without this variable.

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,414 @@
...
+        // //   9) Zinv
+        Zinv(q_dot_k_transpose, zinv_tensors);
+        // //  10) Zinv reciprocal
+        ZinvMultiply(q_dot_k_transpose, zinv_tensors);
+
+        // //  11) drop out
+        DropOut(q_dot_k_transpose, cpu_mha_test_case.drop_out_rate);
+
+        // //  11.1) drop out scalse
+        double drop_out_scale = 1.0 / (1.0 - d_k);
```
CAHEK7: scale depends on `drop_out_rate`.

CAHEK7 (changes requested):

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,513 @@
...
+    trans_mat.par_for_each([&](size_t b_id, size_t sc_id, size_t h_id, size_t sl_id, size_t dk_id) {
+        trans_mat(b_id, sc_id, h_id, sl_id, dk_id) = mat_val(b_id, sc_id, h_id, dk_id, sl_id);
+    });
+}
+
+template <class T>
+void DropOut(tensor<T>& q_dot_k_transpose, const double& drop_out_rate)
+{
+    tensor<T> rand_dis(q_dot_k_transpose.desc.GetLengths());
+    rand_dis.generate(GenData<T>{});
```
CAHEK7: We won't get similar mask tensor on gpu. It would be hardly possible to compare the results.

CAHEK7 (changes requested):
Dropout uses random number generator and if we want to get comparable results, we must use the same generators for GPU and CPU.
Take a look at the dropout tests - they have implemented cpu version of the same generator.

bghimireamd:
> Dropout uses random number generator and if we want to get comparable results, we must use the same generators for GPU and CPU. Take a look at the dropout tests - they have implemented cpu version of the same generator.

Agreed! I will change the code accordingly.

CAHEK7:
I've done some significant code refactoring:

- Separated fp32 and fp8 tests and check f8 version (previously only fp32 version checked).
- Made fp8 test self-sufficient (previously it can't be run without fp32)
- Unbound tests with hardcoded reference data from random number generator
- Used `string_view` instead of strings
- Used proper gtest macros for result verification
- Used float instead of double for some intermediate data (which is supposed to be float)

Also I merged https://github.com/ROCm/MIOpen/pull/2819

CAHEK7 (approved):
mha_helper may use some more generic variants of operations (like `Dot_4D_4D` and all the transposed variants are identical except the index order), but overall it works.
I did some testing against gpu implementation and it works fine.

It's also a base PR for my numerical fwd-mha testing, so it would be nice to merge it before.

CAHEK7:
Sidenote - this code is a subject to change during integration with GPU implementation.
If somebody has any architecture suggestions - they will be taken into account in the following PRs.

Also this PR is a base for bwd-cpu implementation.

test/gtest/mha_helper.hpp
```diff
@@ -0,0 +1,468 @@
...
+// void SetupMask(tensor<T>& mask)
+// {
+//     mask.par_for_each([&](size_t s_id, size_t p_id) {
+//         // make anything above diagonal inf
+//         if(p_id > s_id)
+//         {
+//             mask(s_id, p_id) = -std::numeric_limits<T>::infinity();
+//         }
+//     });
+// }
```
binarman, line 287: nitpick: looks redundant or comment/TODO is missing

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    void init()
+    {
+        d_k = cpu_mha_test_case.problem_dimension / cpu_mha_test_case.num_heads;
+
+        // mask = tensor<T>{
+        //     std::vector<int>{cpu_mha_test_case.sequence_length,
+        //     cpu_mha_test_case.sequence_length}};
+        // SetupMask(mask);
```
binarman, line 122: redundant?

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+        word_position = tensor<InputType>{std::vector<int>{cpu_mha_test_case.batch_size,
+                                                           cpu_mha_test_case.sequence_length,
+                                                           cpu_mha_test_case.problem_dimension}};
+
+        // since Pytorch's Y = X*W_tranpose
+        // cpu_mha_test_case.num_heads, cpu_mha_test_case.problem_dimension, d_k
+        //          need to change the dimension to
+        // cpu_mha_test_case.num_heads, d_k, cpu_mha_test_case.problem_dimension
```
binarman, line 131: and this

Vsevolod1983 (approved):

CAHEK7:
@junliume may I ask to merge it.
It prevents https://github.com/ROCm/MIOpen/pull/2875 to be reviewed, since it contains the changes from this PR.

junliume (approved):

test/tensor_util.hpp
```diff
@@ -122,4 +122,66 @@ void output_tensor_to_bin(const char* fileName, T* data, size_t dataNumItems)
     }
 }
 
+template <typename T>
+void output_tensor_to_screen(const tensor<T>& tensor_val,
```
amberhassaan, line 126: minor: simpler name could be `print_tensor`.

test/gtest/mha_helper.hpp
```diff
@@ -0,0 +1,468 @@
...
+
+float GetF8Scaling(float max_val)
+{
+    constexpr float fp8_E4M3_max = 240.0f;
+
+    return fp8_E4M3_max / max_val;
+}
+
+template <typename T>
+T FindMax4D(const tensor<T>& max_of_tensor)
```
amberhassaan, line 63: A lot of these helper functions on tensors can benefit from variadic lambda parameters like this: 
```
max_of_tensor.for_each([&] (auto... indices) {
   maxVal = std::max(maxVal, max_of_tensor(indices...));
});
```

amberhassaan (commented):
some comments.

atamazov:
@junliume Thanks for renaming this to "[MHA] CPU multi head attention" during merging.

@bghimireamd I highly recommend naming PRs as clearly as possible. For this specific PR I would consider prefixing PR name with `[MHA][tests]` 🤝

PR #2565, opened 2023-11-30
use official ZStd package from Facebook

apwojcik:
It is better to make ZStd a required dependency and use `Findzstd.cmake` from the official package.


DrizztDoUrden (approved):

atamazov:
@apwojcik If ZStd is required and intended to replace Zlib, then I would expect removal of Zlib dependency, but I do not see it here. Why?

atamazov:
@apwojcik 
> It is better to make ZStd a required dependency...

Can you please elaborate _why it is better_ (and maybe better than what)? Thanks!

pfultz2:
> use Findzstd.cmake from the official package.

I think you mean the `zstd-config.cmake`.

apwojcik:
> Can you please elaborate _why it is better_ (and maybe better than what)? Thanks!

When I wrote "better", I meant to use the module configuration scripts provided by the authors instead of us maintaining them.

apwojcik:
> > use Findzstd.cmake from the official package.
> 
> I think you mean the `zstd-config.cmake`.

Yes, that is correct.

apwojcik:
> @apwojcik If ZStd is required and intended to replace Zlib, then I would expect removal of Zlib dependency, but I do not see it here. Why?

As long as the Linux team does not plan to remove it, neither do we. ZStd is complementing existing compression algorithms. I still need to submit the PR utilizing ZStd. This PR is to prepare and introduce a small change, which is easy to back out if necessary.

atamazov:
🟢 No objections!

junliume (approved):

PR #2567, opened 2023-11-30
add a class to allow os-agnostic process execution

apwojcik:


DrizztDoUrden (approved):

apwojcik:
I found a call to `std::system()` in the `TmpDir` class - it is failing on Windows. So, I made the `Process` class available for the whole MIOpen, not just the tests.

junliume (approved):

test/include_inliner.cpp
```diff
@@ -93,19 +56,19 @@ class InlinerTest
         const auto header_src      = test_srcs.path / header_filename;
 
         // clang-format-off
-        std::ofstream(valid_src.c_str()) << "#include <" << header_filename << ">" << std::endl
-                                         << "#include \"" << header_filename << "\"" << std::endl
-                                         << "//inliner-include-optional" << std::endl
+        std::ofstream(valid_src.c_str()) << "#include <" << header_filename << ">\n"
+                                         << "#include \"" << header_filename << "\"\n"
+                                         << "//inliner-include-optional\n"
```
atamazov, line 61: @apwojcik Can you please remind why the change from `std:endl` to `\n` is necessary? The latter seems non-portable. Thanks.

apwojcik: The char `'\n'` is portable between OSes, and compilers convert it correctly (C++ and C specs guarantee that). The use of `'\n'` is recommended instead of `std::endl` for performance reasons. Clang-Tidy can detect that and report a warning. 
The complete explanation is here: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#slio50-avoid-endl

dmikushin:
Hi @apwojcik , what is going to happen if a process managed by ProcessImpl crashes? I mean, return status is one thing, but there is also signal handling. Crash is a signal, and what return status can be received if the process has crashed should be tested (I always forget the exact logic, but it isn't obvious). For your inspiration: https://github.com/dmikushin/task_manager_py/blob/master/src/task_manager.cpp#L107

dmikushin:
@apwojcik , if you allow me to offer a disruptive general comment:

I would develop such OS/platfrom support code for myself for my own education, but definitely not for industrial use, because there are too many risks to take proper care of, and many of them will pass beyond the sight of reviewers, because we are not system engineers by occupation. Therefore I think the work done in this PR should better be in large part copy-pasted from a credible source (with permissive license).

PR #2569, opened 2023-11-30
fix printf type incompatibility between type specifiers

apwojcik:
There are incompatibilities between integer models between Linux and Windows. That results in different type specifiers for long integers in the case of `printf` function. The PR replaces the `printf` function with `std::cout` stream, which resolves the incompatibility issue.


shurale-nkn (approved):
LGTM

driver/rnn_seq_driver.hpp
```diff
@@ -565,11 +565,9 @@ std::vector<int> RNNSeqDriver<Tgpu, Tref>::GetSeqLengthsFromCmdLine()
     if(seq_it != data_seq_lens.end())
     {
         auto padding_val = (seq_it != data_seq_lens.begin()) ? *(seq_it - 1) : seq_len_max;
-        printf("sampl_lens size == %lu is shmaller than time batch_size == %d, padding the rest "
-               "of data with %d\n",
-               static_cast<size_t>(std::distance(data_seq_lens.begin(), seq_it)),
-               batch_size,
-               padding_val);
+        std::cout << "sampl_lens size == " << std::distance(data_seq_lens.begin(), seq_it)
+                  << " is shmaller than time batch_size == " << batch_size
```
atamazov: shmaller 😉

atamazov (approved):
looks ghood ;)

DrizztDoUrden (approved):
I wonder if we can add some check to ban printf entirely in the codebase

driver/rnn_seq_driver.hpp
apwojcik: Done. Thanks!

junliume (approved):

PR #2570, opened 2023-11-30
enable compilation on Windows

apwojcik:
This is the semi-final PR enabling compilation on Windows. The prerequisites (dependency projects from requirements.txt) must be compiled and built separately. This PR also needs all the other Windows PRs submitted to be merged first.
One more PR will be submitted soon to introduce a "one-click" compilation from a single CMake file. That includes MIOpen and dependency projects, too.

### DIFF
--- a/driver/CMakeLists.txt
+++ b/driver/CMakeLists.txt
@@ -27,6 +27,10 @@
 find_package(Threads REQUIRED)
 
 add_executable(MIOpenDriver main.cpp InputFlags.cpp)
+if(WIN32)
+    # Refer to https://en.cppreference.com/w/cpp/language/types for details.
+    target_compile_options(MIOpenDriver PRIVATE $<BUILD_INTERFACE:$<$<CXX_COMPILER_ID:Clang>:-U__LP64__>>)
+endif()
 target_include_directories(MIOpenDriver PRIVATE ../src/kernels)
 target_link_libraries(MIOpenDriver MIOpen Threads::Threads)
 if(NOT MIOPEN_EMBED_DB STREQUAL "")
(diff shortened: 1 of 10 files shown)


junliume (approved):

DrizztDoUrden (approved):

driver/CMakeLists.txt
```diff
@@ -27,6 +27,10 @@
 find_package(Threads REQUIRED)
 
 add_executable(MIOpenDriver main.cpp InputFlags.cpp)
+if(WIN32)
+    # Refer to https://en.cppreference.com/w/cpp/language/types for details.
+    target_compile_options(MIOpenDriver PRIVATE $<BUILD_INTERFACE:$<$<CXX_COMPILER_ID:Clang>:-U__LP64__>>)
```
atamazov, line 32: @apwojcik Can you please suggest why this is needed?

Please also note that
```bash
$ /opt/rocm/llvm/bin/clang -dM -E -x c /dev/null | grep LP64
#define _LP64 1
#define __LP64__ 1
```
so maybe we need to undef _LP64 as well.

test/find_db.cpp
```diff
@@ -232,8 +237,14 @@ struct FindDbTest : test_driver
 
 int main(int argc, const char* argv[])
 {
+#ifdef _WIN32
+    SetEnvironmentVariable("MIOPEN_LOG_LEVEL", "6");
```
atamazov, line 241: @apwojcik Where is the definition of this function?

NeedsMoar: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setenvironmentvariable
It's a Win32 API function, Windows.h should be pulling it in from Winbase

PR #2543, opened 2023-11-21
Sum enhancement in case of inner dim reduce

seungmanhan:
- Minor fixes to layernorm
- Added sum operation and kernel with solver
- Added driver test and gtest for sum
- Compared to ROCm pytorch, there is a performance improvement when sum is performed on the inner dim rather than the last dim

<details><summary>float16</summary>
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">

</head>

<body link="#1155CC" vlink="#1155CC">

op_name | dtype | size | dim | model | direction | ROCm pytorch | MIOpen HIP | Improvement
-- | -- | -- | -- | -- | -- | -- | -- | --
sum | float16 | [8 120 1] | (dim(0)) | bart | fwd | 13680 | 8177 | 1.67
sum | float16 | [8 120 50265] | (dim(0)) | bart | fwd | 383217 | 151147 | 2.54
sum | float16 | [8 1023 1] | (dim(0)) | gpt_neo | fwd | 16912 | 7804 | 2.17
sum | float16 | [8 1023 50257] | (dim(0)) | gpt_neo | fwd | 2638004 | 1275145 | 2.07
sum | float16 | [8 1024 768] | (dim(0)) | gpt_neo | fwd | 51520 | 23449 | 2.20
sum | float16 | [16384 768] | (dim(0)) | gpt2 | fwd | 154496 | 114987 | 1.34
sum | float16 | [8192 3072] | (dim(0)) | gpt | fwd | 122464 | 99271 | 1.23
sum | float16 | [8192 2304] | (dim(0)) | gpt | fwd | 126048 | 93315 | 1.35
sum | float16 | [16384 3072] | (dim(0)) | gpt2 | fwd | 185552 | 166720 | 1.11
sum | float16 | [16384 2304] | (dim(0)) | gpt2 | fwd | 151808 | 154044 | 0.99
sum | float16 | [16 512 1024] | (dim(0)) | bart | fwd | 41936 | 29369 | 1.43
sum | float16 | [48 8 512 512] | (dim(0)) | t5 | fwd | 329040 | 266239 | 1.24
[...]

PR #2571, opened 2023-11-30
[HotFix] Env Var set conflicts between #2543 and #2514

junliume:
conflicts between #2543 and #2514

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 def rocmnode(name) {
-    return 'rocmtest && miopen && (' + name + ')'
+    return '(rocmtest || miopen) && (' + name + ')'
 }
 
 def miopenCheckout()
--- a/test/gtest/sum.cpp
+++ b/test/gtest/sum.cpp
@@ -31,12 +31,12 @@ MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_ALL)
 
 std::string GetFloatArg()
 {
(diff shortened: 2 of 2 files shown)


JehandadKhan (approved):

Jenkinsfile
```diff
@@ -1,5 +1,5 @@
 def rocmnode(name) {
-    return 'rocmtest && miopen && (' + name + ')'
+    return '(rocmtest || miopen) && (' + name + ')'
```
atamazov, line 2: @junliume @reidkwja May I ask why this change is necessary? The latest labeling convention I am aware of is:
- Each node that allowed for use in CI is marked with "rocmtest"
- The nodes that allowed for use by MIOpen are marked with "miopen"

That is, only nodes that have both "rocmtest" and "miopen" are enabled for MIOpen CI.

atamazov (commented):
Node labeling convention is unclear

Jenkinsfile
junliume: @atamazov  I was under the same impression as you but @pfultz2 informed us that `rocmtest` means shared nodes. So if we have MIOpen specific nodes we should not have `rocmtest` label on it.

atamazov: @junliume @pfultz2 So the new convention is:
- The CI nodes marked with "rocmtest" are allowed to be used in any project
- The CI nodes marked with "miopen" are only allowed for use by MIOpen
- ⚠️ From now on, setting both "rocmtest" and "miopen" labels is pointless; all such cases should be revised.

Am I right?

PR #2573, opened 2023-11-30
ROCm Node Fix 

reidkwja:
Fixes the ROCm Node in the CI


junliume:
@reidkwja we may not need this after #2571 Thanks!

PR #2582, opened 2023-12-05
OpenCL to HIP Kernel Convertion

ozturkosu:
Cmake/gtest implementation is in progress now, I will create PR soon.


BradPepersAMD:

PR #2598, opened 2023-12-12
Fix build with disabled CK (after #2523)

CAHEK7:
Properly guard CK usage by MIOPEN_USE_COMPOSABLEKERNEL defines.
Fixes windows build and any othe builds without CK support.  (fix for #2523)

### DIFF
--- a/src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
@@ -339,7 +339,7 @@ void PerformanceConfigHipImplicitGemmGroupFwdXdlops::HeuristicInit(
 
 bool PerformanceConfigHipImplicitGemmGroupFwdXdlops::SetNextValue(const ProblemDescription& problem)
 {
-#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+#if MIOPEN_USE_COMPOSABLEKERNEL
     if(valid_kernels.empty())
     {
         switch(problem.GetInDataType())
--- a/src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
+++ b/src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
@@ -191,7 +191,6 @@ bool ConvHipImplicitGemmGroupFwdXdlops::CheckCKApplicability(
 {
     return IsCKApplicable<DeviceOpGFwdPtrs<DataType>, CKArgs>(problem);
 }
-#endif
 
 #if MIOPEN_ENABLE_AI_KERNEL_TUNING
 static std::vector<std::string> GetKernelAsTokens(const std::string& kernel)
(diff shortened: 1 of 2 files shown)


apwojcik (approved):

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -285,7 +284,8 @@ bool PerformanceConfigHipImplicitGemmGroupFwdXdlops::RunParameterPredictionModel
     }
     return false;
 }
-#endif
+#endif // MIOPEN_ENABLE_AI_KERNEL_TUNING
+#endif // MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
```
atamazov, line 288: Is it really so that AI thing depends on CK???

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -339,6 +339,7 @@ void PerformanceConfigHipImplicitGemmGroupFwdXdlops::HeuristicInit(
 
 bool PerformanceConfigHipImplicitGemmGroupFwdXdlops::SetNextValue(const ProblemDescription& problem)
 {
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
```
atamazov: IIRC `MIOPEN_BACKEND_HIP &&` is not necessary
```suggestion
#if MIOPEN_USE_COMPOSABLEKERNEL
```

atamazov (commented):
Question

atamazov:
@JehandadKhan @CAHEK7 Build without CK seems fragile and breaks frequently. Shall we add smoke stage with -DMIOPEN_USE_COMPOSABLEKERNEL=Off, what do you think?

atamazov:
@CAHEK7 Is this PR windows-specific?

junliume:
> @CAHEK7 Is this PR windows-specific?

Yes, it is to unblock Windows CI. However, as for your question whether it breaks `-DMIOPEN_USE_COMPOSABLEKERNEL=Off`, it has become increasingly difficult to consider CK as "optional" for MIOpen.

junliume (approved):
Good patch IMHO and when resources allow we should check MIOPEN_USE_COMPOSABLEKERNEL

CAHEK7:
@atamazov it's not only windows specific, but I also can't build it on linux without CK (for some reasons I have to do it).

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
CAHEK7: This particular solver? Yes.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
CAHEK7: I guess we have to cleanup this pair - it's widely used.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
atamazov: Oops, I see now. Obvious.

atamazov:
@CAHEK7 Then please remove https://github.com/ROCm/MIOpen/labels/windows

PR #2553, opened 2023-11-27
[Test] Convert conv_igemm_dynamic_dlops etc to gTest

xinlipn:
Converted conv_igemm_dynamic_dlops etc to gTest

PR #2599, opened 2023-12-12
[gtest] Combine gtests into single binary.

cderb:
Update Cmake recipe in gtest folder to combine files into a single binary.
Resolves conflicts between sources when compiled together.

Edits:
Add an `MIOPEN_TEST_DISCRETE` CMake variable which controls whether tests are build individually or as a single binary. Default is `ON` which maintains the current behavior, however, the `Package` CI stage sets it to `OFF` so that build failures can be caught in the CI.


atamazov (commented):
No objections!

test/gtest/conv_test_base.hpp
```diff
@@ -211,3 +142,69 @@ struct ConvFwdSolverTestBase
     miopen::Allocator::ManageDataPtr wei_dev;
     miopen::Allocator::ManageDataPtr out_dev;
 };
+
+template <typename T, typename Tref, bool use_cpu_ref>
+void ConvFwdSolverTestBase<T, Tref, use_cpu_ref>::SetUpImpl(ConvTestCaseBase conv_config,
```
averinevg: Does the linker throw an error when building this code? Like "multiple definitions" or something?

cderb: All multiple definitions errors have been resolved.

CAHEK7:
Seems there is a bunch of non-prepared tests have arrived. You have to fix them too.
I know that you've done it in a different way - renamed lots of structures and so on, but in #2553 I replaced test name with following sed command, because otherwise it requires too much work to do manually:
`parallel sed -i -e "'s/^\(.*TEST_P.*\))/\1_{.})/'" {} ::: smoke*.cpp`
```diff
-TEST_P(Conv2dFloat, FloatTest)
-TEST_P(Conv2dHalf, HalfTest)
+TEST_P(Conv2dFloat, FloatTest_conv_ck_igemm_fwd_v6r1_dlops_nchw)
+TEST_P(Conv2dHalf, HalfTest_conv_ck_igemm_fwd_v6r1_dlops_nchw)
```
I'd recommend to have a separated meeting and decide what shall we do with the naming and fix it later. Now it seems it's better to implement the quickest solution.

cderb:
> Seems there is a bunch of non-prepared tests have arrived. You have to fix them too. 

Tests have been re-named.

test/conv3d.hpp
```diff
@@ -58,3 +59,19 @@ struct conv3d_driver : conv_driver<T>
         this->add(this->out_layout, "out_layout", this->generate_data({"NCDHW"}));
     }
 };
+
+class Conv3dFloat : public testing::TestWithParam<std::vector<std::string>>
```
amberhassaan, line 63: why are all these class definitions exactly the same?

test/gtest/CMakeLists.txt
```diff
@@ -46,7 +44,29 @@ foreach(SOURCE ${SOURCES})
...
 endforeach()
 
-foreach(TEST ${TESTS})
-    get_filename_component(BASE_NAME ${TEST} NAME_WE)
-    add_gtest(${BASE_NAME})
+foreach(SOURCE ${SKIP_TESTS})
+    list(REMOVE_ITEM TESTS ${CMAKE_CURRENT_SOURCE_DIR}/${SOURCE})
 endforeach()
+
+if( DISCRETE_GTEST )
```
amberhassaan: Is this flag "ON" by default or "OFF"?
IMO, this should be "ON" by default, or better yet, we rename the flag to something like "PACKAGE_ALL_GTESTS" and make it off by default.

amberhassaan (commented):
Partial review and some questions.

test/gtest/CMakeLists.txt
cderb: Undefined maps to false. So default is false here.

test/conv3d.hpp
cderb: They are ostensibly configuration definitions split by precision. Though in this instance their data is defined entirely by the test fetching function and as such have no defined members at this level. It's conceivable that they could be summarized into a single class.

test/test_env.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <miopen/env.hpp>
+
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
+
+inline bool IsTestRunWith(const char* float_arg)
+{
+    assert(float_arg != nullptr);
+    const auto& s_envVar = miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG));
+    return (s_envVar.compare(float_arg) == 0);
+}
```
CAHEK7, line 38: Seems it does not make much sense and it's just an artifact of moving the test from CMake.
Nobody runs the test like `MIOPEN_TEST_FLOAT_ARG=--float ./bin/mlir_test`, everyone runs it like `./bin/mlir_test --float`.
Moreover all the other tests have missed that kind of check.

xinlipn: > Nobody runs the test like `MIOPEN_TEST_FLOAT_ARG=--float ./bin/mlir_test`,

@CAHEK7 , my understanding is current gTests still have dependency on env var, and MIOPEN_TEST_FLOAT_ARG should be present and by default set to --float as shown in test/CMakeLists.txt below. Thus this code snippet. Does this make sense?

https://github.com/ROCm/MIOpen/blob/10250d6bcdaef2767d4886b839f639938d15e9e5/test/CMakeLists.txt#L225

> Moreover all the other tests have missed that kind of check.

All tests should include this header, test/test_env.hpp, right?

CAHEK7: Except the fact that nobody is using "MIOPEN_TEST_FLOAT_ARG" env variable except a few freshly converted testcases.
Either all the other testcases are wrong and we must fix them, or `MIOPEN_TEST_FLOAT_ARG` is not a way to switch precision.
```
~/miopen$ git grep MIOPEN_TEST_FLOAT_ARG | grep -e "cpp" -e "\.h"
test/gtest/conv_embed_db.cpp:MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
test/gtest/conv_embed_db.cpp:    const auto& s_envVar = miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG));
test/gtest/conv_hip_igemm_xdlops.cpp:MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
test/gtest/conv_hip_igemm_xdlops.cpp:    const auto& s_envVar = miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG));
test/gtest/conv_igemm_mlir.cpp:MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
test/gtest/conv_igemm_mlir.cpp:    const auto& tmp = miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG));
test/gtest/conv_igemm_mlir_xdlops.cpp:MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
test/gtest/conv_igemm_mlir_xdlops.cpp:    const auto& tmp = miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG));
test/gtest/layernorm.cpp:MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
test/gtest/layernorm.cpp:    const auto& tmp = miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG));
test/gtest/sum.cpp:MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
test/gtest/sum.cpp:    const auto& tmp = miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG));
```

And the most important thing is `set(MIOPEN_TEST_FLOAT_ARG --float)` is not an env variable. It is local cmake variable which is passed to the test as a command line argument.

JehandadKhan:
@CAHEK7 CI passes

CAHEK7:
```
Conflicting files
test/gtest/group_conv3d_bwd.hpp
test/gtest/group_conv3d_wrw.hpp
```
:cry:

amberhassaan:
> ```
> Conflicting files
> test/gtest/group_conv3d_bwd.hpp
> test/gtest/group_conv3d_wrw.hpp
> ```
> 
> 😢

Sorry that's my bad. I should have added namespaces around the test code. Was too focused on the convolution bugs.

junliume:
@amberhassaan here are a few issues let's see if we can deal with quickly: (I suspect it is due to we upgraded CI base docker to ROCm 6.0, we can suppress this warning)
```
[2023-12-21T01:30:45.364Z] /home/jenkins/workspace/LLIBS_MIOpen_cderb_gtest_package/test/gtest/fused_conv_bias_res_add_activ.cpp:73:72: error: use of function template name with no prior declaration in function call with explicit template arguments is a C++20 extension [-Werror,-Wc++20-extensions]

[2023-12-21T01:30:45.364Z]             conv_desc.GetForwardOutputTensor(input.desc, weights.desc, GetDataType<T>());

[2023-12-21T01:30:45.364Z]    
```
should it be
```
        miopen::TensorDescriptor output_desc =
            conv_desc.GetForwardOutputTensor(input.desc, weights.desc, input.desc.GetType());
```

amberhassaan:
> @amberhassaan here are a few issues let's see if we can deal with quickly: (I suspect it is due to we upgraded CI base docker to ROCm 6.0, we can suppress this warning)
> 
> ```
> [2023-12-21T01:30:45.364Z] /home/jenkins/workspace/LLIBS_MIOpen_cderb_gtest_package/test/gtest/fused_conv_bias_res_add_activ.cpp:73:72: error: use of function template name with no prior declaration in function call with explicit template arguments is a C++20 extension [-Werror,-Wc++20-extensions]
> 
> [2023-12-21T01:30:45.364Z]             conv_desc.GetForwardOutputTensor(input.desc, weights.desc, GetDataType<T>());
> 
> [2023-12-21T01:30:45.364Z]    
> ```
> 
> should it be
> 
> ```
>         miopen::TensorDescriptor output_desc =
>             conv_desc.GetForwardOutputTensor(input.desc, weights.desc, input.desc.GetType());
> ```

Will take a look in a bit.

test/gtest/CMakeLists.txt
JehandadKhan: Please check the PR description for updated info on this.

JehandadKhan (approved):

amberhassaan:
> > @amberhassaan here are a few issues let's see if we can deal with quickly: (I suspect it is due to we upgraded CI base docker to ROCm 6.0, we can suppress this warning)
> > ```
> > [2023-12-21T01:30:45.364Z] /home/jenkins/workspace/LLIBS_MIOpen_cderb_gtest_package/test/gtest/fused_conv_bias_res_add_activ.cpp:73:72: error: use of function template name with no prior declaration in function call with explicit template arguments is a C++20 extension [-Werror,-Wc++20-extensions]
> > 
> > [2023-12-21T01:30:45.364Z]             conv_desc.GetForwardOutputTensor(input.desc, weights.desc, GetDataType<T>());
> > 
> > [2023-12-21T01:30:45.364Z]    
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > should it be
> > ```
> >         miopen::TensorDescriptor output_desc =
> >             conv_desc.GetForwardOutputTensor(input.desc, weights.desc, input.desc.GetType());
> > ```
> 
> Will take a look in a bit.

OK, @JehandadKhan has fixed this issue. Thank you.

test/conv3d.hpp
atamazov, line 1: Renaming of .cpp to .hpp silently removes `test_*` executable, and, consequently, `test_* --all` from full tests etc.

test/immed_conv2d.cpp
atamazov, line 1: Ditto

test/pooling2d.hpp
atamazov, line 1: Ditto

atamazov (commented):
🔴 Silently removes some test executables and, consequently, the following tests from "Full test" stages:
- `test_immed_conv3d --all`
- `test_immed_conv2d --all`
- `test_pooling2d --all`

Must be reverted ASAP.

@junliume @cderb @JehandadKhan

test/test_env.hpp
atamazov: @CAHEK7 @cderb `MIOPEN_TEST_FLOAT_ARG` is an environment variable passed to gtests from ctest. Gtests should obey to its value as follows:
- If it is not set, then run tests for all supported datasets
- If it is set to `--float`, or `--half` etc, then only fp32, or fp16 etc. datasets should be run.

More info at #1843

test/pooling2d.hpp
```diff
@@ -24,6 +24,7 @@
  *
  *******************************************************************************/
 
+#include <gtest/gtest.h>
```
atamazov, line 27: ⚓

test/conv3d.hpp
```diff
@@ -23,12 +23,13 @@
  * SOFTWARE.
  *
  *******************************************************************************/
+#include <gtest/gtest.h>
```
atamazov, line 26: ⚓

test/conv3d.hpp
```diff
@@ -59,4 +60,18 @@ struct conv3d_driver : conv_driver<T, ConvApi::Immediate>
...
+{
+};
+
+class Conv3dBFloat16 : public testing::TestWithParam<std::vector<std::string>>
+{
+};
+
+class Conv3dInt8 : public testing::TestWithParam<std::vector<std::string>>
+{
+};
```
atamazov, line 77: These definitions are used by gtests only and should reside in gtest/*.cpp files within respective namespaces.

test/pooling2d.hpp
```diff
@@ -106,4 +107,10 @@ struct pooling2d_driver : pooling_driver<T>
...
 };
 
-int main(int argc, const char* argv[]) { test_drive<pooling2d_driver>(argc, argv); }
+class Pooling2dFloat : public testing::TestWithParam<std::vector<std::string>>
+{
+};
+
+class Pooling2dHalf : public testing::TestWithParam<std::vector<std::string>>
+{
+};
```
atamazov, line 116: Ditto

atamazov (commented):
move gtest-specific classes into gtest/*.cpp

PR #2605, opened 2023-12-14
helper for evn variables update for gtest

CAHEK7:
It's my draft version of this helper. I've made it to fix my comments in #2553 and since it's still in progress I may do some local changes, but the overall design will stay the same.

Examples:
Env variables (this is the example where semicolon separated solvers won't work, since there two FWD solvers and we don't know which will be chosen):
```C++
auto GetTestCases(std::string&& fp_opts)
{
    const auto env_fwd = std::tuple{
        std::pair{ENV(MIOPEN_FIND_MODE), std::string_view("normal")},
        std::pair{ENV(MIOPEN_DEBUG_FIND_ONLY_SOLVER), std::string_view("ConvOclDirectFwd11x11")}};

    const auto env_fwd_gen = std::tuple{
        std::pair{ENV(MIOPEN_FIND_MODE), std::string_view("normal")},
        std::pair{ENV(MIOPEN_DEBUG_FIND_ONLY_SOLVER), std::string_view("ConvOclDirectFwdGen")}};

    const auto vf = fp_opts + " --verbose --disable-backward-data --disable-backward-weights";

    return std::vector{
        // clang-format off
    std::pair{env_fwd,     vf + " --input 1 1 44 44 --weights 1 1 11 11 --pads_strides_dilations 0 0 4 4 1 1"},
    std::pair{env_fwd_gen, vf + " --input 1 1 6 6 --weights 1 1 3 3 --pads_strides_dilations 0 0 2 2 1 1"},
        // clang-format on
    };
}

using TestCase = decltype(GetTestCases({}))::value_type;

class Conv2dFloat : public testing::TestWithParam<std::vector<TestCase>>
{
};
```

Check for compatible GPU arch, more bulletproof for typos like `devName == "GFX900"`:
```C++
static bool IsTestSupportedForDevice()
{
    using e_mask = enabled<Gpu::gfx110X>;
    using d_mask = disabled<Gpu::gfx900, Gpu::gfx906, Gpu::gfx908, Gpu::gfx90A>;
    return IsTestSupportedForDevice<d_mask, e_mask>();
}
```
It matches exact the same logic in CMakeList.txt for easier verification.

> Defaults: GFX900_ENABLED, GFX906_ENABLED, GFX908_ENABLED, GFX90A_ENABLED, GFX94X_DISABLED, GFX103X_DISABLED, GFX110X_DISABLED

Invoking testcases (no need for the switch-case and all that machinery):
```C++
TEST_P(Conv2dFloat, FloatTest)
{
    if(IsTestSupportedForDevice())
        invoke_with_params<conv2d_driver>(Conv2dFloat::GetParam(), tuning_check);
    else
        GTEST_SKIP();
};
```

It also can used for the cases without env variables - just pass empty tuple.

It's critical (or blocker?) since it is related to all the gtest PRs.

@atamazov

### DIFF
--- a/src/include/miopen/env.hpp
+++ b/src/include/miopen/env.hpp
@@ -29,6 +29,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #include <miopen/errors.hpp>
@@ -205,6 +206,13 @@ void UpdateEnvVar(EnvVar, const ValueType& val)
 {
     static_assert(std::is_same_v<typename EnvVar::value_type, ValueType>);
     EnvVar::Ref().UpdateValue(val);
+}
+
+template <typename EnvVar>
+void UpdateEnvVar(EnvVar, const std::string_view& val)
+{
+    EnvVar::Ref().UpdateValue(
+        miopen::internal::ParseEnvVal<typename EnvVar::value_type>::go(val.data()));
 }
 
 } // namespace miopen
--- a/test/gtest/gtest_common.hpp
+++ b/test/gtest/gtest_common.hpp
@@ -1,0 +1,164 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2023 Advanced Micro Devices, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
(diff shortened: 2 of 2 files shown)


src/include/miopen/env.hpp
```diff
@@ -207,6 +208,13 @@ void UpdateEnvVar(EnvVar, const ValueType& val)
     EnvVar::Ref().UpdateValue(val);
 }
 
+template <typename EnvVar>
+void UpdateEnvVar(EnvVar, const std::string_view& val)
```
amberhassaan, line 212: Minor suggestion: rename to ParseEnvVar or something like that and add a comment saying that this is for tests only and normally folks should use `UpdateEnvVar` (the old one).

CAHEK7: I think none of them should be used normally. All that functionality has been introduced exceptionally for the testing purposes.
Let's rename it a bit later, after we finished the most urgent gtests tickets (just to avoid unnecessary changes for everyone).

CAHEK7:
Is the nightly branch super stale? Why does it constantly fail to fast-forward?

src/include/miopen/env.hpp
atamazov: [R] @CAHEK7 @amberhassaan 
> All that functionality has been introduced exceptionally for the testing purposes.

We can enclose this into `miopen::debug` namespace to emphasize this fact.

atamazov:
@CAHEK7 
> Is the nightly branch super stale? Why does it constantly fail to fast-forward?

Does this Q relate to this PR?

cderb (approved):
LGTM

xinlipn:
LGTM

CAHEK7:
@atamazov yes, it is constantly failed and I'm wondering how my PR can affect it.

> @CAHEK7
> 
> > Is the nightly branch super stale? Why does it constantly fail to fast-forward?
> 
> Does this Q relate to this PR?

junliume:
> @atamazov yes, it is constantly failed and I'm wondering how my PR can affect it.
> 
> > @CAHEK7
> > > Is the nightly branch super stale? Why does it constantly fail to fast-forward?
> > 
> > 
> > Does this Q relate to this PR?

Yes that GitHub action is problematic, I fixed it a few times and then it reappears, maybe I should delete that file until a stable solution is found.

junliume (approved):

test/gtest/gtest_common.hpp
```diff
@@ -0,0 +1,164 @@
...
+void tuning_check(const std::string& err)
+{
+    // TEST_TUNING - the test should fail if output contains "Error" or "failed".
+    EXPECT_FALSE(err.find("Error") != std::string::npos || err.find("failed") != std::string::npos);
+    default_check(err);
+}
+
+enum class Gpu : int
+{
+    Default = 0,
```
atamazov, line 52: @CAHEK7 Rename to `none` to allow things like `disabled<Gpu::none>` which looks cleaner than `disabled<Gpu::Default>` (what is the default GPU?)

PR #2574, opened 2023-12-01
use find_package() for Eigen and frugally-deep

apwojcik:
Both libraries provide CMake configuration files, so it is much better to use `find_package()` instead of manually maintaining the header files' discovery.

PR #2608, opened 2023-12-15
Revert "[Windows] use find_package() for Eigen and frugally-deep (#2574)"

junliume:
This reverts commit 4150134faaf9ad095a4d5d1aadb3a61cf29bb7b0.

Issue described at https://github.com/ROCm/MIOpen/issues/2607

We can work on #2574 later but now I need to unblock local dev


apwojcik:
The issue is not caused by the ROCm 5.7.0 but by the CMake version installed in the docker. The command `find_package(... GLOBAL)` requires CMake version >= 3.24

apwojcik (changes requested):
Can we test #2610 before reverting #2574 with this PR

apwojcik:
The issue is not caused by the ROCm 5.7.0 but by the CMake version installed in the docker. The command `find_package(... GLOBAL)` requires CMake version >= 3.24

atamazov:
@apwojcik I will test #2610

junliume:
Unnecessary with #2610

Issue #2607, opened 2023-12-14
[cmake] after #2574 set frugally-deep_FOUND to FALSE so package "frugally-deep" is considered to be NOT FOUND


junliume:
**Update**: Issue reproduced at newer releases, investigating if this is a more common issue than just backward incompatibility

This breaks compatibility with ROCm 5.7.0.

The build in my 5.7.0 docker (rocm/miopen:ci_429410) fails with the following:
```bash
CMake Error at src/CMakeLists.txt:736 (find_package):
  Found package configuration file:

    /usr/local/lib/cmake/frugally-deep/frugally-deepConfig.cmake

  but it set frugally-deep_FOUND to FALSE so package "frugally-deep" is
  considered to be NOT FOUND.

-- Configuring incomplete, errors occurred!
```
I had to `-DMIOPEN_ENABLE_AI_KERNEL_TUNING=Off -DMIOPEN_ENABLE_AI_IMMED_MODE_FALLBACK=Off` to W/A this.

With 6.0 RC (rocm/miopen:ci_fbea27) and 5.7.1 (rocm/miopen:ci_3eb168) dockers, everything is fine.

_Originally posted by @atamazov in https://github.com/ROCm/MIOpen/issues/2574#issuecomment-1856896416_

junliume:
To provide more info, in the failing docker above, the error is because `frugally-deep_GLOBAL_FOUND` is FALSE while `frugally-deep_FIND_REQUIRED_GLOBAL` is TRUE.
I do not know why it worked in one docker but not the other though.

junliume:
@atamazov @JehandadKhan @apwojcik Unfortunately I have to raise this ticket to blocker level, because the recommended way of building MIOpen seems to be broken:
```
export LC_ALL=C.UTF-8 && \
export LANG=C.UTF-8 && \
apt-get update -y && \
apt-get install -y sudo && \
sudo apt-get -y install git git-lfs unzip lbzip2 python3.9-venv pkg-config vim ninja-build half && \
sudo apt update && \
cd /root && git clone -b develop https://github.com/ROCmSoftwarePlatform/MIOpen && cd MIOpen && \
mkdir -p install_dir && cmake -P install_deps.cmake --minimum --prefix /root/MIOpen/install_dir && \
mkdir -p build && \
cd build && \
[...]

PR #2610, opened 2023-12-15
fix find_package(... GLOBAL) for CMake < 3.24

apwojcik:
The fix for the find_package(... GLOBAL) and use of CMake version lower than 3.24.


CMakeLists.txt
```diff
@@ -403,6 +403,11 @@ option(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY "Build with file/folder reorg wit
 set( MIOPEN_INSTALL_DIR miopen)
 set( DATA_INSTALL_DIR ${CMAKE_INSTALL_DATAROOTDIR}/miopen )
 
+if(MIOPEN_ENABLE_AI_KERNEL_TUNING OR MIOPEN_ENABLE_AI_IMMED_MODE_FALLBACK)
+    find_package(frugally-deep CONFIG REQUIRED)
```
atamazov, line 407: So we do not need 3.24 anymore? Good!

atamazov (approved):
LGTM

CMakeLists.txt
apwojcik: Correct :-)

atamazov:
Resolves #2607 and makes #2608 unnecessary.

atamazov:
🟢 Works fine with ROCm 5.7.0 and cmake 3.16.3

junliume:
Thank you for the quick fix! @apwojcik

junliume (approved):

PR #2611, opened 2023-12-15
add_gitignore : adding .gitignore file

bghimireamd:
In this PR we add .gitignore file. Currently the ignores are related to documenation.


Rmalavally (approved):

.gitignore
```diff
@@ -0,0 +1,19 @@
...
+_images/
+_static/
+_templates/
+_toc.yml
+docBin/
+_doxygen/
+_readthedocs/
+ 
+# avoid duplicating contributing.md due to conf.py
+docs/CHANGELOG.md
```
amberhassaan, line 16: why are these lines needed? These could be legit files that we want in the repository.

amberhassaan (commented):
One question and rest is good.

junliume:
@bghimireamd @JehandadKhan @atamazov We discussed about `.gitignore` before and previously decided that it should be user specific and not provided by the repo, e.g. `build` does not necessarily reside in the same directory.
What has changed and now requiring us to add the file?

atamazov:
@junliume I do not know. I have my own .gitignore and do not need this.

bghimireamd:
> @bghimireamd @JehandadKhan @atamazov We discussed about `.gitignore` before and previously decided that it should be user specific and not provided by the repo, e.g. `build` does not necessarily reside in the same directory. What has changed and now requiring us to add the file?

Oh! in that case we can place a hold on this PR. Documentation change was introduced by #2587 , which I believe generates some files during build process. To ignore those files this .gitinore was added.

We can have our own local .gitignore.

CAHEK7:
The good practice is not generating anything in the source directory. Ideally it should be read-only.

amberhassaan:
My humble opinion is that it is common for projects to have a `.gitignore` that serves the need. Folks can modify this `.gitignore` locally or if I recall, there's a way to have a `global .gitignore` that users control: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

CAHEK7:
> My humble opinion is that it is common for projects to have a `.gitignore` that serves the need.

It depends - if you don't generate anything in the build directory - you don't need default project `.gitignore`. No needs - no serve.

Each user makes `.gitignore` for himself based on a particular IDE or specific workflow - putting all possible combinations of IDE's temporary files and directories and something like `build/*` does not look like a good idea.

Issue #2593, opened 2023-12-11
GPU Segfault on 3D Int8 Convolution


JehandadKhan:
When running the convolution described by the following driver command on MI200 platform

``` bash
MIOpenDriver convint8 -n 1 -c 64 --in_d 128 -H 128 -W 128 -k 32 --fil_d 3 -y 3 -x 3 --pad_d 1 -p 1 -q 1 --conv_stride_d 1 -u 1 -v 1 --dilation_d 1 -l 1 -j 1 --spatial_dim 3 -m conv -g 1 -F 1 -t 1 -S 0
```

Running it results in the following:

```
Memory access fault by GPU node-6 (Agent handle: 0x1190f40) on address 0x7f3692c9e000. Reason: Unknown.
```

JehandadKhan:
@atamazov  Can you please take a look?

atamazov:
@JehandadKhan The reason is integer overflow. Navi21:
```bash
MIOpen(HIP): Info [ConvolutionForwardImmediate] solver_id = GemmFwdRest, workspace = 7247757312
... 
MIOpen(HIP): Info2 [Log] Kernel MIOpenUtilKernels4.cl Compile Time, ms: 155.193
MIOpen(HIP): Info2 [run] kernel_name = transpose_packed_MN2NM, global_work_dim = { 3623878656, 1, 1 }, local_work_dim = { 256, 1, 1 }
```
The solver calls
```C++
float transpose_packed_MN2NM(const Handle& handle,
                             int m,
                             int n,
                             int in_offset,
                             int out_offset,
                             ConstData_t in,
                             Data_t out,
                             miopenDataType_t type);
```
and passes 3623878656 to `out_offset` which exceeds INT_MAX.

The fix could be either narrowing the solver's applicability or fixing the math in `transpose_packed_MN2NM()`; the latter is preferable of course.

How urgent is this issue (please assign a label)?

----
Please note that:
- WS size > 4 GiB 
- grid size (3,623,878,656) and that is not far from the HIP limit (4,294,967,295).
- I am not sure if rocBLAS is able to handle matrices > 4GiB, but this is another story.
- ConvDirectNaiveConvFwd is working fine

atamazov:
@JehandadKhan Similar problem happens in `CallGemm()`

PR #2613, opened 2023-12-15
Get rid of legacy 2GiB offset limits in CallGemm*() and transpose*() internal APIs and kernels.

atamazov:
- Get rid of legacy 2GiB offset limit in
  - `CallGemm*()`
  - `transpose_packed_MN2NM()`
  - `transpose_NCHW2CNHW()` - proactive fix
  - `transpose_CNHW2NCHW()` - proactive fix

Resolves issue #2593.

----
- Proposed reviewers:
  - @JehandadKhan
  - @junliume
  - @DrizztDoUrden


atamazov:
Some issues found during manual testing. Triaging in progress. Converted to draft.

src/kernels/MIOpenUtilKernels4.cl
```diff
@@ -512,18 +514,18 @@ __kernel void transpose_packed_MN2NM(const global data_t* in,
...
     {
-        uint m_i = iDiv(i, N);
-        uint n_i = iMod(i, m_i, N);
+        const uint m_i = iDiv(i, N);
+        const uint n_i = iMod(i, m_i, N);
 
-        uint in_offset  = m_i * N + n_i + in_off;
-        uint out_offset = n_i * M + m_i + out_off;
+        const size_t in_offset  = m_i * N + n_i + in_off;
+        const size_t out_offset = n_i * M + m_i + out_off;
```
CAHEK7, line 528: Is it what do you want?
`size_t = ulong( uint( (uint_m_i * uint( int_M )) ) + uint_n_i ) + ulong_in_off;`
I mean ulong conversion happens only during the last addition, all the intermediate calculations happened in `uint32` domain.
Also `size_t` in opencl depends on CL_DEVICE_ADDRESS_BITS and can be 32 or 64, and `long` is always a 64 bit values.
Btw, there is `cl_ulong` type which is also 64bit value and it can be a good replacement of `arg_size_t`. But in common case that sort of interoperability can hardly be maintained.

dmikushin:
Just a general note. In BLAS libraries, it is common to have separate versions for different lengths of indexes. This is called LP64 versus ILP64 ("I" adds 64-bit indexing to the API functions). The rationale to have both of them is that some users want to preserve the ABI, while others need support for larger problems. Don't know how much it is our case or not, I mean to say there is a common practice in this regard that we may want to consider.

atamazov:
@CAHEK7 @JehandadKhan @DrizztDoUrden @junliume All issues resolved, local testing done, ready for review, CI testing and merging.

src/kernels/MIOpenUtilKernels4.cl
atamazov: @CAHEK7 Thanks for comments.
> Is it what do you want?
> size_t = ulong( uint( (uint_m_i * uint( int_M )) ) + uint_n_i ) + ulong_in_off;

Exactly. Computations in 32-bit domain save a bit of performance. Everywhere we assume that up to 3 dimensions can me multiplied in 32-bit domain without overflow:
https://github.com/ROCm/MIOpen/blob/c5a2384dc0f29682ed51aeccf9b981dbdf7e058f/src/solver/pooling/forwardNaive.cpp#L125-L131

> Also size_t in opencl depends on CL_DEVICE_ADDRESS_BITS and can be 32 or 64

Yes. It is 64 in all the devices we do support.

> Btw, there is cl_ulong type which is also 64bit value and it can be a good replacement of arg_size_t. But in common case > that sort of interoperability can hardly be maintained.

This won't help if CL_DEVICE_ADDRESS_BITS is 32 because in this case the device is physically unable to address buffers larger than 4GiB.

atamazov:
@dmikushin ~Yes, LP/ILP64 are supported by rocblas, https://rocm.docs.amd.com/projects/rocBLAS/en/develop/API_Reference_Guide.html#ilp64-interface.~ Only LP64 is supported by rocblas, https://rocm.docs.amd.com/projects/rocBLAS/en/docs-6.0.0/API_Reference_Guide.html#lp64-interface.

But this it irrelevant for this PR as it affects `roclas_gemm_*ex` calls that use `void*` to pass buffer addresses, int32_t for dimensions and int64_t for strides.

dmikushin:
@atamazov ,

> int32_t for dimensions and int64_t for strides

Precisely. Therefore, shouldn't `int64_t` be used in this PR instead of `size_t`?

atamazov:
@dmikushin Let's keep this as is. We can safely assume that 63 bits is enough. Eventually we'll switch to hipBLAS and switch from `size_t` to `hipblas_stride`.

atamazov:
@CAHEK7 @JehandadKhan @DrizztDoUrden @junliume  Ping. Please start CI.

JehandadKhan:
@atamazov  Done

atamazov:
@JehandadKhan ⚠️ This impacts persistent binary kernel cache (kernel db). Would you like me to do something (e.g. rename the kernels) to avoid selection and using of the old kernels (from the outdated binary cache), which would lead to run-time errors?

BTW It may worth adding the https://github.com/ROCm/MIOpen/labels/IMPACTS_KERNEL_DB to the list of labels?

/cc @junliume

JehandadKhan:
> @JehandadKhan ⚠️ This impacts persistent binary kernel cache (kernel db). Would you like me to do something (e.g. rename the kernels) to avoid selection and using of the old kernels (from the outdated binary cache), which would lead to run-time errors?
> 
> BTW It may worth adding the https://github.com/ROCm/MIOpen/labels/IMPACTS_KERNEL_DB to the list of labels?
> 
> /cc @junliume

Lets rename the kernel to avoid any issue, since that addresses the concerns you have.

atamazov:
@JehandadKhan @junliume Decided to rename kernels (see my previous comment).

JehandadKhan (approved):

atamazov:
@junliume 
> @[junliume](https://github.com/junliume) junliume marked this pull request as ready for review

🚀  now I have to create another PR to implement https://github.com/ROCm/MIOpen/pull/2613#issuecomment-1864781888

PR #2560, opened 2023-11-28
Bug fix for 3D Group Conv Backward data and weight update. Failure noticed when pads and strides are not 1

amberhassaan:
3D Group Convolution Backward datd and weights update solvers had a bug. Initially it was thought the bug was in CK but in reality it was in MIOpen. 
The bug stems from a design decision in MIOpen that swaps input and output tensors in backward passes. This is very confusing and bug-prone and needs to be fixed at large ASAP.

PR #2619, opened 2023-12-18
Proper fix for backward passes bwd/wrw for CK group conv 3d

amberhassaan:
The fix I provided in #2560 was not the correct fix, which means the code is still buggy. This PR provides the correct fix. 

I realized while debugging #2429 that our gtests for backward data and weight updates for Group Convolution 3D (CK-based) have a problem. The problem being that the `ProblemDescription`  was not being set up correctly. 

I have fixed the `ProblemDescription` set up in the gtests and added a swap for `in` and `out` strides. Will verify these changes by running the solver directly through the driver. gtests are passing.


junliume (approved):

junliume:
@atamazov @JehandadKhan need your review if we want to move this one quickly before staging.

src/solver/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
```diff
@@ -84,14 +84,15 @@ struct CKArgs
...
-        input  = {G, N, K, Do, Ho, Wo};
+        input  = {G, N, C, Di, Hi, Wi};
+        output = {G, N, K, Do, Ho, Wo};
         weight = {G, K, C, Z, Y, X};
 
         // miopen strides to CK strides
-        auto miopen_in_strides  = problem.GetIn().GetStrides();
-        auto miopen_out_strides = problem.GetOut().GetStrides();
+        // On a backward pass, problem.GetIn() means y(or out),
+        // and problem.GetOut means x(or in)
```
atamazov, line 93: [Notice] For backward pass, x (or dx) tensor contains _output_ data (non-const buffer), y (dy) tensor contains input data (the buffer allowed to be const).

test/gtest/group_conv3d_bwd.cpp
```diff
@@ -54,8 +54,9 @@ void SolverBwd(const miopen::TensorDescriptor& inputDesc,
     const auto tensors =
         miopen::ConvBwdTensors{outputDesc, output, wDesc, weight, inputDesc, input};
 
+    // order for bwd data pass is dy, w, dx
     const auto problem = miopen::conv::ProblemDescription{
-        inputDesc, wDesc, outputDesc, convDesc, miopen::conv::Direction::BackwardData};
+        outputDesc, wDesc, inputDesc, convDesc, miopen::conv::Direction::BackwardData};
```
atamazov, line 59: [Notice] The test (incorrectly) uses notions of "input" and "output", which is not good for convolutions. This is legacy that we should get rid of in the future, I guess.

atamazov (approved):
🟢 LGTM!

src/solver/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
```diff
@@ -123,6 +124,29 @@ struct CKArgs
...
+        };
+#define PRINT_VEC(x) print_vec(#x, x);
+
+        PRINT_VEC(output);
+        PRINT_VEC(out_strides);
+        PRINT_VEC(input);
+        PRINT_VEC(in_strides);
+        PRINT_VEC(weight);
+        PRINT_VEC(wei_strides);
+#endif
```
JehandadKhan, line 149: So far we have never had such code in MIOpen, if we intend on keeping this, we should move the macro to the proper header file. and add it as a higher log level.

src/solver/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
```diff
@@ -120,6 +121,28 @@ struct CKArgs
...
+        };
+#define PRINT_VEC(x) print_vec(#x, x);
+
+        PRINT_VEC(output);
+        PRINT_VEC(out_strides);
+        PRINT_VEC(input);
+        PRINT_VEC(in_strides);
+        PRINT_VEC(weight);
+        PRINT_VEC(wei_strides);
+#endif
```
JehandadKhan, line 145: [same as above]

test/gtest/group_conv3d_bwd.cpp
JehandadKhan: Lets make that the "near future"

JehandadKhan (approved):

PR #2622, opened 2023-12-20
[Jenkins] Add NOMLIR stage. [Workaround] Limit usage of gfx908 nodes in non-nightly builds

junliume:


Jenkinsfile
```diff
@@ -711,6 +712,23 @@ pipeline {
                         buildHipClangJobAndReboot( build_type: 'debug', setup_flags: NOCOMGR_flags, build_cmd: NOCOMGR_build_cmd, test_flags: ' --verbose ')
                     }
                 }
+                stage('Fp32 Hip Debug NOMLIR AnyGPU') {
```
atamazov, line 715: [R] Please update
https://github.com/ROCm/MIOpen/blob/10250d6bcdaef2767d4886b839f639938d15e9e5/Jenkinsfile#L407-L409
MLIR -> NOMLIR

Jenkinsfile
```diff
@@ -466,7 +466,7 @@ pipeline {
             description: "")
         booleanParam(
             name: "TARGET_GFX908",
-            defaultValue: true,
+            defaultValue: env.BRANCH_NAME == env.NIGHTLY_BRANCH ? true : false,
```
atamazov, line 469: [Notice] AnyGPU stages ignore this. AnyGPU stages of non-nightly builds will occupy gfx908 nodes.

[R] Mention that this is a W/A for the MI100 node shortage problem.

atamazov (approved):
🟢 LGTM!

[R] Maybe it is worth renaming this PR to "[Jenkins] Add NOMLIR stage. [Workaround] Limit usage of gfx908 nodes in non-nightly builds".

Jenkinsfile
junliume: @atamazov yes, and it is left on purpose. We want all stages not to be restricted to gfx908, but sometimes anyGPU stages may occupy (but not limited by) gfx908 nodes and if issues found it is still legit.

JehandadKhan (approved):

Issue #2624, opened 2023-12-20
[bug] asm igemm fwd kernel will have computation error when c <=4 and dilation_y > 1


carlushuang:
### Problem Description

asm igemm kernel will have computation error for `c` <=4 and `dilation_y` != 1 case. Both fp16/fp32 NCHW/NHWC have this problem. Affect solver `ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC`
e.g.
```
MIOpenDriver conv -n 2 -c 1 -H 22 -W 22 -k 1 -y 4 -x 4 -p 1 -q 2 -u 4 -v 4 -l 3 -j 2 -m conv -g 1 -F 1 -t 1
#####
PRNG seed: 12345678
MIOpen Forward Conv. Algorithm: 5, Solution: 107/ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC
GPU Kernel Time Forward Conv. Elapsed: 0.014151 ms (average)
stats: name, n, c, ho, wo, x, y, k, flopCnt, bytesRead, bytesWritten, GFLOPs, GB/s, timeMs
stats: fwd-conv4x4u4, 2, 1, 4, 5, 4, 4, 1,  1280, 3936, 160, 0, 0, 0.014151
Forward Convolution FAILED: 0.249294 > 1.5e-05
```

### Operating System

Ubuntu 22.04

### CPU

Ryzen 5950X

### GPU

AMD Instinct MI210

### ROCm Version

ROCm 5.7.1

PR #2632, opened 2023-12-21
[Tests] add unit test for #2624

carlushuang:
add regression unit test for #2624


junliume (approved):

CAHEK7:
It's supposed to be a gtest.

junliume:
> It's supposed to be a gtest.

@carlushuang  @CAHEK7  Yes, I missed that. Indeed all new tests are supposed to be in gtests.
Can it be part of https://github.com/ROCm/MIOpen/pull/2553 ? CC: @xinlipn

Issue #2647, opened 2024-01-01
[CI] need to add "install deb" stage


junliume:
We should be able to add it to an existing stage.
 
             This issue is introduced by #2508 , not detected by CI because we do not have an "install deb" stage.

_Originally posted by @junliume in https://github.com/ROCm/MIOpen/pull/2645#discussion_r1438979806_

PR #2648, opened 2024-01-01
[CI] add deb install to HIP package build stage

junliume:
#2647


junliume:
mmm ... not as easy as I thought it would be ;-(
```
sudo: you do not exist in the passwd database
```

PR #2650, opened 2024-01-02
[HotFix] Adapt to changes to HIPRTC in current Staging/Mainline (future 6.1 RC) that disallow the use of <limits> in kernels

junliume:
as commented below, with this PR, one has to use
> add -DMIOPEN_OVERRIDE_HIP_VERSION_MINOR=1 -DMIOPEN_OVERRIDE_HIP_VERSION_PATCH=24000 

to CMake options

Workaround lifecycle is tracked by #2651

### DIFF
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -60,6 +60,12 @@
 /// With base driver 5.11.32 the errors disappear.
 /// More info at https://github.com/ROCm/MIOpen/issues/1257.
 #define WORKAROUND_ISSUE_1257 (HIP_PACKAGE_VERSION_FLAT >= 4003021331ULL)
+/// For now, use only standard <limits> to avoid possibility of
+/// correctnes or performance regressions.
+/// \todo Test and enable "custom" local implementation.
+/// Hence the workaround is applicable for a range of hipRTC versions 
+/// and potentially in the future
+#define WORKAROUND_HIPRTC_LIMITS (HIP_PACKAGE_VERSION_FLAT >= 6000023464ULL && HIP_PACKAGE_VERSION_FLAT < 6000023494ULL)
 
 MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_COMGR_LOG_CALLS)
 MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_COMGR_LOG_SOURCE_NAMES)
@@ -1302,10 +1308,9 @@ void BuildHip(const std::string& name,
 #endif
         opts.push_back("-DHIP_PACKAGE_VERSION_FLAT=" + std::to_string(HIP_PACKAGE_VERSION_FLAT));
         opts.push_back("-DMIOPEN_DONT_USE_HIP_RUNTIME_HEADERS");
-        /// For now, use only standard <limits> to avoid possibility of
-        /// correctnes or performance regressions.
-        /// \todo Test and enable "custom" local implementation.
+#if !WORKAROUND_HIPRTC_LIMITS
         opts.push_back("-DWORKAROUND_DONT_USE_CUSTOM_LIMITS=1");
+#endif
 #if WORKAROUND_ISSUE_1431
         if((StartsWith(target.Name(), "gfx10") || StartsWith(target.Name(), "gfx11")) &&
            !miopen::comgr::IsWave64Enforced(opts))
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -65,8 +65,7 @@
 /// \todo Test and enable "custom" local implementation.
 /// Hence the workaround is applicable for a range of hipRTC versions
 /// and potentially in the future
-#define WORKAROUND_HIPRTC_LIMITS \
-    (HIP_PACKAGE_VERSION_FLAT >= 6000023464ULL && HIP_PACKAGE_VERSION_FLAT < 6000023494ULL)
+#define WORKAROUND_DONT_USE_CUSTOM_LIMITS (HIP_PACKAGE_VERSION_FLAT < 6001024000ULL) 
 
 MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_COMGR_LOG_CALLS)
 MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_COMGR_LOG_SOURCE_NAMES)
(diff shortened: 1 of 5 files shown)


src/comgr.cpp
```diff
@@ -60,6 +60,13 @@
 /// With base driver 5.11.32 the errors disappear.
 /// More info at https://github.com/ROCm/MIOpen/issues/1257.
 #define WORKAROUND_ISSUE_1257 (HIP_PACKAGE_VERSION_FLAT >= 4003021331ULL)
+/// For now, use only standard <limits> to avoid possibility of
+/// correctnes or performance regressions.
+/// \todo Test and enable "custom" local implementation.
+/// Hence the workaround is applicable for a range of hipRTC versions
+/// and potentially in the future
+#define WORKAROUND_HIPRTC_LIMITS \
+    (HIP_PACKAGE_VERSION_FLAT >= 6000023464ULL && HIP_PACKAGE_VERSION_FLAT < 6000023494ULL)
```
junliume: For reviewers (@atamazov), these way points are the best we know at the moment, they may not be exactly precise but should work with official releases.

atamazov: The HIP version in Staging is usually incorrect. Therefore, we should not rely on these numbers in the code (otherwise we will have more and more of mess in the source code). It's much better to replace the incorrect HIP version numbers with the _correct expected_ ones (in this case 6.1.24xxx) and use the correct version.

atamazov:
```suggestion
#define WORKAROUND_DONT_USE_CUSTOM_LIMITS (HIP_PACKAGE_VERSION_FLAT < 6001024000ULL)
```
Why: https://github.com/ROCm/MIOpen/pull/2650#discussion_r1439907789

src/comgr.cpp
```diff
@@ -1302,10 +1309,9 @@ void BuildHip(const std::string& name,
 #endif
         opts.push_back("-DHIP_PACKAGE_VERSION_FLAT=" + std::to_string(HIP_PACKAGE_VERSION_FLAT));
         opts.push_back("-DMIOPEN_DONT_USE_HIP_RUNTIME_HEADERS");
-        /// For now, use only standard <limits> to avoid possibility of
-        /// correctnes or performance regressions.
-        /// \todo Test and enable "custom" local implementation.
+#if !WORKAROUND_HIPRTC_LIMITS
```
atamazov:
```suggestion
#if WORKAROUND_DONT_USE_CUSTOM_LIMITS
```

src/comgr.cpp
junliume: @atamazov here is a summary of compiler versions and reported in the cmake output:

- [Docker A] ROCm 6.0: 
  - HIP version: 6.0.32830-d62f6a171
  - roc-6.0.0 23483 7208e8d15fbf218deb74483ea8c549c67ca4985e
  -  -- Build with HIP 6.0.23494
 
- [Docker B] The docker with all proposed hipRTC changes:
  - HIP version: 6.0.32830-dc7635e0d
  -  roc-6.0.0 23483 7208e8d15fbf218deb74483ea8c549c67ca4985e
  - -- Build with HIP 6.0.23494
  
- [Docker C] The latest mainline staging docker:
  - HIP version: 6.0.32830-606beddb1 
  - amd-mainline-open 23512 8f0cb360b515076885176a3e5ba09dab6ecefa20
  - -- Build with HIP 6.0.23464

There are a few critical issues:
- From Docker A and Docker B: `hipRTC` versioning is now decoupled from `HIP` package, so our code path control needs to accomodate.
- From Docker C: the reported HIP version is incorrect

atamazov (changes requested):
Why: https://github.com/ROCm/MIOpen/pull/2650#discussion_r1439907789

⚠️ In order to use the library with current staging/mainline compiler, the users have to add `-DMIOPEN_OVERRIDE_HIP_VERSION_MINOR=1 -DMIOPEN_OVERRIDE_HIP_VERSION_PATCH=24000` to CMake options. Please inform QA about this.

src/comgr.cpp
atamazov: Why: Let's make the macro names and semantics the same in host code and device code. Easier to maintain.

atamazov:
@junliume Maybe let's rename PR to "_[HotFix] Adapt to changes to HIPRTC in current Staging/Mainline (future 6.1 RC) that disallow the use of \<limits\> in kernels_"?

src/kernels/miopen_type_traits.hpp
```diff
@@ -76,7 +76,7 @@ struct remove_cv
     typedef typename remove_volatile<typename remove_const<T>::type>::type type;
 };
 
-#if HIP_PACKAGE_VERSION_FLAT >= 6001000000ULL
+#if HIP_PACKAGE_VERSION_FLAT >= 6001024000ULL
```
junliume: @atamazov now with the <limits> header issue resolved, we found more issues with
```
/tmp/comgr-4e3a20/include/miopen_type_traits.hpp:81:8: error: redefinition of 'integral_constant'
   81 | struct integral_constant
      |        ^
/long_pathname_so_that_rpms_can_package_the_debug_info/src/external/clr/hipamd/include/hip/amd_detail/amd_hip_vector_types.h:55:38: note: previous definition is here
   55 | template <class _Tp, _Tp __v> struct integral_constant {
      |                                      ^

```
I guess we have to identify another point where this workaround is needed, and *not the same point* with the limits header.

junliume: We can make it
> #if HIP_PACKAGE_VERSION_FLAT > 6001024000ULL

but this is very fragile ...

src/kernels/miopen_cstdint.hpp
```diff
@@ -30,7 +30,7 @@ typedef signed char int8_t;
 typedef unsigned char uint8_t;
 typedef signed short int16_t;
 typedef unsigned short uint16_t;
-#if HIP_PACKAGE_VERSION_FLAT >= 6001000000ULL
+#if HIP_PACKAGE_VERSION_FLAT >= 6001024000ULL
```
junliume: Similar to https://github.com/ROCm/MIOpen/pull/2650/files#r1439946058

I am afraid this is fragile and might or might not break ...

src/kernels/miopen_limits.hpp
```diff
@@ -41,6 +41,15 @@ namespace std {
 template <typename T>
 class numeric_limits;
 
+template <>
```
junliume, line 44: @atamazov could you take a look? `test_reduce_test` seems to be needing this in the docker where I am testing. Also I may wonder if there could be other data types missing?

src/kernels/miopen_type_traits.hpp
```diff
@@ -26,6 +26,9 @@
 #pragma once
 
 #ifdef MIOPEN_DONT_USE_HIP_RUNTIME_HEADERS
+/// \todo: exact version from which this WA is needed should be updated in release
+/// https://github.com/ROCm/MIOpen/issues/2651
+#define WORKAROUND_USE_MORE_CUSTOM_TYPE_TRAITS (HIP_PACKAGE_VERSION_FLAT > 6001024000ULL)
```
junliume, line 31: @atamazov reviewers be noted that this line **excludes** the assumed (fake) version.

junliume:
replace with #2652

PR #2653, opened 2024-01-04
    correct parameter which prints unused flag in log fusion cmd

jasberc:
Original parameter "false" would be casted to a 0 value std::optional and make unused flag "-S" be printed in fusion cmd logs.


bghimireamd:
LGTM

bghimireamd (approved):

junliume (approved):

PR #2655, opened 2024-01-05
clean up the setting of environment variables cross-platform

apwojcik:
The current MIOpen API for setting environment variables is failing compilation on Windows. This PR is fixing the issue and enhancing the API to propose an OS-agnostic method of setting/removing environment variables. Internally, the API uses POSIX API on Linux and WIN32 API on Windows.

### DIFF
--- a/test/env_utils.hpp
+++ b/test/env_utils.hpp
@@ -1,0 +1,60 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2017 Advanced Micro Devices, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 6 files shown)


DrizztDoUrden (approved):

CAHEK7 (approved):
This particular fix is fine, but unfortunately it fixes the code which should not be written at all.

A short description why:
Recently we've introduced direct access to the cached values of our internal env vars, so there is no need to use environment from the inside of the tests.
Either environment variables should be set from the outside (like `MIOPEN_LOG_LEVEL=7 ./bin/test_conv`) or accessed directly from the inside (like) `miopen::UpdateEnvVar(ENV(MIOPEN_LOG_LEVEL),"7");`.

I'm not requesting to do this change, but if you have some spare time..

atamazov:
@apwojcik Thanks for catching the issue and for providing very good fix, but @CAHEK7 is right. We are going to replace putenv/setenv stuff in tests with direct modification of internally cached values. However it is fine to merge this to fix build errors on Windows.

junliume (approved):

PR #2657, opened 2024-01-05
addkernels: fix operations on path for Windows

apwojcik:
Using `std::string` for manipulating paths is error-prone - it produces paths that do not work on Windows because MIOpen uses the Linux-specific and not portable approach. Moving to `std::filesystem::path` will make it OS-agnostic. Furthermore, we deliberately use `std::filesystem::path` and not `boost::filesystem::path` because the MSVC standard library is incompatible with Boost and requires extra conversion/translation between Boost and Std. The example is when `boost::filesystem::path` is used with `std::ofstream` - GNU has additional construction accepting Boost path, MSVC does not, and we need to convert each time to either Std path or `std::string`.
At the same time, MIOpen is using C++17 standard, and `std::filesystem` has most of the functionality Boost offers.
Later, we will submit PRs for other components in MIOpen, changing `std::string` and `boost::filesystem::path` to `std::filesystem::path`. Currently, the Linux-specific path operations block Windows MIOpen from executing specific scenarios.

### DIFF
--- a/addkernels/addkernels.cpp
+++ b/addkernels/addkernels.cpp
@@ -25,6 +25,7 @@
  *******************************************************************************/
 #include "include_inliner.hpp"
 #include <algorithm>
+#include <filesystem>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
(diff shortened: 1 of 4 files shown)


DrizztDoUrden (approved):
LGTM

JehandadKhan (approved):

CAHEK7 (approved):

junliume (approved):

PR #2652, opened 2024-01-04
[HOTFIX] Adapt to changes in HIP Mainline 417 (possibly future 6.1 RC)

atamazov:
This PR may be considered as a continuation of #2637. We need it due to recent breaking changes in HIP mainline.

Supersedes #2650.

### In short
- Fixes build failures with HIP Mainline 417
- ⚠️ Currently, HIP Mainline builds report incorrect HIP versions.
  - Therefore, in order to properly build the library for the HIP Mainline, the user must override the HIP version in the CMake command line.
  - For Mainline 417 and similar, use `-DMIOPEN_OVERRIDE_HIP_VERSION_MINOR=1 -DMIOPEN_OVERRIDE_HIP_VERSION_PATCH=24000`.
  - For older Mainline builds, `-DMIOPEN_OVERRIDE_HIP_VERSION_MINOR=1 -DMIOPEN_OVERRIDE_HIP_VERSION_PATCH=0` may work better.

### Some details
- HIP Mainline 417 specific fixes
  - Disabled some new OCL & HIP kernel build warnings.
  - [tests] Disabled a testcase from handle_test as [#2600](https://github.com/ROCm/MIOpen/issues/2600) still persists.
    - The testcase will be automatically enabled after week 4.
  - [hipRTC] Used custom implementations instead of standard `<limits>`.
  - [hipRTC] Disabled some of the custom implementations from `<type_traits>` (like `integral_constant` etc).
  - [clang++] Removed "-mcpu" from HIP build options. 
- By-products
  - Fixed issue in `"miopen_limits.h"` that prevented the use of custom implementations.
  - Improved diagnostic messages output onto console after offline build failures.

## :cyclone: Testing

Preconditions:
- Navi21 (gfx1030/36), Ubuntu 20.04.5 LTS
- `-DMIOPEN_TEST_ALL=Off -DBUILD_DEV=On` (Smoke testing)
- `-DMIOPEN_OVERRIDE_HIP_VERSION_MINOR=1 -DMIOPEN_OVERRIDE_HIP_VERSION_PATCH=24000`
- Both online (`-DMIOPEN_USE_COMGR=On -DMIOPEN_USE_HIPRTC=On`) and offline (`-DMIOPEN_USE_COMGR=Off -DMIOPEN_USE_HIPRTC=Off`) kernel compilers were tested.
- ROCm version, docker image, build type, compiler, datatypes tested and results of testing - see table below

  |   |   |   | Results |   |   |   |
[...]

PR #2644, opened 2023-12-28
[ROCm 6.0.1] Automatically activate the new HIPRTC PCH adaptations starting from the 6.0.24000 version. Fix some build errors.

atamazov:
_Cherry picked from commit 4f695d975a2a6de2f167fc2925f3bad79fbaaf98 (see PR #2641)._

Expected to maintain compatibility with ROCm 6.0.1 (which is important for Staging/Mainline testing and future releases)

## :cyclone: Testing

- Navi21 (gfx1030/36), Ubuntu 20.04.5 LTS
- Docker images:
  - `rocm/miopen:hiprtc` (6.1 RC) with `-DMIOPEN_OVERRIDE_HIP_VERSION_PATCH=2400`
  - `rocm/miopen:ci_d868bf` (6.0)
- `-DMIOPEN_TEST_ALL=Off -DBUILD_DEV=On` (Smoke testing)
- Release builds, offline compiler.
- Tested datatypes: FLOAT, HALF, BFLOAT16, INT8

🟢 NO FAILURES

PR #2658, opened 2024-01-05
[ROCm 6.0.1] Fix merge error in #2652 that affects #2644.

atamazov:
Fixes merge error in #2652 that affects the functionality added by #2644. Resolves https://github.com/ROCm/MIOpen/pull/2652#discussion_r1442577220.

----
[Attribution] @junliume
- https://github.com/ROCm/MIOpen/labels/urgency_normal
- Proposed reviewers: @junliume

### DIFF
--- a/src/kernels/miopen_type_traits.hpp
+++ b/src/kernels/miopen_type_traits.hpp
@@ -76,7 +76,7 @@ struct remove_cv
     typedef typename remove_volatile<typename remove_const<T>::type>::type type;
 };
 
-#if HIP_PACKAGE_VERSION_FLAT >= 6001000000ULL && HIP_PACKAGE_VERSION_FLAT < 6001024000ULL
+#if HIP_PACKAGE_VERSION_FLAT >= 6000024000ULL && HIP_PACKAGE_VERSION_FLAT < 6001024000ULL
 template <class T, T v>
 struct integral_constant
 {


junliume (approved):

src/kernels/miopen_type_traits.hpp
```diff
@@ -76,7 +76,7 @@ struct remove_cv
     typedef typename remove_volatile<typename remove_const<T>::type>::type type;
 };
 
-#if HIP_PACKAGE_VERSION_FLAT >= 6001000000ULL && HIP_PACKAGE_VERSION_FLAT < 6001024000ULL
+#if HIP_PACKAGE_VERSION_FLAT >= 6000024000ULL && HIP_PACKAGE_VERSION_FLAT < 6001024000ULL
```
junliume, line 79: @atamazov thanks for catching it! However, now with minor version fixed, `HIP_PACKAGE_VERSION_FLAT < 6001024000ULL` is the only place where we rely on a patch version. Is there a way to get rid of it?

https://github.com/ROCm/MIOpen/issues/2651#issuecomment-1879096776

junliume: It might *just work* because the latest mainline is reporting `Build with HIP 6.1.24013`, just hope that patch version number does not go backwards ;-)

atamazov: @junliume 
> Is there a way to get rid of it?

No, this adapts the library to differences between 6.0, 6.0.1 and 6.1.

>  just hope that patch version number does not go backwards ;-)

If it goes, we override.

PR #2660, opened 2024-01-08
[Fin] Update Submodule

cderb:
Use updated LoadProgram(), SaveBinary() and LoadBinary()
simplify docke.rfile recipe to use miopen ci baseimage
Adding mi300 to arch verification function.


JehandadKhan:
@cderb  Can you please update the description with what changes are being brought in?

junliume (approved):

PR #2663, opened 2024-01-10
Add 2D Group Convolution Backward Data and Weights update solvers. Simplify and unify 3d group conv tests

amberhassaan:
Follow up to #2536 where I use the generic test fixture class (and macros) for 3D group convolution gtests. 

Copy of description from #2536: 
This PR provides CK-based Backward-Data and Backward-Weight update solvers for 2D Group Convolution. This completes the solver picture for 2D group convolutions (except for f8, which has its own solvers).

Also included in this PR is a generic design for the GTest Fixture class that handles 2D and 3D (Fwd,Bwd,Wrw) passes at the same time, so testing code can reuse this class and related macros. A future PR will use this class for 3D grouped conv gtests (arriving shortly).


junliume (approved):

src/solver.cpp
```diff
@@ -607,6 +607,14 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
...
     Register(registry, ++id, Primitive::Batchnorm, batchnorm::BnCKBwdBackward{}.SolverDbId());
     Register(registry, ++id, Primitive::Batchnorm, batchnorm::BnCKFwdTraining{}.SolverDbId());
+    RegisterWithSolver(registry,
+                       ++id,
+                       conv::ConvHipImplicitGemmGroupBwdXdlops{},
+                       miopenConvolutionAlgoImplicitGEMM);
+    RegisterWithSolver(registry,
+                       ++id,
+                       conv::ConvHipImplicitGemmGroupWrwXdlops{},
+                       miopenConvolutionAlgoImplicitGEMM);
```
CAHEK7: The order of ids must not be altered for backward compatibility.
Please move it to the end of the list.

src/include/miopen/handle.hpp
```diff
@@ -211,6 +211,15 @@ struct MIOPEN_EXPORT Handle : miopenHandle
         return result;
     }
 
+    template <class V>
+    void ReadToVec(const Allocator::ManageDataPtr& ddata, V& output_vec)
+    {
+        using T = typename V::value_type;
+        assert(ddata);
+        assert(output_vec.size() > 0);
+        this->ReadTo(output_vec.data(), ddata, output_vec.size() * sizeof(T));
```
CAHEK7, line 208: Should we also check that `ddata` has at least `output_vec.size() * sizeof(T)` bytes allocated?

CAHEK7 (changes requested):
Minor `ReadToVec` comment is here.
But the most important part is to keep all old solver ids unaltered.

src/solver.cpp
amberhassaan: Will do.

src/include/miopen/handle.hpp
amberhassaan: Yes, I wanted that but today we don't store size of the buffer with the ManagedPtr, so I plan on doing this in the near future.

CAHEK7 (approved):

src/solver/conv_hip_implicit_gemm_grouped_bwd_xdlops.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    if(miopen::IsDisabled(ENV(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_GROUP_BWD_XDLOPS)))
+        return false;
+    if(miopen::IsEnabled(ENV(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC)))
+        return false;
+    if(problem.GetInDataType() != problem.GetWeightsDataType() ||
+       problem.GetWeightsDataType() != problem.GetOutDataType() ||
+       problem.GetInDataType() != problem.GetOutDataType())
+        return false;
```
averinevg, line 308:
```suggestion
        return false;
    if(problem.HasNonPackedTensors())
        return false;
```

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
```diff
@@ -0,0 +1,361 @@
...
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    if(miopen::IsDisabled(ENV(MIOPEN_DEBUG_GROUP_CONV_IMPLICIT_GEMM_HIP_WRW_XDLOPS)))
+        return false;
+    if(miopen::IsEnabled(ENV(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC)))
+        return false;
+    if(problem.GetInDataType() != problem.GetWeightsDataType() ||
+       problem.GetWeightsDataType() != problem.GetOutDataType() ||
+       problem.GetInDataType() != problem.GetOutDataType())
+        return false;
```
averinevg, line 304:
```suggestion
        return false;
    if(problem.HasNonPackedTensors())
        return false;
```

src/solver/conv_hip_implicit_gemm_grouped_bwd_xdlops.cpp
```diff
@@ -0,0 +1,367 @@
...
+    [[maybe_unused]] const ProblemDescription& problem) const
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    if(miopen::IsDisabled(ENV(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_GROUP_BWD_XDLOPS)))
+        return false;
+    if(miopen::IsEnabled(ENV(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC)))
+        return false;
+    if(problem.GetInDataType() != problem.GetWeightsDataType() ||
+       problem.GetWeightsDataType() != problem.GetOutDataType() ||
+       problem.GetInDataType() != problem.GetOutDataType())
```
averinevg:
```suggestion
    if(problem.HasMixedDataTypes())
```

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
```diff
@@ -0,0 +1,361 @@
...
+    [[maybe_unused]] const ProblemDescription& problem) const
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    if(miopen::IsDisabled(ENV(MIOPEN_DEBUG_GROUP_CONV_IMPLICIT_GEMM_HIP_WRW_XDLOPS)))
+        return false;
+    if(miopen::IsEnabled(ENV(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC)))
+        return false;
+    if(problem.GetInDataType() != problem.GetWeightsDataType() ||
+       problem.GetWeightsDataType() != problem.GetOutDataType() ||
+       problem.GetInDataType() != problem.GetOutDataType())
```
averinevg:
```suggestion
    if(problem.HasMixedDataTypes())
```

averinevg (changes requested):

amberhassaan:
@averinevg : Please approve the PR. All changes applied.

averinevg (approved):

iq136boy (approved):

PR #2669, opened 2024-01-12
[BF16][FP8][BF8] Fixed some specializations from `<limits>` and `<cmath>`

atamazov:
- For BF16
  - Fixed `numeric_limits<T>::min()`, `denorm_min()`, `quiet_NaN()`, `signaling_NaN()`
- For FP8, BF8
  - Fixed `numeric_limits<T>::min()`, `max()`, `epsilon()`, `quiet_NaN()`
  - Added `numeric_limits<T>::denorm_min()`, , `signaling_NaN()`
  - Fixed `fabs()` for non-IEEE mode
  - Fixed `isfinite()`. This resolves:
    - https://github.com/ROCm/MIOpen/pull/2584#discussion_r1419714785
    - https://github.com/ROCm/MIOpen/pull/2584#discussion_r1419715130

## ⚠️ Other action points

@xinlipn @junliume I think this should be discussed with HIP compiler/library developers who provided the initial version of https://github.com/ROCm/MIOpen/blob/e125c5316516c452be406301f5da9f0f84017ef0/src/kernels/hip_float8.h. Ours and theirs versions must be in sync _and_ of course we do not want the bugs fixed here just to re-appear from the HIP compiler later. Please either discuss the fixes with them or or refer them to me to talk. This AP is https://github.com/ROCm/MIOpen/labels/value_high

----
- https://github.com/ROCm/MIOpen/labels/urgency_high - as I am not aware of any related issues, otherwise it would be blocker.
- https://github.com/ROCm/MIOpen/labels/value_unknown
- Proposed reviewers:
  - @TejashShah for BF16 fixes
  - @xinlipn for F8 fixes
  - @JehandadKhan for general awareness
  - @junliume for general awareness
  - @CAHEK7  for general awareness

### DIFF
--- a/src/kernels/hip_float8.hpp
+++ b/src/kernels/hip_float8.hpp
@@ -507,8 +507,8 @@ public:
 
     static MIOPEN_HIP_HOST_DEVICE miopen_f8::hip_f8<miopen_f8::hip_f8_type::fp8> quiet_NaN()
     {
-        return static_cast<miopen_f8::hip_f8<miopen_f8::hip_f8_type::fp8>>(
-            static_cast<uint8_t>(miopen_f8::get_hip_f8_bias_mode() ? 0X80 : 0x79));
+        return miopen_f8::Generate<miopen_f8::hip_f8<miopen_f8::hip_f8_type::fp8>>(
+            MIOPEN_FP8_IEEE_EXPONENT_BIAS ? 0x7c : 0x80);
     }
 
     static MIOPEN_HIP_HOST_DEVICE miopen_f8::hip_f8<miopen_f8::hip_f8_type::fp8> max()
@@ -536,8 +536,8 @@ public:
 
     static MIOPEN_HIP_HOST_DEVICE miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8> quiet_NaN()
     {
-        return static_cast<miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8>>(
-            static_cast<uint8_t>(miopen_f8::get_hip_f8_bias_mode() ? 0X80 : 0x7d));
+        return miopen_f8::Generate<miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8>>(
+            MIOPEN_FP8_IEEE_EXPONENT_BIAS ? 0x7e : 0x80);
     }
 
     static MIOPEN_HIP_HOST_DEVICE miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8> max()
--- a/src/kernels/hip_float8.hpp
+++ b/src/kernels/hip_float8.hpp
@@ -542,9 +542,8 @@ public:
 
     static MIOPEN_HIP_HOST_DEVICE miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8> max()
     {
-        return static_cast<miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8>>(
-            miopen_f8::Generate<miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8>>(
-                MIOPEN_FP8_IEEE_EXPONENT_BIAS ? 0x7b : 0x7f));
+        return miopen_f8::Generate<miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8>>(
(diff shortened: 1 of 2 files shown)


src/kernels/hip_float8.hpp
```diff
@@ -478,19 +478,26 @@ inline MIOPEN_HIP_HOST_DEVICE bool operator>(const miopen_f8::hip_f8<T>& lhs,
...
     union
     {
         uint8_t bits;
         T value;
     } x;
 
-    x.bits = 0x7F;
+    x.bits = bits;
     return x.value;
 }
```
CAHEK7, line 502: Formally speaking it is UB in `C++` (but allowed and perfectly fine in `C`).
Until C++20 `std::bitcast`, the only true way to do type a punning is `std::memcpy` or any memcpy builtin.

`std::memcpy` perfectly works with registers or any sort of reinterpretation, even if it looks like a memory copy.
https://godbolt.org/z/TPPfqT9s4

CAHEK7 (approved):

src/kernels/hip_float8.hpp
CAHEK7: Thought it works.

atamazov: @CAHEK7 Behavior of this is unspecified in the Standard. But AFAICS this code came from the HIP library, and proven to work for us (we can consider this as documented). 

But this is a bit out of scope of this PR; I am only fixing the values.

junliume (approved):

src/kernels/hip_float8.hpp
CAHEK7: @atamazov 
It is not `unspecified` it is clearly `undefined` in the Standart. The difference between C and C++ is that in C++ union may have members with non-trivial constructors, destructors and so on. As I've said - it works, but still.

I know about that HIP library flaw, it's not a reference, it's a library which uses some highly-likely working trick.
Btw, nvidia has it's own builtin for that sort of casting (I've checked  two implementations of `__shlf_xor` builtins).

https://en.cppreference.com/w/cpp/language/union
I know it's not an official version of the Standard, but it's quite close to be it.

And I'm not asking to change this code right now, I'm just pointing to that common delusion.

atamazov: @CAHEK7 
> It is not unspecified it is clearly undefined

I agree. And that means that our program is ill-formed from the POV of the Standard.

However, in our implementation of the HIP compiler, this behavior is well-defined so far.

I think this should be discussed with HIP compiler/library developers, as proposed in the PR description.

junliume: @atamazov the runtime folks claim no changes were made at least recently to the above mentioned places, and the header files are located at:
https://github.com/ROCm/HIP/blob/355b16b473090b219a6a076cab902b6497ed1883/include/hip/hip_bfloat16.h
https://github.com/ROCm/clr/blob/develop/hipamd/include/hip/amd_detail/amd_hip_bfloat16.h

atamazov: @junliume Thanks. I do not see any issues in https://github.com/ROCm/clr/blob/develop/hipamd/include/hip/amd_detail/amd_hip_bfloat16.h. Although they don't define things from `<limits>` like our hip_bfloat16.h does, so maybe those things are a recent addition we made (and those issues are fixed in this PR).

But what about FP8/BF8 related things? Or header is a copy of
https://github.com/ROCm/MIOpen/blob/3a33fe7a1a2d2d30c6b0d6fe64efbca94d69ba2a/src/kernels/hip_float8.hpp#L34
and I've made a lot of fixes in `<limits>` and `<cmath>` things. Maybe these things are a recent addition to the original code, but I would like to make sure that we and compiler guys are on the same page about what F8 is.

atamazov:
@junliume Let's sync with compiler team about F8: https://github.com/ROCm/MIOpen/pull/2669#discussion_r1455106973

src/kernels/hip_float8.hpp
junliume: @JehandadKhan could you chip in on the F8 header alignment?

atamazov:
@JehandadKhan @junliume Let's sync with compiler team about F8: https://github.com/ROCm/MIOpen/pull/2669#discussion_r1455106973

PR #2671, opened 2024-01-12
Softmax ocl refactoring

Vsevolod1983:
TICKET:
https://github.com/ROCm/MIOpen/issues/2643

SoftmaxForward and SoftmaxBackward are refactored in this PR
Also, network config is changed as well, according to what we discussed with the team


DrizztDoUrden (approved):
LGTM

src/include/miopen/ctc.hpp
```diff
@@ -56,7 +56,7 @@ struct CTCLossDescriptor : miopenCTCLossDescriptor
                                    const int* inputLengths,
                                    miopenCTCLossAlgo_t algo) const;
 
-    void CTCLoss(const Handle& handle,
+    void CTCLoss(Handle& handle,
```
CAHEK7, line 59: Is it related to the softmax?

src/ocl/ctcocl.cpp
```diff
@@ -44,7 +44,7 @@
 
 namespace miopen {
 
-void CTCLossDescriptor::CTCLoss(const Handle& handle,
+void CTCLossDescriptor::CTCLoss(Handle& handle,
```
CAHEK7, line 47: Is it related to the softmax?

src/include/miopen/softmax/invoke_params.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <miopen/invoke_params.hpp>
+#include <miopen/tensor.hpp>
+
+namespace miopen {
+namespace softmax {
+
+struct InvokeParams : public miopen::InvokeParams
+{
+    InvokeParams( const void* alpha_,
+                            const void* beta_,
```
CAHEK7: clang-format

src/include/miopen/softmax/problem_description.hpp
```diff
@@ -0,0 +1,90 @@
...
+{
+    ProblemDescription( const void* alpha_,
+                        const void* beta_,
+                        const TensorDescriptor& xDesc_,
+                        const TensorDescriptor& yDesc_,
+                        miopenSoftmaxAlgorithm_t algorithm_,
+                        miopenSoftmaxMode_t mode_) :
+        alpha(alpha_), beta(beta_), xDesc(xDesc_), yDesc(yDesc_), algorithm(algorithm_), mode(mode_)
+    {
+    }
```
CAHEK7: Could you add that kind of check?
https://github.com/ROCm/MIOpen/blob/5e82b808021308040d6651e0f759d9a50702c859/src/ocl/softmaxocl.cpp#L78

And also verify `alpha_` and `beta_` pointers for `nullptr`?

CAHEK7 (changes requested):
Some minor changes are needed.

src/include/miopen/ctc.hpp
Vsevolod1983: Yes it does. CTC uses softmax and passes handle for softmax calculation. Since this handle is no longer a const for softmax we need to remove const modifier here as well. If we have other options lets discuss.

src/include/miopen/softmax/problem_description.hpp
Vsevolod1983: These checks  are present in  miopenStatus_t SoftmaxForward(...) function
But I will add additional alpha and beta checks.

atamazov: @Vsevolod1983 It seems better to check and block construction of invalid ProblemDescriptions rather than implement the same checks here and there in the code. So please,

> These checks are present in miopenStatus_t SoftmaxForward(...) function

Move these into ctor of PD. Test manually if it works properly.

src/include/miopen/ctc.hpp
atamazov: @Vsevolod1983 is it so that we need to make many other changes in the code in order to keep the handle `const`?

atamazov, line 59: @Vsevolod1983 is it so that we need to make many other changes in the code in order to keep the handle `const`? If yes, and we are not going to do this in this PR, then let's open a ticket and eventually do this in a followup PR.

/cc @CAHEK7

CAHEK7:
I guess it can be converted from draft in order to trigger CI.

src/include/miopen/softmax/problem_description.hpp
Vsevolod1983: @atamazov 
We cannot do it in all cases, because several parameters are not part of ProblemDescription. For example x and y - are data params, they are not included in ProblemDescription ctor, but we need to check them as well (and throws an exception if they are invalid).
Also, we have a case when we throw an exception only after we calculate a lot of intermediate params. So, its not a part of PD ctor as well.

src/include/miopen/ctc.hpp
Vsevolod1983: CTCLoss calls SoftmaxForward which subsequently calls SolverContainers::ExecutePrimitive which accept non const Handle&.
It leads to CTCLoss should accept non const Handle&. Otherwise we need to use const-cast hack.

atamazov: @Vsevolod1983 I see, thanks.

@DrizztDoUrden What about https://github.com/ROCm/MIOpen/pull/1276#discussion_r1455119631?

src/include/miopen/softmax/problem_description.hpp
atamazov: @Vsevolod1983 
> Also, we have a case when we throw an exception only after we calculate a lot of intermediate params. So, its not a part of PD ctor as well.

Does this case indicate invalid problem?

junliume:
> I guess it can be converted from draft in order to trigger CI.

CI started, I also requested the user names to be added to the repo access for more convenient development process.

Update @Vsevolod1983 there are some HIP tidy issues:
[hip_tidy.log](https://github.com/ROCm/MIOpen/files/13970045/hip_tidy.log)

Vsevolod1983:
I also need to remake network config as we discussed with the team. Working on it. So please do not merge for now

src/include/miopen/softmax/problem_description.hpp
Vsevolod1983: @atamazov 
Hi, I moved some checks to ctor of PD class. One check stayed on solver side though. Please have a look.

Also I changed network config, will appreciate if you have a look as well.

Vsevolod1983:
Its ready to review now

Vsevolod1983:
> > I guess it can be converted from draft in order to trigger CI.
> 
> CI started, I also requested the user names to be added to the repo access for more convenient development process.
> 
> Update @Vsevolod1983 there are some HIP tidy issues: [hip_tidy.log](https://github.com/ROCm/MIOpen/files/13970045/hip_tidy.log)

investigating, thank you
UPD: Fixed

src/include/miopen/softmax/problem_description.hpp
```diff
@@ -0,0 +1,153 @@
...
+                       miopenSoftmaxAlgorithm_t algorithm_,
+                       miopenSoftmaxMode_t mode_)
+        : isForward(true),
+          alpha(alpha_),
+          beta(beta_),
+          xdxDesc(xDesc_),
+          yDesc(yDesc_),
+
+          // initalize this reference to some value, but it will not be used for Forward
+          dyDesc(yDesc_),
```
atamazov: [Q] Maybe it is better to initialize it with an empty tensor descriptor? You can create it as a static member of the class and name "unusedDesc".

src/include/miopen/softmax/invoke_params.hpp
```diff
@@ -0,0 +1,133 @@
...
+    std::size_t GetWorkspaceSize() const { return 0; }
+    Data_t GetWorkspace() const { return nullptr; }
+
+public:
+    const void* alpha;
+    const void* beta;
+    miopenSoftmaxAlgorithm_t algorithm;
+    miopenSoftmaxMode_t mode;
+
+    // xdxDesc is used for both forward and bacward
```
atamazov:
```suggestion
    // xdxDesc is used for both forward and backward
```
typo here and below

src/include/miopen/softmax/invoke_params.hpp
```diff
@@ -0,0 +1,133 @@
...
+    const TensorDescriptor& xdxDesc;
```
atamazov: [Q] Maybe it is cleaner to follow the same pattern as used for yDesc? (xDesc, forward_x, backward_x)?

[Q] Or perhaps it is worth creating separate Problems for Forward and Backward? Would this result in improved code clarity?

This is not very important right now, but anyway...

src/solver/softmax/softmax.cpp
```diff
@@ -0,0 +1,340 @@
...
+              vgd,
+              workgroups,
+              batch_size,
+              u_batch_size);
+
+    if(num_batch > 1)
+    {
+        if(isForward)
+        {
+            if((u_batch_size + 1) * 256 > 65536 && yDesc.GetType() == miopenHalf)
```
atamazov, line 189: [Quality] Magic numbers. It this is a legacy stuff, you can simply mark them with `/// \todo Magic numbers` (but of course it is better to fully resolve this review comment).

src/solver/softmax/softmax.cpp
```diff
@@ -0,0 +1,340 @@
...
+                MIOPEN_THROW(miopenStatusBadParm, "Exceed local memory capacity");
+        }
+        else
+        {
+            if((2 * u_batch_size + 1) * 256 > 65536 && yDesc.GetType() == miopenHalf)
```
atamazov, line 195: Ditto

src/solver/softmax/softmax.cpp
```diff
@@ -0,0 +1,340 @@
...
+    auto result = ConvSolution{miopenStatusSuccess};
+
+    auto xDesc = problem.GetXDesc();
+    auto yDesc = problem.GetYDesc();
+
+    auto dxDesc = problem.GetdXDesc();
+    auto dyDesc = problem.GetdYDesc();
+
+    auto alpha     = problem.GetAlpha();
+    auto beta      = problem.GetBeta();
```
atamazov, line 147: [Q] Maybe it is better to make GetAlpha/Beta() return floating-point numbers?

src/solver/softmax/softmax.cpp
```diff
@@ -0,0 +1,340 @@
...
+               int& out_u_batch_size)
+{
+    std::tie(out_n, out_c, out_h, out_w) = tien<4>(in_desc.GetLengths());
+
+    // using workgroup size of 256 by default
+    out_grid_size   = in_mode == MIOPEN_SOFTMAX_MODE_INSTANCE ? out_n : out_n * out_h * out_w;
+    out_spatial_dim = in_mode == MIOPEN_SOFTMAX_MODE_INSTANCE ? 1 : out_h * out_w;
+    out_vector_size = in_mode == MIOPEN_SOFTMAX_MODE_INSTANCE ? out_c * out_h * out_w : out_c;
+    // num_spatial_dims or pixels each workgroup can compute
+    out_num_batch = out_vector_size < 256 ? nextPow2(256 / out_vector_size) : 1;
```
atamazov, line 88: [Quality] Magic numbers. It this is a legacy stuff, you can simply mark them with `/// \todo Magic numbers` (but of course it is better to fully resolve this review comment).

src/solver/softmax/softmax.cpp
```diff
@@ -0,0 +1,340 @@
...
+    out_usefp32 = true;
+    if(in_desc.GetType() == miopenHalf)
+    {
+        out_usefp16 = true;
+        out_usefp32 = false;
+    }
+
+    if(out_num_batch == 1)
+    {
+        out_workgroups = std::min(out_grid_size, 64 * 40 * 8);
```
atamazov, line 102: Ditto

atamazov (commented):
Some questions + comments.

This can be merged provided that performance testing results (baseline vs this PR) are good.

atamazov:
@Vsevolod1983 Good work! What about performance testing?

CAHEK7:
> Some questions + comments.
> 
> This can be merged provided that performance testing results (baseline vs this PR) are good.

Just a sidenote:
since GPU part has not been changed - the only affected part is a host part.

CAHEK7 (approved):

CAHEK7:
@atamazov could you help with perf measurement methodology?
It's not quite obvious what should be measured for this particular set of changes.

src/include/miopen/softmax/problem_description.hpp
CAHEK7: I would ask more generic question - what is the lifetime of all the references? Can we guarantee that it won't be constructed from temporaries?

I'm not quite sure that lifetime extension will work in that case:
```C++
ProblemDescriptor pd(TensorDescriptor());
auto& td_red = pd.getTensortDescriptor();
```

CAHEK7: It is potentially unsafe.
https://godbolt.org/z/vnfM9EYv7

CAHEK7 (changes requested):

src/softmax.cpp
```diff
@@ -0,0 +1,127 @@
...
+        v |= v >> 1;
+        v |= v >> 2;
+        v |= v >> 4;
+        v |= v >> 8;
+        v |= v >> 16;
+        v++;
+        return v;
+    }
+}
+
```
CAHEK7:
```suggestion
```

src/include/miopen/softmax/problem_description.hpp
Vsevolod1983: I will change this on using just Tensors but not const references to tensors.

src/solver/softmax/softmax.cpp
Vsevolod1983: legacy, will put a comment

src/include/miopen/softmax/invoke_params.hpp
Vsevolod1983: Q1: Could you please clarify, how exactly you suggest to make a change?
Q2: I thought about is it better to separate classes or not and concluded its better to have one class, because functionality is really similar. Also, I checked implementation for other operations and saw that this approach is quite commoin. 
I introduced two constructors for more handy usage.

Vsevolod1983:
> @Vsevolod1983 Good work! What about performance testing?

I reworked PD and  InvokeParams and now using TensorDescriptor instead of const ref to it
Also, put changes regarding the rest of the comments(vast majority of them)

Vsevolod1983:
@atamazov 
image
Is it ok ?

atamazov:
@Vsevolod1983 Pictures? 🤣 
> Is it ok ?

Who knows? In order to evaluate how performance has changed, the one needs to compare the baseline vs patch. What are the baseline (currently, this is `develop` at commit 959a7aa0803e64751f36b036795e797fb03563c8) numbers? Let's have a call tomorrow (or on Monday).

src/solver/softmax/softmax.cpp
atamazov: [resolved]

src/solver/softmax/softmax.cpp
atamazov: [resolved]

src/include/miopen/softmax/invoke_params.hpp
atamazov: Ok [resolved]

src/include/miopen/softmax/problem_description.hpp
atamazov: @Vsevolod1983 Oops, yeah, we need copies of TDs. Let's follow the pattern used for convolutions.

CAHEK7 (approved):
Seems fine. Lets take a look at perf measurements.

Vsevolod1983:
Compared PR time on softmax with develop's branch time. So if talk about so called Wall Time we have slight time execution increase on PR which is quite expected.
Wall Time diff is apprx 0.5% for SoftMaxbackward and 2.4% for SoftmaxForward. However, it's worth to mention that it slightly differs from run to run, which is probably also expected.

src/include/miopen/softmax/invoke_params.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+private:
+    void InitializeAlphaBeta(const void* alpha_, const void* beta_)
+    {
+        alpha = 0.0f;
+        beta  = 0.0f;
+
+        if(alpha_ != nullptr)
+        {
+            alpha = *(static_cast<const float*>(alpha_));
```
junliume, line 136: @Vsevolod1983 This line and the Line 141 below:
```
[2024-01-30T19:37:27.888Z] /home/jenkins/workspace/MLLIBS_MIOpen_PR-2671/src/include/miopen/softmax/invoke_params.hpp:136:21: error: implicit conversion 'float' -> bool [readability-implicit-bool-conversion,-warnings-as-errors]

[2024-01-30T19:37:27.888Z]             alpha = *(static_cast<const float*>(alpha_));

[2024-01-30T19:37:27.888Z]                     ^

[2024-01-30T19:37:27.888Z]                     (                                    != 0.0f)
```

atamazov:
## 🌀 Performance testing results

Preconditions:
- Navi21 (gfx1030/36)
- ROCm 6.0 release docker, Ubuntu 20.04.5 LTS
- BUILD_DEV=Off, Release build
- develop ad21ccf7e vs. branch 76057d8e8
- 35 known FP32 configs (borrowed from my convolution tests), tensor sizes vary from 10k to ~2.9M elements
- Binary cache pre-filled with kernels, `-i 1000` used to neutralize the effect of loading kernels from the cache. This is necessary because the softmax driver neither implements the "Aux Wall Time" measurements nor the "warm-up" function.

Results:
- 🟢 GPU Time - jitter around 0.2% (expected)
- 🟢 Wall Time (includes GPU time) - average drop around 0.8%, max drop 6%
  - [Informative] Average drop for Forward is 0.6%, for Backward is 0.9%. This doesn't match the results listed at https://github.com/ROCm/MIOpen/pull/2671#issuecomment-1915292066. I'll share the methodology and scripts with Vsevolod (and everyone who wants to get this too).
- 🟡 Host Overhead Time (computed as `wall_time - gpu_time`) - average drop 4.5%, max drop 9%

Good to go!

src/softmax/problem_description.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+        std::tie(n_y, c_y, h_y, w_y)     = tien<4>(yDesc.GetLengths());
+        std::tie(n_dy, c_dy, h_dy, w_dy) = tien<4>(dyDesc.GetLengths());
+        std::tie(n_dx, c_dx, h_dx, w_dx) = tien<4>(xdxDesc.GetLengths());
+
+        network_config += "n_y" + std::to_string(n_y) + "c_y" + std::to_string(c_y) + "h_y" +
+                          std::to_string(h_y) + "w_y" + std::to_string(w_y) +
+
+                          "n_dy" + std::to_string(n_dy) + "c_dy" + std::to_string(c_dy) + "h_dy" +
+                          std::to_string(h_dy) + "w_dy" + std::to_string(w_dy) +
```
atamazov, line 71: [Notice] 💡 You can use EOL comments to enforce structured formatting, like this:
```C++
network_config += "n_y" + std::to_string(n_y) //
        + "c_y" + std::to_string(c_y) //
        + "h_y" + std::to_string(h_y) //
        + "w_y" + std::to_string(w_y) //
...
```
clang-format will do the rest of the job.

src/softmax/problem_description.cpp
```diff
@@ -0,0 +1,90 @@
...
+        int n_dy, c_dy, h_dy, w_dy;
+        int n_dx, c_dx, h_dx, w_dx;
+
+        std::tie(n_y, c_y, h_y, w_y)     = tien<4>(yDesc.GetLengths());
+        std::tie(n_dy, c_dy, h_dy, w_dy) = tien<4>(dyDesc.GetLengths());
+        std::tie(n_dx, c_dx, h_dx, w_dx) = tien<4>(xdxDesc.GetLengths());
+
+        network_config += "n_y" + std::to_string(n_y) + "c_y" + std::to_string(c_y) + "h_y" +
+                          std::to_string(h_y) + "w_y" + std::to_string(w_y) +
+
```
atamazov, line 69: [Notice] No empty lines please. But you can use EOL comment in empty line to emphasize that it is left empty intentionally.
```C++
       network_config += "n_y" + std::to_string(n_y) + "c_y" + std::to_string(c_y) + "h_y" +
                          std::to_string(h_y) + "w_y" + std::to_string(w_y) +
                          //
                          "n_dy" + std::to_string(n_dy) + "c_dy" + std::to_string(c_dy) + "h_dy" +
                          std::to_string(h_dy) + "w_dy" + std::to_string(w_dy) +
```
but I do not know how clang-format will format this.

src/softmax/problem_description.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if(isForward)
+    {
+        int n_x, c_x, h_x, w_x;
+        int n_y, c_y, h_y, w_y;
+
+        std::tie(n_x, c_x, h_x, w_x) = tien<4>(xdxDesc.GetLengths());
+        std::tie(n_y, c_y, h_y, w_y) = tien<4>(yDesc.GetLengths());
+
+        network_config += "n_x" + std::to_string(n_x) + "c_x" + std::to_string(c_x) + "h_x" +
```
atamazov, line 48: Rewrite this stuff using `ostringstream`, see https://github.com/ROCm/MIOpen/blob/c7fa6f7fbec5f2a76128e2c9ae80f66c802722a6/src/conv/problem_description.cpp#L106

atamazov (changes requested):
Use string streams instead of merging strings. Some formatting tips.

src/softmax/problem_description.cpp
Vsevolod1983: @atamazov 
Merging strings in this style were in initial code which is in develop currently. Scope of this ticket was to introduce solver infrastructure for softmax. Could we improve this in additional ticket / PR ? 

Clang-format was run on this PR.

atamazov: @Vsevolod1983 The "problem description" class is introduced in this PR, so I would prefer fixing right here. However, I won't insist if you would like to postpone this for a while.

**_This PR can be merged as is provided that the next your PR is the one that resolves this review comment._**

atamazov (approved):
🟢 LGTM

src/softmax/problem_description.cpp
atamazov: Agreed to make the fix in the next PR.

Vsevolod1983: Discussed with @atamazov . PR can be merged.

junliume (approved):

src/solver/softmax/softmax.cpp
```diff
@@ -0,0 +1,341 @@
...
+                       in_nstr,
+                       in_cstr,
+                       in_hstr,
+                       out_nstr,
+                       out_cstr,
+                       out_hstr,
+                       params.xdx_offset,
+                       params.y_offset,
+                       alpha,
+                       beta);
```
atamazov, line 286: [Performance] @Vsevolod1983 ⚠️ **This review is written with an assumption that we are passing alpha and beta as kernel arguments, _AND intend to do so in the future._** 

If the above is true, then it does not make sense to create different Invoker instances for different alpha and beta, because kernel binaries remain the same for different alpha and beta. We are simply wasting time.

- Note: I am wondering if this is one of the reasons of 4.5% Host Overhead Time regression mentioned at https://github.com/ROCm/MIOpen/pull/2671#issuecomment-1918034107.

Yes, we _can_ include alpha and beta in PD (ProblemDescription), but they should be of no use there. Specifically:
- thou shalt not serialize them into cache/db keys, including keys for the Invoker cache. 
- thou shalt not use these values in IsApplicable and other solver functions.
- thou shalt not capture them in the InfokerFactory.

The only use of alpha and beta is passing via InvokeParams.

Right here: take alpha and beta values from params:
```C++
                kernel(params.x,
                       ...
                       params.alpha,
                       params.beta);
```
plus some other changes (details omitted).

/cc @CAHEK7 @DrizztDoUrden

atamazov (commented):
Let's use the same Invoker for different alpha/beta.

PR #2677, opened 2024-01-15
fix compilation on Windows

apwojcik:
The Windows stopped compiling tests in Debug because the VS C++ standard library has no additional overloads for `std::isnan()` and `std::isinf()` for integral types. The additional overloads are optional to be provided. They only need to be sufficient to ensure that for their argument `num` of integer type, `std::isnan(num)` has the same effect as `std::isnan(static_cast<double>(num))`. Read for more at https://en.cppreference.com/w/cpp/numeric/math/isnan and https://en.cppreference.com/w/cpp/numeric/math/isinf.


shurale-nkn (approved):
LGTM, but we can remove this check, because the system will report it later anyway. https://github.com/ROCm/MIOpen/blob/4c5bdf7097cd3314c21abaf294b29d5267faa183/test/driver.hpp#L677

junliume (approved):

apwojcik:
> LGTM, but we can remove this check, because the system will report it later anyway.
> 
> https://github.com/ROCm/MIOpen/blob/4c5bdf7097cd3314c21abaf294b29d5267faa183/test/driver.hpp#L677

I removed the asserts as you suggested - this is a better solution.

PR #2691, opened 2024-01-23
[OCL] Use OpenCL 2.0 while compiling kernels

junliume:
if opencl 1.2 kernels are launched with non-uniform block, it needs either MIOpen use -cl-std=2.0, or https://github.com/llvm/llvm-project/pull/79026


JehandadKhan (approved):

atamazov (approved):
LGTM!

PR #2698, opened 2024-01-26
measurement tool useful for debugging of performance regressions

bghimireamd:
* This PR finds the performance hipMemcpy, hipMemset and SetTensor

The performance results for tensor {4, 4, 1115, 1115} in micro sec were : 

hipMemcpy = 29717
hipMemset = 1
testSetTensor = 174

### DIFF
--- a/test/perf_memcpy_memset_setTensor.cpp
+++ b/test/perf_memcpy_memset_setTensor.cpp
@@ -1,0 +1,142 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include "miopen/tensor_ops.hpp"
+#include "tensor_holder.hpp"
+#include "gtest/conv_tensor_gen.hpp"
+#include "get_handle.hpp"
+
+#include <chrono>
+
+#define TEST_SAMPLES 10
+
+const std::vector<size_t> GetInput() { return {4, 4, 1115, 1115}; }
+
+// set the gpu_ptr(device)
+template <typename T>
+long long testSetTensor(void* gpu_ptr, const tensor<T>& input)
+{
+    float zero               = 0.f;
+    auto&& handle            = get_handle();
+    long long total_durtaion = 0;
+
+    for(int i = 0; i < TEST_SAMPLES; ++i)
+    {
+        auto start = std::chrono::high_resolution_clock::now();
+        SetTensor(handle, input.desc, gpu_ptr, &zero);
+        // End time
+        auto end = std::chrono::high_resolution_clock::now();
+
+        // Calculating total duration
+        total_durtaion +=
+            std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+    }
+    return total_durtaion / TEST_SAMPLES;
+}
+
+// copy input data (host) to gpu_ptr(device)
+template <typename T>
+long long testHIPMemcpy(void* gpu_ptr, const tensor<T>& input)
+{
+    long long total_durtaion = 0;
+    for(int i = 0; i < TEST_SAMPLES; ++i)
+    {
+        auto start        = std::chrono::high_resolution_clock::now();
+        auto ret_cpy_hTod = hipMemcpy(static_cast<void*>(gpu_ptr),
+                                      input.data.data(),
+                                      sizeof(T) * input.data.size(),
+                                      hipMemcpyHostToDevice);
+        auto end          = std::chrono::high_resolution_clock::now();
+
+        if(ret_cpy_hTod != hipSuccess)
+        {
+            std::cout << "\n\nhipMemcpy from host to device : Error code :" << ret_cpy_hTod
+                      << "\n\n";
+            exit(1);
+        }
+        total_durtaion +=
+            std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+    }
+    return total_durtaion / TEST_SAMPLES;
+}
+
+// set the gpu_ptr(device)
+template <typename T>
+long long testHIPMemset(void* gpu_ptr, const tensor<T>& input)
+{
+    long long total_durtaion = 0;
+    for(int i = 0; i < TEST_SAMPLES; ++i)
+    {
+        auto start = std::chrono::high_resolution_clock::now();
+        auto ret_cpy_hTod =
+            hipMemset(static_cast<void*>(gpu_ptr), 0, sizeof(T) * input.data.size());
+        auto end = std::chrono::high_resolution_clock::now();
+
+        if(ret_cpy_hTod != hipSuccess)
+        {
+            std::cout << "\n\nhipMemcpy from host to device : Error code :" << ret_cpy_hTod
+                      << "\n\n";
+            exit(1);
+        }
+        total_durtaion +=
+            std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+    }
+    return total_durtaion / TEST_SAMPLES;
+}
+
+int main()
+{
+    std::cout << "\nStart benchmarking ..\n";
+    std::map<std::string, long long> perf_data;
+
+    using T                            = float;
+    miopenTensorLayout_t tensor_layout = miopenTensorNCHW;
+    tensor<T> input;
+    tensor<T> output;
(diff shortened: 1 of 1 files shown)


CAHEK7:
1us for hipMemset of `4 * 4 * 1115 * 1115 * sizeof(float)` ~ 76MB seems strange.
gfx1030 has 476GB/s global memory rate and around 1600GB/s L3 rate.

Now lets do the math: 76MB by 1e-6s means ~76000GB/s memory throughput. And it's not a pure GPU time, it's a roundtrip host time (including hip runtime, driver and scheduling overhead), so pure memory throughput will be even bigger. 
`testSetTensor` looks more real: around 426GB/s - something similar to gfx1030 global memory throughput. But 76Mb fits perfectly into L3, so performance can be better.
Does `hipMemset` synchronous? Can you put explicit synchronization and measure again?

hipMemcpy looks too slow - 16 lines of PCIe4.0 is about 31.5GB/s, while here we've got 2.5Gb/s. Just for the experiment, can you try to use pinned memory (hipHostMalloc).

It would be nice to see HW configuration for the perf metrics.

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+        auto start = std::chrono::high_resolution_clock::now();
+        SetTensor(handle, input.desc, gpu_ptr, &zero);
+        // End time
+        auto end = std::chrono::high_resolution_clock::now();
+
+        // Calculating total duration
+        total_durtaion +=
+            std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+    }
+    return total_durtaion / TEST_SAMPLES;
```
CAHEK7, line 56: Shall we consider FP division?

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+        if(ret_cpy_hTod != hipSuccess)
+        {
+            std::cout << "\n\nhipMemcpy from host to device : Error code :" << ret_cpy_hTod
+                      << "\n\n";
+            exit(1);
+        }
+        total_durtaion +=
+            std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+    }
+    return total_durtaion / TEST_SAMPLES;
```
CAHEK7, line 82: Shall we consider FP division?

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
```
CAHEK7, line 106: Shall we consider FP division?

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+    }
+    // do the performance
+    perf_data["hipMemcpy"]     = testHIPMemcpy(gpu_ptr, input);
+    perf_data["hipMemset"]     = testHIPMemset(gpu_ptr, input);
+    perf_data["testSetTensor"] = testSetTensor(gpu_ptr, input);
+
+    for(auto it : perf_data)
+    {
+        std::cout << it.first << " = " << it.second << "\n";
+    }
```
CAHEK7, line 139:
```suggestion
    for (const auto& [name, val] : perf_data)
    {
        std::cout << name << " = " << val << "\n";
    }
```
structure binding

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+    tensor<T> input;
+    tensor<T> output;
+    void* gpu_ptr;
+
+    input = tensor<T>{tensor_layout, GetInput()};
+    input.generate(GenData<T>{});
+    output.data.resize(input.data.size());
+
+    // allocate memory in GPU
+    auto ret_malloc = hipMalloc(static_cast<void**>(&gpu_ptr), sizeof(T) * input.data.size());
```
CAHEK7, line 125:
```suggestion
    auto ret_malloc = hipMalloc(&gpu_ptr, sizeof(T) * input.data.size());
```
`gpu_ptr` is already `void*` and `&` makes it `void**`.

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    using T                            = float;
+    miopenTensorLayout_t tensor_layout = miopenTensorNCHW;
+    tensor<T> input;
+    tensor<T> output;
+    void* gpu_ptr;
+
+    input = tensor<T>{tensor_layout, GetInput()};
+    input.generate(GenData<T>{});
+    output.data.resize(input.data.size());
```
CAHEK7, line 122: Do we need `output`?

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+// set the gpu_ptr(device)
+template <typename T>
+long long testHIPMemset(void* gpu_ptr, const tensor<T>& input)
+{
+    long long total_durtaion = 0;
+    for(int i = 0; i < TEST_SAMPLES; ++i)
+    {
+        auto start = std::chrono::high_resolution_clock::now();
+        auto ret_cpy_hTod =
+            hipMemset(static_cast<void*>(gpu_ptr), 0, sizeof(T) * input.data.size());
```
CAHEK7, line 94:
```suggestion
            hipMemset(gpu_ptr, 0, sizeof(T) * input.data.size());
```
`gpu_ptr` is already "void*"

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+// copy input data (host) to gpu_ptr(device)
+template <typename T>
+long long testHIPMemcpy(void* gpu_ptr, const tensor<T>& input)
+{
+    long long total_durtaion = 0;
+    for(int i = 0; i < TEST_SAMPLES; ++i)
+    {
+        auto start        = std::chrono::high_resolution_clock::now();
+        auto ret_cpy_hTod = hipMemcpy(static_cast<void*>(gpu_ptr),
```
CAHEK7, line 67:
```suggestion
        auto ret_cpy_hTod = hipMemcpy(gpu_ptr,
```
gpu_ptr is already "void*"

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+                                      input.data.data(),
```
CAHEK7, line 68: It's not a `zero` memory - it contains random data.
It doesn't matter a lot, but anyway.

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+    std::map<std::string, long long> perf_data;
+
+    using T                            = float;
+    miopenTensorLayout_t tensor_layout = miopenTensorNCHW;
+    tensor<T> input;
+    tensor<T> output;
+    void* gpu_ptr;
+
+    input = tensor<T>{tensor_layout, GetInput()};
+    input.generate(GenData<T>{});
```
CAHEK7, line 121: It should be zeroed, not randomly initialized. It's not critical at all.

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+int main()
+{
+    std::cout << "\nStart benchmarking ..\n";
+    std::map<std::string, long long> perf_data;
+
+    using T                            = float;
+    miopenTensorLayout_t tensor_layout = miopenTensorNCHW;
+    tensor<T> input;
```
CAHEK7, line 116:
```suggestion
    tensor<T> input(tensor_layout, GetInput());
```

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+    tensor<T> output;
+    void* gpu_ptr;
+
+    input = tensor<T>{tensor_layout, GetInput()};
```
CAHEK7, line 120:
```suggestion
```

CAHEK7 (commented):
Some C++ related comments. Probably it could useful.

atamazov:
@bghimireamd I suspect this is not a _test_ but a _measurement tool_ useful for debugging of performance regressions etc. Can you please rename the PR accordingly and replace https://github.com/ROCm/MIOpen/labels/testing with https://github.com/ROCm/MIOpen/labels/debugging + https://github.com/ROCm/MIOpen/labels/performance + https://github.com/ROCm/MIOpen/labels/enhancement ? Thanks!

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+const std::vector<size_t> GetInput() { return {4, 4, 1115, 1115}; }
+
+// set the gpu_ptr(device)
+template <typename T>
+long long testSetTensor(void* gpu_ptr, const tensor<T>& input)
+{
+    float zero               = 0.f;
+    auto&& handle            = get_handle();
+    long long total_durtaion = 0;
```
CAHEK7, line 43:
```suggestion
    long long total_duration = 0;
```
Tiny typo

CAHEK7:
There are also a `hipMemset2D` and `hipMemset3D` functions which deal with strides (I hope in an optimal way)

Besides this particular PR - it seems we don't often use async operations. Is there any particular reason for it?

test/perf_memcpy_memset_setTensor.cpp
CAHEK7: I'd say if you are using something different of `std::chrono::nanoseconds`, all kind of integer computations will result in truncations.

For example, for the first iteration it gets 1999ns for `(end-start)` operation, but it results into 1us, for the last 9 iterations it always gets 2999ns (and 2us). And later, after integer division by `TEST_SAMPLES` you may get 19 / 10 = 1us.
Instead of 2.899us. Kind of huge error.
But even taking into account this precision loss, it can't explain so tiny execution time for `hipMemset` - it still ends up into an enormous memory throughput.

Anyway, either `double` should be used for a duration or `std::chrono::nanoseconds` precision for the cast should be used.

bghimireamd:
> There are also a `hipMemset2D` and `hipMemset3D` functions which deal with strides (I hope in an optimal way)
> 
> Besides this particular PR - it seems we don't often use async operations. Is there any particular reason for it?

@CAHEK7 ok will add them too.

test/perf_memcpy_memset_setTensor.cpp
```diff
@@ -0,0 +1,142 @@
...
+        }
+        total_durtaion +=
+            std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+    }
+    return total_durtaion / TEST_SAMPLES;
+}
+
+int main()
+{
+    std::cout << "\nStart benchmarking ..\n";
```
reidkwja, line 111: Instead of cout, it might be preferable to use logging with timestamp...or keep the cout and add timestamp to each output.

junliume:
@bghimireamd could you follow up with the review opinions? We can move forward with this PR. Thanks!

BradPepersAMD:

Issue #2701, opened 2024-01-27
[DB Install] Likely after #2347 `make install` does not install DB files correctly


junliume:
**[Observation]**:
```
about make install at this moment:
Install the project...
-- Install configuration: "Release"
-- Installing: /opt/rocm/share/doc/miopen-hip/LICENSE.txt
-- Installing: /opt/rocm/include/miopen/export.h
-- Installing: /opt/rocm/share/miopen/db/gfx908.tn.model
-- Installing: /opt/rocm/share/miopen/db/gfx908_ConvAsm1x1U_decoder.ktn.model
-- Installing: /opt/rocm/share/miopen/db/gfx908_ConvAsm1x1U_encoder.ktn.model
-- Installing: /opt/rocm/share/miopen/db/gfx908_ConvAsm1x1U_metadata.ktn.model
-- Installing: /opt/rocm/share/miopen/db/gfx908_metadata.tn.model
-- Installing: /opt/rocm/share/miopen/db/gfx90a.tn.model
-- Installing: /opt/rocm/share/miopen/db/gfx90a_ConvHipIgemmGroupFwdXdlops_decoder.ktn.model
-- Installing: /opt/rocm/share/miopen/db/gfx90a_ConvHipIgemmGroupFwdXdlops_encoder.ktn.model
-- Installing: /opt/rocm/share/miopen/db/gfx90a_ConvHipIgemmGroupFwdXdlops_metadata.ktn.model
-- Installing: /opt/rocm/share/miopen/db/gfx90a_metadata.tn.model
-- Up-to-date: /opt/rocm/include
-- Up-to-date: /opt/rocm/include/miopen
-- Installing: /opt/rocm/include/miopen/miopen.h
-- Up-to-date: /opt/rocm/include
-- Up-to-date: /opt/rocm/include/miopen
-- Up-to-date: /opt/rocm/include/miopen/export.h
-- Installing: /opt/rocm/include/miopen/config.h
-- Installing: /opt/rocm/include/miopen/version.h
-- Installing: /opt/rocm/lib/libMIOpen.so.1.0.60100
-- Up-to-date: /opt/rocm/lib/libMIOpen.so.1
-- Set runtime path of "/opt/rocm/lib/libMIOpen.so.1.0.60100" to ""
-- Up-to-date: /opt/rocm/lib/libMIOpen.so
-- Installing: /opt/rocm/lib/cmake/miopen/miopen-targets.cmake
[...]

PR #2702, opened 2024-01-27
[HotFix] Fix DB install after #2347

junliume:
#2701

Reason for Draft: need to take care of embedded DB case


apwojcik (approved):
Thanks for catching this up and for the fix!

junliume:
@apwojcik @JehandadKhan this is not completely fixed yet. If one issues: `make MIOpenDriver` then the DB targets will not be made automatically. They are made only when default `all` is made.

JehandadKhan:
@apwojcik Can you please update this branch so that the db targets are added as dependencies to the `MIOpen` target.

junliume:
@apwojcik there is another issue on `LFS_MISSING_FILES` from
https://github.com/ROCm/MIOpen/blob/06b2bd395f2629b1542d5b2a5d234195f9b41bb8/CMakeLists.txt#L515

This variable gets cleared from scratch every time the function `install_kdb` is called. I guess we need to make it a global variable somehow?

same issue has been discussed here: https://stackoverflow.com/questions/10031953/how-to-set-the-global-variable-in-a-function-for-cmake

apwojcik:
> This variable gets cleared from scratch every time the function `install_kdb` is called. I guess we need to make it a global variable somehow?

@junliume Can you elaborate more on why it is an issue here? I would like to understand the context.

apwojcik:
> @apwojcik @JehandadKhan this is not completely fixed yet. If one issues: `make MIOpenDriver` then the DB targets will not be made automatically. They are made only when default `all` is made.

@junliume @JehandadKhan It is done

junliume:
> > This variable gets cleared from scratch every time the function `install_kdb` is called. I guess we need to make it a global variable somehow?
> 
> @junliume Can you elaborate more on why it is an issue here? I would like to understand the context.

@apwojcik the issue is described in more details here: https://github.com/ROCm/MIOpen/issues/2709

PR #2711, opened 2024-01-31
move the bfloat16 header to the proper guard

JehandadKhan:


atamazov:
@JehandadKhan @junliume Tested with ROCm 5.7.0 (BF16 Full tests) on Navi21, no regressions.

Maybe this header is not necessary at all. Otherwise symbol conflicts ("duplicate definition") errors  would happen during HIPRTC builds.

🟢 Anyway, this is good to go.

JehandadKhan:
> Maybe this header is not necessary at all. Otherwise symbol conflicts ("duplicate definition") errors would happen during HIPRTC builds.

That has been bugging me, I dont quite understand why that did not happen. If HIPRTC has these symbols defined internally then including a file with the same symbols should cause conflict. However, that did not happen!

junliume:
> > Maybe this header is not necessary at all. Otherwise symbol conflicts ("duplicate definition") errors would happen during HIPRTC builds.
> 
> That has been bugging me, I dont quite understand why that did not happen. If HIPRTC has these symbols defined internally then including a file with the same symbols should cause conflict. However, that did not happen!

That's why I think a ticket on runtime to explain is needed.

junliume (approved):

PR #2719, opened 2024-02-02
Do not use HIP runtime headers on Windows

apwojcik:
To compile MIOpen kernels on Windows, we need to turn off the use of HIP-provided runtime header files and workaround for SWDEV_413293. The HIP compiler on Windows does not recognize the compile option `-fno-offload-uniform-block` and is failing compilation.


junliume (approved):

atamazov:
[Informative] @junliume @apwojcik FYI this is not a full-blown fix, but just a workaround for some Windows-specific problems that likely disappear in the future.

Let's formalize future workarounds as specified at #1685.

PR #2722, opened 2024-02-02
Implemented preparsing sqlite db to text format

DrizztDoUrden:
This adds tools/sqlite2txt - a separate program that converts .db perfdb files into .txt for use in ramdb and readonlyramdb classes. SQLite code from miopen has not been reused because it would be very hard to extract just couple of files, plus the tool has to be built before library.
The key used in the internal map represents sqlite column values collection due to how it is being built. Finddb key format cannot be reused there because it contains more information.

Side bonus: .db.txt files are approximately 2 times smaller than .db files.

Works fine on the PCs I have access to and in all configurations that I know of.


junliume:
@DrizztDoUrden please help to resolve the conflicts, and convert from draft when ready. Thanks!

junliume:
@DrizztDoUrden DB install in CMake has changed recently, could you merge the TIP of develop branch?
current issue:
```
[2024-02-02T18:38:31.462Z] /home/jenkins/workspace/MLLIBS_MIOpen_ddu_text-perfdb/test/gtest/cache.cpp:97:20: error: no matching constructor for initialization of 'miopen::KernDb'

[2024-02-02T18:38:31.462Z]     miopen::KernDb empty_db("", false);

[2024-02-02T18:38:31.462Z]                    ^        ~~~~~~~~~
```

CMakeLists.txt
```diff
@@ -487,16 +487,36 @@ set(KERNELS_SOURCE_DIR ${PROJECT_SOURCE_DIR}/src/kernels)
 
 find_program(UNZIPPER lbzcat bzcat REQUIRED)
 file(MAKE_DIRECTORY ${KERNELS_BINARY_DIR})
+add_subdirectory(tools/sqlite2txt)
```
apwojcik: It this tool always required? Can we make it optional?

CMakeLists.txt
```diff
@@ -487,16 +487,36 @@ set(KERNELS_SOURCE_DIR ${PROJECT_SOURCE_DIR}/src/kernels)
...
+    set(PERFDB_SUFFIX "")
+else()
+    set(PERFDB_SUFFIX ".txt")
+endif()
+
 function(unpack_db db_bzip2_file)
     get_filename_component(__fname ${db_bzip2_file} NAME_WLE)
     add_custom_command(OUTPUT ${KERNELS_BINARY_DIR}/${__fname}
-                       COMMAND ${UNZIPPER} -k ${db_bzip2_file} > ${KERNELS_BINARY_DIR}/${__fname})
+                       COMMAND ${UNZIPPER} -ck ${db_bzip2_file} > ${KERNELS_BINARY_DIR}/${__fname})
```
apwojcik: There is no need for `-c` when used with `bzcat` - it breaks Windows when used explicitly (`bzcat.exe` is a regular executable), and when missing on Linux, it does not harm (shell script over `bzip2` and `-c` automatically applied?). Is there a reason you added the `-c` option here? According to the `bzcat` man page, `-c` should be used only for the `bzip2` command. The same man page explains: "bzcat - decompresses files to stdout."

atamazov:
@DrizztDoUrden Let's add this to Jenkinsfile in order to keep SQLilte perf-db functional:
```groovy
stage('Fp32 Hip SqlitePerfdb gfx90a') {
    when {
        beforeAgent true
        expression { params.TARGET_GFX90A }
    }
    options {
        retry(2)
    }
    agent{ label rocmnode("gfx90a") }
    steps{
        buildHipClangJobAndReboot(make_targets: Smoke_targets, setup_flags: "-DMIOPEN_USE_SQLITE_PERF_DB=On")
    }
}
```

CMakeLists.txt
DrizztDoUrden: Removed that -c. For some reason it failed for me on windows initially, but than I have forgotten to remove it after I have fixed the initial issue. Good catch!

CMakeLists.txt
DrizztDoUrden: Made it only included when MIOPEN_ENABLE_SQLITE_PERFDB not enabled (as it wouldn't be needed otherwise).

junliume:
@DrizztDoUrden maybe it could be fixed quickly?
```
test_perfdb (Failed)
```

JehandadKhan:
@DrizztDoUrden  Please update the branch with develop.

junliume (approved):

atamazov:
@junliume @DrizztDoUrden I am afraid this must be reverted.

atamazov:
@DrizztDoUrden (1) The results of conversion produce incorrect result on my machine (one record with key `0x0x0x0x0x0x0x0x0x0x0x0x0x0x0x0x0x0x0x0x0xxx`).

(2) [less important] Many error message like 
```
MIOpen(HIP): Error [Prefetch] Ill-formed record: key not found: /root/github/MLOpen1/build/release.hip-clang/share/miopen/db/gfx1030_36_1.1.0.udb#3
```
because text-based perf-db uses the same name as SQLite-based one (*.udb) and the library tries to re-use existing sqlite db as text.

atamazov:
@junliume HOTFIX will be ready soon.

PR #2723, opened 2024-02-02
MIOpen Documentation for ROCm 6.1

Rmalavally:
Updated and restructured MIOpen documentation for ROCm 6.1.

Pending: Doxygen links for API content

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -940,18 +940,6 @@ COMMAND	$<TARGET_FILE:test_conv2d>	--verbose	--input	8	3	108	108	--weights	63	1	
 
 if(MIOPEN_TEST_DEEPBENCH)
     add_custom_test(test_deepbench_rnn GFX94X_ENABLED GFX103X_ENABLED GFX110X_ENABLED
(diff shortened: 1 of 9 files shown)


docs/install/install.rst
randyh62, line 1: It looks like the header tags are not correct in this file.

Rmalavally: Yes, Randy. Working on it.

docs/index.rst
randyh62, line 1: It looks like the index needs the card layout: Install, Conceptual, Reference, Tutorials

docs/install/install.rst
Rmalavally: @bghimireamd Could you please confirm?

docs/example.rst
randyh62, line 1: Probably can remove this empty file?

docs/index.rst
Rmalavally: Yes. Adding index files.

docs/conceptual/Getting_Started_FusionAPI.md
randyh62, line 1: Is this the same file as the file named the same in the Tutorials folder?

randyh62 (commented):
It looks like the files in the Conceptual folder are the same, or at least named the same as the files in the Tutorials folder? I left some other comments, but think there may be some work to do here.

Rmalavally:
> It looks like the files in the Conceptual folder are the same, or at least named the same as the files in the Tutorials folder? I left some other comments, but think there may be some work to do here.

Yes. Need to clean up.

PR #2724, opened 2024-02-03
[BugFix] Set System KDB journal_mode to Off

cderb:
gfx908 + gfx90a KDB journal_mode was set to record, which caused file system permission errors during runtime, as these files are read only and not user writeable.
Added test to db_sync to catch this error.


test/gtest/db_sync.cpp
```diff
@@ -422,6 +422,26 @@ bool CheckKDBForTargetID(const fs::path& filename)
...
+    std::string journal_mode;
+    while(true)
+    {
+        auto rc = stmt.Step(sql);
+        if(rc == SQLITE_ROW)
+            journal_mode = stmt.ColumnText(0);
+        else if(rc == SQLITE_DONE)
+            break;
+        else if(rc == SQLITE_ERROR || rc == SQLITE_MISUSE)
+            throw std::runtime_error(sql.ErrorMessage());
```
bghimireamd, line 440: will it go into infinite loop if `sqlite3_step` return `SQLITE_BUSY` or `SQLITE_LOCKED`? We could just convert last `else if` to `else` ? But since its just test, I guess we could ignore it :)

junliume: @cderb @bghimireamd I would recommend to be safe and just replace this with `else`. What do you think? CC: @JehandadKhan

junliume (approved):

PR #2690, opened 2024-01-22
Fix compilation on SELS/RHEL after #2657 merged

apwojcik:
The PR fixes compilation on SLES and RHEL, which have `std::filesystem` experimental implementation. It also changes the existing `boost::filesystem` to `std::filesystem` - the MSVC standard library is incompatible with Boost and requires extra conversion/translation between Boost and Std. The example is when `boost::filesystem::path` is used with `std::ofstream` - GNU has additional construction accepting Boost path, MSVC does not, and we need to convert each time to either Std path or `std::string`.
The PR is a prerequisite to the upcoming PRs changing path handling from `std::string` to `fs::path` for most use cases.

**Update:** fix https://github.com/ROCm/MIOpen/pull/2687

PR #2732, opened 2024-02-06
fix build after #2657 and #2690 (boost::filesystem)

CAHEK7:
#2657 introduced `std::filesystem` and #2690 replaced most of `boost::filesystem` code with `std` but there are a few ifdefs where this replacement was missing.

It's not very critical for the releases, because most of the cases are under the `!MIOPEN_ENABLE_SQLITE_KERN_CACHE` (disabled kernel cache) define. But it may affect development environment.

@atamazov @JehandadKhan

### DIFF
--- a/src/hip/handlehip.cpp
+++ b/src/hip/handlehip.cpp
@@ -530,7 +530,7 @@ Program Handle::LoadProgram(const std::string& program_name,
                            program_name,
                            params);
 #else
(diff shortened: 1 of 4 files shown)


apwojcik (approved):
Thanks for catching this up!

junliume (approved):

atamazov (approved):
🟢 LGTM

PR #2737, opened 2024-02-08
enable softmax solver based on attention-softmax implementation

CAHEK7:
This PR enables special version of softmax tailored for attention layer but can be applicable for the normal softmax as well.

Tested of gfx1030. Compared with the current openCL imeplementation.

It shows performance gain (up to 1.8-2x) for relatively small problems (`17 < c*h*w < 4096`) for any batch size.
For tiny problems (`c*h*w < 16`) it shows perf gain for small batch sizes (`<1024`) 
For problems where `c*h*w > 4096` it shows the same performance as openCL version and becomes memory-bound.

```bash
./bin/MIOpenDriver softmax -A 1 -B 0 -F 1 -t 1 -i 11 -m 0 -V 0 -H 1 -c 1 -W $PROBLEM_SIZE -n $BATCH_SIZE
```

Some heuristics have been applied.

<details>

<summary>Perf graphs</summary>

image
image
image

</details>


src/solver/softmax/attn_softmax.cpp
```diff
@@ -0,0 +1,150 @@
...
+namespace miopen::solver::softmax {
+
+namespace {
+template <typename T, typename S = std::enable_if_t<std::is_unsigned_v<T>, T>>
+constexpr inline S Ceil(const T val, const T div)
+{
+    return (val - 1 + div) / div;
+}
+
+constexpr uint32_t nextPow2(uint32_t v)
```
Vsevolod1983, line 47: We already have nextPow2 function implemented in regular softmax solver. Could you please reuse that code ?

CAHEK7: Exactly, I just wanted to highlight the problem. I've explicitly moved `nextPow2` in the regular softmax solver into anonymous namespace and I'm aware of this implementation.

We already had lots of implementations of the same or similar functions function across the project.
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/softmax/softmax.cpp#L42
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/pooling/forwardNaive.cpp#L54
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/include/miopen/pooling.hpp#L39
more generic "nextPow2":
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/conv_winoRxS.cpp#L106
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/pooling/backward2d.cpp#L113
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/pooling/forward2d.cpp#L86
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/pooling/forwardNd.cpp#L73

As well as integer ceil division:
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/conv_winoRxS.cpp#L100
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/conv_winoRxS_fused.cpp#L50
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/batchnorm/problem_description.cpp#L171
and around 300 other places with just an inline version of this function, for example:
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/pooling/backwardNd.cpp#L112.
https://github.com/ROCm/MIOpen/blob/efaafebc51c4db6f0f2731df7e2e366f5b0952fd/src/solver/pooling/forwardNd.cpp#L144

And there is no place for such kind of helper functions in the project. @atamazov probably you have any ideas?

CAHEK7: I think it has to be fixed properly - move all the identical functions not just from here, but from all the project, into a dedicated helper header, well document the header ant let everyone read it.

CAHEK7: Issue created https://github.com/ROCm/MIOpen/issues/2748

Meanwhile `nextPow2` from the regular softmax is not mathematically correct - it should not be a signed integer. It works for positive values but does not make sense for the negative (and even it triggers `signed integer` overflow, which is UB).
Unsigned version makes more sense and may trigger `unsigned integer` overflow which is perfectly defined.

Vsevolod1983 (approved):

Kirpich30000 (approved):
LGTM

junliume (approved):

PR #2744, opened 2024-02-09
Replace ROCmSoftwarePlatform links with ROCm after repo consolidation

junliume:
Previously they are linked to ROCm after consolidation.
This is low priority, we do see an issue with some repos in backward compatibility path. e.g. install `half` in pre ROCM 6.0 docker


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "fin"]
     path = fin
-    url = https://github.com/ROCmSoftwarePlatform/MIFin.git
+    url = https://github.com/ROCm/MIFin.git
```
dgaliffiAMD: Because we're all in the same organization now, you may define this as a relative path. For example,
```suggestion
    url = .../MIFin
```

PR #2745, opened 2024-02-09
Removing unmaintained release notes

saadrahim:


LisaDelaney (approved):

junliume (approved):

junliume:
Usually I would prefer archiving it rather than delete, but this file is severely outdated already

@bghimireamd could you use the info (like the date of release) and carry over to `CHANGELOG.md` ?

atamazov:
@junliume @bghimireamd AFAICS the only info that exists in https://github.com/ROCm/MIOpen/blob/develop/docs/releasenotes.md while missing in https://github.com/ROCm/MIOpen/blob/develop/CHANGELOG.md is dates. However
- Usually we do not publish dates in CHANGELOG.md
- Dates of releases can be easily found at https://github.com/ROCm/MIOpen/tags
- Dates in releasenotes.md do not match the dates at https://github.com/ROCm/MIOpen/tags.
  - Example: rocm-4.2.0 tag is dated May 10, 2021, but releasenotes has May, 17.

I would recommend not spending time on changing the CHANGELOG format and transferring information about release dates from releasenotes.md there.

bghimireamd:
@atamazov is correct. All information from releasenotes.md were carried over to CHANGELOG.md minus the dates. So, we can get rid of releasenotes.md

PR #2751, opened 2024-02-14
[MHA] Implement Tensor Descriptors for MIOpen Backend API

### LINKED ISSUE #2704: [MHA] Implement Tensor Descriptors for MIOpen Backend API
Implement cuDNN-like API for creating leaf nodes in the calculation graph and gtests for them.

MIOPEN_BACKEND_TENSOR_DESCRIPTOR
  - [x] Builder
  - [x] C API

scerzh:


src/include/miopen/backend_api.hpp
```diff
@@ -0,0 +1,62 @@
...
+    miopenStatus_t m_Status;
+    bool m_IsVirtual;
+    bool m_UniqueIdSet;
+    bool m_DataTypeSet;
+    bool m_ByteAlignmentSet;
+    bool m_DimensionsSet;
+    bool m_StridesSet;
+    bool m_VectorCountSet;
+    bool m_VectorizedDimensionSet;
+    bool m_RaggedOffsetDescSet;
```
amberhassaan: This is good practice. We haven't adopted a standard style in MIOpen yet but I'm trying to convince folks to do so. 
My preference:
Use `mMemberName` instead of `m_MemberName` as the former conveys the same and saves the typing of `_`.

include/miopen/miopen.h
```diff
@@ -5730,6 +5730,536 @@ MIOPEN_EXPORT miopenStatus_t miopenGroupNormForward(miopenHandle_t handle,
...
+ * The type of structure that it points to is determined by the argument when allocating the memory
+ * for the opaque structure using miopenBackendCreateDescriptor().
+ *
+ * Attributes of a descriptor can be set using miopenBackendSetAttribute(). After all required
+ * attributes of a descriptor are set, the descriptor can be finalized by miopenBackendFinalize().
+ * From a finalized descriptor, one can query its queryable attributes using
+ * miopenBackendGetAttribute(). Finally, the memory allocated for a descriptor can be freed using
+ * miopenBackendDestroyDescriptor().
+ */
+typedef void* miopenBackendDescriptor_t;
```
amberhassaan: Use `MIOPEN_DECLARE_OBJECT(miopenBackendDescriptor)` and that defines an empty base struct `miopenBackendDescriptor_t`. You can see examples in this file and the definition too.

src/backend_api.cpp
```diff
@@ -0,0 +1,268 @@
...
+                              miopenBackendDescriptor_t* descriptor)
+{
+    switch(descriptorType)
+    {
+    case MIOPEN_BACKEND_TENSOR_DESCRIPTOR:
+        // TODO: add logging
+        // TODO: get rid of unnecessary throws
+        return miopen::try_(
+            [&] {
+                miopen::backend_api::BackendTensorDescriptor* newDescriptor =
```
amberhassaan: use `miopen::deref(descriptor)` here. Well, now that I think of it, may be we need to tweak the `miopen::deref` or create a slightly modified version that allows multiple levels of class hierarchy.

src/backend_api.cpp
```diff
@@ -0,0 +1,268 @@
...
+                    new(std::nothrow) miopen::backend_api::BackendTensorDescriptor();
+
+                if(!newDescriptor)
+                {
+                    throw miopen::Exception(miopenStatusAllocFailed);
+                }
+
+                try
```
amberhassaan: why is this needed? we should not use `try/catch` as that's what `miopen::try_` is for.

src/backend_api.cpp
```diff
@@ -0,0 +1,268 @@
...
+                                                    miopenBackendAttributeType_t attributeType,
+                                                    int64_t elementCount,
+                                                    void* arrayOfElements)
+{
+    return miopen::try_(
+        [&] {
+            miopen::backend_api::BackendDescriptor* theDescriptor =
+                static_cast<miopen::backend_api::BackendDescriptor*>(descriptor);
+
+            miopenStatus_t status = theDescriptor->SetAttribute(
```
amberhassaan: I'm not sure if every method on the internal class should return `miopenStatus_t`.

src/include/miopen/backend_api.hpp
```diff
@@ -0,0 +1,62 @@
...
+    BackendDescriptor() : m_IsFinalized(false) {}
+    virtual ~BackendDescriptor()                               = default;
+    virtual miopenStatus_t SetAttribute(miopenBackendAttributeName_t attributeName,
+                                        miopenBackendAttributeType_t attributeType,
+                                        int64_t elementCount,
+                                        void* arrayOfElements) = 0;
+    virtual miopenStatus_t Finalize()                          = 0;
+
+protected:
+    bool m_IsFinalized;
```
amberhassaan: Initialize to `false` here.

src/include/miopen/backend_api.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+namespace miopen {
+
+namespace backend_api {
+
+class BackendDescriptor
+{
+public:
+    BackendDescriptor() : m_IsFinalized(false) {}
+    virtual ~BackendDescriptor()                               = default;
```
amberhassaan: define the destructor in `backend_api.cpp` as that allows the compiler to `anchor` the virtual destructor.

src/include/miopen/backend_api.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+private:
+    miopenTensorDescriptor_t m_Descriptor;
+    std::vector<int64_t> m_Dimensions;
+    std::vector<int64_t> m_Strides;
+    int64_t m_UniqueId;
+    int64_t m_ByteAlignment;
+    int64_t m_VectorCount;
+    int64_t m_VectorizedDimension;
+    miopenBackendDescriptor_t m_RaggedOffsetDesc;
```
amberhassaan: We should remove this for now as we are not going to support it.

src/include/miopen/backend_api.hpp
```diff
@@ -0,0 +1,62 @@
...
+                                        miopenBackendAttributeType_t attributeType,
+                                        int64_t elementCount,
+                                        void* arrayOfElements) override;
+    virtual miopenStatus_t Finalize() override;
+    inline miopenStatus_t Status() const noexcept { return m_Status; }
+    inline bool IsVirtual() const noexcept { return m_IsVirtual; }
+
+private:
+    miopenTensorDescriptor_t m_Descriptor;
+    std::vector<int64_t> m_Dimensions;
```
amberhassaan: We are duplicating information here that will eventually be in `miopen::TensorDescriptor`. I can see why `Finalize()` requires doing so but let's think of ways to get rid of it.

amberhassaan (commented):
Some initial comments.

src/include/miopen/backend_api.hpp
CAHEK7: (a short explanation)
https://llvm.org/docs/CodingStandards.html#provide-a-virtual-method-anchor-for-classes-in-headers

src/backend_api.cpp
```diff
@@ -0,0 +1,268 @@
...
+    {
+        return miopenStatusNotInitialized;
+    }
+
+    switch(attributeName)
+    {
+    case MIOPEN_ATTR_TENSOR_UNIQUE_ID:
+        if(attributeType == MIOPEN_TYPE_INT64 && elementCount == 1)
+        {
+            m_UniqueId    = *reinterpret_cast<int64_t*>(arrayOfElements);
```
CAHEK7: Nitpicking, but cast from `void *` to `T *` can be a `static_cast<T *>`

src/backend_api.cpp
```diff
@@ -0,0 +1,268 @@
...
+            m_ByteAlignmentSet = true;
+            return miopenStatusSuccess;
+        }
+        else
+        {
+            return miopenStatusBadParm;
+        }
+
+    case MIOPEN_ATTR_TENSOR_DIMENSIONS:
+        if(attributeType == MIOPEN_TYPE_INT64 && elementCount == 4)
```
CAHEK7: `elementCount == 4` is a part of backend api? It looks like hardcoded 4D tensor.
The same is for `MIOPEN_ATTR_TENSOR_STRIDES` case.

In any case I'd like to generalize this check for dimensions and strides and in case of any changes it will be updated simultaneously.

src/backend_api.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+miopenStatus_t BackendTensorDescriptor::Finalize()
+{
+    if(!m_UniqueIdSet || !m_DataTypeSet || !m_ByteAlignmentSet || !m_DimensionsSet ||
+       !m_StridesSet || (m_VectorCountSet && !m_VectorizedDimensionSet))
+    {
+        return miopenStatusBadParm;
+    }
+
+    if(m_Dimensions.size() != 4 || m_Strides.size() != 4)
```
CAHEK7: Right now it is hardcoded to be 4d, should it be handled in a more common case?

src/backend_api.cpp
amberhassaan: I did look into it. The backend API does not limit to 4d or 5d. They can be any positive number, I believe. Similarly, any dimension can be vectorized as opposed to just `channel` dim.

scerzh: CUDNN library's possible values are from 3 to 8: https://docs.nvidia.com/deeplearning/cudnn/archives/cudnn-897/developer-guide/index.html#tensor-descriptor
MIOPEN's are from 1 to 5.
The layout checks are done by TensorDescriptor, so we can omit them here.

src/backend_api.cpp
scerzh: will remove it and rely on TensorDescriptor's layout checking

src/include/miopen/backend_api.hpp
```diff
@@ -0,0 +1,62 @@
...
+                              miopenBackendAttributeType_t attributeType,
+                              int64_t elementCount,
+                              void* arrayOfElements) override;
+    virtual void finalize() override;
+    virtual void getAttribute(miopenBackendAttributeName_t attributeName,
+                              miopenBackendAttributeType_t attributeType,
+                              int64_t requestedElementCount,
+                              int64_t* elementCount,
+                              void* arrayOfElements) override;
+    virtual void execute(miopenHandle_t handle, miopenBackendDescriptor_t varianPack) override;
```
amberhassaan: minor: s/varianPack/variantPack/

src/include/miopen/backend_api_builders.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <miopen/miopen.h>
+#include <miopen/tensor.hpp>
+#include <vector>
+
+namespace miopen {
+
+namespace backend_api {
+
+class TensorDescriptorEx : public TensorDescriptor
```
amberhassaan: Seems like a good idea. Eventually we might fold it into the original TensorDescriptor class. Something to keep in mind.

src/backend_api.cpp
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    return miopen::try_([&] {
+        auto& outputDescriptor = miopen::deref(descriptor);
+        void* address          = descriptor;
+
+        switch(descriptorType)
+        {
+        case MIOPEN_BACKEND_TENSOR_DESCRIPTOR:
+            if(std::align(alignof(miopen::backend_api::BackendTensorDescriptor),
```
amberhassaan: good

src/backend_api.cpp
```diff
@@ -0,0 +1,328 @@
...
+                          sizeof(miopen::backend_api::BackendTensorDescriptor),
+                          address,
+                          sizeInBytes) &&
+               address == descriptor)
+            {
+                new(std::launder(&outputDescriptor)) miopen::backend_api::BackendTensorDescriptor();
```
amberhassaan: good. I didn't know `std::launder` existed.

src/backend_api.cpp
```diff
@@ -0,0 +1,328 @@
...
+        },
+        false);
+}
+
+extern "C" miopenStatus_t miopenBackendDestroyDescriptor(miopenBackendDescriptor_t descriptor)
+{
+    return miopen::try_([&] { miopen_destroy_object(descriptor); }, false);
+}
+
+extern "C" miopenStatus_t miopenBackendInitialize(miopenBackendDescriptor_t descriptor,
```
amberhassaan: I do not like this API and not keen on supporting it but your implementation seems right.

src/backend_api.cpp
```diff
@@ -0,0 +1,328 @@
...
+                                                  miopenBackendDescriptorType_t descriptorType,
+                                                  size_t sizeInBytes)
+{
+    MIOPEN_LOG_FUNCTION(descriptorType, sizeInBytes);
+
+    if(!descriptor)
```
amberhassaan: now `descriptor` is a `struct` and not a pointer.

src/backend_api.cpp
```diff
@@ -0,0 +1,328 @@
...
+    {
+        return miopenStatusBadParm;
+    }
+
+    return miopen::try_([&] {
+        auto& outputDescriptor = miopen::deref(descriptor);
+        void* address          = descriptor;
```
amberhassaan: same as above.

src/backend_api.cpp
```diff
@@ -0,0 +1,328 @@
...
+BackendDescriptor::~BackendDescriptor() {}
+
+BackendTensorDescriptor::BackendTensorDescriptor()
+    : mBuilder(new TensorBuilder()), mDescriptor(nullptr)
+{
+}
+
+BackendTensorDescriptor::~BackendTensorDescriptor()
+{
+    delete mBuilder;
```
amberhassaan: good practice to set a pointer to `nullptr` after `delete`. This avoids double deletion.

src/backend_api.cpp
```diff
@@ -0,0 +1,328 @@
...
+}
+
+namespace miopen {
+
+namespace backend_api {
+
+BackendDescriptor::~BackendDescriptor() {}
+
+BackendTensorDescriptor::BackendTensorDescriptor()
+    : mBuilder(new TensorBuilder()), mDescriptor(nullptr)
```
amberhassaan: We may need to review this design decision. For example, we could use `std::unique_ptr`, or go pointer-less. In pointer-less design, we `move` the builder into the descriptor by making `build` an `r-value` method like this: 
```
TensorDescriptorEx build() && { ... }
``` 
and then invoking as:
```
auto desc = std::move(builder).build();
```
The pros are that we don't have to worry about pointers. Cons are that an empty builder still takes up space. Let's stay with your current design but at the same time think about potential issues.

src/backend_api.cpp
```diff
@@ -0,0 +1,328 @@
...
+                                           int64_t requestedElementCount,
+                                           int64_t* elementCount,
+                                           void* arrayOfElements)
+{
+    if(!mFinalized)
+    {
+        MIOPEN_THROW(miopenStatusNotInitialized);
+    }
+
+    const auto& dimensions = mDescriptor->GetLengths();
```
amberhassaan: minor: move declaration closer to use.

amberhassaan (commented):
@scerzh : overall I'm happier with the new `Builder` based design. Some more comments. When you pass CI, you can remove the `Draft` label and open it up for other reviewers.

src/backend_api.cpp
scerzh: these are due to declarations are not allowed inside a switch()

src/backend_api.cpp
scerzh: `descriptor` here is the function's argument which is a pointer.
image

src/backend_api.cpp
amberhassaan: Can't you get around that problem by putting the `case` body inside `{}`?

scerzh: You are right. There are existing {} already that can be used.

src/backend_api.cpp
scerzh: Also we can consider std::optional with the same pros and cons.
And we can consider std::variant for both objects

src/include/miopen/backend_api_builders.hpp
```diff
@@ -0,0 +1,56 @@
+#pragma once
+
+#include <miopen/miopen.h>
+#include <miopen/tensor.hpp>
+#include <memory>
+#include <vector>
+
+namespace miopen {
+
+namespace backend_api {
```
amberhassaan: optional suggestions: 
rename `backend_api` to `backend` or `backendapi`.

src/include/miopen/backend_api.hpp
```diff
@@ -0,0 +1,64 @@
+#pragma once
+
+#include <miopen/miopen.h>
+#include <miopen/tensor.hpp>
+#include <miopen/backend_api_builders.hpp>
```
amberhassaan: rename this file to something like `backend_tensor.hpp` and include this header file `backend_api.hpp` in `backend_tensor.hpp`. This to me is the more logical `include order`. 
Then in `backend_api.cpp` you can include both the header files.

amberhassaan (approved):
Please address the include order comment and then it's good to go by me. I'd like @CAHEK7 to review and approve when he comes back. Thanks.

src/backend_api.cpp
amberhassaan: my bad. you're right.

src/include/miopen/backend_api.hpp
scerzh: agree

src/include/miopen/backend_api_builders.hpp
scerzh: Agree. In cudnn these APIs are called Graph API
image
Is it better to name it `graphapi`?

amberhassaan: yeah, I personally like `graphapi` and I don't like the name `backend` but seems like cudnn also chose wrong and named every enum and api with `BackendXYZ`. So, I say let's rename our internal `namespace` to `graphapi` and then we can think about what to do with the word `Backend` in the public API and enums. This entire API feels like a series of bad decisions ;).

scerzh: What about filenames? I suggest the following renames:
backend_api.hpp -> graphapi.hpp
backend_api.cpp -> graphapi.cpp
backend_tensor.hpp -> graphapi_tensor.hpp
backend_tensor.cpp -> graphapi_tensor.hpp

amberhassaan: I'm OK with that. @JehandadKhan : please provide your input on this.

CAHEK7 (approved):

junliume (approved):

junliume:
Ping @JehandadKhan for the last round of review

amberhassaan (approved):
@scerzh : minor convention related comment: namespace `miopen::graphapi` should match the directory structure `miopen/graphapi`, so we should put our header files and src files under a `graphapi` subdirectory. Sorry for thinking of this late but I suggest we do this going forward.

src/include/miopen/graphapi/graphapi.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#pragma once
+
+#include <miopen/miopen.h>
+#include <miopen/object.hpp>
+
```
atamazov, line 30: For int64_t
```suggestion
#include <miopen/object.hpp>

#include <cstdint>
```

src/CMakeLists.txt
```diff
@@ -87,6 +87,8 @@ set( MIOpen_Source
     activ_api.cpp
     api/find2_0_commons.cpp
     argmax_api.cpp
+    graphapi/graphapi.cpp
+    graphapi/graphapi_tensor.cpp
```
atamazov: Please keep this list in the sorted order

src/include/miopen/graphapi/graphapi_tensor.hpp
```diff
@@ -0,0 +1,189 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#pragma once
+
+#include <miopen/graphapi/graphapi.hpp>
+
+#include <vector>
```
atamazov, line 32:
```suggestion
#include <cstdint>
#include <vector>
```

test/gtest/graphapi_tensor.cpp
```diff
@@ -0,0 +1,204 @@
+#include <miopen/miopen.h>
```
atamazov, line 26: Please insert license header.

src/graphapi/graphapi.cpp
```diff
@@ -0,0 +1,157 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include <miopen/graphapi/graphapi.hpp>
+#include <miopen/graphapi/graphapi_tensor.hpp>
+#include <miopen/errors.hpp>
+#include <miopen/logger.hpp>
```
atamazov: Please sort this list.

atamazov (approved):
🟡  Looks almost good. Thanks!

src/CMakeLists.txt
atamazov: [resolved]

src/graphapi/graphapi.cpp
atamazov: [resolved]

src/include/miopen/graphapi/graphapi.hpp
atamazov: [resolved]

src/include/miopen/graphapi/graphapi_tensor.hpp
atamazov: [resolved]

test/gtest/graphapi_tensor.cpp
atamazov: [resolved]

atamazov (approved):
🟢 LGTM!

src/include/miopen/graphapi/graphapi.hpp
scerzh: I didn't notice because it was probably included indirectly. Done.

src/CMakeLists.txt
scerzh: done

src/include/miopen/graphapi/graphapi_tensor.hpp
scerzh: done

src/graphapi/graphapi.cpp
scerzh: done

test/gtest/graphapi_tensor.cpp
scerzh: done

src/graphapi/graphapi_tensor.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+Tensor TensorBuilder::build() &&
+{
+    if(!mUniqueIdSet || !mDataTypeSet || !mDimensionsSet || !mStridesSet ||
+       mTensor.getDimensions().size() != mTensor.getStrides().size())
+    {
+        MIOPEN_THROW(miopenStatusBadParm);
+    }
+
+    return std::move(mTensor);
```
amberhassaan: Better to not use `std::move` here and let the compiler pick.

src/include/miopen/graphapi/graphapi_tensor.hpp
```diff
@@ -0,0 +1,190 @@
...
+public:
+    TensorBuilder& setDataType(miopenDataType_t dataType) &;
+    TensorBuilder& setDim(const std::vector<int64_t>& dimensions) &;
+    TensorBuilder& setDim(std::vector<int64_t>&& dimensions) &;
+    TensorBuilder& setStride(const std::vector<int64_t>& strides) &;
+    TensorBuilder& setStride(std::vector<int64_t>&& strides) &;
+    TensorBuilder& setId(int64_t id) &;
+    TensorBuilder& setVirtual(bool isVirtual) &;
+
+    TensorBuilder&& setDataType(miopenDataType_t dataType) &&
```
amberhassaan, line 103: setters should not return an `r-value` reference or be marked as `r-value` methods.

src/include/miopen/graphapi/graphapi_tensor.hpp
```diff
@@ -0,0 +1,190 @@
...
+#include <miopen/graphapi/graphapi.hpp>
+
+#include <cstdint>
+#include <vector>
+
+namespace miopen {
+
+namespace graphapi {
+
+class Tensor
```
amberhassaan, line 38: Previously, this class extended from the existing `TensorDescriptor`. Why are we no longer doing that?

src/include/miopen/graphapi/graphapi_tensor.hpp
```diff
@@ -0,0 +1,190 @@
...
+    {
+    }
+
+    miopenDataType_t getDataType() const noexcept { return mDataType; }
+    const std::vector<int64_t>& getDimensions() const noexcept { return mDimensions; }
+    const std::vector<int64_t>& getStrides() const noexcept { return mStrides; }
+    int64_t getId() const noexcept { return mId; }
+    bool getVirtual() const noexcept { return mVirtual; }
+
+    Tensor& setDataType(miopenDataType_t dataType) noexcept
```
amberhassaan: we don't need `setters` here because that's what the builder is for.

src/include/miopen/graphapi/graphapi_tensor.hpp
scerzh: this is for preserving r-value references in sequences like 
`auto result = Builder().setSomething().setSomethingElse().setAnotherSomething().build();`

src/graphapi/graphapi_tensor.cpp
scerzh: mTensor is an l-value here and won't be moved without std::move.
copy-elision eliminates additional creation; mTensor was created already and its creation cannot be eliminated.

without std::move the object to be returned will always be copy-constructed.

src/include/miopen/graphapi/graphapi_tensor.hpp
scerzh: I saw a TensorParameters in cudnn_frontend class with these capabilities. I think we will be able to reuse these setters.

src/include/miopen/graphapi/graphapi_tensor.hpp
amberhassaan: cool.

src/include/miopen/graphapi/graphapi_tensor.hpp
amberhassaan: Usually setters open the door to creating a logical invalid state in an object so I recommend we don't open that door and everyone goes through the builder route.

src/graphapi/graphapi_tensor.cpp
amberhassaan: seems like you're right.

src/include/miopen/graphapi/graphapi_tensor.hpp
scerzh: OK makes sense

src/include/miopen/graphapi/graphapi_tensor.hpp
scerzh: Added `operator miopen::TensorDescriptor()`

src/include/miopen/graphapi/graphapi_tensor.hpp
scerzh: Done

src/include/miopen/graphapi/graphapi_tensor.hpp
```diff
@@ -0,0 +1,167 @@
...
+          mStrides(std::move(strides)),
+          mId(id),
+          mDataType(dataType),
+          mVirtual(isVirtual)
+    {
+    }
+
+    operator miopen::TensorDescriptor() const
+    {
+        return {mDataType,
```
amberhassaan, line 73: one minor but important detail is to specify `layout` type to the TensorDescriptor. We should perhaps store the `layout` type with `Tensor` class too, as most of our performant kernels use `NHWC` while the default layout is `NCHW`.

scerzh: Layout is not provided on tensor creation in Graph API. I think this is not a topic of this PR.

JehandadKhan:
@junliume  Can we please merge this PR, since a lot of other PRs depend on it.

src/graphapi/graphapi_tensor.cpp
atamazov: [Quality] @CAHEK7 @amberhassaan @scerzh Please notice that std:move changes the state of mTensor to the moved-from state, which makes it unusable. The above means that after `build()` the object should not be used anymore. Do we guarantee/enforce that? Maybe it is worth adding the ~IsBuilt or~ `IsUnusable` flag and check it where appropriate?

atamazov (commented):
Potential problem? https://github.com/ROCm/MIOpen/labels/urgency_normal

src/graphapi/graphapi_tensor.cpp
scerzh: There are 2 overloads of `build()`. The declaration of this one is enough here.
```C++
Tensor TensorBuilder::build() &&
```
it is making the whole builder object unusable: you should not access the builder after it, including the proposed call of `IsUnusable()`; cppcheck or similar analyzers should trigger errors if you do. For the same reason STL doesn't have `IsUnusable()`-like methods for its containers.

amberhassaan: @atamazov : the `isFinalized()` method is what you want from `isUsable()`. Once the object has been finalized, the builder is no longer there. We have thought of replacing the builder with a `std::optional<builder-type>` and marking the std::optional false after `Finalize()`. And we have discussed keeping a pointer to `builder` and making it null. I think all are more or less equivalent solutions.

CAHEK7: > There are 2 overloads of `build()`. The declaration of this one is enough here.
> 
> ```c++
> Tensor TensorBuilder::build() &&
> ```
> 
> it is making the whole builder object unusable: you should not access the builder after it, including the proposed call of `IsUnusable()`; cppcheck or similar analyzers should trigger errors if you do. For the same reason STL doesn't have `IsUnusable()`-like methods for its containers.

@atamazov  @amberhassaan  @scerzh 
I would argue about that.
`std::move` itself does not do anything, it just forces r-value semantic to be called by static cast to r-value reference. As well as there is no officially defined "moved-from" state for the object.
Move semantic is broken is C++ because it does not force any specific behavior after moving - the only requirement is to left the object in a valid for destruction state, but you must check owning resource state in the destructor to identify do you own it or it has been moved, so you must be able to call all the appropriate functions to do this.

It's up to user-defined move operators to decide the state of the moved object and for example for `std::unique_ptr` the state after move is totally defined. 

So adding `bool mIsUnusable` and throw an exception from the `build()&&` if the object has already been moved or not yet properly constructed does not contradict any specification or standard. At least we will have this behavior defined. It just must not be a `isUsable()` method, the check should be done inside the `build()&&` method.

atamazov: @amberhassaan @scerzh @CAHEK7 Yeah, the idea is like to set the flag and throw an exception in all methods of the class if it is set (except dtor).

scerzh: OK, I see the value in the idea.

We don't need an additional flag. We already have flags causing `build()&` and `build()&&` to throw if something required wasn't set. We can reset the state of the builder to initial.

Issue #2752, opened 2024-02-15
[hipRTC] MIOpen Custom Limits missing members such as lowest()


junliume:
**[Observation]**:
Several ONNX UT has failed with the message
```
/tmp/comgr-01aacb/include/data_type.hpp:1025:86: error: no member named 'lowest' in 'std::numeric_limits<float>'
 1025 |     __host__ __device__ static constexpr T Lowest() { return std::numeric_limits<T>::lowest(); }
      |                                                              ~~~~~~~~~~~~~~~~~~~~~~~~^
/tmp/comgr-01aacb/include/reduction_operator.hpp:87:84: note: in instantiation of member function 'ck::NumericLimits<float>::Lowest' requested here
   87 |     __device__ static constexpr T GetReductionZeroVal() { return NumericLimits<T>::Lowest(); };
      |                                                                                    ^
/tmp/comgr-01aacb/include/gridwise_generic_2d_reduction_direct_threadwise.hpp:85:40: note: in instantiation of member function 'ck::reduce::Max<float>::GetReductionZeroVal' requested here
   85 |         const auto zeroVal = opReduce::GetReductionZeroVal();
      |                                        ^
```
It is requested here:
https://github.com/ROCm/MIOpen/blob/b44f182b731532b5c32eb37a2df16b18efe4d94f/src/composable_kernel/composable_kernel/include/utility/data_type.hpp#L1025

**[More Discussion]**:
@atamazov @JehandadKhan  we cannot just re-invent all variations of member functions for all variations of datatypes on our own, right? 
https://en.cppreference.com/w/cpp/types/numeric_limits

**[Related PRs]**:
https://github.com/ROCm/MIOpen/pull/2637
https://github.com/ROCm/MIOpen/pull/2705

here is just a few examples (not a complete list)
- https://github.com/ROCm/MIOpen/blob/b44f182b731532b5c32eb37a2df16b18efe4d94f/test/check_numerics_test.cpp#L146
- https://github.com/ROCm/MIOpen/blob/b44f182b731532b5c32eb37a2df16b18efe4d94f/test/check_numerics_test.cpp#L152
- https://github.com/ROCm/MIOpen/blob/b44f182b731532b5c32eb37a2df16b18efe4d94f/test/rnn_seq_api.hpp#L1472
- https://github.com/ROCm/MIOpen/blob/b44f182b731532b5c32eb37a2df16b18efe4d94f/test/soft_max.cpp#L116
[...]

PR #2753, opened 2024-02-15
[hipRTC] add lowest() for float to MIOpen custom limits

junliume:
#2752

### DIFF
--- a/src/kernels/miopen_limits.hpp
+++ b/src/kernels/miopen_limits.hpp
@@ -47,7 +47,9 @@ public:
 
     static constexpr __device__ float min() noexcept { return 0x1p-126f; }
 
+#if HIP_PACKAGE_VERSION_FLAT >= 6001024024ULL
     static constexpr __device__ float lowest() noexcept { return -0x1.fffffep+127; }
+#endif
 };
 
 template <>
--- a/src/kernels/miopen_limits.hpp
+++ b/src/kernels/miopen_limits.hpp
@@ -46,6 +46,8 @@ public:
     static constexpr __device__ float max() noexcept { return 0x1.FFFFFEp+127f; }
 
     static constexpr __device__ float min() noexcept { return 0x1p-126f; }
+
+    static constexpr __device__ float lowest() noexcept { return -0x1.fffffep+127; }
 };
 
 template <>


src/kernels/miopen_limits.hpp
```diff
@@ -46,6 +46,8 @@ class numeric_limits<float>
     static constexpr __device__ float max() noexcept { return 0x1.FFFFFEp+127f; }
 
     static constexpr __device__ float min() noexcept { return 0x1p-126f; }
+
+    static constexpr __device__ float lowest() noexcept { return -0x1.fffffep+127; }
```
junliume, line 51: This line *might* need to be protected by `HIP_PACKAGE_VERSION_FLAT` but let's see if it works as is.

junliume: Added anyway since the earliest trace of issue comes together with the one below on int

JehandadKhan (approved):

src/kernels/miopen_limits.hpp
atamazov: @junliume It is highly unlikely that in some HIP release (or release candidate) we have PCH with `lowest()`, while `min()` and `max()` are missing there. Therefore I recommend removing the conditional wrapper (`#if HIP_PACKAGE_VERSION_FLAT >= 6001024024ULL` and `#endif`).

amberhassaan:
@junliume , @atamazov : slightly off topic but what do you guys think of using constants from `<cfloat>` (https://en.cppreference.com/w/cpp/header/cfloat) which is what is used to define `numeric_limists` (see further down the page (https://en.cppreference.com/w/cpp/types/numeric_limits). This way we don't have to maintain constants ourselves?

atamazov:
@amberhassaan (Maybe I've misunderstood your idea, but) hipRTC does not allow any standard headers to be used in compiled programs, including `<cfloat>`. See https://reviews.llvm.org/D100652 for more info.

amberhassaan:
> @amberhassaan (Maybe I've misunderstood your idea, but) hipRTC does not allow any standard headers to be used in compiled programs, including `<cfloat>`. See https://reviews.llvm.org/D100652 for more info.

No, you're right. I did think of it after typing my comment. I was hoping that we can slip in some simple C header that just defines some constants but technically we can't rely on that. The only other option would be to embed these system headers with MIOpen the way we do for user-defined include files.

PR #2762, opened 2024-02-19
Upgrade class TmpDir

apwojcik:
This is the PR in a series to fix tests on Windows.

A few improvements to the `TmpDir` class and integration with the `fs::path` class. Additionally, on Windows, the first attempt to remove the directory might not be successful. Therefore, the `TmpDir` class is retrying removal up to 5 times.


johnnynunez:
@junliume I've seen that AMDMIGraphx will have support for windows os in rocm 6.1. MiOpen too?

atamazov:
@apwojcik 
> A few improvements to the `TmpDir` class and integration with the `fs::path` class.

Can you please elaborate improvements?

> Additionally, on Windows, the first attempt to remove the directory might not be successful. Therefore, the `TmpDir` class is retrying removal up to 5 times.

I guess this peculiarity is described somewhere in Windows docs. Can you please provide a link?

apwojcik:
> Can you please elaborate improvements?
* use `std::string_view` instead of copying `std:string` in the constructor and methods (90% of use cases are literals),
* prefer move semantics instead of deleting the copy constructor and assignment operator,
* integration with `fs::path` for easy path construction under the temp directory
* automatically casting to `fs::path` and `std::string` when a path or string is needed (e.g., to pass as an argument to a function - `fs::path` does conversion to `std::stirng` automatically* on Linux, but not on Windows)
* Execute returns exit code of a child process for the upper layer to decide how to proceed
* There was a minor bug in the constructor - for empty prefixes, the resulting name of the directory was `"***--***"` instead of `"***-***"`.

All the above is used by upcoming changes (PRs) to fix MIOpen execution on Windows.

> I guess this peculiarity is described somewhere in Windows docs. Can you please provide a link?

Depending on the implementation, one of these functions is used underneath (or both)
* https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-removedirectoryw
* https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-deletefilew

\* GNU standard library has overloaded methods accepting both `fs::path` and `std::string`, MSVC standard library does not most cases.

averinevg:
> Additionally, on Windows, the first attempt to remove the directory might not be successful. Therefore, the `TmpDir` class is retrying removal up to 5 times.

@apwojcik looks like it doesn't work very reliably :smile:. It might be worth using this approach only for Windows, and note that this is a workaround.

apwojcik:
This is not a workaround. I can make it only for Windows. However, it is better to maintain a common code base. The change is not breaking Linux, and it does not introduce extra latency - it will only fall into the `catch` block if, e.g., the process is trying to remove a directory and does not have proper privileges (which is not a case here).

averinevg:
> This is not a workaround. I can make it only for Windows. However, it is better to maintain a common code base. The change is not breaking Linux, and it does not introduce extra latency - it will only fall into the `catch` block if, e.g., the process is trying to remove a directory and does not have proper privileges (which is not a case here).

@apwojcik I think that Windows problems are better left on the Windows side. At the same time, it would be nice to leave a comment on what caused such a strange trick (and describe not only what is happening, but also why). This would be useful for the next person who wants to change this code.

atamazov:
@apwojcik Thanks for the explanations.

> This is not a workaround. I can make it only for Windows. 

I am afraid this is Windows-specific W/A, because `filesystem::remove_all()` does not say anything about possibility of failures that can be fixed by simple delay+repetition. Ditto DeleteFileW and RemoveDirectoryW.

atamazov:
@apwojcik 
> * automatically casting to ... `std::string` when a path or string is needed.

I am not sure if automatically casting file paths to `std::string` is really useful. I remember a case where the lack of such an automatic cast forced me to replace `std::string` with `fs::path`, and that ended up simplifying the code elsewhere.

apwojcik:
> @apwojcik
> 
> > * automatically casting to ... `std::string` when a path or string is needed.
> 
> I am not sure if automatically casting file paths to `std::string` is really useful. I remember a case where the lack of such an automatic cast forced me to replace `std::string` with `fs::path`, and that ended up simplifying the code elsewhere.

Agree. This operator is temporary. Otherwise, I would have to add more changes that are not closely related to this PR. I prefer to keep it minimal. I already have the PR changing `std::string` to use `fs::path`, which will remove that overload, too.

apwojcik:
@junliume @atamazov @averinevg, Can we move forward with this PR? I have a list of PRs waiting to be merged after this one.

junliume (approved):

atamazov (commented):
No objections. Sorry for delaying the review.

PR #2764, opened 2024-02-20
[HIPRTC] Provide option to add/remove include directories to/from compiler flags

raramakr:
The hip compiler flags are getting embedded in MIOpen shared library and the isystem include directories in the compiler flags are hard coded paths. For the ROCm use case, build scripts will set the option to OFF, so that include directories will not be added to compiler flags. This will help in removing the hard coded path from the library By default the option is set to ON.


CMakeLists.txt
```diff
@@ -215,6 +215,7 @@ endif()
 # HIP is always required
 find_package(hip REQUIRED PATHS /opt/rocm)
 message(STATUS "Build with HIP ${hip_VERSION}")
+option(APPEND_INTERFACE_INCLUDE_DIRECTORIES "Append include directories to compiler flags" ON)
```
atamazov: Prefix with "MIOPEN". Make the option name more clear.
```suggestion
option(MIOPEN_HIP_COMPILER_USE_SYSTEM_INCLUDE_DIRECTORIES "Append include directories to compiler flags" ON)
```

atamazov: This option should be automatically set OFF when hipRTC compiler is used (i.e. when `MIOPEN_USE_HIPRTC` is set to ON). Use `set_var_to_condition` to conditionally set the default value (the technique is described at https://github.com/ROCm/MIOpen/pull/2694#discussion_r1482098759)

atamazov (changes requested):

atamazov:
@raramakr Please prepend PR's title with [HIPRTC].

raramakr:
Thank you @atamazov for the review comment. Will  update the patch based on the comments

CMakeLists.txt
raramakr: Updated the patch with the suggested name

raramakr: Done

junliume:
@atamazov @JehandadKhan if no objections, we will merge it when you approve.

CMakeLists.txt
```diff
@@ -261,6 +252,22 @@ set(MIOPEN_hip_VERSION ${MIOPEN_hip_VERSION_MAJOR}.${MIOPEN_hip_VERSION_MINOR}.$
 set_var_to_condition(MIOPEN_USE_HIPRTC_DEFAULT (MIOPEN_USE_COMGR AND (MIOPEN_hip_VERSION VERSION_GREATER_EQUAL 5)))
 option(MIOPEN_USE_HIPRTC "Use HIPRTC to build HIP kernels instead of COMGR" ${MIOPEN_USE_HIPRTC_DEFAULT})
 
+# Do not append system include directories to HIP compiler flags when HIPRTC is used
+set_var_to_condition(MIOPEN_HIP_COMPILER_USE_SYSTEM_INCLUDE_DIRECTORIES_DEFAULT
+              (NOT (MIOPEN_USE_HIPRTC AND (MIOPEN_hip_VERSION VERSION_GREATER_EQUAL 6.1.40091))))
```
atamazov, line 257: `6.1.40091` seems questionable. IIRC the patch version format is yywwd, which means that this is 6.1 built at Sunday week 09 year 2040. As a result, system includes will be always used until 6.2 (please check am I correct or not)

If we want to disable system includes starting from 6.1, then let's use "6.1.0".

CMakeLists.txt
```diff
@@ -261,6 +252,22 @@ set(MIOPEN_hip_VERSION ${MIOPEN_hip_VERSION_MAJOR}.${MIOPEN_hip_VERSION_MINOR}.$
...
+target_flags(HIP_COMPILER_FLAGS hip::device)
+# Remove cuda arch flags
+string(REGEX REPLACE --cuda-gpu-arch=[a-z0-9]+ "" HIP_COMPILER_FLAGS "${HIP_COMPILER_FLAGS}")
+string(REGEX REPLACE --offload-arch=[a-z0-9:+-]+ "" HIP_COMPILER_FLAGS "${HIP_COMPILER_FLAGS}")
+# Skip library paths since hip will incorrectly treat it as a source file
+string(APPEND HIP_COMPILER_FLAGS " ")
+foreach(_unused RANGE 2)
+    string(REGEX REPLACE " /[^ ]+\\.(a|so) " " " HIP_COMPILER_FLAGS "${HIP_COMPILER_FLAGS}")
+endforeach()
+
```
atamazov, line 270: This change seems unnecessary. Please move this block of code to its original location.

atamazov (changes requested):

CMakeLists.txt
raramakr: we are relying on hip_version rather than rocm release version.

The hip_version.h file  from 6.1.0 RC2  build shows the version as below. So hip version corresponding to 6.1.0 is 6.1.40091
#define HIP_VERSION_MAJOR 6
#define HIP_VERSION_MINOR 1
#define HIP_VERSION_PATCH 40091

MIOPEN_hip_VERSION is populated as below. 
set(MIOPEN_hip_VERSION ${MIOPEN_hip_VERSION_MAJOR}.${MIOPEN_hip_VERSION_MINOR}.${MIOPEN_hip_VERSION_PATCH})

In [6.0.2 RC1 GA] the corresponding version is 6.0.24015

CMakeLists.txt
raramakr: The HIP_COMPILER_FLAGS are populated by calling target_flags(HIP_COMPILER_FLAGS hip::device) , which  will add system include directories
But the condition to add/skip-adding  system include directories should be set before calling the target_flags. That is the reason for moving the code block

CMakeLists.txt
raramakr: I can check with HIP team on this - 40091 (YearWeekDay)
But the 6.1.0 release is giving the same and that is the reason for using it

atamazov: @raramakr yes, please check, maybe they changed the format due to some reason. 

[resolved]

CMakeLists.txt
atamazov: @raramakr thanks

[resolved]

atamazov (approved):
🟢 LGTM

CMakeLists.txt
raramakr: Sure @atamazov and thank you for reviewing the patch
If any changes in hip_VERSION , will upload a separate patch to update the same

junliume (approved):

CMakeLists.txt
atamazov: @raramakr 🔴 hip_VERSION should maintain backward compatibility. The goal is that any new version of MIOpen works with some previous version of HIP correctly. If HIP guys change the format, and MIOpen adapts to it, then MIOpen may loose the capability to with with older HIP. So the question is important.

**_Can you please check with HIP team if they changed the format or not._** So far, I am assuming that `40091` is just a glitch. It should not lead to any problems, I guess, unless they will release ROCm 6.1.1 with "correct" HIP version like 6.1.24xxx (but this can be manually fixed).

/cc @junliume

raramakr: sure @atamazov

raramakr: @atamazov @junliume, From hip team
The format of the 5 digit is :- YdddX.
Y = last digit of the year. So 4 for 2024. 9 for 2029.
ddd = day in year. So goes 001 to 366.
X = is initially set to 0 on every mainline promotion. Increments by 1 if we ever make a cherry-pick to the point release.

In 6.0.2 the patch version is 32831, but cmake is reporting hip_VERSION as 6.0.24015 (24015- HIP_VERSION_GITDATE)
In 6.1.0 the patch version is 40091 and cmake is reporting hip_VERSION  as  6.1.40091. (40091- HIP_VERSION_PATCH)
There was change in 6.1 for HIP patch version reported via CMake , and that is why the difference in behavior
-    VERSION "${HIP_VERSION_MAJOR}.${HIP_VERSION_MINOR}.${**HIP_VERSION_GITDATE**}" (6.0)
+    VERSION "${HIP_VERSION_MAJOR}.${HIP_VERSION_MINOR}.${**HIP_VERSION_PATCH**}".   (6.1)

atamazov: @raramakr [sarcasm to HIP guys] So what we will do on the 2029/2030 tripping point? Let's have another kind of Y2K problem each 10 years instead of 100 years, right? 🤣

atamazov: @raramakr Seriously, thanks for finding this out and informing us 🤝

atamazov: [resolved] But please do not hide this review comment.

raramakr: Since the comparison is between versions(Major.Minor.patch) of each releases,  the patch version should not be an issue, since major and minor will change with each release

junliume: > @raramakr [sarcasm to HIP guys] So what we will do on the 2029/2030 tripping point? Let's have another kind of Y2K problem each 10 years instead of 100 years, right? 🤣

Thanks @raramakr for sharing! This has puzzled me for a long time :) After year of 2029 the major version would be very different then.

PR #2770, opened 2024-02-24
[HotFix] fix failed error bugs in conv backward weight solvers

iq136boy:
Fix the bug: the numerical error is larger than threshold in the ck-based backward weight convolution. 

MIOpen/build# ./bin/MIOpenDriver conv -n 1 -c 64 -H 130 -W 130 -k 3 -y 3 -x 3 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1 -F 4
MIOpenDriver conv -n 1 -c 64 -H 130 -W 130 -k 3 -y 3 -x 3 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -t 1 -F 4
PRNG seed: 12345678
MIOpen Backward Weights Conv. Algorithm: 5, Solution: 156/ConvHipImplicitGemmGroupWrwXdlops
GPU Kernel Time Backward Weights Conv. Elapsed: 0.005003 ms (average)
stats: bwdw-conv3x3u1, 1, 64, 128, 128, 3, 3, 3,  56623104, 0, 0, 11319, 0, 0.005003
Backward Convolution Weights FAILED: 0.892519 > 0.01

### DIFF
--- a/src/include/miopen/solver/implicitgemm_ck_util.hpp
+++ b/src/include/miopen/solver/implicitgemm_ck_util.hpp
@@ -29,6 +29,7 @@
 #include <miopen/conv/data_invoke_params.hpp>
 #include <miopen/conv/wrw_invoke_params.hpp>
 #include <miopen/batched_transpose_sol.hpp>
+#include <miopen/tensor_ops.hpp>
 
 #if MIOPEN_USE_COMPOSABLEKERNEL
 #include <ck/utility/data_type.hpp>
@@ -96,6 +97,25 @@ bool IsCKApplicable(const ProblemDescriptionType& problem)
         ptrs.begin(), ptrs.end(), [&args](auto& ptr) { return args.IsSupportedBy(ptr); });
 }
 
+#if MIOPEN_BACKEND_HIP
+inline void ProfilingRecordStart(const Handle& handle, HipEventPtr& start, HipEventPtr& stop)
+{
+    start = make_hip_event();
+    stop  = make_hip_event();
+    hipEventRecord(start.get(), handle.GetStream());
+}
+
+inline void ProfilingRecordStop(const Handle& handle, HipEventPtr& start, HipEventPtr& stop)
+{
+    hipEventRecord(stop.get(), handle.GetStream());
+    hipEventSynchronize(stop.get());
+    float mS = 0.0f;
+    hipEventElapsedTime(&mS, start.get(), stop.get());
+    handle.ResetKernelTime();
+    handle.AccumKernelTime(mS);
+}
+#endif
+
 template <typename DeviceOpType,
           typename CKArgsType,
           typename CastType,
@@ -114,15 +134,36 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
     }
 
     ConvSolution result;
-    result.invoker_factory =
-        [ck_args     = CKArgsType{problem},
-         sh_conv_ptr = std::shared_ptr{std::move(*ptr_iter)}](const std::vector<Kernel>&) mutable {
-            return [ck_args = std::move(ck_args), sh_conv_ptr = std::move(sh_conv_ptr)](
-                       const Handle& handle, const AnyInvokeParams& primitive_parameters) {
-                const auto& data_ctx = primitive_parameters.CastTo<CastType>();
-                auto argument_ptr    = ck_args.MakeArgPtr(sh_conv_ptr, data_ctx.tensors);
-                auto invoker_ptr     = sh_conv_ptr->MakeInvokerPointer();
+    result.invoker_factory = [ck_args     = CKArgsType{problem},
+                              sh_conv_ptr = std::shared_ptr{std::move(*ptr_iter)}](
+                                 const std::vector<Kernel>&) mutable {
+        return [ck_args = std::move(ck_args), sh_conv_ptr = std::move(sh_conv_ptr)](
+                   const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            const auto& data_ctx = primitive_parameters.CastTo<CastType>();
+            auto argument_ptr    = ck_args.MakeArgPtr(sh_conv_ptr, data_ctx.tensors);
+            auto invoker_ptr     = sh_conv_ptr->MakeInvokerPointer();
 
+            if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
+            {
+                auto zero           = 0.0f;
+                const auto& tensors = data_ctx.tensors;
+#if MIOPEN_BACKEND_HIP
+                HipEventPtr start = nullptr;
+                HipEventPtr stop  = nullptr;
+                if(handle.IsProfilingEnabled())
+                {
+                    ProfilingRecordStart(handle, start, stop);
+                }
+#endif
+                SetTensor(handle, tensors.dwDesc, tensors.dw, &zero);
+                invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), false});
+#if MIOPEN_BACKEND_HIP
+                if(handle.IsProfilingEnabled())
+                    ProfilingRecordStop(handle, start, stop);
+#endif
+            }
+            else
+            {
                 const auto enable_profiling = handle.IsProfilingEnabled();
                 float elapsed_time =
                     invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), enable_profiling});
@@ -131,8 +172,9 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
                     handle.ResetKernelTime();
                     handle.AccumKernelTime(elapsed_time);
                 }
-            };
+            }
         };
+    };
     return result;
 }
 
@@ -583,6 +625,11 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
             handle.ResetKernelTime();
 
             const auto& data_ctx = primitive_parameters.CastTo<CastType>();
+            if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
+            {
+                auto zero = 0.0f;
+                SetTensor(handle, data_ctx.tensors.dwDesc, data_ctx.tensors.dw, &zero);
+            }
 
             if(!data_ctx.workSpace)
             {
@@ -632,15 +679,33 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
                                                    tr_ptrs[0]->GetBufferPtr(),
                                                    tr_ptrs[1]->GetBufferPtr(),
                                                    tr_ptrs[2]->GetBufferPtr());
-            float conv_time   = 0;
-            conv_time += invoker_ptr->Run(argument_ptr.get(),
-                                          {handle.GetStream(), handle.IsProfilingEnabled()});
-
-            if(handle.IsProfilingEnabled())
+            if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
             {
-                handle.AccumKernelTime(conv_time);
+#if MIOPEN_BACKEND_HIP
+                HipEventPtr start = nullptr;
+                HipEventPtr stop  = nullptr;
+                if(handle.IsProfilingEnabled())
+                {
+                    ProfilingRecordStart(handle, start, stop);
+                }
+#endif
+                invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), false});
+#if MIOPEN_BACKEND_HIP
+                if(handle.IsProfilingEnabled())
+                    ProfilingRecordStop(handle, start, stop);
+#endif
             }
+            else
+            {
+                float conv_time = 0;
+                conv_time += invoker_ptr->Run(argument_ptr.get(),
+                                              {handle.GetStream(), handle.IsProfilingEnabled()});
 
+                if(handle.IsProfilingEnabled())
+                {
+                    handle.AccumKernelTime(conv_time);
+                }
+            }
             output_tr_inst.ConvertTo(handle, kernels, conv_tensors);
         };
     };
--- a/src/include/miopen/solver/implicitgemm_ck_util.hpp
+++ b/src/include/miopen/solver/implicitgemm_ck_util.hpp
@@ -644,8 +644,6 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
 
             output_init_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
 
-            /// \todo: Fix NHWC Wrw invokers to also issue a zero-out kernel. Will
-            /// need SetTensor() to properly zero out non-packed tensors
             if(output_tr_inst.GetConvOperandTag() == internal::ConvOperandTag::Weights)
             {
                 output_tr_inst.ZeroOutBuffer();
--- a/src/include/miopen/solver/implicitgemm_ck_util.hpp
+++ b/src/include/miopen/solver/implicitgemm_ck_util.hpp
@@ -98,7 +98,7 @@ bool IsCKApplicable(const ProblemDescriptionType& problem)
 }
 
 #define WORKAROUND_CK_ISSUE_1184 1
-#ifdef WORKAROUND_CK_ISSUE_1184
+#if WORKAROUND_CK_ISSUE_1184
 struct HipEventProfiler
 {
     const Handle& handle;
@@ -107,10 +107,7 @@ struct HipEventProfiler
     HipEventPtr stop;
 
     HipEventProfiler(const Handle& handle_)
-        : handle(std::move(handle_)),
-          event_time(0.0f),
-          start(make_hip_event()),
-          stop(make_hip_event())
+        : handle(handle_), event_time(0.0f), start(make_hip_event()), stop(make_hip_event())
     {
         hipEventRecord(start.get(), handle.GetStream());
     }
@@ -152,9 +149,7 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
                 auto argument_ptr    = ck_args.MakeArgPtr(sh_conv_ptr, data_ctx.tensors);
                 auto invoker_ptr     = sh_conv_ptr->MakeInvokerPointer();
                 {
-#ifdef WORKAROUND_CK_ISSUE_1184
                     HipEventProfiler pfr(handle);
-#endif
                     if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
                     {
                         auto zero           = 0.0f;
@@ -637,15 +632,14 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
                 std::swap(conv_tensors.x, conv_tensors.y);
                 std::swap(conv_tensors.xDesc, conv_tensors.yDesc);
             }
-#ifdef WORKAROUND_CK_ISSUE_1184
             HipEventProfiler pfr(handle);
-#endif
             input1_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
 
             input2_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
 
             output_init_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
 
+            /// \todo: Will need SetTensor() to properly zero out non-packed tensors
             if(output_tr_inst.GetConvOperandTag() == internal::ConvOperandTag::Weights)
             {
                 output_tr_inst.ZeroOutBuffer();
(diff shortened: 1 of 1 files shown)


junliume:
@zjing14 @JehandadKhan ping for review

junliume (approved):

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -583,6 +625,11 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
             handle.ResetKernelTime();
 
             const auto& data_ctx = primitive_parameters.CastTo<CastType>();
+            if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
+            {
+                auto zero = 0.0f;
+                SetTensor(handle, data_ctx.tensors.dwDesc, data_ctx.tensors.dw, &zero);
```
carlushuang: isn't this will clear zero of all the wrw solvers? which will result in perf regression

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -632,15 +679,33 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
...
-            conv_time += invoker_ptr->Run(argument_ptr.get(),
-                                          {handle.GetStream(), handle.IsProfilingEnabled()});
-
-            if(handle.IsProfilingEnabled())
+            if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
             {
-                handle.AccumKernelTime(conv_time);
+#if MIOPEN_BACKEND_HIP
+                HipEventPtr start = nullptr;
+                HipEventPtr stop  = nullptr;
```
carlushuang: can you explain why we need to add extra HipEventPtr variable `start`/`stop` considering we have these hipevent inside handler?
This PR looks like only move the hipevent and profiling related stuff out of handler, and how could it solve the precision issue of wrw kernel?

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -114,15 +134,36 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
-        [ck_args     = CKArgsType{problem},
-         sh_conv_ptr = std::shared_ptr{std::move(*ptr_iter)}](const std::vector<Kernel>&) mutable {
-            return [ck_args = std::move(ck_args), sh_conv_ptr = std::move(sh_conv_ptr)](
-                       const Handle& handle, const AnyInvokeParams& primitive_parameters) {
-                const auto& data_ctx = primitive_parameters.CastTo<CastType>();
-                auto argument_ptr    = ck_args.MakeArgPtr(sh_conv_ptr, data_ctx.tensors);
-                auto invoker_ptr     = sh_conv_ptr->MakeInvokerPointer();
+    result.invoker_factory = [ck_args     = CKArgsType{problem},
+                              sh_conv_ptr = std::shared_ptr{std::move(*ptr_iter)}](
+                                 const std::vector<Kernel>&) mutable {
```
atamazov: @iq136boy This is formatting-only change. The "Clang Format" CI stage should fail, but it didn't. Why?

atamazov:
@iq136boy Am I correct that this PR is intended to implement the following TODO?
https://github.com/ROCm/MIOpen/blob/a592725cbf84a6706456220ef287bfd485ca6320/src/include/miopen/solver/implicitgemm_ck_util.hpp#L615-L616

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: @carlushuang We need to zero out the wrw buffer, otherwise the result will be wrong by calling the CK kernel. But I do not need to call the setTensor here since the buffer has been zeroed out for the NCHW layout. I will fix it.

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: @atamazov To be honest, I have no idea. I have called clang-format to set the format for the modified file locally before I push it to the github. And the following clang-format command get empty set: 

clang-format-12 -style=file ../src/include/miopen/solver/implicitgemm_ck_util.hpp | diff - ../src/include/miopen/solver/implicitgemm_ck_util.hpp

Which means the "clang format" should be good?

iq136boy:
> @iq136boy Am I correct that this PR is intended to implement the following TODO?
> 
> https://github.com/ROCm/MIOpen/blob/a592725cbf84a6706456220ef287bfd485ca6320/src/include/miopen/solver/implicitgemm_ck_util.hpp#L615-L616

@atamazov Yes. That is correct.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -114,15 +134,36 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
+                {
+                    ProfilingRecordStart(handle, start, stop);
+                }
+#endif
+                SetTensor(handle, tensors.dwDesc, tensors.dw, &zero);
+                invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), false});
+#if MIOPEN_BACKEND_HIP
+                if(handle.IsProfilingEnabled())
+                    ProfilingRecordStop(handle, start, stop);
+#endif
```
atamazov: Please use approach 1 from https://github.com/ROCm/MIOpen/pull/2429#discussion_r1360795493 to accumulate time.

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: Ditto.

atamazov (changes requested):

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: @carlushuang  In the currently implementation, we use the profiling time that get directly from ck-api (ck invoker run fuction). When the ck-profiling is enabled, the ck wrw kernel introduces the errors that cause the precision issue. The ck-profiling should be set to false to avoid the precision issue but the time will be 0 in this case. I tried to utilize the handler to get the time. However, the current timer scheme in handler does not work properly, including approach 1 in this [comment](https://github.com/ROCm/MIOpen/pull/2429#discussion_r1360795493) Artem mentioned. 
So I proposed this workaround that using hipevent to get the correct time for the ck kernels. I will push the latest changes soon.

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: > Please use approach 1 from [#2429 (comment)](https://github.com/ROCm/MIOpen/pull/2429#discussion_r1360795493) to accumulate time.

Unfortunately, this approach does not work. I explained [here](https://github.com/ROCm/MIOpen/pull/2770#discussion_r1506089187)

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: @iq136boy Thanks for the explanation. Continued at https://github.com/ROCm/MIOpen/pull/2770#discussion_r1509099090.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -606,6 +643,12 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
                 std::swap(conv_tensors.xDesc, conv_tensors.yDesc);
             }
 
+            HipEventPtr start = nullptr;
+            HipEventPtr stop  = nullptr;
+            if(handle.IsProfilingEnabled())
+            {
+                ProfilingRecordStart(handle, start, stop);
+            }
```
atamazov: [Recommendation] Rework ProfilingRecordStart/ProfilingRecordStop into a class that holds HipEventPtrs and a reference to handle.
- Ctor should check IsProfilingEnabled() and start profiling, like ProfilingRecordStart().
- Dtor should check IsProfilingEnabled() and then stop profiling and write results into handle (just like ProfilingRecordStop()).

So all the benchmarking stuff in invoker could be reduced to construction of an object just before running the kernels and _implicit_ destruction of that object after the last kernel is enqueued.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -114,15 +132,33 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
+                   const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            const auto& data_ctx = primitive_parameters.CastTo<CastType>();
+            auto argument_ptr    = ck_args.MakeArgPtr(sh_conv_ptr, data_ctx.tensors);
+            auto invoker_ptr     = sh_conv_ptr->MakeInvokerPointer();
+
+            if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
+            {
+                auto zero           = 0.0f;
+                const auto& tensors = data_ctx.tensors;
+                SetTensor(handle, tensors.dwDesc, tensors.dw, &zero);
```
atamazov: Issue: SetTensor() runs a kernel, but that kernel is excluded from benchmarking. Please fix this.

atamazov: Let's restructure the code to avoid code duplication:
```C++
{
    Profiler t; // see the Recommendation above
    if(WrW)
    {
        ...set output buffer to 0...
    }
    ....run CK kernel....
} // end of the scope - implicitly destructs t
```
So we do not need to create a copy of the code that launches the CK kernel..

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -114,15 +132,33 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
+                if(handle.IsProfilingEnabled())
+                {
+                    ProfilingRecordStart(handle, start, stop);
+                }
+                invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), false});
+                if(handle.IsProfilingEnabled())
+                    ProfilingRecordStop(handle, start, stop);
+            }
+            else
+            {
```
atamazov: [Warning] At line 164 we are still getting elapsed time from CK, which, are you've said, leads to bugs (or the problem in CK impacts only WrW kernels?) Anyway, resolving my previous review comment would fix this.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -606,6 +643,12 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
                 std::swap(conv_tensors.xDesc, conv_tensors.yDesc);
             }
 
+            HipEventPtr start = nullptr;
```
atamazov: Quote from https://github.com/ROCm/MIOpen/pull/2770#discussion_r1506089187:

> When the ck-profiling is enabled, the ck wrw kernel introduces the errors that cause the precision issue.

- Please open an issue in CK repo and add a link to it to the description of this PR.
- Please create the `WORKAROUND_CK_ISSUE_XXX` with value 1 and wrap all the W/A related code in `#if WORKAROUND_CK_ISSUE_XXX ... #endif`. See #1685 for more info.

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: At line 660 below there is:
https://github.com/ROCm/MIOpen/blob/81f29c00af6dcc3e9053390953373dc55dc45117/src/include/miopen/solver/implicitgemm_ck_util.hpp#L658-L663
Please delete _"Fix NHWC Wrw invokers to also issue a zero-out kernel"_ from the todo comment. The second statement is still needed - it is about of replacing ZeroOutBuffer() with SetTensor() AFAICS.

/cc @amberhassaan

atamazov (changes requested):

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [obsolete, hide as if it is resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [obsolete, hide as if it is resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: @iq136boy Likely so unless your clang-format tool or ./.clang-format file is corrupt or something else. Please check what CI prints during the "Clang Format" stage. **_This is important._**

atamazov:
@iq136boy Please mark this with https://github.com/ROCm/MIOpen/labels/IMPACTS_FIND_DB

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [Q][Important 🔴] Do we need to clean output buffer for **_all_** CK-based WrW convolutions?

atamazov (changes requested):

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: (continued at https://github.com/ROCm/MIOpen/pull/2770#discussion_r1509099090)

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: Of course I am not against replacing ZeroOutBuffer() with SetTensor() right here, but this is a bit out of scope of this PR, so I am not insisting on anything. Up to you.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: @atamazov : yes, as things stand today. We'll soon have support for `alpha & beta` scaling parameters in CK convolutions, which when used will not need zeroing out of the weights buffer.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -96,6 +97,23 @@ bool IsCKApplicable(const ProblemDescriptionType& problem)
         ptrs.begin(), ptrs.end(), [&args](auto& ptr) { return args.IsSupportedBy(ptr); });
 }
 
+inline void ProfilingRecordStart(const Handle& handle, HipEventPtr& start, HipEventPtr& stop)
```
amberhassaan: As @atamazov said, this is best handled as a class that provides `Start()`, `Stop()` and `GetTime()` and holds the `hip events` as members.

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: [resolved]

src/include/miopen/solver/implicitgemm_ck_util.hpp
atamazov: @amberhassaan Clarification: Right now I am proposing the simplest solution that has everything in ctor and dtor. More stuff can be provided later when/if we need better flexibility.

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: > Issue: SetTensor() runs a kernel, but that kernel is excluded from benchmarking. Please fix this.

I will fix it. Thanks.

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: > Quote from [#2770 (comment)](https://github.com/ROCm/MIOpen/pull/2770#discussion_r1506089187):
> 
> > When the ck-profiling is enabled, the ck wrw kernel introduces the errors that cause the precision issue.
> 
> * Please open an issue in CK repo and add a link to it to the description of this PR.
> * Please create the `WORKAROUND_CK_ISSUE_XXX` with value 1 and wrap all the W/A related code in `#if WORKAROUND_CK_ISSUE_XXX ... #endif`. See [Lifecycle of workarounds #1685](https://github.com/ROCm/MIOpen/discussions/1685) for more info.

Yes. I am about to do it. I have discussed with CK team about it. It seems that the issue I found is a none trivial task for them. That is why I am trying to find some workaround solution from hostside.

iq136boy:
> @iq136boy Please mark this with [ IMPACTS_FIND_DB](https://github.com/ROCm/MIOpen/labels/IMPACTS_FIND_DB)

done

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: > [Warning] At line 164 we are still getting elapsed time from CK, which, are you've said, leads to bugs (or the problem in CK impacts only WrW kernels?) Anyway, resolving my previous review comment would fix this.

Yes. Only wrw kernels has the issue. I will fix and restructure the code base your comments.

iq136boy:
@atamazov Thanks for your comments and code restructure recommendations. I will work to address them.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -96,6 +97,34 @@ bool IsCKApplicable(const ProblemDescriptionType& problem)
...
+#ifdef WORKAROUND_CK_ISSUE_1184
+struct HipEventProfiler
+{
+    const Handle& handle;
+    float event_time;
+    HipEventPtr start;
+    HipEventPtr stop;
+
+    HipEventProfiler(const Handle& handle_)
+        : handle(std::move(handle_)),
```
atamazov: I do not think we need std::move here. Or I am missing something?

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -96,6 +97,34 @@ bool IsCKApplicable(const ProblemDescriptionType& problem)
         ptrs.begin(), ptrs.end(), [&args](auto& ptr) { return args.IsSupportedBy(ptr); });
 }
 
+#define WORKAROUND_CK_ISSUE_1184 1
+#ifdef WORKAROUND_CK_ISSUE_1184
```
atamazov:
```suggestion
#if WORKAROUND_CK_ISSUE_1184
```

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -122,14 +151,17 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
-                const auto enable_profiling = handle.IsProfilingEnabled();
-                float elapsed_time =
-                    invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), enable_profiling});
-                if(enable_profiling)
                 {
-                    handle.ResetKernelTime();
-                    handle.AccumKernelTime(elapsed_time);
+#ifdef WORKAROUND_CK_ISSUE_1184
+                    HipEventProfiler pfr(handle);
+#endif
```
atamazov: We do not have "normal" implementation here yet, and therefore disabling the workaround should lead to build errors. The simplest way to achieve this:
```suggestion
                    HipEventProfiler pfr(handle);
```

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -605,15 +637,15 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
                 std::swap(conv_tensors.x, conv_tensors.y);
                 std::swap(conv_tensors.xDesc, conv_tensors.yDesc);
             }
-
+#ifdef WORKAROUND_CK_ISSUE_1184
+            HipEventProfiler pfr(handle);
+#endif
```
atamazov: Ditto
```suggestion
            HipEventProfiler pfr(handle);
```

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -605,15 +637,15 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
...
             input1_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
 
             input2_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
 
             output_init_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
 
-            /// \todo: Fix NHWC Wrw invokers to also issue a zero-out kernel. Will
-            /// need SetTensor() to properly zero out non-packed tensors
```
atamazov, line 616: Please keep the second statement of this TODO. It is about replacing ZeroOutBuffer with SetSensor here

atamazov (changes requested):

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: You are right. It is a reference. No need std::move. Fixed.

atamazov (approved):
🟢 LGTM!

cderb (approved):
LGTM

junliume (approved):

PR #2773, opened 2024-02-28
Adding Link Dependencies 

arvindcheru:
Changes proposed
- Add link dependencies to fix build issue

### DIFF
--- a/tools/sqlite2txt/CMakeLists.txt
+++ b/tools/sqlite2txt/CMakeLists.txt
@@ -2,5 +2,5 @@ add_executable(sqlite2txt
         main.cpp
 )
 
-target_link_libraries(sqlite2txt SQLite::SQLite3 dl)
+target_link_libraries(sqlite2txt SQLite::SQLite3 pthread dl)
 clang_tidy_check(sqlite2txt)
--- a/tools/sqlite2txt/CMakeLists.txt
+++ b/tools/sqlite2txt/CMakeLists.txt
@@ -2,5 +2,5 @@ add_executable(sqlite2txt
         main.cpp
 )
 
-target_link_libraries(sqlite2txt SQLite::SQLite3)
+target_link_libraries(sqlite2txt SQLite::SQLite3 dl)
 clang_tidy_check(sqlite2txt)


junliume (approved):
@arvindcheru it looks okay but what was the issue? How to reproduce it?

atamazov:
@junliume I think this is to resolve https://github.com/ROCm/MIOpen/pull/2769#issuecomment-1965416826

arvindcheru:
> @arvindcheru it looks okay but what was the issue? How to reproduce it?

Thank you. Yes it is to address the issue shared by [atamazov](https://github.com/atamazov), it was reproduced in our local BAAS build test-4257

tools/sqlite2txt/CMakeLists.txt
```diff
@@ -2,5 +2,5 @@ add_executable(sqlite2txt
         main.cpp
 )
 
-target_link_libraries(sqlite2txt SQLite::SQLite3)
+target_link_libraries(sqlite2txt SQLite::SQLite3 pthread dl)
```
atamazov, line 5: The missing symbols from pthread and dl are referenced by sqlite. There is no direct sqlite2txt's dependence on those.

JehandadKhan:
This issue should be fixed in the SQLite CMake module, since the code in MIOpen does not depend on pthreads or dlopen, its rather `SQLite::SQLite3` which depends on these components. Therefore, these should be coming from the SQLite CMake module. 

cc @atamazov

JehandadKhan (approved):
As a workaround, this is fine, however, as mentioned in my comment, it should be fixed in the Cmake module itself rather than here.

atamazov:
@JehandadKhan FindSQlite3.cmake is identical on both ROCm 5.7 (where the build fails, gcc 9.4.0) and 6.0 (where everything is fine, gcc 11.4.0). However CMake test CMAKE_HAVE_LIBC_PTHREAD fails on 5.7 and passes on 6.0, which means that on 6.0 thread support functions are compiled into libc library directly.

Maybe FindSQlite3.cmake is missing `target_link_libraries(SQLite::SQLite3 PRIVATE pthread dl)` which would make the transitive dependency explicit, but I am not sure that the fix is that trivial, because SQLite3 can be built in different configurations, and some of them may not require threads or dl.

Anyway I guess that fixing stuff provided by CMake is out of the scope of our work ;)

/cc @arvindcheru

atamazov (approved):
🟢

DrizztDoUrden (approved):

PR #2776, opened 2024-02-29
Softmax for find20

Vsevolod1983:
1. Softmax primitive is implemented in terms of Find 2.0 API 
2. GTest is added for testing Find 2.0 wrapper over Softmax for both SoftmaxForward and SoftmaxBackward

TICKETS: 
LWPMIOPEN-519
LWPMIOPEN-557


CAHEK7 (approved):

Vsevolod1983:
Hi @junliume 
I can see "Read The Docs" is the only fail in my pull request. Clicking on Details brings me to 404 not found page.
Should I worry about this fail or its non-blocking thing ?

junliume:
@atamazov @JehandadKhan please review since `miopen.h` is modified. Thanks!

Gentle ping to @DrizztDoUrden for find 2.0 review

atamazov:
@Vsevolod1983 Will review tomorrow, sorry for delaying.

> `docs/readthedocs.com:advanced-micro-devices-miopen — Read the Docs build failed!`

Please check this. Changes in miopen.h may lead to such failures.

src/softmax.cpp
```diff
@@ -33,8 +33,55 @@
...
 
+#include <nlohmann/json.hpp>
+
 namespace miopen {
 
+extern "C" miopenStatus_t miopenCreateSoftmaxDescriptor(miopenSoftmaxDescriptor_t* activDesc)
+{
+
+    MIOPEN_LOG_FUNCTION(activDesc);
+    return miopen::try_([&] { miopen::deref(activDesc) = new miopen::SoftmaxDescriptor(); });
```
DrizztDoUrden:
```suggestion
    return miopen::try_([&] {
        auto& desc = miopen::deref(activDesc);
        desc = new miopen::SoftmaxDescriptor();
    });
```
This way memory would not leak if deref throws

DrizztDoUrden (approved):
LGTM with one minor suggestion

include/miopen/miopen.h
```diff
@@ -5314,6 +5319,11 @@ typedef enum
     miopenTensorBiasY        = 9,
     miopenTensorBias         = 10,
 #endif
+    miopenTensorSoftmaxX  = 11,
+    miopenTensorSoftmaxY  = 12,
+    miopenTensorSoftmaxDX = 13,
+    miopenTensorSoftmaxDY = 14,
```
atamazov: [Recommendation] I highly recommend moving these new members above the BETA API stuff and renumber items. This way, non-BETA members will have stable values regardless of what is happening with BETA things.

src/problem.cpp
```diff
@@ -431,6 +463,60 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
...
+    std::vector<solver::softmax::SoftmaxSolver*> solvers;
+
+    solvers.push_back(&attnSoftmaxSolver);
+    solvers.push_back(&regularSoftmaxSolver);
+
+    for(auto solver : solvers)
+    {
+        if(!solver->IsApplicable(ctx, problem_description))
+        {
+            continue;
```
atamazov, line 494: [R] It may worth logging non-applicable Solvers at Info2 level, like we do for convolution solvers.

src/softmax.cpp
```diff
@@ -33,8 +33,57 @@
...
 
+#include <nlohmann/json.hpp>
+
 namespace miopen {
 
+extern "C" miopenStatus_t miopenCreateSoftmaxDescriptor(miopenSoftmaxDescriptor_t* activDesc)
+{
+    MIOPEN_LOG_FUNCTION(activDesc);
+    return miopen::try_([&] {
+        auto& desc = miopen::deref(activDesc);
```
atamazov: [R] `activDesc` seems inadequate in this context, right? 😉

src/problem.cpp
```diff
@@ -277,6 +282,33 @@ activ::ProblemDescription Problem::AsActivation() const
...
+    const auto& softmax_desc = boost::get<SoftmaxDescriptor>(operator_descriptor);
+
+    float alpha = softmax_desc.GetAlpha();
+    float beta  = softmax_desc.GetBeta();
+
+    softmax::ProblemDescription problem_description =
+        (GetDirection() == miopenProblemDirectionForward)
+            ? softmax::ProblemDescription(
+                  &alpha,
+                  &beta,
```
atamazov, line 296: [Performance] @Vsevolod1983 See https://github.com/ROCm/MIOpen/pull/2671#pullrequestreview-1937760018

src/problem.cpp
```diff
@@ -431,6 +463,60 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
...
+
+    const auto algo = AlgorithmName{"Softmax"};
+
+    solver::softmax::AttnSoftmax attnSoftmaxSolver;
+    solver::softmax::Softmax regularSoftmaxSolver;
+
+    std::vector<solver::softmax::SoftmaxSolver*> solvers;
+
+    solvers.push_back(&attnSoftmaxSolver);
+    solvers.push_back(&regularSoftmaxSolver);
```
atamazov: [Can be postponed] I think we can statically initialize `solvers`.

src/problem.cpp
```diff
@@ -431,6 +463,60 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
...
+        if(!solver->IsApplicable(ctx, problem_description))
+        {
+            continue;
+        }
+
+        auto solution = Solution();
+
+        /// \todo time measurement will be done later. For now we set less time for attention
+        /// softmax and slightly bigger for regular
+        solution.SetTime(solver == &attnSoftmaxSolver ? 0.0f : 1.0f);
```
atamazov: No zero time, please.

[Q] Are we really going to benchmark the applicable Solutions? Or just use some kind of heuristics like WTI computations, neural nets or so?

atamazov (changes requested):

include/miopen/miopen.h
Vsevolod1983: Renumbering existing items in enums are usually dangerous in case if they get serialized somewhere.
Are we sure they are not serialized ?

src/problem.cpp
Vsevolod1983: What do you suggest instead of zero time? 0.1f ?
As we told with @CAHEK7  probably later we add some time management, not a highly priority task though.

atamazov: @Vsevolod1983 `? 1.0f : 2.0f);`

include/miopen/miopen.h
atamazov: @Vsevolod1983 We do not care about BETA_API stuff. At least it has lower importance than "normal" API things, right?

Vsevolod1983: ok

src/problem.cpp
Vsevolod1983: I removed alpha and beta from network config in softmax primitive. Is it enough for the this PR ?

src/problem.cpp
atamazov: @DrizztDoUrden Can you please suggest @Vsevolod1983 the technique for doing this? Thanks.

src/softmax.cpp
atamazov: [resolved]

src/problem.cpp
atamazov: [resolved]

src/problem.cpp
```diff
@@ -490,14 +490,15 @@ Problem::FindSolutionsImpl(Handle& handle,
     {
         if(!solver->IsApplicable(ctx, problem_description))
         {
+            MIOPEN_LOG_I2("solver " << solver->SolverDbId() << "is not applicable for the given problem description");
```
atamazov: Space before "is" is missing. Please borrow message format from there: https://github.com/ROCm/MIOpen/blob/cee33a7aaa3dbbe850d2580a72fff8774ddbc131/src/include/miopen/find_solution.hpp#L209

include/miopen/miopen.h
```diff
@@ -5310,19 +5310,20 @@ typedef enum
     miopenTensorConvolutionX      = 1,
     miopenTensorConvolutionW      = 2,
     miopenTensorConvolutionY      = 3,
+    miopenTensorSoftmaxX  = 4,
```
atamazov: [Informative] I suspect that clang-format will fail. See https://github.com/ROCm/MIOpen/wiki/How-to-format-code

atamazov (changes requested):

src/problem.cpp
atamazov: @Vsevolod1983 I would begin with removing alpha/beta from PD (and then fix build errors by forwarding alpha/beta via InvokeParams). However I think we need to discuss https://github.com/ROCm/MIOpen/pull/2671#pullrequestreview-1937760018 first.

/cc @CAHEK7 @DrizztDoUrden

include/miopen/miopen.h
atamazov: [resolved]

src/problem.cpp
atamazov: [resolved]

atamazov:
@Vsevolod1983 Please hide all resolved review comments to save space (my review comments are resolved when marked with `[resolved]`), thanks!

src/problem.cpp
Vsevolod1983: Should we do this kind of refactoring as a separate ticket / PR ?

CAHEK7: We need. Actually a/b affects kernel compilation, like it does for convolutions. There are some optimizations with default a/b values, so we need at least a "default a/b" flag in the problem description and in the network config.

Vsevolod1983: I restored alpha and beta in network config in this PR until we decide what exactly we want to do next.

atamazov: [Informative] @Vsevolod1983 @CAHEK7 it's interesting that "default alpha" optimization won't give us much, but "default beta" opt removes 1 global memory read, which may be substantial (./src/kernels/MIOpenSoftmax.cl)

src/problem.cpp
atamazov: [resolved]

atamazov (approved):
🟢 LGTM!

Postponed topics:
- https://github.com/ROCm/MIOpen/pull/2776#discussion_r1525541568 - alpha/beta thing
- https://github.com/ROCm/MIOpen/pull/2776#discussion_r1525544719 - static init of vector of solvers

src/problem.cpp
CAHEK7: @atamazov right now `default alpha` gives much since it enables attention softmax solver, which is faster and does not support a/b in terms of normal softmax operation.

Both values are used in the `isApplicable` method and can't be just removed.

junliume (approved):

src/problem.cpp
DrizztDoUrden: The most basic primitive for this is https://github.com/ROCm/MIOpen/blob/9db0c265deb57b33d1a5669264157bc2700bd6d8/src/include/miopen/find_solution.hpp#L181

src/include/miopen/softmax.hpp
```diff
@@ -28,12 +28,47 @@
...
+    }
+
+    friend std::ostream& operator<<(std::ostream& stream, const SoftmaxDescriptor& x);
+
+    friend void to_json(nlohmann::json& json, const SoftmaxDescriptor& descriptor);
+    friend void from_json(const nlohmann::json& json, SoftmaxDescriptor& descriptor);
+
+private:
+    float alpha;
+    float beta;
```
DrizztDoUrden, line 67: I would rather store them as `bool`. This way later it would be harder to accidentally use them instead of actual values. On subsequent runs float values here would be incorrect anyway.
```suggestion
    bool has_alpha;
    bool has_beta;
```

src/solution.cpp
```diff
@@ -232,6 +241,91 @@ void Solution::RunImpl(Handle& handle,
...
+        }
+        auto ret = found->second;
+        if(!ret.descriptor.has_value())
+            ret.descriptor = problem_casted.GetTensorDescriptorChecked(name, name_str);
+        return ret;
+    };
+
+    const softmax::ProblemDescription problem_description = problem_casted.AsSoftmax();
+
+    float alpha                        = softmax_desc.GetAlpha();
```
DrizztDoUrden, line 268: You can't use these values, they are outdated on subsequent calls. User should be able to invoke the same solution multiple times with different a/b, unless they were defaulted in find.

DrizztDoUrden (changes requested):

DrizztDoUrden:
If we merge it with a/b set through PD, it would be kind of a breaking change later. To do so it has to be wrapped in `MIOPEN_BETA_API`

src/problem.cpp
DrizztDoUrden: This is postponable in contrast to a/b problem

DrizztDoUrden:
I have somewhat hacky solution for passing scalar arguments at find and run time here: https://github.com/ROCm/MIOpen/pull/2825

DrizztDoUrden:
I'll split scalar argument passing to a separate [PR](https://github.com/ROCm/MIOpen/pull/2826) today

Vsevolod1983:
I will use MIOPEN_BETA_API for this PR to have it merged. When we get understanding on how to refactor, lets do it in a separate ticket

Vsevolod1983:
@DrizztDoUrden 
I put miopenSoftmaxCreateProblem declaration under MIOPEN_BETA_API. Tensor's enum values as well.

Btw, just an observation. Our library can be built only with this beta api flag. 
We have   add_definitions("-DMIOPEN_BETA_API=1") in cmakelist.txt file.
Without this flag library cannot be compiled.

DrizztDoUrden (approved):
LGTM

PR #2782, opened 2024-03-04
Minor softmax improvements

Vsevolod1983:
These are minor  improvements for PR-2671 https://github.com/ROCm/MIOpen/pull/2671#discussion_r1472813196
which @atamazov recently requested to do.
It includes:
1. Minor changes for softmax problem description
2. Changes for softmax test driver (which are actually done by Artem himself).

TICKET:
LWPMIOPEN-534


CAHEK7 (approved):

junliume (approved):

src/softmax/problem_description.cpp
```diff
@@ -45,14 +47,11 @@ NetworkConfig ProblemDescription::MakeNetworkConfig() const
...
-
-                          "n_y" + std::to_string(n_y) + "c_y" + std::to_string(c_y) + "h_y" +
-                          std::to_string(h_y) + "w_y" + std::to_string(w_y);
+        ss << "n_x" << n_x << "c_x" << c_x << "h_x" << h_x << "w_x" << w_x;
+        ss << "n_y" << n_y << "c_y" << c_y << "h_y" << h_y << "w_y" << w_y;
 
-        network_config += "xpk" + std::to_string(static_cast<int>(xdxDesc.IsPacked())) + "ypk" +
-                          std::to_string(static_cast<int>(yDesc.IsPacked()));
+        ss << "xpk" << static_cast<int>(xdxDesc.IsPacked());
+        ss << "ypk" << static_cast<int>(yDesc.IsPacked());
```
atamazov, line 54: [Notice] @Vsevolod1983 @CAHEK7 @averinevg (1) We do not store full information about unpacked tensors, only the flag. Also we do not store information about tensor layouts, at all. The above means that we'll need to extend the network config when support for unpacked tensors or support for different layouts is required.

If this is not done, then, for example, MIOpen won't be able to distinguish between two invokers when unpacked tensors are used (same NCHW but different strides).

@Vsevolod1983 @CAHEK7 Does it worth a ticket, what do you think?

(2) It may worth extending the common tensor descriptor abstraction to support serialization (with custom prefix). This would allow for uniform serialization of TDs in different primitives + for removal of duplicated code. Currently each primitive uses its own code for serialization of TDs. @averinevg If this makes sense, then can you please open a ticket?

/cc @junliume @JehandadKhan

Vsevolod1983: @atamazov 
We can do it in another PR. Actually this PR does not change anything. Its only about using ostream instead of strings concatenation. So this PR does not change network config structure.

atamazov: @Vsevolod1983 Correct. However, the problem will become real when/if we will support upacked tensors (see (1) above) or serialized PD will be used as a key for the persistent databases (as this change in the db key will lead to backward incompatibility of the databases).

I'll open a ticket.

atamazov: [resolved] since tickets have been opened.

PR #2785, opened 2024-03-05
[CI][test-perf] MIOpenDriver to use rocrand to init buffers. Do not init output buffers. Use non-DEV build in perf test.

atamazov:
Accelerate performance testing on CI.
- [driver][conv] Added the `--gpualloc/-G` option. Avoids copying huge buffers from host to device and back. Instead of initialization of host buffers with subsequent copying, initializes buffers on the host directly by rocRand. Verification is disabled in this mode.
- [driver][conv] Another optimization: Do not initialize output buffers.
- [tests] Engaged the new driver's option in `test_perf.py`.
- [CI] Use BUILD_DEV=Off when package_build=true. This enables binary cache.

### Performance testing results
- Navi21, ROCm 6.0
  - Baseline: 149 mins
  - `--gpualloc 1` and other driver changes: 117 mins (~1.3X)
  - BUILD_DEV=Off: 78 mins (another ~1.5X)

Total acceleration is ~1.9X.

[informative] Potential optimization of `test_perf.py`: Do not test the same convolution config more than once. Currently we have 4897 configs under test, but 1126 of those are duplicates. This may give another 1.3X acceleration.

----
- https://github.com/ROCm/MIOpen/labels/performance
- https://github.com/ROCm/MIOpen/labels/testing
- Proposed reviewers:
  - @JehandadKhan
  - @pfultz2 
  - @CAHEK7


atamazov:
@junliume @JehandadKhan ⚠️ Please note that Jenkinsfile changes from fork repos are NOT tested. Can you please create a copy of this branch in _this repo_ and test it. Thanks!

junliume:
@atamazov this error looks very strange IMHO, maybe because rocRAND is not installed by default.
```
[2024-03-05T19:37:14.992Z] [100%] Linking CXX executable ../bin/MIOpenDriver
[2024-03-05T19:37:14.992Z] ld.lld: error: unable to find library -lrocrand
```

CMakeLists.txt
```diff
@@ -412,6 +412,10 @@ endif()
 
 option(Boost_USE_STATIC_LIBS "Use boost static libraries" ON)
 set(BOOST_COMPONENTS filesystem)
+if(MIOPEN_BUILD_DRIVER)
+    # boost core is a header only component that can't be found by find_package
+    # list(APPEND BOOST_COMPONENTS core)
```
junliume, line 419: @atamazov all commented out, so WIP?

driver/CMakeLists.txt
```diff
@@ -25,15 +25,17 @@
 ################################################################################
 
 find_package(Threads REQUIRED)
+find_package(rocrand REQUIRED)
+message(STATUS "rocrand_VERSION= ${rocrand_VERSION}")
 
-add_executable(MIOpenDriver main.cpp InputFlags.cpp)
+add_executable(MIOpenDriver main.cpp InputFlags.cpp rocrand_wrapper.cpp)
```
junliume, line 29: @atamazov in the top-most CMakeLists.txt we need to make rocrand a dependency

junliume: and ultimately make `MIOpen package` be dependent on `rocRAND`

driver/conv_driver.hpp
```diff
@@ -201,16 +210,197 @@ bool readBufferFromFile(T* data, size_t dataNumItems, const char* fileName)
...
+
+    void AllocOnHost(miopenTensorDescriptor_t t)
+    {
+        host = tensor<Tgpu>(miopen::deref(t));
+        if(is_gpualloc) // We do not need host data.
+        {
+            host.data.clear();
+            host.data.shrink_to_fit(); // To free host memory.
+        }
+    }
```
CAHEK7, line 232: What shall we do in an `else` case?
Its name is `AllocOnHost` but it does deallocation for `is_gpualloc` or does nothing. That may be confusing.

driver/conv_driver.hpp
```diff
@@ -201,16 +210,197 @@ bool readBufferFromFile(T* data, size_t dataNumItems, const char* fileName)
...
+            /// Verification cache would be broken otherwise.
+            auto val = generator();
+            if(do_write)
+                GetVector()[i] = val;
+        }
+    }
+
+    status_t AllocOnDevice(stream, context_t ctx, const size_t sz)
+    {
+        dev = std::unique_ptr<GPUMem>(new GPUMem(ctx, sz, sizeof(Tgpu)));
```
CAHEK7: It's not critical, but consider useing `std::make_unique`
```suggestion
        dev = std::make_unique<GPUMem>(ctx, sz, sizeof(Tgpu));
```
Also if a reallocation is required, it would be better to explicitly free the buffer before the new allocation, because in this particular case it requires `(oldSz + sz) * sizeof(Tgpu)` amount of memory instead of just a `sz`. May or may not be critical.
```suggestion
        dev.reset();
        dev = std::unique_ptr<GPUMem>(new GPUMem(ctx, sz, sizeof(Tgpu))); //or std::make_unique()
```

driver/conv_driver.hpp
```diff
@@ -201,16 +210,197 @@ bool readBufferFromFile(T* data, size_t dataNumItems, const char* fileName)
...
+
+    template <typename T>
+    status_t AllocOnDevice(stream, context_t ctx, const size_t sz, std::vector<T>& init)
+    {
+        static_assert(std::is_same<T, float>::value           //
+                          || std::is_same<T, int32_t>::value, //
+                      "Before enabling more types, check thoroughly.");
+        dev = std::unique_ptr<GPUMem>(new GPUMem(ctx, sz, sizeof(T)));
+        return STATUS_SUCCESS;
+    }
```
CAHEK7: `init` is not used, and also check the comments from the previous `AllocOnDevice`

CMakeLists.txt
atamazov: @junliume The first line of the comment explains why the second line is commented out.

driver/conv_driver.hpp
atamazov: In gpualloc mode it does nothing because we do not need buffers on the host. And yes, the design may confuse the reader, but the root reason of this is clumsy initial design of the driver. At least we have a kind of imperfect abstraction now; let's improve this stuff in the future.

driver/conv_driver.hpp
atamazov: @CAHEK7 Reallocation is not required at all yet. By design, each buffer is allocated only once at the beginning and then freed once at the end. Thanks for this comment anyway.

CMakeLists.txt
atamazov: @junliume All this stuff can be removed, of course. I keep this only to inform the reader.

driver/CMakeLists.txt
atamazov: @junliume 
> in the top-most CMakeLists.txt

I am wondering why all this can't be done right in driver/CMakeLists.txt, but will do ;)

atamazov: @junliume Fixed in 9ada41aab14643469c55c0d00251bf36a962f89f

atamazov:
@junliume @CAHEK7 Ready for the next review & CI testing round (reminder: if you are going to kick off CI, then please notice https://github.com/ROCm/MIOpen/pull/2785#issuecomment-1979107530)

NB: Now the PR includes some unrelated fixes, as explained at:
- https://github.com/ROCm/MIOpen/pull/2772#discussion_r1515221540
- https://github.com/ROCm/MIOpen/pull/2772#issuecomment-1981943257

driver/conv_driver.hpp
atamazov: [Gentle reminder] @CAHEK7 Please mark this review thread as "resolved" or provide some other kind of feedback.

driver/conv_driver.hpp
atamazov: [Gentle reminder] @CAHEK7 Please mark this review thread as "resolved" or provide some other kind of feedback.

driver/CMakeLists.txt
junliume: @atamazov IMHO in order to fix this issue we need to add a line of `rocrand` right below this:
https://github.com/ROCm/MIOpen/blob/cdb4c740c361cc75d053d4b69463ecb2c813468e/Dockerfile#L52

junliume: and some where near https://github.com/ROCm/MIOpen/blob/cdb4c740c361cc75d053d4b69463ecb2c813468e/CMakeLists.txt#L484-L486
```
set(MIOPEN_PACKAGE_REQS "${MIOPEN_PACKAGE_REQS}, rocrand")
```

driver/CMakeLists.txt
```diff
@@ -26,14 +26,14 @@
...
 if(WIN32)
     # Refer to https://en.cppreference.com/w/cpp/language/types for details.
     target_compile_options(MIOpenDriver PRIVATE $<BUILD_INTERFACE:$<$<CXX_COMPILER_ID:Clang>:-U__LP64__>>)
 endif()
 add_dependencies(MIOpenDriver generate_kernels)
 target_include_directories(MIOpenDriver PRIVATE ../src/kernels)
-target_link_libraries(MIOpenDriver MIOpen Threads::Threads)
+target_link_libraries(MIOpenDriver MIOpen Threads::Threads rocrand)
```
junliume, line 36: @atamazov found an weird issue with
```
[100%] Linking CXX executable ../bin/MIOpenDriver
ld.lld: error: unable to find library -lrocrand
```
here is how to reproduce:
- rocm/miopen:ci_a9eebf
- CXX=/opt/rocm/llvm/bin/clang++ CXXFLAGS='-Werror'  cmake -DMIOPEN_TEST_FLAGS=' --disable-verification-cache ' -DCMAKE_BUILD_TYPE=debug -DCMAKE_CXX_FLAGS_DEBUG='-g -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=undefined -Wno-option-ignored ' -DBUILD_DEV=On -DMIOPEN_USE_MLIR=ON -DMIOPEN_GPU_SYNC=Off -DMIOPEN_BACKEND=HIPNOGPU -DMIOPEN_INSTALL_CXX_HEADERS=On -DCMAKE_PREFIX_PATH=/opt/rocm ..
- make -j$(nproc)

driver/conv_driver.hpp
CAHEK7: What do you think about `make_unique`?

driver/conv_driver.hpp
```diff
@@ -201,16 +210,197 @@ bool readBufferFromFile(T* data, size_t dataNumItems, const char* fileName)
...
+
+template <typename Tgpu>
+class GpumemVector
+{
+    std::unique_ptr<GPUMem> dev;
+    std::vector<Tgpu> host;
+    bool is_gpualloc = false;
+
+public:
+    void SetGpuallocMode(bool v) { is_gpualloc = v; }
```
CAHEK7, line 350: Why can't it be a part of constructor?
To avoid defining all of them?

driver/conv_driver.hpp
```diff
@@ -633,6 +816,29 @@ int ConvDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
...
         }
     }
+
+    is_gpualloc = (inflags.GetValueInt("gpualloc") == 1);
+
+    if(is_gpualloc && inflags.GetValueInt("verify") == 1)
+    {
+        std::cerr << "Error: '--gpualloc 1' should not be used with enabled verification. Add "
+                     "'--verify 0' to options."
+                  << std::endl;
```
CAHEK7, line 826: It's a bit artificial limitation, especially for convolution with gpu-based verification. In this case we can get benefits even for regular runs.
(probably it should be fixed in a subsequent PR)

driver/conv_driver.hpp
```diff
@@ -919,6 +1125,15 @@ int ConvDriver<Tgpu, Tref>::AddCmdLineArgs()
...
                          "\n<invalid name> Use Find() API",
                          "string");
+    inflags.AddInputFlag("gpualloc",
+                         'G',
+                         "0",
+                         "Controls allocation and initialization buffers on GPU and CPU."
+                         "\n0 Init input buffers on CPU and copy them to GPU. After convolution"
+                         "\n  is executed, copy output buffer to CPU (Default)."
+                         "\n1 No copying. Use hipMalloc to allocate and rocrand to init buffers"
+                         "\n  directly on GPU. Verification (-V 1) won't succeed in this mode.",
```
CAHEK7, line 1135: Check my previous comment about this limitation. I guess it can be allowed with gpu-based verification.

driver/conv_driver.hpp
```diff
@@ -1449,19 +1651,18 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
...
+                bool read = false;
+                if(!biasFileName.empty())
+                    read = readBufferFromFile<float>(b_int8.data(), b_sz, biasFileName.c_str());
+                if(!read)
+                    for(int i = 0; i < b_sz; i++)
+                        b_int8[i] = static_cast<float>(i % 8) + prng::gen_canonical<float>();
             }
-
-            b_dev->ToGPU(q, b_int8.data());
+            std::ignore = b.AllocOnDeviceAndInit(q, ctx, b_sz, b_int8);
```
CAHEK7, line 1665: Why do we need to explicitly ignore the result? It is not marked with `[[nodiscard]]` attribute.
Anyway, all the other cases are checked to `STATUS_SUCCESS` or do not have explicit `std::ignore`.
Could you make it consistent?

driver/conv_driver.hpp
```diff
@@ -1433,12 +1635,12 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
     bool dataRead = false;
     if(is_fwd || is_wrw)
         if(!inFileName.empty())
-            dataRead = readBufferFromFile<Tgpu>(in.data.data(), in_sz, inFileName.c_str());
+            dataRead = readBufferFromFile<Tgpu>(in.GetVectorData(), in_sz, inFileName.c_str());
```
CAHEK7, line 1638: We will get a segfault if `!inFileName.empty()` and `is_gpualloc`.

driver/conv_driver.hpp
```diff
@@ -1433,12 +1635,12 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
...
 
     bool weiRead = false;
     if(is_fwd || is_bwd)
         if(!weiFileName.empty())
-            weiRead = readBufferFromFile<Tgpu>(wei.data.data(), wei_sz, weiFileName.c_str());
+            weiRead = readBufferFromFile<Tgpu>(wei.GetVectorData(), wei_sz, weiFileName.c_str());
```
CAHEK7, line 1643: We will get a segfault if `!weiFileName.empty()` and `is_gpualloc`.

driver/conv_driver.hpp
```diff
@@ -1470,130 +1671,120 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
         bool doutRead = false;
         if(is_bwd || is_wrw)
             if(!doutFileName.empty())
-                doutRead = readBufferFromFile<Tgpu>(dout.data.data(), out_sz, doutFileName.c_str());
+                doutRead =
+                    readBufferFromFile<Tgpu>(dout.GetVectorData(), out_sz, doutFileName.c_str());
```
CAHEK7, line 1675: We will get a segfault if `!doutFileName.empty()` and `is_gpualloc`.

driver/conv_driver.hpp
CAHEK7: But unfortunately it is still misleading and error prone.
You explicitly call `AllocOnHost` and later you may think that `GetVectorData` may return some valid pointer and will get a segfault.

You've already wrote that kind of code even if you are aware about this behavior.

driver/conv_driver.hpp
```diff
@@ -1470,130 +1671,120 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
...
-            }
 
+            b.AllocOnHost(biasTensor);
+            db.AllocOnHost(b_sz);
+            db_host = tensor<Tref>(miopen::deref(biasTensor));
+
+            // Init tensor on host
+            bool b_read = false;
             if(!biasFileName.empty())
+                b_read = readBufferFromFile<Tgpu>(b.GetVectorData(), b_sz, biasFileName.c_str());
```
CAHEK7, line 1701: `b.AllocOnHost` does not allocate any data, and subsequent `readBufferFromFile<Tgpu>(b.GetVectorData(),...` will fail with segfault.
The code looks correct at the first glance, but under the hood it has 2 pitfalls - `AllocOnHost` does not allocate any buffers and `GetVectorData` explicitly returns `nullptr`.

driver/conv_driver.hpp
```diff
@@ -1470,130 +1671,120 @@ int ConvDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
...
         if(is_fwd || is_bwd)
-            dumpBufferToFile<Tgpu>("dump_wei.bin", wei.data.data(), wei_sz);
+            dumpBufferToFile<Tgpu>("dump_wei.bin", wei.GetVectorData(), wei_sz);
         if(inflags.GetValueInt("bias") != 0)
-            dumpBufferToFile<Tgpu>("dump_bias.bin", b.data.data(), b.data.size());
-
+            dumpBufferToFile<Tgpu>("dump_bias.bin", b.GetVectorData(), b.GetVectorSize());
         if(is_bwd || is_wrw)
-            dumpBufferToFile<Tgpu>("dump_dout.bin", dout.data.data(), out_sz);
+            dumpBufferToFile<Tgpu>("dump_dout.bin", dout.GetVectorData(), out_sz);
```
CAHEK7, line 1751: That may also segfault for `is_gpualloc`.

driver/CMakeLists.txt
atamazov: @junliume Thanks, I'll work on it. FYI with ci_d868bf everything is fine.

driver/rocrand_wrapper.hpp
```diff
@@ -0,0 +1,65 @@
...
+namespace gpumemrand {
+
+int gen_0_1(double* buf, size_t sz);
+int gen_0_1(float* buf, size_t sz);
+int gen_0_1(half_float::half* buf, size_t sz);
+
+template <typename T>
+int gen_0_1(T* buf, size_t sz)
+{
+    std::cout << "Warning: gpumemrand functions are supported only for double, float and half. GPU "
```
CAHEK7, line 55: But there is a `uniform_int_distribution`. I mean it's not for a (0;1] range, but the range can be adjusted for integers.

CAHEK7 (dismissed):

driver/conv_driver.hpp
atamazov: @CAHEK7 Currently objects are constructed before options are parsed, i.e. we do not know if gpualloc mode is set.

driver/conv_driver.hpp
atamazov: @CAHEK7 IIRC we do not have GPU-based verification yet. Yes, we compute reference on GPU, but buffers are copied back to the host for rms computation. But maybe I am wrong. And yes, this functionality can be added later.

driver/conv_driver.hpp
atamazov: @CAHEK7 Let's do it later, https://github.com/ROCm/MIOpen/pull/2785#discussion_r1516258252

driver/rocrand_wrapper.hpp
atamazov: @CAHEK7 gpualloc is needed for fp32 and fp16 for now.

driver/conv_driver.hpp
CAHEK7: Even in case of cpu-based verification - we can generate input data on gpu, generate reference data on gpu and copy only result back to CPU for rms computation. I think it makes sense.
But I completely agree that it should be done in separated PRs.

driver/CMakeLists.txt
junliume: > @junliume Thanks, I'll work on it. FYI with ci_d868bf everything is fine.

@atamazov  then likely manually installing `rocrand` in the docker has messed up the library.

atamazov: @junliume The reason of the issue is `-DMIOPEN_BACKEND=HIPNOGPU`. Which seems logical (as rocrand uses GPU). I am going to disable building the driver for the HIPNOGPU backend (because what's the use for the driver without GPU?)

driver/conv_driver.hpp
atamazov: @CAHEK7 Yes, but not this time please. Previously, error codes here were ignored, and I do not touch this part of the functionality. Experience suggests that correct error code handling may require substantial effort and time. If this is worth a ticket, then please open it ;)

driver/conv_driver.hpp
atamazov: @CAHEK7 No, we'll get a message:
```
Ignored input file <FILENAME> - no host buffer to copy data into (nullptr)
```

driver/conv_driver.hpp
atamazov: @CAHEK7 https://github.com/ROCm/MIOpen/pull/2785#discussion_r1518058280

driver/conv_driver.hpp
atamazov: @CAHEK7 https://github.com/ROCm/MIOpen/pull/2785#discussion_r1518058280

driver/conv_driver.hpp
atamazov: @CAHEK7 
> AllocOnHost does not allocate any buffers

`GpumemTensor` and `GpumemVector` were intentionally designed to ignore any host-buffer related actions in gpumem mode. This is to minimize changes in the bulk of the driver. With this in mind, the no-op AllocOnHost no longer looks like a pitfall.

> GetVectorData explicitly returns nullptr.

This is done intentionally, to inform `readBufferFromFile` to do nothing.

driver/conv_driver.hpp
atamazov: @CAHEK7 It prints
```
Ignored ouput file <FILENAME> - no host buffer to dump data from (nullptr)
```

driver/conv_driver.hpp
atamazov: @CAHEK7 Done in c28d88037

driver/CMakeLists.txt
atamazov: @junliume Done in 671215b67

driver/conv_driver.hpp
atamazov: @CAHEK7 Yes. The goal of this PR is only introducing the new functionality. Before this change we had one big bucket of worms, now we have two smaller ones. Some progress as a by-product, right?

driver/conv_driver.hpp
atamazov: @CAHEK7 Well, actually we use only type of it. Identifier removed in 82e6ac961

atamazov:
@junliume @CAHEK7 Ready for the next review & CI testing round.

Reminder: if you are going to kick off CI, then please notice [#2785 (comment)](https://github.com/ROCm/MIOpen/pull/2785#issuecomment-1979107530). This `atamazov:ci_perf_test_performance` should be trivially mergeable into `ROCm:ci_perf_test_performance`.

driver/conv_driver.hpp
CAHEK7: Kind of, but why should we create two buckets of worms and a cup instead of getting rid of cup right now. Why we should introduce a new functionality which is kind of dangerous from the beginning?

driver/conv_driver.hpp
CAHEK7: In any case - here you ignored it explicitly, in the other place implicitly (without `std::ignore`). Could you make it consistent?

driver/conv_driver.hpp
CAHEK7: Ok

CAHEK7 (approved):

atamazov:
@JehandadKhan CI has passed and we have approval. Maybe we can re-enable regular perf testing when this PR is merged in (and maybe on each merge to develop -- depending on how long perf testing takes)

driver/conv_driver.hpp
atamazov: @CAHEK7 The danger with metaphors is that they can be easily misunderstood. This PR does not _create_ a bucket of worms. It sorts existing worms into buckets, making it easier to get rid of them in the future. The fewer worms in a bucket, the easier it is to notice them, and the easier it is to get rid of them (the easiest way is if each worm sits in its own bucket).

But maybe this metaphor is completely inappropriate for this case. AFAICS the driver needs to be redesigned, but that work is clearly out of the scope or this PR and does not fit into my mind anyway 🤷‍♂️ 

Thanks for reviewing.

driver/rocrand_wrapper.hpp
apwojcik, line 55: @atamazov If only the three prototypes above are available from Rocrand, we should report an error at compile time instead of runtime. We can go the fancy way or just use static_assert().

junliume: @atamazov it sounds like a minor revision, could you help to resolve it?

atamazov: @apwojcik If we use static_assert then the driver build would always fail because we instantiate the driver class for all datatypes.

rocRand is used to initialize input buffers due to:
https://github.com/ROCm/MIOpen/blob/82e6ac961018b9759c86e327b5dbf8a244076ab0/driver/conv_driver.hpp#L285-L291
However there is no real evidence that garbage or zeroes in the input buffers affect performance of our kernels. I tried FP16/FP32 kernels and do not see any effect. Well, it is _better_ to initialize, but we can live without that, most likely. That's why I use warning instead of error at run time.

apwojcik: Understood. No further comments.

junliume (approved):

PR #2787, opened 2024-03-05
Remove SetTensorLayout 

bghimireamd:
SetTensorLayout is being used in some places in MIOpen. This function is not necessary since tensor class's constructor takes care of this functionality.

### DIFF
--- a/test/gtest/cat.hpp
+++ b/test/gtest/cat.hpp
@@ -100,15 +100,6 @@ std::vector<CatTestCase> CatTestConfigs()
     // clang-format on
 }
 
-inline int32_t SetTensorLayout(miopen::TensorDescriptor& desc)
-{
-    std::vector<std::size_t> lens = desc.GetLengths();
-    std::vector<int32_t> int32_t_lens(lens.begin(), lens.end());
-
-    // set the strides for the tensor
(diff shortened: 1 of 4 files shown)


CAHEK7 (approved):
Looks good.

But this particular fix shows up a potential problem of using `SetTensorNd` - it may truncate tensor length to a signed 32 bits value while under the hood it is unsigned 64 bits value.

Tensor helper functions must be revisited and adjusted to properly handle big tensors.

seungmanhan (approved):
I didn't know there was a SetTensorLayout() that wasn't deleted, Look good to me.

bghimireamd:
> `SetTensorNd` - it may truncate tensor length to a signed 32 bits value while under the hood it is unsigned 64 bits value.

Created Issue for this https://github.com/ROCm/MIOpen/issues/2790

amberhassaan (approved):
Looks good. @bghimireamd As a follow up, we should remove many of the constructors in `tensor_holder.hpp` that don't take `layout` type as a parameter. In my mind, one should always specify the layout. So the `tensor` constructor in `tensor_holder` should be either taking `(layout, dims)` or `(layout, dims, strides)`.

junliume (approved):

PR #2793, opened 2024-03-07
Fixing linking issue for sqlite2txt on Windows

urpetkov-amd:
Fixing linking issue for Windows which is related to previously merged PR #2773 .

Adding `pthread` and `dl` to target link libraries caused blockage on Windows build with linking issue because `thread.lib` and `dl.lib` are not supported on Windows. 

Replaced `pthread` with `Threads::Threads` and added support for `dl` library for Linux.

### DIFF
--- a/tools/sqlite2txt/CMakeLists.txt
+++ b/tools/sqlite2txt/CMakeLists.txt
@@ -2,5 +2,10 @@ add_executable(sqlite2txt
         main.cpp
 )
 
-target_link_libraries(sqlite2txt SQLite::SQLite3 pthread dl)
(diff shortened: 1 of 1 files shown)


arvindcheru (commented):
Thank you for the sharing the observation from Windows, and the changes proposed.
A query, Do we need only new line added, or shall we enable based on the platform like below if both required?
if (WIN32)
    target_link_libraries(sqlite2txt SQLite::SQLite3 Threads::Threads)
else()
    target_link_libraries(sqlite2txt dl)
endif()

atamazov:
@DrizztDoUrden fyi

junliume (approved):

PR #2794, opened 2024-03-07
RNN back weights update

shurale-nkn:
Changed reduction kernel to CK.

Backward Weights time table

image

40-50% less time

### DIFF
--- a/driver/lstm_verify_gemm.hpp
+++ b/driver/lstm_verify_gemm.hpp
@@ -493,7 +493,7 @@ void RunLSTMForwardGEMMCPUVerify(miopenHandle_t handle,
                     else
                     {
                         size_t prec_shift = li * batch_n * hy_stride +
-                                         (bacc - in_n.at(ti - 1)) * hy_stride + bi * 4 * hy_h;
+                                            (bacc - in_n.at(ti - 1)) * hy_stride + bi * 4 * hy_h;
 
                         hid_state.at(hid_shift + (bacc + bs) * hy_stride + bi * 4 * hy_h + h) +=
                             activfunc(hid_state.at(hid_shift + (bacc + bs) * hy_stride + hy_h + h),
@@ -1069,7 +1069,7 @@ void RunLSTMBackwardDataGEMMCPUVerify(
                         if(ti < seqLength - 1)
                         {
                             size_t pretime_shift = li * batch_n * hy_stride +
-                                                (baccbi - in_n[seqLength - 2 - ti]) * hy_stride;
+                                                   (baccbi - in_n[seqLength - 2 - ti]) * hy_stride;
 
                             dh_state[hid_shift + (baccbi + bs) * hy_stride + bi * 4 * hy_h + hy_h +
                                      h] +=
@@ -1124,7 +1124,7 @@ void RunLSTMBackwardDataGEMMCPUVerify(
                             {
                                 size_t pretime_shift =
                                     li * batch_n * hy_stride +
-                                                    (baccbi + in_n[seqLength - 1 - ti]) * hy_stride;
+                                    (baccbi + in_n[seqLength - 1 - ti]) * hy_stride;
 
                                 dh_state[hid_shift + (baccbi + bs) * hy_stride + 5 * hy_h + h] +=
                                     dh_state[hid_shift + (baccbi + bs) * hy_stride + bi * 4 * hy_h +
@@ -1338,7 +1338,7 @@ void RunLSTMBackwardWeightGEMMCPUVerify(std::vector<Tgpu>& in,
                                         bool hx_is_null = false)
 {
     size_t batch_n = sumvc(in_n);
-    int numlayer = bidirection ? hy_d / 2 : hy_d;
+    int numlayer   = bidirection ? hy_d / 2 : hy_d;
     size_t bacc; // accumulation of batch
     int bi = bidirection ? 2 : 1;
 
@@ -1401,7 +1401,7 @@ void RunLSTMBackwardWeightGEMMCPUVerify(std::vector<Tgpu>& in,
     }
 
     size_t wei_shift_bias = (in_h + hy_h + (bi * hy_h + hy_h) * (numlayer - 1)) * wei_stride;
-    int wei_len        = wei_shift_bias;
+    int wei_len           = wei_shift_bias;
     if(biased)
     {
         int in_bias = 2;
--- a/driver/lstm_verify_gemm.hpp
+++ b/driver/lstm_verify_gemm.hpp
@@ -38,10 +38,10 @@ void RunLSTMForwardGEMMCPUVerify(miopenHandle_t handle,
                                  bool hx_is_null = false,
                                  bool cx_is_null = false)
 {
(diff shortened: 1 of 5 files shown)


junliume:
@shurale-nkn LSTM related unit tests are failing: [CI log](http://micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLIBS/pipelines/MIOpen/branches/rnn_bwweights/runs/4/nodes/1282/log/?start=0)
it might be CK kernel issue, please issue tickets to kernel folks in that case

shurale-nkn:
> @shurale-nkn LSTM related unit tests are failing: [CI log](http://micimaster.amd.com/blue/rest/organizations/jenkins/pipelines/MLLIBS/pipelines/MIOpen/branches/rnn_bwweights/runs/4/nodes/1282/log/?start=0) it might be CK kernel issue, please issue tickets to kernel folks in that case

 Yes segfault because of bug in CK. Our reduction kernel not able to work with unaligned pointer. Made WA in RNN

driver/lstm_verify_gemm.hpp
```diff
@@ -141,8 +141,8 @@ void RunLSTMForwardGEMMCPUVerify(miopenHandle_t handle,
     // forward emulator
     for(int li = 0; li < numlayer; li++)
     {
-        int hid_shift = li * batch_n * hy_stride;
-        int hx_shift  = li * in_n.at(0) * h_stride;
+        size_t hid_shift = li * batch_n * hy_stride;
```
atamazov, line 144: [Informative] @shurale-nkn Just to remind: if batch_n were not of the size_t type, then changing the type of hid_shift from int to size_t won't prevent overflow (in this line), because li and hy_stride are of type int. But now, thanks the result of `(li * batch_n * hy_stride)` is of type size_t, you can use `auto` in the definition of hid_shift.

driver/lstm_verify_gemm.hpp
```diff
@@ -141,8 +141,8 @@ void RunLSTMForwardGEMMCPUVerify(miopenHandle_t handle,
...
+        size_t hx_shift  = li * in_n.at(0) * h_stride;
```
atamazov, line 145: [Question] @shurale-nkn Just curious: what would be the the type of `(li * in_n.at(0) * h_stride)` if Tgpu is `float`? I am not aware of the real-life values of li, in_at(0) and h_stride, but operation like `(int * float * int)` may lead to loss of precision due to rounding (I suspect that the type of `int * float` is `float`, so all the computations are happen in float domain).

src/rnn/rnn_util.cpp
```diff
@@ -27,9 +27,19 @@
 #include <miopen/rnn.hpp>
 #include <miopen/rnn_util.hpp>
 #include <algorithm>
+#include <miopen/env.hpp>
+
+MIOPEN_DECLARE_ENV_VAR_UINT64(MIOPEN_RNNWRW_REDUCTION)
```
atamazov, line 32: [Recommendation] @shurale-nkn I suspect that
```C++
MIOPEN_DECLARE_ENV_VAR(MIOPEN_RNNWRW_REDUCTION, uint64_t, 1)
```
should define cached variable with default value 1.

atamazov:
@shurale-nkn Is there a fallback from CK to whatever is was before for old GPU types (unsupported by CK)?

driver/lstm_verify_gemm.hpp
atamazov: And I am not even asking about case when Tgpu is half ;)

driver/lstm_verify_gemm.hpp
shurale-nkn: > [Informative] @shurale-nkn Just to remind: if batch_n were not of the size_t type, then changing the type of hid_shift from int to size_t won't prevent overflow (in this line), because li and hy_stride are of type int. And now, as the result of `(li * batch_n * hy_stride)` is of type size_t, so you can use `auto` in the definition of hid_shift.

@atamazov as long as we using rocblas with offset in int we can and will overflow, This is not an attempt to fix something. it's just that I've made changes locally for some variables. Globally, I know that this code does not support large values, but changing this will require not only correct type conversion.

driver/lstm_verify_gemm.hpp
shurale-nkn: > [Question] @shurale-nkn Just curious: what would be the the type of `(li * in_n.at(0) * h_stride)` if Tgpu is `float`? I am not aware of the real-life values of li, in_at(0) and h_stride, but operation like `(int * float * int)` may lead to loss of precision due to rounding (I suspect that the type of `int * float` is `float`, so all the computations are happen in float domain).

...in_n is a vector of int . This is stride in number of elements. And it used only as index

src/rnn/rnn_util.cpp
shurale-nkn: Ok, but in next commit or in next PR. Depending on CI result

shurale-nkn:
> @shurale-nkn Is there a fallback from CK to whatever is was before for old GPU types (unsupported by CK)?

This does not follow from the question, but my answer will be: I use the standard MIOpen API. It should work on everything that supports MIOpen.

driver/lstm_verify_gemm.hpp
atamazov: @shurale-nkn Ah, I see, my mistake. So this is `(int * int * int)`, which should fit into `int` if the multiply_dims_overflow_assumption is valid:
https://github.com/ROCm/MIOpen/blob/d069763dfd7fa9fd53c3bbfd0513c5c572333c05/src/solver/pooling/forwardNaive.cpp#L125-L136

So you just would like to change the type of the result to size_t due to some reason.
[resolved]

shurale-nkn: Yes, don't worry too much about this file. His code coverage is enough for our solution, there are many problems, but so far it is enough.

atamazov:
@shurale-nkn 
> This does not follow from the question, but my answer will be: I use the standard MIOpen API. It should work on everything that supports MIOpen.

I've asked because PR description says: "_Changed reduction kernel to CK_" but I know that CK kernels are often not supported for old targets.

driver/lstm_verify_gemm.hpp
atamazov: @shurale-nkn In other words, you've changed `int` to `size_t` in many places due to some reason that is not related to fixing potential or real overflows. Okay.
[resolved]

atamazov:
@shurale-nkn Please hide resolved review comments to avoid littering this page, thanks.

shurale-nkn:
> @shurale-nkn
> 
> > This does not follow from the question, but my answer will be: I use the standard MIOpen API. It should work on everything that supports MIOpen.
> 
> I've asked because PR description says: "_Changed reduction kernel to CK_" but I know that CK kernels are often not supported for old targets.

There are many types of CK, this one specifically is contained in the MIOpen source and should work, but for any case, there are 3 alternatives in PR code to maintain functionality.

junliume (approved):

PR #2769, opened 2024-02-23
Fixed incorrectly generated files

DrizztDoUrden:
This is a hotfix for issue described here: https://github.com/ROCm/MIOpen/pull/2722#issuecomment-1961900323

3 things are fixed:
- text db files now contain correct keys
- any updates to sqlite2txt now force regeneration of db.txt files
- text user dbs now have a separate extension

PR #2800, opened 2024-03-12
[HotFix] Changed text perfdbs to be actually installed when enabled #2722

DrizztDoUrden:
Sqlite perfdb files were installed instead of txt files even when not enabled


cderb (approved):

junliume (approved):

atamazov:
This is continuation of #2722 and #2769.

PR #2802, opened 2024-03-12
[HotFix] Fix unpackdb after merging #2800

apwojcik:
PR #2800 broke the configuration of a CMake build on Windows.

PR #2803, opened 2024-03-12
[HotFix] fix unpackdb issue in cmakefile

junliume:
after #2800

fix the issue:
```
CMake Error at CMakeLists.txt:566 (unpack_db):
  unpack_db Function invoked with incorrect arguments for function named:
  unpack_db
  CMakeLists.txt:585 (install_kdb)

```


CMakeLists.txt
```diff
@@ -563,7 +563,7 @@ function(install_kdb FILE_NAME COMPONENT_NAME)
     if(__contents MATCHES "version")
         list(APPEND LFS_MISSING_FILES ${__fname})
     else()
-        unpack_db(${FILE_NAME})
+        unpack_db(${FILE_NAME} FILE_SUFFIX)
```
atamazov, line 566: [Recommendation]
```suggestion
        unpack_db(${FILE_NAME} UNUSED_FILE_SUFFIX)
```

atamazov: The design of this stuff is a bit clumsy, so I would also recommend adding a comment after line 532:
```
if(NOT MIOPEN_USE_SQLITE_PERFDB AND __extension STREQUAL ".db")
```
that should inform the reader that this branch is dedicated to a special case of unpacking ~system-find-db~ **_system-perf-db_** files converted to text format.

/cc @DrizztDoUrden

atamazov (approved):
LGTM albeit some recommendations

CMakeLists.txt
atamazov: UPD: system-perf-db files

junliume: @atamazov should we close this PR and use #2802 instead? That's my plan

PR #2804, opened 2024-03-12
Adapt logging functionality to Windows

apwojcik:
The clang compiler on Windows and MSVC does not recognize the GCC-specific `__PRETTY_FUNCTION__` preprocessor macro. The closest similar macro is `__FUNCSIG__`. Additionally, we use `std::string_view` for literals, which is 25% faster than `const char*` and requires no memory allocation. It also seamlessly supports `std::string`, `char*`, and `char[]`.

### DIFF
--- a/src/include/miopen/lock_file.hpp
+++ b/src/include/miopen/lock_file.hpp
@@ -42,6 +42,7 @@
 #include <mutex>
 #include <shared_mutex>
 #include <string>
+#include <string_view>
 
 namespace miopen {
 
@@ -75,7 +76,7 @@ public:
     }
     void lock()
     {
-        LockOperation("lock", MIOPEN_GET_FN_NAME(), [&]() { std::lock(access_mutex, flock); });
+        LockOperation("lock", MIOPEN_GET_FN_NAME, [&]() { std::lock(access_mutex, flock); });
     }
 
     void lock_shared()
@@ -83,7 +84,7 @@ public:
         access_mutex.lock_shared();
         try
         {
-            LockOperation("shared lock", MIOPEN_GET_FN_NAME(), [&]() { flock.lock_sharable(); });
+            LockOperation("shared lock", MIOPEN_GET_FN_NAME, [&]() { flock.lock_sharable(); });
         }
         catch(...)
         {
@@ -93,7 +94,7 @@ public:
 
     bool try_lock()
     {
-        return TryLockOperation("lock", MIOPEN_GET_FN_NAME(), [&]() {
+        return TryLockOperation("lock", MIOPEN_GET_FN_NAME, [&]() {
             return std::try_lock(access_mutex, flock) != 0;
         });
     }
@@ -104,7 +105,7 @@ public:
             return false;
 
         if(TryLockOperation(
-               "shared lock", MIOPEN_GET_FN_NAME(), [&]() { return flock.try_lock_sharable(); }))
+               "shared lock", MIOPEN_GET_FN_NAME, [&]() { return flock.try_lock_sharable(); }))
             return true;
         access_mutex.unlock();
         return false;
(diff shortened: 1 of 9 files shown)


apwojcik:
The build in CI is failing due to this error.
```
[2024-03-12T22:58:28.644Z] Error while preparing SQL statement: database disk image is malformed
[2024-03-12T22:58:28.644Z] Statement: {SELECT solver, params, spatial_dim, in_channels, in_h, in_w, in_d, fil_h, fil_w, fil_d, out_channels, batchsize, pad_h, pad_w, pad_d, conv_stride_h, conv_stride_w, conv_stride_d, dilation_h, dilation_w, dilation_d, bias, group_count, layout, data_type, direction FROM perf_db INNER JOIN config ON perf_db.config = config.id}
```

apwojcik:
```
[2024-03-12T22:58:29.075Z] make[2]: *** [CMakeFiles/generate_gfx90a68_db_txt.dir/build.make:74: share/miopen/db/gfx90a68.db.txt] Aborted (core dumped)
...
[2024-03-12T22:58:29.075Z] make[2]: *** [CMakeFiles/generate_gfx90878_db_txt.dir/build.make:74: share/miopen/db/gfx90878.db.txt] Aborted (core dumped)
...
[2024-03-12T22:58:29.075Z] make[2]: *** [CMakeFiles/generate_gfx90a6e_db_txt.dir/build.make:74: share/miopen/db/gfx90a6e.db.txt] Aborted (core dumped)
```

junliume:
> database disk image is malformed

@cderb I do see this error messages a few times, when I try to access the DB file from a docker, or access file from shared/mounted location. Usually it gets resolved by leveling up access permission.

CAHEK7:
It looks good, but ~~I'm curious, why `std::string_view` is faster than `const char*`?~~
~~I agree that it is faster than `const std::string&` - that was the change in this PR and `const std::string&` requires memory allocation.~~
Ok, because `std::string_view` have its length and does not require `/0` support. Lots of string manipulation algorithms can have benefits or simplifications or even explicit constant propagation optimizations.

apwojcik:
> It looks good, but I'm curious, why `std::string_view` is faster than `const char*`? I agree that it is faster than `const std::string&` - that was the change in this PR and `const std::string&` requires memory allocation.

@CAHEK7 The purpose of `std::string_view` is to avoid copying data owned by another instance when only a non-mutable view is required (99% of cases as a function parameter). It is faster than 'const char *' because there is no need for a length to be calculated. On the other hand, before performing any operation with `const char*`, you must calculate the length each time. You can read more here: https://stackoverflow.com/a/40129198 (both pros and cons).

In modern C++ code, `std::string_view` should replace nearly all uses of `const std::string&` as function parameters.

General rules for using `std::string_view`: function argument always pass by copy or const copy, never call the constructor on literals - use conversion operator instead (e.g., `auto s = "ABC"sv`; ) or let the compiler automatically construct from a literal.

apwojcik:
@CAHEK7 This is a great example of optimization with `std::string_view` (from updated env.hpp):
```c++
constexpr std::string_view LoggingParseFunction(const std::string_view func,
                                                const std::string_view pretty_func)
{
    if(func != "operator()")
        return func;
    // lambda
    const auto pf_tail = pretty_func.substr(0, pretty_func.find_first_of('('));
    return pf_tail.substr(1 + pf_tail.find_last_of(':'));
}
```
Both arguments are literal, and both calls to the `substr()` are O(1). Furthermore, the whole function is executed at compile time—there is no runtime execution for that function (the way it is used currently).

CAHEK7 (approved):
@apwojcik thanks for the detailed explanation!

atamazov (approved):
🟢 LGTM!
@apwojcik Thanks for the hi-quality patches, Artur!

DrizztDoUrden (approved):

junliume (approved):

PR #2805, opened 2024-03-12
Use std::vector<char> for binary blobs

apwojcik:
This PR is because MIOpen uses binary data with streams (e.g., save to a file). On Windows, when `std::string` is used with a stream, various translation operations happen (depending on a stream mode set, e.g., converting `\n` to `\r\n`, code page or character encoding conversions, etc.). That results in the corruption of data. By changing to `std::vector<char>`, we overcome this problem.
It is ok to use `std::string` for binary data in general. However, at least on Windows, this data type can hold only up to 1.9GB of data.


src/load_file.cpp
```diff
@@ -1,17 +1,52 @@
...
-    const std::ifstream t(s);
-    std::stringstream buffer;
-    buffer << t.rdbuf();
-    return buffer.str();
+    std::ifstream in(path, std::ios::binary | std::ios::ate);
+    if(!in.is_open())
+        MIOPEN_THROW("Unable to open file: " + path.string());
+    in.unsetf(std::ios::skipws);
+    auto size = in.tellg();
+    in.seekg(0, std::ios::beg);
```
CAHEK7: I think `auto size = fs::file_size(path)` is better.

src/sqlite_db.cpp
```diff
@@ -349,12 +349,13 @@ std::string SQLite::Statement::ColumnText(int idx)
         reinterpret_cast<const char*>(sqlite3_column_text(pImpl->ptrStmt.get(), idx)), bytes};
 }
 
-std::string SQLite::Statement::ColumnBlob(int idx)
+std::vector<char> SQLite::Statement::ColumnBlob(int idx)
 {
-    auto ptr = sqlite3_column_blob(pImpl->ptrStmt.get(), idx);
+    auto ptr = reinterpret_cast<const char*>(sqlite3_column_blob(pImpl->ptrStmt.get(), idx));
```
CAHEK7:
```suggestion
    auto ptr = static_cast<const char*>(sqlite3_column_blob(pImpl->ptrStmt.get(), idx));
```
void* -> any* can be casted directly, without reinterpretation.

CAHEK7 (commented):
Minor suggestions. Also have you tried to use `std::byte` instead of char?
Potentially it's a more suitable type for binary blobs.

junliume:
@apwojcik submodule MIFin needs to be updated with this PR

src/load_file.cpp
apwojcik: I prefer to keep `tellg()`—the file is already opened, and with `file_size`, on Linux, uses `stat` to read `st_size`, but Windows uses the `GetFileSize()` function, which requires opening the file first. The size is only required to estimate roughly how much memory we need to preallocate. It doesn't have to be precise, but for the binary stream, `tellg()` returns the number of bytes from the beginning of the file (both Linux and Windows). It is not valid for other stream modes on Windows.

CAHEK7: I'm no fully convinced.
`GetFileSize()` requires file handle, which can be just an identifier to get all the file information from the filesystem, while `std::ifstream in(path, std::ios::binary | std::ios::ate);` really opens the file and rewinds its internal state to the end of the file (which can be potentially zero-cost, but who knows). Then you need to rewind file back (which is also can be zero-cost, but so `GetFileSize()` can be). Moreover, MSDN says:

> You can use SetFilePointerEx to determine the length of a file. To do this, use FILE_END for dwMoveMethod and seek to location zero. The file offset returned is the length of the file. However, this practice can have unintended side effects, such as failure to save the current file pointer so that the program can return to that location. It is simpler and safer to use the [GetFileSizeEx](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfilesizeex) function instead.

Nevertheless the code with rewinding looks unclear. Especially the `std::ios::skipws` part which does not make much sense since the filepointer is already at the end, but this flag potentially alters stream state and may affect reading from it (probably not for the binary stream).

apwojcik: I ran quick performance measurements on Linux and Windows with `gtell()` and `file_size()`. The test reads 100 times the file of size 400MB. The results revealed that both methods are equally performant (approximately 10ms in favor of `file_size`).
However, I discovered that determining file size is not an issue here. The actual slowdown is copying data from file to vector. I changed to allocate the whole vector and used `in.read()`. I observed x3 faster file reading on Linux and x7 faster file reading on Windows. With that change, Windows reads files faster than Linux (for the mentioned test, ~300ms for Linux and ~225ms for Windows per single file read).

CAHEK7 (approved):

apwojcik:
> @apwojcik submodule MIFin needs to be updated with this PR

@junliume PR submitted ROCm/MIFin#101

src/bz2.cpp
```diff
@@ -55,15 +59,16 @@ void check_bz2_error(int e, const std::string& name)
...
-std::string compress(std::string s, bool* compressed)
+std::vector<char> compress(const std::vector<char>& v, bool* compressed)
 {
-    std::string result = s;
-    unsigned int len   = result.size();
-    auto e             = BZ2_bzBuffToBuffCompress(&result[0], &len, &s[0], s.size(), 9, 0, 30);
+    auto result      = v;
+    unsigned int len = result.size();
+    auto e           = BZ2_bzBuffToBuffCompress(
+        result.data(), &len, const_cast<char*>(v.data()), v.size(), 9, 0, 30);
```
junliume, line 68: @apwojcik HIP tidy is complaining about this line  
error: do not use const_cast [cppcoreguidelines-pro-type-const-cast,-warnings-as-errors]

src/bz2.cpp
```diff
@@ -72,11 +77,12 @@ std::string compress(std::string s, bool* compressed)
...
 
-std::string decompress(std::string s, unsigned int size)
+std::vector<char> decompress(const std::vector<char>& v, unsigned int size)
 {
-    std::string result(size, 0);
+    std::vector<char> result(size, 0);
     unsigned int len = result.size();
-    auto e           = BZ2_bzBuffToBuffDecompress(&result[0], &len, &s[0], s.size(), 0, 0);
+    auto e           = BZ2_bzBuffToBuffDecompress(
+        result.data(), &len, const_cast<char*>(v.data()), v.size(), 0, 0);
```
junliume, line 88: Ditto

src/bz2.cpp
CAHEK7: I guess HIP clang-tidy ignores fully-qualified exception `// NOLINTBEGIN(cppcoreguidelines-pro-type-const-cast)`
For me only global exception works, just `// NOLINTBEGIN`.

junliume (approved):

PR #2810, opened 2024-03-14
[Tests] Convert three regression tests to gTests

xinlipn:
Three new gTests created  from cTests, ie.

1.test_regression_half_vega
2.test_regression_issue_2012
3.test_regression_half_mi200


test/gtest/regression_half_mi200.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include "get_handle.hpp"
+
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS)
+
+namespace regression_half_mi200 {
+void SetupEnvVar(void)
+{
+    miopen::UpdateEnvVar(ENV(MIOPEN_FIND_MODE), std::string("normal"));
+    miopen::UpdateEnvVar(ENV(MIOPEN_DEBUG_FIND_ONLY_SOLVER), std::string("1"));
```
CAHEK7: Should it be like this?
```suggestion
    miopen::UpdateEnvVar(ENV(MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS), 1);
```

test/gtest/regression_half_mi200.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void GetArgs(const std::string& param, std::vector<std::string>& tokens)
+{
+    std::stringstream ss(param);
+    std::istream_iterator<std::string> begin(ss);
+    std::istream_iterator<std::string> end;
+    while(begin != end)
+        tokens.push_back(*begin++);
+}
```
CAHEK7:
```suggestion
std::vector<std::string> GetArgs(const std::string& param)
{
    std::stringstream ss(param);
    std::istream_iterator<std::string> begin(ss);
    std::istream_iterator<std::string> end;
    return {begin, end};
}
```

test/gtest/regression_issue_2012.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <miopen/env.hpp>
+#include "get_handle.hpp"
+
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS)
+
+namespace regression_issue_2012 {
+void SetupEnvVar(void)
+{
+    miopen::UpdateEnvVar(ENV(MIOPEN_DEBUG_FIND_ONLY_SOLVER), std::string("1"));
```
CAHEK7:
```suggestion
    miopen::UpdateEnvVar(ENV(MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS), 1);
```

test/gtest/regression_half_vega.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void GetArgs(const std::string& param, std::vector<std::string>& tokens)
+{
+    std::stringstream ss(param);
+    std::istream_iterator<std::string> begin(ss);
+    std::istream_iterator<std::string> end;
+    while(begin != end)
+        tokens.push_back(*begin++);
+}
```
CAHEK7: See half_mi200 suggestion.

test/gtest/regression_issue_2012.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+void GetArgs(const std::string& param, std::vector<std::string>& tokens)
+{
+    std::stringstream ss(param);
+    std::istream_iterator<std::string> begin(ss);
+    std::istream_iterator<std::string> end;
+    while(begin != end)
+        tokens.push_back(*begin++);
+}
```
CAHEK7: See half_mi200 suggestion.

test/gtest/regression_half_vega.cpp
```diff
@@ -0,0 +1,114 @@
...
+std::vector<std::string> GetTestCases(void)
+{
+    const std::string& cmd       = "test_conv3d ";
+    const std::string& float_arg = miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG));
+    const std::string& conv_verbose_b =
+        float_arg + " --verbose --disable-forward --disable-backward-weights";
+
+    // clang-format off
+    return std::vector<std::string>{
+        {cmd + conv_verbose_b + " --input 64 256 28 28 --weights 64  64  1 1 --pads_strides_dilations 0 0 1 1 1 1 --group-count 4"}
```
CAHEK7:
```suggestion
add_custom_test(test_regression_half_vega SKIP_UNLESS_ALL FLOAT_DISABLED HALF_ENABLED GFX908_DISABLED GFX90A_DISABLED
# Issue #1956.
    ENVIRONMENT MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER='GemmBwdRest'
    COMMAND $<TARGET_FILE:test_conv3d> ${TEST_CONV_VERBOSE_B} --cmode conv --pmode default --group-count 1 --batch_size 2 --input_channels 64 --output_channels 32 --spatial_dim_elements 128 128 128 --filter_dims 3 3 3 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW ${MIOPEN_TEST_FLAGS_ARGS}
)
```

test/gtest/regression_issue_2012.cpp
```diff
@@ -0,0 +1,118 @@
...
+    using namespace miopen::debug;
+    using e_mask = enabled<Gpu::Default>;
+    using d_mask = disabled<Gpu::gfx900, Gpu::gfx906, Gpu::gfx90A>;
+    return ::IsTestSupportedForDevMask<d_mask, e_mask>();
+}
+
+void Run2dDriver(void)
+{
+    if(!(IsTestSupportedForDevice() &&
+         miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG)) == "--float"))
```
CAHEK7, line 82: It should not be here. issue 2012 does not have `float` limitation.

test/gtest/regression_issue_2012.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "../conv2d.hpp"
+#include <miopen/miopen.h>
+#include <gtest/gtest_common.hpp>
+#include <gtest/gtest.h>
+#include <miopen/env.hpp>
+#include "get_handle.hpp"
+
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS)
+
```
CAHEK7:
```suggestion
```

test/gtest/regression_half_mi200.cpp
```diff
@@ -0,0 +1,114 @@
...
+ *******************************************************************************/
+#include "../conv2d.hpp"
+#include <miopen/miopen.h>
+#include <gtest/gtest_common.hpp>
+#include <gtest/gtest.h>
+#include <miopen/env.hpp>
+#include "get_handle.hpp"
+
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS)
```
CAHEK7:
```suggestion
```

test/gtest/regression_half_vega.cpp
```diff
@@ -0,0 +1,114 @@
...
+    using d_mask = disabled<Gpu::gfx908, Gpu::gfx90A>;
+    return ::IsTestSupportedForDevMask<d_mask, e_mask>();
+}
+
+void Run2dDriver(void)
+{
+    if(!(IsTestSupportedForDevice()                      //
+         && (miopen::IsUnset(ENV(MIOPEN_TEST_ALL))       // standalone run
+             || (miopen::IsEnabled(ENV(MIOPEN_TEST_ALL)) // or --float full tests enabled
+                 && miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG)) == "--half"))))
```
CAHEK7, line 83: Please double check this condition.
At least `miopen::IsEnabled(ENV(MIOPEN_TEST_ALL))` must not skip the test, it must enable it.
As we as `miopen::GetStringEnv(ENV(MIOPEN_TEST_FLOAT_ARG)) == "--half"`

CAHEK7 (changes requested):

test/gtest/regression_half_mi200.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+using TestCase = decltype(GetTestCases())::value_type;
+
+class ConfigWithHalf_regression_half_mi200 : public testing::TestWithParam<std::vector<TestCase>>
+{
+};
+
+bool IsTestSupportedForDevice()
+{
+    using namespace miopen::debug;
```
CAHEK7:
```suggestion
```

test/gtest/regression_half_vega.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+using TestCase = decltype(GetTestCases())::value_type;
+
+class ConfigWithHalf_regression_half_vega : public testing::TestWithParam<std::vector<TestCase>>
+{
+};
+
+bool IsTestSupportedForDevice()
+{
+    using namespace miopen::debug;
```
CAHEK7:
```suggestion
```

test/gtest/regression_issue_2012.cpp
```diff
@@ -0,0 +1,118 @@
...
+using TestCase = decltype(GetTestCases())::value_type;
+
+class ConfigWithFloat_regression_issue_2012 : public testing::TestWithParam<std::vector<TestCase>>
+{
+};
+
+bool IsTestSupportedForDevice()
+{
+    using namespace miopen::debug;
```
CAHEK7:
```suggestion
```

test/CMakeLists.txt
```diff
@@ -1020,52 +1020,6 @@ add_custom_test(smoke_solver_ConvMlirIgemm_W GFX900_DISABLED GFX908_DISABLED GFX
...
     COMMAND $<TARGET_FILE:test_conv2d> ${TEST_CONV_VERBOSE_W} --input 64 64  28 28 --weights 64  64  1 1 --pads_strides_dilations 0 0 1 1 1 1 ${MIOPEN_TEST_FLAGS_ARGS}
 )
-
-add_custom_test(test_regression_half_vega SKIP_UNLESS_ALL FLOAT_DISABLED HALF_ENABLED GFX908_DISABLED GFX90A_DISABLED
-# Issue #1956.
-    ENVIRONMENT MIOPEN_FIND_MODE=normal MIOPEN_DEBUG_FIND_ONLY_SOLVER='GemmBwdRest'
-    COMMAND $<TARGET_FILE:test_conv3d> ${TEST_CONV_VERBOSE_B} --cmode conv --pmode default --group-count 1 --batch_size 2 --input_channels 64 --output_channels 32 --spatial_dim_elements 128 128 128 --filter_dims 3 3 3 --pads_strides_dilations 1 1 1 1 1 1 1 1 1 --trans_output_pads 0 0 0 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW ${MIOPEN_TEST_FLAGS_ARGS}
-)
-# Test case for issue #1956 uses huge tensors, therefore:
-set_tests_properties(test_regression_half_vega PROPERTIES RUN_SERIAL On)
```
CAHEK7, line 1030: Hmm, interesting. That may be hard to support in gtest-parallel infrastructure. @atamazov @JehandadKhan  FYI

[it does not block this PR, just an observation]

test/gtest/regression_half_mi200.cpp
xinlipn: @CAHEK7 , Yes, thanks for picking it up. Using "true" to replace 1 probably is better to avoid casting

test/gtest/regression_issue_2012.cpp
xinlipn: Correction, MIOPEN_DEBUG_FIND_ONLY_SOLVER indeed needs to be replaced with MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS

test/gtest/regression_half_mi200.cpp
CAHEK7: "true" is probably fine, just need to check if we accept bool for `UpdateEnvVar`.

test/gtest/regression_issue_2012.cpp
CAHEK7: Please see the comment above, "1" should be string for the string `MIOPEN_DEBUG_FIND_ONLY_SOLVER` which must contain a solver name. For the boolean `MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS` it should be integer value.

xinlipn: > Please see the comment above, "1" should be string for the string `MIOPEN_DEBUG_FIND_ONLY_SOLVER` which must contain a solver name. For the boolean `MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS` it should be integer value.

Yes, that's correct. the MIOPEN_DEBUG_FIND_ONLY_SOLVER shouldn't be here and MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS should be assigned 1 or true.

test/gtest/regression_half_mi200.cpp
xinlipn: Hi @CAHEK7 , this is interface is indeed cleaner. One thing is the following line would need to be changed to 

https://github.com/ROCm/MIOpen/blob/a583373aea46ee7b3b3cc0e13ad1b331ae018e4a/test/gtest/regression_half_mi200.cpp#L94

e.g. 
std::vector<std::string> tokens = std::move(GetArgs(test_value));
In this case, the compiler would give a warning "moving a temporary object prevents copy elision warning for the following code"

Or, 
std::vector<std::string> tokens = GetArgs(test_value);

This copies the string to tokens unless the compiler performs copy elision silently. Do you think the original design is a little more efficient? Thanks

CAHEK7: Copy elision is a mandatory part of the standard and works perfect in simple cases like `return vector{};`
Thats why the compiler complains about `std::move(GetArgs(...))`.
So
```C++
std::vector<std::string> tokens = GetArgs(test_value);
```
Meantime I did some measurements, the version with copy elision is a bit faster.

test/gtest/regression_half_vega.cpp
xinlipn: @CAHEK7 , could you help clarify this comment? Newly added code seems to be for CMakeLists.txt. Thanks

CAHEK7: It's a reference from CMakeList, so you can compare the newly added testcase and the testcase from CMakeList.

test/gtest/regression_issue_2012.cpp
xinlipn: @CAHEK7 Please correct me if I'm wrong. Seems "--float" is the default

https://github.com/ROCm/MIOpen/blob/cee33a7aaa3dbbe850d2580a72fff8774ddbc131/test/CMakeLists.txt#L230

test/gtest/regression_half_vega.cpp
xinlipn: @CAHEK7 , this logic is a bit convoluted. Please refer to the discussion here. 
https://github.com/ROCm/MIOpen/pull/2545#discussion_r1478689194

test/gtest/regression_issue_2012.cpp
CAHEK7: Ah, it's a part of inverted condition. Ok

test/gtest/regression_half_vega.cpp
CAHEK7: It's very easy to mess with parenthesis and operations order in this condition.

CAHEK7:
FYI: https://github.com/ROCm/MIOpen/pull/2816
`MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS` is not required at all.

xinlipn:
> FYI: #2816 `MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS` is not required at all.

@CAHEK7 . Thanks for the info. MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS is removed from test_regression_issue_2012 but kept for 
test_regression_half_mi200. Thus, I have updated regression_issue_2012.cpp

https://github.com/ROCm/MIOpen/blob/20dcde0ed5a5f266aadcf1a75e4c460e33f326df/test/CMakeLists.txt#L1055

test/gtest/regression_issue_2012.cpp
xinlipn: @CAHEK7 Removed MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS

test/gtest/regression_half_mi200.cpp
xinlipn: @CAHEK7 Seems MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS is kept for test_regression_half_mi200 in test/CMakeLists.txt and no pending PR's removing it after a quick search.

CAHEK7:
> > FYI: #2816 `MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS` is not required at all.
> 
> @CAHEK7 . Thanks for the info. MIOPEN_DEBUG_IMPLICIT_GEMM_FIND_ALL_SOLUTIONS is removed from test_regression_issue_2012 but kept for test_regression_half_mi200. Thus, I have updated regression_issue_2012.cpp

I guess Artem just forgot to remove that env variable from the second test.
He entirely removed the variable from the convolution, and it does not affect anything at all. Let's see.

CAHEK7 (approved):
Looks good.

junliume (approved):

test/CMakeLists.txt
atamazov: @CAHEK7 @xinlipn @JehandadKhan Yes. The `RUN_SERIAL` tests should be either put into a different binary or selected with `--gtest-filter`. I suspect that in any case we'll need to run gtest-parallel twice, like this:
```bash
$ .../gtest-parallel miopen_gtest --gtest_filter=<REGEX-TO-EXCLUDE-ALL-SERIAL-TESTS>
$ .../gtest-parallel miopen_gtest --gtest_filter=<REGEX-TO-INCLUDE-ONLY-SERIAL-TESTS> --workers=1
```

atamazov, line 1030: 🐛 [Quiet change] The loss of `RUN_SERIAL` may lead to unstable behavior.

In test/gtest/CMakeLists.txt: When `MIOPEN_TEST_DISCRETE` is set to ON, for the test `test_regression_half_vega`, the `RUN_SERIAL` property must be set to ON. See also: https://cmake.org/cmake/help/latest/module/GoogleTest.html#command:gtest_discover_tests

atamazov (commented):
🔴 @xinlipn Post-merge review

/cc @CAHEK7

Issue #2808, opened 2024-03-13
[conv][fp16][bf16] Gemm solvers fail with big tensors


atamazov:
The most likely reason rocBLAS API limitations (some parameters are passed as ints, which means 2G limit). Another possible reason overflow due to is natural limitations of the FP16 datatype, but this is definitely not the case for BF16 (which fails as well with the same configs), so overflow is unlikely.

Currently the problem is avoided by old `WORKAROUND_MLOPEN_ISSUE_1430`. It makes Gemm solvers inapplicable by limiting WS size for non-FP32 types. The valid approach could be checking parameters of GEMM calls and making the solver inapplicable if they exceed INT_MAX.

[Notice] It is likely that the problem is inherent to FP32 as well, but I am unable to test this due to limited memory on the GPUs that I have on hand. `WORKAROUND_MLOPEN_ISSUE_1430` proactively protects Gemm solvers for this case but with twice bigger WS limit.

### Examples (ROCm 6.0.0, Navi21):

<details><summary><b><i>-n 1 -c 48 --in_d 256 -H 128 -W 128 -k 16 --fil_d 3 -y 3 -x 3</i></b></summary>

```bash
$ export MIOPEN_DEBUG_FIND_ONLY_SOLVER="GemmWrwUniversal;GemmBwdRest;GemmFwdRest"
$ export MIOPEN_WORKAROUND_ISSUE_2808=0
$ MIOPEN_FIND_MODE=1 ./bin/MIOpenDriver convfp16 -n 1 -c 48 --in_d 256 -H 128 -W 128 -k 16 --fil_d 3 -y 3 -x 3 --pad_d 1 -p 1 -q 1 \
--conv_stride_d 1 -u 1 -v 1 --dilation_d 1 -l 1 -j 1 --spatial_dim 3 -m conv -g 1 -F 7 -t 1 -i 1
Forward Convolution FAILED: 0.880373 > 0.0082
Backward Convolution Data Verifies OK on GPU reference (0.000102885 < 0.0082)
Backward Convolution Weights FAILED: 0.891237 > 0.0164
$ ./bin/test_conv3d --half --verbose --disable-verification-cache \
--batch_size 1 --input_channels 48 --output_channels 16 --spatial_dim_elements 256 128 128 --filter_dims 3 3 3 \
--pads_strides_dilations 1 1 1 1 1 1 1 1 1
FAILED: -nan
Max diff: nan
[...]

PR #2811, opened 2024-03-14
[conv][FP32] Extend applicability of GemmBwdRest and GemmFwdRest for big WS sizes

### LINKED ISSUE #2789: MAX_MEM_ALLOC_SZ limited to 7287183769
The hard-coded limit of MAX_MEM_ALLOC_SZ <= 7287183769 in https://github.com/ROCm/MIOpen/blob/release/rocm-rel-5.4/src/solver/gemm.cpp#L203 is detrimental as it limits the applicability of GemmFwdRest solvers. When the workspace size exceeds this limit, the search falls back on the naive direct conv which is much slower.
I can't see from issue #1430 why this limit needs to be hard-coded. Can it be removed? 

E.g. a 3d conv matching the key signature
`32-128-128-128-3x3x3-32-128-128-128-1-1x1x1-1x1x1-1x1x1-0-NCDHW-FP32-F`
on my gfx908 GPU gives 
```
MIOpen(HIP)   0.004: Info [FindConvFwdAlgorithm] FW Chosen Algorithm: GemmFwdRest , 7247757312, 52.4841
```

but small change to 
`33-128-128-128-3x3x3-33-128-128-128-1-1x1x1-1x1x1-1x1x1-0-NCDHW-FP32-F`
gives
```
MIOpen(HIP)   0.017: Info2 [GetWorkspaceSize] 7474249728 > 7287183769
...
MIOpen(HIP)   0.005: Info [FindConvFwdAlgorithm] FW Chosen Algorithm: ConvDirectNaiveConvFwd , 0, 2399.31
```

atamazov:
This change is NFC except that:
- Doubled workspace limit for the "last chance" FP32 GEMM Forward and Backward convolution solvers.
  - Under control of `MIOPEN_WORKAROUND_ISSUE_2789`, enabled by default.
  - This resolves #2789

NFC style/quality changes:
- Common code from GEMM solvers moved to the solver/gemm_common module.
- A bit better logging in Gemm's GetWorkspaceSize() funcs
- Formalized WORKAROUND_MLOPEN_ISSUE_1430.
  - See https://github.com/AMDComputeLibraries/MLOpen/issues/1430
- [driver] Added logging of hipMalloc/Free at Info2 level.

Manually tested. No new tests added since large configs are already covered by existing tests.

----
- https://github.com/ROCm/MIOpen/labels/workaround
- Proposed reviewers: @JehandadKhan @DrizztDoUrden @ritwickJ @sspencespence


src/solver/gemm_common.cpp
```diff
@@ -0,0 +1,112 @@
...
+    const auto m = h.GetMaxMemoryAllocSize();
+    auto limit   = std::numeric_limits<std::size_t>::max();
+
+#if WORKAROUND_MLOPEN_ISSUE_1430
+    const auto device = h.GetTargetProperties().Name();
+    if(device == "gfx906" || device == "gfx900")
+    {
+        // Workaround for MLOpen issue 1430. Vega20 fails to access GPU memory
+        // larger than the return value of GetMaxMemoryAllocSize() of Vega10.
+        limit = std::min(limit, static_cast<std::size_t>(7287183769));
```
CAHEK7:
```suggestion
        limit = std::min(limit, wa_2808_16bit_limit);
```

src/solver/gemm_bwd.cpp
```diff
@@ -228,9 +182,10 @@ size_t GemmBwd1x1_stride2::GetWorkspaceSize(const ExecutionContext& context,
...
 
-    if(gemm_trans > MAX_MEM_ALLOC_SZ)
+    if(gemm_trans > gemm::MaxMemAllocSz(handle, problem))
     {
-        MIOPEN_LOG_I2(gemm_trans << " > " << MAX_MEM_ALLOC_SZ);
+        MIOPEN_LOG_I2("GemmBwd1x1_stride2: " << gemm_trans << " > "
+                                             << gemm::MaxMemAllocSz(handle, problem));
         return 0;
     }
     return gemm_trans;
```
CAHEK7, line 191: What do you think?
`gemm_common.hpp:`
```C++
size_t GetMaxPossibleMem(size_t requested_size, auto handle, auto problem, std::string_view solver_name)
{
    auto max_possible_size = gemm::MaxMemAllocSz(handle, problem);
    if(requested_size > max_possible_size)
    {
        MIOPEN_LOG_I2(solver_name << ": " << requested_size << " > " << max_possible_size);
        return 0;
    }
    return requested_size;
}
```

```C++
return GetMaxPossibleMem(gemm_trans, handle, problem, "GemmBwd1x1_stride2");
```

src/solver/gemm.cpp
```diff
@@ -539,9 +494,10 @@ size_t GemmFwd1x1_0_1_int8::GetWorkspaceSize(const ExecutionContext& context,
                                          std::multiplies<std::size_t>()) *
                          GetTypeSize(wDesc.GetType()) * conv.group_count;
 
-    if(ws_size > MAX_MEM_ALLOC_SZ)
+    if(ws_size > gemm::MaxMemAllocSz(handle, problem))
```
CAHEK7, line 497: See gemm_bwd suggestion.

src/solver/gemm_wrw.cpp
```diff
@@ -376,9 +356,12 @@ size_t GemmWrwUniversal::GetWorkspaceSize(const ExecutionContext& context,
                                          std::multiplies<std::size_t>()) *
                          conv.group_count;
 
-    if(ws_size > MAX_MEM_ALLOC_SZ)
+    if(ws_size > gemm::MaxMemAllocSz(handle, problem))
```
CAHEK7, line 359: See gemm_bwd suggestion.

CAHEK7 (approved):
Some minor suggestions.

src/solver/gemm_common.cpp
atamazov: I want independent value here.

src/solver/gemm_bwd.cpp
atamazov: @CAHEK7 I can do that if you insist (but please do not, if you can ;)

atamazov: @CAHEK7 What I actually think is: returning 0 is a gemm-specific temporary hack that I would like to eliminate. In IsApplicable() we should get WS size and return false if it exceeds allowed max. I will do this while working on #2808.

atamazov: @CAHEK7 ...and I think that for fixing 2808, IsApplicable should also include imitation of rocblas launch (to detect overflows), while this gemm::MaxMemAllocSz() will de-evolute down to handle.GetMaxMemoryAllocSize().

src/solver/gemm_common.cpp
CAHEK7: Then declare it properly)
```
if (42 == status)
```

atamazov: There is comment that explains the semantics of the literal, and the literal is used only once.

BTW 42 has well-known meaning ))

atamazov: @CAHEK7 BTW are you insisting or just discussing?

atamazov: If the former, then I'll simply follow your recommendation

CAHEK7: I just want to have everything consistent.
All those values are used only once and instead of declaring them, `wa_2808_16bit_limit` can be used directly.
```
    constexpr auto wa_2808_other_limit    = wa_2808_16bit_limit * 2;
    constexpr auto wa_2809_fp32_wrw_limit = wa_2808_16bit_limit;
```
But this literal looks exactly the same as `wa_2808_16bit_limit` and it look suspicions even if it has a comment, because the comment does not clarify anything - is `7287183769` the limit of `GetMaxMemoryAllocSize()`? Why it is hardcoded if there is a function which returns that value? 
Why it says Vega20, then refers to Vega10?

atamazov: @CAHEK7 
> But this literal looks exactly the same as wa_2808_16bit_limit...

I am not changing existing limits because 
- We do not know the root reason of the 2808 and 2809
- Increasing the value may lead to correctness errors
- Decreasing the value may lead to performance issues

That's why only limit for FP32 Fwd/Bwd was changed (and carefully tested).

The values for targets other than Vega10/20 will be changed (and hopefully removed at all) after the full-blown fix.

WORKAROUND_MLOPEN_ISSUE_1430 will remain until we drop support for Vega20.

> Why it is hardcoded if there is a function which returns that value?
> Why it says Vega20, then refers to Vega10?

As comment explains, we found that "Vega20 fails to access GPU memory larger than the return value of GetMaxMemoryAllocSize() of Vega10". `7287183769` is the value that GetMaxMemoryAllocSize() returns on Vega10.

CAHEK7: I'm not asking about changing the numbers, I'm talking about magic numbers which are not clear from the code. And I'm talking about the fact that you named some of them and did not some others.
That explanation is great and useful, but it will stay here in the comments to the pull request, not in the code.

It's up to you to name it properly or leave everything as it is - I approved this PR since the beginning.

atamazov: @CAHEK7 it's your right (and even duty) to ask questions if something is unclear.

> And I'm talking about the fact that you named some of them and did not some others.

The reasons for not naming literal under WORKAROUND_MLOPEN_ISSUE_1430 are:
- It's hard to invent name that explains the semantics better than the existing comment
- the literal is used only once
- the w/a it is not intended to be fixed and will be removed together with gfx906 support.

Thanks for reviewing!

atamazov:
@junliume @CAHEK7 @JehandadKhan  Can you please restart CI? Thanks!

cderb (approved):

junliume:
@atamazov @cderb DB Sync tests have failed, and we should update DB because of this PR?

atamazov:
@junliume This shouldn't be necessary. Let me update from latest develop.

atamazov:
@junliume @CAHEK7 @JehandadKhan @cderb Updated. Please someone start CI.

atamazov:
@junliume 
> @atamazov @cderb DB Sync tests have failed, and we should update DB because of this PR?

If dbsync test is still failing, please provide me with a full log of that CI stage. Thanks!

sspencespence:
I have been trying to build and test but get an error during build like:
``` 
Scanning dependencies of target generate_gfx90a68_db
[  3%] Generating share/miopen/db/gfx90a68.db
tools/sqlite2txt/main.cpp:172:17: error: static assertion failed: unsupported type
                static_assert(false, "unsupported type");
                ^             ~~~~~
1 error generated.
make[2]: *** [tools/sqlite2txt/CMakeFiles/sqlite2txt.dir/build.make:82: tools/sqlite2txt/CMakeFiles/sqlite2txt.dir/main.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2254: tools/sqlite2txt/CMakeFiles/sqlite2txt.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
seems related. if, as you suggested, updating db isn't necessary, then do I need to build it and is there a way to skip it?

junliume:
[fail.log](https://github.com/ROCm/MIOpen/files/14858292/fail.log)
@atamazov here is the full log

atamazov:
@junliume Increasing the WS size limit for a GEMM solver of leads to wider applicability of it. But the applicability can only be widened for the "rest-of" kind GEMM solvers, since there are (legacy) dependencies between the applicability of GEMM solvers, for example:
https://github.com/ROCm/MIOpen/blob/37c8b9894b4c727e63fda308ffb896f31a21cd4d/src/solver/gemm.cpp#L1015-L1030

As you can see, expanding the applicability of GemmFwd1x1_0_1 will result in narrowing the applicability of GemmFwdRest. The latter leads to errors that require expensive re-tuning.

I'll provide the proper fix today.

atamazov:
@JehandadKhan @DrizztDoUrden @ritwickJ @sspencespence @CAHEK7 @junliume PR fixed (see https://github.com/ROCm/MIOpen/pull/2811#issuecomment-2041491592 for more info), description and relevant tickets updated. You are welcome to review.

Someone please start CI, thanks!

atamazov:
@sspencespence 
> I have been trying to build and test but get an error during build like...
> ...
> ...seems related. if, as you suggested, updating db isn't necessary, then do I need to build it and is there a way to skip it?

No, this is not about updating databases, this is about building MIOpen.

I suspect this is because you're using outdated ROCm or have outdated dependencies installed (likely sqlite3). I recommend switching to ROCm 6.0.0 and installing fresh dependencies as shown at https://github.com/ROCm/MIOpen?tab=readme-ov-file#installing-dependencies (see https://github.com/ROCm/MIOpen/blob/develop/requirements.txt)

Or you can try to stay with ROCm 5.4.2. In this case I recommend downloading relevant MIOpen sources (https://github.com/ROCm/MIOpen/releases/tag/rocm-5.4.1) and applying patch from https://github.com/ROCm/MIOpen/issues/2789#issuecomment-2041536687.

sspencespence:
Thanks for the clarification! I'll ask again if I come across other issues.On 7 Apr 2024 19:46, Artem Tamazov ***@***.***> wrote:
@sspencespence

I have been trying to build and test but get an error during build like...
...
...seems related. if, as you suggested, updating db isn't necessary, then do I need to build it and is there a way to skip it?

No, this is not about updating databases, this is about building MIOpen.
I suspect this is because you're using outdated ROCm or have outdated dependencies installed (likely sqlite3). I recommend switching to ROCm 6.0.0 and installing fresh dependencies as shown at https://github.com/ROCm/MIOpen?tab=readme-ov-file#installing-dependencies (see https://github.com/ROCm/MIOpen/blob/develop/requirements.txt)
Or you can try to stay with ROCm 5.4.2. In this case I recommend downloading relevant MIOpen sources (https://github.com/ROCm/MIOpen/releases/tag/rocm-5.4.1) and applying patch from #2789 (comment).

CAHEK7 (approved):

junliume:
CI restarted

junliume (approved):

PR #2817, opened 2024-03-18
pass alpha and beta to conv 3D solver

bghimireamd:
This PR attempts to pass alpha and beta to CK solvers


atamazov:
@bghimireamd Would you like to pass alpha
- (a) to the Invoker, and pass alpha as a kernel argument at launch time, or
- (b) to the Solver (via problem description) and compile in (embed) alpha into the kernel at compile time?

JehandadKhan:
To the kernel at launch time!

src/include/miopen/conv/problem_description.hpp
```diff
@@ -248,6 +252,21 @@ struct ProblemDescription : ProblemDescriptionBase
     bool IsDirectionBackwardWrW() const { return direction == conv::Direction::BackwardWeights; }
     std::string GetDirectionStr() const;
 
+    const void* GetAlpha() const { return alpha; }
+    const void* GetBeta() const { return beta; }
+
+    bool IsDefaultAlphaBetaType() const
```
amberhassaan: I recall us identifying the following cases for `alpha/beta`: 
1. Default: both are null or alpha==1 and beta == 0
2. Alpha==x and beta == 0. This case will be handled by adding a `Scale` operation to CK instance
3. Alpha==x and beta==y. This case will be handled by adding a `BiLinear` operation to CK instance. 
4. Alpha==0 and beta==y should not occur but we should check for it as well. 

cc: @JehandadKhan .

src/include/miopen/conv/data_invoke_params.hpp
```diff
@@ -38,28 +38,38 @@ struct DataInvokeParams : InvokeParams
     Data_t workSpace;
     std::size_t workSpaceSize;
     bool gfx90aFp16alt;
+    const void* alpha;
```
amberhassaan: We need a wrapper class to store constants like `alpha/beta` because their data type can be different. This class can have a void pointer and a `miopenDataType_t`. Then it can have a method like `getVal()` that dereferences the pointer based on the data type. 

The datatype itself is determined by the tensor types. See: https://docs.nvidia.com/deeplearning/cudnn/latest/developer/misc.html#scaling-parameters

amberhassaan (changes requested):
Two important fixes: 
1. Proper handling of the cases of alpha/beta
2. Wrapper classes for pointers to alpha/beta.

src/include/miopen/conv/data_invoke_params.hpp
bghimireamd: By the way, CK solver only supports alpha and beta as[ float](https://github.com/ROCm/composable_kernel/blob/develop/include/ck/tensor_operation/gpu/element/binary_element_wise_operation.hpp#L349). Based in cudnn the only types alpha and beta can only be is float (int, half, float) or double.

amberhassaan: I see it as a `CK` limitation. We should at least have the infrastructure in place to support the spec correctly instead of assuming `float` everywhere. If the user inputs `double` tensors, then we shouldn't use `CK` kernels (which is how the solvers are set up).

src/include/miopen/scalar_mul.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+#include <miopen/common.hpp>
+#include <miopen/miopen.h>
+
+#include <cmath>
+#include <limits>
+#include <cassert>
+
+template <typename T>
+inline bool isCloseToZero(T value)
```
amberhassaan: I don't think we need these. We can trust the user to pass zero and one exactly.

src/include/miopen/scalar_mul.hpp
```diff
@@ -0,0 +1,125 @@
...
+    return std::abs(value - T(1)) <= std::numeric_limits<T>::epsilon();
+}
+
+namespace miopen {
+template <int default_val = 1>
+struct ScalarMul
+{
+    ScalarMul()
+    {
+        mVal  = static_cast<float>(default_val);
```
amberhassaan: should cast to `double`

src/include/miopen/scalar_mul.hpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+template <typename T>
+inline bool isCloseToOne(float value)
+{
+    return std::abs(value - T(1)) <= std::numeric_limits<T>::epsilon();
+}
+
+namespace miopen {
+template <int default_val = 1>
```
amberhassaan: Don't like this choice. Let's call the class `Scalar` and take default value in the constructor. If we have `Scalar<0>` and `Scalar<1>`, now these are two types and you cannot assign one to the other without explicitly defining some kind of constructor or assignment operator or type conversion operator. So, not having a template parameter makes life easier.

src/include/miopen/scalar_mul.hpp
```diff
@@ -0,0 +1,125 @@
...
+        {
+            return GetAsFloat();
+        }
+        else if constexpr(std::is_same_v<double, T>)
+        {
+            return GetAsDouble();
+        }
+        else
+        {
+            throw std::runtime_error("ERROR: only expected float or double\n");
```
amberhassaan: Prefer `MIOPEN_THROW`.

src/include/miopen/scalar_mul.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+private:
+    double mVal;
+    miopenDataType_t mType;
+};
+
+using Alpha = ScalarMul<>;
+using Beta  = ScalarMul<0>;
+
+enum class EncodeAlphaBeta
```
amberhassaan: let's move this enum to `conv` namespace and call it `AlphaBetaCase`.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -35,6 +35,31 @@
 #include <ck/utility/data_type.hpp>
 #endif // MIOPEN_USE_COMPOSABLEKERNEL
 
+template <typename T>
```
amberhassaan: This class needs to go in `miopen::solver::internal` namepsace.

src/ocl/convolutionocl.cpp
```diff
@@ -512,23 +524,27 @@ void ConvolutionDescriptor::ConvolutionForward(Handle& handle,
...
         ValidateGroupCount(xDesc, wDesc, *this);
 
         const auto algorithm_name = AlgorithmName{ConvolutionAlgoToDirectionalString(
             static_cast<miopenConvAlgorithm_t>(algo), conv::Direction::Forward)};
-
-        const auto problem =
-            conv::ProblemDescription{xDesc, wDesc, yDesc, *this, conv::Direction::Forward};
+        Alpha alpha_val(alpha, AlphaBetaStorageDataType(xDesc));
+        Beta beta_val(beta, AlphaBetaStorageDataType(xDesc));
+        const auto problem = conv::ProblemDescription{
```
amberhassaan, line 535: Problem description should also cache the `AlphaBetaCase` enum so that we can use it to quickly make decisions.

amberhassaan (commented):
Some comments.

src/include/miopen/scalar_mul.hpp
bghimireamd: I needed this to find the enum case in [AlphaBetaCase](https://github.com/ROCm/MIOpen/blob/83cfe579b5f29c29d5a54202d7f10b86f0370afc/src/include/miopen/scalar_mul.hpp#L99).

src/ocl/convolutionocl.cpp
bghimireamd: now initialized in constructor of[ problem_description.hpp](https://github.com/ROCm/MIOpen/blob/83cfe579b5f29c29d5a54202d7f10b86f0370afc/src/include/miopen/conv/problem_description.hpp#L166)

src/include/miopen/scalar.hpp
```diff
@@ -0,0 +1,113 @@
...
+#include <miopen/miopen.h>
+
+#include <cmath>
+#include <limits>
+#include <cassert>
+
+namespace miopen {
+struct Scalar
+{
+    Scalar(double default_val = 1.0)
```
amberhassaan: single value constructors almost always must be marked with `explicit` or you'll get implicit conversions from `double` to `Scalar`.

src/include/miopen/scalar.hpp
```diff
@@ -0,0 +1,113 @@
...
+    double GetAsDouble() const { return mVal; }
+
+    miopenDataType_t GetType() const { return mType; }
+
+private:
+    double mVal;
+    miopenDataType_t mType;
+};
+
+namespace conv {
```
amberhassaan: This chunk can go in `conv/problem_description.hpp` at the top. It does not belong in this file.

src/include/miopen/scalar.hpp
```diff
@@ -0,0 +1,113 @@
...
+#include <cmath>
+#include <limits>
+#include <cassert>
+
+namespace miopen {
+struct Scalar
+{
+    Scalar(double default_val = 1.0)
+    {
+        mVal  = static_cast<double>(default_val);
```
amberhassaan: initialize these members outside of `{}` after `:`.

src/include/miopen/conv/problem_description.hpp
```diff
@@ -155,9 +158,12 @@ struct ProblemDescription : ProblemDescriptionBase
           weights_layout(ComputeWeightsLayout()),
           out_layout(ComputeOutLayout()),
           direction(direction_),
-          bias(bias_)
+          bias(bias_),
+          alpha(alpha_),
+          beta(beta_)
     {
         HeuristicUpdateLayouts();
+        alpha_beta_case = GetAlphaBetaCase(alpha, beta);
```
amberhassaan: initialize outside the constructor after line `163`.

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
```diff
@@ -66,25 +66,28 @@ using DeviceOpGFwdPtrs =
 
 namespace {
 
+template <typename DataType>
```
amberhassaan: Don't think you still need this, do you?

test/gpu_reference_kernel.cpp
```diff
@@ -422,7 +422,7 @@ struct gpu_reference_conv_2d : gpu_reference_kernel_base
                 rand_tensor_integer(in);
                 rand_tensor_integer(wei);
                 /// \ref copy_non_packed_output_before_convolution
-                rand_tensor_integer(out);
+                // rand_tensor_integer(out);
```
amberhassaan: why comment these lines?

bghimireamd: reverted.

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
bghimireamd: removed this.

src/include/miopen/scalar.hpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    explicit Scalar(double default_val = 1.0)
+        : mVal(static_cast<double>(default_val)), mType(miopenDouble)
+    {
+    }
+
+    // Any type of data in ptr are converted to double
+    Scalar(ConstData_t ptr, miopenDataType_t type = miopenDouble, double default_val = 1.0)
+    {
+        double temp = 1.0;
```
bghimireamd: @amberhassaan had to the initialization of mVal this way to avoid "_runtime error: load of misaligned address_". If float's address is passed we this "misaligned address" error.

amberhassaan: No, this is wrong because you're copying 8 bytes from a `float`'s address, so you get 4 bytes of crap. Let's use a `std::variant<double,float>` for `mVal` and then we read a `float*` as `float.

src/include/miopen/scalar.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include <miopen/miopen.h>
+
+#include <cmath>
+#include <limits>
+#include <cassert>
+
+namespace miopen {
+struct Scalar
+{
+    explicit Scalar(double default_val = 1.0)
```
amberhassaan: I'd like to remove this constructor altogether.

src/include/miopen/scalar.hpp
```diff
@@ -0,0 +1,70 @@
...
+            memcpy(&temp, ptr, sizeof(double));
+        }
+        else
+        {
+            temp = default_val;
+        }
+        mVal = temp;
+    }
+
+    float GetAsFloat() const { return (float)mVal; }
```
amberhassaan: Use `static_cast`. Never use `C-style` casts.

amberhassaan (commented):
replying to initialization issue.

src/include/miopen/scalar.hpp
bghimireamd: If I remove this constructor I will have compiler error because of https://github.com/ROCm/MIOpen/blob/84907b23bc3974173ef410ea0c62abc85e8efb1a/src/include/miopen/conv/problem_description.hpp#L158

src/conv/problem_description.cpp
```diff
@@ -44,6 +44,16 @@ EncodeDataTypesForKey(miopenDataType_t in, miopenDataType_t weights, miopenDataT
     return GetDataTypeName(in) + GetDataTypeName(weights) + GetDataTypeName(out);
 }
 
+bool isCloseToZero(double value)
```
amberhassaan: The reason I'm for removing these functions is because `0.0` and `1.0` can be represented perfectly in floating point notation, so we don't need to guess if the user passed `0.999999999999999999999`. We trust that the user will pass `1.0` if that's what he or she intends.

src/include/miopen/scalar.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+    Scalar(ConstData_t ptr, miopenDataType_t type);
+
+    float GetAsFloat() const;
+    double GetAsDouble() const;
+
+    miopenDataType_t GetType() const { return mType; }
+
+private:
+    std::variant<double, float> mVal;
```
amberhassaan: I would remove `variant` and simplify `GetAsFloat` and `GetAsDouble` by storing value as `double` and `double` can be easily cast down to `float`.

src/scalar.cpp
```diff
@@ -0,0 +1,68 @@
...
+    }
+    else
+    {
+        MIOPEN_THROW("ERROR: Only accepts float or double type for now.");
+    }
+}
+
+float Scalar::GetAsFloat() const
+{
+    if(mType == miopenFloat)
```
amberhassaan: put assert here that `mType == miopenFloat`. or throw an exception.

src/solver/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
```diff
@@ -131,13 +132,16 @@ struct CKArgs
...
+    CKArgs(const CKArgs&)     = default;
+    CKArgs(CKArgs&&) noexcept = default;
     CKArgs& operator=(const CKArgs&) = default;
 
     template <typename ConvPtr>
     auto MakeArgPtr(const ConvPtr& conv_ptr, ConstData_t in, ConstData_t w, Data_t out) const
     {
+        // using T = typename ::internel::AlphaBetaTypeMapper<DataType>::type;
+        // T alpha_val =  alpha.GetVal<T>();
+        // T beta_val  =  beta.GetVal<T>();
```
amberhassaan, line 144: TODO: create three different DeviceOp types and use `AlphaBetaCase` or `std::is_same_v` to decide which `MakeArgumentPointer` to invoke.

src/include/miopen/conv/problem_description.hpp
```diff
@@ -248,6 +270,11 @@ struct ProblemDescription : ProblemDescriptionBase
     bool IsDirectionBackwardWrW() const { return direction == conv::Direction::BackwardWeights; }
     std::string GetDirectionStr() const;
 
+    Scalar GetAlpha() const { return alpha; }
```
amberhassaan, line 273: minor: return `const Scalar&`.

src/include/miopen/conv/problem_description.hpp
```diff
@@ -248,6 +270,11 @@ struct ProblemDescription : ProblemDescriptionBase
...
+    Scalar GetBeta() const { return beta; }
+
+    AlphaBetaCase GetAlphaBetaEnumCase() { return alpha_beta_case; }
```
amberhassaan: remove `Enum` from the name.

src/include/miopen/conv/data_invoke_params.hpp
```diff
@@ -38,28 +39,38 @@ struct DataInvokeParams : InvokeParams
     Data_t workSpace;
     std::size_t workSpaceSize;
     bool gfx90aFp16alt;
+    Scalar mAlpha;
```
amberhassaan, line 42: minor: perhaps stick with the naming style of existing members until we decide on a style guide and upgrade the entire class at that point.

src/include/miopen/solver/problem_description_interpreter.hpp
```diff
@@ -226,6 +226,16 @@ struct ProblemInterpreter
         return problem.GetPadW();
     }
 
+    static Scalar GetAlpha(const miopen::conv::ProblemDescription& problem)
```
amberhassaan, line 229: return `const Scalar&` here too.

src/scalar.cpp
```diff
@@ -0,0 +1,55 @@
...
+    {
+        mVal = *static_cast<const double*>(ptr);
+    }
+    else
+    {
+        MIOPEN_THROW("ERROR: Only accepts float or double type for now.");
+    }
+}
+
+float Scalar::GetAsFloat() const
```
amberhassaan, line 47: These can be in the header file now as they are simple and putting them in the header inlines them easily.

junliume:
**Notice to the reviewers:** replaced with consolidated https://github.com/ROCm/MIOpen/pull/2863

bghimireamd:
should review https://github.com/ROCm/MIOpen/pull/2863 instead.

PR #2819, opened 2024-03-18
Forward MHA find2.0 interface and implementation

CAHEK7:
Dropout is missing right now, because it requires random state initialization. It will be added with the backward pass.

Interface and numeric tests will be added in separated PRs.


CAHEK7:
I did some preliminary local testing (based on https://github.com/ROCm/MIOpen/pull/2563).
Further test integration will be done in separated PR.

CAHEK7:
@atamazov @DrizztDoUrden @amberhassaan can you take a look?

include/miopen/miopen.h
```diff
@@ -346,6 +346,11 @@ MIOPEN_DECLARE_OBJECT(miopenDropoutDescriptor);
  */
 MIOPEN_DECLARE_OBJECT(miopenReduceTensorDescriptor);
 
+/*! @ingroup mha
+ * @brief Creates the miopenMHADescriptor_t type
+ */
+MIOPEN_DECLARE_OBJECT(miopenMHADescriptor);
```
atamazov: [Recommendation] This is still BETA API, isn't it?

include/miopen/miopen.h
```diff
@@ -5310,18 +5315,37 @@ typedef enum
     miopenTensorConvolutionX      = 1,
     miopenTensorConvolutionW      = 2,
     miopenTensorConvolutionY      = 3,
+
+    miopenTensorMHAK                  = 4,
```
atamazov: Ditto

include/miopen/miopen.h
atamazov: [Recommendation] Here and everywhere: Despite the fact that the old APIs are done differently, I prefer keeping camelCase even for acronyms:
```suggestion
MIOPEN_DECLARE_OBJECT(miopenMhaDescriptor);
```
What do you think?

include/miopen/miopen.h
```diff
@@ -5346,6 +5370,40 @@ MIOPEN_EXPORT miopenStatus_t miopenCreateConvProblem(miopenProblem_t* problem,
...
+
+MIOPEN_EXPORT miopenStatus_t miopenSetMHADescriptor(miopenMHADescriptor_t mhaDesc, float scale);
+
+/*! @brief Gets the MHA descriptor details
+ *
+ * Retrieves all of the descriptor details for the MHA.
+ *
+ * @param mhaDesc               Pointer to a MHA descriptor
+ * @param scale                 Scale (output)
+ * @return             miopenStatus_t
```
atamazov: Formatting.

src/CMakeLists.txt
```diff
@@ -134,6 +134,8 @@ set( MIOpen_Source
     load_file.cpp
     lock_file.cpp
     logger.cpp
+    mha/problem_description.cpp
+    mha/mha_descriptor.cpp
     lrn_api.cpp
```
atamazov: Sorting
```suggestion
    lrn_api.cpp
    mha/mha_descriptor.cpp
    mha/problem_description.cpp
```

src/CMakeLists.txt
```diff
@@ -409,6 +412,7 @@ if( MIOPEN_BACKEND MATCHES "OpenCL" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN
         kernels/miopen_limits.hpp
         kernels/miopen_type_traits.hpp
         kernels/miopen_utility.hpp
+        kernels/miopen_rocrand.hpp
```
atamazov: ditto
```suggestion
        kernels/miopen_rocrand.hpp
        kernels/miopen_type_traits.hpp
        kernels/miopen_utility.hpp
```

src/include/miopen/mha/invoke_params.hpp
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    const MHAInputDescsForward& GetDescs() const { return mhaInputDescsForward; }
+    const MHADataForward& GetData() const { return mhaDataForward; }
+
+    std::size_t GetWorkspaceSize() const { return workSpaceSize; }
+    Data_t GetWorkspace() const { return workSpace; }
+
+private:
+    MHAInputDescsForward mhaInputDescsForward;
```
atamazov: Can you please remind me why do we need a copy of `MHAInputDescsForward`? It's size is not that small.

/cc @DrizztDoUrden

atamazov: [Recommendation] I suspect that all data members can be made const.

src/mha/mha_descriptor.cpp
```diff
@@ -0,0 +1,49 @@
+#include <miopen/mha/mha_descriptor.hpp>
```
atamazov, line 26: Please add license header.

src/include/miopen/mha/mha_descriptor.hpp
```diff
@@ -0,0 +1,62 @@
...
+ *******************************************************************************/
+#ifndef MIOPEN_MHADESCRIPTOR_HPP_
+#define MIOPEN_MHADESCRIPTOR_HPP_
+
+#include <miopen/common.hpp>
+#include <miopen/miopen.h>
+#include <miopen/object.hpp>
+#include <nlohmann/json_fwd.hpp>
+
+#include "mha.hpp"
```
atamazov: Fixed ordering, path to mha.hpp, see #2638 
```suggestion
#include <miopen/common.hpp>
#include <miopen/mha/mha.hpp>
#include <miopen/miopen.h>
#include <miopen/object.hpp>

#include <nlohmann/json_fwd.hpp>
```

atamazov (changes requested):
shallowish partial review

include/miopen/miopen.h
Vsevolod1983: Actually, its not. But we can discuss it if there are different opinions

src/include/miopen/mha/invoke_params.hpp
Vsevolod1983: It's better to not rely on an user of this class. I think we cannot always be sure that such a  reference is valid.

CAHEK7: I agree, but...
It can be vulnerable to `std::laundry`, especially in C++20 and if there is any kind of type-erasure or reinterpretation. It's a gray zone, I did not dig deep enough.

Vsevolod1983: I made it const actually, like in other invoke params

include/miopen/miopen.h
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

include/miopen/miopen.h
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

include/miopen/miopen.h
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

include/miopen/miopen.h
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

src/CMakeLists.txt
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

src/CMakeLists.txt
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

src/include/miopen/mha/invoke_params.hpp
atamazov: @Vsevolod1983 @DrizztDoUrden I do not see why. Let's discuss this topic offline.

src/mha/mha_descriptor.cpp
Vsevolod1983: Added

src/include/miopen/mha/invoke_params.hpp
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

src/mha/mha_descriptor.cpp
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

src/include/miopen/mha/mha_descriptor.hpp
atamazov: [resolved] @Vsevolod1983 pls. hide this comment

src/include/miopen/mha/invoke_params.hpp
Vsevolod1983: @atamazov 
Actually, pretty same reason why we don't use const references to Tensor Descriptors but use copy of Tensor Descriptors instead.
Was discussed here:
https://github.com/ROCm/MIOpen/pull/2671#discussion_r1466920140

Vsevolod1983: Could we merge this PR to unblock other PRs creation ?

atamazov: @Vsevolod1983 This comment cannot be considered resolved, but this is not a blocker to merging this PR.

atamazov: [future][performance] To be discussed offline.

src/kernels/miopen_rocrand.hpp
```diff
@@ -0,0 +1,63 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#pragma once
+
+namespace miopen::prng {
+// borrowed from rocrand, since hiprtc/comgr has some problems to compile kernels with rocrand
```
atamazov, line 29: [Q] Do we plan to re-use rocrand?

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -24,33 +24,33 @@
...
-    DST tmp;
-    /// TODO: wait for C++20 and use std::bit_cast
-    /// until then it's the true way of type puning
-    __builtin_memcpy(&tmp, &val, sizeof(DST));
-    /// unions can work for C99 and later but not for C++,
-    /// though compilers widely and unoficially support it
-    return tmp;
-}
-
 constexpr float max_op(float a, float b) { return a > b ? a : b; };
```
atamazov, line 40: [informative] What about `std::max()`?

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -24,33 +24,33 @@
...
-constexpr float max_abs_op(float a, float b) { return max_op(abs(a), abs(b)); };
 constexpr float plus_op(float a, float b) { return a + b; };
-constexpr float identety_op(float a) { return a; };
-} // namespace
+
+/// atomically calcutates maximum of non-negative ordered values
+/// it produces wrong results for negatve values or nans
+/// but it is a final amax reducton step and we expect only non-negative ordered values
```
atamazov, line 45: [R] Typo etc
```suggestion
/// Atomically calculates maximum of non-negative ordered values.
/// Produces wrong results for negatve values or nans,
/// but it is a final amax reducton step and we expect only non-negative ordered values.
```

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -275,51 +264,70 @@ extern "C" __global__ void SoftMaxBlock(const float* in,
...
+        local_val = (lid < seq_len) ? expf(local_val - r_max) : 0;
+
+        float r_sum = 1.0f / reductionBlock<NumWarps>(local_val, plus_op, lid, laneId, warpId);
 
-        float r_sum = 1.f / reductionBlock<NumWarps>(local_val, plus_op, lid, laneId, warpId);
+        local_val *= r_sum;
+
+        // it is supposed to be maximum of absolut values
+        // but after the exponent it is already non-negative
+        // so just a maximum can be used
```
atamazov, line 275: [R]
```suggestion
        // It is supposed to be maximum of absolute values,
        // however we do not need abs() because expf() above produces
        // non-negative value. Plain max() is enough.
```

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -338,21 +346,28 @@ extern "C" __global__ void SoftMaxCommon(const float* in,
...
         for(uint32_t loop_lid = lid; loop_lid < seq_len; loop_lid += blockDim.x)
         {
-            float local_val = exp(line[loop_lid] - r_max) * r_sum;
-            res[loop_lid]   = local_val;
-            r_Amax          = max_abs_op(r_Amax, local_val);
+            float local_val = expf(line[loop_lid] * descaler - r_max) * r_sum;
+
+            // it is supposed to be maximum of absolut values
+            // but after the exponent it is already non-negative
+            // so just a maximum can be used
```
atamazov, line 365: [R] Ditto

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -24,33 +24,33 @@
...
 constexpr float max_op(float a, float b) { return a > b ? a : b; };
-constexpr float max_abs_op(float a, float b) { return max_op(abs(a), abs(b)); };
 constexpr float plus_op(float a, float b) { return a + b; };
-constexpr float identety_op(float a) { return a; };
-} // namespace
+
+/// atomically calcutates maximum of non-negative ordered values
+/// it produces wrong results for negatve values or nans
+/// but it is a final amax reducton step and we expect only non-negative ordered values
+__device__ float atomicMaxOfAbsolutValues(float* addr, float value)
```
atamazov, line 46: [clarity]
```suggestion
__device__ float atomicMaxOfNonNegative(float* addr, float value)
```

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -24,33 +24,33 @@
...
+{
+    // ordered non-negatve and even infinity values can be compared as integers
```
atamazov, line 48: [R] For IEEE-754-compatible formats, when two floats are compared as UNSIGNED, the following is valid:
```
+0.0 < +denorm < +norm < +INF < +SNAN < +QNAN < -0.0 < -denorm < -norm < -INF < -SNAN < -QNAN
```
So if you use unsigned comparison, then both positive and negative NaNs and INFs are considered greater than positive normal and denormal numbers, which is good for MAX operation. But -0.0 is > than +norm/+denorm, which is wrong.

When two floats are compared as SIGNED:
```
-0.0 < -denorm < -norm < -INF < -SNAN < -QNAN < +0.0 < +denorm < +norm < +INF < +SNAN < +QNAN < 
```
This produces "valid" results with -0.0, but results with -NANs/-INFs will be incorrect (which means that -NANs/-INFs may disappear after atomic max, and that may camouflage correctness issues, if any).

Clearing of MSB guarantees valid result with both -0.0 and -NAN/-INF, but this is impossible to do during atomic max. Therefore I recommend careful analyzing of possible inputs:
- If we are pretty sure that -0.0/-INF/-NAN can't appear on input, then we are fine with both signed and unsigned comparison.
- If there is possibility of -0.0, then I would recommend using SIGNED comparison.
- If only -INF/-NAN may appear, then it is better to use UNSIGNED
- If everything is possible (-0.0/-INF/-NAN) then let's use SIGNED. This guarantees valid results with -0.0, but who really cares about disappearing -INF/-NAN? Seriously, if we have -INF/-NAN then it is likely that we also have +INF/+NAN, which do not disappear.

It's up to you what to do with the above considerations. I guess that at least you may wish to add some comments to the code.

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -362,12 +377,61 @@ extern "C" __global__ void SoftMaxCommon(const float* in,
...
+    const uint32_t lid          = threadIdx.x;
+    const uint32_t laneId       = lid % warpSize;
+    const uint32_t warpId       = lid / warpSize;
+
+    r_Amax = reductionBlock<NumWarps>(r_Amax, max_op, lid, laneId, warpId);
+    if(lid == 0)
+    {
+        atomicMaxOfAbsolutValues(Amax, r_Amax);
+    }
+}
```
atamazov, line 437: [R] github shows missing EOL

atamazov (commented):
Device code reviewed. No objections.

src/kernels/MIOpenSoftmaxAttn.cpp
atamazov: Current name give false impression that it performs `max(abs(...))`.

src/kernels/miopen_rocrand.hpp
CAHEK7: We do, at least its kernel code.
RNG node allows the user to run different prng algorithms and we are not going to implement all of them.

Host code does not allow to initialize prng GPU state.

src/kernels/MIOpenSoftmaxAttn.cpp
CAHEK7: I'm aware about this behavior. This function must be called only for non-negative-ordered numbers.

It does not work with NANs, since `fmax(val, NAN) = val` (because all the comparisons `val` vs `NAN` if always `false`), but this function returns a `NAN`, and it does not work with the negative values, which has already been stated in the comment.
C++ std: https://en.cppreference.com/w/cpp/numeric/math/fmax
CUDA: https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE_1g6e7516db46be25c33fb26e203287f2a3

As for the input data - prior to that atomic operation we do `fmax` operation with initial `0` and even if there is any `NAN` in the input data, it will be filtered and replaced by 0 or any other positive value.

atamazov: @CAHEK7 
> This function must be called only for non-negative-ordered numbers.

Yes, but the question is: can we provide a guarantee that this function will be called with such a numbers?

If not, then another question: Do we want NaNs to disappear or to remain? The answer depend on what is required from the softmax primitive.

> As for the input data - prior to that atomic operation we do `fmax` operation with initial `0` and even if there is any `NAN` in the input data, it will be filtered and replaced by 0 or any other positive value.

Is it what we really want?

src/kernels/miopen_rocrand.hpp
atamazov: [resolved]

atamazov (approved):
🟢 Good to merge provided that the following comments will be eventually resolved either in discussion threads right here, or in the future PR(s):
- [ ] https://github.com/ROCm/MIOpen/pull/2819#discussion_r1543138431 - [Q] copying of PD into Invoker
- [x] https://github.com/ROCm/MIOpen/pull/2819#discussion_r1546628019 - [Q] max_op: std:max or...?
- [x] https://github.com/ROCm/MIOpen/pull/2819#discussion_r1546632569 - [R] comments
- [x] https://github.com/ROCm/MIOpen/pull/2819#discussion_r1546643193 - [R] comments
- [x] https://github.com/ROCm/MIOpen/pull/2819#discussion_r1546645308 - [R] comments
- [x] https://github.com/ROCm/MIOpen/pull/2819#discussion_r1546648661 - function name must be fixed
- [x] https://github.com/ROCm/MIOpen/pull/2819#discussion_r1546692620 - [Q] what we want to do with input and intermediate NANs/INFs.
- [ ] https://github.com/ROCm/MIOpen/pull/2819#discussion_r1548102760 - [Bug] Write required workspace size into Solution

src/kernels/MIOpenSoftmaxAttn.cpp
CAHEK7: > Yes, but the question is: can we provide a guarantee that this function will be called with such a numbers?

Yes, because the value prior this atomic operation acquired from `fmax(0, val)` which filters all the `NANs`

> Is it what we really want?

Kind of, the only case where it can produce different result against normal `fmax` reduction is the buffer of all `NANs` - we will produce 0. In the case where we have at least 1 normal number we will match.

The documentation does not say anything explicit about that case.

src/include/miopen/mha/invoke_params.hpp
```diff
@@ -0,0 +1,63 @@
...
+#pragma once
+
+#include <miopen/invoke_params.hpp>
+#include <miopen/tensor.hpp>
+#include "mha.hpp"
+
+namespace miopen {
+namespace mha {
+
+struct InvokeParams : public miopen::InvokeParams
```
amberhassaan, line 36: minor suggestion: rename to `MhaInvokeParams` to be inline with other derived classes of `InvokeParams`.

src/solver/mha/mha_solver.cpp
```diff
@@ -0,0 +1,306 @@
...
+                    handle_.ResetKernelTime();
+                    handle_.AccumKernelTime(mS);
+                }
+            };
+        };
+
+    return result;
+}
+
+bool Mha::MayNeedWorkspace() const { return true; }
```
amberhassaan, line 300: not sure if I saw it but `GetSolution()` should also store a size value in `result.workspace_size`.

amberhassaan (commented):
Looks ok. Please add gtests before merging.

Vsevolod1983:
> Looks ok. Please add gtests before merging.

GTest for MHA Find 2.0 wrapper is in a separate branch. We need to merge this PR in order to create a GTest PR

src/include/miopen/mha/invoke_params.hpp
Vsevolod1983: Well, I 'm looking at other primitives....
softmax - just "InvokeParams" class under softmax namespace.
pooling  - "FwdInvokeParams" class under pooling namespace
reduce - "InvokeParams" class under reduce namespace,
etc

amberhassaan (approved):

src/include/miopen/mha/invoke_params.hpp
amberhassaan: I guess then we're not consistent in our repo. We have `WrwInvokeParams` and `DataInvokeParams` for convolution and then fusion stuff is named similarly I believe.

amberhassaan: I am of the opinion that we should name the classes explicitly instead of relying on what namespace a certain `InvokeParams` is in and sometimes it may also lead to collisions that are not detected. 
CC: @atamazov , @DrizztDoUrden

CAHEK7: @amberhassaan 
What kind of non-detected collisions do you mean?
If we do not overuse `using namespace` - namespace is a good solution.

CAHEK7:
@junliume if you have a change, could we merge it. It blocks few other PRs with the tests and backward pass as well.

junliume (approved):

src/include/miopen/mha/invoke_params.hpp
atamazov: @Vsevolod1983 

> softmax - just "InvokeParams" class under softmax namespace.
> pooling - "FwdInvokeParams" class under pooling namespace

Maybe this is because pooling requires different InvokeParams for Bwd, while softmax doesn't?

@amberhassaan @CAHEK7 
> If we do not overuse `using namespace` - namespace is a good solution.

I totally agree.

src/solver/mha/mha_solver.cpp
atamazov: @CAHEK7 @amberhassaan ...except that if the solution does not require workspace, then it may rely on the default (which is 0).

src/kernels/MIOpenSoftmaxAttn.cpp
atamazov: @CAHEK7 Let me summarize: **_The primitive does not guarantee correct (well-defined) results when input is invalid (i.e. contains INFs or NANs)._** Correct behavior in this case means that if input is invalid then output should be invalid as well. Invalid output means INF or NAN; it is questionable that output 0 should be considered invalid (it is meaningless, but formally valid).

**_The current implementation is such that intermediate INFs and NANs do not appear provided that the input is valid._**

Please unhide this comment.

[resolved]

PR #2821, opened 2024-03-19
Add GraphAPI Operation class

scerzh:
The class is a base for all Graph API operations and is a node in the graph.


scerzh:
@amberhassaan Could you review this small commit; I'm going to base all my branches upon it. I saw you started the same class in your [amber/graph](https://github.com/ROCm/MIOpen/tree/amber/graph) branch.

src/include/miopen/graphapi/graphapi_operation.hpp
```diff
@@ -0,0 +1,131 @@
...
+    bool hasOutput(Tensor* pTensor)
+    {
+        return std::find_if(mOutputs.begin(), mOutputs.end(), [pTensor](const auto& item) {
+                   return std::get<1>(item) == pTensor;
+               }) != mOutputs.end();
+    }
+
+private:
+    std::vector<std::tuple<ArgumentId, Tensor*>> mInputs;
+    std::vector<std::tuple<ArgumentId, Tensor*>> mOutputs;
```
amberhassaan: prefer `std::pair` to `std::tuple` and you should still be able to use `std::get<>`. `pair` has a simple layout while `tuple` doesn't.

src/include/miopen/graphapi/graphapi_operation.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include <algorithm>
+#include <tuple>
+#include <vector>
+
+namespace miopen {
+
+namespace graphapi {
+
+class Operation
```
amberhassaan, line 38: You should rename this class to something `Convolution` related as the implementation is limited to `Convolutions` it seems.

src/include/miopen/graphapi/graphapi.hpp
```diff
@@ -49,6 +51,7 @@ class BackendDescriptor : public miopenBackendDescriptor
                               int64_t* elementCount,
                               void* arrayOfElements) = 0;
     virtual void execute(miopenHandle_t handle, miopenBackendDescriptor_t variantPack);
+    virtual Operation* getOperation();
```
amberhassaan, line 54: Not all `BackendDescriptors` are `Operation Nodes`, so I'm not sure if this method belongs here.

amberhassaan (commented):
Some quick comments.

src/include/miopen/graphapi/graphapi_operation.hpp
scerzh: Not limited. This is a base for other operations as well. For each of the inputs we should add a value of `Operation::ArgumentId` enum or reuse an existing ones, for example `ArgumentId::X` for any single input operation.

src/include/miopen/graphapi/graphapi_operation.hpp
scerzh: agree, `pair` will do

src/include/miopen/graphapi/graphapi.hpp
scerzh: I partially agree with you.

My point is the following. Graph API operates on pointers and doesn't know the type of the descriptor. The base class implementation returns `nullptr`. A descendant class who is an operation descriptor should override it and return a valid pointer.

Another option is to define a separate base class for all Operations and move `getOperation()` there. It will cost an additional dynamic_cast:

```c++
class BackendOperationGraphDescriptor : public BackendDescriptor
{
public:
    void setAttribute(...) override
    {
        BackendDescriptor* pBaseDescr    = deref(static_cast<miopenBackendDescriptor_t*>(arrayOfElements));
        auto& baseDescr = deref(pBaseDescr);
        auto& opDescr = dynamic_cast<ProposedBaseForOperations&>(baseDescr);
        opGraphBuilder.setOperation(opDescr.getOperation());
    }
};
``` 

Seems like not a big price for a good logic. I'll move the function.

amberhassaan: may be we should leave it as is for now and come back to it in the near future.

src/include/miopen/graphapi/graphapi_operation.hpp
amberhassaan: let's leave it to the derived operation classes to associate the tensors to their inputs. At the base operation node, all we care about is the `input tensors` and `output tensors`.

scerzh: Then we don't need `pair`

CAHEK7: outdated

~~X-Y-W sounds like convolution parameters.
Also the minimum number of parameters for operations is 2: 1 input and 1 output tensor.
But for the RNG it can be even 0 input tensors and 1 output (and up to 3 scalar parameters, but it depends, each of them can be an input tensor).
Are you going to enumerate all kind of operand ids here?~~

src/include/miopen/graphapi/graphapi_operation.hpp
```diff
@@ -0,0 +1,131 @@
...
+        {
+            std::get<1>(*existing) = pTensor;
+        }
+        return *this;
+    }
+    Operation& setOutput(ArgumentId argumentId, Tensor* pTensor)
+    {
+        auto existing =
+            std::find_if(mOutputs.begin(), mOutputs.end(), [argumentId](const auto& item) {
+                return std::get<0>(item) == argumentId;
```
CAHEK7: May I recommend `std::get<ArgumentId>(item)`?

scerzh:
No need for a PR. Just a branch will be used.

PR #2822, opened 2024-03-19
Add alignment to the workspace pointer passed to the reduction kernel

qianfengz:
Add alignment guarantee for the workspace pointer passed from the user of the ReduceTensor API before passing the workspace pointer to the kernel. 

This PR only fix the bug for the `dynamic composable kernel` since the `static composable kernel` is regarded as `obsolete`


cderb (approved):
LGTM

src/reducetensor.cpp
```diff
@@ -1031,6 +1031,10 @@ void ReduceTensorDescriptor::ReduceTensor(const Handle& handle,
                                        std::to_string(static_cast<int>(use_padding.first)) +
                                        std::to_string(static_cast<int>(use_padding.second));
 
+        int alignment = sizeof(int);
+        Data_t workspace_aligned = reinterpret_cast<Data_t>(
+            (reinterpret_cast<unsigned long>(workspace) + alignment - 1) / alignment * alignment);
+
```
CAHEK7: Probably `std::align` is better?
Btw, `reinterpret_cast<unsigned long>(workspace)` does truncation on windows, since `sizeof(unsigned long) == 4`
@apwojcik

Also I have doubts that `sizeof(int)` is the best alignment. Probably it works, but 256byte alignment could be better from a performance point of view. May I ask you to double check this?

src/reducetensor.cpp
```diff
@@ -568,7 +568,7 @@ std::size_t ReduceTensorDescriptor::GetWorkspaceSize(const Handle& handle,
     std::size_t wsSizeInBytes =
         !need_indices ? workspace_size * detail::GetDataTypeSize(inDesc.GetType())
                       : workspace_size * (detail::GetDataTypeSize(inDesc.GetType()) + sizeof(int)) +
-                            64 + sizeof(int);
+                            64 + sizeof(int) + sizeof(int);
```
CAHEK7: Am I right and you wanted to cover any misalignment for the buffer with additional `sizeof(int)`?
I guess it's better to use a single global constant for the buffer and for the further alignment routine.

CAHEK7 (dismissed):
It breaks windows functionality.

src/reducetensor.cpp
shurale-nkn: The main reason is to ensure the correctness of the kernel.

I've also recommended using std::align.

apwojcik: For reference, Windows data ranges: https://learn.microsoft.com/en-us/cpp/cpp/data-type-ranges?view=msvc-170 (applies to HIP compiler, too).
Like @CAHEK7 mentioned, `unsigned long` is 32bit on Windows (even on a 64bit operating system) but 64bit on Linux. Please use `unsigned long long`, which size is 64bit on both Windows and Linux.

CAHEK7: Or `uintptr_t` (it is optional, but widely supported). Or `std::align`.

src/reducetensor.cpp
qianfengz: The workspace is divided into 4 buffers inside the kernel.

src/reducetensor.cpp
qianfengz: Updated, please check,  Thanks

src/reducetensor.cpp
CAHEK7: It looks like each of them should be aligned properly.
Actually, I'm going to do some sort of `workspace` container which will hold multiple properly aligned raw buffers from the workspace.
But I'm not sure about timeline and what it will be adopted by all the algorithms.

qianfengz:
One question,   should we use `clang-format-12` or `clang-format-10` for formatting the MIOpen C++ codes?   The Readme here tells to use `clang-format-10`,  but the CI failure indicates using `clang-format-12`

shurale-nkn (approved):
LGTM

junliume:
> One question, should we use `clang-format-12` or `clang-format-10` for formatting the MIOpen C++ codes? The Readme here tells to use `clang-format-10`, but the CI failure indicates using `clang-format-12`

It should be `clang-format-12` now, I will update the readme.

junliume (approved):

src/reducetensor.cpp
```diff
@@ -1031,6 +1032,10 @@ void ReduceTensorDescriptor::ReduceTensor(const Handle& handle,
                                        std::to_string(static_cast<int>(use_padding.first)) +
                                        std::to_string(static_cast<int>(use_padding.second));
 
+        int alignment            = 64;
+        Data_t workspace_aligned = std::align(
+            alignment, workspaceSizeInBytes - alignment, workspace, workspaceSizeInBytes);
```
CAHEK7: It doesn't look like we need to subtract alignment here: `workspaceSizeInBytes - alignment`, `std::align` does it automatically and store the resulted size into the last parameter. But it does not break anything and works fine right now.

CAHEK7 (approved):
Looks fine.

src/reducetensor.cpp
qianfengz: The two parameters `size` and  `space` of `std::align` are confusing and seems no use here.  My codes here refer to the one used in `src/ocl/rnnocl.cpp`

atamazov:
@qianfengz 
> Add alignment guarantee for the workspace pointer passed from the user of the ReduceTensor API before passing the workspace pointer to the kernel.

Can you please provide some explanation how it could be that the pointer is not aligned properly? I'm wondering if we shouldn't align incoming workspace pointers in other primitives. Or there is something special about the reduction primitive?

Well, I've seen this:
> The workspace is divided into 4 buffers inside the kernel.

But if all pointer math happens within the library, then we have MultiBufferWorkspaceTraits in buffer_info.hpp that can handle everything (otherwise it can be extended with necessary stuff).

Or it is the user's application that performs some pointer math and passes modified pointer to the library? Thanks!

src/reducetensor.cpp
atamazov: [Recommendation] `std::align` changes the value of `workspaceSizeInBytes`, so we should not use its value after this (or use it with caution). Well, actually `workspaceSizeInBytes_aligned` gets has different semantics after this call, something alike `workspaceSizeInBytes_aligned`.

This preserves `workspaceSizeInBytes` and introduces explicit "aligned" kind:

```suggestion
        size_t workspaceSizeInBytes_aligned = workspaceSizeInBytes;
        Data_t workspace_aligned = std::align(
            alignment, workspaceSizeInBytes - alignment, workspace, workspaceSizeInBytes_aligned);
```

Or it may be even better to directly update `workspace`, so the semantically bound pair {`workspace`, `workspaceSizeInBytes`} remains in sync after this statement:

```suggestion
        workspace = std::align(alignment, workspaceSizeInBytes - alignment, workspace, workspaceSizeInBytes);
```

atamazov (changes requested):
🟠 I need more info to approve this, please see https://github.com/ROCm/MIOpen/pull/2822#issuecomment-2016591252

atamazov:
@qianfengz @junliume 
> This PR only fix the bug for the `dynamic composable kernel` since the `static composable kernel` is regarded as `obsolete`

Does this mean that after this PR we still have some remaining bug in the `static composable kernel`?

src/reducetensor.cpp
shurale-nkn: > It doesn't look like we need to subtract alignment here: `workspaceSizeInBytes - alignment`, `std::align` does it automatically and store the resulted size into the last parameter. But it does not break anything and works fine right now.

Using the same value in the 2nd and last arg would be an incorrect way to use std:align.
The second argument of the function is the size of the structure after alignment. The fourth argument is the size of the buffer in which the aligned pointer should be located. If the buffer size is insufficient for alignment, the function returns nullptr and does not change the pointer or workspaceSizeInBytes.

workspaceSizeInBytes = real_size + (space for alignment(64));
therefore, it is necessary to send only the actual size of the aligned structure to the second argument.

shurale-nkn: @atamazov  `std::align`, on success, changes the value of the `workspace` pointer to the aligned value and decreases space argument by the number of bytes used for alignment. Thus, the value obtained from std::align is either 0 at fail or identical to new state of `workspace` at success.

shurale-nkn:
@atamazov 
> 
> > Add alignment guarantee for the workspace pointer passed from the user of the ReduceTensor API before passing the workspace pointer to the kernel.
> 
> Can you please provide some explanation how it could be that the pointer is not aligned properly? I'm wondering if we shouldn't align incoming workspace pointers in other primitives. Or there is something special about the reduction primitive?
> 
> Well, I've seen this:
> 
> > The workspace is divided into 4 buffers inside the kernel.
> 
> But if all pointer math happens within the library, then we have MultiBufferWorkspaceTraits in buffer_info.hpp that can handle everything (otherwise it can be extended with necessary stuff).
> 
> Or it is the user's application that performs some pointer math and passes modified pointer to the library? Thanks!

The library does not require the user to provide a buffer aligned in any way. According to the specification, the pointer should only point to an area of sufficient size.
The library can change the specification, but imposing such restrictions leads to a decrease in the library's applicability and compatibility with some applications and standards. Including inside the library itself.

CAHEK7:
> The library can change the specification, but imposing such restrictions leads to a decrease in the library's applicability and compatibility with some applications and standards. Including inside the library itself.

What kind of standards and applications do you mean which allows and endorse non-naturally aligned memory allocations? It would be nice to provide any kind of example. C++ allows it, but you have to force this behavior by special attributes\pragmas, etc - even std::malloc allocates memory with [std::max_align_t](https://en.cppreference.com/w/cpp/types/max_align_t) alignment which is maximum alignment to hold the maximum scalar type (usually 16 bytes).
cudaMalloc (and hipMalloc) are also align memory: "The allocated memory is suitably aligned for any kind of variable." but it behaves even strictly - usually it allocates by 4k chunks.
If the user messed up with the pointers and provided non-naturally aligned buffer - it doesn't sound like an applicability or compatibility limitation, it looks like a bad coding practice.

But in any case, we can just request additional memory for each workspace and do alignment inside the library, it's not a big deal. But it's not a question about applicability or compatibility, we will just always request additional 255 bytes for each workspace, even if it wll be perfectly aligned.

atamazov:
@shurale-nkn 
> The library does not require the user to provide a buffer aligned in any way. 

Unfortunately, it does. In fact, library needs "properly aligned" buffers, but it is not stated explicitly in the documentation. The implied limitations may be even stricter. For example, the library built for the OpenCL backend requires that the workspace is allocated using clCreateBuffer* calls (well, more correctly, it should denote valid R/W cl_mem object).

> According to the specification, the pointer should only point to an area of sufficient size. 

Unfortunately, I don't remember any statements in our documentation that would specify pointer requirements in this form. IIRC we are not telling anything about required alignment. I would say this not the specification but a gap in it.

cuDNN docs say: "_Make sure that all user-provided tensors, workspace, and reserve space are aligned to
128-bit boundaries_"

> The library can change the specification, but imposing such restrictions leads to a decrease in the library's applicability and compatibility with some applications and standards. Including inside the library itself.

What I'm trying to explain is that the library has always had an implicit limitation, which can be informally stated as follows: _the pointers to the device memory passed to the library must be created by appropriate device runtime memory allocation calls. MIOpen/HIP allows also passing addresses that point to a logical sub-buffers somewhere within previously allocated device memory, provided that the addresses are aligned as if the sub-buffers were allocated by the relevant HIP runtime memory allocation calls._

Appending the specification won't impose additional restrictions, it will only make real restrictions more clear for the end user.

UPDATE: Of course, the restrictions also affect the new developments we are currently working on, and this can be inconvenient. In order to provide the necessary flexibility, I would recommend developing new external APIs in a way that allows robust usage of the library even if the library is built with OpenCL backend.

shurale-nkn:
> What kind of standards and applications do you mean which allows and endorse non-naturally aligned memory allocations? It would be nice to provide any kind of example. C++ allows it, but you have to force this behavior by special attributes\pragmas, etc - even std::malloc allocates memory with [std::max_align_t](https://en.cppreference.com/w/cpp/types/max_align_t) alignment which is maximum alignment to hold the maximum scalar type (usually 16 bytes). cudaMalloc (and hipMalloc) are also align memory: "The allocated memory is suitably aligned for any kind of variable." but it behaves even strictly - usually it allocates by 4k chunks. If the user messed up with the pointers and provided non-naturally aligned buffer - it doesn't sound like an applicability or compatibility limitation, it looks like a bad coding practice.
> 
> But in any case, we can just request additional memory for each workspace and do alignment inside the library, it's not a big deal. But it's not a question about applicability or compatibility, we will just always request additional 255 bytes for each workspace, even if it wll be perfectly aligned.

Firstly, the library does not work with full memory allocations, but only with limited memory areas that the user points to. This can be part of bigger buffer of some type, or object placed with an offset from the beginning of the allocated segment. Also, in the compiler, data alignment during allocation occurs to increase efficiency, and not for the correctness of reads.

Secondly, in this case, the pointer alignment was equal to the data type specified in the descriptor. So workspace was PTR_ELEM_ALIGNED. But as already described, this program ignores the data type defined in the descriptor and always uses 32b reads for workspace ptr .

In cudnn, if special alignment of a special buffer is required, this is specified separately in the documentation. In RNN, for example, there is no such restriction and the user can transfer a buffer with any alignment. For ordinary tensors, an alignment equal to the size of the data type is assumed.

shurale-nkn:
@atamazov 
> cuDNN docs say: "_Make sure that all user-provided tensors, workspace, and reserve space are aligned to 128-bit boundaries_"

This is only for Tensor Core Operations support (https://docs.nvidia.com/deeplearning/cudnn/developer/core-concepts.html#tensor-core-operations)

src/reducetensor.cpp
```diff
@@ -568,7 +569,7 @@ std::size_t ReduceTensorDescriptor::GetWorkspaceSize(const Handle& handle,
     std::size_t wsSizeInBytes =
         !need_indices ? workspace_size * detail::GetDataTypeSize(inDesc.GetType())
                       : workspace_size * (detail::GetDataTypeSize(inDesc.GetType()) + sizeof(int)) +
-                            64 + sizeof(int);
+                            64 + sizeof(int) + 64;
```
atamazov: define just above `ReduceTensorDescriptor::GetWorkspaceSize()`:
```C++
// This is WS requirement of the dynamic reduction.
// We must enforce it especially when reduction is used internally.
constexpr std::size_t workspaceAlignRequirementBytes = 4; // or 64 if you wish so but I do not see any reason
```
then
```suggestion
                            64 + sizeof(int) + workspaceAlignRequirementBytes;
```

src/reducetensor.cpp
atamazov: Actually we do not need `workspace_aligned`.
```suggestion
        if(nullptr == std::align(                                  //
            workspaceAlignRequirementBytes,                        //
            workspaceSizeInBytes - workspaceAlignRequirementBytes, //
            workspace,                                             //
            workspaceSizeInBytes))
        {
            MIOPEN_THROW(...);
        }
```
and revert all the changes below.

atamazov (changes requested):
🟡 _This can be merged as is, but I highly recommend resolving comments from this review:_

- Remove implicit dependence
- Remove unnecessary variable, make code cleaner
- Handle potential errors

atamazov:
@qianfengz The answers for the questions asked at https://github.com/ROCm/MIOpen/pull/2822#issuecomment-2016591252 are not required anymore. Discussed offline with @shurale-nkn

qianfengz:
> @qianfengz @junliume
> 
> > This PR only fix the bug for the `dynamic composable kernel` since the `static composable kernel` is regarded as `obsolete`
> 
> Does this mean that after this PR we still have some remaining bug in the `static composable kernel`?

Yes, static composable kernel is ignored by this PR.   The other question I have asked, should we remove the static composable kernel reduction codes ?

atamazov:
@qianfengz @junliume

> > > This PR only fix the bug for the `dynamic composable kernel` since the `static composable kernel` is regarded as `obsolete`
> > 
> > Does this mean that after this PR we still have some remaining bug in the `static composable kernel`?
> 
> Yes, static composable kernel is ignored by this PR.

Thanks for the answer. My current understanding is that this is not a bug at all. Currently many other primitives require DWORD alignment for GPU buffers and this is not a problem.

What this PR does is a kind of adaptation of dynamic reduction primitive for some internal use (in RNN). Of course this can be done by means of `MultiBufferWorkspaceTraits` (see https://github.com/ROCm/MIOpen/pull/2822#issuecomment-2016591252), but for now current approach looks a bit better, at least for @shurale-nkn.

> The other question I have asked, should we remove the static composable kernel reduction codes ?

I don't know. If static implementation is no longer used and if it is worse in every way than the dynamic one, then of course it should be removed. But is it so? To be investigated, I think.

src/reducetensor.cpp
atamazov: [Informative] May be replaced by
```
workspace &= ~(workspaceAlignRequirementBytes - 1); // clear some LSBs of the pointer
workspaceSizeInBytes -= workspaceAlignRequirementBytes; // remove all spare space
```

shurale-nkn: No, NO, NO definitely this is wrong.
`workspace &= ~(workspaceAlignRequirementBytes - 1);`  INCORRECT
The pointer value will be less than the value passed by the user. This way we will get a pointer to the area before or equal to the user ptr.
The aligned value should always be >= user value.

atamazov (approved):
🟢 @shurale-nkn Thank you!

src/reducetensor.cpp
atamazov: [informative] @shurale-nkn Good catch! yeah we should do this first:
```
workspace += workspaceAlignRequirementBytes; // move ptr forwards...
```

PR #2823, opened 2024-03-19
Add filesystem utility functions

apwojcik:
The PR adds the following OS-agnostic filesystem utility function.
* `make_object_file_name()` - appends to a filename `.o` on Linux and `.obj` on Windows,
* `make_executable_name()` - appends to a filename `.exe` on Windows, nothing on Linux,
* `make_static_library_name()` - prepends a filename with a `lib` prefix and appends `.a` on Linux, appends `.lib` on Windows,
* `make_dynamic_library_name()` - prepends a filename with a `lib` prefix and appends `.so` on Linux. spprnfd `.dll` on Windows.

You don't need to separate the parent path from a filename; the functions will work correctly. For example, when you call `make_dynamic_library_name("<parent_path>/sample")`, the result on Linux is `<parent_path>/libsample.so` and on Windows is `<parent_path>\sample.dll`.

In addition, the PR implements two operator overloads for concatenating strings with `fs::path`, resulting in a string object. The call to `.string()` is unnecessary, e.g., when you want to report the path in the `MIOPEN_THROW` messages.

As for logging the PR changes from `+` to `<<`, which natively supports `fs::path`.

PS. You may still see some calls to `.string()` from `fs::path`. The upcoming PR, which changes `std::string` to `fs::path` to work on paths, will remove those (I need this PR to be merged first).


junliume:
@apwojcik MIFin, which is a submodule of MIOpen, needs to be updated to accommodate this change simultaneously. Thanks!

CAHEK7 (approved):

DrizztDoUrden (approved):

junliume:
@cderb could you help to fix FIN when @apwojcik is not available this week?

junliume:
FIN is updated in https://github.com/ROCm/MIOpen/pull/2805, if that PR is fine and merge first, we can update this PR from develop

junliume:
@apwojcik please resolve conflict by merging the latest develop branch.

PR #2826, opened 2024-03-20
Find 2.0 scalar run-time parameters

DrizztDoUrden:
This allows passing run-time parameters to find 2.0 primitives. On find they are not mandatory, but can be attached to miopenFindOptions_t. On run they are passed via `miopenTensorArgument_t` with appropriate id and nullptr descriptor field and are considered mandatory, i.e. would produce an error if missing.


DrizztDoUrden:
@atamazov it may be interesting for you

Vsevolod1983 (approved):

include/miopen/miopen.h
```diff
@@ -5324,6 +5324,7 @@ typedef enum
     miopenTensorSoftmaxDY    = 14,
 #endif
 
+    miopenScalarArgument = 1 << 16,
```
atamazov:
```suggestion
    miopenTensorArgumentIsScalar = 1U << 31,
```
### Why renamed
To use common prefix in the new enum member. Looks terrible and irrational, unless we rename enum
- `miopenTensorArgumentId_t` -> `miopenArgumentId_t`

and then members
- `miopenTensorConvolutionX` -> `miopenArgumentTensorConvolutionX`

etc, but this is what we have right now.

### Why 16 to 31

- Pros: Larger contiguous range of spare ID values
- Cons: None (no cost at runtime)

Why not 32 but 31 and why 1U you already know I guess ;)

atamazov: [Q] How does the user specify the datatype of the scalar argument? Implied by the enum member?

atamazov (changes requested):
Almost good

CAHEK7 (approved):

junliume:
@DrizztDoUrden we have some build issues with this PR:
```
[2024-03-22T16:36:03.792Z] /home/jenkins/workspace/IBS_MIOpen_ddu_find2_scalar-args/src/api/find2_0_commons.cpp:262:12: error: enumeration value 'miopenScalarArgument' not handled in switch [-Werror,-Wswitch]

[2024-03-22T16:36:03.792Z]   262 |     switch(tensor.id)

[2024-03-22T16:36:03.792Z]       |            ^~~~~~~~~
```

include/miopen/miopen.h
DrizztDoUrden: AFAIK we can't rename enum, sadly.
I don't know why I even chose 16, great catch.

DrizztDoUrden: Since this PR doesn't introduce any scalar parameters on its own, user technically doesn't have any way to pass one. But the parameters are meant to be added as `someScalarArg = miopenTensorArgumentIsScalar | someNumber`, so user just calls it by `someScalarArg`. Separate flag is kind of an implementation detail that has no reason to hide it as it makes it obvious, which parameters are scalars. Datatype is inferred from name as of now, but in theory we could support e.g. creating and passing tensor descriptor with size 1 and some metadata (like type) for scalars.

DrizztDoUrden:
I don't understand why windows CI breaks
p.s.: I mean it's the certificate error that doesn't go away

junliume (approved):

PR #2837, opened 2024-03-23
[Windows] fix handling environment variables

apwojcik:
The PR fixes handling environment variables on Windows and adds some improvements to the interface.
Below are the updated functions

* `env::value(VAR)` reads a value from VAR or returns the default value if VAR is not defined
* **`env::value_or(VAR, alt)`** reads a value from VAR or returns the `alt` value if VAR is not defined
* `env::enabled(VAR)` returns `true` if VAR is defined and holds a `true` value
* `env::disabled(VAR)` returns `true` if VAR is defined and holds a `false` value 
* `env::name(VAR)` returns a string with a VAR name
* `env::clear(VAR)` undefines (unsets) a VAR for the local process (does not affect global system settings)
* `env::update(VAR, value)` updates the value of a VAR for the local process. If VAR is not yet defined, it creates it
* `if(VAR)` ... or `if(!VAR)` will tell you if VAR is defined or not defined, respectively

The difference between the previous function's interface and the new one is that the new one does automatic type casting (see examples). That includes reading values (`env::value()` and `env::value_or()`) and updating variables (`env::update()`).

Below is the updated declaration macro
* `MIOPEN_DECLARE_ENV_VAR_*(name, [default_value], [create_if_missing])` compared to the previous definition, the declaration macros has two additional and optional arguments - the default value (if not specified, the default for the data type will be used), and flag which will define a variable if it is missing (not already defined). The default for the flag is not to create the variable.

Examples (see the updated code for more examples)
``` c++
// declare SOME_VARIABLE with default value 0ULL
MIOPEN_DECLARE_ENV_VAR_UINT64(SOME_VARIABLE)

// declare OTHER_VARIABLE with default value "lorem ipsum"
MIOPEN_DECLARE_ENV_VAR_STR(OTHER_VARIABLE, "lorem ipsum")

// declare FLAG_VARIABLE with default value false, and create it if missing
// Notice: the default value needs to be provided if we want to use `create_if_missing` argument
MIOPEN_DECLARE_ENV_VAR_BOOL(FLAG_VARIABLE, false, true)

// example POD for automatic casting
enum some_type {
...
};

// no need for explicit casting from `unsigned long long` to `some_type`
// the compiler will do that automatically
some_type v = env::value(SOME_VARIABLE);

// read the value of OTHER_VARIABLE, and if not defined, ignore the default value
// with declaration macro and use "dolor sit amet instead
auto str = env::value_or(OTHER_VARIABLE, "dolor sit amet");

MIOPEN_DECLARE_ENV_VAR_UIN64(CHOOSE_ALGORITHM, 1)
...
struct Algorithm {
   int value;
   ...
   // override the default value for an algorithm with `value` if CHOOSE_ALGORITHM does not exists
   // notice: no explicit cast is needed between `unsigned long long` and ` algorithmType` data type
   algorithmType getAlgorithm() const { return env::value_or(CHOOSE_ALGORITHM, value); }
  
   // use the default value for an algorithm if CHOOSE_ALGORITHM does not exists
   algorithmType getDefaultAlgorithm() const { return env::value(CHOOSE_ALGORITHM); }
};

enum Flags {
  A = 0,
  B = 1,
...
 K
};

// No need for casting to `unsigned long long`
MIOPEN_DECLARE_ENV_VAR_UINT64(ANOTHER_FLAG, Flags::D)
...
env::update(ANOTHER_FLAG, Flags::K);

// Read default PATH from env variable or use default like this
MIOPEN_DECLARE_ENV_VAR_STR(USER_PATH, fs::temp_directory_path().string())
```


src/ocl/gcn_asm_utils.cpp
```diff
@@ -73,7 +73,7 @@ static std::string CleanupPath(const char* p);
 
 std::string GetGcnAssemblerPathImpl()
 {
-    const auto& asm_path_env_p = miopen::GetStringEnv(ENV(MIOPEN_EXPERIMENTAL_GCN_ASM_PATH));
+    const auto& asm_path_env_p = env::value(MIOPEN_EXPERIMENTAL_GCN_ASM_PATH);
```
junliume: @apwojcik issue with build `HipNoGPU Debug Build Test`:
```
[2024-03-25T05:51:19.005Z] /var/jenkins/workspace/MLLIBS_MIOpen_upgrade_environ/src/ocl/gcn_asm_utils.cpp:76:34: error: use of undeclared identifier 'env'; did you mean 'miopen::env'?

[2024-03-25T05:51:19.005Z]    76 |     const auto& asm_path_env_p = env::value(MIOPEN_EXPERIMENTAL_GCN_ASM_PATH);

[2024-03-25T05:51:19.005Z]       |                                  ^~~

[2024-03-25T05:51:19.005Z]       |                                  miopen::env
```

junliume:
@apwojcik HIP tidy is complaining about `ctuOneDefinitionRuleViolation`, could you help to fix that?

src/include/miopen/env.hpp
```diff
@@ -26,195 +26,193 @@
...
+} // end namespace detail
 
-// static inside function hides the variable and provides
-// thread-safety/locking
-// Used in global namespace
-#define MIOPEN_DECLARE_ENV_VAR(name, type, default_val)                            \
-    namespace miopen::env {                                                        \
-    struct name                                                                    \
+#define MIOPEN_DECLARE_ENV_VAR(_name, _type, ...)                                  \
+    [[maybe_unused]] static const struct __struct_##_name                          \
```
CAHEK7: ~That sort of declaration is bloating binary size, because it has a unique copy of that struct in each file where the same env var is used.
Just a~
```
struct _name
{
};
```
~combines the same struct declaration into one copy in the binary.
That is a critical for the tests where we use `MIOPEN_TEST_ALL` env variable almost in every test and then link everything into one binary.~

~Also it impacts performance since it has a unique copy of the singleton the same env variable and all the copies will be initialized individually. Seems that's why you've introduced `create_if_missing` flag to synchronize all the copies.~

Self-answering. That explicit instantiation is useful for omitting struct instantiation for the function calls, like `GetVal(MIOPEN_TEST_ALL)` instead of `GetVal(ENV(MIOPEN_TEST_ALL))` or `GetVal(MIOPEN_TEST_ALL{})`
And for this particular class without any members there is no problems for the instantiation - it is either small or will be optimized out, at least the will happen for inplace instantiation during a function call.

CAHEK7 (dismissed):
Overall commit looks great, but I have just few questions about overall design:
1, What kind of "windows problem" did you try to solve,
2. ~~Why did you use `static const struct NAME {} name;` declaration of the environment variables.~~

(2) is already answered. The only thing is not quite clear - why do we need to set\update actual OS environment variable and use `create_if_missing` flag. Usually, we care about OS environment variable only during the initialization and any updates should be performed on an internal variable state without going to OS level.

I'm not strictly blocking this PR, I just would like to have some clarifications and suspect that some functionality may not be required.

src/include/miopen/env.hpp
atamazov: +1 each envvar object should be defined only once.

atamazov:
@apwojcik Can we have a smaller PR that contains only functional changes and another NFC PR that contains only renames and namespace changes?

src/include/miopen/env.hpp
CAHEK7: > +1 each envvar object should be defined only once.

We've already discussed this.

junliume:
@apwojcik could you help to follow up on this PR? Thanks!

junliume:
> @apwojcik HIP tidy is complaining about `ctuOneDefinitionRuleViolation`, could you help to fix that?

This problem still exists, please check the latest CI status. Thanks!

.clang-tidy
```diff
@@ -186,6 +189,7 @@ Checks: >-
   -readability-simplify-boolean-expr,
   -readability-suspicious-call-argument,
   -readability-uppercase-literal-suffix,
+  -ctuOneDefinitionRuleViolation
```
CAHEK7: I'm not sure that's a good idea.
Is it 100% false-positive? Otherwise it should be fixed, not suppressed.

pfultz2: Yea these dont look like false positives.

apwojcik: I will revert that change. It was just an experiment, and it did not work.

junliume:
@apwojcik are we still working on this change requested?

apwojcik:
> @apwojcik are we still working on this change requested?

Yes, but currently, I have other priorities. I will be back to this PR.

junliume:
@CAHEK7 @atamazov @pfultz2 last call of review. Thanks!

CAHEK7 (approved):
I don't think that `setEnvironmentVariable` is required at all, it was misused in the tests and must be removed, but now it is even widely used. `env::update(VAR, value)` must be used insted.

But this PR is too huge and fragile, lets merge it and deal with `setEnvironmentVariable` later.

apwojcik:
@junliume, please ignore the Windows build failure; the #3043 will fix that.

driver/CBAInferFusion_driver.hpp
```diff
@@ -601,7 +601,7 @@ int CBAInferFusionDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
     else
         out_sz = in_sz; // This is for N+A so the output is the same as the input size
 
-    if(miopen::IsEnabled(ENV(MIOPEN_DRIVER_PAD_BUFFERS_2M)))
+    if(env::enabled(MIOPEN_DRIVER_PAD_BUFFERS_2M))
```
atamazov, line 604: [Question][Quality] @CAHEK7 @apwojcik Do we really want to change from CamelCase to snake_case for functions?

atamazov, line 604: [R][Quality] Removal of "Is" works with "enabled" and "disabled", but this is not good as a general solution. For example "unset(VAR)" is ambiguous: is it not clear if it queries the state of VAR or changes it.

CAHEK7: @atamazov, the problem is that we already have everything messed up. Let's recall `miopen::deref`, `miopen::rms_range`, ` miopen::compress`, nomen illis legio. I have no ideas what is right.

CAHEK7: A pitfall of irregular verb "set". I agree, that may be confusing.

apwojcik: There are no `env::unset()` nor `env::set()` functions because of the ambiguity. To clear a variable (removes from a process environment variables), one needs to call `env::clear()`, and to set a variable or update its value, one need to call `env::update()`. To check if the variable exists (has been set in the environment), one needs to query the name, e.g., `if (MIOPEN_ALL_TESTS) { ... }` for checking if `MIOPEN_ALL_TESTS` is set (exist in the environment, but not reading a value), and `if(!MIOPEN_ALL_TESTS) { ... }` for checking is it is not set/does not exist. Please see the description of the PR for more details.

apwojcik:
> I don't think that `setEnvironmentVariable` is required at all, it was misused in the tests and must be removed, but now it is even widely used. `env::update(VAR, value)` must be used insted.
> 
> But this PR is too huge and fragile, lets merge it and deal with `setEnvironmentVariable` later.

The `setEnvironmentVariable` function is needed. It is used under the hood because it is a cross-platform implementation of setting/updating environment variables.
My first attempt was with `env::update()`, but that broke one definition rule, and I got clang-tidy errors. To make it work, we need to define all environment variables in a single source file and create an interface to import names into units where we need to work with values (DEFINE and DECLARE macros).

CAHEK7:
> > I don't think that `setEnvironmentVariable` is required at all, it was misused in the tests and must be removed, but now it is even widely used. `env::update(VAR, value)` must be used insted.
> > But this PR is too huge and fragile, lets merge it and deal with `setEnvironmentVariable` later.
> 
> The `setEnvironmentVariable` function is needed. It is used under the hood because it is a cross-platform implementation of setting/updating environment variables. My first attempt was with `env::update()`, but that broke one definition rule, and I got clang-tidy errors. To make it work, we need to define all environment variables in a single source file and create an interface to import names into units where we need to work with values (DEFINE and DECLARE macros).

We were able to handle ODR violation before and it behaved as expected. We don't need OS for updating our internal state, that's important thing.
Initially I've thought about `DEFINE` and `DECLARE` macro, but we've found a solution without it.

junliume (approved):

Issue #2839, opened 2024-03-24
[MHA] Implement point-wise operation descriptor in Graph API


scerzh:
The following descriptors are to be implemented.

MIOPEN_BACKEND_POINTWISE_DESCRIPTOR
- [x] Builder and tests
- [x] C functions and tests

MIOPEN_BACKEND_OPERATION_POINTWISE_DESCRIPTOR
- [x] Builder and tests
- [x] C functions and tests

atamazov:
@scerzh I guess either the topmost comment should be beefed up with more checkboxes or this ticket should be closed, right?

scerzh:
> @scerzh I guess either the topmost comment should be beefed up with more checkboxes or this ticket should be closed, right?

The last checkbox is still not merged #2886. The PR is linked to this issue and will close it automatically after merging.

PR #2854, opened 2024-03-27
[MHA] Implement MIOPEN_BACKEND_POINTWISE_DESCRIPTOR

scerzh:
Issue [#2839](https://github.com/ROCm/MIOpen/issues/2839)
The following tasks were implemented.

MIOPEN_BACKEND_POINTWISE_DESCRIPTOR
- [x] Builder and tests
- [x] C functions and tests

C API -> BackendPointwiseDescriptor -> PointwiseBuilder -> Pointwise

BackendPointwiseDescriptor, a polymorphic descriptor class, is utilized by C-API functions to handle create/set/finalize/get/execute requests.

Pointwise class holds parameters for a pointwise operation (which is created separately).


scerzh:
@atamazov , you are welcome to review.
@DrizztDoUrden , could you review our new enum in miopen.h, please?

urgency_blocker is set as further development for [#2839](https://github.com/ROCm/MIOpen/issues/2839) is blocked by this PR

junliume:
@scerzh please update the feature branch by resolving the conflicts, also the latest CI seems to be having issues.

scerzh:
@junliume @apwojcik  all CI jobs succeeded except Windows, which logs show an SSL/connection error:
```
[2024-04-01T17:58:07.205Z] Downloading http://sourceware.org/pub/bzip2/bzip2-1.0.8.tar.gz
[2024-04-01T17:58:07.205Z] 
[2024-04-01T17:58:07.524Z] Unexpected error: <class 'OSError'>
[2024-04-01T17:58:07.524Z] [Errno socket error] [Errno socket error] [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)
[2024-04-01T17:58:07.524Z] Failed to build package bzip2@1.0.8
```
This type of issue doesn't seem to be caused by the PR.

include/miopen/miopen.h
```diff
@@ -6136,6 +6136,111 @@ typedef enum
...
 } miopenBackendAttributeType_t;
 
+/*! @brief Intended poinwise math operation for a pointwise operation descriptor
+ *
+ * An enumerated type to indicate the intended pointwise math operation in the backend pointwise
+ * operation descriptor
+ */
+typedef enum
+{
+    MIOPEN_POINTWISE_ADD,        /*!< A pointwise addition between two tensors is computed. */
```
DrizztDoUrden, line 6146: [suggestion] It's easier to read large comments when they take full line, so I recommend placing them before the enum element rather than at the end of the line

src/graphapi/pointwise.cpp
```diff
@@ -0,0 +1,251 @@
...
+}
+
+void BackendPointwiseDescriptor::setAttribute(miopenBackendAttributeName_t attributeName,
+                                              miopenBackendAttributeType_t attributeType,
+                                              int64_t elementCount,
+                                              void* arrayOfElements)
+{
+    if(mFinalized)
+    {
+        MIOPEN_THROW(miopenStatusNotInitialized);
```
DrizztDoUrden, line 50: I am not sure `miopenStatusNotInitialized` is a good error code here. `miopenStatusInvalidValue` is probably a proper one here.

DrizztDoUrden (approved):
Minor suggestions, nothing blocking

src/graphapi/pointwise.cpp
scerzh: Agree with you, but I decided to return the same error as described in [CUDNN Graph API docs](https://docs.nvidia.com/deeplearning/cudnn/api/cudnn-graph-library.html#cudnnbackendsetattribute). Do you think we should deviate from it here?

scerzh: I suggest we should not unless we have a reason for that.

include/miopen/miopen.h
scerzh: Agree. If Jun insists on overcoming windows build errors before merging, I will fix comments in this PR, otherwise in the next one.

src/graphapi/graphapi.cpp
```diff
@@ -139,6 +144,18 @@ extern "C" miopenStatus_t miopenBackendInitialize(miopenBackendDescriptor_t desc
 
         switch(descriptorType)
         {
+        case MIOPEN_BACKEND_POINTWISE_DESCRIPTOR:
+            if(std::align(alignof(miopen::graphapi::BackendPointwiseDescriptor),
```
amberhassaan, line 148: you may want to refactor it into a single function that works for all descriptor types. Less repetition that way.

amberhassaan (approved):
Looks good.

src/graphapi/pointwise.cpp
DrizztDoUrden: It is misleading, but I guess it would be OK to keep the same one they have

src/graphapi/graphapi.cpp
junliume: Let's take this into the next refinement PR :)

junliume (approved):

include/miopen/miopen.h
scerzh: Done in https://github.com/ROCm/MIOpen/pull/2792

src/graphapi/graphapi.cpp
scerzh: Done in https://github.com/ROCm/MIOpen/pull/2792

test/gtest/graphapi_gtest_common.hpp
```diff
@@ -98,8 +98,8 @@ class GTestDescriptorVectorAttribute : public GTestDescriptorAttribute
                                    int64_t invalidCount,
                                    std::initializer_list<ValueType> values)
         : mValues(values),
-          mInvalidTypeValues(std::max(1ul, values.size())),
-          mInvalidCountValues(std::max(1l, invalidCount),
+          mInvalidTypeValues(std::max(static_cast<decltype(values.size())>(1), values.size())),
+          mInvalidCountValues(std::max(static_cast<decltype(invalidCount)>(1), invalidCount),
```
atamazov: @scerzh Yes, we should never use UL and L in _integer_ literals. Static casts or ULL/LL are fine.

scerzh: @atamazov what about L or F in floating point literals?

atamazov (commented):
🟢 LGTM after all

test/gtest/graphapi_gtest_common.hpp
atamazov: @scerzh Word "integer" added. Good catch, thanks))

PR #2855, opened 2024-03-27
Graph API: Operation Graph matching

amberhassaan:
Algorithms and tests to establish that two operation graphs are the same (isomorphic).


src/graphapi/opgraph.cpp
```diff
@@ -94,5 +95,220 @@ OpGraph OpGraphBuilder::build() &&
...
+    {
+        auto [it, _ignore] = paths_by_size.emplace(p.size(), VecOfPaths{});
+        it->second.emplace_back(std::move(p));
+    }
+
+    return paths_by_size;
+}
+
+/*
+auto sumPathSizes(const VecOfPaths& all_paths) {
```
Vsevolod1983, line 203: Do we need this commented out code

src/graphapi/opgraph.cpp
```diff
@@ -94,5 +95,220 @@ OpGraph OpGraphBuilder::build() &&
...
+    return l_names == r_names;
+}
+
+bool checkSameDegreeVecs(const OpGraph& left, const OpGraph& right)
+{
+    auto l_degs = left.getInOutDegrees();
+    auto r_degs = right.getInOutDegrees();
+
+    /*
+    auto sort_deg_vec = [] (auto& deg_vec) {
```
Vsevolod1983, line 170: Do we need this commented out code

Vsevolod1983 (approved):

PR #2863, opened 2024-04-01
Integrate conv bilinear ck solver lwp 633

bghimireamd:
Integrate CK's 3D bilinear solvers.

This PR is combination of https://github.com/ROCm/MIOpen/pull/2817 and https://github.com/ROCm/MIOpen/pull/2912


junliume:
@bghimireamd `GroupConv3D_BackwardWeights_half_Bilinear_Suite` is producing NaN, please check

src/conv/problem_description.cpp
```diff
@@ -80,6 +108,21 @@ std::string ProblemDescription::GetDirectionStr() const
...
+std::string ProblemDescription::GetAlphaBetaCaseStr() const
+{
+    std::string s;
+
+    switch(GetAlphaBetaCase())
+    {
+    case AlphaBetaCase::BILINEAR: s = "Bilinear"; break;
+    case AlphaBetaCase::SCALE: s = "Scale"; break;
+    case AlphaBetaCase::DEFAULT: s = "Default"; break;
+    default: assert(false);
```
DrizztDoUrden: This should probably return a string like "Invalid" or "Unsupported". It would be much harder to notice an empty string if this happens in release. It is best to check for errors where the check is expected (like in ctor of the ProblemDescriptor) rather than to rely on it happening as a side-effect of string conversion in debug build.

src/include/miopen/scalar.hpp
```diff
@@ -0,0 +1,56 @@
...
+        assert(mType == miopenFloat);
+        return static_cast<float>(mVal);
+    }
+    double GetAsDouble() const { return mVal; }
+
+    miopenDataType_t GetType() const { return mType; }
+
+private:
+    double mVal;
+    miopenDataType_t mType;
```
DrizztDoUrden, line 53: You can use `std::variant<float, double>`. This way no conversion is required and type is kind of stored internally.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -155,8 +161,27 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
                         const auto& tensors = data_ctx.tensors;
                         SetTensor(handle, tensors.dwDesc, tensors.dw, &zero);
+
+                        void* gpu_buf    = nullptr;
+                        size_t num_bytes = sh_conv_ptr->GetWorkSpaceSize(argument_ptr.get());
+                        if(num_bytes != 0)
+                        {
+                            // allocate gpu workspace memory
+                            HIP_CHECK(hipMalloc(&gpu_buf, num_bytes));
+                            assert(gpu_buf != nullptr);
```
DrizztDoUrden: This should be `if(...) thow ...` since the following code would likely fail hard if the assertion fails in release. Solver selection mechanism relies on explicit failure to provide a correct solution.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -658,7 +683,28 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
...
-            invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), false});
+            if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
+            {
+                void* gpu_buf    = nullptr;
+                size_t num_bytes = sh_conv_ptr->GetWorkSpaceSize(argument_ptr.get());
+                if(num_bytes != 0)
+                {
+                    // allocate gpu workspace memory
+                    HIP_CHECK(hipMalloc(&gpu_buf, num_bytes));
+                    assert(gpu_buf != nullptr);
```
DrizztDoUrden: ditto

DrizztDoUrden (dismissed):

src/conv/problem_description.cpp
bghimireamd: when you say check error in ctor, you mean do MIOPEN_THROW inside [ClassifyAlphaBeta](https://github.com/ROCm/MIOpen/blob/d303cb3d70a0f841de91c1f39f9f74c6da6be364/src/include/miopen/conv/problem_description.hpp#L180) instead of returning [ERROR_STATE](https://github.com/ROCm/MIOpen/blob/d303cb3d70a0f841de91c1f39f9f74c6da6be364/src/conv/problem_description.cpp#L93)? That way we don't have to handle default case?

src/include/miopen/scalar.hpp
bghimireamd: @amberhassaan  I am open to this change.

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: updated

src/conv/problem_description.cpp
```diff
@@ -80,6 +108,21 @@ std::string ProblemDescription::GetDirectionStr() const
     return s;
 }
 
+std::string ProblemDescription::GetAlphaBetaCaseStr() const
+{
+    std::string s;
+
+    switch(GetAlphaBetaCase())
+    {
+    case AlphaBetaCase::BILINEAR: s = "Bilinear"; break;
```
amberhassaan: better to `return "whatever"; `  than assign `s` and `break`.

src/include/miopen/scalar.hpp
amberhassaan: @DrizztDoUrden : we did try using `std::variant` and it did not add much value while complicating the code. Normally I'd argue for `std::variant` too, but in this case, we have a `bigger storage type` double that encompasses `float`, so not using `std::variant` leads to simpler code without questionable conversions.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -155,8 +161,30 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
                         auto zero           = 0.0f;
                         const auto& tensors = data_ctx.tensors;
                         SetTensor(handle, tensors.dwDesc, tensors.dw, &zero);
+
+                        void* gpu_buf    = nullptr;
+                        size_t num_bytes = sh_conv_ptr->GetWorkSpaceSize(argument_ptr.get());
```
amberhassaan: @bghimireamd : what is this code all about? Please add comments to explain what's going on.

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -102,6 +103,32 @@ inline __device__ __host__ int8_t cast_to(const int32_t& val)
     return static_cast<int8_t>(val & 0xff);
 }
 
+inline __device__ __host__ bool IsZero(double val) { return fabs(val) < DBL_EPSILON; }
```
amberhassaan: let's simplify these checks too. We can assume that the user passed exact values for `0.0` and `1.0`.

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -28,6 +28,7 @@
 #include <hip/hip_runtime.h>
 #include <hip/hip_bfloat16.h>
 #endif
+#include <cfloat> // has DBL_EPSILON
```
amberhassaan: Can't include `<cfloat>` I believe because `hipRTC` does not support `std` headers.

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -102,6 +103,32 @@ inline __device__ __host__ int8_t cast_to(const int32_t& val)
...
 
+inline __device__ __host__ bool IsZero(double val) { return fabs(val) < DBL_EPSILON; }
+
+inline __device__ __host__ bool IsOne(double val)
+{
+    return fabs(val - static_cast<double>(1.0) < DBL_EPSILON);
+}
+
+template <typename dst_data_t, typename acc_data_t>
+inline __device__ void alphaBetaUpdate(dst_data_t* __restrict__ p_array,
```
amberhassaan: start name with verb. `applyAlphaBetaUpdate`.

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -102,6 +103,32 @@ inline __device__ __host__ int8_t cast_to(const int32_t& val)
...
+                                       double beta,
+                                       size_t index)
+{
+    if(IsOne(alpha) && IsZero(beta))
+    {
+        p_array[index] = cast_to<acc_data_t, dst_data_t>(value);
+        return;
+    }
+
+    acc_data_t val_alpha_beta =
```
amberhassaan: This is wrong. Imagine if `acc_data_t` is `int32_t` and we're casing `double` to `int32_t`. The correct way is to cast everything to `double` and then to `dst_data_t` in the second line.

src/solver/conv_direct_naive_conv.cpp
```diff
@@ -332,6 +332,784 @@ void conv_internal::DebugPrintTensorStrides(const TensorDescriptor& inDesc,
     printOneStrideVec("outDesc = ", outDesc.GetStrides());
 }
 
+namespace conv_internal {
+::miopen::solver::ConvSolution
+GetConv2DFWDSolution(const ExecutionContext& ctx, const ::miopen::conv::ProblemDescription& problem)
+{
+    ::miopen::solver::ConvSolution result;
+
+    int hi          = problem.GetInHeight();
```
amberhassaan: We need to use `ProblemInterpreter` here because it corrects for the case when descriptors (and buffers) are swapped in the backward passes.

src/solver/conv_direct_naive_conv.cpp
```diff
@@ -332,6 +332,784 @@ void conv_internal::DebugPrintTensorStrides(const TensorDescriptor& inDesc,
     printOneStrideVec("outDesc = ", outDesc.GetStrides());
 }
 
+namespace conv_internal {
```
amberhassaan, line 335: @bghimireamd : good job on consolidating stuff here. I think we need to go further and make these into solver classes. Not sure why we haven't done that. Please create an issue for this.

src/solver/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
```diff
@@ -142,26 +200,100 @@ struct CKArgs
 
     template <typename ConvPtr>
     auto MakeArgPtr(const ConvPtr& conv_ptr, Data_t in, ConstData_t w, ConstData_t out) const
+    {
+        // The function parameter "const ConvPtr&" modifies the "typename ConvPtr"
+        // adding "const" and "reference". So had to adjust them during comparison.
+        if constexpr(std::is_same_v<std::remove_const_t<std::remove_reference_t<ConvPtr>>,
```
amberhassaan: let's simplify this check. We know that `conv_ptr` is a smart pointer, so you can do something like: 
```
using P = std::remove_pointer_t<decltype(conv_ptr.get())>; // might need to add remove_ref_t and remove_const_t
if (std::is_same_v<P, DeviceOp...>)
```

src/solver/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
```diff
@@ -142,26 +200,100 @@ struct CKArgs
...
+        else if constexpr(std::is_same_v<std::remove_const_t<std::remove_reference_t<ConvPtr>>,
+                                         std::unique_ptr<DeviceOpGBwdScale<DataType>>> ||
+                          std::is_same_v<std::remove_const_t<std::remove_reference_t<ConvPtr>>,
+                                         std::shared_ptr<DeviceOpGBwdScale<DataType>>>)
+        {
+            return MakeScaleArgPtr(conv_ptr, in, w, out);
+        }
+        else
+        {
+            return MakeDefaultArgPtr(conv_ptr, in, w, out);
```
amberhassaan, line 217: add static_assert() that conv_ptr is indeed the `PassThrough` type.

amberhassaan (commented):
Some comments.

src/conv/problem_description.cpp
DrizztDoUrden: I mean in the constructor of problem description, but yes, not having an "error" case if it is not functional is probably better.

bghimireamd: now doing [MIOPEN_THROW ](https://github.com/ROCm/MIOpen/blob/b444156dc86ba3111ebb9aab0d84a53f5996e2be/src/conv/problem_description.cpp#L118)in default case.

junliume:
@bghimireamd please help to resolve the conflicts. Thanks!

junliume (approved):

src/include/miopen/miopen_internal.h
```diff
@@ -104,6 +106,23 @@ MIOPEN_EXPORT miopenStatus_t miopenGetConvolutionFindMode(
...
+ * @param  size_t output channel
+ * @param  size_t output tensor length
+ * @param  size_t buffer_size for CK Backward weights work space
+ */
+extern "C" miopenStatus_t
+miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t ab_case,
+                                                   miopenDataType_t data_type,
+                                                   size_t C,
+                                                   size_t K,
+                                                   size_t output_tensor_size,
```
amberhassaan, line 123: This function should just take the tensor descriptors and extract `C`, `K`, `output_size or output num elements` and `data type` from tensor descriptors.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -140,26 +147,31 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
+            auto invoker_ptr     = sh_conv_ptr->MakeInvokerPointer();
+            {
+                HipEventProfiler pfr(handle);
+                if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
                 {
-                    HipEventProfiler pfr(handle);
-                    if constexpr(std::is_same<CastType, miopen::conv::WrWInvokeParams>::value)
+                    auto zero           = 0.0f;
+                    const auto& tensors = data_ctx.tensors;
+                    SetTensor(handle, tensors.dwDesc, tensors.dw, &zero);
```
amberhassaan, line 164: With the `alpha/beta` support, this zeroing out should not be required. @bghimireamd , please confirm with @bartekxk

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -140,26 +147,31 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
+
+                    if(data_ctx.workSpace)
```
amberhassaan, line 166: add a debug mode check to verify if the workspace should have been provided, i.e., catch the scenario where workspace was expected but not set.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -477,10 +490,23 @@ auto MakeTaggedTransposeInstances(ConvSolution& result,
                                        output_solver.GetKernelInfo(),
                                        output_init_solver.GetKernelInfo()});
 
+    if(ck_buff_des)
+    {
+        MultiBufferWorkspaceTraits wt({input1_solver.GetOutputTensorSize(),
+                                       input2_solver.GetOutputTensorSize(),
+                                       output_solver.GetOutputTensorSize(),
+                                       ck_buff_des->ck_size});
+        ck_buff_des->ck_offset = wt.GetOffset(3);
```
amberhassaan, line 499: looks good.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -539,24 +565,50 @@ inline void DebugPrintConvTensors(const ConvTensors& conv_tensors)
...
+// packed size in bytes
+inline size_t GetPackedSize(const TensorDescriptor& td)
+{
+    return td.GetElementSize() * GetTypeSize(td.GetType());
+}
+
+inline size_t GetCKAlphaBetaWorkspace(const miopen::conv::ProblemDescription& problem)
+{
+    std::size_t buff_size;
+    miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(problem.GetAlphaBetaCase(),
```
amberhassaan, line 577: I'm thinking the public API should call this function `GetCKAlphaBetaWorkspace` but I'm not feeling too strongly about it.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -711,8 +763,112 @@ ConvSolution InitInvokerFactoryWrwNCHW(const ExecutionContext& ctx,
     using Input2 = internal::CKTransposeInputOp<ND, internal::ConvOperandTag::Output>;
     using Output = internal::CKTransposeOutputOp<ND, internal::ConvOperandTag::Weights>;
 
-    return InitInvokerFactoryNCHW<DeviceOpType, CKArgsType, CastType>(
-        ctx, problem, kernel_id, Input1{}, Input2{}, Output{});
+    assert(problem.IsLayoutDefault());
+
+    ConvSolution result;
```
amberhassaan, line 768: This giant duplication of code can be avoided by checking if the `output_tr_inst.GetConvOperandTag() == internal::ConvOperandTag::Weights` and adding workspace related logic there. See for example line 681 above.

src/solver/conv_direct_naive_conv.cpp
```diff
@@ -332,6 +332,784 @@ void conv_internal::DebugPrintTensorStrides(const TensorDescriptor& inDesc,
...
+::miopen::solver::ConvSolution
+GetConv2DFWDSolution(const ExecutionContext& ctx, const ::miopen::conv::ProblemDescription& problem)
+{
+    ::miopen::solver::ConvSolution result;
+
+    int hi          = ProblemInterpreter::GetInputHeightHi(problem);
```
amberhassaan, line 341: I am more confident that now naive conv is correct :).

test/gtest/group_conv.hpp
```diff
@@ -497,20 +525,26 @@ std::vector<miopenTensorLayout_t> GetLayoutValues()
...
+           GroupConv##ndim##D_##dir##_##type##_##ab_case##_Test)                        \
+    {                                                                                   \
+        RunSolver();                                                                    \
+    }                                                                                   \
+    INSTANTIATE_TEST_SUITE_P(                                                           \
+        GroupConv##ndim##D_##dir##_##type##_##ab_case##_Suite,                          \
+        GroupConv##ndim##D_##dir##_##type##_##ab_case,                                  \
+        testing::Combine(                                                               \
+            testing::ValuesIn(GroupConvTestConfig<ndim>::GetConfigs<Direction::dir>()), \
+            testing::ValuesIn({alpha}),                                                 \
```
amberhassaan, line 543: You can put a few values of `alpha` and `beta` here that cover all the cases and then you don't have to modify any of the `.cpp` test files or macros below.

amberhassaan (commented):
Some late but important comments. @bghimireamd .

src/include/miopen/solver/implicitgemm_ck_util.hpp
bartekxk: Hi, yes. If Bilinear or SCale is used then zeroing out is not needed.

bghimireamd: @amberhassaan good catch!

src/include/miopen/miopen_internal.h
bghimireamd: Since there is no access to ProblemDescription wouldn't it be hard to decipher C and K (because of different layout it could be more error prone) ?

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: The issue was when `CKBWDWeightBufferDescriptor` is created and passed through lambda function but not used in some cases (fwd and bwd) compiler throws warning, which becomes error in `make analyze`.

amberhassaan: Let's start a draft PR and point me to it. I'd like to try a couple of ideas. I feel it can be done.

src/include/miopen/miopen_internal.h
amberhassaan: So, yes, it will be more complex because of our layouts, but it's also future proof. 
Secondly, you're asking the user to provide `output_tensor_size` which is more error prone to the user. It depends on the layout (packed vs non-packed) and can be confusing (are you asking bytes or elements?).

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: @bartekxk for Scale too? Because scale does not use previous data.

src/include/miopen/miopen_internal.h
bghimireamd: ok. I am asking elements. Will compute bytes based on `data_type`

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: ok will do in https://github.com/ROCm/MIOpen/pull/2972

src/include/miopen/miopen_internal.h
amberhassaan: yes, I know you're asking for num elements, but the main point being, make the interface easier and future proof by taking descriptor pointers instead of `C`, `K` etc. What if you need one of `G, D, H, W` in the future?

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: @amberhassaan  But we do not have miopen::conv::ProblemDescription access in miopen.h

src/include/miopen/miopen_internal.h
bghimireamd: good point! That way at least API remains consistent.

atamazov:
@bghimireamd 
> This PR is combination of #2817 and #2912

Does this mean that all the review comments in #2817 and #2912 are resolved right there or the reviewers have to double their work in the new PR?

/cc @junliume

bghimireamd:
@atamazov yes issues in #2817 and #2912 were resolved.

src/include/miopen/miopen_internal.h
```diff
@@ -104,6 +106,23 @@ MIOPEN_EXPORT miopenStatus_t miopenGetConvolutionFindMode(
...
+ *
+ * @param miopenAlphaBetaCase_t type of alpha beta case
+ * @param  miopenDataType_t data type
+ * @param  size_t input channel
+ * @param  size_t output channel
+ * @param  size_t output tensor length
+ * @param  size_t buffer_size for CK Backward weights work space
+ */
+extern "C" miopenStatus_t
+miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t ab_case,
```
atamazov, line 119: @bghimireamd @amberhassaan @junliume Excuse me for late asking, but what is the purpose of this function? The question seems important to me.

amberhassaan: @atamazov : this came up when supporting `alpha/beta` scaling parameters for convolution recently. All of our current `GetWorkspaceSize` functions do not take `alpha` or `beta` but we ran into a situation where the nature of these parameters affected the workspace size calculation, so we added this new function. 

@bghimireamd : this should probably be named `miopenConvolutionAB...`. CK is an implementation detail that users don't care about.

atamazov: @amberhassaan @bghimireamd Aha, this `miopenConvolutionCKBackwardWeightsGetWorkSpaceSize()` is only for internal use in our library, right? If this is so, then I do not see any reason to declare this functionality in the "internal C API" header. This file contains some APIs that we created but then decided not to open to end users for a while (I don't remember exactly why). In any case, miopen_internal.h contains candidates for miopen.h. Can we implement it as a C++ function, as usual? That would simplify the code, I guess.

amberhassaan: It is also intended as a `beta API` for external users when they are using scaling parameters. May be we should put it in `miopen.h` under the `beta API` section?

bghimireamd: @atamazov @amberhassaan   should I move it to `miopen.h` under `beta API` or just move it as function inside `impilicitgemm_ck_util.hpp`?

atamazov:
⚠️ @bghimireamd Unfortuantely I see some bugs in this 3000+ lines PR. But this PR combines new code and NFC refactorings (like code moves), and this makes reviewing terribly hard to me. PLEASE newer do this in the future _unless_ you provide clean instructions for the reviewers how to filter out refactorings. 

/cc @amberhassaan

src/include/miopen/miopen_internal.h
atamazov: @amberhassaan @bghimireamd Unfortunately, I am unable to find the description of the key design choices in this PR. Therefore I had to review this PR in some more detail to answer the question.

The short answer: No. We need to introduce some other thing into our public API.

### Longer answer

⚠️ The problem is that A/B (alpha/beta) is currently unknown to `miopenFindConvolution*`, so it works as if A/B have default values. This violates the guarantees we must provide to the subsequent `miopenConvolution*`, so it won't work as expected if non-default A/B passed. This problem cannot be resolved with things like miopenConvolutionCKBackwardWeightsGetWorkSpaceSize().

AFAICS the correct way to implement this feature (for Find 1.0 API) includes updating the public API. We have to add possibility to store "a sign of non-default A/B" (something like `ab_case` from this PR) in the ConvolutionDescriptor. This way "a sign of non-default A/B" becomes known to both `miopenFindConvolution*` and `miopenGetWorkspaceSize*`. This opens possibility to produce correct Invokers during Find phase, and also makes additional API calls like "get additional workspace size" unnecessary.

[Notice] FYI current ImmediateMode API does not allow passing alpha/beta.

/cc @junliume @JehandadKhan @CAHEK7

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -116,9 +139,13 @@ inline __device__ __host__ int8_t cast_to(const int32_t& val)
 /// at index 3 while stride for C is at index 0. This is different from how the
 /// tensor descriptors store strides, which is always NCHW order, left-to-right.
 
+/// alpha and beta are double to ensure high precision.
+
 template <bool ASSUME_PACKED, typename src_data_t, typename acc_data_t, typename dst_data_t>
 inline __device__ void naive_conv_fwd_nchw(const src_data_t* __restrict__ p_in,
                                            const src_data_t* __restrict__ p_wei,
+                                           const double alpha,
+                                           const double beta,
```
atamazov, line 148: kdb incompatibility

atamazov: @bghimireamd This is relatively easy to fix: Simply rename the kernel (and update the solver). That would lead to cache misses, but these are relatively harmless. You can even avoid cache misses if you keep the kernel with old signature, but this is a bit more work (and not worth it since the KDB needs to be regenerated anyway -- to cache new kernels).

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -102,6 +102,29 @@ inline __device__ __host__ int8_t cast_to(const int32_t& val)
...
+    if(IsOne(alpha) && IsZero(beta))
+    {
+        p_array[index] = cast_to<acc_data_t, dst_data_t>(value);
+        return;
+    }
+
+    double val_alpha_beta =
+        cast_to<double, acc_data_t>(alpha) * value +
+        cast_to<double, acc_data_t>(p_array[index]) * cast_to<double, acc_data_t>(beta);
+    p_array[index] = cast_to<double, dst_data_t>(val_alpha_beta);
```
atamazov, line 125: It should be
```C++
    acc_data_t val_alpha_beta =
        cast_to<double, acc_data_t>(alpha) * value +
        cast_to<dst_data_t, acc_data_t>(p_array[index]) * cast_to<double, acc_data_t>(beta);
    p_array[index] = cast_to<acc_data_t, dst_data_t>(val_alpha_beta);
```
- [performance] for `val_alpha_beta`, `double` is excessive since all the computations are performed in `acc_data_t` domain
- [correctness] the value read from `p_array[index]` is of the `dst_data_t` type, not `double`.

bghimireamd: @atamazov 
[performance] you recommend float?
[correctness] is already fixed in latest develop branch https://github.com/ROCm/MIOpen/blob/develop/src/kernels/gpu_reference_kernel/naive_conv.cpp#L122C5-L124C66

src/kernels/gpu_reference_kernel/naive_conv.cpp
amberhassaan: @atamazov : how does this change affect KDB? I'm missing some details in my understanding.

amberhassaan:
> ⚠️ @bghimireamd Unfortuantely I see some bugs in this 3000+ lines PR. But this PR combines new code and NFC refactorings (like code moves), and this makes reviewing terribly hard to me. PLEASE newer do this in the future _unless_ you provide clean instructions for the reviewers how to filter out refactorings.
> 
> /cc @amberhassaan

I'm with you on that. This PR ran into some planning issues where we had 3 PRs initially but non of them could be merged stand-alone in a way where functionality would get tested, so it was either merge untested functionality or merge the three PRs into one. We went with the latter (due to time crunch). Hopefully we'll plan things better in the future.

src/kernels/gpu_reference_kernel/naive_conv.cpp
atamazov: > [performance] you recommend float?

`acc_data_t val_alpha_beta = ...`

src/kernels/gpu_reference_kernel/naive_conv.cpp
atamazov: @amberhassaan We supply "precompiled kernels package" which actually contains KDB files. The users can download and install the package on their systems. After this PR, KDB becomes outdated, but MIOpen will try to use it. And if it try to load and launch use outdated `naive_conv_fwd_nchw`, then HIP runtime will fail to do so because of mismatch of the number of parameters.

[Informative] IIRC, QA engineers use precompiled kernels packages during some testing rounds.

The per-user binary cache (UKDB)  is not affected because pathnames of UKDB files contain MIOpen version including git hash.

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -102,6 +102,29 @@ inline __device__ __host__ int8_t cast_to(const int32_t& val)
...
+{
+    if(IsOne(alpha) && IsZero(beta))
+    {
+        p_array[index] = cast_to<acc_data_t, dst_data_t>(value);
+        return;
+    }
+
+    double val_alpha_beta =
+        cast_to<double, acc_data_t>(alpha) * value +
+        cast_to<double, acc_data_t>(p_array[index]) * cast_to<double, acc_data_t>(beta);
```
atamazov, line 124: Thou shalt not read p_array if `IsZero(beta) == true` as it may contain junk (incl. NaNs or Infs) 💡

PR #2869, opened 2024-04-02
workaround SEDEV_451681

carlushuang:
this fix issue https://github.com/ROCm/MIOpen/issues/2867

### DIFF
--- a/src/solver/conv_asm_implicit_gemm_gtc_wrw_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_wrw_nhwc.cpp
@@ -864,6 +864,25 @@ bool ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC::IsApplicable(
         return false;
 #endif
 
+#if WORKAROUND_ISSUE_2867
+    {
+        const int hi        = problem.GetOutHeight();
+        const int wi        = problem.GetOutWidth();
+        const int k         = problem.GetInChannels();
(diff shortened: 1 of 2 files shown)


junliume (approved):

junliume (approved):

JehandadKhan (approved):

src/solver/conv_asm_implicit_gemm_gtc_wrw_nhwc.cpp
```diff
@@ -864,6 +864,25 @@ bool ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC::IsApplicable(
         return false;
 #endif
 
+#if WORKAROUND_ISSUE_2867
+    {
```
atamazov, line 868: [Notice] @bghimireamd Adding this would allow re-using the existing `MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_WRW_GTC_XDLOPS_NHWC` env var to override the W/A.
```C++
#if WORKAROUND_ISSUE_2867
if(!miopen::IsEnabled(ENV(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_ASM_WRW_GTC_XDLOPS_NHWC)))
{
    ... // checks
       return false;
}
#endif
```
This is useful for checking the library with newer versions of HIP runtime, compiler etc without need to rebuild & reinstall MIOpen. I am not sure if this is necessary for this specific case, just to let you know.

PR #2870, opened 2024-04-02
Skip fusions tests when xnack is enabled

cderb:


test/gtest/cba_find2_infer.cpp
```diff
@@ -103,6 +105,8 @@ void RunTunableSolver(miopen::FusedProblem& problem,
     handle.Finish();
 }
 
+bool SkipTest() { return miopen::IsEnabled(ENV(MIOPEN_TEST_GPU_XNACK_ENABLED)); }
```
JehandadKhan: An env var is not required, you can query the handle to know if xnack is enabled or not

JehandadKhan (dismissed):

junliume (approved):

PR #2872, opened 2024-04-03
MHA Forward Find 2.0 Wrapper Test

Vsevolod1983:
This is a GTest for Find 2.0 wrapper over MHA primitive.

This test tests only wrapper itself but not algorithms. It comapres results from Find 2.0 api  with the results from calling invoker directly (without Find 2.0 ).


test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+    bool isFloatTensor;
+    tensor<float> tensorFloat;
+
+    Allocator::ManageDataPtr gpuBuffer;
+
+    // TODO Unsued for now
+    // tensor<unit64_t> tensorUint64;
+};
+
+typedef std::shared_ptr<TensorStruct> TensorStructPtr;
```
CAHEK7: For this particular test I'm wondering why `shared`?
`std::map` and `MhaFind20Test` are the owners of the pointers and we are not going to pass them somewhere else or extent their lifetime.

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+        }
+
+        std::cerr << "Finished testing miopenGetSolution<Attribute>." << std::endl;
+    }
+
+    void TestRunSolutionsForward(Handle& handle, const std::vector<miopenSolution_t>& solutions)
+    {
+        std::cerr << "Testing a forward solution..." << std::endl;
+
+        const unsigned int numTensors = tensors.size();
```
CAHEK7: Probably?
```suggestion
        const auto\size_t numTensors = tensors.size();
```

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+        auto arguments = std::make_unique<miopenTensorArgument_t[]>(numTensors);
+
+        std::vector<miopenTensorDescriptor_t> descVector(numTensors);
+
+        int i = 0;
+        for(const auto& it : tensors)
```
CAHEK7, line 119: It looks a bit clearer.
```suggestion
        for(const auto& [id, tensor] : tensors)
```

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+            std::cerr << "Run via solver infrastructure executed!" << std::endl;
+
+            for(const auto& id : output_ids)
+            {
+                double error = miopen::rms_range(outputTensorResults[id]->tensorFloat.data,
+                                                 tensors[id]->tensorFloat.data);
+                const double tolerance = 1e-3;
+
+                EXPECT_TRUE(std::isfinite(error) && error <= tolerance)
+                    << "Outputs do not match each other. Error:" << error;
```
CAHEK7:
```suggestion
                EXPECT_TRUE(std::isfinite(error));
                EXPECT_LE(error, tolerance);
```
`EXPECT_LE` prints both values in case of condition violation.

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+    void CreateTensor(miopenTensorArgumentId_t id,
+                      unsigned int n = 1,
+                      unsigned int c = 1,
+                      unsigned int h = 1,
+                      unsigned int w = 1,
+                      bool generate  = true)
+    {
+        // TODO Unused for now
+        bool floatFlag = !IsDropoutTensorId(id);
+        tensors[id]    = TensorStructPtr(new TensorStruct(floatFlag));
```
CAHEK7: Probably?
```suggestion
        tensors[id]    = std::make_shared<TensorStruct>(floatFlag);
```
Or `make_unique`.

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+        tensors[id]->tensorFloat = tensor<float>{test_n, test_c, test_h, test_w};
```
CAHEK7: Actually most of the tensors must be `[1,1,1,1]`, but this function ignores input parameters.
```suggestion
        tensors[id]->tensorFloat = tensor<float>{n, c, h, w};
```

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+            CreateTensor(miopenTensorMhaDescaleV);
+            CreateTensor(miopenTensorMhaDescaleS);
+            CreateTensor(miopenTensorMhaScaleS);
+            CreateTensor(miopenTensorMhaScaleO);
+
+            CreateTensor(miopenTensorMhaDropoutProbability);
+            CreateTensor(miopenTensorMhaDropoutSeed);
+            CreateTensor(miopenTensorMhaDropoutOffset);
+
+            CreateTensor(miopenTensorMhaO, test_n, test_c, test_h, test_w);
```
CAHEK7:
```suggestion
            CreateTensor(miopenTensorMhaO, test_n, test_c, test_h, test_w, false);
```
This is also an output tensor, like all the following.

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+            // to compare with
+            for(const auto& id : output_ids)
+            {
+                const TensorStructPtr& tensorStructPtr = tensors[id];
+                tensorStructPtr->tensorFloat.data      = handle.Read<float>(
+                    tensorStructPtr->gpuBuffer, tensorStructPtr->tensorFloat.data.size());
+
+                TensorStructPtr& ptr = outputTensorResults[id];
+
+                ptr              = TensorStructPtr(new TensorStruct());
```
CAHEK7: Probably?
```suggestion
                ptr              = std::make_shared<TensorStruct>(true);
```
Or `make_unique`.

CAHEK7 (changes requested):

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+            CreateTensor(miopenTensorMhaK, test_n, test_c, test_h, test_w);
+            CreateTensor(miopenTensorMhaQ, test_n, test_c, test_h, test_w);
+            CreateTensor(miopenTensorMhaV, test_n, test_c, test_h, test_w);
+
+            CreateTensor(miopenTensorMhaDescaleK);
+            CreateTensor(miopenTensorMhaDescaleQ);
+            CreateTensor(miopenTensorMhaDescaleV);
+            CreateTensor(miopenTensorMhaDescaleS);
+            CreateTensor(miopenTensorMhaScaleS);
+            CreateTensor(miopenTensorMhaScaleO);
```
CAHEK7, line 275: Mathematically it depends on K, Q, V input tensors and softmax\output results from the previous step.
Also `miopenTensorMhaScaleS` is `1 / miopenTensorMhaDescaleS`.
But since we don't care much about numerical meaning, the only restriction is to set it non-zero values.

Non zeroes are important, and I'd suggest to explicitly generate `1`, something like this:
```
tensor<float>{1, 1, 1, 1}.generate([](auto x, auto y, auto z, auto w) { return 1; });
```

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+        std::cerr << "Finished testing backward solution functions." << std::endl;
+    }
+
+    void Finalize() { EXPECT_EQUAL(miopenDestroyProblem(problem), miopenStatusSuccess); }
+
+private:
+    bool IsDropoutTensorId(miopenTensorArgumentId_t id) const
+    {
+        return id == miopenTensorMhaDropoutProbability || id == miopenTensorMhaDropoutSeed ||
```
CAHEK7: `miopenTensorMhaDropoutProbability` is fp32 tensor. Should be excluded from here.

test/gtest/mha_find20.cpp
```diff
@@ -0,0 +1,395 @@
...
+               id == miopenTensorMhaDropoutOffset;
+    }
+
+    void CreateTensor(miopenTensorArgumentId_t id,
+                      unsigned int n = 1,
+                      unsigned int c = 1,
+                      unsigned int h = 1,
+                      unsigned int w = 1,
```
CAHEK7: For the attention it is NHSD - batch size, number of heads, sequence length, embedding dimension size.

test/gtest/mha_find20.cpp
Vsevolod1983: Done

test/gtest/mha_find20.cpp
Vsevolod1983: changed to unique_ptr

test/gtest/mha_find20.cpp
Vsevolod1983: For me such a syntax construction looks really heavy for its really simple purpose its intended for. So, would prefer my option

CAHEK7 (approved):

CAHEK7:
I would like to have it merged prior to #2875.

junliume (approved):

PR #2873, opened 2024-04-03
[MHA] Implement MIOPEN_BACKEND_OPERATION_RNG_DESCRIPTOR

### LINKED ISSUE #2842: [MHA] Implement random number generator descriptor in Graph API

MIOPEN_BACKEND_RNG_DESCRIPTOR
- [x] Builder and tests
- [x] C functions and tests

MIOPEN_BACKEND_OPERATION_RNG_DESCRIPTOR
- [x] Builder and tests
- [x] C functions and tests

scerzh:
Issue #2842 

Tasks:

MIOPEN_BACKEND_OPERATION_RNG_DESCRIPTOR
- [x] Builder and tests

C API -> BackendOperationRngDescriptor (polymorphic) -> OperationRngBuilder -> OperationRng

BackendOperationRngDescriptor, a polymorphic descriptor class, is utilized by C-API functions to handle create/set/finalize/get/execute requests. It hasn't been developed in this PR and is coming in the next one.

Builder pattern is used for set and finalize API requests. Builder is part of this PR.

OperationRng class represents a random number generator node in an operation graph.

After merging #2874 into this PR another task is completed:

MIOPEN_BACKEND_OPERATION_RNG_DESCRIPTOR
- [x] C functions and tests

and this PR closes #2842


test/gtest/graphapi_operation_rng.cpp
```diff
@@ -0,0 +1,335 @@
...
+                                   ValidatedValue<Tensor*>,
+                                   ValidatedValue<std::variant<int64_t, Tensor*>>,
+                                   ValidatedValue<Tensor*>>;
+
+} // namespace
+
+class GraphApiOperationRngBuilder : public testing::TestWithParam<DescriptorTuple>
+{
+protected:
+    bool attrsValid;
```
Vsevolod1983: In class  OperationRng you use prefix "m" for variables. Why in this class you dont use this prefix ?

Vsevolod1983: Should we put public section first, and protected, private after publicr ?
Usually, people use the class and need public interface first of all, so its more convinient to put it first.

But this is a topic for discussion of course. Not sure do we have some rules about it/

Vsevolod1983 (approved):

test/gtest/graphapi_operation_rng.cpp
scerzh: I don't use `m` prefix in GTest Fixtures to emphasize input parameters of a test.
Also I didn't use `m` for `execute` in another Fixture to show that it is a functional object.

Another reason. When you write TEST_P body you treat these members as local (or global) variables. When I don't want TEST_P to use a member, then I put `m` prefix.

scerzh: There will be no public section in this class. It is a Fixture; TEST_P seems to create a descendant class internally and puts its (TEST_P's) body in one of class's methods so protected members will be available, and they are perceived by you as local variables when you write TEST_P body.

Vsevolod1983: I thought we want to use "m" prefix for class members, does not matter what this param is intended for. Just to distinguish them from  local function variables.
But ok.

amberhassaan: I agree with @Vsevolod1983 . We should just use `mVarName` for members consistently, structs or classes, doesn't matter.

amberhassaan: This is something we haven't agreed as a team on, and we are not enforcing, so you can delay making this change if you want.

scerzh: done.
And in the next PR too.
But I'm not sure we should apply this rule to public members.

scerzh:
Changed the title to reflect merging #2874 into this PR

CAHEK7 (approved):

scerzh:
@junliume Can we merge this PR? Other PRs are going to reuse some parts of it.

junliume (approved):

PR #2875, opened 2024-04-04
[MHA] Added forward numeric test. Fix some bugs in cpu and gpu implementations. Resolved some post-merge comments.

CAHEK7:
It is a numeric test for the forward multi-head attention pass.

~~It's based on https://github.com/ROCm/MIOpen/pull/2563, so once it has been merged, the number of changes been reduced.~~
It also fixes leftovers from that comment https://github.com/ROCm/MIOpen/pull/2819#pullrequestreview-1973929801 and a few bugs in GPU and CPU implementations.

@atamazov


CAHEK7:
~~I'm waiting for #2563 to be merged making a review process a bit easier.~~
Done.

Vsevolod1983 (approved):

src/solver/mha/mha_solver.cpp
```diff
@@ -160,7 +160,7 @@ ConvSolution Mha::GetSolution(const ExecutionContext& context,
         return static_cast<void*>(static_cast<std::byte*>(buffer) + ws.GetOffset(part_idx));
     };
 
-    local_threads  = RoundUpToMultiple<uint64_t>(std::min<uint64_t>(nhsd, 256), 32);
+    local_threads  = std::clamp<uint32_t>(nextPow2(nhsd), warpSize, 256);
```
atamazov, line 163: This is not the same, e.g.
- RoundUpToMultiple(min(`90`,256), 32) == `96`
- std::clamp(nextPow2(`90`), 32, 256) == `128`

atamazov (commented):
is it intentional?

src/solver/mha/mha_solver.cpp
CAHEK7: Yes, I need the second one.

CAHEK7:
> is it intentional?

Yes. Under the hood it requires a number of warps to be power of 2.

src/solver/mha/mha_solver.cpp
atamazov: [resolved]

atamazov:
@CAHEK7 I recommend renaming this to seomthing like: `[MHA] Fixed SoftmaxAttn kernel. Fixed MakeNetworkConfig(). Added forward numeric test. Resolved some post-merge comments.`

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -408,15 +409,15 @@ extern "C" __global__ void ScaleReduce(const float* __restrict__ in,
 
     auto in_ptr    = in + gid;
     auto out_ptr   = out + gid;
-    const auto end = in_ptr + nhsd;
+    const auto end = in + nhsd;
```
cderb, line 412: This seems like a significant change? How does in relate to in_ptr? Was this a bug before?

src/mha/problem_description.cpp
```diff
@@ -37,11 +37,20 @@ NetworkConfig ProblemDescription::MakeNetworkConfig() const
...
-    ss << "mhafwd-";
+    ss << "mha";
 
     if(isForward)
     {
-        // TODO Implement
+        ss << "fwd-";
+        for(auto s : mhaInputDescsForward.oDesc.GetLengths())
+        {
+            ss << s << "x";
```
cderb, line 47: Is the extra x on the end of the Lengths / Strides for the Type?
ex. mhafwd-1x1x1xFP32

CAHEK7: Yep. It's just a DB key.

src/kernels/MIOpenSoftmaxAttn.cpp
CAHEK7: It was a bug. 'end' pointer counts from the beginning - from `begin` pointer (which is `in`).
`in_ptr` is a shifted by global thread id, It can index current element, but can't be used for the `end` calculatons.

cderb (approved):

CAHEK7:
@junliume seems it's ready.

DrizztDoUrden (approved):

iq136boy (approved):

test/gtest/mha_helper.hpp
```diff
@@ -62,10 +62,9 @@ float GetF8Scaling(float max_val)
 template <typename T>
 T FindMax4D(const tensor<T>& max_of_tensor)
```
amberhassaan, line 63: rename to `AbsoluteMax` as we're doing the `abs()` and then `4D` is no longer a limitation due to the variadic lambda.

bghimireamd (approved):
LGTM

test/gtest/cpu_multi_head_attention.hpp
```diff
@@ -105,8 +131,8 @@ struct CPUMHATest : public ::testing::TestWithParam<CPUMHATestCase>
 
         const auto [error, max_diff] = calcStats(final_transformed_attention);
         // CI clang-tidy treats is as "boolean value assigned to float"
-        const double error_threshold    = ((std::is_same_v<OutputType, float>) ? 1e-7 : 1e-2);
-        const double max_diff_threshold = ((std::is_same_v<OutputType, float>) ? 1e-6 : 1e-1);
+        const double error_threshold    = ((std::is_same_v<OutputType, float>) ? 1e-7 : 5e-2);
```
amberhassaan, line 134: is the first one relative error? or what is the difference between the two?

amberhassaan (approved):
Did not review kernel code very carefully but the rest looks good.

test/gtest/cpu_multi_head_attention.hpp
bghimireamd: I think one is for float and other is for fp8? Since fp8 has lower precision hence the lower tolerance.

test/gtest/mha_helper.hpp
CAHEK7: I guess it can be done in separated PR with backward CPU implementation (just to avoid any potential conflicts and CI restart).
@bghimireamd

test/gtest/cpu_multi_head_attention.hpp
CAHEK7: If you mean different thresholds for the errors - @bghimireamd has already answered.
If you mean error and max_diff then the first one is rms and the second one is maximum absolute difference (
`sqrt(sum((x1-y1)^2...) / n)` and `max(abs(x1-y1)...)`)
Since rms hides outliers, I'd like to control maximum difference as well.

If you mean why I changed fp8 threshold - I changed a pipeline to precalculate some scaling coefficient where it's possible, but some others can't be precalculated - they go from the previous iteration of the algorithm and since we do just one iteration it is 1.0f and it introduced extra error.

amberhassaan: @CAHEK7 : thanks. That's what I was looking for.

junliume (approved):

PR #2876, opened 2024-04-04
removing gfx908 and vega builds node from smoke tests

reidkwja:
removes gfx908 and vega from getting picked up for smoke tests runs.


atamazov:
@reidkwja this is not an enhancement, this is workaround.

Jenkinsfile
```diff
@@ -815,12 +815,12 @@ pipeline {
                 stage('Fp32 Hip AnyGPU') {
```
atamazov: Here and everywhere: if only gfx90a is left, please use "gfx90a" in the stage name instead of "AnyGPU"

Jenkinsfile
atamazov, line 1: Line 470:
```groovy
        booleanParam(
            name: "TARGET_GFX908",
            defaultValue: env.BRANCH_NAME == env.NIGHTLY_BRANCH ? true : false,
description: "")
```
This totally excludes MI100 nodes from testing except nightly. Who will triage and fix the bugs found during nightly?

If we are excluding MI100s from smoke tests now, then I highly recommend enabling them in Full tests, **_in develop at least._**

atamazov (changes requested):

Jenkinsfile
atamazov: /cc @junliume

cderb: I'll echo this comment.
Limiting the smoke tests to gfx90a is acceptable in my opinion, because the test stage is then more deterministic in the CI.
However, there is still a need to identify architecture specific bugs which may arise. Perhaps the full test is the stage to address this in. gfx908 and vega tests will still be necessary in the regular CI to provide immediate feedback for developers.

junliume (approved):

PR #2879, opened 2024-04-08
[MHA] Implement MIOPEN_BACKEND_OPERATION_POINTWISE_DESCRIPTOR Builder

scerzh:
Issue #2839

Tasks:

MIOPEN_BACKEND_OPERATION_POINTWISE_DESCRIPTOR
- [x] Builder and tests

C API -> BackendOperationPointwiseDescriptor (polymorphic) -> OperationPointwiseBuilder -> OperationPointwise

BackendOperationPointwiseDescriptor, a polymorphic descriptor class, is utilized by C-API functions to handle create/set/finalize/get/execute requests.

OperationPointwise class represents a pointwise node in an operation graph.

### DIFF
--- a/src/graphapi/pointwise.cpp
+++ b/src/graphapi/pointwise.cpp
@@ -120,6 +120,8 @@ void BackendPointwiseDescriptor::setAttribute(miopenBackendAttributeName_t attri
         setFloatOrDouble(&PointwiseBuilder::setSoftPlusBeta);
         break;
 
+    case MIOPEN_ATTR_POINTWISE_SWISH_BETA: setFloatOrDouble(&PointwiseBuilder::setSwishBeta); break;
+
     case MIOPEN_ATTR_POINTWISE_AXIS:
         if(attributeType == MIOPEN_TYPE_INT64 && elementCount == 1)
         {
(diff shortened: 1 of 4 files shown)


src/graphapi/pointwise.cpp
```diff
@@ -246,6 +252,416 @@ void BackendPointwiseDescriptor::getAttribute(miopenBackendAttributeName_t attri
...
+        }
+        break;
+
+    /* TODO: Implement the remaining cases
+     */
+    case MIOPEN_POINTWISE_ERF:
+    case MIOPEN_POINTWISE_GEN_INDEX: MIOPEN_THROW(miopenStatusNotImplemented);
+
+    default: MIOPEN_THROW(miopenStatusNotImplemented);
+    }
```
scerzh, line 660: @CAHEK7 @DrizztDoUrden @amberhassaan Colleages, could you check the switch statement, please? Are the pointwise operations grouped correctly, are their inputs and outputs correct, what group should we put the last ones to; and do we support all of the operations or should we throw miopenStatusNotImplemented in certain cases?

scerzh:
@junliume urgency has been set to unblock the next PR for MIOPEN_BACKEND_OPERATION_POINTWISE_DESCRIPTOR

CAHEK7 (approved):

junliume (approved):

src/graphapi/pointwise.cpp
atamazov: @scerzh 
> should we throw miopenStatusNotImplemented in certain cases?

If some operation is listed in the API but not implemented, then it seems obvious that we should throw `miopenStatusNotImplemented` instead of `miopenStatusBadParm`. So I recommend exchanging BadParm <-> Not Implemented in this switch.

PR #2880, opened 2024-04-08
[MHA] Implement MIOPEN_BACKEND_OPERATION_REDUCTION_DESCRIPTOR

### LINKED ISSUE #2841: [MHA] Implement reduction descriptor in Graph API

MIOPEN_BACKEND_REDUCTION_DESCRIPTOR
- [x] Builder and tests
- [x] C functions and tests

MIOPEN_BACKEND_OPERATION_REDUCTION_DESCRIPTOR
- [x] Builder and tests
- [x] C functions and tests

scerzh:
Issue #2841

Tasks

MIOPEN_BACKEND_OPERATION_REDUCTION_DESCRIPTOR
- [x] Builder and tests
- [x] C functions and tests

C API -> BackendOperationReductionDescriptor (polymorphic) -> ReductionOperationBuilder -> OperationReduction

BackendOperationReductionDescriptor, a polymorphic descriptor class, is utilized by C-API functions to handle create/set/finalize/get/execute requests.

OperationReduction class represents a reduction node in an operation graph.

closes #2841


src/graphapi/reduction.cpp
```diff
@@ -134,6 +136,65 @@ void BackendReductionDescriptor::getAttribute(miopenBackendAttributeName_t attri
...
+{
+    if(src != nullptr)
+    {
+        dst = src;
+    }
+    else
+    {
+        MIOPEN_THROW(miopenStatusBadParm);
+    }
+}
```
CAHEK7: Seems this function is widely used. Can it be moved to a common header?

CAHEK7 (approved):
Some minor suggestion.
It's probably a part of https://github.com/ROCm/MIOpen/issues/2748

src/graphapi/reduction.cpp
scerzh: Now it is used in 2 cpp files. I'll propagate its usage into existing cpps and place it into graphapi.hpp in one of the next PRs.

scerzh: Also instead of this function we can write:
```C++
dst = &deref(src);
```
I find it less readable. What do you think, @CAHEK7 ?

CAHEK7: For me `defer` doesn't look like a perfect solution too. Current one looks good.

Or
```C++
dst = checkPtr(src);
```

scerzh: Your suggestion is more flexible:
```C++
dst = checkPtr(src);
```
Done so in #2886

scerzh: Done here as well

scerzh:
@junliume Can we merge this one?

junliume (approved):

PR #2884, opened 2024-04-10
Remove extra db path quotes introduced in #2823

### LINKED ISSUE #2881: MIOPEN_USER_DB_PATH=. no longer works as it should after #2823
With `MIOPEN_USER_DB_PATH=.` MIOpen creates a new directory (`'"."'`) in the current directory and puts db files into that directory instead of the current directory. Probably caused by https://github.com/ROCm/MIOpen/pull/2823

CC @junliume @atamazov

CAHEK7:
Fixes #2881 

#2823 removed `.string()` from `fs::path` redirected to `std::stingstream` and it introduced extra quotes to that path which caused #2881.
This is the most obvious and noticeable place, but since we mixing `std::string`, `fs::path`, and streams that but may reappear in different places.
Probably some logs can be broken, since #2823 removed `.string()` from lots of logging functions, but logs are not so critical and sometimes that's a good change, since it will quote a full path.

But reverting #2823 may not help - there are some other PRs where that sort of changes can be done. Looks like all the `fs::path` PRs should be revisited.

@atamazov  @junliume

### DIFF
--- a/src/find_db.cpp
+++ b/src/find_db.cpp
@@ -219,7 +219,7 @@ std::string FindDbRecord_t<TDb>::GetUserPath(Handle& handle, const std::string& 
 {
 #if !MIOPEN_DISABLE_USERDB
     std::ostringstream ss;
-    ss << GetUserDbPath() << '/';
+    ss << GetUserDbPath().string() << '/';
     ss << handle.GetDbBasename();
     ss << '.' << GetUserDbSuffix();
     if(!path_suffix.empty())


atamazov (approved):
🟢 LGTM provided that this resolves the issue.

apwojcik:
The issue occurs because the `std::filesystem::path` (and `boost::filesystem::path`, too) do the normalization before writing to a string stream, and in case of the empty path, it resolves to `.` (calling `.string()` just returns `""`). Soon, I will submit a PR that builds the user db path entirely with `fs::path` (no string streams in between) because the current implementation will not work on Windows. However, I needed to merge #2823 first.
Read more about the normalization algorithm here: https://en.cppreference.com/w/cpp/filesystem/path

apwojcik (approved):

CAHEK7:
> The issue occurs because the `std::filesystem::path` (and `boost::filesystem::path`, too) do the normalization before writing to a string stream, and in case of the empty path, it resolves to `.` (calling `.string()` just returns `""`). Soon, I will submit a PR that builds the user db path entirely with `fs::path` (no string streams in between) because the current implementation will not work on Windows. However, I needed to merge #2823 first. Read more about the normalization algorithm here: https://en.cppreference.com/w/cpp/filesystem/path

Cppreference says nothing about explicit path normalization for stream output. It says that it uses `std::quoted` for stream output, that's it.
I didn't try it on windows, but there is no difference except quotes between `path` and `path.string()` on linux: https://godbolt.org/z/dP5Es4n1e
it prints `|""||` - just quotes for `path` and empty string for `path.string()`, without adding an extra `.` for `stream << path`.

cderb (approved):

junliume (approved):

PR #2888, opened 2024-04-11
Consider workspace constraints when loading convolution solutions from database

tbennun:
Trying to find a convolution algorithm with a workspace size constraint works when the database needs to be regenerated. This PR enables using that same constraint when the database already exists.

This is important for memory-limited environments with large neural networks.

### DIFF
--- a/src/ocl/convolutionocl.cpp
+++ b/src/ocl/convolutionocl.cpp
@@ -204,7 +204,8 @@ static void ShrinkToFind10Results(std::vector<PerfField>& found)
 
 static inline std::vector<PerfField> FindConvolution(const ExecutionContext& ctx,
                                                      const conv::ProblemDescription& problem,
-                                                     const AnyInvokeParams& invoke_ctx)
+                                                     const AnyInvokeParams& invoke_ctx,
+                                                     const int requestAlgoCount)
 {
     auto results         = std::vector<PerfField>{};
     auto sol             = boost::optional<miopenConvSolution_t>{};
@@ -214,7 +215,16 @@ static inline std::vector<PerfField> FindConvolution(const ExecutionContext& ctx
     if(findMode.IsFast(ctx) || findMode.IsHybrid(ctx))
     {
         auto fallback = bool{};
-        auto sols     = conv.GetSolutions(ctx, problem, 1, &fallback);
+        auto sols     = conv.GetSolutions(ctx, problem, requestAlgoCount, &fallback);
+
+        // Remove solutions for which the given workspace size is insufficient
+        sols.erase(std::remove_if(sols.begin(),
+                                  sols.end(),
+                                  [&](const miopenConvSolution_t& entry) {
+                                      return invoke_ctx.GetWorkspaceSize() < entry.workspace_size;
+                                  }),
+                   sols.end());
+
         // override the normal find with immed mode with env var
         if(!sols.empty() && (!(findMode.IsHybrid(ctx) && fallback) ||
                              miopen::IsEnabled(ENV(MIOPEN_DEBUG_FORCE_IMMED_MODE_FALLBACK))))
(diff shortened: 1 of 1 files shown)


atamazov:
@tbennun Please explain why you need this feature. Please include a use case scenario (at high level; I do not need a complete C++ program or alike). Thanks.

atamazov (changes requested):
This PR includes a quiet change of the public API. Such a change must be made explicit at least (explanation: https://github.com/ROCm/MIOpen/pull/2888#issuecomment-2053755444).

tbennun:
@atamazov Thank you for reviewing.

> @tbennun Please explain why you need this feature. Please include a use case scenario (at high level; I do not need a complete C++ program or alike). Thanks.

A deep learning library trains a neural network. Since the network has high demands on GPU memory (e.g., activation tensor size), the workspace allocation is limited to X GiB. If the database does not exist, calling `miopenFindConvolutionForwardAlgorithm` with a given workspace size will (correctly) choose algorithms that conform to the workspace size (in [EvaluateInvokers](https://github.com/ROCm/MIOpen/blob/develop/src/conv/solver_finders.cpp#L241)).
However, if a database already exists, it will only choose the fastest one, failing or segfaulting later due to insufficient workspace.

> This PR includes a quiet change of the public API. Such a change must be made explicit at least.

I was not aware that this PR changes public API, as the behavior already occurs the first time `miopenFindConvolution*` is called.

atamazov:
@tbennun 
> Trying to find a convolution algorithm with a workspace size constraint works when the database needs to be regenerated.

It may randomly work in some specific cases, but IIRC the Find API still requires workspace to be not smaller than the respective GetWorkspaceSize API call informs. We thought about allowing Find to return the "best possbile" results if the workspace size is insufficient, but for now the decision is to keep API "strict". The reason is that this may prevent the most performant kernels to be found and used, which is against the purpose of the library -- it exists to deliver performance (perhaps this reasoning is somewhat superficial and we are open for discussion).

atamazov:
@tbennun Oops, I see EvaluateInvokers, let me check why this is so...

atamazov:
@tbennun Yes we have this relaxation, it was implemented in https://github.com/AMDComputeLibraries/MLOpen/pull/2499 (private MIOpen repo). I need to discuss this with the team.

/cc @DrizztDoUrden

atamazov:
@tbennun Right now you can use 
```bash
export MIOPEN_FIND_MODE=Normal
```
That would enforce the library to ignore existing find-db records during Find calls. The new records will be written into the user-find-db, and these new records override the old ones.

After the network is run once, you can unset MIOPEN_FIND_MODE. This is recommended because the Normal Find mode has substantial host-side overhead which may slowdown your network.

[Notice] The by-product of this approach is that you'll likely get the best possible performance for your specific system, because in Normal Find Mode the library builds and benchmarks kernels right on your HW.

atamazov:
@tbennun 
> I was not aware that this PR changes public API, as the behavior already occurs the first time miopenFindConvolution* is called.

My review was incorrect and now dismissed.

atamazov:
@tbennun 
> I was not aware that this PR changes public API, as the behavior already occurs the first time miopenFindConvolution* is called.

Currently the behavior of Find when WS is too small is unspecified. In some cases the library works, is some others it doesn't. This is not good. Your PR makes (or at least is intended to make) the behavior well-specified. _It is in this sense this PR changes the API._

From the practical POV this means that we need to update the documentation in miopen.h

atamazov:
@tbennun After some internal discussion the decision is made to officially support the cases when WS is too small. I'll provide the proper fix soon. The PR will include:
- Some changes similar to this PR
- Appropriate warnings
- Changes in documentation in miopen.h.

/cc @DrizztDoUrden @JehandadKhan @junliume

tbennun:
Thank you, @atamazov! Should I close this PR?

atamazov:
@tbennun Let me open another PR first. I will ask for closing this one explicitly. Thanks.

JehandadKhan:
@atamazov  Any objections to merging this PR? Especially, are there any regression risks for existing functionality ?

atamazov:
@JehandadKhan No objections. I will reimplement it soon to make it more "right" etc

atamazov:
Risks are possible, but low.

JehandadKhan (approved):

Issue #2843, opened 2024-03-24
[MHA] Implement MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR in Graph API


scerzh:
The following descriptors are to be implemented.

MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR
  - [x] C functions and tests
  - [x] #2914

This descriptor should query pattern matcher for available solvers/engines capable of doing the requested calculations. The usercode chooses a solver/engine from the list, a read only attribute of the descriptor, and passes its index as an attribute to the next MIOPEN_BACKEND_ENGINE_DESCRIPTOR descriptor.

PR #2894, opened 2024-04-16
[MHA] Implement MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR C API Interface

scerzh:
Implementation of the following tasks from issue #2843 
- [x] C functions and tests


junliume:
@scerzh please check CI status for the new test failure. Thanks!

src/graphapi/enginefinder.cpp
```diff
@@ -0,0 +1,167 @@
...
+                                        builder.addNode(node);
+                                        ops.push_back(apiDescriptor);
+                                        return;
+                                    }
+                                }
+                                MIOPEN_THROW(miopenStatusBadParm);
+                            });
+
+            mOps            = std::move(ops);
+            mOpGraphBuilder = std::move(builder);
```
amberhassaan: This does not seem right. A user can add multiple nodes by calling `setAttribute` multiple times and we should keep accumulating them.

src/graphapi/enginefinder.cpp
```diff
@@ -0,0 +1,167 @@
...
+            *elementCount                                  = 1;
+            *static_cast<miopenHandle_t*>(arrayOfElements) = mHandle;
+        }
+        else
+        {
+            MIOPEN_THROW(miopenStatusBadParm);
+        }
+        break;
+
+    case MIOPEN_ATTR_OPERATIONGRAPH_OPS:
```
amberhassaan: The `getAttribute` here is also strange in the sense that it's only useful if the user specifies the count as the number of nodes in the graph. If it's lesser, we don't know which nodes to return. I think we shouldn't support `getAttribute` for this case but I'm not insisting. I think the case where count == num nodes is useful but then there is not way to get the count of nodes from a graph in the C API.

src/include/miopen/graphapi/enginefinder.hpp
```diff
@@ -0,0 +1,65 @@
+/*******************************************************************************
```
amberhassaan: let's rename the file to something better, e.g., `opgraph_descriptor.hpp` to indicate that it implements the `descriptor` behavior. Same with the `.cpp` file.

amberhassaan (commented):
Initial comments.

src/include/miopen/graphapi/enginefinder.hpp
scerzh: agree

src/graphapi/enginefinder.cpp
scerzh: Disagree. The attribute is set in one step by passing the whole array of nodes.

src/graphapi/enginefinder.cpp
scerzh: We should return descriptors the same as user has set them but not going beyond the provided storage cutting off the tail. The number of descriptors we actually have should be returned in *elementCount. A user can correct the storage size by calling getAttribute with requestedCount=0 and receiving the actual number in *elementCount; then the user calls it again with the right requestedCount value.

src/graphapi/enginefinder.cpp
amberhassaan: Not sure if the API documentation says that. There can be situations where we cannot add all nodes at once or don't want to add all nodes at once.

amberhassaan: But then maybe you're right. `Set` semantics would mean that you do it in one go. That's just bad API design but let's go with this.

src/include/miopen/graphapi/operationgraph_descriptor.hpp
```diff
@@ -0,0 +1,65 @@
...
+namespace graphapi {
+
+class BackendOperationGraphDescriptor : public BackendDescriptor
+{
+private:
+    miopenHandle_t mHandle = nullptr;
+    std::vector<miopenBackendDescriptor_t> mOps; // to return them in getAttribute
+    OpGraphBuilder mOpGraphBuilder;
+    OpGraph mOpGraph;
+    std::vector<miopenSolution_t> mSolutions;
```
scerzh: @atamazov @DrizztDoUrden @amberhassaan Can we use `miopen::Solution` here?
It seems like `Engine` in graph API is the same as `miopen::Solution`. We need a C++ wrapper for `miopenSolution_t` and that C-interface of it. Isn't it better to directly use `miopen::Solution` without pointers?

CAHEK7 (approved):

src/include/miopen/graphapi/operationgraph_descriptor.hpp
amberhassaan: I'm of the opinion that we create a wrapper to have a name `Engine` that is compatible with the graph API. It also allows us to move away from `Solution` in the future, if the need arises.

src/graphapi/graphapi.cpp
```diff
@@ -73,6 +74,9 @@ miopenBackendCreateDescriptor(miopenBackendDescriptorType_t descriptorType,
         case MIOPEN_BACKEND_OPERATION_RNG_DESCRIPTOR:
             outputDesciptor = new miopen::graphapi::BackendOperationRngDescriptor(); break;
 
+        case MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR:
+            outputDesciptor = new miopen::graphapi::BackendOperationGraphDescriptor(); break;
```
Vsevolod1983: is it a typo? "desciptor, looks like "r" is missing

Vsevolod1983 (dismissed):

src/graphapi/graphapi.cpp
CAHEK7: Everywhere)

src/include/miopen/graphapi/operationgraph_descriptor.hpp
atamazov: @scerzh 
> Can we use `miopen::Solution` here? It seems like `Engine` in graph API is the same as `miopen::Solution`.

Good question. I am also a bit curious why that opaque object handle from public API is used for internal needs. I would prefer holding actual objects by-value.

> We need a C++ wrapper for miopenSolution_t 

It seems like `miopen::Solution` provides the necessary functionality.

> and that C-interface of it

We have it in miopen.h already of course. But in graph API engines are always hidden behind various backend descriptors and function calls, so AFAICS we do not need any dedicated C API for it.

@amberhassaan 
> I'm of the opinion that we create a wrapper to have a name `Engine` that is compatible with the graph API.

AFAICS engines are not exposed in the graph API, pls. see my previous comment, thanks.

amberhassaan: @atamazov : we have plans to release graph API as a C++ API as well.

src/graphapi/graphapi.cpp
scerzh: Yes, it is a typo =)

src/include/miopen/graphapi/operationgraph_descriptor.hpp
atamazov: @amberhassaan Ah, I see. Well, I thought we can fork https://github.com/NVIDIA/cudnn-frontend and replace their backend with ours. If this is acceptable (AFAICS their license allows that)...

amberhassaan: That is another approach too that needs more discussion. I, so far, don't like the idea of adding our backend to their frontend API, because we don't support the backend API in full yet and there's some stuff around engines, knobs etc., that does not make sense and we might not want to support it or we might not know how to support it.

scerzh: @amberhassaan a C++ class representing the descriptor and a builder for it have been added to support MHA forward C++ test

src/graphapi/graphapi.cpp
scerzh: done

src/graphapi/operationgraph_descriptor.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include <miopen/errors.hpp>
+#include <miopen/graphapi/operationgraph_descriptor.hpp>
+
+#include <algorithm>
+
+namespace miopen {
+
+namespace graphapi {
+
+OperationGraph::OperationGraph(miopenHandle_t handle, const OpGraph& opGraph) : mHandle(handle)
```
amberhassaan: Some minor issues:
-- `deref(miopenHandle_t` so that this class stores `Handle` pointer
-- A bit of a naming problem where we now have `OpGraph` and `OperationGraph`. We could call this class `OpGraphDesc` since it deals with descriptor related stuff. I'm open to other ideas as my suggestion is not that great.

src/graphapi/operationgraph_descriptor.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    /* TODO: to be implemented in
+     * [MHA] Implement MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR read only attributes #2914
+     * https://github.com/ROCm/MIOpen/issues/2914
+     */
+    (void)handle;
+    (void)opGraph;
+}
+
+OperationGraph::OperationGraph(miopenHandle_t handle, OpGraph&& opGraph) : mHandle(handle)
```
amberhassaan: why not just define it as `default` and let the compiler generate it for us?

src/graphapi/operationgraph_descriptor.cpp
```diff
@@ -0,0 +1,237 @@
...
+    return *this;
+}
+
+OperationGraph OperationGraphBuilder::build() &
+{
+    if(mHandle == nullptr || !mOpsSet)
+    {
+        MIOPEN_THROW(miopenStatusBadParm);
+    }
+    return {mHandle, OpGraphBuilder(mOpGraphBuilder).build()};
```
amberhassaan: not defining a proper `move` constructor breaks this line.

atamazov: Interesting. In spite of the implicitly-defined move ctor is suitable in this case, this feature is "fragile": the compiler won't define it in many cases (e.g. if someone adds a user-defined dtor etc). So I would recommend using the explicitly-defaulted move ctor for the sake of so called robustness ;)

/cc @amberhassaan @scerzh

src/graphapi/operationgraph_descriptor.cpp
scerzh: I think then we will have to add a suffix to all existing classes related to descriptors: Tensor, Convolution, Pointwise, Rng, VariantPack...

OpGraph seems to be used internally only and it doesn't have a corresponding descriptor. We can move it into an internal namespace.

Another option is to merge these 2 classes into one.

src/graphapi/operationgraph_descriptor.cpp
scerzh: it won't use `handle` argument

src/graphapi/operationgraph_descriptor.cpp
scerzh: ~~I didn't get you. What class are you talking about? (what class does move ctor belong?)~~

Ah, I see.

src/graphapi/operationgraph_descriptor.cpp
scerzh: And another line of classes are BackendTensorDescriptor, BackendVariantPackDescriptor already has -descriptor suffix, this will be confusing with OperationGraphDesc. Moreover I thought we might remove Backend- prefix in future as this API is no longer called Backend API.

Also I didn't get you about deref, do you suggest to deref `handle`?

src/graphapi/operationgraph_descriptor.cpp
scerzh: And this ctor will be developed further by Bibek

src/graphapi/operationgraph_descriptor.cpp
scerzh: For me this `OperationGraph` class looks more like `EngineList`, may be call it this way?

The disadvantage of `EngineList` is that it is too far from its enum name `MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR`.

CuDNN frontend API has such a class and they called it OperationGraph: https://github.com/NVIDIA/cudnn-frontend/blob/1b0b5eac540b7f8fd19b18f1e6b8427c95503348/include/cudnn_frontend_OperationGraph.h#L265-L266

src/graphapi/operationgraph_descriptor.cpp
amberhassaan: @atamazov : I don't think manually written code improves robustness over something automatically generated. We have a simple class here with two members that have well-defined `move` semantics, so setting the `move` constructor to `default` is the best approach IMO as it involves writing less code.

src/graphapi/operationgraph_descriptor.cpp
amberhassaan: Why wait for Bibek (@bghimireamd ) when we can just set it to `default` and call it a day? What special operation needs to happen here? The `OpGraph` has a well-defined `move` constructor.

src/graphapi/operationgraph_descriptor.cpp
amberhassaan: I think the most sensible thing here is to combine `OpGraph` and `OperationGraph` classes and their builders. A couple of points:
1. I do not think of `OpGraph` as an internal class. It's the operation graph for C++ users
2. CUDNN made the decision to put the `handle` in the graph. It does not make sense to me (they must have a reason). To me graph is just a description of operations and their dependencies. It shouldn't even be returning a list of `Engines`, but since we're following them blindly, we do what the API demands.

scerzh: Agree

src/graphapi/operationgraph_descriptor.cpp
scerzh: If we are going to combine OpGraph and OperationGraph, then this code will be redesinged.

src/graphapi/operationgraph_descriptor.cpp
scerzh: done

src/graphapi/operationgraph_descriptor.cpp
scerzh: done

atamazov: [note] @amberhassaan just for clarity
> so setting the `move` constructor to `default` is the best

That's what I am proposing as well (explicitly-defaulted move ctor)

amberhassaan: OK, my bad. I thought you were preferring hand-written `move` constructor to `default` one.

junliume (approved):

PR #2895, opened 2024-04-17
[MHA] backward pass

Vsevolod1983:
1. Mha backward solver
2. Find 2.0 Wrapper over mha solver
3. A GTest for Mha backward Find 2.0 wrapper.
4. Dropout support for forward and backward pass (+test)

This is a shared PR (mine and Alexander's)


src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -150,6 +146,11 @@ __device__ float reductionCommon(const float* __restrict__ line,
 
     return reductionBlock<NumWarps>(reduced_val, op, lid, laneId, warpId);
 };
+
+__device__ bool doDropout(float dropout, rocrand_device::xorwow_engine* state)
```
atamazov: [Q] Maybe it is worth adding `static inline`?

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -150,6 +146,11 @@ __device__ float reductionCommon(const float* __restrict__ line,
...
+{
+    return (dropout > 0.0f && prng::xorwow_uniform(state) < dropout);
```
atamazov, line 157: [Readability]
```suggestion
    return (dropout > 0.0f && dropout > prng::xorwow_uniform(state));
```

src/CMakeLists.txt
```diff
@@ -266,7 +266,8 @@ set( MIOpen_Source
     solver/layernorm/forward_layernorm.cpp
     solver/layernorm/forward_layernorm2d_ck.cpp
     solver/layernorm/forward_layernorm4d_ck.cpp
-    solver/mha/mha_solver.cpp
+    solver/mha/mha_solver_forward.cpp
+    solver/mha/mha_solver_backward.cpp
```
atamazov: [R] Sort
```suggestion
    solver/mha/mha_solver_backward.cpp
    solver/mha/mha_solver_forward.cpp
```

src/mha/problem_description.cpp
```diff
@@ -39,18 +39,50 @@ NetworkConfig ProblemDescription::MakeNetworkConfig() const
 
     ss << "mha";
 
+    auto print_strides = [&ss](const TensorDescriptor& desc) {
+        for(auto d : desc.GetStrides())
```
atamazov: [R] Maybe
```suggestion
        for(const auto& d : desc.GetStrides())
```

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -323,12 +321,16 @@ extern "C" __global__ void SoftMaxCommon(const float* in,
     const uint32_t laneId       = lid % warpSize;
     const uint32_t warpId       = lid / warpSize;
     const float descaler        = (descale_Q ? *descale_Q : 1.0f) * (descale_K ? *descale_K : 1.0f);
-    const float dropout         = (dropout_P && rng_state) ? (*dropout_P) : 0.0f;
-    const float scaler          = (scale_S ? *scale_S : 1.0f) * (1.0f - dropout);
+    const float dropout         = (dropout_P && seed && offset) ? (*dropout_P) : 0.0f;
+    const float scaler          = (scale_S ? *scale_S : 1.0f) / (1.0f - dropout);
     const bool save_stats       = M && Z && lid == 0;
```
atamazov: [Readability]
```suggestion
    const bool save_stats       = !(M && Z && lid);
```

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -323,12 +321,16 @@ extern "C" __global__ void SoftMaxCommon(const float* in,
     const uint32_t laneId       = lid % warpSize;
     const uint32_t warpId       = lid / warpSize;
     const float descaler        = (descale_Q ? *descale_Q : 1.0f) * (descale_K ? *descale_K : 1.0f);
-    const float dropout         = (dropout_P && rng_state) ? (*dropout_P) : 0.0f;
-    const float scaler          = (scale_S ? *scale_S : 1.0f) * (1.0f - dropout);
+    const float dropout         = (dropout_P && seed && offset) ? (*dropout_P) : 0.0f;
+    const float scaler          = (scale_S ? *scale_S : 1.0f) / (1.0f - dropout);
```
atamazov: [Future][Optimization] `dropout` is scalar (uniform-per-all-lanes) value, as well as the presence of scaler/descalers. However I am not sure if the compiler is able to use this information to improve performance when dropout/scaling is not necessary. Let's remember (maybe even open a ticket) this and investigate optimization possibilities in the future.

src/kernels/miopen_rocrand.hpp
```diff
@@ -25,39 +25,51 @@
...
 
-inline constexpr unsigned int xorwow_next(xorwow_state* state)
-{
-    const unsigned int t = state->x[0] ^ (state->x[0] >> 2);
-    state->x[0]          = state->x[1];
-    state->x[1]          = state->x[2];
-    state->x[2]          = state->x[3];
-    state->x[3]          = state->x[4];
-    state->x[4]          = (state->x[4] ^ (state->x[4] << 4)) ^ (t ^ (t << 1));
+#ifndef WORKAROUND_IGNORE_ROCRAND_INCLUDES
```
atamazov: [R] If we have a ticket for rocrand, then it would be better to notice it somehow (WORKAROUND_SWDEV_XXX or just a comment). We can even open a ticket at https://github.com/ROCm/rocRAND -- that repo seems alive.

src/solver/mha/mha_solver_backward.cpp
```diff
@@ -0,0 +1,475 @@
...
+ *******************************************************************************/
+
+#include <miopen/mha/solvers.hpp>
+
+#include <miopen/mha/invoke_params.hpp>
+#include <miopen/datatype.hpp>
+#include <miopen/kernel_build_params.hpp>
+#include <miopen/target_properties.hpp>
+#include <miopen/float_equal.hpp>
+#include <miopen/gemm_v2.hpp>
```
atamazov, line 34: [R] Sort this list.

[Note] There should be no need for <miopen/mha/solvers.hpp> to be the first one in the list.

src/solver/mha/mha_solver_backward.cpp
```diff
@@ -0,0 +1,475 @@
...
+
+#include <algorithm>
+#include <vector>
+#include <tuple>
+
+// disable __device__ qualifiers
+#ifdef FQUALIFIERS
+#error rocrand FQUALIFIERS defined externally, probably one of rocrand device header included prior to this
+#endif
+#define FQUALIFIERS inline
```
atamazov: [Q] Another issue in rocrand?

src/solver/mha/mha_solver_backward.cpp
```diff
@@ -0,0 +1,475 @@
...
+    return !miopen::IsDisabled(ENV(MIOPEN_DEBUG_ATTN_NAIVE_BWD)) && //
+           !problem.IsForward() &&                                  //
+           S <= std::numeric_limits<uint32_t>::max() &&             //
+           D <= std::numeric_limits<uint32_t>::max() &&             //
+           descsForward.kDesc.IsPacked() &&                         //
+           descsForward.qDesc.IsPacked() &&                         //
+           descsForward.vDesc.IsPacked() &&                         //
+           descsForward.oDesc.IsPacked() &&                         //
+           descsForward.doDesc.IsPacked() &&                        //
+           MIOPEN_USE_GEMM;
```
atamazov: [Recommendation][Readability] Usually I move operators to the next line, like this:
```C++
    return !miopen::IsDisabled(ENV(MIOPEN_DEBUG_ATTN_NAIVE_BWD)) //
           && !problem.IsForward()                              //
           ...
           && descsForward.doDesc.IsPacked()                     //
           && MIOPEN_USE_GEMM;
```

src/solver/mha/mha_solver_backward.cpp
```diff
@@ -0,0 +1,475 @@
...
+}
+
+ConvSolution MhaBackward::GetSolution(const ExecutionContext& context,
+                                      const miopen::mha::ProblemDescription& problem) const
+{
+    auto result = ConvSolution{miopenStatusSuccess};
+
+    auto [N, H, S, D] = miopen::tien<4>(problem.GetDescsBackward().kDesc.GetLengths());
+    uint32_t emb_dim  = D;
+    uint32_t seq_len  = S;
```
atamazov: [Note] clang-tidy may complain here. Also explicit cast emphasizes the presence of some intent.

src/kernels/MIOpenSoftmaxAttn.cpp
```diff
@@ -436,3 +431,327 @@ extern "C" __global__ void ScaleReduce(const float* __restrict__ in,
...
+extern "C" __global__ void ScaleRowReduceBlock(const float* __restrict__ dO,
+                                               const float* __restrict__ O,
+                                               float* __restrict__ out,
+                                               const float* __restrict__ descale_dO,
+                                               const float* __restrict__ descale_O,
+                                               const float* __restrict__ dropout_P,
+                                               uint32_t d,
+                                               uint64_t nhs)
+{
+    constexpr uint32_t NumWarps = THREADS / warpSize;
```
atamazov, line 495: [Q] Is it guaranteed that `THREADS % warpSize == 0`? If not, then we have to handle this case somehow (add 1 at least + possible more). Or this is guaranteed by the host code?

src/kernels/miopen_rocrand.hpp
```diff
@@ -25,39 +25,51 @@
...
+// rocrand_uniform.h has too many dependencies and device code
 inline constexpr float uniform_distribution(unsigned int v)
 {
-    constexpr float ROCRAND_2POW32_INV = 2.3283064e-10f;
-    return ROCRAND_2POW32_INV + (static_cast<float>(v) * ROCRAND_2POW32_INV);
+    constexpr float rocrand_2pow32_inv =
+#ifdef ROCRAND_2POW32_INV
+        ROCRAND_2POW32_INV;
+#else
+        2.3283064e-10f;
```
atamazov, line 66: [R] Modern literal for old men 🕶️ 
```suggestion
        0x1p-32f;
```

atamazov (commented):
I do not see anything bad, but have some questions/advices.

src/kernels/MIOpenSoftmaxAttn.cpp
CAHEK7: Nameless namespace is a modern way of `static`.
Probably `inline` or even `__force_inline__` is a good suggestion.

src/mha/problem_description.cpp
CAHEK7: A common pitfall. Sure.

src/kernels/MIOpenSoftmaxAttn.cpp
CAHEK7: 1) I would like to explicitly state that `lid` is index and must be zero.
2) It is `(M) && (Z) && (lid == 0)`, not `(M && Z && lid) == 0`
3) probably parenthesis is a good solution for readability.

src/solver/mha/mha_solver_backward.cpp
CAHEK7: Not quite an issue.
Actually it's not needed here at all, since I don't allocate random state anymore.

But for the test I have to use some rocrand functions in a host environment, but they did not mark anything as a `__host__` functions.
I can copy-paste them, but don't see any points right now - that define does the trick.

src/kernels/MIOpenSoftmaxAttn.cpp
CAHEK7: Compiler just put this value on a register - either 0 or something else. I'm not sure what kind of optimization do you expect from it?
Like omitting operations at all? That's hardly possible.

The only thing which is forgotten is `__launch_bounds__(THREADS)`

src/solver/mha/mha_solver_backward.cpp
CAHEK7: clang-format \ clang-tidy must do this. Somebody has to turn on this rule on and close this discussion once and forever.

> [Note] There should be no need for <miopen/mha/solvers.hpp> to be the first one in the list.

I guess it's copy-pasted from the other solver. There is an `iwyu` tool, stated for "include-what-you-use", but I have no ideas how to make it work with our specific clang.

src/solver/mha/mha_solver_backward.cpp
CAHEK7: As prefixes, ok.

src/solver/mha/mha_solver_backward.cpp
CAHEK7: Luckily it does not, but you are right.

src/kernels/MIOpenSoftmaxAttn.cpp
CAHEK7: It is "guaranteed" by the host code and a common sense, but I'll add `static_assert` here.

src/kernels/miopen_rocrand.hpp
CAHEK7: probably, but since it is copied from `rocrand`, I'd prefer to keep it similar.

src/kernels/MIOpenSoftmaxAttn.cpp
CAHEK7: I'm not quite sure, for me `random_value < threshold` looks more like a statement `the value is in the range [0; threshold)`.
Actually I don't care much,

src/kernels/MIOpenSoftmaxAttn.cpp
atamazov: Optimization won't give anything substantial [resolved]

src/solver/mha/mha_solver_backward.cpp
atamazov: [Recommendation] @CAHEK7 Can you please add a comment about this.

src/solver/mha/mha_solver_forward.cpp
```diff
@@ -65,8 +58,11 @@ constexpr S RoundUpToMultiple(T val, T mul)
     return Ceil(val, mul) * mul;
 }
 
-constexpr uint32_t nextPow2(uint32_t v)
+template <typename T>
+constexpr T nextPow2(T v)
 {
+    static_assert(std::is_integral_v<T>);
```
atamazov: [R] I am afraid this function works correctly only with unsigned types (or positive signed values). If I am right then we need `assert(v>=0)` for signed integer types.

atamazov (approved):
🟢 LGTM!

src/solver/mha/mha_solver_forward.cpp
CAHEK7: assert is a bad option for constexpr functions, but we can always do `static_assert(std::is_unsigned_v<>T);`

Vsevolod1983: Changed

src/kernels/miopen_rocrand.hpp
```diff
@@ -25,39 +25,51 @@
...
-    state->x[4]          = (state->x[4] ^ (state->x[4] << 4)) ^ (t ^ (t << 1));
+#ifndef WORKAROUND_IGNORE_ROCRAND_INCLUDES
+#define WORKAROUND_IGNORE_ROCRAND_INCLUDES 1
+#endif
 
-    state->d += 362437;
+#if WORKAROUND_IGNORE_ROCRAND_INCLUDES == 1
+// disable math.h from rocrand (it conflicts with hiptrc)
+// NOLINTNEXTLINE
+#define _GLIBCXX_MATH_H 1
```
junliume, line 40: @Vsevolod1983 this is not working properly with offline compiler, i.e. `-DMIOPEN_USE_COMGR=Off`

CAHEK7 (approved):

junliume (approved):

src/kernels/miopen_rocrand.hpp
Vsevolod1983: @CAHEK7 ^^

CAHEK7: @junliume it passes NO_COMGR tests in CI.
And we tested it with @atamazov locally in different configurations.

PR #2896, opened 2024-04-18
ConvProblemDescription: fix GetInSize(), GetOutSize() and GetWeightsSize()

averinevg:
Currently, strides are not taken into account in `GetInSize()`, `GetOutSize()` and `GetWeightsSize()`. This PR fixes the issue.

Additional changes:

- Some code refactoring in the `ProblemDescription`
- TensorDescriptor: Misleading `GetSize()` has been replaced with `GetNumDims()`
- Some dead code has been removed

### DIFF
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -331,9 +331,15 @@ const std::vector<std::size_t>& TensorDescriptor::GetLengths() const { return le
 
 const std::vector<std::size_t>& TensorDescriptor::GetStrides() const { return strides; }
 
-int TensorDescriptor::GetSize() const
+unsigned TensorDescriptor::GetNumDims() const
 {
     return lens.size();
+}
+
+// This name is misleading, all the usages shold be replaced with GetNumDims()
+int TensorDescriptor::GetSize() const
+{
+    return GetNumDims();
 }
 
 std::size_t TensorDescriptor::GetElementSize() const
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -333,7 +333,7 @@ const std::vector<std::size_t>& TensorDescriptor::GetStrides() const { return st
 
 unsigned TensorDescriptor::GetNumDims() const { return lens.size(); }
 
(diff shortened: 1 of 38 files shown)


CAHEK7 (approved):

atamazov:
@averinevg 
> * Tensor: Added `GetNumDims()` as a replacement for `GetSize()`

Maybe get rid of GetSize at once?

src/tensor.cpp
```diff
@@ -331,15 +331,13 @@ const std::vector<std::size_t>& TensorDescriptor::GetLengths() const { return le
...
-    return lens.size();
-}
+unsigned TensorDescriptor::GetNumDims() const { return lens.size(); }
+
+// This name is misleading, all the usages should be replaced with GetNumDims()
+int TensorDescriptor::GetSize() const { return GetNumDims(); }
 
 std::size_t TensorDescriptor::GetElementSize() const
 {
-    assert(lens.size() == strides.size());
```
atamazov, line 342: What was wrong with this assert?

atamazov (approved):
🟢 LGTM! Can we add a regression test or this is a proactive fix?

src/tensor.cpp
averinevg: It is already checked in the constructor.

averinevg:
@atamazov 
> Can we add a regression test or this is a proactive fix?

I noticed this while working on another pull request, it's a proactive fix.

atamazov (approved):
💚 LGGTM

junliume (approved):

test/tensor_util.hpp
```diff
@@ -74,114 +71,4 @@ void operate_over_subtensor_impl(const data_operator_t<T>& r_data_operator,
     }
 }
 
-template <typename T>
```
bghimireamd, line 77: Why did we delete these function?

averinevg: Looks like it was used for debugging, or someone accidentally left it here. In any case, this code is not used anywhere, and I decided to remove it.

PR #2897, opened 2024-04-18
Gtest nchw nhwc

randyspauldingamd:
Updated nchw_nhwc test to gtest. I used a minimalistic strategy; no changes were made to the existing tests.

- moved test/gpu_nchw_nhwc_transpose.cpp -> test/gtest/gpu_nchw_nhwc_transpose.hpp
- made changes to gpu_nchw_nhwc_transpose.hpp to use gtest instead of ctest
- added macros to define tests and added test/gtest/gpu_nchw_nhwc_transpose.cpp, as the cpp seems to be required in order for CMake to identify the test and generate the executable


test/gtest/gpu_nchw_nhwc_transpose.hpp
atamazov, line 1: This PR silently removes `test_gpu_nchw_nhwc_transpose` executable. Is this side effect desirable?

test/gtest/gpu_nchw_nhwc_transpose.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "gpu_nchw_nhwc_transpose.hpp"
+
+DEFINE_DEFAULT_TO_NHWC_TEST(float);
+DEFINE_DEFAULT_TO_NHWC_TEST(uint16_t);
+DEFINE_DEFAULT_TO_NHWC_TEST(uint8_t);
+
+DEFINE_NHWC_TO_DEFAULT_TEST(float);
+DEFINE_NHWC_TO_DEFAULT_TEST(uint16_t);
+DEFINE_NHWC_TO_DEFAULT_TEST(uint8_t);
```
atamazov: `gpu_nchw_nhwc_transpose.hpp` is included only once. Please move macros to the end of it, remove this file and rename `gpu_nchw_nhwc_transpose.hpp` to `gpu_nchw_nhwc_transpose.cpp`.

atamazov (changes requested):

atamazov:
@randyspauldingamd BTW how long does this test take to run?

randyspauldingamd:
[AMD Official Use Only - General]

Thanks for the suggestion! I was mimicking a pattern used by Amber. I don’t know a lot of details about cmake and gtest, but indeed, not having the .cpp file does prevent the executable from being built.

I believe it would be possible to eliminate extraneous cpp files via cmake changes. That would be separate work that should not hold up this PR.

Cheers,

………………………………………………………..
Randy J. Spaulding
SMTS, Software Development Engr.
Advanced Micro Devices, Inc.
Machine and Software Dev (2-MSD)

atamazov:
@randyspauldingamd Thanks for quick feedback. Can you please provide it in usual places on the github page, when you have time? Thanks.

randyspauldingamd:
@junliume I created this PR in my fork intending to use it for practice, not realizing that github had selected the original repo as the destination by default. Since it exists, it ok to submit PR's this way, or should I re-submit it directly to the MIOpen repo?

randyspauldingamd:
> @randyspauldingamd Thanks for quick feedback. Can you please provide it in usual places on the github page, when you have time? Thanks.

I think that answers the question I was just asking Jun, haha. Yes, I'll remove this one, and submit it in MIOpen instead.

randyspauldingamd:
> @randyspauldingamd BTW how long does this test take to run?

It takes 550-700 ms depending on spinup time which is most of it. Interestingly, it is much faster than the original tests (appx 2.5 seconds) even though there were no changes made and all of the original tests are being executed. It seems that the ctest infrastructure restarts the GPUs or ROCm many times. These gtests only start them once.

test/gtest/gpu_nchw_nhwc_transpose.cpp
randyspauldingamd: Hi Artem, I just realized that I misread your comment the other day. Yes, it makes sense to move all of the code into the cpp file. I'll do that.

randyspauldingamd:
deleting this PR...

PR #2898, opened 2024-04-18
KernelTuningNet for MI300/200 ConvHipIGemmGrouped Solvers

Dmantri98:
MI300 Runtimes:

Forward:
Mean of Default GPU Kernel Time: 0.6508 ms
Standard Deviation of Default GPU Kernel Time: 2.6097ms
Mean of Model GPU Kernel Time: 0.3782ms
Standard Deviation of Model GPU Kernel Time: 2.0900ms

Backward Data:
Mean of Default GPU Kernel Time: 0.5457ms
Standard Deviation of Default GPU Kernel Time: 2.1012ms
Mean of Model GPU Kernel Time: 0.2464ms
Standard Deviation of Model GPU Kernel Time: 1.1348ms

Backward Weights:
Mean of Default GPU Kernel Time: 22.6547ms
Standard Deviation of Default GPU Kernel Time: 182.6347ms
Mean of Model GPU Kernel Time: 12.79862ms
Standard Deviation of Model GPU Kernel Time: 133.5387ms

Avg Model Run Time: 1077.5548 microseconds.

MI200 Runtimes (Forward is already integrated):

Backward Data:
Mean of Default GPU Kernel Time: 2.11048ms
Standard Deviation of Default GPU Kernel Time: 7.4350ms
Mean of Model GPU Kernel Time: 0.6051ms
Standard Deviation of Model GPU Kernel Time: 2.8523ms

Backward Weights:
Mean of Default GPU Kernel Time: 12.0965ms
Standard Deviation of Default GPU Kernel Time: 57.8867ms
Mean of Model GPU Kernel Time: 5.9298ms
Standard Deviation of Model GPU Kernel Time: 26.5806ms

Avg Model Run Time: 998.2971 mircoseconds.

These models address an issue related to model performance where there is a warmup loop, so even though it may seem costly to run the model in comparison to the speed improvements, the invokers will be cached providing a speed improvement across the entire model run.


msaudulhassan:
For MI300, I'm not sure if integrating KTN would make things better. For example, in the forward direction, KTN-predicted kernel takes 0.38ms to execute while the default kernel takes 0.65ms. However, running KTN itself takes about 1.08ms. So if we use KTN in the forward direction, it'd take a total of 0.38+1.08 ms, which is worse than using the default kernel. However, I understand that running KTN is a one-time thing, and thus the time can get amortized over multiple calls. Is there a way to measure the amortized performance over a practical(-ish) workload?

src/kernels/gfx942_ConvHipIgemmGroupXdlops_metadata.ktn.model
```diff
@@ -0,0 +1,71 @@
...
+            "17": "Filter1x1Stride1Pad0",
+            "18": "OddC",
+            "19": "Filter1x1Pad0",
+            "20": "8",
+            "21": "4",
+            "22": "2",
+            "23": "32",
+            "24": "Default",
+            "25": "2",
+            "26": "Filter1x1Stride1Pad0",
```
msaudulhassan, line 35: I am under the impression that the network predicts tuning parameters as tokens (the keys in this dictionary) and then those tokens get mapped to actual tuning parameters (the values in this dictionary). If that's the case, I'm not sure I understand what, for example, "26" mapping to "Filter1x1Stride1Pad0" means?

msaudulhassan: Could this be a bug in data processing?

src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -616,9 +618,19 @@ bool ModelSetParams(const std::string& arch,
...
-    for(std::size_t i = 0; i < model->metadata.num_tuning_params; ++i)
+    std::string dir;
+    switch(direction)
+    {
+    case miopen::conv::Direction::Forward: dir = "fwd"; break;
+    case miopen::conv::Direction::BackwardData: dir = "bwd"; break;
+    case miopen::conv::Direction::BackwardWeights: dir = "wrw"; break;
+    default: return false;
+    }
+    std::size_t num_tuning_params = model->metadata.num_tuning_params[dir];
```
msaudulhassan, line 632: In `src/kernels/gfx908_ConvAsm1x1U_metadata.ktn.model`, only "fwd" is present under "num_tuning_parameters". So if the GPU is gfx908 and the solver is ConvAsm1x1U and `direction` is `miopen::conv::Direction::BackwardData`, line 632 would error out

src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -606,6 +607,7 @@ std::shared_ptr<Model> GetModel(const std::string& arch, const std::string& solv
 
 bool ModelSetParams(const std::string& arch,
                     const std::string& solver,
+                    miopen::conv::Direction direction,
```
msaudulhassan, line 610: minor: direction can be made `const`

src/kernels/gfx942_ConvHipIgemmGroupXdlops_metadata.ktn.model
Dmantri98: It's a tuning parameter. Sort of how language models predict an index within the output vocabulary and they then get tokenized to a string. It's not actually predicting token/values in and of themselves but rather indices that correspond to a given token.

src/conv/heuristics/ai_heuristics.cpp
Dmantri98: Both "fwd" and "bwd" are defined in the metadata file.

Dmantri98:
> For MI300, I'm not sure if integrating KTN would make things better. For example, in the forward direction, KTN-predicted kernel takes 0.38ms to execute while the default kernel takes 0.65ms. However, running KTN itself takes about 1.08ms. So if we use KTN in the forward direction, it'd take a total of 0.38+1.08 ms, which is worse than using the default kernel. However, I understand that running KTN is a one-time thing, and thus the time can get amortized over multiple calls. Is there a way to measure the amortized performance over a practical(-ish) workload?

This should be done by the testing team, I am not sure what customer workloads would be, but to my understanding there is usually a warm-up run done before running workloads. These warm-up runs would not count towards benchmarks our customers would use.

src/kernels/gfx942_ConvHipIgemmGroupXdlops_metadata.ktn.model
Dmantri98: For example this is a full CK config:
DeviceGroupedConvFwdMultipleABD_Xdl_CShuffle<256, 128, 128, 16, Default, 32, 32, 2, 2, 4, 4, 4, 1, 1>
The "Default" is a tuning parameter.

src/conv/heuristics/ai_heuristics.cpp
msaudulhassan: What about `wrw`? Also, src/kernels/gfx90a_ConvHipIgemmGroupFwdXdlops_metadata.ktn.model only has `fwd`

src/kernels/gfx942_ConvHipIgemmGroupXdlops_metadata.ktn.model
msaudulhassan: I see. And to double check, is `Filter1x1Stride1Pad0` a tuning parameter, too?

msaudulhassan: Ah another thing. Both "17" and "26" map to "Filter1x1Stride1Pad0". I believe you encoded the tuning parameters and then inverted the encodings to get the `decodings` dictionary in your metadata file. Assuming Filter1x1Stride1Pad0 is indeed a tuning parameter, I'm confused as to how it could have been encoded to two different ordinals

Dmantri98: So, this is a byproduct of a decision made a while back. Instead of just tokenizing all unique values throughout the entire vocabulary I would tokenize all unique values within each timestep, as you can see there are multiple "32"s, "64"s,  etc. The overlap of non-numerical tuning parameters has to do with the fact that the model "ConvHipIgemmXdlops" is trained to produce configs/tuning strings for each solver wrw/bwd/fwd and these solvers have the "type", (non-numerical tuning parameter such as "Filter1x1Stride1Pad0"), located at a different timestep/index within the config string.
An example being:

Forward:
DeviceGroupedConvFwdMultipleABD_Xdl_CShuffle<256, 128, 128, 16, Default, 32, 32, 2, 2, 4, 4, 4, 1, 1>

Backward Data:
DeviceGroupedConvBwdDataMultipleD_Xdl_CShuffle_v1<128, 128, 32, 32, 8, 8, Default, 32, 32, 2, 1, 8, 8, 1, 1>

As you can see the type is in a different position within the config string. Now I could revert this decision and tokenize only unique values that are unique throughout the entire vocabulary, but the improvement in the model would have to be tested empirically. I don't know if it would actually lead to better performance.

src/conv/heuristics/ai_heuristics.cpp
Dmantri98: ConvHipIgemmGroupFwdXdlops is only a forward direction solver. ConvAsm1x1U solver is not applicable for backward weights (https://github.com/ROCm/MIOpen/pull/2898/files#diff-d69d92a5ae6963da140aa45d0ec4ef7317f368a93fbc6076c783142ef3be7d82L535).

msaudulhassan (approved):
lgtm

cderb (approved):

src/solver/conv_hip_implicit_gemm_grouped_bwd_xdlops.cpp
```diff
@@ -216,15 +217,139 @@ bool ConvHipImplicitGemmGroupBwdXdlops::CheckCKApplicability(
...
+
+    if(eraseBegin != heuristic_indexes.begin())
+    {
+        heuristic_indexes.erase(eraseBegin, heuristic_indexes.end());
+        return true;
+    }
+    return false;
+}
+
+static std::vector<float> GetFeatures(const ProblemDescription& problem, std::size_t num_cu)
```
junliume: @Dmantri98 [[maybe_unused]] for num_cu?

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
```diff
@@ -212,15 +214,138 @@ bool ConvHipImplicitGemmGroupWrwXdlops::CheckCKApplicability(
...
+
+    if(eraseBegin != heuristic_indexes.begin())
+    {
+        heuristic_indexes.erase(eraseBegin, heuristic_indexes.end());
+        return true;
+    }
+    return false;
+}
+
+static std::vector<float> GetFeatures(const ProblemDescription& problem, std::size_t num_cu)
```
junliume: Ditto

junliume (approved):

PR #2899, opened 2024-04-18
Unify 'include amd_comgr.h' between Windows and Linux

apwojcik:
Windows MIOpen moved to HIP SDK 6.0 and no longer supports HIP SDK 5.7 out of the box. However, HIP SDK 5.7 might still be used to compile MIOpen, but with slight modifications to the source code.

### DIFF
--- a/src/comgr.cpp
+++ b/src/comgr.cpp
@@ -37,11 +37,7 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/stringutils.hpp>
 
-#if !defined(_WIN32)
 #include <amd_comgr/amd_comgr.h>
-#else
-#include <amd_comgr.h>
-#endif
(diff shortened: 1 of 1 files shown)


src/comgr.cpp
```diff
@@ -37,11 +37,7 @@
 #include <miopen/solver/implicitgemm_util.hpp>
 #include <miopen/stringutils.hpp>
 
-#if !defined(_WIN32)
```
atamazov, line 40: If you would like to keep backward compatibility with HIP SDK 5.7, then 
```C++
#if !defined(_WIN32) && HIP_PACKAGE_VERSION_MAJOR < 6
```
would do the job.

atamazov (approved):
🟢 LGTM

junliume (approved):

PR #2905, opened 2024-04-19
fix execution of a HIP compiler on Windows

apwojcik:
Windows uses the WIN32 API to execute a subprocess. No command shell is spawned, so stream redirections or setting process environment variables do not work. Instead, appropriate parameters need to be provided to the CreateProcess() WIN32 call.


src/hip/hip_build_utils.cpp
```diff
@@ -174,20 +174,20 @@ static fs::path HipBuildImpl(boost::optional<TmpDir>& tmp_dir,
...
-        const std::string cmd        = env + std::string(" ") + MIOPEN_HIP_COMPILER;
-        const std::string args       = params + filename + " -o " + bin_file + redirector;
-        tmp_dir->Execute(cmd, args);
+        std::string args = params + filename + " -o " + bin_file;
+#ifndef _WIN32
+        // Windows uses WIN32 API to execute a subprocess. None command shell is spawned.
+        if(testing_mode)
+            args += " 1>/dev/null 2>&1";
+#endif
+        tmp_dir->Execute(MIOPEN_HIP_COMPILER, args, env);
```
atamazov: [Notice] @apwojcik AFAICS `env` remains empty. Seems like this is a rudiment of the past. So the support for setting the enviroment in `TmpDir::Execute()` is not really needed. I guess you may wish to remove it to simplify the code. Of course, this is up to you.

src/hip/hip_build_utils.cpp
```diff
@@ -30,6 +30,7 @@
 #include <miopen/exec_utils.hpp>
 #include <miopen/logger.hpp>
 #include <miopen/env.hpp>
+#include <miopen/stringutils.hpp>
 #include <miopen/solver/implicitgemm_util.hpp>
```
atamazov: [Recommendation] Keep sorted.
```suggestion
#include <miopen/solver/implicitgemm_util.hpp>
#include <miopen/stringutils.hpp>
```

atamazov (approved):
🟢 LGTM

src/hip/hip_build_utils.cpp
CAHEK7: No. We must turn clang-format automatic sorting on and forget about such comments.
As simple as https://github.com/ROCm/MIOpen/blob/0334279c5214329acd8f7dfe20b31182e17c6bdb/.clang-format#L74

junliume:
@apwojcik and @atamazov this change is meeting with issues when offline compiler is been used:
```
-DMIOPEN_USE_COMGR=Off
```
and 
```
 sh: 1: Syntax error: "(" unexpected
 ```

File: src/hip/hip_build_utils.cpp (hunk 2 above)
Comment (atamazov) [reply src/hip/hip_build_utils.cpp]: @CAHEK7 Enabling this right here would result in too many unrelated changes. Let's do this separately, in some future PR. Proper manual sorting in advance prevents possibility of issues there. Agree?

Comment (atamazov) [reply src/hip/hip_build_utils.cpp]: @CAHEK7 Anyway this is just a recommendation that does not require explicit feedback.

**Comment by atamazov** [comment]
@apwojcik Please try to fix this yourself or ask me for assistance (up to you).

File: src/hip/hip_build_utils.cpp [hunk 3]
```diff
@@ -174,20 +171,20 @@ static fs::path HipBuildImpl(boost::optional<TmpDir>& tmp_dir,
 #endif
 
     // hip version
-    params +=
-        std::string(" -DHIP_PACKAGE_VERSION_FLAT=") + std::to_string(HIP_PACKAGE_VERSION_FLAT);
-
-    params += " ";
+    params += " -DHIP_PACKAGE_VERSION_FLAT=" MIOPEN_STRINGIZE(HIP_PACKAGE_VERSION_FLAT) " ";
```
atamazov: `HIP_PACKAGE_VERSION_FLAT` expands to a parenthesized expression. Please use std::to_string. Thanks.

atamazov (changes requested):
The likely reason of https://github.com/ROCm/MIOpen/pull/2905#issuecomment-2070998270

junliume (approved):

src/hip/hip_build_utils.cpp
```diff
@@ -174,20 +171,20 @@ static fs::path HipBuildImpl(boost::optional<TmpDir>& tmp_dir,
...
+        std::string args = params + filename + " -o " + bin_file;
+#ifndef _WIN32
+        // Windows uses WIN32 API to execute a subprocess. None command shell is spawned.
+        if(testing_mode)
+            args += " 1>/dev/null 2>&1";
+#endif
+        tmp_dir->Execute(MIOPEN_HIP_COMPILER, args);
         if(!fs::exists(bin_file))
-            MIOPEN_THROW("Failed cmd: '" + cmd + "', args: '" + args + '\'');
+            MIOPEN_THROW("Failed cmd: '" MIOPEN_HIP_COMPILER "', args: '" + args + '\'');
```
junliume, line 187: ref: #2958

atamazov: @junliume @apwojcik Yeah, this is optimization by-product of this PR that leads to an issue. It is similar to the one with `HIP_PACKAGE_VERSION_FLAT` (see https://github.com/ROCm/MIOpen/pull/2905#discussion_r1575449537 above). We should assume that `MIOPEN_HIP_COMPILER` expands to an expression that can be used as an initializer for std:string, but the assumption here is that it expands to a string literal.

PR #2906, opened 2024-04-19
fix lock file path creation on Windows

apwojcik:
The character `:` is reserved on Windows and cannot be used for constructing a path except when it follows a drive letter.


src/lock_file.cpp
```diff
@@ -89,6 +90,12 @@ LockFile::LockFile(const char* path_, PassKey) : path(path_)
 
 LockFile& LockFile::Get(const char* path)
 {
+#ifdef _WIN32
+    // The character ':' is reserved on Windows and cannot be used for constructing
+    // a path except when it follows a drive letter.
+    std::string file{std::regex_replace(path, std::regex(":memory:"), "memory_")};
+    path = file.c_str();
```
atamazov, line 97: [Question] @JehandadKhan @Dmantri98 Is is related to the memvfs used to hold SQLite databases in memory? And do we use normal FS for locking?

src/lock_file.cpp
```diff
@@ -89,6 +90,12 @@ LockFile::LockFile(const char* path_, PassKey) : path(path_)
 
 LockFile& LockFile::Get(const char* path)
 {
+#ifdef _WIN32
+    // The character ':' is reserved on Windows and cannot be used for constructing
+    // a path except when it follows a drive letter.
+    std::string file{std::regex_replace(path, std::regex(":memory:"), "memory_")};
```
atamazov, line 96: [Recommendation] @apwojcik AFAICS the following special symbols are allowed for the all filesystems of interest:
- `# $ % & ' ( ) * + - ^ _  { } ~`.

If we see disallowed symbol, then we can translate it to something like `{{decimal-code}}`. For ASCII colon, this gives `{{58}}`. This would reduce probability of accidental matching to some other file.

atamazov, line 96: [Recommendation] @apwojcik Maybe `ReplaceString()` from stringutils.hpp would work faster than regex?

atamazov (approved):
🟡 LGTM

src/lock_file.cpp
apwojcik: Yes, that is related to the memvfs, and we use normal FS locking. I am not super happy with this fix. If you think it is not required, I would gladly replace it with something else, like not locking when working with an in-memory database.

src/lock_file.cpp
apwojcik: Unfortunately, we cannot replace it based on the character. For example, for the path `C:\Some\Path\:memory:\and\the_rest.lock`, the first `:` character has to stay. The other occurrences should be replaced. However, the "drive letter" may actually be a multi-letter signature (e.g., `AZ:\`) for large storage systems.
It is even more complicated because the path may use UNC format: `\\?\C\Some\Path\:memory:\and\the_rest.lock` or even worse `\\?\Volume{4c1b02c1-d990-11dc-99ae-806e6f6e6963}\Some\Path\:memory:\...`, etc. So, just finding the first occurrence will not work. First, we need to know the format used; things are getting complicated quickly. The easiest way is to create paths only with charters allowed on both Linux and Windows.

apwojcik: > * `# $ % & ' ( ) * + - ^ _  { } ~`.

That is only valid when using NTFS. On FAT, those are not allowed.

src/lock_file.cpp
JehandadKhan: IIUC the special path `:memory:` never results in file IO because sqlite would consume it completely. @apwojcik  Are you saying that the sqlite behavior on windows is different. 

Keep in mind that even on Linux a file name `:memory:` would not be created.

src/lock_file.cpp
JehandadKhan: Please see comment above

JehandadKhan (dismissed):
The lockfile name should never have `:memory:` in it to begin with. Therefore I dont understand why this is necessay? 

If the above string does leak in to the lock file name, that would be a bug and needs to be triaged as such.

apwojcik:
> The lockfile name should never have `:memory:` in it to begin with. Therefore, I don't understand why this is necessary.
> 
> If the above string leaks into the lock file name, that would be a bug and needs to be triaged.

@JehandadKhan It is leaking into the lock file, and Windows is trying to create such a file and reports an error because, as I mentioned, `:` cannot be used except following the drive letter. What is the next step? This is a blocker on Windows. Without this PR, we are unable to run MIOpen on Windows.

src/lock_file.cpp
apwojcik: > IIUC the special path `:memory:` never results in file IO because sqlite would consume it completely. @apwojcik Are you saying that the sqlite behavior on windows is different.
> 
> Keep in mind that even on Linux a file name `:memory:` would not be created.

@JehandadKhan On Windows, it is trying to create such a path. That is the reason why we came up with this PR.

junliume (approved):
@JehandadKhan if the review opinion is not blocking, we can move forward with this PR

PR #2909, opened 2024-04-19
[Tests] Make would fail with no device error without GPUs

xinlipn:
Steps to reproduce:

1. docker run -it --rm --name binarygtest -v /home/shinlee/MIOpenG/MIOpen:/MIOpen rocm/miopen:ci_8dc537

2. rm -rf build/

3. mkdir build; cd build

4. export ROCM_PATH=/opt/rocm 
5. cmake -DMIOPEN_BACKEND=HIP -DCMAKE_CXX_COMPILER="${ROCM_PATH}/llvm/bin/clang++" -DCMAKE_C_COMPILER="${ROCM_PATH}/llvm/bin/clang" -DCMAKE_PREFIX_PATH="${ROCM_PATH};${ROCM_PATH}/hip;${DEPS_DIR}" -DHIP_OC_COMPILER="${ROCM_PATH}/bin/clang-ocl" -DMIOPEN_TEST_DISCRETE=OFF ..

6. make -j$(nproc)

Failure:
[make_fail.log](https://github.com/ROCm/MIOpen/files/15045859/make_fail.log)

Pass with changes:
[make_pass.log](https://github.com/ROCm/MIOpen/files/15045869/make_pass.log)


atamazov:
[IMPORTANT] @xinlipn @junliume @JehandadKhan Do we need to run gtests via CTest when `MIOPEN_TEST_DISCRETE` is set to OFF?

atamazov:
The root of the problem is that our non-discrete test requires a GPU to even run `--gtest_list_tests`. Is this correct behavior, what do you think? Maybe we should investigate the reason of it.

atamazov:
@xinlipn @junliume @JehandadKhan 

The original CMakeLists.txt has the detection of the "NO GPU case" and then
```
if(MIOPEN_NO_GPU)
    set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params
            test_test_errors test_type_name test_tensor_test test_sqlite_perfdb test_sequences
            test_pooling3d test_perfdb)
endif()
```
Most likely the root reason of the issue is that MIOPEN_NO_GPU related stuff was not ported to gtests.

- (A) If we do not need non-discrete gtest to work without GPU, then we may
  - skip discovering, like in this PR
  - engage `DISCOVERY_MODE PRE_TEST` in `gtest_discover_tests`, so the gtest won't be run during the build step.
- (B) If we want the non-discrete gtest to work without GPU under CMake/CTest control but _**unconditionally fail** when run in standalone mode_, then we may simply pass the `MIOPEN_NO_GPU` via environment and check it when necessary (when gtest is run under CTest, see [#1843](https://github.com/ROCm/MIOpen/issues/1843)). This is similar to (A) but a bit simpler.
- (C) If we want the non-discrete gtest to work without GPU (like the original make check), then we need to fully port MIOPEN_NO_GPU functionality.

I suspect that for (B) and (C) we may need to engage `DISCOVERY_MODE PRE_TEST` in any case.

xinlipn:
> [IMPORTANT] @xinlipn @junliume @JehandadKhan Do we need to run gtests via CTest when `MIOPEN_TEST_DISCRETE` is set to OFF?

Hi @atamazov , The idea was to use gTest to replace CTest eventually. So we wouldn't need to initiate gTests via cTest

atamazov (approved):
LGTM!

xinlipn:
> The root of the problem is that our non-discrete test requires a GPU to even run `--gtest_list_tests`. Is this correct behavior, what do you think? Maybe we should investigate the reason of it.

We probably should check and guard GPU call for gTests on non-GPU configuration to avoid this issue, I'll create a separate ticket/PR

xinlipn:
> @xinlipn @junliume @JehandadKhan
> 
> The original CMakeLists.txt has the detection of the "NO GPU case" and then
> 
> ```
> if(MIOPEN_NO_GPU)
>     set(SKIP_ALL_EXCEPT_TESTS test_include_inliner test_kernel_build_params
>             test_test_errors test_type_name test_tensor_test test_sqlite_perfdb test_sequences
>             test_pooling3d test_perfdb)
> endif()
> ```
> 
> Most likely the root reason of the issue is that MIOPEN_NO_GPU related stuff was not ported to gtests.
> 
> * (A) If we do not need non-discrete gtest to work without GPU, then we may
>   
>   * skip discovering, like in this PR
>   * engage `DISCOVERY_MODE PRE_TEST` in `gtest_discover_tests`, so the gtest won't be run during the build step.

Hi @atamazov , just to complete the discussion. non-discrete gtest does need GPU to run. So (A) applies. In the meantime, I'm not sure if `DISCOVERY_MODE PRE_TEST` is needed since gtest isn't running during build stage as logs below. I'm keeping this changes here for now

https://github.com/ROCm/MIOpen/blob/c0e672c11d2f2bde0af14ad3194e05b9ffa9ead5/test/gtest/CMakeLists.txt#L33

1.With GPU and DISCOVERY_MODE set to PRE_TEST
[gpu_with_PRE_Test.log](https://github.com/ROCm/MIOpen/files/15132890/gpu_with_PRE_Test.log)

2.With GPU and don't set DISCOVERY_MODE
[gpu_without_PRE_Test.log](https://github.com/ROCm/MIOpen/files/15132892/gpu_without_PRE_Test.log)

3.Without GPU and DISCOVERY_MODE set to PRE_TEST
[nogpu_with_PRE_Test.log](https://github.com/ROCm/MIOpen/files/15132898/nogpu_with_PRE_Test.log)

4.Without GPU and don't set DISCOVERY_MODE
[nogpu_without_PRE_Test.log](https://github.com/ROCm/MIOpen/files/15132907/nogpu_without_PRE_Test.log)

junliume (approved):

PR #2912, opened 2024-04-22
Enable alpha beta in conv reference kernel

bghimireamd:
This PR enables alpha and beta in naïve reference GPU kernel. 
`out = val*alpha + out*beta`

This PR will be merged after https://github.com/ROCm/MIOpen/pull/2817


src/include/miopen/conv/data_invoke_params.hpp
```diff
@@ -38,15 +38,15 @@ struct DataInvokeParams : InvokeParams
     Data_t workSpace;
     std::size_t workSpaceSize;
     bool gfx90aFp16alt;
-    const void* alpha;
-    const void* beta;
+    ConstData_t alpha;
```
amberhassaan: `Data_t` and `ConstData_t` in our codebase refer to buffer types. We shouldn't use that for scalars. I made a comment in #2817 where I want wrapper classes for `Scalar` parameters that contain data type info as well.

src/solver/conv_direct_naive_conv_bwd.cpp
```diff
@@ -163,6 +165,14 @@ ConvSolution ConvDirectNaiveConvBwd::GetSolution(const ExecutionContext& ctx,
                 auto out_strides =
                     conv_internal::MakeStrideArray<5>(conv_internal::SplitStrideCtoGC(
                         group, tensors.outDesc.GetStrides(), G_stride_idx));
+                float alpha_val = 1.0;
```
amberhassaan: We cannot assume `float` as a data type for `alpha/beta`. See: https://docs.nvidia.com/deeplearning/cudnn/latest/developer/misc.html#scaling-parameters

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -119,6 +119,8 @@ inline __device__ __host__ int8_t cast_to(const int32_t& val)
 template <bool ASSUME_PACKED, typename src_data_t, typename acc_data_t, typename dst_data_t>
 inline __device__ void naive_conv_fwd_nchw(const src_data_t* __restrict__ p_in,
                                            const src_data_t* __restrict__ p_wei,
+                                           const float alpha,
```
amberhassaan: The type of `alpha/beta` for naive kernel should be `accumulator type` as that's `double` for floating point inputs and `int32_t` for integer inputs sucha as `int8_t`.

test/gtest/nonpack_conv3d_fwd.hpp
```diff
@@ -44,46 +44,58 @@ struct NonPackTestCase : Conv3DTestCase
...
+                1,
+                576,
+                192,
+                64,
+                98304,
+                1,
+                32768,
+                2048,
+                128,
+                1.0, // alpha
```
amberhassaan: let's not add `alpha/beta` here. There's a better way using `Combine`. See: https://github.com/ROCm/MIOpen/blob/develop/test/gtest/fused_conv_bias_res_add_activ.cpp#L229-L235

amberhassaan (dismissed):
I want to get rid of passing around `alpha/beta` as `const void*` where it's not necessary. The only place where it's necessary is where we interface with public API. Our testing code can specify the types better and for internal code, we use wrapper classes.

src/kernels/gpu_reference_kernel/naive_conv.cpp
```diff
@@ -234,21 +236,27 @@ inline __device__ void naive_conv_fwd_nchw(const src_data_t* __restrict__ p_in,
         {
             size_t o_idx = static_cast<size_t>(iho) * wo + static_cast<size_t>(iwo);
 
-            p_out[o_idx] = cast_to<acc_data_t, dst_data_t>(value);
+            acc_data_t val_alpha_beta =
```
amberhassaan: This is good but I don't think we are quite complying with the spec. I was wrong to suggest that we can use `acc_data_t` for `alpha/beta`. In reality, it should be `float` or `double`. I think the more correct solution would be to use another template parameter for `alpha/beta`.

junliume:
@bghimireamd @amberhassaan @dpsm @powderluv @rasmith 
This is a late feature, please help expedite your review, we plan to merge soon. Thanks!

amberhassaan:
> @bghimireamd @amberhassaan @dpsm @powderluv @rasmith This is a late feature, please help expedite your review, we plan to merge soon. Thanks!

@junliume : @bghimireamd will put out a combined PR soon that contains tests etc. Let's all review that one.

bghimireamd:
@amberhassaan combined PR is ready for review https://github.com/ROCm/MIOpen/pull/2863

junliume:
**Notice to the reviewers:** replaced with consolidated https://github.com/ROCm/MIOpen/pull/2863

bghimireamd:
should review https://github.com/ROCm/MIOpen/pull/2863 instead.

PR #2913, opened 2024-04-22
MHA find graph solutions, Plan execution and forward graph API test

### LINKED ISSUE #2914: [MHA] Implement MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR read only attributes
`MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR` holds a list of engines suitable to solve a given operation graph. It is represented by `OperationGraph` class introduced in #2894.

Engine is an equivalent of our `Solution` and represented by `Engine` class that also introduced in #2894; it has a constructor taking a `Solution` to be created from.

`OperationGraph` has a constructor that is planned to:
- use #2855 to check whether the graph matches one of the known graphs
- if it does retrieve input tensors and other needed parameters and create a `Problem`
- find a list of `Solutions`
- populate the list of `Engines`

amberhassaan:
This PR contains:
1. Infrastructure to create executable engine and support `ExecutionPlan::execute()`
2. MHA forward test with the graph API (C++ API). 
3. MHA Forward test with C API

closes #2914


amberhassaan:
@Vsevolod1983 : the PR compiles. You can merge this with your PR when you are ready to run your test. I'll be debugging my test in the mean time.

src/include/miopen/graphapi/execution_plan.hpp
```diff
@@ -62,10 +61,17 @@ class ExecutionPlan
     const EngineCfg& getEngineCfg() const noexcept { return mEngineCfg; }
     EngineCfg& getEngineCfg() noexcept { return mEngineCfg; }
     const std::vector<int64_t>& getIntermediateIds() const noexcept { return mIntermediateIds; }
-    int64_t getWorkspaceSize() const { return mWorkspaceSize; }
     std::string getJsonRepresentation() const;
 
-    void execute(const VariantPack& variantPack);
+    void execute(const VariantPack& variantPack)
```
scerzh: `miopenBackendExecute()` takes a `handle` in its arguments:
https://github.com/ROCm/MIOpen/blob/d6caccd3ec5d0ba08a3aa5f9e6df3e5780b71461/include/miopen/miopen.h#L6847

I wonder if we should add a `handle` to the signature here. I don't know exactly why CuDNN has several places to pass a `handle`. Might be because when an `execution_plan` is serialized and then restored from JSON it looses the `handle` value. For now we can have its value either from opgraph or from setAttribute call.

We can wait until we implement serialization to figure out if we need it here.

src/include/miopen/graphapi/engineheur.hpp
```diff
@@ -84,16 +84,22 @@ class BackendEngineHeurDescriptor : public BackendDescriptor
...
+            : Base(engineCfg, &mOwnedEngineDescriptorInstance),
+              mOwnedEngineDescriptorInstance(Base::getEngineCfg().getEngine(), opGraphDescriptor)
+        {
+        }
+
+        OwnedEngineCfgDescriptor(const OwnedEngineCfgDescriptor& other)     = default;
+        OwnedEngineCfgDescriptor(OwnedEngineCfgDescriptor&& other) noexcept = default;
+        ;
+        OwnedEngineCfgDescriptor& operator=(const OwnedEngineCfgDescriptor& other) = default;
+        OwnedEngineCfgDescriptor& operator=(OwnedEngineCfgDescriptor&& other) noexcept = default;
```
scerzh, line 102: Default versions don't handle `Base::mEngineDescriptor` correctly. Could you double check this decision, please?

scerzh (commented):
Some initial comments

src/graphapi/find_engine.cpp
```diff
@@ -0,0 +1,336 @@
...
+    {
+        if(p->matches(graph))
+        {
+            MIOPEN_LOG_I("Matched against pattern: " << p->name());
+            return p->getEngines(graph);
+        }
+    }
+
+    MIOPEN_LOG_E("No patterns matched against the graph");
+    MIOPEN_THROW("Solution not found");
```
scerzh: Throwing here leads to `miopenBackendFinalize()` returning miopenStatusUnknownError, but usercode expects miopenStatusSuccess in this case. To detect no engines `miopenGetAttribute()` with `MIOPEN_ATTR_OPERATIONGRAPH_ENGINE_GLOBAL_COUNT = 0` is used.

I don't think this is a blocking issue, but should be discussed.

src/graphapi/find_engine.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+        return graph_gen->graph();
+    }
+
+    std::shared_ptr<TensorInfoMap> extractFind20Tensors(const OpGraph& graph,
+                                                        float* attn_scale) const
+    {
+
+        assert(attn_scale);
+        assert(&graph);
```
CAHEK7: That is an UB, a reference must exist.
Compiler just deletes this assertion: https://godbolt.org/z/3onG4hMT8

src/include/miopen/errors.hpp
```diff
@@ -66,6 +66,7 @@ template <class... Params>
 #define MIOPEN_THROW(...)                                     \
     do                                                        \
     {                                                         \
+        std::abort(); /* TODO: Remove  */                     \
```
scerzh: Don't forget to remove

src/convolution_api.cpp
```diff
@@ -197,7 +197,7 @@ extern "C" miopenStatus_t miopenGetConvolutionFindMode(const miopenConvolutionDe
 }
 
 extern "C" miopenStatus_t
-miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t alpha_beta_case,
+miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(miopenAlphaBetaCase_t alpha_beta_case,
```
atamazov, line 200: [Q] Why this is necessary?

src/graphapi/find_engine.cpp
CAHEK7: I'm talking about
`extractFind20Tensors(const OpGraph& graph...`
and the following check
`assert(&graph);`
@amberhassaan if you are going to push any other changes, please remove this assertion too.
Or don't forget to remove it in a subsequent PR.

UPD: you have to fix a merge conflict and restart CI.
It's better to remove the assertion in this PR.

src/include/miopen/errors.hpp
amberhassaan: done

src/graphapi/find_engine.cpp
amberhassaan: ok, working on both now.

src/graphapi/engine.cpp
```diff
@@ -32,52 +32,111 @@ namespace miopen {
...
+
+size_t GraphExecutorFind20::getWorkspaceSize() const
 {
-    mOpGraph = checkPtr(opGraph);
-    return *this;
+    size_t workspace_size = 0;
+    auto s                = miopenGetSolutionWorkspaceSize(mSolution, &workspace_size);
+    MIOPEN_THROW_IF(s != miopenStatusSuccess, "get solution workspace size failed");
+
+    return workspace_size;
```
atamazov: [R] Exceptions are not free from the performance POV even if not thrown.
```suggestion
    return miopen::deref(mSolution).GetWorkspaceSize();
```
This should be enough, I guess.

CAHEK7 (approved):

src/graphapi/engine.cpp
```diff
@@ -32,52 +32,111 @@ namespace miopen {
...
+void GraphExecutorFind20::execute(miopenHandle_t handle, const VariantPack& vpk)
 {
-    if(globalIndex >= 0)
+
+    std::vector<miopenTensorArgument_t> tens_args;
+
+    auto num = vpk.getTensorIds().size();
+    assert(num == vpk.getDataPtrs().size());
+
+    // TODO(amber) : verify that variant pack has all the expected input and output
```
atamazov, line 50: [R] I would use the following format:
```C++
/// \todo <...stuff...> --amberhassaan <date>
```

src/graphapi/engine.cpp
```diff
@@ -32,52 +32,111 @@ namespace miopen {
...
+    // tensors
+    for(auto i = 0ull; i < num; ++i)
```
atamazov: [R] You do not need to derive the type of variable from `auto... = ...ull` ;)
```suggestion
    for(std::size_t i = 0; i < num; ++i)
```

src/graphapi/engineheur.cpp
```diff
@@ -143,15 +137,13 @@ void BackendEngineHeurDescriptor::finalize()
 
     mEngineHeur = mBuilder.build();
 
-    auto& engineCfgs = mEngineHeur.getResults();
+    const auto& engineCfgs = mEngineHeur.getResults();
```
atamazov, line 140: 👍

src/graphapi/find_engine.cpp
```diff
@@ -0,0 +1,341 @@
...
+    {
+        if(p->matches(graph))
+        {
+            MIOPEN_LOG_I("Matched against pattern: " << p->name());
+            return p->getEngines(graph);
+        }
+    }
+
+    MIOPEN_LOG_E("No patterns matched against the graph");
+    MIOPEN_THROW("Solution not found");
```
atamazov: [R] Output of MIOPEN_LOG_E won't be visible anyway (on most Linux systems at least)
```suggestion
    MIOPEN_THROW("No patterns matched against the graph. Solution not found");
```

src/graphapi/opgraph.cpp
```diff
@@ -292,26 +279,31 @@ bool isIsomorphic(const OpGraph& left, const OpGraph& right)
 {
     if(left.numNodes() != right.numNodes())
     {
+        MIOPEN_LOG_I("test failed due to num nodes being different");
```
atamazov, line 282: [R] (in many places of this PR) I would use `MIOPEN_LOG_I2`, it is the one which is intended to log detailed info.

src/graphapi/opgraph.cpp
```diff
@@ -349,32 +341,36 @@ void BackendOperationGraphDescriptor::setAttribute(miopenBackendAttributeName_t
...
-                                  descriptors.push_back(apiDescriptor);
-                                  nodes.push_back(backendDescriptor.getOperation());
-                              }
-                              else
-                              {
-                                  MIOPEN_THROW(miopenStatusBadParm);
-                              }
-                          });
+            std::for_each_n(static_cast<miopenBackendDescriptor_t*>(arrayOfElements),
+                            elementCount,
```
atamazov: - This partially reverts #2973
```suggestion
            // for_each_n is not available on RHEL/SLES, see issue #2973
            std::for_each(static_cast<miopenBackendDescriptor_t*>(arrayOfElements),
                          static_cast<miopenBackendDescriptor_t*>(arrayOfElements) + elementCount,
```

src/graphapi/opgraph.cpp
```diff
@@ -349,32 +341,36 @@ void BackendOperationGraphDescriptor::setAttribute(miopenBackendAttributeName_t
...
+                                    MIOPEN_THROW(miopenStatusBadParm, "descriptor not finalized");
+                                }
+                            });
 
             if(!internal::noRepetitions(nodes))
             {
-                MIOPEN_THROW(miopenStatusBadParm);
+                for(size_t i = 0; i < nodes.size(); ++i)
+                {
+                    std::cout << "node: " << nodes[i] << ", desc: " << descriptors[i] << std::endl;
```
atamazov: [R] Use `std::cerr`. Maybe put this under `#ifndef NDEBUG`. Check if it works.

atamazov (changes requested):
Partial review. One fix is mandatory , others can wait (recommendations - [R]).

src/convolution_api.cpp
amberhassaan: `const` here adds almost no value so I removed it.

src/graphapi/opgraph.cpp
amberhassaan: Forgot to remove these prints. Will remove.

src/convolution_api.cpp
amberhassaan: Also wanted to match the `declaration` in the header file because `clang-tidy` or `cppcheck` was complaining about it.

src/include/miopen/graphapi/engineheur.hpp
```diff
@@ -84,16 +84,22 @@ class BackendEngineHeurDescriptor : public BackendDescriptor
...
+                                 miopenBackendDescriptor_t opGraphDescriptor)
+            : Base(engineCfg, &mOwnedEngineDescriptorInstance),
+              mOwnedEngineDescriptorInstance(Base::getEngineCfg().getEngine(), opGraphDescriptor)
+        {
+        }
+
+        OwnedEngineCfgDescriptor(const OwnedEngineCfgDescriptor& other)     = default;
+        OwnedEngineCfgDescriptor(OwnedEngineCfgDescriptor&& other) noexcept = default;
+        ;
+        OwnedEngineCfgDescriptor& operator=(const OwnedEngineCfgDescriptor& other) = default;
```
scerzh, line 101:
```suggestion
        OwnedEngineCfgDescriptor(const OwnedEngineCfgDescriptor& other)     = delete;
        OwnedEngineCfgDescriptor(OwnedEngineCfgDescriptor&& other) noexcept = default;
        ;
        OwnedEngineCfgDescriptor& operator=(const OwnedEngineCfgDescriptor& other) = delete;
```
This introduces a bug: the value of `BackendEngineCfgDescriptor::mEngineDescriptor` will be invalid after copying, so I suggest to `delete` copying methods.

scerzh (approved):

src/convolution_api.cpp
atamazov: @amberhassaan Okay. Well, `const` only affects the function body, disabling modification of the parameter there. It is useless in the declaration, and that is why static analysis tools often complain if `const` used in both declaration and definition.

[R] Return `const` to the definition, just to make diff size smaller. Up to you.

src/include/miopen/graphapi/engineheur.hpp
atamazov: @scerzh @amberhassaan 
> the value of `BackendEngineCfgDescriptor::mEngineDescriptor` will be invalid after copying

Maybe I do not understand the whole picture, but shall we also disable copy ctor and operator= in `BackendEngineCfgDescriptor` then?

src/include/miopen/graphapi/opgraph.hpp
```diff
@@ -100,6 +100,14 @@ class OpNode
     friend class OpGraph;
 
 protected:
+    static Edge makeEdge(OpNode* n, Tensor* t) { return Edge{n, t}; }
+    static Edge makeEdge(const OpNode* n, const Tensor* t)
+    {
+        return Edge{
+            const_cast<OpNode*>(n), // NOLINT (cppcoreguidelines-pro-type-const-cast)
+            const_cast<Tensor*>(t)  // NOLINT (cppcoreguidelines-pro-type-const-cast)
```
atamazov, line 108: [Q] I think you have a plan to get rid of this stuff in the future, right?

src/include/miopen/graphapi/util.hpp
```diff
@@ -0,0 +1,266 @@
...
+#include <unordered_map>
+
+namespace miopen {
+namespace graphapi {
+
+inline std::string tensorIdAsStr(const Tensor* tens_ptr)
+{
+
+    int64_t id = tens_ptr->getId();
+    char* b    = reinterpret_cast<char*>(&id);
```
atamazov, line 43: [R] This won't work on big-endian machines.

src/include/miopen/graphapi/util.hpp
```diff
@@ -0,0 +1,266 @@
...
+        that.mFn = emptyFn;
+    }
+
+    HeapPtrDeleter& operator=(HeapPtrDeleter&& that) noexcept
+    {
+        this->mFn(); // destruct self.
+        this->mFn = std::move(that.mFn);
+        that.mFn  = emptyFn;
+        return *this;
+    }
```
atamazov, line 104: [R] Check for self-assignment, consider using copy-and-swap idiom?

src/include/miopen/graphapi/util.hpp
```diff
@@ -0,0 +1,266 @@
...
+    Make(const std::vector<DummyNodeGenSpec>& node_specs)
+    {
+        return std::unique_ptr<PatternGraphGenerator>(new PatternGraphGenerator(node_specs));
+    }
+
+    const auto& graph() const { return mGraph; }
+};
+
+// TODO(Amber): move this function out so that other find 2.0 code can use it
+inline std::string tensorEnumIdToStr(miopenTensorArgumentId_t id)
```
atamazov, line 219: [R] return `std::string_view`.

atamazov (approved):
🟢 LGTM provided that the recommendations are resolved (discussed to consensus) in the future

src/include/miopen/graphapi/opgraph.hpp
amberhassaan: No, I'm afraid these are necessary.

src/include/miopen/graphapi/engineheur.hpp
scerzh: @atamazov No, `BackendEngineCfgDescriptor` is copyable because the value of `BackendEngineCfgDescriptor::mEngineDescriptor` is provided by a user (who created the Engine descriptor).

`OwnedEngineCfgDescriptor` contains the Engine descriptor (as a member) and that pointer should always point to it.

Awkward solution, but not many lines of code.

scerzh: Now I think move c-tor and move `operator=` also don't work
```suggestion
        OwnedEngineCfgDescriptor(const OwnedEngineCfgDescriptor& other)     = delete;
        OwnedEngineCfgDescriptor(OwnedEngineCfgDescriptor&& other) noexcept = delete;
        ;
        OwnedEngineCfgDescriptor& operator=(const OwnedEngineCfgDescriptor& other) = delete;
```
or may be better to undo this chunk and the corresponding one in cpp-file.

src/include/miopen/graphapi/util.hpp
amberhassaan: yeah, `move-and-swap` idiom since copy is disabled :).

src/include/miopen/graphapi/engine.hpp
```diff
@@ -27,16 +27,98 @@
...
+        return std::unique_ptr<GraphPatternExecutor>(p);
+    }
+};
+
 class Engine
 {
 private:
-    Solution mSolution;
+    std::shared_ptr<GraphPatternExecutor> mExecutor;
+    OpGraph* mGraph;
```
scerzh:
```suggestion
    OpGraph* mGraph      = nullptr;
```
Fix uninitialized member though it is not highlighted by tidy or cppcheck

src/include/miopen/graphapi/util.hpp
CAHEK7: It seems it doesn't matter, since `uint64_t` is just a container for 8 chars. In a typical flow it is filled from `std::string_view` and here we just construct that string back. But in a common case it can be broken even for the little endian. `c_str()` and c-style string operations will be broken with zero less significant byte and both C strings and C++ `streams` will suffer from special characters.

I think it should be either bulletproof or should not be a part of the library (not in `include/miopen`).
We can use it as it is in the test, but if it's a part of the library and can be misused - it will be misused.

junliume (approved):

atamazov:
@amberhassaan @CAHEK7 
> LGTM provided that the recommendations are resolved (discussed to consensus) in the future

Does it worth creating a ticket dedicated to leftovers (like my reviews)? Just to remember

src/include/miopen/graphapi/util.hpp
atamazov: @CAHEK7 +1

amberhassaan:
> @amberhassaan @CAHEK7
> 
> > LGTM provided that the recommendations are resolved (discussed to consensus) in the future
> 
> Does it worth creating a ticket dedicated to leftovers (like my reviews)? Just to remember

https://github.com/ROCm/MIOpen/pull/3010 started to address follow up work.

PR #2917, opened 2024-04-24
Basic mechanism for tuning in ExecutePrimitive

DrizztDoUrden:
This allows to add tuning support for more primitives. As originally batchnorm was requested, I have also included a perfdb definition for that one.

For such primitives tuning behaviour can be only controlled via `MIOPEN_FIND_ENFORCE` as the original API has no parameters controlling the behaviour in contrast to convolution/find 2.0.

To use it in a specific primitive a function like
```C++
auto GetDb(const miopen::ExecutionContext& ctx, const miopen::some_primitive::ProblemDescriptionTag&)
    -> miopen::PerformanceDb
```
is needed, where `ProblemDescriptionTag` is some empy type from which `some_primitive::ProblemDescription` is derived and is just marking the primitive for the db paths. And any solver derived from `TunableSolverTrait` including `TunableSolverMixin`-derivatives would use tuning when requested. If GetDb was not implemented, but SolverContainer has such a solver, a compiler error would be produced.

The following `using` is an easy way to skip duplicating a long name on a per-solver basis
```C++
template <class PerformanceConfig>
using SomePrimitiveTunableSolver =
    TuningSolverMixin<ExecutionContext, miopen::some_primitive::ProblemDescription, PerformanceConfig>;
```

Side-bonus: this doesn't instantiate db until the first applicable tunable solver is encountered with enforce allowing it, so no time wasted when it is not required.

Test for the generic mechanism is included.


src/include/miopen/find_solution.hpp
```diff
@@ -169,10 +175,18 @@ ConvSolution FindSolution(Solver s,
                           const std::optional<FindOptions>& options = std::nullopt)
 {
     static_assert(sizeof(Solver) == sizeof(SolverBase), "Solver must be stateless");
-    static_assert(std::is_base_of<SolverBase, Solver>{}, "Not derived class of SolverBase");
+    static_assert(std::is_base_of<SolverBase, Solver>{} && !std::is_same_v<SolverBase, Solver>,
+                  "Not derived class of SolverBase");
```
averinevg: `SolverBase` is an abstract class so this check is redundant

src/include/miopen/solver.hpp
```diff
@@ -196,82 +199,90 @@ struct ConvTunableSolverBase : SolverMixin<ExecutionContext, miopen::conv::Probl
...
+
+namespace conv {
+
+/// Typedef for convolution non-tunable solvers
+using ConvSolver = NonTunableSolverBase<ExecutionContext, miopen::conv::ProblemDescription>;
+
+/// Typedef for convolution tunable solvers
+template <class PerformanceConfig>
+using ConvTunableSolver =
+    TunableSolverMixin<ExecutionContext, miopen::conv::ProblemDescription, PerformanceConfig>;
```
averinevg, line 284: It would be nice to make changes for fused solvers as well. `struct FusionTunableSolverBase` can be removed and `struct FusionTunableSolver` can be replaced with alias.

src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -178,6 +180,9 @@ struct ProblemDescription : ProblemDescriptionBase
 
     NetworkConfig MakeNetworkConfig() const override;
 
+    friend auto GetDb(const miopen::ExecutionContext& ctx,
+                      const miopen::batchnorm::ProblemDescription&) -> miopen::PerformanceDb;
```
averinevg: Do we need this declaration?

test/gtest/execute_primitive.cpp
```diff
@@ -0,0 +1,235 @@
...
+    ASSERT_FALSE(results.get_default_perf_cfg_invoked);
+    ASSERT_TRUE(results.search_invoked);
+    ASSERT_EQ(results.perf_cfg_value, TestPerfConfig::searched_value);
+}
+
+int main(int argc, char** argv)
+{
+    testing::InitGoogleTest(&argc, argv);
+    return RUN_ALL_TESTS();
+}
```
JehandadKhan: This will break the single binary for gtest. Please remove so that GTest would use default `main`.

src/include/miopen/find_solution.hpp
```diff
@@ -163,16 +171,24 @@ template <class Solver, class Context, class Problem, class Db>
...
 {
     static_assert(sizeof(Solver) == sizeof(SolverBase), "Solver must be stateless");
     static_assert(std::is_base_of<SolverBase, Solver>{}, "Not derived class of SolverBase");
+
+    decltype(auto) db_getter = [&]() -> decltype(auto) {
+        if constexpr(std::is_invocable_v<Db>)
+            return db;
+        else
+            return [&]() -> std::decay_t<Db>& { return db; };
+    }();
```
JehandadKhan, line 187: @DrizztDoUrden  Can you please help me understand what is going on here and why ?

JehandadKhan (dismissed):

src/include/miopen/find_solution.hpp
DrizztDoUrden: Yes, sure! This makes a compile-time check of where `Db` is a type with an `operator()` with no parameters. If so, it returns the value in db. Otherwise it returns a lamda that, in turns, returns reference to the value of db. This does not have a check that Db is an actual data base type, but that is implied. I have considered making a pseudo-concept to do the check, but I do not think it is feasible here: the build would fail anyway and code would not be trivial for a proper check as dbs do not have unified API. I can make it if needed to make this lamda architecturally correct.

test/gtest/execute_primitive.cpp
DrizztDoUrden: Sure, would fix on Tuesday, as that is the next work day in Serbia. I can make it ASAP if needed.

src/include/miopen/batchnorm/problem_description.hpp
DrizztDoUrden: Yes, that is the way to detect primitive related perf-dbs in `ExecutePrimitive` via ADL. I can change it if you have another idea of how to make it plug'n'play, I genuinely do not.
With ADL approach any primitive with `ExecutePrimitive` only needs some `GetDb` declaration discoverable via ADL and at least one tunable solver to make tuning/perfdb work out of the box via a well tested path.

src/include/miopen/solver.hpp
DrizztDoUrden: I have missed that, great catch! Would fix with another comments.

JehandadKhan:
It can surely wait!

src/include/miopen/batchnorm/problem_description.hpp
averinevg: It would be nice to note this in the comment to this function

averinevg (commented):
> Problem description here is not supposed to be used inside in any way and is just marking the primitive for the db paths

Let's comment on this to make it clearer

cderb (approved):
lgtm

averinevg (approved):
LGTM

junliume (approved):

PR #2919, opened 2024-04-24
[MHA] Implement several graph API descriptors

### LINKED ISSUE #2920: Implementation of MIOPEN_BACKEND_ENGINE_DESCRIPTOR

### LINKED ISSUE #2921: Implementation of MIOPEN_BACKEND_ENGINECFG_DESCRIPTOR

### LINKED ISSUE #2922: Implementation of MIOPEN_BACKEND_EXECUTION_PLAN_DESCRIPTOR

scerzh:
The list of the descriptors:
- MIOPEN_BACKEND_ENGINE_DESCRIPTOR
- MIOPEN_BACKEND_ENGINECFG_DESCRIPTOR
- MIOPEN_BACKEND_EXECUTION_PLAN_DESCRIPTOR

closes #2920 
closes #2921 
closes #2922


junliume (approved):

JehandadKhan (approved):

src/include/miopen/graphapi/engine.hpp
```diff
@@ -49,8 +51,52 @@ class Engine
...
+class EngineBuilder
+{
+private:
+    const OpGraph* mOpGraph = nullptr;
+    int64_t mGlobalIndex    = -1;
+    int32_t mSmCount        = 0;
+    bool mGlobalIndexSet    = false;
+
+public:
+    EngineBuilder& setOpGraph(const OpGraph* opGraph);
```
amberhassaan, line 72: engine needs to keep a graph pointer. I can add it in my latest PR.

src/include/miopen/graphapi/enginecfg.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+class EngineCfg
+{
+private:
+    /* we don't use a pointer here to allow a user
+     * to have several configs for an Engine. Each
+     * config might modify its Engine in future so
+     * their instance of Engine shouldn't be shared
+     */
+    Engine mEngine;
```
amberhassaan, line 43: I'm not sure I agree with this. I think this should be a pointer. Backend API can set engine with a pointer, but also call `getAttribute` on the same engine, so we should be returning the same engine pointer. 
Further, a user may modify a `knob` via the `EngineConfig` but read it via the `Engine` (bizarre API choice IMO), but to support it means we keep a pointer.

src/include/miopen/graphapi/execution_plan.hpp
```diff
@@ -0,0 +1,138 @@
...
+namespace graphapi {
+
+class ExecutionPlan
+{
+private:
+    /* we don't use a pointer for mEngineCfg
+     * because we need to support serialization
+     * and deserialization
+     */
+    EngineCfg mEngineCfg;
```
amberhassaan, line 47: same here. Should be a `pointer`.

amberhassaan (commented):
@scerzh : some important concerns to address here. Please prioritize fixing these as my branch depends on `Engine, EngineConfig and ExecutionPlan`.

src/include/miopen/graphapi/execution_plan.hpp
scerzh: A user expects to be able to load an execution plan from JSON and run it without building the graph from the beginning; this along with the pointer will cause difficulties in memory management.

src/include/miopen/graphapi/engine.hpp
scerzh: If your PR depends on it, add it in your PR. Also I can add it in my next [GTEST] PR.

src/include/miopen/graphapi/enginecfg.hpp
scerzh: I thought the instance of Engine provided by a user should not be modified by setting a knob and reading the same knob from the Engine should return the default value; and EngineCfg represents an engine's config without touching the source engine instance.

But the behavior is not described in detail in the docs so you may be right.

scerzh: This engine descriptor and its way of 'creating' it by index seems to purpose the case when a user wants to go through available engines and try different knob values to find a suitable engine/its config. Then the user can save found combination by serialization and in future reuse it without creating all the needed descriptors and providing only variant pack with uids/data pointers (even no need of tensor descriptors).

If a user wants the lib to find the best engine/its config, there is engineheir descriptor.

scerzh: engineheir descriptor returns list of enginecfg descriptors skipping 'engine' step.

src/include/miopen/graphapi/execution_plan.hpp
scerzh: This place requires some managed pointer, shared_ptr, I think. Should we then switch to `shared_ptr`s throughout the existing graph API code?

PR #2929, opened 2024-04-25
[MHA] add test for the backward pass

CAHEK7:
Add numeric test for multihead attention backward pass.

Skip unsupported cases earlier for forward and backward passes (do not generate data and compute reference tensors).


test/gtest/gpu_mha_backward.cpp
```diff
@@ -0,0 +1,441 @@
...
+    else
+    {
+        return {};
+    }
+}
+
+inline std::vector<TestCase> GetFullTestCases()
+{
+    if((miopen::IsUnset(ENV(MIOPEN_TEST_ALL)) || miopen::IsEnabled(ENV(MIOPEN_TEST_ALL))) &&
+       CheckFloatArg("--float"))
```
DrizztDoUrden:
```suggestion
    if(miopen::IsDisabled(ENV(MIOPEN_TEST_ALL)))
        return {};

    if(CheckFloatArg("--float"))
```

[Suggestion] This way it would be harder to accidentally add an else that ignores `MIOPEN_TEST_ALL`

DrizztDoUrden (approved):
LGTM with one postponable suggestion

junliume (approved):

PR #2936, opened 2024-04-30
[Workaround] Disable conv_hip_implicit_gemm_grouped_bwd_xdlops

msaudulhassan:


src/solver/conv_hip_implicit_gemm_grouped_bwd_xdlops.cpp
```diff
@@ -460,7 +460,7 @@ bool ConvHipImplicitGemmGroupBwdXdlops::IsApplicable(
     [[maybe_unused]] const ProblemDescription& problem) const
 {
 #if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
-    if(miopen::IsDisabled(ENV(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_GROUP_BWD_XDLOPS)))
+    if(miopen::IsEnabled(ENV(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_GROUP_BWD_XDLOPS)))
```
junliume, line 463: @msaudulhassan interesting way to disable a solver ;-P I like it that we can test the fix later by enabling that env var explicitly.

msaudulhassan: @junliume PS: I'll set this PR to Open once I have tested it on the detectron2 workload

junliume:
@zjing14 @JehandadKhan : I think we still need this WA until the kernel issues is fully investigated and fixed.

msaudulhassan:
> @zjing14 @JehandadKhan : I think we still need this WA until the kernel issues is fully investigated and fixed.

Yes we'd still need it even if TunaNet is fixed. Unfortunately, I have been mainly been focusing on fixing Tunanet thus far and haven't been able to get to this. The PR fixing TunaNet should be open by the day's end and then I can resume this PR

junliume (approved):

PR #2938, opened 2024-05-01
Gtest nchw nhwc

randyspauldingamd:


randyspauldingamd:
Notes:
- float8 is not supported (throws exceptions)
- any other types we should add?
- runtime is excessive (appx. 78s on fw-11/MI100). This could be greatly reduced by restructuring the parameters and using at least one loop inside the RunTest() functions.

randyspauldingamd:
`to_miopen_data_type` and its specializations are duplicated. I tried moving them into their own file, but the #include structure is very incomplete and this created errors in many other modules. It would be a nightmare.

test/gtest/gpu_nchw_nhwc_transpose.cpp
```diff
@@ -0,0 +1,658 @@
...
+    static std::string get() { return "nhwc2nchw"; }
+};
+
+template <typename T>
+struct to_miopen_data_type
+{
+};
+
+template <>
+struct to_miopen_data_type<float>
```
amberhassaan: why do you need this? It's already there in `tensor_holder.hpp`.

test/gtest/gpu_nchw_nhwc_transpose.cpp
```diff
@@ -0,0 +1,658 @@
...
+struct to_miopen_data_type<uint8_t>
+{
+    static miopenDataType_t get() { return miopenInt8; }
+};
+
+static constexpr int RAND_INTEGER_MAX = 120;
+static constexpr int RAND_INTEGER_MIN = -88;
+
+template <typename T>
+void rand_tensor_integer(tensor<T>& t, int max = RAND_INTEGER_MAX, int min = RAND_INTEGER_MIN)
```
amberhassaan: remove and replace with `tensor.generate()` from `tensor_holder.hpp`.

src/include/miopen/tensor_layout.hpp
```diff
@@ -35,6 +35,26 @@
 
 namespace miopen {
 
+enum tensor_layout_t
```
amberhassaan, line 38: This enum is a duplicate of `miopenTensorLayout_t` in `miopen.h`, so let's remove it and use `miopenTensorLayout_t`. You can split that as a separate PR that merges before this PR since the scopes are different.

test/gtest/gpu_nchw_nhwc_transpose.cpp
```diff
@@ -0,0 +1,658 @@
...
+
+using namespace batched_transpose;
+
+template <typename T, class TRANSPOSE_SOL>
+struct LayoutTransposeTest_2D
+    : public ::testing::TestWithParam<
+          std::tuple<uint32_t, uint32_t>>
+{
+protected:
+    miopenHandle_t              handle{};
```
amberhassaan: use `miopen::Handle` directly and then you don't have to `create` it or `destroy` it.

test/gtest/gpu_nchw_nhwc_transpose.cpp
```diff
@@ -0,0 +1,658 @@
...
+        miopenDestroy(handle);
+    }
+};
+
+#define DEFINE_TransposeTest_2D(type, sol)                                                          \
+struct TransposeTest_2D_ ## sol ## _ ## type                                                        \
+    : public TransposeTest_2D<type, miopen::sol>                                                    \
+{                                                                                                   \
+protected:                                                                                          \
+    void SetUp() override { TransposeTest_2D<type, miopen::sol>::SetUp(); }                         \
```
amberhassaan: why do you need to `override` these methods if you're only calling the base class versions.

amberhassaan (commented):
@randyspauldingamd : the two cpp files under `test/gtest` seem similar. Are they duplicates? 
2. Please remove `hipMalloc/hipMemcpy` calls and use `handle.Read()` and `handle.Write()` instead. 
3. Left some other comments related to duplicate classes and enums etc.

test/gtest/gpu_nchw_nhwc_transpose.cpp
CAHEK7: It comes from the original file, but agree, it should be replaced with `struct miopen_type;` from `tensor_holder.hpp`

test/gtest/gpu_nchw_nhwc_transpose.cpp
```diff
@@ -0,0 +1,658 @@
...
+                                testing::ValuesIn(transpose_dims::get_channel_size())               \
+                            )                                                                       \
+);
+
+#define DEFINE_2D_TYPED_TESTS(sol)          \
+DEFINE_TransposeTest_2D(float, sol);        \
+DEFINE_TransposeTest_2D(float16, sol);      \
+DEFINE_TransposeTest_2D(bfloat16, sol);     \
+DEFINE_TransposeTest_2D(uint16_t, sol);     \
+DEFINE_TransposeTest_2D(uint8_t, sol);      \
```
CAHEK7: Due to our `tensor<T>` abstraction flaw we must do strange exercises like this one. But actually it must not be a compile-time parameter.
CPU implementations can be explicitly instantiated and tensor type is actually a runtime enum parameter, so it's not a big deal to do a runtime dispatch for those types and simplify everything by moving the type into one of the `testing::Combine` term.

test/gtest/gpu_nchw_nhwc_transpose.cpp
```diff
@@ -0,0 +1,658 @@
...
+    uint32_t n;
+    uint32_t c;
+
+    virtual void SetUp() override
+    {
+        miopenCreate(&handle);
+
+        auto [_n, _c] = GetParam();
+        n = _n;
+        c = _c;
```
CAHEK7: Since `n` and `c` already exist, `std::tie` looks more suitable here.

test/gtest/layout_transpose.cpp
```diff
@@ -0,0 +1,654 @@
...
+#include "random.hpp"
+
+#define ASSERT_HIP_SUCCESS(x) ASSERT_EQ((x), hipSuccess)
+
+namespace batched_transpose {
+
+template <typename T>
+void cpu_ncdhw2ndhwc(T* dst, T* src, uint64_t N, uint64_t C, uint64_t D, uint64_t H, uint64_t W)
+{
+    for(uint64_t i_n = 0; i_n < N; i_n++)
```
CAHEK7: This file looks very similar to `gpu_nchw_nhwc_transpose.cpp`.
Can we gather identical parts into one header and use it everywhere?

CAHEK7 (changes requested):
Overall it looks good, since most of the code came from the old test.
But the test itself seems too old-fashioned or badly designed with direct HIP memory management, handlers creation, etc.

Probably a legacy code should be revisited and rewritten in this PR.
Or maybe this PR is just a moving the test to gtest infra without any improvements or significant changes.
This topic should be discussed together with compile-time tensor types.

My main concern is a similar code in `gpu_nchw_nwhc_transpose` and `layout_transpose` files, probably it can be improved.

CAHEK7:
Also someone should start jenkins build manually. Sometimes it does not start automatically.

test/gtest/layout_transpose.cpp
randyspauldingamd: gpu_nchw_nhwc_transpose.cpp is just an old copy. I used git rename to change it to layout_transpose, but something went wonky and the old file stuck around temporarily.

test/gtest/gpu_nchw_nhwc_transpose.cpp
randyspauldingamd: fixed (in layout_transpose.cpp)

test/gtest/gpu_nchw_nhwc_transpose.cpp
randyspauldingamd: Good catch. I just didn't trust gtest and wanted to be certain they get called. The overrides were not needed.

test/gtest/gpu_nchw_nhwc_transpose.cpp
randyspauldingamd: done (in layout_transpose.cpp)

test/gtest/gpu_nchw_nhwc_transpose.cpp
randyspauldingamd: Very nice, yeah, I don't know the original author's motivation for those, but `miopen_type` was an easy drop-in replacement.

test/gtest/gpu_nchw_nhwc_transpose.cpp
randyspauldingamd: fixed in layout_transpose

test/gtest/gpu_nchw_nhwc_transpose.cpp
randyspauldingamd: That's beyond my knowledge, so it won't make it into this PR--but I'll mention it to the team as a possible future improvement.

src/include/miopen/tensor_layout.hpp
randyspauldingamd: Sure, it makes perfect sense to do that. However, these tests do not use `tensor_layout_t` directly, so there's no need to connect that work with this PR. Just let me know when to pull it into a sprint; it would also be a good task for @BrianHarrisonAMD.

amberhassaan: @randyspauldingamd : I am fine with creating a follow up PR and task for it but please assign it to yourself. @iq136boy will work with @BrianHarrisonAMD to figure out what he'll work on next. CC: @JehandadKhan

randyspauldingamd:
@CAHEK7 could you please finish your review and remove the block on this PR?

test/gtest/layout_transpose.cpp
```diff
@@ -0,0 +1,566 @@
...
+                    static_cast<int>(h),
+                    static_cast<int>(w)
+                };
+
+                std::vector<int> tensor_strides;
+
+                std::string layout_default = miopen::tensor_layout_get_default(4);
+                std::string layout_string  = tensor_layout_to_string(miopen::tensor_layout_nchw);
+
+                miopen::tensor_layout_to_strides(
```
amberhassaan, line 295: the body of the loop from here on is similar to the body of the loop in the `3D test`. Please factor out the common code into helper functions.

CAHEK7 (approved):

test/gtest/layout_transpose.cpp
randyspauldingamd: I considered that as well and decided the benefit does not justify the cost. It is very important that this PR gets merged before the end of the sprint. If you would like to discuss further refactoring, let's do that next sprint and do any new work under another ticket.

amberhassaan: let's do it in the next sprint. You should create a new ticket.

amberhassaan (approved):

test/gtest/layout_transpose.cpp
randyspauldingamd: I should have added that those helper functions will probably be useful in several other unit test files. Doing it the right way is far from a trivial exercise.

randyspauldingamd (commented):
All good!

randyspauldingamd:
CI was broken. Replaced by https://github.com/ROCm/MIOpen/pull/2982

PR #2953, opened 2024-05-07
[Windows] fix NOGPU backend compilation

apwojcik:
Broken compilation on Windows.
Data types `uint64_t` and `uint32_t` are unavailable on Windows (MSVC). The PR changes to `unsigned long long`, a 64-bit data type on Windows and Linux, and to `unsigned int`, a 32-bit data type on Windows and Linux. Also, the HIP SDK on Windows uses `clang` as an OpenCL compiler.
On Windows, when using `std::underlying_type_t` and `std::ostream` requires to include explicitly desired header files (`type_traits` and `ostream`).


averinevg (dismissed):
> Data types uint64_t and uint32_t are unavailable on Windows (MSVC)

@apwojcik You could look in our library sources and see that these data types are used in many places and there are no portability issues with them.
This is probably a problem with a missing include, perhaps this compiler processes them differently.
Please also note that this file is not part of any kernel.

junliume:
> `uint32_t` are unavailable on Windows

@apwojcik agree with @averinevg 's comment above, I'm not familiar with MSVC but very strange that these standard types are not supported, unless it is really an outdated version.

apwojcik:
@junliume @averinevg it is done

junliume (approved):

src/include/miopen/conv/kernel_interface/winograd_kernel_interface.hpp
```diff
@@ -26,6 +26,10 @@
 
 #pragma once
 
+#include <type_traits>
+#include <ostream>
+#include <stdint.h>
```
CAHEK7, line 31:
```suggestion
#include <cstdint>
```

CAHEK7 (approved):

src/include/miopen/conv/kernel_interface/winograd_kernel_interface.hpp
apwojcik: On Windows, `<cstdint>` puts those types in the `std` namespace. So, instead of `uint32_t`, we should use `std::uint32_t`. I wanted to minimize the changes required.

PR #2961, opened 2024-05-14
[CK] Replace ENV with MIOPEN_ENV and bump CK commit hash by updating requirements.txt

bghimireamd:
update ck commit hash to pass [bwd weight half convolution bilinear tests](https://github.com/ROCm/MIOpen/pull/2863).

Had to rename [MIOpen's #define ENV ](https://github.com/ROCm/MIOpen/blob/develop/src/include/miopen/env.hpp#L156) to [#define MIOPEN_ENV](https://github.com/ROCm/MIOpen/blob/426d0a17627e7d6fba555a27be8ffe612c25c97e/src/include/miopen/env.hpp#L156) because it was having name collision with recent introduction of CK's macro with same name [#define ENV](https://github.com/ROCm/composable_kernel/blob/aab4439f0f5e8c2c94c440aa103cecd1b74e3133/include/ck/utility/env.hpp#L127)

### DIFF
--- a/driver/CBAInferFusion_driver.hpp
+++ b/driver/CBAInferFusion_driver.hpp
@@ -601,7 +601,7 @@ int CBAInferFusionDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
     else
         out_sz = in_sz; // This is for N+A so the output is the same as the input size
 
-    if(miopen::IsEnabled(ENV(MIOPEN_DRIVER_PAD_BUFFERS_2M)))
+    if(miopen::IsEnabled(MIOPEN_ENV(MIOPEN_DRIVER_PAD_BUFFERS_2M)))
     {
         PadBufferSize(wei_sz, sizeof(Tgpu));
     }
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -145,7 +145,7 @@ private:
 
 static inline void AdjustWorkspacesizeVariableFromEnv(std::size_t& sz)
 {
-    auto adj = miopen::Value(ENV(MIOPEN_DRIVER_CONV_WORKSPACE_SIZE_ADJUST));
+    auto adj = miopen::Value(MIOPEN_ENV(MIOPEN_DRIVER_CONV_WORKSPACE_SIZE_ADJUST));
     if(adj == 0ULL)
         return; // nop
     auto sz_save = sz;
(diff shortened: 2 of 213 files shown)


junliume (approved):

driver/CBAInferFusion_driver.hpp
```diff
@@ -601,7 +601,7 @@ int CBAInferFusionDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
     else
         out_sz = in_sz; // This is for N+A so the output is the same as the input size
 
-    if(miopen::IsEnabled(ENV(MIOPEN_DRIVER_PAD_BUFFERS_2M)))
+    if(miopen::IsEnabled(MIOPEN_ENV(MIOPEN_DRIVER_PAD_BUFFERS_2M)))
```
atamazov, line 604: This must be reverted and fixed in CK

atamazov:
@bghimireamd [Informative]

> Had to rename [MIOpen's #define ENV ](https://github.com/ROCm/MIOpen/blob/develop/src/include/miopen/env.hpp#L156) to [#define MIOPEN_ENV](https://github.com/ROCm/MIOpen/blob/426d0a17627e7d6fba555a27be8ffe612c25c97e/src/include/miopen/env.hpp#L156) because it was having name collision with recent introduction of CK's macro with same name [#define ENV](https://github.com/ROCm/composable_kernel/blob/aab4439f0f5e8c2c94c440aa103cecd1b74e3133/include/ck/utility/env.hpp#L127)

Thanks for fixing the problem.

We do not export macro as a part of our public API and therefore it is OK to define whatever we want.

If CK exports its own macro (`ENV`) in its public API, then it is CK who must take care about possible collisions with their users, for example by adding and unique prefix (like `CK_` or `COMPOSABLE_KERNELS_`). Or, better, they should avoid exporting macros in their API.

I'll take care about this.

atamazov:
@junliume @JehandadKhan @illsilin FYI previous comment.

driver/CBAInferFusion_driver.hpp
amberhassaan: I agree that CK is at fault for exposing `ENV` but it doesn't hurt us to add our prefix to our macros. I'll leave it up to @bghimireamd to see if he wants to go through with changing a large set of files.

amberhassaan:
> @bghimireamd [Informative]
> 
> > Had to rename [MIOpen's #define ENV ](https://github.com/ROCm/MIOpen/blob/develop/src/include/miopen/env.hpp#L156) to [#define MIOPEN_ENV](https://github.com/ROCm/MIOpen/blob/426d0a17627e7d6fba555a27be8ffe612c25c97e/src/include/miopen/env.hpp#L156) because it was having name collision with recent introduction of CK's macro with same name [#define ENV](https://github.com/ROCm/composable_kernel/blob/aab4439f0f5e8c2c94c440aa103cecd1b74e3133/include/ck/utility/env.hpp#L127)
> 
> Thanks for fixing the problem.
> 
> We do not export macro as a part of our public API and therefore it is OK to define whatever we want.
> 
> If CK exports its own macro (`ENV`) in its public API, then it is CK who must take care about possible collisions with their users, for example by adding and unique prefix (like `CK_` or `COMPOSABLE_KERNELS_`). Or, better, they should avoid exporting macros in their API.
> 
> I'll take care about this.

I didn't realize when reviewing that it's already merged. I say we keep `MIOPEN_ENV` as that's better than `ENV` which is very generic and may cause collisions with some 3rd party library, but at the same time, we make CK fix their naming.

junliume:
> @junliume @JehandadKhan @illsilin FYI previous comment.

CK side is fixing the issue: https://github.com/ROCm/composable_kernel/pull/1296

PR #2965, opened 2024-05-15
[Windows] graphapi: fix compilation due to conflicting definitions of std::min between MSVC and Clang

apwojcik:
Once again, the `graphapi` component broke compilation on Windows. The PR prepares a better workaround for the issue HIP SDK faces on Windows.
The `std::min()` problem concerns only the `int64_t` data type, and the workaround addresses only this data type.


junliume:
@apwojcik Thanks! It would be great if our Windows CI is more reliable, so that we will not merge if it has failed :)

I will remind the team about the `minimal` workaround on certain datatypes too.

junliume (approved):

src/graphapi/engineheur.cpp
```diff
@@ -197,7 +197,7 @@ void BackendEngineHeurDescriptor::getAttribute(miopenBackendAttributeName_t attr
         {
             *elementCount = mResults.size();
             std::transform(mResults.begin(),
-                           mResults.begin() + std::min(*elementCount, requestedElementCount),
```
junliume, line 200: @apwojcik it should be fixed in Clang, right? Is the compiler team aware of this issue?

PR #2968, opened 2024-05-15
[MHA] Implement ExecutionPlan execute function

kikimych:
Binding Engine with arguments


amberhassaan:
@kikimych : this seems like a duplication of work I was doing for #2913 . We should close it as I provide an implementation of `find_engines()` and `ExecutionPlan::execute()`. 
CC: @JehandadKhan , @junliume

BradPepersAMD:

PR #2971, opened 2024-05-17
[MHA] turn on fp8 matmul for mha

CAHEK7:
Moved some common mha code into separated file and made a wrapper for rocblas gemm to run native fp8 matrix multiplication.

Current miopen gemm wrapper supports only casted operations and it is a subject for refactoring, adding one more operation type will make that wrapper overcomplicated.


src/solver/mha/mha_solver_forward.cpp
```diff
@@ -109,45 +73,59 @@ bool MhaForward::IsApplicable([[maybe_unused]] const ExecutionContext& context,
...
-           && descsForward.vDesc.IsPacked()                             //
-           && descsForward.oDesc.IsPacked()                             //
-           && descsForward.mDesc.IsPacked()                             //
-           && descsForward.zInvDesc.IsPacked()                          //
-           && MIOPEN_USE_GEMM;
+    const miopen::mha::MhaInputDescsForward& descsFwd = problem.GetDescsForward();
+
+    auto [N, H, S, D] = miopen::tien<4>(descsFwd.kDesc.GetLengths());
+
+    return MIOPEN_USE_GEMM                                                          //
```
amberhassaan: Is this a macro? Perhaps we should cast it to `bool` ?

amberhassaan (approved):
Looks good.

src/solver/mha/mha_solver_forward.cpp
CAHEK7: Probably.
It is extensively used in `#if MIOPEN_USE_GEMM`, so it must be implicitly castable to bool in all cases.

test/gtest/gpu_mha_forward.cpp
```diff
@@ -76,58 +78,55 @@ struct TestCase
...
-            {8, 7, 270, 1, 0.0f},
-            {1, 1, 1, 1, 0.0f},
-            {3, 5, 32, 7, 0.8f},
-            {2, 2, 64, 128, 0.8f},
-            {2, 1, 128, 4, 0.8f},
-            {2, 7, 256, 31, 0.8f},
-        };
-    }
-    else
+    if(!(CheckFloatArg("--float") || CheckFloatArg("--float8")))
```
CAHEK7, line 83: Since we are still using a strange way to limit the tests, right now there is no way to turn on only fp8 tests.
Moreover having a dedicated build for just fp8 tests seems strange - why we should rebuild whole library just for it?

@atamazov @JehandadKhan @xinlipn what do you think?

Vsevolod1983 (approved):

test/gtest/gpu_mha_forward.cpp
atamazov: @CAHEK7 I would recommend implementing this feature just like we've implemented it for other data types. I believe we'll have more and more float8 tests in the future. And unification helps to save the most critical resource (our brains).

😅 It's not that I think we're idiots))

junliume (approved):

PR #2972, opened 2024-05-17
[CONV] Follow up for conv bilinear

bghimireamd:
* Address comments from https://github.com/ROCm/MIOpen/pull/2863
* enable `half` datatype in backward weight


Jenkinsfile
```diff
@@ -469,7 +469,7 @@ pipeline {
             description: "")
         booleanParam(
             name: "TARGET_GFX908",
-            defaultValue: env.BRANCH_NAME == "develop" ? true : false,
+            defaultValue: true,
```
bghimireamd: ignore this. I will revert this change if needed. This stage only ran in develop branch so had to remove it temporarily to make sure tests passing in my branch.

Jenkinsfile
```diff
@@ -284,7 +284,7 @@ def buildHipClangJob(Map conf=[:]){
             }
 
             withDockerContainer(image: image, args: dockerOpts + ' -v=/var/jenkins/:/var/jenkins') {
-                timeout(time: 150, unit:'MINUTES')
+                timeout(time: 600, unit:'MINUTES')
```
bghimireamd: had to increase this time so CI does not timout.

amberhassaan: 4x is rather dramatic. Do we need that much?

src/convolution_api.cpp
```diff
@@ -198,39 +198,48 @@ extern "C" miopenStatus_t miopenGetConvolutionFindMode(const miopenConvolutionDe
 
 extern "C" miopenStatus_t
 miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t alpha_beta_case,
-                                                   miopenDataType_t data_type,
-                                                   size_t C,
-                                                   size_t K,
-                                                   size_t output_tensor_size,
+                                                   const miopenTensorDescriptor_t inputTensorDesc,
```
amberhassaan, line 201: add `convolution descriptor` as well, even though we don't need it (yet).

src/convolution_api.cpp
```diff
@@ -198,39 +198,48 @@ extern "C" miopenStatus_t miopenGetConvolutionFindMode(const miopenConvolutionDe
...
         {
-        case miopenInt32:
-        case miopenFloat:
-        case miopenHalf:
-        case miopenBFloat16:
-        case miopenInt8:
-        case miopenFloat8:
-        case miopenBFloat8: byte_size = 4; break;
-        case miopenDouble: byte_size = 8;
+            switch(data_type)
```
amberhassaan, line 225: I believe there's a function somewhere that gives size for type.

src/convolution_api.cpp
```diff
@@ -198,39 +198,48 @@ extern "C" miopenStatus_t miopenGetConvolutionFindMode(const miopenConvolutionDe
...
 
-    size_t byte_size = 0;
-    if(alpha_beta_case == BILINEAR || alpha_beta_case == SCALE ||
-       ((data_type == miopenHalf) && ((C & 1) != 0 || (K & 1) != 0 /* Test if odd*/)))
-    {
-        switch(data_type)
+        assert(in_spatial_dims == miopen::deref(outputTensorDesc).GetNumDims());
+
+        size_t C = std::get<1>(
+            miopen::GetNCDHW(in_spatial_dims, miopen::deref(inputTensorDesc).GetLengths()));
```
amberhassaan, line 216: so this works for all layouts?

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -129,6 +134,15 @@ struct HipEventProfiler
 };
 #endif
 
+inline bool ShouldAllocateWorkSpaceBufferForWRW(const miopen::conv::ProblemDescription& problem)
+{
+    return problem.GetSpatialDims() == 3 &&
```
amberhassaan: `prob.Is3d()`? 
Also, break down the checks into separate bools for better readability. May be add informative comments as well.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -148,22 +162,32 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
 
     ConvSolution result;
     result.invoker_factory = [ck_args     = CKArgsType{problem},
+                              problem_    = problem,
                               sh_conv_ptr = std::shared_ptr{std::move(*ptr_iter)}](
                                  const std::vector<Kernel>&) mutable {
-        return [ck_args = std::move(ck_args), sh_conv_ptr = std::move(sh_conv_ptr)](
-                   const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+        return [ck_args     = std::move(ck_args),
+                problem_    = std::move(problem_),
```
amberhassaan: do not capture a large struct like `problem` by value. Rather let's pick the pieces we need. You can compute them outside the lambda and then capture them by value.

test/gtest/group_conv.hpp
```diff
@@ -548,3 +560,25 @@ std::vector<miopenTensorLayout_t> GetLayoutValues()
     DEFINE_GROUP_CONV_TEST(2, alpha, beta, type, dir, ab_case)
 #define DEFINE_GROUP_CONV3D_TEST(type, dir, alpha, beta, ab_case) \
     DEFINE_GROUP_CONV_TEST(3, alpha, beta, type, dir, ab_case)
+
+//// 2D
+// Bilinear where alpha = 2.2 and beta 3.3
+#define DEFINE_GROUP_CONV2D_BILINEAR_TEST(type, dir) \
```
amberhassaan: Don't like these new macros split by case. Define some values for `alpha` and `beta` (2 each) separately on lines 555 and 556 and don't define any new macros.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -632,8 +656,16 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
     auto ck_args = CKArgsType{problem};
 
     auto conv_ptrs = DeviceOpType::GetInstances();
-    auto ptr_iter  = FindConvPtrByID(conv_ptrs, kernel_id);
 
+    std::shared_ptr<CKBWDWeightBufferDescriptor> _ck_buff_des;
+
+    if(problem.GetSpatialDims() == 3 && problem.IsDirectionBackwardWrW())
+    {
+        _ck_buff_des =
```
amberhassaan: why does a simple struct like this need to be a `shared_ptr`? Let's see if we can avoid it. You can for example use `std::optional<CK...>` and that does not allocate anything on the heap (just an object + a bool).

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -491,7 +516,7 @@ auto MakeTaggedTransposeInstances(ConvSolution& result,
                                   const Input1TposeOp& input1_op,
                                   const Input2TposeOp& input2_op,
                                   const OutputTposeOp& output_op,
-                                  CKBWDWeightBufferDescriptor* ck_buff_des)
+                                  std::optional<CKBWDWeightBufferDescriptor>& ck_buff_des)
```
amberhassaan, line 529: should be const ref?

src/convolution_api.cpp
bghimireamd: yes since we save lens in NCHW format.

test/gtest/group_conv.hpp
```diff
@@ -43,6 +43,13 @@ struct GroupConvTestConfig
 {
 };
 
+// Define specific combinations of alpha beta case.
+static std::vector<std::tuple<float, float>> combinations = {
```
amberhassaan: This needs a better name, but let me push a fix the way I wanted it.

amberhassaan (approved):
I'm gonna push a change. Rest seems OK.

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: ck_buff_des is being modified to get ck_offset size.

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: `prob.Is3d()` since ck is currently only supporting 3d kernel.

src/convolution_api.cpp
bghimireamd: I couldn't find the function whose return byte_size based on size of accumulator.

src/conv/problem_description.cpp
```diff
@@ -187,6 +187,19 @@ void ProblemDescription::MakeNetworkConfig(std::string& conf_key) const
...
+    // was retrived. So, had to add alpha value to
+    // network config to avoid picking wrongly cached kernel.
+    auto roundUpBilinearValues = [](double val) -> std::string {
+        std::ostringstream stream;
+        stream << std::fixed << std::setprecision(1) << val;
+        return stream.str();
+    };
+    ss << 'x' << roundUpBilinearValues(GetAlpha().GetAsDouble());
+    ss << 'x' << roundUpBilinearValues(GetBeta().GetAsDouble());
+
```
bghimireamd: @amberhassaan  @DrizztDoUrden  test were failing since wrong cached kernel were used, so had to do this.

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: prefer calling `Is3d()` to `X == 3`.

src/conv/problem_description.cpp
amberhassaan: No, this is not acceptable. We need to find out why the wrong kernel is being picked up. One possible common cause would be that we are caching certain variables in the Invoker and those cached values end up being wrong when the problem changes. We need to find out if that's happening and what those values are. Put prints inside the Invoker for failing tests.

bghimireamd: what if problem is same but the alpha and beta values are different?

Jenkinsfile
junliume: I just saw this :) I observed some time outs on gfx908, and I think 210 mins will be sufficient.
10 hours is indeed a bit dramatic

junliume (approved):

src/convolution_api.cpp
amberhassaan: I was thinking of `GetTypeSize()` in `src/include/.../tensor.hpp`. It doesn't quite do what you're doing here because it returns the actual size in bytes and not 4 bytes for types < 4. 

You could take `GetTypeSize()` and round it to the next multiple of 4, but that IMO is too clever and less clear.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -146,33 +164,74 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
                     SetTensor(handle, tensors.dwDesc, tensors.dw, &zero);
-
-                    if(data_ctx.workSpace)
-                    {
-                        sh_conv_ptr->SetWorkSpacePointer(argument_ptr.get(), data_ctx.workSpace);
-                    }
+                }
+                // use captured value, other wise getting warning
+                // "lambda capture is not used" since this variable is only used in assert.
+                (void)should_allocated_wrw_buffer;
```
amberhassaan, line 203: just fyi: our style in the library seems to be `std::ignore = should_...` elsewhere. It's up to you if you want to use that.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -490,7 +549,7 @@ auto MakeTaggedTransposeInstances(ConvSolution& result,
                                        output_solver.GetKernelInfo(),
                                        output_init_solver.GetKernelInfo()});
 
-    if(ck_buff_des)
+    if(ck_buff_des.has_value())
```
amberhassaan, line 552: fyi: `if (ck_buf_des)` also works the same way.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -588,7 +648,7 @@ inline size_t GetWorkspaceSizeLayoutTransformConv(const miopen::conv::ProblemDes
 {
     if(problem.IsLayoutNHWC())
     {
-        if(problem.GetDirection() == ::miopen::conv::Direction::BackwardWeights)
+        if(problem.Is3d() && problem.GetDirection() == ::miopen::conv::Direction::BackwardWeights)
```
amberhassaan, line 651: `problem.IsDirectionBackwardWrW()`.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -597,7 +657,7 @@ inline size_t GetWorkspaceSizeLayoutTransformConv(const miopen::conv::ProblemDes
 
     assert(problem.IsLayoutDefault());
 
-    if(problem.GetDirection() == ::miopen::conv::Direction::BackwardWeights)
+    if(problem.Is3d() && problem.GetDirection() == ::miopen::conv::Direction::BackwardWeights)
```
amberhassaan, line 660: same as above.

src/solver/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
```diff
@@ -292,15 +302,18 @@ struct CKArgs
...
     {
-        return MakeArgPtr(conv_ptr, tensors.out, tensors.w, tensors.in);
+        return MakeArgPtr(conv_ptr, tensors.out, tensors.w, tensors.in, alpha, beta);
     }
 
     template <typename ConvPtr>
     bool IsSupportedBy(const ConvPtr& conv_ptr) const
     {
-        auto arg_ptr = MakeArgPtr(conv_ptr, nullptr, nullptr, nullptr);
+        auto arg_ptr = MakeArgPtr(conv_ptr, nullptr, nullptr, nullptr, 1.0f, 0.0f);
```
amberhassaan, line 316: suggestion: perhaps we should define constants for `CONV_DEFAULT_ALPHA` and `CONV_DEFAULT_BETA` as you are using `1.0` and `0.0` in other places like the constructor for `ProblemDescription` and `*InvokeParams`.

amberhassaan (approved):
Minor comments mostly but PR is good to go.

junliume:
@bghimireamd @amberhassaan we will watch CI for this one, and the remaining review issues are non-blocking and maybe dealt with in follow up enhacements.

PR #2976, opened 2024-05-20
[HotFix] add two missing cases in #2869

junliume:
With the fix in #2869, it still failed for the following configs in Tensorflow unit tests:
```
./bin/MIOpenDriver conv -n 1 -c 1 -H 1 -W 5 -k 1 -y 3 -x 6 -p 2 -q 1 -u 2 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 1

./bin/MIOpenDriver conv -n 1 -c 1 -H 1 -W 1 -k 1 -y 2 -x 3 -p 1 -q 2 -u 1 -v 2 -l 1 -j 1 -m conv -g 1 -F 4 -t 1
```

### DIFF
--- a/src/solver/conv_asm_implicit_gemm_gtc_wrw_nhwc.cpp
+++ b/src/solver/conv_asm_implicit_gemm_gtc_wrw_nhwc.cpp
@@ -877,8 +877,12 @@ bool ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC::IsApplicable(
         const auto pad_h    = problem.GetPadH();
         const auto pad_w    = problem.GetPadW();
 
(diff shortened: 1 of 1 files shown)


junliume:
@carlushuang  and @aska-0096 : do you have more general ways of checking applicability?

carlushuang:
Hi @junliume I think this is fine, as long as we just WA this issue in the solver, so explicitly check the size is OK

junliume:
@carlushuang and @aska-0096 : could you check if the change indeed worked around the given two cases? If so could you approve this PR? Thanks!

JehandadKhan (approved):

PR #2982, opened 2024-05-21
Rs/layout transpose

randyspauldingamd:
Replacing PR #2938 because it's not working from the fork.

https://ontrack-internal.amd.com/browse/LWPMIOPEN-792


CAHEK7 (approved):

amberhassaan (approved):

src/include/miopen/tensor_layout.hpp
```diff
@@ -35,6 +35,26 @@
 
 namespace miopen {
 
+enum tensor_layout_t
+{
+    tensor_layout_nchw,
+    tensor_layout_ncdhw,
+    tensor_layout_nhwc,
+    tensor_layout_ndhwc,
+};
```
averinevg: Why don't we use the existing enum for layouts?

amberhassaan: I raised this point. @randyspauldingamd will fix this as a follow up PR.

averinevg: Let's create a ticket for this so we don't forget

averinevg: Also let's not add things to the library that are only used in the tests

randyspauldingamd: Good point @averinevg . I moved it here because I couldn't find a suitable existing file under `test`. I'm moving `tensor_layout_to_string` into a new file, gtest/tensor_layout_util.hpp. PR will be up this evening. `tensor_layout_t` no longer exists, ofc. Any alternate suggestions before I get too far into that work?

averinevg: My point was that you are trying to invent something that already exists. In addition to being a waste of time, it could lead to confusion. Since this will be fixed, I have no complaints. But it’s still better to prevent the presence of such “unfinished” pull requests in the `develop`, especially when the problem is known and the solution is trivial.

randyspauldingamd: Oh @averinevg, you missed the history of this. `tensor_layout_t` is not my code, I just refactored it to remove existing duplicates in the tests. The duplication of `miopenTensorLayout_t` was not identified before it was too late in the sprint to justify holding up this PR. So yes, the proper approach is to handle this immediately in a followup PR.

averinevg (changes requested):
I suggest combining https://github.com/ROCm/MIOpen/pull/2982 and https://github.com/ROCm/MIOpen/pull/2988 into single pull request

randyspauldingamd:
Merged 2988 into this PR

src/tensor.cpp
```diff
@@ -392,8 +392,16 @@ std::string TensorDescriptor::GetLayout_str() const
     case miopenTensorCHWNc8: return "CHWNc";
     case miopenTensorNCDHW: return "NCDHW";
     case miopenTensorNDHWC: return "NDHWC";
+    default: MIOPEN_THROW(miopenStatusInternalError, "Unknown tensor layout");
```
averinevg, line 396:
```suggestion
```

src/tensor.cpp
```diff
@@ -392,8 +392,16 @@ std::string TensorDescriptor::GetLayout_str() const
...
     }
-    MIOPEN_THROW(miopenStatusInternalError, "Unknown tensor layout");
+
+    return "";
```
averinevg:
```suggestion
    MIOPEN_THROW(miopenStatusInternalError, "Unknown tensor layout");
```

src/tensor.cpp
```diff
@@ -392,8 +392,16 @@ std::string TensorDescriptor::GetLayout_str() const
...
+}
+
+std::string TensorDescriptor::GetLayout_str() const
+{
+    auto str = GetLayoutStr(this->tensorLayout);
+    return str;
```
averinevg:
```suggestion
    return GetLayoutStr(this->tensorLayout);
```

src/include/miopen/tensor.hpp
```diff
@@ -193,6 +193,7 @@ struct MIOPEN_EXPORT TensorDescriptor : miopenTensorDescriptor
 
     miopenDataType_t GetType() const;
     miopenTensorLayout_t GetLayout_t() const;
+    static std::string GetLayoutStr(miopenTensorLayout_t layout);
```
averinevg, line 197: :+1:

averinevg (changes requested):

src/tensor.cpp
randyspauldingamd: This is intentional, it informs future devs that processing should _not_ continue after the switch. The return "" later isn't necessary, but is left to catch future potential bugs.

src/tensor.cpp
randyspauldingamd: This is intentional, it informs future devs that processing should _not_ continue after the switch. The return "" later isn't necessary, but is left to catch future potential bugs.

src/tensor.cpp
randyspauldingamd: The existing code is better for debugging.

src/tensor.cpp
averinevg: What kinds of bugs can occur in this place?

src/tensor.cpp
averinevg: What kinds of bugs can occur in this place?

src/tensor.cpp
averinevg: Let's _Keep it Short and Simple_. This variable is not needed here, and what can be debugged in this function?

src/tensor.cpp
randyspauldingamd: I don't understand what you're asking here? The MIOPEN_THROW needs to stay. Do you want me to remove the return?

randyspauldingamd: done

averinevg (approved):

randyspauldingamd:
Ready to be merged!

junliume (approved):

PR #2984, opened 2024-05-21
[Testing] Convert MIOpenDriver tests to gTest framework

BrianHarrisonAMD:
Purpose of PR:

- Convert existing MIOpenDriver tests to utilize gtest framework.

How this was tested:

- Ran the converted tests on a gfx90a device with different environment variables set to confirm it's running tests as expected.
- Running CI pipeline to confirm tests are running as expected in different conditions.
- Confirmed failing tests are caught by malforming arguments to cause failure messages. 

Notes:

- Associated previous [PR](https://github.com/ROCm/MIOpen/pull/2878).


BrianHarrisonAMD:
Just realized that the environment variables used in the tests used to be variables set during the CMake step, and there isn't anything actually setting the environment variables during the testing stage. I think we will need to add quite a few environment variables to Jenkins to make sure they are set correctly during the run.

src/include/miopen/process.hpp
```diff
@@ -40,15 +41,22 @@ struct Process
...
 };
 
 struct ProcessAsync
 {
-    ProcessAsync(const fs::path& cmd, std::string_view args = "", const fs::path& cwd = "");
+    ProcessAsync(const fs::path& cmd,
+                 std::string_view args                                             = "",
+                 const fs::path& cwd                                               = "",
+                 std::ostream* out                                                 = nullptr,
+                 std::map<std::string, std::string> additionalEnvironmentVariables = {});
```
randyspauldingamd: (already resolved after Teams conversation) Perhaps we should add an alias, something along the lines of:
`using ProcessEnvironmentMap = std::map<std::string, std::string>;`

test/gtest/miopendriver_gemm.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_WITH_MIOPENDRIVER)
+MIOPEN_DECLARE_ENV_VAR_STR(MIOPENDRIVER_MODE_GEMM)
+
+namespace miopendriver_gemm {
+
+std::vector<std::string> GetTestCases()
+{
+    const std::string& modeGemmnArg = miopen::GetStringEnv(MIOPEN_ENV(MIOPENDRIVER_MODE_GEMM));
```
randyspauldingamd, line 44: Is there a typo in `modeGemmnArg` ?

randyspauldingamd (approved):
I verified that all the test cases have been reproduced.

test/gtest/miopendriver_regression_issue_2047.cpp
```diff
@@ -0,0 +1,96 @@
...
+    return ::IsTestSupportedForDevMask<d_mask, e_mask>();
+}
+
+void RunMIOpenDriver()
+{
+    bool runTestSuite = miopen::IsEnabled(MIOPEN_ENV(MIOPEN_TEST_WITH_MIOPENDRIVER)) &&
+                        IsTestSupportedForDevice() &&
+                        (miopen::GetStringEnv(MIOPEN_ENV(MIOPEN_TEST_FLOAT_ARG)) == "--float" ||
+                         miopen::GetStringEnv(MIOPEN_ENV(MIOPEN_TEST_FLOAT_ARG)) == "--half" ||
+                         miopen::GetStringEnv(MIOPEN_ENV(MIOPEN_TEST_FLOAT_ARG)) == "--bf16" ||
```
atamazov, line 75: @BrianHarrisonAMD 🔴 I am afraid this should be `--bfloat16` 🤝

BrianHarrisonAMD: You are absolutely correct, and it's wrong in miopendriver_conv2d_trans.cpp as well.
Ill get that fixed.

BrianHarrisonAMD: Added the fix to PR #3011 since this is the branch where environment variables will be set correctly from CMake.

Thanks for the heads up!

test/gtest/miopendriver_conv2d_trans.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+void RunMIOpenDriver()
+{
+    bool runTestSuite = miopen::IsEnabled(MIOPEN_ENV(MIOPEN_TEST_WITH_MIOPENDRIVER)) &&
+                        IsTestSupportedForDevice() &&
+                        miopen::IsEnabled(MIOPEN_ENV(MIOPEN_TEST_ALL)) &&
+                        (miopen::GetStringEnv(MIOPEN_ENV(MIOPEN_TEST_FLOAT_ARG)) == "--float" ||
+                         miopen::GetStringEnv(MIOPEN_ENV(MIOPEN_TEST_FLOAT_ARG)) == "--half" ||
+                         miopen::GetStringEnv(MIOPEN_ENV(MIOPEN_TEST_FLOAT_ARG)) == "--bf16");
+
```
atamazov, line 88: These tests won't be run in standalone mode, but we need this (the whole idea of switching to google test is to be able to run tests without CTest).

We should have three tests, one per each datatype. When `IsUnset(MIOPEN_ENV(MIOPEN_TEST_ALL)`, all three tests should be run (if `IsTestSupportedForDevice() == true`), see #1843.

BrianHarrisonAMD: Since these call the driver, and it relies on the environment variable, we likely need to separate these into three tests, and then set the associated conv argument with each test as part of running the driver. That way you can run them when TEST_ALL is not set, with the proper environment, and otherwise only run them selectively when they are set. 

Does that sound correct?

BrianHarrisonAMD: This should be addressed in the gTest environment variable PR #3011. 
Please take a look when you get a chance!

atamazov: @BrianHarrisonAMD Thanks for quick feedback, Brian.

> Does that sound correct?

Yes, but let's see the PR ;)

atamazov: @BrianHarrisonAMD The fix in #3011 looks good 👍

BrianHarrisonAMD: Thanks for the feedback!

PR #2985, opened 2024-05-22
fix ck instance

bghimireamd:
switch to Amd-develop branch of CK instance that supports correct instance of DeviceGroupedConvFwdMultipleABD.


junliume (approved):
@bghimireamd it would be better if we can indicate *when* this backward incompatibility starts in CK. It looks like a hotfix when we update the CK commit hash in requirements.txt

PR #2988, opened 2024-05-23
Rs/tensor layout

randyspauldingamd:
Removing a duplicated enum, `tensor_layout_t`. This was only used in Unit Tests, so the helper function `tensor_layout_to_string` was also moved into a new utility file.

NOTE: there is extra code in the original diff that is part of PR #2982 which is ready for merge. This will be updated after that PR has been merged.

NOTE: ci did not automatically launch for this PR. If anyone has any ideas why, please let me know!


averinevg:
> Removing a duplicated enum, `tensor_layout_t`. This was only used in Unit Tests, so the helper function `tensor_layout_to_string` was also moved into a new utility file.

I'm afraid this won't work. All commits of the pull request are squashed, and in the end only the diff is merged. It turns out that you cannot delete something that is not yet in the `develop` at the time of your commit. Or are you planning to do a rebase? (which should also only be used as a last resort)

averinevg (changes requested):
I suggest combining https://github.com/ROCm/MIOpen/pull/2982 and https://github.com/ROCm/MIOpen/pull/2988 into single pull request

randyspauldingamd:
> > Removing a duplicated enum, `tensor_layout_t`. This was only used in Unit Tests, so the helper function `tensor_layout_to_string` was also moved into a new utility file.
> 
> I'm afraid this won't work. All commits of the pull request are squashed, and in the end only the diff is merged. It turns out that you cannot delete something that is not yet in the `develop` at the time of your commit. Or are you planning to do a rebase? (which should also only be used as a last resort)

Hmm, my git knowledge isn't that great and I'm happy to get some guidance. I've never used rebase and didn't intend to. I created this branch from the rs/layout_transpose branch. Will that prevent the commits from being squashed? At any rate, they will still exist in my local repo, right? So I think in the worst case I could push a dummy commit to recover them.

averinevg:
@randyspauldingamd You can read about `--squash` option [here](https://git-scm.com/docs/git-merge#Documentation/git-merge.txt---squash).

test/gtest/tensor_layout_util.hpp
```diff
@@ -0,0 +1,45 @@
...
+{
+    switch(layout)
+    {
+    case miopenTensorNCHW: return "NCHW";
+    case miopenTensorNCDHW: return "NCDHW";
+    case miopenTensorNHWC: return "NHWC";
+    case miopenTensorNDHWC: return "NDHWC";
+    default: MIOPEN_THROW("Unsupported tensor layout");
+    }
+}
```
averinevg, line 43: [Note] We already have `TensorDescriptor::GetLayout_str()`. There is no static method there, it might be worth adding.

randyspauldingamd:
There isn’t much detail there, but it seems to confirm my suggestion.

amberhassaan:
> I suggest combining #2982 and #2988 into single pull request

@randyspauldingamd : I see what's going on. @averinevg  is right. You can combine the two by either merging the branch for 2988 into the branch for 2982 (in this case, close 2988) or by rebasing 2988 on top of 2982 (in this case, close 2982). First option may be easier because folks tend to trip on rebase more easily than merge.

randyspauldingamd:
Merged into 2982

PR #2991, opened 2024-05-23
Int64 type introduction and basic  support

Vsevolod1983:
We need to have Int64 type in miopenDataType_t enum in order to be able to create correct Mha Forward Graph in terms of Graph Api.
For example: dropout seed and dropout offset parameters in MHA are represented by int64.

### DIFF
--- a/src/convolution_api.cpp
+++ b/src/convolution_api.cpp
@@ -218,7 +218,8 @@ miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t a
         case miopenInt8:
         case miopenFloat8:
         case miopenBFloat8: byte_size = 4; break;
-        case miopenDouble: byte_size = 8;
+        case miopenDouble:
+        case miopenInt64: byte_size = 8; break;
         }
         *buffer_size = byte_size * output_tensor_size;
     }
--- a/src/include/miopen/tensor.hpp
+++ b/src/include/miopen/tensor.hpp
@@ -105,6 +105,7 @@ inline std::size_t GetTypeSize(miopenDataType_t d)
     case miopenFloat8:
     case miopenBFloat8: return 1;
     case miopenDouble: return 8;
+    case miopenInt64: return 8;
     }
     MIOPEN_THROW("Unknown or unsupported data type");
 }
--- a/src/include/miopen/tensor.hpp
+++ b/src/include/miopen/tensor.hpp
@@ -104,7 +104,7 @@ inline std::size_t GetTypeSize(miopenDataType_t d)
     case miopenInt8:
     case miopenFloat8:
     case miopenBFloat8: return 1;
-    case miopenDouble: return 8;
+    case miopenDouble:
     case miopenInt64: return 8;
     }
     MIOPEN_THROW("Unknown or unsupported data type");
--- a/src/solver/mlir_common.cpp
+++ b/src/solver/mlir_common.cpp
@@ -58,6 +58,7 @@ static const char* DTypeName(miopenDataType_t ty)
     case miopenInt8: return "i8";
     case miopenFloat8: return "fp8";
     case miopenBFloat8: return "bfp8";
+    case miopenInt64: return "i64";
     }
     MIOPEN_THROW(miopenStatusInternalError, "Value outside of datatype enum");
 }
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -52,6 +52,7 @@ bool IsDataTypeSupported(miopenDataType_t t)
     case miopenInt8:
     case miopenBFloat16:
     case miopenDouble: return true;
+    case miopenInt64: return false;
     }
     return false;
 }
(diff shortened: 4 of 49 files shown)


amberhassaan (approved):
Looks good. @junliume : please prioritize merging this as it's needed for MHA graph API PR #2913 to land.

amberhassaan:
@atamazov : please prioritize reviewing this PR as it's needed for #2913

src/convolution_api.cpp
```diff
@@ -218,7 +218,8 @@ miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t a
         case miopenInt8:
         case miopenFloat8:
         case miopenBFloat8: byte_size = 4; break;
-        case miopenDouble: byte_size = 8;
+        case miopenDouble:
+        case miopenInt64: byte_size = 8; break;
```
averinevg, line 222: [Not related to this PR]

@bghimireamd Please take a look at this function. Is everything fine with `miopenHalf`, `miopenBFloat16`, `miopenInt8`, `miopenFloat8` and `miopenBFloat8`? In my superficial opinion, `byte_size` should have different values for these cases.

src/include/miopen/tensor.hpp
```diff
@@ -105,6 +105,7 @@ inline std::size_t GetTypeSize(miopenDataType_t d)
     case miopenFloat8:
     case miopenBFloat8: return 1;
     case miopenDouble: return 8;
+    case miopenInt64: return 8;
```
averinevg, line 108:
```suggestion
    case miopenDouble:
    case miopenInt64: return 8;
```

src/tensor.cpp
```diff
@@ -52,6 +52,7 @@ bool IsDataTypeSupported(miopenDataType_t t)
     case miopenInt8:
     case miopenBFloat16:
     case miopenDouble: return true;
+    case miopenInt64: return false;
```
averinevg: Why is this set to false?

src/convolution_api.cpp
bghimireamd: Yes, this is correct. This byte size is for the accumulator type (specifically for CK kernel's wrw case). The accumulator types for [`miopenHalf`, `miopenBFloat16`, `miopenInt8`, `miopenFloat8` and `miopenBFloat8`] is of byt_size 4. But for `miopenDouble` and `miopenInt64` will be byte_size 8. So this code make sense.

src/solver/mlir_common.cpp
```diff
@@ -58,6 +58,7 @@ static const char* DTypeName(miopenDataType_t ty)
     case miopenInt8: return "i8";
     case miopenFloat8: return "fp8";
     case miopenBFloat8: return "bfp8";
+    case miopenInt64: return "i64";
```
CAHEK7, line 61: Is there anybody familiar with MLIR? Just to double check that `i64` actually exists.
@JehandadKhan @atamazov

CAHEK7:
`test/tensor_holder.hpp` should be updated as well.
It's not critical to do right now, but you will face some problems later, trying to use `tensor<int64_t>` type.

bghimireamd (approved):

CAHEK7:
I volunteered to fix the rest of switch-cases.

src/tensor.cpp
Vsevolod1983: Because adding support for tensor<T> class was out of scope for this PR. But probably have it as true also possible

Vsevolod1983:
> `test/tensor_holder.hpp` should be updated as well. It's not critical to do right now, but you will face some problems later, trying to use `tensor<int64_t>` type.

We discussed that in this PR we want only new value introduced in the datatype enum. Correct me If I'm wrong.

src/solver/mlir_common.cpp
atamazov: @CAHEK7 Even if not, there should be no problem, this change is good.

atamazov:
@CAHEK7 
> you will face some problems later, trying to use `tensor<int64_t>` type.

Only when we need this int64 tensors (which is not so yet ;)

test/gtest/tensor_api.cpp
```diff
@@ -47,7 +47,7 @@
 
 // miopenLastDataType must be changed if new data types are added
 #define miopenFirstDataType miopenHalf
-#define miopenLastDataType miopenBFloat8
+#define miopenLastDataType miopenInt64
```
atamazov, line 50: [R] https://github.com/ROCm/MIOpen/labels/value_low Introduce `miopenDataTypeLast_` in miopen.h right in this PR (and I do not see any need to guard it with BETA macro).

atamazov (approved):
🟢 LGTM! I do not see anything suspicious.

atamazov:
@amberhassaan 
> @atamazov please prioritize reviewing this PR as it's needed for #2913

Thanks for explicit @mention 👍

test/gtest/tensor_api.cpp
CAHEK7: `miopenInt64` is the last type in both cases (beta/not-beta).
Now it is even safer, since it does not require `MIOPEN_BETA` macro.

Actually I'm a bit against of `first-last` values, especially how it is used in the test.
For the common case or iteration over the enum it may be helpful, but definitely not for this test.

CAHEK7:
All the tests passed except broken MI300 node (see #2998).

junliume (approved):

PR #2995, opened 2024-05-23
RNN MS solution env update

shurale-nkn:
now it is possible to force MS solver selection, if MS is applicable. (For benchmarks)


src/ocl/rnnocl.cpp
```diff
@@ -48,15 +48,24 @@ bool RNNForwardMSIsSupported([[maybe_unused]] const RNNDescriptor& desctiptor,
...
     {
         return true;
     }
 #endif // MIOPEN_USE_GEMM&& MIOPEN_BACKEND_HIP
     return false;
 }
 
+bool RNNForwardMSIsFast(miopenDataType_t dataT, const int seqLen)
+{
+    if(miopen::IsEnabled(MIOPEN_ENV(MIOPEN_RNNFWD_exp)))
```
JehandadKhan, line 61: @junliume

junliume (approved):

PR #2891, opened 2024-04-15
Adjustments for the latest assembler (e.g. latest changes in the upstream clang)

atamazov:
- COMgr amdgcn assembly: Removed "-mno-xnack"
  - This resolves #2851
- Added WORKAROUND_ROCMCOMPILERSUPPORT_ISSUE_67 for https://github.com/ROCm/ROCm-CompilerSupport/issues/67
- Offline (clang) amdgcn assembly:
  - Removed "-mno-xnack".
  - Added features to "-mcpu" value (e.g. `"-mcpu=gfx908"` -> `"-mcpu=gfx908:xnack-"`,  see [here](https://llvm.org/docs/AMDGPUUsage.html#amdgpu-target-features) for more info)

## :cyclone: Performance/correctness testing results `develop` vs this PR

Preconditions:
- System & build configurations:
  - Navi21/36 ROCm 6.0 docker, online gcnasm (COMgr)
  - Navi21/36 ROCm 6.0 docker, offline gcnasm (/opt/rocm/llvm/bin/clang)
  - gfx906/60 ROCm 6.0 docker, online gcnasm (COMgr)
  - gfx906/60 ROCm 6.0 docker, offline gcnasm (/opt/rocm/llvm/bin/clang)
  - gfx906/60 ROCm 5.7 docker, online gcnasm (COMgr)
  - gfx906/60 ROCm 5.7 docker, offline gcnasm (/opt/rocm/llvm/bin/clang)
- Common system & build parameters: Ubuntu 20.04.5 LTS, BUILD_DEV=On, Release build, 
- 93 known FP32 convolution configs
- Only assembly kernels enabled

No performance or correctness regressions.

Verdict: :green_circle: GO!

----
[Attribution] @junliume @JehandadKhan
- https://github.com/ROCm/MIOpen/labels/enhancement
- https://github.com/ROCm/MIOpen/labels/urgency_high
- Proposed reviewers:
  - @JehandadKhan 
  - @cderb 
  - @shurale-nkn 
  - @littlewu2508

PR #2996, opened 2024-05-24
KDB updates for latest assembler

cderb:
This PR updates db_sync and KDB files to support assembler changes and changes made in #2891 
Kernel argument changes to LoadProgram function replicated in db_sync test.


junliume:
manually verified fixes of cases in https://github.com/ROCm/MIOpen/pull/2891#discussion_r1612503660

test/gtest/db_sync.cpp
```diff
@@ -563,7 +564,16 @@ void CheckDynamicFDBEntry(size_t thread_index,
...
+                    if(miopen::EndsWith(kern.kernel_file, ".s"))
+                    {
+                        compile_options +=
+                            " -mcpu=" +
+                            miopen::LcOptionTargetStrings{handle.GetTargetProperties()}.targetId;
+                    }
+                    else
+                    {
+                        compile_options += " -mcpu=" + handle.GetDeviceName();
+                    }
```
CAHEK7, line 576:
```suggestion
                    compile_options += " -mcpu=";
                    if(miopen::EndsWith(kern.kernel_file, ".s"))
                    {
                        compile_options +=
                            miopen::LcOptionTargetStrings{handle.GetTargetProperties()}.targetId;
                    }
                    else
                    {
                        compile_options += handle.GetDeviceName();
                    }
```
[R] For me it looks a bit better.

test/gtest/db_sync.cpp
```diff
@@ -734,10 +745,20 @@ void CheckFDBEntry(size_t thread_index,
...
+                            {
+                                compile_options +=
+                                    " -mcpu=" +
+                                    miopen::LcOptionTargetStrings{handle.GetTargetProperties()}
+                                        .targetId;
+                            }
+                            else
+                            {
+                                compile_options += " -mcpu=" + handle.GetDeviceName();
+                            }
```
CAHEK7, line 761:
```suggestion
                            compile_options += " -mcpu="
                            if(miopen::EndsWith(kern.kernel_file, ".s"))
                            {
                                compile_options +=
                                    miopen::LcOptionTargetStrings{handle.GetTargetProperties()}
                                        .targetId;
                            }
                            else
                            {
                                compile_options += handle.GetDeviceName();
                            }
```

CAHEK7 (approved):

junliume:
@cderb there are quite a few entries not found for gfx90a, could you help to double check? Thanks!

cderb:
> @cderb there are quite a few entries not found for gfx90a, could you help to double check? Thanks!

@junliume @JehandadKhan  this is a side effect of the targetid being added to the-mcpu argument.
The kernels were generated on a machine with '-mcpu=gfx90a:xnack-'
While the ci machine appears to be using '-mcpu=gfx90a:sramecc+:xnack-'
This of course causes a miss when looking up these arguments.

junliume:
@atamazov can we revert this change?
> targetid being added to the-mcpu argument

atamazov:
@junliume 
> @atamazov can we revert this change?
> 
> > targetid being added to the-mcpu argument

I do not think so. More info here: https://github.com/ROCm/MIOpen/pull/2309#issuecomment-1675431339.

I am investigating what needs to be done.

atamazov (approved):
🟢 LGTM

atamazov:
~However, I highly recommend getting back to using different databases for targets with different xnack and sramecc features. Otherwise we potentially losing performance.~ This is already partially implemented, KDB names include target features, see https://github.com/ROCm/MIOpen/pull/2157#discussion_r1617894860. How the kernels are compiled for KDB is another question...

See also https://github.com/ROCm/MIOpen/pull/2891#discussion_r1616343986

Note that we don't need to generate DBs for all possible combinations. We need to do this only for the most common combinations.

cderb:
> @atamazov can we revert this change?
> 
> > targetid being added to the-mcpu argument

@junliume if the kernel arguments change is reverted, then the code changes in this PR are no longer necessary. The kdb changes would also be inaccurate. If the assembler changes are causing an issue, then I can generate the kdbs once more. Otherwise we can throw out this PR.

junliume:
> @junliume
> 
> > @atamazov can we revert this change?
> > > targetid being added to the-mcpu argument
> 
> I do not think so. More info here: [#2309 (comment)](https://github.com/ROCm/MIOpen/pull/2309#issuecomment-1675431339).
> 
> I am investigating what needs to be done.

@atamazov actually I have the opposite impression, since we build kernel caches using offline compiler (not hipRTC), then the code generated without `sramecc` or `xnack` should work in a more general way: e.g. offline compiler generated kernel caches without `sramecc` or `xnack` specifics, like "gfx90a" alone, should work on gfx90a regardless of how the GPU reports these detailed features.

**update**: discussed https://github.com/ROCm/MIOpen/pull/2891#discussion_r1616343986

@atamazov  @JehandadKhan  please let me know if the above impression is mistaken. Thanks!

```
If a GPU reports XNACK feature then the code generated without xnack will likely work incorrectly.
If a GPU does not report XNACK then the code generated with xnack may work slower than it potentially can.
If a GPU reports SRAMECC feature then the code generated without sramecc will likely work incorrectly.
If a GPU does not report SRAMECC then the code generated with sramecc may work slower than it potentially can.
```

cderb:
Confirmed that dbsync will pass with no kdb changes if we revert this change:
https://github.com/ROCm/MIOpen/blob/4f5ed42bad90523c6236c3b16def127a386b2794/src/hip/handlehip.cpp#L500-L503

atamazov:
https://github.com/ROCm/MIOpen/pull/2996#issuecomment-2133976237 is updated (clarified)

atamazov:
@junliume 
> since we build kernel caches using offline compiler (not hipRTC),...

The reason is different, I guess. Offline compiler supports `xnack` target feature at least (well, there is some space for improvement wrt `sramecc`). But when Tuna builds kernels for KDBs, it uses some special hackery (`MIOPEN_DEVICE_ARCH`) to substitute actual GPU type and other target info with anything that Tuna wants. This allows to build kernels even on GPUless machines.

> then the code generated without `sramecc` or `xnack` should work in a more general way: e.g. offline compiler generated kernel caches without `sramecc` or `xnack` specifics, like "gfx90a" alone, should work on gfx90a regardless of how the GPU reports these detailed features

Yep, and this is not free from the performance POV (however I do not know the actual impact).

If we want to get the best performance for certain targets (I believe we do want this), then we need to
- identify the exact list of targets that we need to produce system KDBs
- update Tuna and related stuff to produce KDBs for these targets

<details><summary><h3>IIRC the full list of potential targets for KDB is like this:</h3></summary>

```
gfx900
gfx906:sramecc+:xnack-
gfx906:sramecc+:xnack+
-----------------------------------------------------
gfx908:sramecc+:xnack-
gfx908:sramecc+:xnack+
-----------------------------------------------------
gfx90a:sramecc-:xnack-
gfx90a:sramecc-:xnack+
gfx90a:sramecc+:xnack-
gfx90a:sramecc+:xnack+
-----------------------------------------------------
gfx940/941/942:sramecc-:xnack-
gfx940/941/942:sramecc-:xnack+
gfx940/941/942:sramecc+:xnack-
gfx940/941/942:sramecc+:xnack+
-----------------------------------------------------
gfx1030
gfx1031
gfx1100
gfx1101
gfx1102
```
</details>

/cc @junliume @cderb @JehandadKhan

atamazov:
@junliume 
> since we build kernel caches using offline compiler (not hipRTC),...

The reason is different, I guess. Offline compiler supports `xnack` target feature at least (well, there is some space for improvement wrt `sramecc`). But when Tuna builds kernels for KDBs, it uses some special hackery (`MIOPEN_DEVICE_ARCH`) to substitute actual GPU type and other target info with anything that Tuna wants. This allows to build kernels even on GPUless machines.

> then the code generated without `sramecc` or `xnack` should work in a more general way: e.g. offline compiler generated kernel caches without `sramecc` or `xnack` specifics, like "gfx90a" alone, should work on gfx90a regardless of how the GPU reports these detailed features

Well, actually, if xnack is not specified but the GPU has target feature. then compiler assumes xnack+. And this is not free from the performance POV. I do not know the actual impact for our kernels, but it seems like _current system KDB has some space for improvement_.

The same considerations apply to sramecc.

If we want to get the best performance for certain targets (I believe we do want this), then we need to
- identify the exact list of targets that we need to produce system KDBs
- update Tuna and related stuff to produce KDBs for these targets

<details><summary><h3>IIRC the full list of potential targets for KDB is like this:</h3></summary>

```
gfx900
gfx906:sramecc+:xnack-
gfx906:sramecc+:xnack+
-----------------------------------------------------
gfx908:sramecc+:xnack-
gfx908:sramecc+:xnack+
-----------------------------------------------------
gfx90a:sramecc-:xnack-
gfx90a:sramecc-:xnack+
gfx90a:sramecc+:xnack-
gfx90a:sramecc+:xnack+
-----------------------------------------------------
gfx940/941/942:sramecc-:xnack-
gfx940/941/942:sramecc-:xnack+
gfx940/941/942:sramecc+:xnack-
gfx940/941/942:sramecc+:xnack+
-----------------------------------------------------
gfx1030
gfx1031
gfx1100
gfx1101
gfx1102
```
</details>

Again, we don't need to generate KDBs for all possible combinations, only for the important ones.

/cc @junliume @cderb @JehandadKhan

atamazov:
@junliume I am going to open a ticket about "KDB optimization" soon.

cderb:
Discarding this change.

Issue #3001, opened 2024-05-28
Regenerating KDB in the `develop` branch is currently not possible.


atamazov:
Regenerating KDB with new ASM kernels in the `develop` is needed to fix #2851 (the latter may be done by reverting #3002 + regen ASM kernels altogether).

Regenerating KDB with new OCL and HIP kernels in the `develop` is needed for #3003.

atamazov:
[Attribution] @junliume
- https://github.com/ROCm/MIOpen/labels/urgency_high
- Proposed assignee: ?

atamazov:
@junliume I think this can be closed when we are ready to regenerate kernels in KDB for develop branch.

CAHEK7:
Bump priority based on a dependent PR

atamazov:
@CAHEK7 your case it a bit different. My PR needs only ASM kernels, your PR - HIP and OCL ones. I'll rename the ticket.

atamazov:
@CAHEK7 @junliume Issue description updated.

PR #3003, opened 2024-05-29
[MHA] turn on 'negative zero as nan' for FP8 numbers

CAHEK7:
Switch defaults to HW supported mode for soft-float calculations.
It does not affect GPU code, since it is already used the only available HW mode, but it affects all the host calculations like CPU verification code.

Related to #2971 and blocked by #3001
Probably related to #2996

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -39,7 +39,7 @@ endif()
 #     remain in the future)  perform final conversion (and rounding) of FP32
 #     to BF16 results. This affects the main functionality of the library.
 option( MIOPEN_USE_RNE_BFLOAT16 "Sets rounding scheme for bfloat16 type" ON )
-option( MIOPEN_FP8_IEEE_EXPONENT_BIAS "Sets the FP8 exponent bias to IEEE" ON)
+option( MIOPEN_FP8_IEEE_EXPONENT_BIAS "Sets the FP8 exponent bias to IEEE" OFF)
 option( MIOPEN_FP8_CLIPPING "Sets the FP8 clipping" ON)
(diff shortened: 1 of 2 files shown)


junliume:
@CAHEK7 @atamazov so we need to regenerate KDB caches after this change too?

***update***: `Dbsync gfx90a` test indeed is failing with this PR

CAHEK7:
After exactly this branch.
I reverted this change in my previous branch and disabled my fp8 test if that macro is 1.
It doesn't block fp8 mha, since GPU code has hw compatible implementation, but any host-based internal miopen fp8 calculations won't be correct, that's why I disabled the test.

**UPD**: the main problem is that the macro goes to all the kernels, even if they do not use it and do not support FP8 at all. It invalidates almost everything even on the targets without FP8.

junliume:
@CAHEK7 and @cderb will `gfx942`'s KDB have the same issues after this PR? We cannot merge it before confirming.

cderb:
> @CAHEK7 and @cderb will `gfx942`'s KDB have the same issues after this PR? We cannot merge it before confirming.

gfx942 dbsync is passing when I run it by hand.

junliume (approved):

atamazov:
@CAHEK7 @JehandadKhan 

> **UPD**: the main problem is that the macro goes to all the kernels, even if they do not use it and do not support FP8 at all. It invalidates almost everything even on the targets without FP8.

I think that this "selection of F8 type" shall be completely removed in the future. We shall switch to _distinct_ F8 types. That would automatically get rid of all this mess.

[Informative] Right now I foresee three F8 types:
- The F8 type which works on current HW (`bf8_fnuz`)
- Two F8 types that comply the OFP8 spec (FP8 in E4M3 format and BF8 in E5M2).

PR #3006, opened 2024-05-30
[MHA] [GTEST] Add MIOPEN_BACKEND_ENGINE_DESCRIPTOR tests

### LINKED ISSUE #2923: Tests for MIOPEN_BACKEND_ENGINE_DESCRIPTOR

scerzh:
closes #2923 

Also fixes a couple of violations of API requirements for finalize().

### DIFF
--- a/test/gtest/graphapi_engine.cpp
+++ b/test/gtest/graphapi_engine.cpp
@@ -1,0 +1,134 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


randyspauldingamd:
Ready to be merged!

test/gtest/graphapi_engine.cpp
```diff
@@ -0,0 +1,134 @@
...
+     * We have other tests (for example MHA) that cover attrsValid=true case but they
+     * cover the whole API code.
+     */
+
+    execute.descriptor.attrsValid = false;
+    execute.descriptor.textName   = "MIOPEN_BACKEND_ENGINE_DESCRIPTOR";
+    execute.descriptor.type       = MIOPEN_BACKEND_ENGINE_DESCRIPTOR;
+    execute.descriptor.attributes = {&attrOpGraph, &attrGlobalIndex, &attrSmCount};
+
+    execute();
```
randyspauldingamd, line 133: Similar to my comments on the other PR, we should add tests for replacing `attrOpGraph` and `attrGlobalIndex`.

scerzh: Yes, you are right.

`attrGlobalIndex` becomes invalid here because mock `OpGraph` returns an empty array of `Engines`. To overcome it we also should somehow mock `patterns` variable in `findEngines()` function.

`BackendOperationGraphDescriptor` class depends on concrete class `OpGraph` which depends on concrete class `MHA_Fwd_F8_Pattern` (through the call of `OpGraph::initEngines()` and `findEngines()`)

The dependency on `MHA_Fwd_F8_Pattern` should be replaced by a dependency on abstract `GraphPatternMatcher` by passing the vector of GraphPatternMatchers through this sequence above of dependent classes. There should be a point where we create a concrete class instances, seems like it is `miopenBackendCreateDescriptor()`.

This refactoring is not the point of this PR and should be planned separately.

Also we can write tests which cover the whole process instead of this particular step. We have MHA test that goes that way.

CAHEK7 (approved):

Vsevolod1983 (approved):

test/gtest/graphapi_engine.cpp
amberhassaan: Yes, currently hard to mock patterns and executors. Let me think about it as I'm not sure if it's worth allowing such a thing. We might have a baseline graph executor that executes every valid graph. This may simplify testing here.

amberhassaan (approved):

scerzh:
@junliume Can we merge this PR as it is needed by feature delivery?

randyspauldingamd (approved):
My comments were only intended to start a dialogue about future improvements. I see no reason to hold up this PR.

junliume (approved):

Issue #3005, opened 2024-05-30
MIOpen unit test link issue : ld.lld: error: undefined symbol: dladdr and undefined reference due to --no-allow-shlib-undefined


junliume:
@atamazov could you help to check if this issue relates to MIOpen change or compiler change? Thanks!

```
[2024-05-30T06:51:49.711Z] [ 75%] Linking CXX executable ../../bin/test_miopendriver_conv2d_trans
[2024-05-30T06:51:49.711Z] cd /long_pathname_so_that_rpms_can_package_the_debug_info/data/driver/MLOpen/build_hip/test/gtest && /usr/local/lib/python3.8/dist-packages/cmake/data/bin/cmake -E cmake_link_script CMakeFiles/test_miopendriver_conv2d_trans.dir/link.txt --verbose=1
[...]

PR #3007, opened 2024-05-30
[W/A] Link dl to MIOpen tests and fix GTest version in requirements

BrianHarrisonAMD:
Fix testing compilation issue where -ldl is missing when compiling gtest binaries.
I believe this is the same issue as [PR 2177](https://github.com/ROCm/MIOpen/pull/2177), but now with test binaries.

In addition, we also had to move the gtest dependency to requirements.txt. This change was necessary, due to environments that didn't install gtest linking to the wrong gtest binary which was causing build issues.

ref: #3005

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ ROCm/FunctionalPlus@v0.2.18-p0
 ROCm/eigen@3.4.0
 ROCm/frugally-deep@9683d557eb672ee2304f80f6682c51242d748a50
 ROCm/composable_kernel@e4112de7303af1601c6590b964d0df2b6a7f7d32 -DCMAKE_BUILD_TYPE=Release -DINSTANCES_ONLY=ON
+google/googletest@v1.14.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,3 @@
 ROCm/rocm-recipes@329203d79f9fe77ae5d0d742af0966bc57f4dfc8
 -f requirements.txt
 danmar/cppcheck@2.12.1
-google/googletest@v1.14.0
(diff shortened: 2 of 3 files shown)


junliume (approved):

requirements.txt
```diff
@@ -8,3 +8,4 @@ ROCm/FunctionalPlus@v0.2.18-p0
 ROCm/eigen@3.4.0
 ROCm/frugally-deep@9683d557eb672ee2304f80f6682c51242d748a50
 ROCm/composable_kernel@e4112de7303af1601c6590b964d0df2b6a7f7d32 -DCMAKE_BUILD_TYPE=Release -DINSTANCES_ONLY=ON
+google/googletest@v1.14.0
```
junliume, line 11: For the reviewers (ref: #3005)

- Does MIOpen *require* googletest to function? No, which is why this line was in `dev-requirements.txt` instead.
- Why are we putting it back to `requirements.txt`? (1) cost is relatively minimal (to build googletest); (2) it saves a lot of compatibility check issues.

amberhassaan: what compatibility checks?

BrianHarrisonAMD: This comment has a more detailed response @amberhassaan:
https://github.com/ROCm/MIOpen/issues/3005#issuecomment-2141005498

Realistically, gtest isn't needed in requirements.txt, but anyone building and running the tests for MIOpen would need to install the dev-requirements as well which wasn't a step that was known / happening in all cases. It was working before without installing dev-requirements, since it was finding the gtest installed with conda. Adding it here guarantees it'll work and link correctly with the expected version.

amberhassaan: I still don't get it. May be @junliume  or @JehandadKhan  can provide some context. Why did we separate `requirements.txt` and `dev-requirements.txt`? if the goal of `requirements.txt` is to enable building `libMIOpen.so` only, then `google-test` is not required and we shouldn't add any such dependency.

BrianHarrisonAMD: I think that's exactly right, but the way it was being ran and tested didn't include installing the dev requirements which lead to this issue. I think it's a valid choice to not add gtest, and instead have people change how they install / run the MIOpen test suite, but the issue is that the way they were doing it before just happened to work, and now it won't unless we add gtest to the requirements.txt.

PR #3008, opened 2024-05-30
Up CK version in requirements

cderb:
Move composable_kernel requirement to current tip of develop.


junliume:
@cderb we usually pick CK's more `stable` branch `amd-develop` if possible.

For example, at this moment, the TIP of CK `develop` branch does not have https://github.com/ROCm/composable_kernel/pull/1313 which supposes to fix issues from https://github.com/ROCm/composable_kernel/pull/1300 which has been included in develop, I will generally consider the current TIP of its develop branch not-stable.

What would be the reason to update CK commit hash in this case?

cderb:
@junliume fair enough. There was a need to update the ck repo in a branch for miopen to generate tunings for this ticket https://ontrack-internal.amd.com/browse/SWDEV-463985
I was told the tip of develop for ck would be the best option for doing this. I'm pushing forward with this branch, but if there is a better version of ck to use for this, i can regear it.

alexandraBara:
@cderb  is there another update here? can we resolve the conflict please so we can push?

junliume (approved):

PR #3010, opened 2024-05-31
Follow up fixes for 2913 (Graph API find solution and execute plan)

amberhassaan:
- [ ] add back `const` to function definition (https://github.com/ROCm/MIOpen/pull/2913#discussion_r1619325866)
- [ ] fix TODO format (https://github.com/ROCm/MIOpen/pull/2913#discussion_r1619106651)
- [ ] replace MIOPEN_LOG_I with MIOPEN_LOG_I2 (https://github.com/ROCm/MIOpen/pull/2913#discussion_r1619164246)
- [ ] Fix OwnedEngineConfigDesc class (https://github.com/ROCm/MIOpen/pull/2913#discussion_r1619506633)
- [ ] Improve assignment operator (https://github.com/ROCm/MIOpen/pull/2913#discussion_r1619409013)
- [ ] Use std::string_view (https://github.com/ROCm/MIOpen/pull/2913#discussion_r1619423637)


src/include/miopen/graphapi/util.hpp
```diff
@@ -90,16 +91,23 @@ struct HeapPtrDeleter
...
-        this->mFn(); // destruct self.
-        this->mFn = std::move(that.mFn);
-        that.mFn  = emptyFn;
+        if(this != &that)
+        {
+            HeapPtrDeleter tmp{std::move(that)};
+            swap(*this, tmp);
+        }
         return *this;
     }
```
CAHEK7, line 107: The problem is that the previous version is actually throw. As well as the current destructor:
```
~HeapPtrDeleter() { mFn(); }
```

> Invoking the target of an empty std::function results in `std::bad_function_call` exception being thrown.

`emptyFm` must not be used, I mean it can be, but it does not make much sense, since the check for `non-empty std::function` is required.

This particular code is fine now (with an `std::swap`), but move constructor and destructor must be fixed as well as `emptyFn` must be removed.
You may keep it if you want, but even a `const Fn` checks for the function state in the destructor and thus generates unnecessary code - it's a tine overhead, but for no reason at all.

src/include/miopen/graphapi/util.hpp
```diff
@@ -79,7 +79,8 @@ struct HeapPtrDeleter
 {
     using Fn         = std::function<void()>;
     const Fn emptyFn = []() {};
-    Fn mFn           = emptyFn;
+
+    Fn mFn = emptyFn;
```
CAHEK7: `emptyFn` should be removed, see another comment for the details.

CAHEK7 (changes requested):

CAHEK7 (approved):

junliume (approved):

PR #3013, opened 2024-06-02
Manually export symbols from MIOpen DLL

apwojcik:
We can no longer link the Debug version of MIOpen DLL on Windows with the CMake's property `WINDOWS_EXPORT_ALL_SYMBOLS`. 

DLL and EXE files can only export up to 65356 symbols. Recently, we exceeded that limit by ~10k symbols for the Debug version of MIOpen. Therefore, with this PR, we are changing auto-exporting to manual exporting of symbols to gain control over that process.

Unfortunately, we all need to pay attention and export only symbols comprising official MIOpen API and only a subset of internal symbols required for testing.

**To export the official MIOpen API, use the `MIOPEN_EXPORT` macro. To export internals required for testing, use MIOPEN_INTERNALS_EXPORT (on Windows, those are not visible in the release/deployment version of MIOpen).**


seungmanhan (commented):
I confirmed the t5layernorm, addlayernorm, layernorm, sum, reduceextreme, groupnorm,  changes.
Would I add it separately for places that do not have MIOPEN_INTERNALS_EXPORT?

dmikushin:
Hi @apwojcik , in addition to this particular problem of too many exported symbols, testing and debugging a large DLL is difficult by many means. Looking into potential further challenges, I'd rather suggest to ship a static library for development. I mean, ship DLL as you do for the customers, and additionally ship static LIB for development. Then any test executable could be linked to it, without any manual selection of exported symbols. Please make sure the static linking is done in as-needed mode, so that the binary size does not grow too much due to incorporating unused symbols.

apwojcik:
> I confirmed the t5layernorm, addlayernorm, layernorm, sum, reduceextreme, groupnorm, changes. Would I add it separately for places that do not have MIOPEN_INTERNALS_EXPORT?

Not yet. Only when you see tests failing to link because of undefined symbols. You do not have that issue on Linux, and only Windows is affected.

apwojcik:
> Hi @apwojcik , in addition to this particular problem of too many exported symbols, testing and debugging a large DLL is difficult by many means. Looking into potential further challenges, I'd rather suggest to ship a static library for development. I mean, ship DLL as you do for the customers, and additionally ship static LIB for development. Then any test executable could be linked to it, without any manual selection of exported symbols. Please make sure the static linking is done in as-needed mode, so that the binary size does not grow too much due to incorporating unused symbols.

We tried before shipping the static library for development. The result of the experiment was that each test file was about 700MB because almost the whole library was linked into the binary. After the build was complete, the whole MIOpen folder was about 80GB, so we abandoned that.

The issue is not that the Debug DLL is huge. The issue is that the PE header (the structure describing how to load and run a library/executable) has only a 16-bit field to store the number of exported symbols (simplified description). The size of the library does not change.

The original method used CMake's property EXPORT_ALL_SYMBOLS and exported everything. With this PR, we are exporting only those that are required.

The mechanism of not exporting internal symbols when tests are not built I added with earlier PR. This PR changes from automatic to manual control only.

dmikushin:
> We tried before shipping the static library for development. The result of the experiment was each test file size was about 700MB because almost the whole library was linked into the binary.

It indicates that the those tests are using the frontend API. The tests that run over selected internal functions or classes should not grow that large. So, you may only need to gather the frontend API tests into a single gtest executable to resolve this problem.

apwojcik:
> It indicates that the those tests are using the frontend API. The tests that run over selected internal functions or classes should not grow that large. So, you may only need to gather the frontend API tests into a single gtest executable to resolve this problem.

We may try that, but not right now. We are currently tight on schedule.

apwojcik:
> > It indicates that the those tests are using the frontend API. The tests that run over selected internal functions or classes should not grow that large. So, you may only need to gather the frontend API tests into a single gtest executable to resolve this problem.
> 
> We may try that, but not right now. We are currently tight on schedule.

To clarify, Linux is unaffected - all symbols are exported. The Windows is only impacted by that change.

dmikushin:
> all symbols are exported

Btw, this is not a commonly used solution either. Even the opensource shared libraries export the intended API only. No matter what it is, glibc or kernel32, all dlls export only what they intend to, not everything.

Sorry, I don't want to be annoying, I just see a rolling snowball of weird design decisions, that eventually will require even more weird design decisions :)

apwojcik:
> > All symbols are exported
> 
> Btw, this is not a commonly used solution either. Even the opensource shared libraries export the intended API only. No matter what it is, glibc or kernel32, all dlls export only what they intend to, not everything.
> 
> Sorry, I don't want to be annoying, I just see a rolling snowball of weird design decisions, that eventually will require even more weird design decisions :)

I agree. However, MIOpen was not initially designed with Windows in mind, and on Linux, all symbols are visible by default, too. We will gradually move to the proper cross-OS design. Right now, we are blocked and cannot Debug.

src/convolution_api.cpp
```diff
@@ -113,12 +114,14 @@ static inline auto MakeWrWCtxAndProblem(miopenHandle_t handle,
     return std::make_tuple(std::move(ctx), std::move(problem));
 }
 
+MIOPEN_EXPORT
```
amberhassaan, line 117: Why do the `.cpp` files need `MIOPEN_EXPORT` attribute? Doesn't it suffice to just put it on the headers (internal or external, i.e., stuff under `src/include`).

include/miopen/miopen.h
```diff
@@ -6094,7 +6094,7 @@ MIOPEN_EXPORT miopenStatus_t miopenT5LayerNormForward(miopenHandle_t handle,
  * @param sizeInBytes              Pointer to data to return the minimum workspace size
  * @return                         miopenStatus_t
  */
-extern "C" miopenStatus_t
+MIOPEN_EXPORT miopenStatus_t
```
amberhassaan, line 6097: Why did you remove `extern "C"` specifier?

amberhassaan (commented):
some questions for my understanding.

include/miopen/miopen.h
apwojcik: Because the entire file is enclosed already. 
```
#ifdef __cplusplus__
extern "C" {
#endif
...
#ifdef __cplusplus__
}
#endif
```
There is no need for explicit `extern "C"`.

src/convolution_api.cpp
apwojcik: No header file provides the prototype (declaration) of those functions.

amberhassaan: Doesn't that mean that we don't need to export these symbols? Because the only way to use such a symbol would be through a public or internal header.

apwojcik: If you see `MIOPEN_EXPORT` or `MIOPEN_INTERNALS_EXPORT` in the header or source file, that means we need to export those symbols because something (e.g., MIOpenDriver, examples, tests, etc.) is using them. All of the exports you see are relevant. Any of them missing will cause the Windows build to fail to link. There will be more in the future.

apwojcik: To be more precise, you don't need a header file declaring exported functions to use them in your code. You may put a function declaration in your code, and the linker will pick it up automatically from the list of exported functions of a certain library that has that function exported.

amberhassaan: Yeah, the pleasures of `C`, but going forward, perhaps, we should enforce such a paradigm that only stuff in headers will be `EXPORT`ed and if you need it, put it in a header. CC: @JehandadKhan , @junliume

apwojcik:
> It indicates that the those tests are using the frontend API. The tests that run over selected internal functions or classes should not grow that large. So, you may only need to gather the frontend API tests into a single gtest executable to resolve this problem.

In the experiment I mentioned, we split the API from the internals. The API was only linked to the DLL and not included in the static library. So, the API tests were interfaced to the DLL and for the internals to the static library. Still, each binary was huge.

junliume (approved):

PR #3015, opened 2024-06-03
Batchnorm forward inference CK for NCHW

xinlipn:
Enable NCHW batchnorm using CK


BradPepersAMD:

PR #3018, opened 2024-06-03
Change std::string to fs::path for handling files

apwojcik:
Change handling file paths from `std::string` to `fs::path` for better cross-OS portability.


src/include/miopen/execution_context.hpp
```diff
@@ -273,23 +269,20 @@ struct ExecutionContext
...
+        const auto filename = GetStream().GetDbBasename() +
 #if MIOPEN_ENABLE_SQLITE && MIOPEN_USE_SQLITE_PERFDB
-        filename << "_" << SQLitePerfDb::MIOPEN_PERFDB_SCHEMA_VER << ".udb";
+                              "_" + SQLitePerfDb::MIOPEN_PERFDB_SCHEMA_VER + ".udb";
 #else
-        filename << "." << GetUserDbSuffix() << ".udb.txt";
+                              "." + GetUserDbSuffix() + ".udb.txt";
 #endif
-        return (udb / filename.str()).string();
+        return udb / filename;
```
DrizztDoUrden: Prefix got lost here
Also why was ostringstream removed? It should have less allocations than concatenation chain.

DrizztDoUrden (dismissed):

src/include/miopen/execution_context.hpp
apwojcik: > Prefix got lost here

I submitted the PR too early. I am working to fix that right now (and compilation).

> Also why was ostringstream removed? It should have less allocations than concatenation chain.

It is the opposite. Please read this StackOverflow for more details: https://stackoverflow.com/questions/14741144/stdstringstream-vs-stdstring-for-concatenating-many-strings

apwojcik: The benchmark: https://quick-bench.com/q/iR8hY3MmHXCSXe5vvY28C-6mQfY
Lower means faster — `std::string` concatenation is way faster than using `std::stringstream` for the same operation.

junliume:
@DrizztDoUrden @averinevg @atamazov last call of review before merging this PR. Thanks!

apwojcik:
@junliume CI is failing because there is `No space left on device`.

cmake/embed.cmake
```diff
@@ -94,9 +94,9 @@ const std::unordered_map<std::string, std::pair<const char*,const char*>>& ${EMB
 #pragma clang diagnostic ignored \"-Wreserved-identifier\"
 #include <${EMBED_NAME}.hpp>
 ${EXTERNS}
-const std::unordered_map<std::string, std::pair<const char*,const char*>>& ${EMBED_NAME}()
+const std::unordered_map<fs::path, std::string_view>& ${EMBED_NAME}()
 {
-    static const std::unordered_map<std::string, std::pair<const char*,const char*>> result = {${INIT_KERNELS}};
+    static const std::unordered_map<fs::path, std::string_view> result = {${INIT_KERNELS}};
```
CAHEK7: Doesn't it break compatibility with old distros?
Like that one https://github.com/ROCm/MIOpen/pull/3035 ?
Also, I'd suggest to use that solution: https://github.com/ROCm/MIOpen/pull/3035#issuecomment-2156807075

src/include/miopen/simple_hash.hpp
```diff
@@ -27,15 +27,17 @@
...
 
 namespace miopen {
 struct SimpleHash
 {
-    size_t operator()(const std::pair<std::string, std::string>& p) const
+    size_t operator()(const std::pair<fs::path, std::string>& p) const
     {
-        using std::hash;
-        return (hash<std::string>()(p.first) ^ hash<std::string>()(p.second));
+        return (std::hash<std::string>()(p.first.string()) ^ std::hash<std::string>()(p.second));
```
CAHEK7:
```suggestion
        return (fs::hash_value(p.first) ^ std::hash<std::string>()(p.second));
```

CAHEK7 (changes requested):
It looks like the first comment is important, and, if CI will be restarted in any case, the second comment can be fixed too.

CAHEK7 (approved):

junliume (approved):

cmake/embed.cmake
```diff
@@ -84,19 +84,19 @@ function(generate_embed_source EMBED_NAME)
 
     file(WRITE "${PARSE_HEADER}" "
 #include <string>
-#include <unordered_map>
+#include <map>
 #include <utility>
-const std::unordered_map<std::string, std::pair<const char*,const char*>>& ${EMBED_NAME}();
```
junliume, line 89: Thanks! This will save us from some potential issues on legacy distributions :)

src/include/miopen/execution_context.hpp
atamazov: @apwojcik Thanks, this is interesting.

> The benchmark: https://quick-bench.com/q/iR8hY3MmHXCSXe5vvY28C-6mQfY

I've tried it in different configurations:
- c++17 or c++20
- Optimization: None or O3
- libstdc++(GNU) or libc++(LLVM)
- Modifications of the test:
  - Original code
  - 10 times longer strings
  - many writes to stream/string
  - ostringstream or stringstream

The benchmark shows that std::string is 14...220 times better than std::stringstream, depending on the configuration.

It's interesting that compiler optimizations negatively affect performance: the original `BM_StringConcat` reports ~640 "cpu_time" (whatever that means), but after change to O3 it shows ~910. For `BM_StringStream` things are even worse: "gpu_time" regresses from 16k to 177k.

@DrizztDoUrden Shall we open a ticket dedicated to replacing stringstreams with strings, what do you think?

src/anyramdb.cpp
```diff
@@ -52,27 +52,18 @@ static std::chrono::seconds GetLockTimeout() { return std::chrono::seconds{60};
...
-    // the calling app lifetime. Size of each is very small, and there couldn't
-    // be many of them (max number is number of _different_ GPU board installed
-    // in the user's system, which is _one_ for now). Therefore the total
-    // footprint in heap is very small. That is why we can omit deletion of
-    // these objects thus avoiding bothering with MP/MT syncronization.
-    // These will be destroyed altogether with heap.
-    auto instance = new AnyRamDb{path};
-    instances.emplace(path, instance);
-    return *instance;
+    return *instances.emplace(path, std::make_unique<AnyRamDb>(path)).first->second;
```
atamazov, line 66: @msaudulhassan just FYI.

atamazov:
@apwojcik Looks good.

atamazov:
@junliume @JehandadKhan Is is possible to launch a performance testing round on CI with d208a9a3351eb9048a4a2b3f9d34fefbae0f6fda (develop right after this commit)?

PR #3019, opened 2024-06-04
rs/transpose refactor

randyspauldingamd:
Just tidying these tests up a bit. 

Design summary:

- I looked through several tests and did not find other use cases, so changes are exclusive to layout_transpose tests.
- Since other `Solutions` expose a `Make*InvokerFactory` helper, I added it to the `BatchedTransposeSolution` public API.
- It would make a lot of sense to restructure batched transpose similarly to other Solutions by moving it into its own folder and creating files such as invoke_params.hpp, problem_description.hpp, and solvers.hpp. However, I decided these really aren't intended to be standalone solvers, and chose the "simpler is better" approach.

Deadline for this code review is 12PM PDT, 6/5/2024. Any comments received after this time will be performed under a follow-up PR next sprint.

### DIFF
--- a/src/include/miopen/batched_transpose_sol.hpp
+++ b/src/include/miopen/batched_transpose_sol.hpp
@@ -28,10 +28,26 @@
 
 #include <miopen/miopen.h>
 #include <miopen/errors.hpp>
+#include <miopen/invoke_params.hpp>
 #include <miopen/kernel_info.hpp>
 #include <miopen/op_kernel_args.hpp>
 #include <miopen/execution_context.hpp>
 #include <vector>
+
+struct transpose_invoke_param : public miopen::InvokeParams
+{
+    ConstData_t src = nullptr;
+    Data_t dst      = nullptr;
+
+    transpose_invoke_param(ConstData_t src_, Data_t dst_) : src(src_), dst(dst_) {}
+    transpose_invoke_param(miopen::InvokeType type_, ConstData_t src_, Data_t dst_)
+        : InvokeParams{type_}, src(src_), dst(dst_)
+    {
+    }
+
+    Data_t GetWorkspace() const { return nullptr; }
+    std::size_t GetWorkspaceSize() const { return 0; }
+};
 
 namespace miopen {
 
@@ -65,9 +81,43 @@ struct BatchedTransposeSolution
     int num_cu;
 
     BatchedTransposeParam kernel_param_heuristic;
+
+    InvokerFactory MakeBatchedTransposeInvokerFactory();
 };
 
-struct TransposeSolutionDefault2Nhwc : public BatchedTransposeSolution
+struct BatchedTransposeSolution2D : public BatchedTransposeSolution
+{
+    BatchedTransposeSolution2D(const ExecutionContext& ctx_,
+                               miopenDataType_t data_type_,
+                               uint32_t batch_,
+                               uint32_t channels_,
+                               uint32_t height_,
+                               uint32_t width_ = 1)
+        : BatchedTransposeSolution(ctx_, data_type_, batch_, channels_, height_ * width_)
+    {
+        MIOPEN_THROW_IF(size_t(height_ * width_) != (size_t(height_) * size_t(width_)),
+                        "integer overflow");
+    }
+};
+
+struct BatchedTransposeSolution3D : public BatchedTransposeSolution
+{
+    BatchedTransposeSolution3D(const ExecutionContext& ctx_,
+                               miopenDataType_t data_type_,
+                               uint32_t batch_,
+                               uint32_t channels_,
+                               uint32_t depth_,
+                               uint32_t height_ = 1,
+                               uint32_t width_  = 1)
+        : BatchedTransposeSolution(ctx_, data_type_, batch_, channels_, depth_ * height_ * width_)
+    {
+        MIOPEN_THROW_IF(size_t(depth_ * height_ * width_) !=
+                            (size_t(depth_) * size_t(height_) * size_t(width_)),
+                        "integer overflow");
+    }
+};
+
+struct TransposeSolutionDefault2Nhwc : public BatchedTransposeSolution2D
 {
     TransposeSolutionDefault2Nhwc(const ExecutionContext& ctx_,
                                   miopenDataType_t data_type_,
@@ -75,13 +125,13 @@ struct TransposeSolutionDefault2Nhwc : public BatchedTransposeSolution
                                   uint32_t c_,
                                   uint32_t h_,
                                   uint32_t w_)
-        : BatchedTransposeSolution(ctx_, data_type_, n_, c_, h_ * w_)
+        : BatchedTransposeSolution2D(ctx_, data_type_, n_, c_, h_, w_)
     {
         MIOPEN_THROW_IF(size_t(h_ * w_) != (size_t(h_) * size_t(w_)), "integer overflow");
     }
 };
 
-struct TransposeSolutionNhwc2Default : public BatchedTransposeSolution
+struct TransposeSolutionNhwc2Default : public BatchedTransposeSolution2D
 {
     TransposeSolutionNhwc2Default(const ExecutionContext& ctx_,
                                   miopenDataType_t data_type_,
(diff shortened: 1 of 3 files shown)


test/gtest/layout_transpose.cpp
```diff
@@ -232,20 +232,9 @@ struct transpose_dims
...
-    }
-
-    Data_t GetWorkspace() const { return nullptr; }
-    std::size_t GetWorkspaceSize() const { return 0; }
-};
+template <typename T>
+static auto gen_value =
+    [](auto... is) { return static_cast<T>(prng::gen_A_to_B(RAND_INTEGER_MIN, RAND_INTEGER_MAX)); };
 
 } // namespace batched_transpose
```
CAHEK7: I would like to rename this namespace into nameless and get rid of static. It will hide the whole block to be visible from the outside.

Nameless namespace does exactly the same what `static` does in the cpp file - it makes everything private to the object file.

CAHEK7:
May I also ask you to duplicate the content from internal jira to the PR since it's an important information for the reviewers.

It's an opensource project and there are some external collaborators. (and even if I have access to the internal jira, it's quite cumbersome to use it outside of amd infrastructure)

amberhassaan:
> Just tidying these tests up a bit. Please read the notes in the Issue prior to code reviews:
> 
> https://ontrack-internal.amd.com/browse/LWPMIOPEN-807
> 
> Deadline for code reviews is 12PM PDT, 6/5/2024. Any comments received after this time will be performed under a follow-up PR next sprint.

Avoid internal links on public github. Also, please post the message above regarding review deadline in the `daily` channel.

src/include/miopen/batched_transpose_sol.hpp
```diff
@@ -28,11 +28,27 @@
...
 #include <miopen/op_kernel_args.hpp>
 #include <miopen/execution_context.hpp>
 #include <vector>
 
+struct transpose_invoke_param : public miopen::InvokeParams
+{
+    ConstData_t src = nullptr;
+    Data_t dst      = nullptr;
+
+    transpose_invoke_param(ConstData_t src_, Data_t dst_) : src(src_), dst(dst_) {}
```
amberhassaan: This constructor doesn't call the base constructor explicitly. Are we calling the `no-arg` constructor here? Better to make it explicit.

test/gtest/layout_transpose.cpp
```diff
@@ -232,20 +232,9 @@ struct transpose_dims
...
-        : InvokeParams{type_}, src(src_), dst(dst_)
-    {
-    }
-
-    Data_t GetWorkspace() const { return nullptr; }
-    std::size_t GetWorkspaceSize() const { return 0; }
-};
+template <typename T>
+static auto gen_value =
+    [](auto... is) { return static_cast<T>(prng::gen_A_to_B(RAND_INTEGER_MIN, RAND_INTEGER_MAX)); };
```
amberhassaan, line 220: convert to a function please.

amberhassaan: something like: 
```
template <typename T, typename... Args>
T genValue(Args... args) {
   return something;
}
```

test/gtest/layout_transpose.cpp
```diff
@@ -363,6 +331,21 @@ struct LayoutTransposeTest_3D : public ::testing::TestWithParam<std::tuple<uint3
...
+    miopen::Invoker PrepareLayoutTransposeInvoker(const miopen::ExecutionContext& ctx_,
+                                                  miopenDataType_t data_type_,
+                                                  uint32_t n_,
+                                                  uint32_t c_,
+                                                  uint32_t d_,
+                                                  uint32_t h_,
+                                                  uint32_t w_)
+    {
+        TRANSPOSE_SOL transpose_sol(ctx_, data_type_, n_, c_, d_, h_, w_);
+        auto invoker_factory = transpose_sol.MakeBatchedTransposeInvokerFactory();
```
amberhassaan, line 324: convert these lines to a method (e.g., `PrepareInvoker()`) on the class `BatchedTransposeSol`.

test/gtest/layout_transpose.cpp
```diff
@@ -260,6 +249,20 @@ struct LayoutTransposeTest_2D : public ::testing::TestWithParam<std::tuple<uint3
...
     uint32_t c;
 
+    miopen::Invoker PrepareLayoutTransposeInvoker(const miopen::ExecutionContext& ctx_,
+                                                  miopenDataType_t data_type_,
+                                                  uint32_t n_,
+                                                  uint32_t c_,
+                                                  uint32_t h_,
+                                                  uint32_t w_)
+    {
+        TRANSPOSE_SOL transpose_sol(ctx_, data_type_, n_, c_, h_, w_);
```
amberhassaan, line 240: same here. This should be a method on the `BatchedTransposeSolution` class.

amberhassaan (commented):
some comments.

test/gtest/layout_transpose.cpp
scerzh: We can keep `genValue` as an object of class with template `operator()` if we need to prevent `genValue` from participating in ADL. @randyspauldingamd was it the intention?

amberhassaan: Actually, I realize that `gen_value` needs to be a functor, so let's leave it as a lambda. @scerzh , your suggestion also means the same thing but more typing, so lambda offers the lazy way out.

test/gtest/layout_transpose.cpp
randyspauldingamd: done

test/gtest/layout_transpose.cpp
randyspauldingamd: This is only used in tensor::generate, and I don't see any use cases taking a function. Exploratory work is out of scope for this PR, how about we discuss this as future work?

src/include/miopen/batched_transpose_sol.hpp
randyspauldingamd: Not my code. InvokeParams is a very basic wrapper struct defining no explicit ctors and containing only an InvokeType which defaults to Run. I'm not familiar with the language vs. compiler-specific rules behavior regarding auto-generated ctors, but it does work, so OK.

amberhassaan: You should always call some base class constructor in a derived constructor as that's a good practice and structs and classes are exactly the same in C++ except for default visibility.

test/gtest/layout_transpose.cpp
randyspauldingamd: I actually had it there initially, but I moved it because it is specific to testing and other Solvers do not provide such helpers in their public API.

In the past, I have seen such test helpers placed into a friend class. I don't have strong feelings any way, but this does seem the simplest, and better from a consistency standpoint. Of course, the point is code improvement, so consistency is not a big motivator. It does add a couple orders of magnitude to the scope of the suggestion though.

test/gtest/layout_transpose.cpp
randyspauldingamd: See discussion on the other one.

randyspauldingamd:
> > Just tidying these tests up a bit. Please read the notes in the Issue prior to code reviews:
> > https://ontrack-internal.amd.com/browse/LWPMIOPEN-807
> > Deadline for code reviews is 12PM PDT, 6/5/2024. Any comments received after this time will be performed under a follow-up PR next sprint.
> 
> Avoid internal links on public github. Also, please post the message above regarding review deadline in the `daily` channel.

Thanks for the reminder--yeah, that was silly :)

src/include/miopen/batched_transpose_sol.hpp
CAHEK7: It is called by compiler implicitly - the default one.

CAHEK7 (approved):

test/gtest/layout_transpose.cpp
amberhassaan: I think anyone using the `BatchedTransposeSolution` class would want this method and it would avoid repetition as seen in your own test case.

scerzh (approved):

junliume (approved):

test/gtest/layout_transpose.cpp
amberhassaan: @randyspauldingamd , do you want me to do this fix as a follow up PR or would you rather do it yourself?

randyspauldingamd: ~~`Sounds good, I'll move it later today. I'll wait to push until any other comments have came in.`~~
Oops, I wrote that before I noticed it had already been merged. I'll sneak it into the next PR.

PR #3025, opened 2024-06-05
[MHA] enable fp8 for backward mha pass

CAHEK7:


src/solver/mha/mha_solver_backward.cpp
```diff
@@ -49,23 +47,26 @@ namespace mha {
...
     // the second MatMuls (N*H*S*S)[T] * (N*H*S*D) = (N*H*S*D)
     // dOxO row reduction (N*H*S*1)
-    return MultiBufferWorkspaceTraits{NHS * S * get_data_size(miopenFloat), // first matmuls
-                                      NHS * S * get_data_size(miopenFloat), // first matmuls
-                                      NHS * std::max(S, D) *
-                                          get_data_size(miopenFloat), // reduction and second matmul
-                                      NHS * D * get_data_size(miopenFloat),  // second matmuls
-                                      NHS * D * get_data_size(miopenFloat)}; // second matmuls
+    return MultiBufferWorkspaceTraits{
+        NHS * S * get_data_size(miopenFloat),                             // fp32 QxK and fp32 S
```
amberhassaan, line 57: off topic: I realized that `get_data_size()` is a duplicate of `GetTypeSize()` in tensor.hpp. We should delete the former in favor of the latter.

src/solver/mha/mha_solver_backward.cpp
```diff
@@ -85,52 +86,75 @@ bool MhaBackward::IsApplicable([[maybe_unused]] const ExecutionContext& context,
...
+           && descsBwd.zInvDesc.GetType() == miopenFloat              //
+           && descsBwd.kDesc.GetType() == descsBwd.qDesc.GetType()    //
+           && descsBwd.kDesc.GetType() == descsBwd.vDesc.GetType()    //
+           && descsBwd.kDesc.GetType() == descsBwd.oDesc.GetType()    //
+           && descsBwd.kDesc.GetType() == descsBwd.dqDesc.GetType()   //
+           && descsBwd.kDesc.GetType() == descsBwd.dkDesc.GetType()   //
+           && descsBwd.kDesc.GetType() == descsBwd.dvDesc.GetType()   //
+           && ((descsBwd.kDesc.GetType() == miopenFloat)              //
+               || (USE_ROCBLAS_EX3                                    //
+                   && (MIOPEN_FP8_IEEE_EXPONENT_BIAS == 0)            //
```
amberhassaan, line 117: what is the significance of this check?

test/gtest/gpu_mha_backward.cpp
```diff
@@ -78,63 +80,62 @@ struct TestCase
...
 }
 } // namespace
 
+template <typename T>
 class Test_Bwd_Mha : public testing::TestWithParam<TestCase>
 {
+    static_assert(std::is_same_v<T, float> || std::is_same_v<T, float8>);
+
 protected:
+    using dO_T = std::conditional_t<std::is_same_v<T, float>, float, bfloat8>;
```
amberhassaan, line 137: so `dO_T` can only be `bfloat8` and not `float8`?

test/gtest/gpu_mha_backward.cpp
```diff
@@ -178,12 +179,13 @@ class Test_Bwd_Mha : public testing::TestWithParam<TestCase>
             tensors[id] = std::move(tmp);
         };
 
-        auto GenScaledTensor = [](auto... nhsd) {
-            auto val_full = tensor<float>{nhsd...};
+        auto GenScaledTensor = [](auto type, auto... nhsd) {
+            using ResultT = std::decay_t<decltype(type)>;
+            auto val_full = tensor<ResultT>{nhsd...};
             val_full.for_each([&](auto... id) {
                 // backward pass is very sensible to input data due to possible subtraction of
```
amberhassaan, line 186: s/sensible/sensitive/

test/gtest/gpu_mha_backward.cpp
```diff
@@ -178,12 +179,13 @@ class Test_Bwd_Mha : public testing::TestWithParam<TestCase>
             tensors[id] = std::move(tmp);
         };
 
-        auto GenScaledTensor = [](auto... nhsd) {
-            auto val_full = tensor<float>{nhsd...};
+        auto GenScaledTensor = [](auto type, auto... nhsd) {
```
amberhassaan, line 182: consider combining with `GetScaledTensor` and `ScaledTensor` in `mha_helper.hpp` as it will help with writing `graph API` tests for backward MHA.

amberhassaan (approved):
Some minor comments and questions.

src/solver/mha/mha_solver_backward.cpp
CAHEK7: We support this solver for fp32 types always, and for fp8 type if rocblas supports it.

test/gtest/gpu_mha_backward.cpp
CAHEK7: Correct.
It is by cudnn documentation, and also we can save some workspace size, since FP32xBF8 gemm is possible while FP32xF8 is not, and we need additional buffer to make a conversion.
Right now we do it in-place.

test/gtest/gpu_mha_backward.cpp
CAHEK7: Sure. Just keep in the mind that it's an opposite function, it generates fp8\bf8 data and then invent some scaling/descaling values.

junliume (approved):

test/gtest/gpu_mha_backward.cpp
amberhassaan: I noticed that it looked different. May be we should have two functions, each of which generates a `ScaledTensor` object in different ways.

src/solver/mha/mha_solver_backward.cpp
CAHEK7: Let me think about it. The problem is that I either have to pass tensor descriptor to that function and re-calculate all the parameters once again, or I can pass tensor descriptor just to get the TypeSize(), but actually I've already got the type earlier. It also breaks some sort of codestyle - some parts will be `sizeof(float)`, some others `GetTypeSize()` or `outputTypeSize`.
I'm not sure that there is just a simple replacement.

test/gtest/gpu_mha_backward.cpp
CAHEK7: #3040

PR #3032, opened 2024-06-07
Add a CMake option to control whether to strip debug symbols

xuantengh:
There is a [team](https://salsa.debian.org/rocm-team) trying to bring ROCm stack into the Debian official package archive. When packaging MIOpen, we find that it strips the debug symbols by default when building under release mode. Debian prefers to keep these symbols to produce an extra package (e.g., `libmiopen-dbgsym.deb` here) for the users to [obtain a full backtrace](https://wiki.debian.org/HowToGetABacktrace) when debugging.

Here we introduce a new CMake option `MIOPEN_STRIP_SYMBOLS` to control whether the debug symbols are stripped when building under release. Thus one can pass `-DMIOPEN_STRIP_SYMBOLS=OFF` when configuring CMake to achieve this.


junliume (approved):

CMakeLists.txt
```diff
@@ -103,10 +103,11 @@ set(MIOPEN_ENABLE_SQLITE_BACKOFF On CACHE BOOL "")
 
 option( BUILD_DEV "Build for development only" OFF)
 option(MIOPEN_ENABLE_FIN "Enable the fin driver for MIOpen"  OFF)
+option(MIOPEN_STRIP_SYMBOLS "Strip symbols in release mode" ON)
 
 
 # Strip symbols for release
-if(NOT WIN32 AND NOT APPLE)
+if(MIOPEN_STRIP_SYMBOLS AND NOT WIN32 AND NOT APPLE)
```
amberhassaan, line 110: @junliume : why are we doing this? There is value in having debug symbols in release builds.

junliume: @amberhassaan it just provides an option to do so if I see correctly

amberhassaan: @junliume : the switch is ON by default. I'd like it to be off by default so that we an debug release builds more effectively.

xuantengh: > the switch is ON by default. I'd like it to be off by default so that we an debug release builds more effectively.

MIOpen will strip the symbols before, so the default value is set as `ON` to match this default behavior.

PR #3033, opened 2024-06-07
Fix memory leak when using deref()

scerzh:
`deref()` is often used improperly, without taking into account that the right part of operator `=` in c++17 is evaluated before the left part. Prior c++17 the order of evaluation was undefined.

The following code will produce a memory leak if `desc` is `null` because `deref` throws an exception when memory is already allocated (and object is created). `destructor` won't be called possibly producing even more issues.

```C++
miopen::deref(desc) = new SomeDescriptor();
```

A branch _`sg/deref_mem_leak_demo`_ was created with the gtest file `deref_mem_leak.cpp` to demonstrate this misuse of `deref`.

Take a look and try its first commit [Demonstrate a memory leak with deref()](https://github.com/ROCm/MIOpen/commit/5c9c4596372212946ce8977307a401a04564f4d6) which shows the above issue for `miopenCreateTensorDescriptor()` by counting leaked objects and failing a corresponding `TEST`.

Also a fix was added with the same `TEST` in this branch's second commit [Fix memory leak with deref() for TensorDescriptor](https://github.com/ROCm/MIOpen/commit/e7005f24b3edabc440401660f238a9912f34912a)

I was searching for the instances of this issue using the following cmd
```bash
git grep -n -e 'deref([^(^)]*) *= *new'
```
I'm encouraging you to suggest other instances or search techniques.

Do you think if there is a way of writing a test to detect this issue without altering the source class (like in the demo branch)?

### DIFF
--- a/src/rnn.cpp
+++ b/src/rnn.cpp
@@ -276,18 +276,19 @@ std::vector<int> RNNDescriptor::pTensorLengthsCalculation(const TensorDescriptor
 
 RNNDescriptor::RNNDescriptor()
 {
-    nLayers                     = 1;
-    hsize                       = 0;
-    nHiddenTensorsPerLayer      = 0;
-    rnnMode                     = miopenRNNTANH;
-    dirMode                     = miopenRNNunidirection;
-    biasMode                    = miopenRNNNoBias;
-    algoMode                    = miopenRNNdefault;
-    inputMode                   = miopenRNNlinear;
-    dataType                    = miopenFloat;
-    typeSize                    = 4;
-    workspaceScale              = 1;
-    miopen::deref(&dropoutDesc) = new miopen::DropoutDescriptor();
+    nLayers                = 1;
+    hsize                  = 0;
+    nHiddenTensorsPerLayer = 0;
+    rnnMode                = miopenRNNTANH;
+    dirMode                = miopenRNNunidirection;
+    biasMode               = miopenRNNNoBias;
+    algoMode               = miopenRNNdefault;
+    inputMode              = miopenRNNlinear;
+    dataType               = miopenFloat;
+    typeSize               = 4;
+    workspaceScale         = 1;
+    auto& desc             = miopen::deref(&dropoutDesc);
+    desc                   = new miopen::DropoutDescriptor();
 }
 
 RNNDescriptor::RNNDescriptor(int hsz,
@@ -337,15 +338,16 @@ RNNDescriptor::RNNDescriptor(int hsz,
         typeSize = dType == miopenHalf ? 2 : 4;
     }
 
-    hsize                       = hsz;
-    nLayers                     = layers;
-    inputMode                   = inMode;
-    dirMode                     = bidir;
-    rnnMode                     = rmode;
-    algoMode                    = amode;
-    biasMode                    = bmode;
-    dataType                    = dType;
-    miopen::deref(&dropoutDesc) = new miopen::DropoutDescriptor();
+    hsize      = hsz;
+    nLayers    = layers;
+    inputMode  = inMode;
+    dirMode    = bidir;
+    rnnMode    = rmode;
+    algoMode   = amode;
+    biasMode   = bmode;
+    dataType   = dType;
+    auto& desc = miopen::deref(&dropoutDesc);
+    desc       = new miopen::DropoutDescriptor();
 
     switch(rmode)
     {
--- a/src/rnn.cpp
+++ b/src/rnn.cpp
@@ -276,19 +276,18 @@ std::vector<int> RNNDescriptor::pTensorLengthsCalculation(const TensorDescriptor
 
 RNNDescriptor::RNNDescriptor()
 {
-    nLayers                = 1;
-    hsize                  = 0;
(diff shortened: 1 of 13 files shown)


scerzh:
@atamazov could you take a look?

src/rnn.cpp
```diff
@@ -276,18 +276,19 @@ std::vector<int> RNNDescriptor::pTensorLengthsCalculation(const TensorDescriptor
...
+    rnnMode                = miopenRNNTANH;
+    dirMode                = miopenRNNunidirection;
+    biasMode               = miopenRNNNoBias;
+    algoMode               = miopenRNNdefault;
+    inputMode              = miopenRNNlinear;
+    dataType               = miopenFloat;
+    typeSize               = 4;
+    workspaceScale         = 1;
+    auto& desc             = miopen::deref(&dropoutDesc);
+    desc                   = new miopen::DropoutDescriptor();
```
scerzh: seems like not needed here

src/rnn.cpp
```diff
@@ -337,15 +338,16 @@ RNNDescriptor::RNNDescriptor(int hsz,
...
+    hsize      = hsz;
+    nLayers    = layers;
+    inputMode  = inMode;
+    dirMode    = bidir;
+    rnnMode    = rmode;
+    algoMode   = amode;
+    biasMode   = bmode;
+    dataType   = dType;
+    auto& desc = miopen::deref(&dropoutDesc);
+    desc       = new miopen::DropoutDescriptor();
```
scerzh: seems like not needed here

scerzh (commented):
Good `derefs` were marked

junliume (approved):

atamazov:
@scerzh Thanks for this finding! 👍 

> The following code will produce a memory leak if `desc` is `null`...

Am I understanding correctly that this only happens when the client's app uses the library in a certain way? And then the library returns an error code + unwanted side effect (memleak)?

atamazov:
@scerzh @CAHEK7 @DrizztDoUrden @JehandadKhan @junliume For a memleak to happen, the user must pass `nullptr` (instead of a pointer to `miopenSomething_t`) to `miopenCreateSomething()`.

In this case, the library allocates memory for an object, but cannot store its address in the object handle. As a result, the user's program won't be able to perform any operations with object.

These operation include destroying an object. In other words, there is no appropriate/valid call to `miopenDestorySomething()` in the user's program.

Consequently the user's program is ill-formed from the MIOpen's POV. 

In this case returning error code to the user is above expectations, and memleak is acceptable. This is anyway better than quiet memleaks without any error messages that happen if some `miopenDestorySomething()` is missing (as a result of a user's mistake, which is essentially the equivalent situation).

~So I don't think this patch is necessary. The commit may be reverted or left as is, but I don’t see any practical need to follow the "first convert the opaque handle into an object, and then allocate memory" policy in the future...~

UPDATE: ~...in this layer of the library.~

UPDATE: see https://github.com/ROCm/MIOpen/pull/3033#issuecomment-2156173180

**In any case, thanks to Sergei for the work done.** Attention to detail is very important. And this inspired me to repeat a chain of reasoning (made several years ago) once again and share it with the team.

CAHEK7:
Strongly disagree, it must not be reverted.
We must properly handle and report malformed data - just a garbage result may be acceptable, but memory (or other resources) leaks, nullptr dereferences or other nasty thing must not. It's a potential security issue.

@atamazov I'm a bit surprised seeing a revert recommendation and a recommendation to ignore such behavior. Why would we stop checking for a nullpointer in `miopen::deref` and just dereference it? It is ill-formed data from everybody's POV and must not happen ever.

atamazov:
@CAHEK7 
> Strongly disagree, it must not be reverted.

Am I understand correctly that you disagree with the following: "_The commit may be reverted or left as is_"?

> Why would we stop checking for a nullpointer in `miopen::deref` and just dereference it?

I never recommended that.

> memory (or other resources) leaks ... It's a potential security issue.

This is more interesting. If a user's application has vulnerabilities that allow an attacker to force the app to call MIOpen like "miopenCreateSomething(nullptr)", then the attacker could cause resource exhaustion.

atamazov:
Finally, the commit is good and should not be reverted (for the sake of "theoretical" danger, but what the heck). Let's follow the "first dereference ~convert the opaque handle into an object~, and then allocate memory" policy in this layer of the library.

scerzh:
> Finally, the commit is good and should not be reverted (for the sake of "theoretical" danger, but what the heck). Let's follow the "first convert the opaque handle into an object, and then allocate memory" policy in this layer of the library.

I don't see the following to the "first convert the opaque handle into an object here, and then allocate memory" policy here. Because it's not a handle that is passed to `miopenCreate...` functions but it is a pointer to a handle; and a plain dereference of a pointer is happening here.

CAHEK7:
> I never recommended that.

I recommend that, because it's the way how to show that the discussion about this PR went too far. Adding some nonsense and that's how it may look like (explicitly hyperbolized):
 
> Passing nullptr is not gonna happen ever, we can do whatever we want. It's just a "theoretical" danger, so why we should check for nullptr?

But finally we converged to the final conclusion.

Issue #3034, opened 2024-06-08
Tip of `develop` branch fail to build on older Linux distributions


junliume:
issues observed other than Ubuntu 22 or RHEL9 (these are relatively new):

[Build_Fail_Redhat.log](https://github.com/user-attachments/files/15748806/Build_Fail_Redhat.log)

Examples:
```
[2024-06-08T15:02:18.554Z] /opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../include/c++/9/bits/hashtable_policy.h:1384:9: error: type 'const std::hash<std::filesystem::path>' does not provide a call operator
[2024-06-08T15:02:18.554Z]  1384 |         return _M_h1()(__k);
[2024-06-08T15:02:18.554Z]       |                ^~~~~~~
```
```
[2024-06-08T15:02:12.767Z] /opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../include/c++/9/bits/unordered_map.h:228:28: error: call to implicitly-deleted default constructor of 'hasher' (aka 'std::hash<std::filesystem::path>')
[2024-06-08T15:02:12.767Z]   228 |                     const hasher& __hf = hasher(),
[2024-06-08T15:02:12.767Z]       |                                          ^
```
```
[2024-06-08T15:02:13.023Z] /opt/rh/devtoolset-9/root/usr/lib/gcc/x86_64-redhat-linux/9/../../../../include/c++/9/bits/hashtable_policy.h:1382:16: error: static assertion failed due to requirement 'std::__is_invocable<const std::hash<std::filesystem::path> &, const std::filesystem::path &>{}': hash function must be invocable with an argument of key type
[2024-06-08T15:02:13.023Z]  1382 |         static_assert(__is_invocable<const _H1&, const _Key&>{},
[2024-06-08T15:02:13.023Z]       |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

junliume:
@atamazov @scerzh @CAHEK7 @JehandadKhan could you take a look at this issue? I think Ubuntu 20.04 can also reproduce this issue
RHEL 9 or Ubuntu 22.04 does not see these issues. Most likely due to incompatibility of older glibc++ versions in these legacy Linux distributions.

We should provide a workaround.

List of suspects which are causing these issues:
[...]

PR #3035, opened 2024-06-09
Workaround: Ubuntu 20.04 and RHEL no standard hash for std::filesystem::path

junliume:
#3034

### DIFF
--- a/src/kernels/kernel.cpp.in
+++ b/src/kernels/kernel.cpp.in
@@ -24,7 +24,7 @@
  *
  *******************************************************************************/
 #include <algorithm>
-#include <unordered_map>
+#include <map>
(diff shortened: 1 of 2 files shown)


apwojcik (approved):

CAHEK7:
I'm just curious why we did not use that solution:
```C++
struct Hash
{
    auto operator()(const std::filesystem::path& p) const noexcept
    {
        return std::filesystem::hash_value(p);
    }
};

std::unordered_map<std::filesystem::path, int, Hash> foo;
```
It perfectly works on gcc 7.1 and clang 7.0.0 which are pretty old.
@junliume  @apwojcik shall I resubmit a proper fix?

PR #3039, opened 2024-06-10
standalone kernels test + gtest

sgundabo:
Basic test to use MIOpen to compile HIP and OpenCL kernels and verify correctness by comparing the output against CPU implementation.


test/vector_add.cpp
```diff
@@ -0,0 +1,343 @@
...
+bool compare_equal(T r1, T r2)
+{
+    return r1 == r2;
+}
+
+template <typename T>
+bool verify_tensor(tensor<T>& t_gpu1, tensor<T>& t_gpu2)
+{
+    EXPECT(t_gpu1.data.size() == t_gpu2.data.size());
+    auto idx          = miopen::mismatch_idx(t_gpu1.data, t_gpu2.data, compare_equal<T>);
```
amberhassaan: `compare_equal` can be a `lambda`. Good time to learn about those.

src/kernels/MIOpenVecAdd.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "float_types.h"
+
+template <typename T>
+inline __device__ void vectorAdd_function( T* a,
+                                        T* b,
+                                        T* c)
+{
+    int i = blockIdx.x * blockDim.x + threadIdx.x;
+    c[i] = a[i] + b[i];
```
CAHEK7: I guess array boundaries should be checked here, unless you can guarantee that the size can be precisely expressed by `blockIdx.x * blockDim.x + threadIdx.x`

src/kernels/MIOpenVecAddOCL.cl
```diff
@@ -0,0 +1,36 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+__kernel void vector_add(__global const float *x, 
+                         __global const float *y, 
+                         __global float* restrict z)
+{
+    // get index of the work item
+    int index = get_global_id(0);
```
amberhassaan, line 32: add check for `array size` . Something like `if (index < size)`.

src/kernels/MIOpenVecAdd.cpp
```diff
@@ -0,0 +1,45 @@
...
+#endif
+
+#include "float_types.h"
+
+template <typename T>
+inline __device__ void vectorAdd_function( T* a,
+                                        T* b,
+                                        T* c)
+{
+    int i = blockIdx.x * blockDim.x + threadIdx.x;
```
amberhassaan: add check for array size.

src/kernels/MIOpenVecAdd.cpp
```diff
@@ -0,0 +1,45 @@
...
+    c[i] = a[i] + b[i];
+}
+
+extern "C" __global__ void vectorAdd_kernel(void* a, void* b, void* c)
```
amberhassaan: take params as `float*`. No point taking `void*`. That way you avoid `reinterpret_cast`.

src/kernels/MIOpenVecAdd.cpp
```diff
@@ -0,0 +1,45 @@
...
+{
+    vectorAdd_function<float>(reinterpret_cast<float*>(a), reinterpret_cast<float*>(b), reinterpret_cast<float*>(c));
```
amberhassaan: template functions can be called without `<float>` type explicit type args if the function args can be used to deduce the type T.

amberhassaan (commented):
Some initial comments. I'd like to convert it to `gtest`.

src/kernels/MIOpenVecAdd.cpp
CAHEK7: All the casts from `void*` into `T*` can be a `static_cast`. But in this particular case, since you've hardcoded float into the template functions, even `void*` is not required and can be `float*` from the beginning.

src/kernels/MIOpenVecAddOCL.cl
CAHEK7: https://github.com/ROCm/MIOpen/pull/3039#discussion_r1633469273

src/kernels/MIOpenVecAdd.cpp
sgundabo: Yeah, this was just an initial implementation trying to get MIOpen to compile both OpenCL and HIP kernels. I will fix the kernels to account for boundary conditions.

CAHEK7:
Just a side note for upcoming OpenCL to HIP porting - there is no need to verify the algorithm itself (i.e. check against CPU implementation). 
The most important check is that the both GPU versions are working exactly the same in terms of the numbers and performance.
Tiny fluctuations are possible, but if there is a significant perf drop or numerical disparity - the new implementation must be thoroughly checked.

test/vector_add.cpp
```diff
@@ -0,0 +1,343 @@
...
+        for(size_t i = 0; i < r.data.size(); i++)
+            r[i] = 0;
+
+        auto dstC_dev  = handle.Write(r.data);
+
+
+        std::string program_name = "MIOpenVecAddOCL.cl";
+        std::string kernel_name  = "vector_add";
+
+        std::string network_config = "NW_CONFIG_PLACEHOLDER";
```
amberhassaan: this should be a descriptive and somewhat unique name that identifies the operation. Something like `standalone_kernel_vector_add_cl` and then similar for hip. The string makes up the key for the kernel cache, so it needs to be unique.

src/kernels/MIOpenVecAdd.cpp
```diff
@@ -0,0 +1,42 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#ifndef MIOPEN_DONT_USE_HIP_RUNTIME_HEADERS
+#include <hip/hip_fp16.h>
+#include <hip/hip_runtime.h>
+#endif
+
+extern "C" __global__ void vector_add_hip(const float * a, const float * b, float * c, uint vec_size)
```
amberhassaan: use `size_t` for vec_size.

src/kernels/MIOpenVecAddOCL.cl
```diff
@@ -0,0 +1,39 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+__kernel void vector_add_ocl(__global const float *x, 
+                         __global const float *y, 
+                         __global float* restrict z,
+                         uint vec_size
```
amberhassaan: `size_t`

test/gtest/vec_add.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include <miopen/env.hpp>
+#include "vec_add.hpp"
+
+namespace vecadd {
```
amberhassaan, line 205: combine the `.h` and `.cpp` file into one `.cpp` file. No need for separate header file in this case (unless parts of the header file are being used elsewhere).

test/gtest/vec_add.cpp
```diff
@@ -0,0 +1,50 @@
...
+using namespace vecadd;
+
+TEST_P(VecAddTestFloat, VecAddTestFw)
+{
+    RunTestOCL();
+    VerifyOCL();
+    RunTestHIP();
+    VerifyHIP();
+
+    // This verifies OCL and HIP results.
```
amberhassaan: this is what we want actually. But verifying against CPU is an optional add on.

test/vector_add.cpp
```diff
@@ -0,0 +1,313 @@
+/*******************************************************************************
```
amberhassaan: you can remove this file now.

amberhassaan (commented):
More review comments.

src/kernels/MIOpenVecAddOCL.cl
sgundabo: OpenCL does not support the use of size_t in kernel parameters, so I had uint for both kernels for consistency.

test/gtest/vec_add.cpp
sgundabo: Yes, I understand. But, I figured that in the context of the vector_add example, it would make sense to compare against the CPU. But still left the OCL vs HIP function as an option.

amberhassaan: I suggest enabling this test as that's how we will be testing kernels going forward.

src/kernels/MIOpenVecAddOCL.cl
amberhassaan: Just use an 8-byte equivalent like `unsigned long` which is supported in OpenCL. My issue with `uint` is that it's not a proper type name and most size things in C++ are 8-bytes unsigned (i.e. size_t). `uint` is like street slang :).

amberhassaan (approved):
good to go @junliume

src/kernels/MIOpenVecAddOCL.cl
sgundabo: So according to [OpenCL Data Types](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#built-in-scalar-data-types) `size_t ` in OpenCL returns "The `unsigned integer` type of the result of the `sizeof `operator." and the `unsigned integer` type is 4-bytes in the OpenCL 3.0 spec.

Your comment makes even more sense now!

junliume:
> good to go @junliume

clang format issue fixed, another round of CI kicked off

src/kernels/MIOpenVecAdd.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+    // Get the index of the current element
+    size_t index = blockIdx.x * blockDim.x + threadIdx.x;
+
+    // Check if the index is within the range of the vector size
+    if(index < vec_size)
+    {
+        c[index] = a[index] + b[index]; // Add the two elements
+    }
+}
```
amberhassaan, line 41: seems to be missing `<CR>` at the end of the file.

amberhassaan:
@junliume : please merge this PR when you get a chance.

junliume (approved):

PR #3043, opened 2024-06-10
fix compilation without rocblas

apwojcik:
Fix the compilation while `MIOPEN_USE_ROCBLAS` is `OFF`. We need to strip off the rocblas library in UAI, so this one is important.

### DIFF
--- a/src/ocl/rnnocl.cpp
+++ b/src/ocl/rnnocl.cpp
@@ -42,17 +42,17 @@ namespace miopen {
 
 namespace {
 
+#if MIOPEN_USE_ROCBLAS && MIOPEN_BACKEND_HIP
+
 bool RNNForwardMSIsSupported([[maybe_unused]] const RNNDescriptor& desctiptor,
                              [[maybe_unused]] bool use_dropout)
 {
-#if MIOPEN_USE_GEMM && MIOPEN_BACKEND_HIP
     if(desctiptor.rnnMode == miopenLSTM && desctiptor.algoMode == miopenRNNdefault &&
        !use_dropout && desctiptor.nLayers > 1 && desctiptor.dirMode == miopenRNNunidirection &&
        desctiptor.inputMode != miopenRNNskip)
     {
         return true;
     }
-#endif // MIOPEN_USE_GEMM&& MIOPEN_BACKEND_HIP
     return false;
 }
 
@@ -242,6 +242,8 @@ miopenStatus_t ReducAddBias(miopen::Handle& handle,
 
     return miopenStatusSuccess;
 }
+
+#endif // MIOPEN_USE_ROCBLAS && MIOPEN_BACKEND_HIP
 
 } // namespace
 
@@ -1078,8 +1080,6 @@ void RNNDescriptor::RNNForwardMS(Handle& handle,
     (void)y;
     (void)hy;
     (void)cy;
-    (void)reserveSpace;
-    (void)reserveSpaceSize;
 
     MIOPEN_THROW("GEMM is not supported");
 #endif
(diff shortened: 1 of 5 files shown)


src/solver/mha/mha_solver_backward.cpp
```diff
@@ -90,8 +90,8 @@ bool MhaBackward::IsApplicable([[maybe_unused]] const ExecutionContext& context,
 
     auto [N, H, S, D] = miopen::tien<4>(descsBwd.kDesc.GetLengths());
 
-    return MIOPEN_USE_GEMM                                            //
-           && !miopen::IsDisabled(ENV(MIOPEN_DEBUG_ATTN_NAIVE_BWD))   //
+#if MIOPEN_USE_ROCBLAS
```
CAHEK7: It does not contain any rocblas-specific functionality, except the `USE_ROCBLAS_EX3` which should be properly defined with or without rocblas.
And currently, prior to #3030 `MIOPEN_USE_GEMM` is equal to `MIOPEN_USE_ROCBLAS` so it is not applicable and must not be excluded in that way.

CAHEK7 (changes requested):
Could you please explain what is the problem with `isApplicable` methods?

src/solver/mha/mha_solver_backward.cpp
apwojcik: When `MIOPEN_USE_GEM` (aka `MIOPEN_USE_ROCBLAS`) is 0, the compiler warns that the rest of the expression has no effect. When `-Werror` is additionally being set, compilation breaks.

apwojcik: Here is the effect of `MIOPEN_USE_GEMM` equal to `0` (false) when compiling.

```
C:\...\MIOpen\src\solver\mha\mha_solver_backward.cpp:115:17: warning: code will never be executed [-Wunreachable-code]
  115 |            && ((descsBwd.kDesc.GetType() == miopenFloat)              //
      |                 ^~~~~~~~
C:\...\MIOpen\src\solver\mha\mha_solver_backward.cpp:114:15: warning: code will never be executed [-Wunreachable-code]
  114 |            && descsBwd.kDesc.GetType() == descsBwd.dvDesc.GetType()   //
      |               ^~~~~~~~
C:\...\MIOpen\src\solver\mha\mha_solver_backward.cpp:113:15: warning: code will never be executed [-Wunreachable-code]
  113 |            && descsBwd.kDesc.GetType() == descsBwd.dkDesc.GetType()   //
      |               ^~~~~~~~
C:\...\MIOpen\src\solver\mha\mha_solver_backward.cpp:112:15: warning: code will never be executed [-Wunreachable-code]
  112 |            && descsBwd.kDesc.GetType() == descsBwd.dqDesc.GetType()   //
...
```
There are about 40 of these kinds of messages in a row.

CAHEK7: Ahh, but it is intentional. It looks like false-positive for me.
Ok, don't want fighting with the compiler, let's just exclude it by ifdef. But there will be a merge conflict with #3030

CAHEK7 (approved):

src/ocl/rnnocl.cpp
```diff
@@ -1078,8 +1082,6 @@ void RNNDescriptor::RNNForwardMS(Handle& handle,
     (void)y;
     (void)hy;
     (void)cy;
-    (void)reserveSpace;
```
junliume, line 1081: @apwojcik why these two lines of changes?

junliume: We can let it go as long as hiptidy is not complaining, but it looks like not necessary.

junliume (approved):

PR #3045, opened 2024-06-11
Rs/conv immediate

randyspauldingamd:
Ready for review

- I decided to keep the TYPED_TEST. It's a bit clunky due to gtest limitations, and somewhat limited in scalability, but compared to alternatives it isn't all that bad and it will be a useful example in the future.
- It would be nice to add `miopenTranspose` mode, but the CPU code causes core dumps in this mode. I wasn't able to find the cause in the time available, so this is left for future work.


test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,904 @@
...
+    TypeDefs<miopen::conv::Direction::BackwardData,     float16,    float16,    layout>,    \
+    TypeDefs<miopen::conv::Direction::BackwardData,     bfloat16,   bfloat16,   layout>,    \
+                                                                                            \
+    TypeDefs<miopen::conv::Direction::BackwardWeights,  float,      float,      layout>,    \
+    TypeDefs<miopen::conv::Direction::BackwardWeights,  float16,    float16,    layout>,    \
+    TypeDefs<miopen::conv::Direction::BackwardWeights,  bfloat16,   bfloat16,   layout>
+// clang-format on
+
+typedef ::testing::Types<MakeTypeDefs(miopenTensorNCHW), MakeTypeDefs(miopenTensorNHWC)>
+    Implementations2d;
```
CAHEK7:
```suggestion
using Implementations3d = ::testing::Types<MakeTypeDefs(miopenTensorNCDHW), MakeTypeDefs(miopenTensorNDHWC)>;
```

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,904 @@
...
+template <miopen::conv::Direction direction,
+          typename TREF,
+          typename TOUT,
+          miopenTensorLayout_t tensor_layout>
+struct TypeDefs
+{
+    static constexpr miopen::conv::Direction get_direction() { return direction; };
+    typedef TREF TRef;
+    typedef TOUT Tout;
+    static constexpr miopenTensorLayout_t get_tensor_layout() { return tensor_layout; };
```
CAHEK7: I don't think it's significant difference between `constexpr` function and static member, I just prefer static member to keep using this aggregate uniformly across the types and constants. Moreover all the `std::` constants are constants, not functions, check `std::is_same_v` implementation.
But `using` must be used instead of `typedef`. It is more straight and potentially more flexible.
Inverted typedef logic is kind of ugly.
```suggestion
    static constexpr miopen::conv::Direction Direction = direction;
    using TRef = TREF;
    using Tout = TOUT;
    static constexpr miopenTensorLayout_t Layout = tensor_layout;
```

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,904 @@
...
+};
+
+template <class Types>
+class ReferenceConv2d : public ReferenceConvBase
+{
+protected:
+    template <miopen::conv::Direction direction,
+              typename TRef,
+              typename Tout,
+              miopenTensorLayout_t tensor_layout>
```
CAHEK7:
```suggestion
```

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,904 @@
...
+    void run()
+    {
+        auto run_conv_2d = [&](int n,
```
CAHEK7:
```suggestion
    {
        using TRef = Types::TRef;
        using Tout = Types::Tout;

        static constexpr auto direction        = Types::Direction;
        static constexpr auto tensor_layout = Types::Layout;
        auto run_conv_2d = [&](int n,
```

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,904 @@
...
+    Implementations2d;
+
+TYPED_TEST_CASE(ReferenceConv2d, Implementations2d);
+
+TYPED_TEST(ReferenceConv2d, Test)
+{
+    typedef typename TypeParam::TRef TRef;
+    typedef typename TypeParam::Tout Tout;
+
+    this->template run<TypeParam::get_direction(), TRef, Tout, TypeParam::get_tensor_layout()>();
```
CAHEK7:
```suggestion
    run();
```

CAHEK7 (changes requested):
I suggested some simplifications.

But first of all, `direction` and `layout` must not be template parameters. It slows down compilation and bloats the code. That's why the biggest problem is not a passing multiple template arguments into `TYPED_TEST`, but a mixing value-parametrized `TEST_P` and type-parametrized `TYPED_TEST` approaches.

Typed approach is not scalable at all and can handle just a few dozens of hand-written cases, while we may need much more and generate it.

CAHEK7:
I would suggest not mixing type-parametrized and value-parametrized tests in our test suite without really strong reason - we've already got quite complex construction and let's not overcomplicate it more.

Most of the tests relies on value parametrization and if we need different types, it can be solved manually, or I can provide some macros - but in this case we can still use standardized test declaration and we can use all the power of value-parametrized test like passing list of test cases or even combine and mix the cases.

@JehandadKhan may I also suggest to stop any testing related activities until we finalize the test design documentation and creat some sort of plan how we will fix current gtests, port legacy remains to gtest and add new tests. I expect another round of quite significant changes in the infrastructure.

test/gtest/conv_immediate.cpp
randyspauldingamd: accepted

test/gtest/conv_immediate.cpp
randyspauldingamd: accepted

test/gtest/conv_immediate.cpp
randyspauldingamd: accepted

test/gtest/conv_immediate.cpp
randyspauldingamd: accepted

CAHEK7:
May I ask you to update the test according to [the document](https://github.com/ROCm/MIOpen/wiki/GTest-development) (either by updating the document, or the test, or even both)

test/gtest/cartesian_product.hpp
```diff
@@ -0,0 +1,104 @@
...
+        {
+            for(auto b2 : B)
+            {
+                C.push_back(std::tuple_cat(a, std::make_tuple(b, b2)));
+            }
+        }
+    }
+
+    return C;
+}
```
CAHEK7, line 69: Everything in this file looks like reinvention of `testing::Combine`. I'm not sure we have to do this, but probably it could be useful if we can extend it with filtering and limiting capabilities.
But even in this case we can always alter the result of `testing::Combine` unless we have too many cases that `testing::Combine` can fail.

test/gtest/cartesian_product.hpp
```diff
@@ -0,0 +1,104 @@
...
+#pragma once
+
+#include <tuple>
+#include <vector>
+
+namespace test {
+
+// converts a vector of a type to a vector of tuples of the type
+template <typename T>
+inline std::vector<std::tuple<T>> vec_to_tuple(std::vector<T> in)
```
CAHEK7, line 35:
```suggestion
inline std::vector<std::tuple<T>> vec_to_tuple(const std::vector<T>& in)
```
Passing `std::vector` by value look suspicious. Is it intended?

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,532 @@
...
+    {
+        miopenTensorDescriptor_t desc;
+
+        EXPECT_TRUE(miopenCreateTensorDescriptor(&desc) == miopenStatusSuccess);
+        EXPECT_TRUE(
+            miopenSetTensorDescriptor(desc, type, lens.size(), lens.data(), strides.data()) ==
+            miopenStatusSuccess);
+
+        return desc;
+    }
```
CAHEK7, line 158: It does not depend on `T`, but each distinct type will generate it's own set of those functions.
Though it's not a big deal.

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+    void read_gpu_data(miopen::Handle& handle, const miopen::Allocator::ManageDataPtr& ddata)
+    {
+        check      = tensor<T>{descriptor.GetLengths(), descriptor.GetStrides()};
+        check.data = handle.Read<T>(ddata, check.data.size());
+    }
+
+    tensor<T> check{};
+    tensor<T> host;
+    miopen::TensorDescriptor descriptor;
```
CAHEK7, line 178: I'm not sure that a separated `miopen::TensorDescriptor` required, since `tensor<T>` has already included it.
https://github.com/ROCm/MIOpen/blob/bd993eb488bac41b4dd977cfc60db7dc0ba689fa/test/tensor_holder.hpp#L143

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,532 @@
...
+        descriptor   = miopen::TensorDescriptor{miopen_type<T>{}, lens, strides};
+        host         = tensor<T>{lens, strides}.generate(gen_value<T>);
+    }
+
+    ~layout_data() {}
+
+    void read_gpu_data(miopen::Handle& handle, const miopen::Allocator::ManageDataPtr& ddata)
+    {
+        check      = tensor<T>{descriptor.GetLengths(), descriptor.GetStrides()};
+        check.data = handle.Read<T>(ddata, check.data.size());
```
CAHEK7, line 173: It's not efficient, since we allocate twice more memory that we need.
I'd suggest to use `handle.ReadTo` directly to the already allacated vector. Currently `Read<T>` creates a new one and replaces previously allocated.

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,532 @@
...
+        // Channel size must be multiple of group size
+        auto sz = get_group_size();
+        std::transform(sz.begin(), sz.end(), sz.begin(), [&](int s) { return s * 4; });
+
+        return sz;
+    }
+
+    static std::vector<int> get_batch_size() { return {1, 2}; }
+
+    static bool same(std::vector<int> v) { return std::equal(v.begin() + 1, v.end(), v.begin()); }
```
CAHEK7, line 263: Passing `std::vector` by value looks suspicious.
Also there it an UB if `v` is empty.

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,532 @@
...
+            int wo = conv_out_size(wi, px, dx, fx, sx);
+
+            if(fy > hi || fx > wi || (fy - 1) < py || (fx - 1) < px || ho <= 0 || wo <= 0 ||
+               c % g != 0 || k % g != 0)
+                continue;
+            if((fx == 3 && fy == 5) || (fx == 5 && fy == 3))
+                continue;
+
+            f(n, {wi, hi}, c, k, {fy, fx}, {py, px}, {sy, sx}, {dy, dx}, g);
+        }
```
CAHEK7, line 297: The problem of this approach is that - we don't have separated test cases and if something got failed in the middle, we have to rerun aaaall the cases.

We have a bunch of similar problems in the tests, where we misuse `testing::Value` instead of `testing::ValueIn`, let's don't create a new one.

CAHEK7 (changes requested):
I still have doubts that we have to introduce one more way to run such tests, especially if there is no ability to run some particular configuration.
If you decide to go this way, don't forget do describe this approach in the documentation and put some recommendations why and when we should use `TYPED_TEST` instead of this `TEST_P` approach with types: https://github.com/ROCm/MIOpen/blob/bd993eb488bac41b4dd977cfc60db7dc0ba689fa/test/gtest/gpu_mha_backward.cpp#L421

randyspauldingamd (commented):
misclick

test/gtest/conv_immediate.cpp
randyspauldingamd: Nice, I didn't know about ReadTo. Any time difference wasn't noticeable, but it's still a warm and fuzzy.

test/gtest/conv_immediate.cpp
randyspauldingamd: It's no longer in use, I'll delete it.

test/gtest/cartesian_product.hpp
randyspauldingamd: It was not intended. I _think_ it's the same with clang (due to inline), but yours is better anyway (clearer, more portable).

test/gtest/cartesian_product.hpp
randyspauldingamd: There was some massive overhead using TEST_P and testing::Combine which I did not want to investigate. This is appx. 8x faster (8-ish seconds vs 60-ish).

test/gtest/conv_immediate.cpp
randyspauldingamd: Valid point, but the overhead associated with those test cases is horrible for a test space this vast (it was over 80k test cases at one point). I usually add custom instrumentation anyway, so adding a filter or custom data to isolate the bad ones before rebuilding is trivial.

test/gtest/conv_immediate.cpp
randyspauldingamd: Interesting. I've never had a reason to be concerned about bloat in test code. I would personally prefer this over something like an unnamed namespace unless it was a ridiculous size difference. But that's just me :)

randyspauldingamd:
> I still have doubts that we have to introduce one more way to run such tests, especially if there is no ability to run some particular configuration. <snip>

There are some great discussion points here--I'm super busy the next couple weeks, but I would be interested in throwing a meeting together at some point.

If you only have a single type parameter, I think a strong argument can be made that TYPED_TEST is cleaner. Here's another example I looked at:

https://stackoverflow.com/a/64048808

It doesn't work in this case--and I also see a strong argument against the ugliness of this multi-parameter solution. In this case, the wall clock improvement wins though (as I pointed out in another comment).

randyspauldingamd:
> May I ask you to update the test according to [the document](https://github.com/ROCm/MIOpen/wiki/GTest-development) (either by updating the document, or the test, or even both)

Cool, I had heard some discussion about this but I wasn't aware that it was live. I think there are some interesting discussions to come, so I'll hold off on the documentation for a while. I will see if I can fix the naming tomorrow though.

test/gtest/cartesian_product.hpp
CAHEK7: > I _think_ it's the same with clang (due to inline)

It is not. And there no chance for that kind of optimization even for the perfectly inlined code - std::vector constructor may throw, alter global state and has side-effects - you can't remove memory allocation.

test/gtest/cartesian_product.hpp
CAHEK7: If TYPED_TEST is 8x faster, then why we are still using TEST_P? Probably it is worth to investigate the problem and propose a better solution?
Is it `testing::Combine`? If it's not, then you can use `testing::Combine` for cartesian product and this code is no needed.
Is it `TEST_P` and its internal infrastructure? If it's not, then your cartesian product must be used in every test.
Or maybe it's a test structure and resource managing and you can give some recommendations how to do it in a better way.

test/gtest/conv_immediate.cpp
CAHEK7: The tests take more compilation time than the library itself. We shall think about this too, as well as about the wall clock execution time.

Btw, unnamed namespace and static member function are not equivalent.

test/gtest/conv_immediate.cpp
CAHEK7: Sure, it's not a problem for tiny tensors. As well as replacing `std::vector<T>` by `const std::vector<T>&` in the `vet_to_tuple` won't show you any difference.

Since you've provided some abstraction, it would be nice to have a good abstraction without obvious pitfalls, or those pitfalls must be visible and clearly documented.

test/gtest/conv_immediate.cpp
CAHEK7: Sure, not a big deal. May I ask you to investigate that horrible overhead and provide some recommendations when each approach should be used and why?

randyspauldingamd: That investigation is far from trivial, so (just to be clear) it will not be part of this PR. There was another team member who pointed this out not long after I saw it so there may already be a ticket (@BrianHarrisonAMD maybe?). If not, I'll make a ticket and we can discuss priority during planning.

BrianHarrisonAMD: I don't think I ever added anything.

Are you saying that the testing combine macros is leading to the delay when executing tests through make check? 
I think we see this delay even if they should be skipped.

Oddly, I don't notice any delay when executing the tests through the executables that are generated.
I would have expected them to behave the same.

CAHEK7: Then we postpone this PR until that investigation has not been done. We can't just randomly do various things. We've already got the test broken due to this.

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+namespace {
+
+static constexpr int RAND_INTEGER_MAX       = 5;
+static constexpr int RAND_INTEGER_MIN       = -4;
+static constexpr float MAX_INTEGER_INTERVAL = 4.f;
+
+template <typename T>
+auto gen_value =
+    [](auto... is) { return static_cast<T>(prng::gen_A_to_B(RAND_INTEGER_MIN, RAND_INTEGER_MAX)); };
```
BrianHarrisonAMD, line 60: Mostly just curious, but is there a reason to use a templated lambda variable instead of a method for the generator?

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+using Implementations2d =
+    ::testing::Types<MakeTypeDefs(miopenTensorNCHW), MakeTypeDefs(miopenTensorNHWC)>;
+
+TYPED_TEST_CASE(ReferenceConv2d, Implementations2d);
+
+TYPED_TEST(ReferenceConv2d, Test) { this->run(); }
+
+template <class Types>
+struct ReferenceConv3d : ReferenceConvBase<Types>
```
BrianHarrisonAMD, line 522: Do these tests cover the same functionality as [gpu_reference_kernel.cpp](https://github.com/ROCm/MIOpen/blob/da4ba7f838c8cd97e5a59e12c873f060e78d627c/test/gpu_reference_kernel.cpp)?

If yes, then should we remove the gpu_reference_kernel tests with this change?

test/gtest/conv_immediate.cpp
CAHEK7: If you pass is into `.generate()` method, it supplies input lambda with indexes based on the tensor's dimensions. `prng::` generator does not take any arguments, so we have to wrap it into lambda which ignores input indexes and just returns random value for each call.

BrianHarrisonAMD: Oh sorry, I meant I just found it surprising that gen_value was a lambda variable instead of a method like:

```cpp
template <typename T>
T GenValue(auto... is){ return static_cast<T>(prng::gen_A_to_B(RAND_INTEGER_MIN, RAND_INTEGER_MAX)); }
```

test/gtest/cartesian_product.hpp
amberhassaan: Also, what's the reasoning behind two copies of `B`?

test/gtest/conv_immediate.cpp
```diff
@@ -0,0 +1,532 @@
...
+            miopenStatusSuccess);
+
+        return desc;
+    }
+
+    layout_data(int _n, std::vector<int> _dims, int _c, miopenTensorLayout_t _tensor_layout)
+    {
+        auto lens    = get_layout_lengths(_n, _c, _dims);
+        auto strides = get_strides(lens, _dims.size(), _tensor_layout);
+        descriptor   = miopen::TensorDescriptor{miopen_type<T>{}, lens, strides};
```
amberhassaan, line 164: as @CAHEK7  suggested, it's better to use `tensor<T>` than `TensorDescriptor`.

amberhassaan (commented):
a couple of comments. Overall, I agree with @CAHEK7 that we should investigate the source of overhead and then arrive at a solution.

test/gtest/conv_immediate.cpp
randyspauldingamd: > Do these tests cover the same functionality as [gpu_reference_kernel.cpp](https://github.com/ROCm/MIOpen/blob/da4ba7f838c8cd97e5a59e12c873f060e78d627c/test/gpu_reference_kernel.cpp)?
> 
> If yes, then should we remove the gpu_reference_kernel tests with this change?

Yes, this is a direct replacement/port to gtest. I've just kept the original for reference temporarily.

test/gtest/conv_immediate.cpp
randyspauldingamd: This was existing code, and I don't have an opinion. I went with the conservation of effort approach and simply relocated it.

test/gtest/conv_immediate.cpp
randyspauldingamd: Oops, I missed this one. This was left over from an earlier revision that reused buffers. I think you're right, I'll try to make it go away.

test/gtest/conv_immediate.cpp
randyspauldingamd: Yes, please refer to my reply below.

randyspauldingamd: On the 'two copies of B' question: I was just replicating the original test cases which use repeated values for h and w for image, filter, and pad sizes. It ends up running two squares and two rectangles of each. For example, image sizes tested are [9x9], [14x9], [9x14], and [14x14].

junliume:
@randyspauldingamd can we follow up and avoid a hang PR? Thanks!

PR #3057, opened 2024-06-12
[HotFix] fix the setworkspace issue for ck wrw solvers

iq136boy:
The PR is to fix the setworkspace issue for ck wrw solvers for some input configs that need to call the optimized ck kernels.

### DIFF
--- a/src/convolution_api.cpp
+++ b/src/convolution_api.cpp
@@ -216,8 +216,6 @@ miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t a
 {
     MIOPEN_LOG_FUNCTION(alpha_beta_case, outputTensorDesc);
     return miopen::try_([&] {
-        (void)convDesc; // -warn
-        assert(miopen::deref(convDesc).spatialDim == 3);
         miopenDataType_t data_type = miopen::deref(outputTensorDesc).GetType();
         size_t in_spatial_dims     = miopen::deref(inputTensorDesc).GetNumDims();
 
@@ -229,9 +227,14 @@ miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t a
             miopen::GetNCDHW(in_spatial_dims, miopen::deref(outputTensorDesc).GetLengths()));
         size_t output_tensor_size = miopen::deref(outputTensorDesc).GetElementSize();
 
+        int group_count = miopen::deref(convDesc).GetGroupCount();
+
         size_t byte_size = 0;
         if((alpha_beta_case == BILINEAR || alpha_beta_case == SCALE ||
-            ((data_type == miopenHalf) && ((C & 1) != 0 || (K & 1) != 0 /* Test if odd*/))))
+            ((data_type == miopenHalf) &&
+             (((C / group_count == 1) && (K / group_count == 1)) || ((C & 1) != 0 || (K & 1) != 0
+                                                                     /* Test if odd*/
+                                                                     )))))
         {
             switch(data_type)
             {
--- a/src/convolution_api.cpp
+++ b/src/convolution_api.cpp
@@ -232,7 +232,7 @@ miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t a
         size_t byte_size = 0;
         if((alpha_beta_case == BILINEAR || alpha_beta_case == SCALE ||
             ((data_type == miopenHalf) &&
-             (((C / group_count == 1) && (K / group_count == 1)) || ((C & 1) != 0 || (K & 1) != 0
+             (((C / group_count == 1) || (K / group_count == 1)) || ((C & 1) != 0 || (K & 1) != 0
                                                                      /* Test if odd*/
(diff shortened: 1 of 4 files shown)


src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -798,6 +711,95 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
...
+        MIOPEN_LOG_E("PerformanceConfig kernel '" + kernel_id + "' does not exist.");
+        return {miopenStatusInvalidValue};
+    }
+
+    if constexpr(std::is_same_v<CastType, miopen::conv::WrWInvokeParams>)
+    {
+        miopenAlphaBetaCase_t alpha_beta_case = problem.GetAlphaBetaCase();
+        [[maybe_unused]] bool should_allocated_wrw_buffer =
+            ShouldAllocateWorkSpaceBufferForWRW(problem);
+        size_t spatial_dim = problem.GetSpatialDims();
```
bghimireamd, line 736: I guess we can remove `spatial_dim` variable now since no one is using it.

src/convolution_api.cpp
```diff
@@ -229,9 +227,14 @@ miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t a
             miopen::GetNCDHW(in_spatial_dims, miopen::deref(outputTensorDesc).GetLengths()));
         size_t output_tensor_size = miopen::deref(outputTensorDesc).GetElementSize();
 
+        int group_count = miopen::deref(convDesc).GetGroupCount();
+
         size_t byte_size = 0;
         if((alpha_beta_case == BILINEAR || alpha_beta_case == SCALE ||
-            ((data_type == miopenHalf) && ((C & 1) != 0 || (K & 1) != 0 /* Test if odd*/))))
+            ((data_type == miopenHalf) &&
+             (((C / group_count == 1) && (K / group_count == 1)) || ((C & 1) != 0 || (K & 1) != 0
```
amberhassaan: I can't read this anymore. Please create local variables with descriptive names as to what's being tested for and why (add comments too).

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: > I guess we can remove this variable now since no one is using it.

I prefer to keep it. The reason is that we might need to do the ck-based solver combination later since ck might deprecate the 2d instances and only keep 3d ones. We need the the dim variable to distinguish the 2d and 3d configs when we run them on 3d kernels.

src/convolution_api.cpp
amberhassaan: ok, I get it now but still want comments. Compute `C_per_group` and `K_per_group`. Those values being odd is what matters and that simplifies the check to `isOdd(C_per_group) || isOdd(K_per_group)`.

amberhassaan (approved):
one minor change requested.

src/convolution_api.cpp
JehandadKhan: I am curious why this detail is leaking into the API implementation side? Shouldnt the solvers just handle this on their own? Why do we have a conditional guarding this?

CAHEK7: It looks even more strange, since it is specific CK Backward Weight Workspace api function..

src/solver/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
```diff
@@ -309,6 +309,9 @@ struct CKArgs
     bool IsSupportedBy(const ConvPtr& conv_ptr) const
     {
         auto arg_ptr = MakeArgPtr(conv_ptr, nullptr, nullptr, nullptr, 1.0f, 0.0f);
+        // Creat dummy workspace to pass the ck IsSupportedArgument check.
+        int dummy_var = 1;
```
CAHEK7, line 313: I haven't looked int CK implementation and I'm curious about few questions:
Is it critical to pass `1`?
Can it be `0`?
Is it workspace size?
Does CK allocate workspace?

bghimireamd: It really does not matter what value we assign. [CK just needed a pointer](https://github.com/ROCm/composable_kernel/blob/develop/include/ck/tensor_operation/gpu/device/impl/device_grouped_conv_bwd_weight_two_stage_xdl_cshuffle.hpp#L1338) not to be null.

src/convolution_api.cpp
bghimireamd: only CK's 2d and 3d Backward Weight solver needed this trick. I would say its leaking from CK.

src/solver/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
iq136boy: @CAHEK7 yes. CK just need a not-null pointer to check. It does matter what value we give here.

junliume (approved):

PR #3059, opened 2024-06-14
fix wrw hip error

bghimireamd:
* Create function `CKWrwRequireWorkspace` 
* Only call CK's `SetWorkSpacePointer` in `IsSupportedBy` if workspace is required.

### DIFF
--- a/src/convolution_api.cpp
+++ b/src/convolution_api.cpp
@@ -232,9 +232,9 @@ miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t a
                                          size_t K,
                                          miopenDataType_t data_type,
                                          miopenAlphaBetaCase_t alpha_beta_case) {
-            auto is_odd     = [](int num) { return num % 2 != 0; };
-            int C_per_group = C / G;
-            int K_per_group = K / G;
+            auto is_odd        = [](int num) { return num % 2 != 0; };
+            size_t C_per_group = C / G;
+            size_t K_per_group = K / G;
 
             return (alpha_beta_case == BILINEAR || alpha_beta_case == SCALE) ||
                    (data_type == miopenHalf &&
--- a/src/convolution_api.cpp
+++ b/src/convolution_api.cpp
@@ -208,7 +208,7 @@ extern "C" miopenStatus_t miopenGetConvolutionFindMode(const miopenConvolutionDe
 }
 
 MIOPEN_EXPORT extern "C" miopenStatus_t
-miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t alpha_beta_case,
+miopenConvolutionABBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t alpha_beta_case,
                                                    const miopenTensorDescriptor_t inputTensorDesc,
                                                    const miopenTensorDescriptor_t outputTensorDesc,
                                                    const miopenConvolutionDescriptor_t convDesc,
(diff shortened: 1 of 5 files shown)


src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -558,9 +558,9 @@ inline size_t GetCKAlphaBetaWorkspace(const miopen::conv::ProblemDescription& pr
 inline bool CKWrwRequireWorkspace(
     size_t G, size_t C, size_t K, miopenDataType_t data_type, miopenAlphaBetaCase_t alpha_beta_case)
 {
-    auto is_odd     = [](int num) { return num % 2 != 0; };
-    int C_per_group = C / G;
-    int K_per_group = K / G;
+    auto is_odd        = [](size_t num) { return num % 2 != 0; };
+    size_t C_per_group = C / G;
+    size_t K_per_group = K / G;
```
amberhassaan, line 563: We do not need to check `==1` because guess what `1` is odd :).

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -558,9 +558,9 @@ inline size_t GetCKAlphaBetaWorkspace(const miopen::conv::ProblemDescription& pr
 inline bool CKWrwRequireWorkspace(
     size_t G, size_t C, size_t K, miopenDataType_t data_type, miopenAlphaBetaCase_t alpha_beta_case)
 {
-    auto is_odd     = [](int num) { return num % 2 != 0; };
-    int C_per_group = C / G;
-    int K_per_group = K / G;
+    auto is_odd        = [](size_t num) { return num % 2 != 0; };
```
amberhassaan, line 561: why are we replicating the logic. I'd like to avoid it by creating some function that is called in both places.

bghimireamd: Its same logic (check if odd) I was just making it more readable.

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: For example if `C` = 64 and `G` = 2, `C_per_group` = 32. So, C_per_group can be any number.

src/include/miopen/solver/implicitgemm_ck_util.hpp
bghimireamd: I was hoping to place `CKWrwRequireWorkspace` in one place so that both `src/convolution_api.cpp` and `src/include/miopen/solver/implicitgemm_ck_util.hpp` could use them though but not sure which file to choose.

src/include/miopen/solver/implicitgemm_ck_util.hpp
bartekxk: It is enough to check if C_per_group or K_per_group are odd. From CK point of view we doesnt care about C and K. It is only about C_per_group or K_per_group .

bartekxk (commented):
LGTM

src/convolution_api.cpp
```diff
@@ -221,20 +221,29 @@ miopenConvolutionCKBackwardWeightsGetWorkSpaceSize(const miopenAlphaBetaCase_t a
 
         assert(in_spatial_dims == miopen::deref(outputTensorDesc).GetNumDims());
 
+        int G    = miopen::deref(convDesc).GetGroupCount();
```
atamazov, line 224: General comment to the `miopenConvolutionCKBackwardWeightsGetWorkSpaceSize()` function. As you can see in the other cases, the functions that implement C API contain only thin C/C++ API adaptation layer (and possibly some trivial transformations, like we have for transposed convolutions), and actual job is performed elsewhere. Can you please follow this design principle?

src/include/miopen/solver/implicitgemm_ck_util.hpp
amberhassaan: OK, yeah, I didn't notice that. Good catch @bartekxk . @bghimireamd you need to input `C_per_group` to `isOdd`, not `C`.

src/convolution_api.cpp
atamazov: [Notice] Resolving a comment at https://github.com/ROCm/MIOpen/pull/2863#discussion_r1640547009 would automatically resolve this one.

junliume (approved):

PR #3061, opened 2024-06-17
fix kdb issue caused by naive alpha and beta  support

bghimireamd:
fix KDB incompatibility that was caused by https://github.com/ROCm/MIOpen/pull/2863

### DIFF
--- a/src/solver/conv_direct_naive_conv.cpp
+++ b/src/solver/conv_direct_naive_conv.cpp
@@ -129,12 +129,12 @@ std::string ConvDirectNaiveConvKernelName(const ProblemDescription& problem)
 #ifndef NDEBUG // enable in debug mode only
     if(env::enabled(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_USE_PACKED_KERNELS))
     {
-        kernel_name << "naive_conv_packed_";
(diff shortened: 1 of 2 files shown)


src/solver/conv_direct_naive_conv.cpp
```diff
@@ -129,12 +129,12 @@ std::string ConvDirectNaiveConvKernelName(const ProblemDescription& problem)
 #ifndef NDEBUG // enable in debug mode only
     if(env::enabled(MIOPEN_DEBUG_CONV_DIRECT_NAIVE_USE_PACKED_KERNELS))
     {
-        kernel_name << "naive_conv_packed_";
+        kernel_name << "naive_conv_ab_packed_";
```
atamazov, line 132: [Notice]  Just in case: "packed" kernels are not renamed in this PR (yet). Of course, I see that this is draft..

bghimireamd: @atamazov renamed both kernel names

atamazov (approved):
LGTM

atamazov:
@bghimireamd @junliume Please check if this resolves the issue and update the topmost comment. Thanks.

iq136boy (approved):

junliume (approved):

PR #3068, opened 2024-06-19
Fix MIOpenDriver gemm with column ordering

BrianHarrisonAMD:
Running MIOpenDriver gemm with column ordering specified was generating incorrect gemm_descriptions which would fail. This change sets the leading dimensions properly for column ordering to allow for -C 1 argument to work correctly.

For example the following would fail previously:

./bin/MIOpenDriver gemm -C 1 -m 256 -n 512 -k 1024 -i 1 -V 1 

  what():  rocBlas error encountered
Aborted (core dumped)


amberhassaan (approved):
looks good.

junliume (approved):

PR #3069, opened 2024-06-19
Fix "Crunching the latest data, just for you. Hang tight…"

averinevg:
Contributors page get stuck with "Crunching the latest data, just for you. Hang tight…". This happens due to large file sizes.


junliume (approved):
@averinevg how do I trigger these warnings? I have no issue with the fix while am interested in the context.

PR #3072, opened 2024-06-24
[ROCm 6.2] UAI and Windows backport

apwojcik:


src/comgr.cpp
```diff
@@ -241,15 +240,15 @@ namespace hip {
 #if PCH_IS_SUPPORTED
 static bool IsPchEnabled()
 {
-    return !miopen::IsDisabled(MIOPEN_ENV(MIOPEN_DEBUG_COMGR_HIP_PCH_ENFORCE));
+    static bool IsPchEnabled() { return !env::disabled(MIOPEN_DEBUG_COMGR_HIP_PCH_ENFORCE); }
```
junliume: @apwojcik we have a build issue here. I guess it is just a merge conflict issue hence two lines above :)
```
/MIOpen/src/comgr.cpp:243:32: error: function definition is not allowed here
  243 |     static bool IsPchEnabled() { return !env::disabled(MIOPEN_DEBUG_COMGR_HIP_PCH_ENFORCE); }
      |                                ^
1 error generated when compiling for gfx1030.

```

junliume: [Resolved]

junliume (approved):

PR #3073, opened 2024-06-24
C API Mha Backward Test (raw code)

Vsevolod1983:
1. Mha Forward test refactored to get common code base for Mha Forward and Mha Backward.
2. Mha Backward Graph creation
3. Cpu verification run

NOTICE: 
This backward test is only partly tested because functionality itself (graph matching) is not fully ready yet. 
I would like to have this code reviewed and merged, so guys who are developing functionality might use it in their development.

When graph matching is ready I will use another PR to apply all fixes required on this test.

### DIFF
--- a/test/gtest/graphapi_capi_mha_backward.cpp
+++ b/test/gtest/graphapi_capi_mha_backward.cpp
@@ -1,0 +1,184 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include "graphapi_capi_mha_common.hpp"
+
+using namespace capi_test_mha_common;
+
+template <typename T>
+class MhaBackwardTest : public MhaCommonTest
+{
+    static_assert(std::is_same_v<T, float> || std::is_same_v<T, float8>);
+    using dO_T = std::conditional_t<std::is_same_v<T, float>, float, bfloat8>;
+
+protected:
+    virtual void MakeRealTensorsAndFillData(miopen::Handle& handle) override
+    {
+        auto q = test::cpu::GenScaledTensorBackward<T>(n, h, s, d);
+        auto k = test::cpu::GenScaledTensorBackward<T>(n, h, s, d);
+        auto v = test::cpu::GenScaledTensorBackward<T>(n, h, s, d);
+
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaQ, false, test_n, test_h, test_s, test_d).
+            InitAndWriteToGPU(handle, std::move(q.mTensor));
+
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaK, false, test_n, test_h, test_s, test_d).
+            InitAndWriteToGPU(handle, std::move(k.mTensor));
+
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaV, false, test_n, test_h, test_s, test_d).
+            InitAndWriteToGPU(handle, std::move(v.mTensor));
+
+        float sScale = 1.0f;
+        float sDescale = 1.0f;
+
+        float oScale = 1.0f;
+        float oDescale = 1.0f;
+
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaDescaleQ).InitAndWriteToGPU(q.mDescale);
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaDescaleK).InitAndWriteToGPU(k.mDescale);
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaDescaleV).InitAndWriteToGPU(v.mDescale);
+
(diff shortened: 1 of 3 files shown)


test/gtest/graphapi_capi_mha_backward.cpp
```diff
@@ -0,0 +1,430 @@
...
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaAmaxDQ);
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaAmaxDK);
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaAmaxDV);
+        MakeAndAddRealTensorDescriptor(miopenTensorMhaAmaxDS);
+
+        dQDescRef = tensor<T>{m_testN, m_testH, m_testS, m_testD};
+        dKDescRef = tensor<T>{m_testN, m_testH, m_testS, m_testD};
+        dVDescRef = tensor<T>{m_testN, m_testH, m_testS, m_testD};
+
+        test::cpu::MultiHeadAttentionBackwardDataf8(
```
amberhassaan, line 149: function name suggests that you'll be creating tensors, but you're also calling cpu reference.

amberhassaan (approved):
Looks OK. I haven't verified the accuracy of the graph in detail.

test/gtest/graphapi_capi_mha_backward.cpp
Vsevolod1983: Yeah its convenient  to do it here, because this call uses a lot from previous calculations. Probably worth to consider slightly rename it

test/gtest/graphapi_capi_mha_backward.cpp
```diff
@@ -0,0 +1,430 @@
...
+        auto mm0 = MakeGapiVirtualTensorDesc(m_testN, m_testH, m_testS, m_testS);
+
+        MakeMatmul(m_realTensorMap[miopenTensorMhaQ]->m_gapiDesc,
+                   m_realTensorMap[miopenTensorMhaK]->m_gapiDesc,
+                   mm0);
+
+        // MakePointwise function will automatically create a tensor for output,
+        // if output is nullptr
+        DescriptorWrapperPtr pwS0;
+        MakePointwise(MIOPEN_POINTWISE_IDENTITY, mm0, nullptr, pwS0, false, m_attentionScale);
```
scerzh, line 194: `MIOPEN_POINTWISE_IDENTITY` replaces `MIOPEN_POINTWISE_MUL` here. I see the same approach for the forward path. This decision seems to have been discussed previously but I missed it.

@Vsevolod1983, was the intention here to make the attention scale optional?

test/gtest/graphapi_capi_mha_backward.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+        DescriptorWrapperPtr mult1;
+        DescriptorWrapperPtr pwS3, pwS4;
+        MakePointwise(MIOPEN_POINTWISE_MUL, mult0, rnd, mult1);
+
+        // _TODO there is a 1/(1-p) probability on the picture
+        MakePointwise(MIOPEN_POINTWISE_MUL,
+                      mult1,
+                      m_realTensorMap[miopenTensorMhaDropoutProbability]->m_gapiDesc,
+                      pwS3);
```
scerzh, line 226: This part is completely missing in the MHA graph formed by CuDNN's frontend API: [CuDNN code](https://github.com/NVIDIA/cudnn-frontend/blob/98ca4e1941fe3263f128f74f10063a3ea35c7019/include/cudnn_frontend/node/sdpa_fp8_bwd.h#L226-L230)

It's quite interesting: CuDNN doesn't have RNG part.

test/gtest/graphapi_capi_mha_backward.cpp
```diff
@@ -0,0 +1,430 @@
...
+        MakePointwise(
+            MIOPEN_POINTWISE_MUL, mm2, m_realTensorMap[miopenTensorMhaDescaleDO]->m_gapiDesc, pwS7);
+        MakePointwise(
+            MIOPEN_POINTWISE_MUL, pwS7, m_realTensorMap[miopenTensorMhaDescaleV]->m_gapiDesc, pwS8);
+        MakePointwise(MIOPEN_POINTWISE_MUL, pwS8, rnd, pwS9);
+
+        MakePointwise(MIOPEN_POINTWISE_MUL,
+                      pwS9,
+                      m_realTensorMap[miopenTensorMhaDropoutProbability]->m_gapiDesc,
+                      pwS10);
```
scerzh, line 266: This part is also missing in CuDNN's version as in the comment above [CuDNN code](https://github.com/NVIDIA/cudnn-frontend/blob/98ca4e1941fe3263f128f74f10063a3ea35c7019/include/cudnn_frontend/node/sdpa_fp8_bwd.h#L261-L265)

test/gtest/graphapi_capi_mha_backward.cpp
```diff
@@ -0,0 +1,430 @@
...
+            MIOPEN_REDUCE_TENSOR_MAX, mult3, m_realTensorMap[miopenTensorMhaAmaxDS]->m_gapiDesc);
+        MakePointwise(MIOPEN_POINTWISE_MUL,
+                      mult3,
+                      m_realTensorMap[miopenTensorMhaScaleDS]->m_gapiDesc,
+                      pwS15);
+
+        ////////////////// Center-bottom, Right-bottom ////////////////////////////
+        ///////////////////////////////////////////////////////////////////////////
+        auto mm3 = MakeGapiVirtualTensorDesc(m_testN, m_testH, m_testS, m_testD);
+        MakeMatmul(pwS15, m_realTensorMap[miopenTensorMhaK]->m_gapiDesc, mm3);
```
scerzh, line 307: @Vsevolod1983 The first `matmul` with input `K` is above and was using `K` as if it is already transposed. This is the second one which takes not transposed `K` and if `K` is already transposed we should add a `reshape` node before `matmul` either here to untranspose `K` or before the first `matmul` to flip the assumption about transposing of `K` and to stick closer to the graph from the docs.

scerzh (approved):
Looks good. To parallel the work on MHA we need to merge the PR. The comments are addressing the expected follow up PR.

junliume (approved):

test/gtest/graphapi_capi_mha_backward.cpp
Vsevolod1983: @scerzh 
As far as I remember you cannot pass attention scale in MUL node as a float parameter. In identity node we use alpha param to pass attn scale

test/gtest/graphapi_capi_mha_backward.cpp
Vsevolod1983: Once we get reshape node merged I will add it. Probably worth to discuss everything though

test/gtest/graphapi_capi_mha_backward.cpp
Vsevolod1983: We decided to stick with the graph from the official cudnn web site for now, right ?

test/gtest/graphapi_capi_mha_backward.cpp
scerzh: I was asking because in CuDNN's version `attention scale` is an optional node.

Yes, `MUL` node only accepts tensors, and you have to create a tensor `[1, 1, 1, 1]` for a single float value. Broadcasting will propagate the value.

test/gtest/graphapi_capi_mha_backward.cpp
scerzh: right

test/gtest/graphapi_capi_mha_backward.cpp
Vsevolod1983: If I understand correctly Attention scale is a value not a tensor. Also, its a value in Find 2.0 interface. That's why it was decided to do it  We can discuss all these details on the meeting with Alexander and Amber

PR #3078, opened 2024-06-25
Implemented readonly-/ramdb preloading

DrizztDoUrden:
If convolution perf and find dbs have not been started loading yet, handle constructor starts a separate thread that does that. When a db is requested in some library call like find, instead of just creating a new instance it first checks if preload has started and waits for its result if needed.


src/include/miopen/db_preload.hpp
```diff
@@ -0,0 +1,85 @@
...
+    DbPreloadStates(const DbPreloadStates&) = delete;
+    auto operator=(const DbPreloadStates&) -> DbPreloadStates& = delete;
+    DbPreloadStates(DbPreloadStates&&)                         = delete;
+    auto operator=(DbPreloadStates&&) -> DbPreloadStates& = delete;
+};
+
+auto GetDbPreloadStates() -> DbPreloadStates&;
+
+template <class Db>
+inline auto GetPreloadedDb(const fs::path& path) -> std::unique_ptr<Db>
```
CAHEK7: Since `Db` can only be `RamDB` and `ReadOnlyRamDb`, I think it makes sense to move the function body into `cpp` file and explicitly instantiate this function for those two types.
It doesn't seem like a performance critical code or may get any benefits from inlining (except a bloating binary and longer compilation time).

src/handle_common.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include <miopen/readonlyramdb.hpp>
+
+#include <array>
+#include <mutex>
+
+namespace miopen {
+template <class Db>
+static void
+StartPreloadingDb(DbPreloadStates& states, DbKinds db_kind, const fs::path& path, bool is_system)
+{
```
CAHEK7: Personally, I prefer unnamed namespaces over static. At least it creates explicit section of the file-local functions, classes, etc., and creates a uniform way to declare file-local types (which is not possible usnig static).
```suggestion
namespace {
template <class Db>
void
StartPreloadingDb(DbPreloadStates& states, DbKinds db_kind, const fs::path& path, bool is_system)
{
...
}
}

namespace miopen {
```

src/handle_common.cpp
```diff
@@ -0,0 +1,124 @@
...
+    ExecutionContext ctx{this};
+
+    auto& states = GetDbPreloadStates();
+
+    if(states.started_loading)
+        return;
+
+    std::unique_lock<std::mutex> lock(states.mutex, std::defer_lock);
+
+    if(!lock.try_lock() || states.started_loading)
```
CAHEK7: Can we make `started_loading` atomic and swap the condition?

src/include/miopen/db_preload.hpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    auto& states = GetDbPreloadStates();
+
+    std::unique_lock<std::mutex> lock{states.mutex};
+
+    if(!states.started_loading)
+        return nullptr;
```
CAHEK7: See a comment about atomic `started_loading`.
We can return earlier before even try to lock the mutex. But I didn't check is it worth or not.

CAHEK7 (approved):

DrizztDoUrden:
@CAHEK7 I have reverted making the flag atomic and placing it before the mutex because it looses its original purpose xD

BradPepersAMD:

PR #3081, opened 2024-06-26
[Workaround] fix mem fault caused by GemmBwdRest solver by using 64bit index

bghimireamd:
Currenly `GemmFwdRest`, `GemmBwdRest` and `GemmWrwUniversal` does not support tensor of size higher then 2GB (limited by its `int` offset) plus they seem to be slow when tensor size are bigger. This PR to make these solver extend beyond its pervious limit of tensor size greater then 2GB [issue](https://github.com/ROCm/MIOpen/issues/3076). Performance improvement PR for large tensor size will come soon.


src/solver/gemm.cpp
```diff
@@ -863,6 +863,13 @@ bool GemmFwdRest::IsApplicable(const ExecutionContext& context,
                                const ProblemDescription& problem) const
 {
 #if MIOPEN_USE_GEMM
+    const size_t two_gb_memory = 2ULL * 1024 * 1024 * 1024;
+    if((problem.GetInSize() >= two_gb_memory) || (problem.GetWeightsSize() >= two_gb_memory) ||
+       (problem.GetOutSize() >= two_gb_memory))
+    {
+        return false;
+    }
+
```
JehandadKhan: Please add this to the base class instead of the derived classes. I see WRW missing from this PR as well

JehandadKhan (changes requested):

atamazov:
> Currenly `GemmFwdRest` and `GemmBwdRest` does not support tensor of size higher then 2GB (limited by its `int` offset) plus they seem to be slow when tensor size are bigger.

But actually they do. This 2GiB limitation is an overkill that may lead to some unwanted side effects. The conditions are more complex.

bghimireamd:
> > Currenly `GemmFwdRest` and `GemmBwdRest` does not support tensor of size higher then 2GB (limited by its `int` offset) plus they seem to be slow when tensor size are bigger.
> 
> But actually they do. This 2GiB limitation is an overkill that may lead to some unwanted side effects. The conditions are more complex.

but if tensor size is bigger then 2GB it would cause Memory access fault because of the limitation caused by [int.](https://github.com/ROCm/MIOpen/blob/2de28ab6870fede26536fd9fa0fde51d2d5542e5/src/ocl/utilocl.cpp#L420) 

As per your comment [here](https://github.com/ROCm/MIOpen/issues/3076#issuecomment-2192604372), we decided to take the quick fix route, which would be faster than the partial fix, let alone a full-blown fix.

(phase 1) quick fix (current PR) <  (phase 2) partial fix (just change the offset from int to long) <  (phase 3) full blown fix (take care of performance)

src/solver/gemm.cpp
bghimireamd: only `GemmFwdRest`, `GemmBwdRest` and  `GemmWrwUniversal` use the `int` limited offset [`Im2ColGPU`](https://github.com/ROCm/MIOpen/blob/develop/src/ocl/utilocl.cpp#L43) and [`Col2ImGPU`](https://github.com/ROCm/MIOpen/blob/develop/src/ocl/utilocl.cpp#L43). So, I avoided placing the check directly in base class, since it might restrict other solvers.

junliume:
@atamazov and @JehandadKhan this PR is a quick workaround, and we will need to implement some full fix in mainline later.

However, @cderb this PR has triggered `Dbsync gfx90a` failure which is expected

junliume:
@bghimireamd could you resolve the Dbsync gfx90a failure issue with @cderb ? Thanks!

atamazov:
@bghimireamd @junliume You warned. I recommend limiting this W/A to 2D only at least. I'll try to create some alternative W/A tomorrow.

atamazov:
@bghimireamd @junliume I am still working on the full-fledged fix; hope to roll it out today.

bghimireamd:
closed since solved by #3099

PR #3082, opened 2024-06-26
[MHA] Implement MIOPEN_BACKEND_OPERATION_RESHAPE_DESCRIPTOR

scerzh:
MIOPEN_BACKEND_OPERATION_RESHAPE_DESCRIPTOR appears in MHA backward path graph so we are adding its implementation.

C API -> BackendOperationReshapeDescriptor (polymorphic) -> OperationReshapeBuilder -> OperationReshape

BackendOperationReshapeDescriptor, a polymorphic descriptor class, is utilized by C-API functions to handle create/set/finalize/get/execute requests.

OperationReshape class represents a reshape node in an operation graph.


src/graphapi/reshape.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+OperationReshapeBuilder& OperationReshapeBuilder::setY(Tensor* y)
+{
+    mOperationReshape.mY = checkPtr(y);
+    return *this;
+}
+
+OperationReshape OperationReshapeBuilder::build()
+{
+    if(mOperationReshape.mX != nullptr && mOperationReshape.mY != nullptr &&
```
Vsevolod1983: Do you have checks somewhere regarding sizes of input and output tensor? For example if input tensor is "n x h x 2 x 3"
then output tensor should be "n x h x 3 x 2".

@CAHEK7  @amberhassaan what do you think guys. Would be nice to have, right ?

Vsevolod1983 (changes requested):

src/graphapi/reshape.cpp
amberhassaan: yes, as long as we only support `transpose` in `reshape`.

scerzh: I was thinking about the condition here. Are we supporting a transpose only? Are we restricting it to 4-dimensional tensors? And are we restricting it for the last 2 dimensions?

Vsevolod1983: In MatMul node,  you guys, restricted minimum number of dimensions to 2, which probably makes sense. 
So may do in a pretty same way. But, yes, if you have a tensor with dims [n,h,s,d] then transposed version of this tensor would be [n,h,d,s] in my understanding.
This is in case of Transpose operation.

Vsevolod1983 (approved):

src/graphapi/reshape.cpp
amberhassaan: In MIOpen we support `NCHW` to `NHWC` transpose, so that suggests that `transpose` can be more general than that. Similarly, `reshape` can do things beyond `transpose`, e.g., treat a `[n, h, w, c]` tensor as `n, h*w, c, 1` tensor. So the question then becomes: should we support `reshape` in the long run or is it a temporary thing for BWD MHA?

Vsevolod1983: I believe we can merge it and enhance later if needed.

src/include/miopen/graphapi/reshape.hpp
```diff
@@ -0,0 +1,95 @@
...
+                      void* arrayOfElements) override;
+    void finalize() override;
+    void getAttribute(miopenBackendAttributeName_t attributeName,
+                      miopenBackendAttributeType_t attributeType,
+                      int64_t requestedElementCount,
+                      int64_t* elementCount,
+                      void* arrayOfElements) override;
+    OpNode* getOperation() override;
+
+    const OperationReshape* getOperationReduction() const { return &mOperationReshape; }
```
amberhassaan: typo here: Says `OperationReduction`.

src/graphapi/reshape.cpp
scerzh: I'm going to restrict `reshape` node to `transpose` for the last 2 dimensions. But the check if a `reshape` node is a `transpose` node should nevertheless be repeated in graph matching/extracting because `reshape` descriptor might be extended in the future.

amberhassaan (commented):
one important typo fix.

src/include/miopen/graphapi/reshape.hpp
scerzh: oh yes, thanks

src/graphapi/reshape.cpp
amberhassaan: @scerzh : can you please introduce an enum that specifies whether the reshape operation is `reshape` or `transpose`? Or do we guess from `tensor dims` (per cuDNN)?

scerzh: The user provides only dims so we should guess from them. But still we can add an enum property which in any case will be calculated during `finalize` and also it will avoid code duplication. I think it's a good idea.

junliume (approved):

PR #3084, opened 2024-06-26
KernelTuningNet Integration for CK FWD and WRW solvers

msaudulhassan:
This PR is based on https://github.com/ROCm/MIOpen/pull/3051 and https://github.com/ROCm/MIOpen/pull/3062. https://github.com/ROCm/MIOpen/pull/3051 adds KernelTuningNet (KTN) for both CK FWD and CK WRW solvers, however, the KTN for CK WRW is buggy. https://github.com/ROCm/MIOpen/pull/3062 adds the correct KTN for CK WRW solvers.

This PR takes the CK FWD from https://github.com/ROCm/MIOpen/pull/3051 and CK WRW from https://github.com/ROCm/MIOpen/pull/3062


src/include/miopen/solver/wrw_implicitgemm_ck_util.hpp
```diff
@@ -0,0 +1,950 @@
+/*******************************************************************************
```
amberhassaan, line 1: why are we creating a new copy of a file?

src/conv/heuristics/ai_heuristics.cpp
```diff
@@ -628,9 +629,14 @@ bool ModelSetParams(const std::string& arch,
     case miopen::conv::Direction::BackwardWeights: dir = "wrw"; break;
     default: return false;
     }
-    std::size_t num_tuning_params = model->metadata.num_tuning_params[dir];
-    for(std::size_t i = 0; i < num_tuning_params; ++i)
+
+    std::size_t i                 = 0;
+    std::size_t num_tuning_params = 1;
```
amberhassaan: you can still use a `for` loop here. e.g. 
```
for (size_t i = 0, num_tuning_params = 1; i < num_tuning_params; ++i) {
```
`for` is preferred in terms of avoiding human errors and has better readability in general. The `do-while` has `i` and `num_tuning_params` escaping into the outer scope.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -219,15 +219,14 @@ static std::vector<std::string> GetKernelAsTokens(const std::string& kernel)
     return tokens;
 }
 
-void PerformanceConfigHipImplicitGemmGroupFwdXdlops::InitHeuristicKernelIDs()
+void PerformanceConfigHipImplicitGemmGroupFwdXdlops::InitHeuristicKernelIDs(const std::string& type)
```
amberhassaan, line 222: Please add a comment (doxygen style) as to what `type` signifies.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -283,26 +291,25 @@ GetFeatures(const ProblemDescription& problem, std::size_t num_cu, const std::st
         features[17] = num_cu;
         return features;
     }
-    std::size_t n = 18;
+    std::size_t n = 17;
```
amberhassaan, line 294: lots of magic numbers here with not documentation.

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
```diff
@@ -283,26 +291,25 @@ GetFeatures(const ProblemDescription& problem, std::size_t num_cu, const std::st
...
-    features[9 * n + 9]   = problem.GetPadH();
-    features[10 * n + 10] = problem.GetPadW();
-    features[11 * n + 11] = problem.GetKernelStrideH();
-    features[12 * n + 12] = problem.GetKernelStrideW();
-    features[13 * n + 13] = problem.GetDilationH();
-    features[14 * n + 14] = problem.GetDilationW();
-    features[15 * n + 15] = problem.GetBatchSize();
-    features[16 * n + 16] = problem.GetInDataType() == miopenFloat ? 2.0 : 1.0;
-    features[17 * n + 17] = problem.GetGroupCount();
+    features[0]           = problem.GetInChannels();
```
amberhassaan, line 296: why are we populating only the diagonal entries in an `n*n` matrix?

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
```diff
@@ -400,15 +526,26 @@ bool PerformanceConfigHipImplicitGemmGroupWrwXdlops::SetNextValue(const ProblemD
         assert(!valid_kernels.empty());
         return true;
     }
-    if((index + 1) < valid_kernels.size())
+    do
```
amberhassaan, line 534: not clear as to why you need `do..while(false)` when just adding `{}` suffices.

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
```diff
@@ -250,22 +308,63 @@ static std::vector<std::string> GetKernelAsTokens(const std::string& kernel)
...
+            if(!valid_kernels.empty())
+                return true;
+            return false;
+        }
+        if(idx > 0)
+            idx--;
+        if(idx >= 12)
+            idx += 2;
+        if(((idx == 15 && (valid_kernels[heuristic_indexes[0]].find(
+                               "DeviceGroupedConvBwdWeight_Xdl_CShuffle") != std::string::npos)) ||
```
amberhassaan: if `CK` decides to change the name of their kernels, we have to come find all the locations where we hardcoded it.

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
```diff
@@ -212,12 +227,55 @@ struct CKArgs
 };
 } // namespace
 
+template <int L, int H>
```
amberhassaan, line 230: all of these functions are strange. They don't quite (or just) do what the name suggests. I'd like these to be re-designed. Seems like there's a range (1, 128) where you either iterate linearly and loop around to `1` or you iterate in powers of `2`.

amberhassaan (changes requested):
Blocking because we are duplicating a large file (950 lines) called `implicitgemm_ck_util.h`.

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
JehandadKhan: @amberhassaan  It doesnt, notice the `break` statements in there. Its basically a bunch of `goto` but since `use of goto considered harmful` they use the `do ... while(false)` construct

amberhassaan: I feel more sad :(.

amberhassaan: Here's an alt version:
```
if (!NextTwoPower...) {
    ...
} else if (!NextLinear..) {
    ...
} else {
   return false;
}
```

src/solver/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp
Dmantri98: That's the embedding method I created. This diagonal matrix scales embedding vectors (a representation for each convolution parameter) with it's value.

src/solver/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp
Dmantri98: I'll have it check the size of the kernels within valid_kernel instead.

iq136boy:
Close this PR. Issue was solved in [PR#3091](https://github.com/ROCm/MIOpen/pull/3091)

PR #3088, opened 2024-06-28
Dropout kernel OpenCL to HIP + gtest

sgundabo:
Initial attempt at translating the Dropout OpenCL Kernel to HIP with a GTEST, with hardcoded PRNG matrices replaced with rocrand function calls.


test/gtest/dropout_ocl_hip.cpp
```diff
@@ -0,0 +1,825 @@
...
+#include <algorithm>
+#include <iostream>
+
+#include "get_handle.hpp"
+#include "random.hpp"
+#include "driver.hpp"
+#include "dropout_util.hpp"
+
+template <typename T>
+inline void SquashPairedTensor(const std::vector<T> x_len,
```
amberhassaan: use `const ref`, i.e. `const std::vector<T>&`.

src/kernels/MIOpenDropoutHIP.cpp
```diff
@@ -0,0 +1,374 @@
...
+        size_t i4    = gid % dim4;
+        size_t i4_rd = i4 / RD_BLCK;
+
+        size_t x_idx = i0 * in_str0 + i1 * in_str1 + i2 * in_str2 + i3 * in_str3 +
+                       i4_rd * RD_BLCK; // Calculate the index of the output tensor
+        size_t y_idx = i0 * out_str0 + i1 * out_str1 + i2 * out_str2 + i3 * out_str3 +
+                       i4_rd * RD_BLCK; // Calculate the index of the input tensor
+
+        // Read RD_BLCK number of _FLOAT data from y and store it in dat_blk
+        *(reinterpret_cast<T*>(dat_blk)) = *(reinterpret_cast<const T*>(y + out_offset + y_idx));
```
amberhassaan, line 297: why do you need this? Seems like the `data_blk` array is of type `F` and so is the `x` and `y` inputs.

src/kernels/MIOpenDropoutHIP.cpp
```diff
@@ -0,0 +1,374 @@
...
+#ifndef MIOPEN_USE_FP32
+#define MIOPEN_USE_FP32 0
+#endif
+
+#ifndef MIOPEN_USE_FP16
+#define MIOPEN_USE_FP16 0
+#endif
+
+#if MIOPEN_USE_FP16 == 1
+#define _FLOAT half
```
amberhassaan: Might be useful to replace these defines with kernels of different types. See the last lines in `src/kernels/gpu_reference_kernel/naive_conv.cpp`. 
The idea being that we instantiate the templated implementation with all supported data types and compile them all at once because they are in a single file.

src/kernels/MIOpenDropoutHIP.cpp
```diff
@@ -0,0 +1,374 @@
...
+ * The mask is either generated and saved in the reserved space, or an existing mask is loaded into
+ * registers from the reserved space. The loaded mask is applied to the data in the registers. The
+ * processed data is then written to the output. Reading and writing of input data can be performed
+ * as vectorized operations, with the extent of vectorization determined by `#RD_BLOCK`.
+ */
+template <typename F, typename T, typename B, bool MASK = false, bool RSVSP = false>
+__forceinline__ __device__ void dropoutfw(const rocrand_state_xorwow* state,
+                                          float dropout,
+                                          float scale,
+                                          int dim1,
```
amberhassaan, line 108: `dims` numbering should also start at `0` for consistency.

amberhassaan: You can also pass `dims` and `strides` as `int dim[N]` where `N` is a template parameter. Then we can handle `ncdhw`, i.e. 3D spatial tensors as well.

src/kernels/MIOpenDropoutHIP.cpp
```diff
@@ -0,0 +1,374 @@
...
+    uchar is_kept[RD_BLCK]; // Register space to store the mask for the dropout
+
+    size_t sid = threadIdx.x + blockIdx.x * blockDim.x;
+    rocrand_state_xorwow cur_state; // Read the state of the current thread
+    cur_state = state[sid];
+
+    for(size_t gid = threadIdx.x + blockIdx.x * blockDim.x; gid < total_work;
+        gid += blockDim.x * gridDim.x)
+    {
+        size_t i0    = gid / (dim1 * dim2 * dim3 * dim4);
```
amberhassaan: I've seen this type of computation (converting `gid` to a tensor index) being done a few times. We should make a note for now and create a helper function or a class for this kind of conversion.

src/kernels/MIOpenDropoutHIP.cpp
```diff
@@ -0,0 +1,374 @@
...
+        }
+        else
+        { // If MASK is enabled then read the mask from the reserveSpace
+            *(reinterpret_cast<B*>(is_kept)) = *(reinterpret_cast<const B*>(
+                reserveSpace + rsvsp_offset + gid - i4 + i4_rd * RD_BLCK));
+        }
+        // Apply the mask to the data and scale it with the scale factor.
+        for(size_t i = 0; i < RD_BLCK; ++i)
+        {
+            dat_blk[i] = static_cast<bool>(is_kept[i]) ? dat_blk[i] * (F)scale : (F)0;
```
amberhassaan: avoid `C` style casts always.

src/kernels/MIOpenDropoutHIP.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+/**
+ * @brief Applies dropout backward operation to the input data.
+ * The mask is either generated or an existing mask is loaded into
+ * registers from the reserveSpace. The loaded mask is applied to the data in the registers. The
+ * processed data is then written to the output. Reading and writing of input data can be performed
+ * as vectorized operations, with the extent of vectorization determined by #RD_BLOCK.
+ */
+template <typename F, typename T, typename B, bool PRNG = false>
+__forceinline__ __device__ void dropoutbw(const rocrand_state_xorwow* state,
```
amberhassaan, line 251: seems like `dropoutfw` and `dropoutbw` are virtually the same function and all that matters is swapping pointers and strides?

src/kernels/MIOpenDropoutHIP.cpp
```diff
@@ -0,0 +1,374 @@
...
+#endif
+
+#if MIOPEN_USE_FP16 == 1
+#define _FLOAT half
+#endif
+#if MIOPEN_USE_FP32 == 1
+#define _FLOAT float
+#endif
+
+#ifndef RUN_FORWARD
```
amberhassaan, line 49: I think we can remove this define?

src/kernels/MIOpenDropoutHIP.cpp
```diff
@@ -0,0 +1,374 @@
...
+#define RUN_FORWARD 0
+#endif
+
+#ifndef RUN_INIT_PRNG
```
amberhassaan, line 53: same, can be removed and the host side can choose to launch the `Init` kernel or not.

test/gtest/dropout_ocl_hip.cpp
amberhassaan: need to document this code so I can understand what's going on. Please add comments.

test/gtest/dropout_ocl_hip.cpp
```diff
@@ -0,0 +1,825 @@
...
+
+    if(!std::equal(in_len.begin(), in_len.end(), out_len.begin()))
+    {
+        MIOPEN_THROW("Input/Output tensor lengths do not match");
+    }
+}
+
+void InitPRNGState(miopen::Handle& handle,
+                   const miopen::DropoutDescriptor& DropoutDesc,
+                   bool use_hip = false)
```
amberhassaan: avoid `default` parameters as Aleksandr suggested.

test/gtest/dropout_ocl_hip.cpp
```diff
@@ -0,0 +1,825 @@
...
+    handle.AddKernel(kernel_name, network_config, program_name, kernel_name, vld, vgd, params)(
+        pstates,
+        dropout,
+        amp_scale,
+        static_cast<int>(in_len[1]),
+        static_cast<int>(in_len[2]),
+        static_cast<int>(in_len[3]),
+        static_cast<int>(in_len[4]),
+        dy,
+        static_cast<int>(out_str[0]),
```
amberhassaan, line 576: take strides as `size_t` because we might need to support `large tensors` > 2GB.

test/gtest/dropout_ocl_hip.cpp
```diff
@@ -0,0 +1,825 @@
...
+                    DropoutDesc,
+                    use_hip);
+
+    din_gpu.data = handle.Read<T>(din_dev, din.data.size());
+    return din_gpu;
+}
+
+struct DropoutTestCase
+{
+    bool mask_flag;
```
amberhassaan: why is one a `bool` and the other an `int`?

amberhassaan (commented):
Initial comments.

src/kernels/MIOpenDropoutHIP.cpp
CAHEK7: But if I don't need FP16 implementation at all? The whole compiled binary blob gets into the kernel cache for only the one specific kernel depicted by the network config. Network config says "FP32", so why we should spend time compiling "FP16" and waste space putting this FP16 stuff into the kernel cache.

src/kernels/MIOpenDropoutHIP.cpp
CAHEK7: See https://github.com/ROCm/MIOpen/pull/3088/files/9888ac8efe7cc7ea3213c93c09edb1ebd57a0fe1#r1679668996

src/kernels/MIOpenDropoutHIP.cpp
CAHEK7: See
https://github.com/ROCm/MIOpen/pull/3088/files/9888ac8efe7cc7ea3213c93c09edb1ebd57a0fe1#r1679668996

src/kernels/MIOpenDropoutHIP.cpp
sgundabo: I think the dims are numbered correctly, it is a flattened 5D tensor.

src/kernels/MIOpenDropoutHIP.cpp
sgundabo: Yes, they are. Well, I personally think having more #defines or if constexprs within the kernel ruins the readability. Does this cause any other issues other than MIOpen needing extra space to inline the kernel src?

src/kernels/MIOpenDropoutHIP.cpp
sgundabo: Well, those are necessary for vectorized reads/writes, if and when supported by the hardware.

sgundabo:
@CAHEK7 @amberhassaan 

Please find the profiling results attached below.

[DropoutPerf.csv](https://github.com/user-attachments/files/16257288/DropoutPerf.csv)
DropoutPerf_FP32
DropoutPerf_FP16

src/kernels/MIOpenDropoutHIP.cpp
CAHEK7: It depends. Probably you don't even need ifdefs and just swap strides and tensors on a host side.

test/gtest/dropout_ocl_hip.cpp
CAHEK7: It seems `size_t` based indexing introduces significant perf degradation (up to 2 times).
The way how to do it properly has to be evaluated and thoroughly tested. I guess it has to be a separated task.

CAHEK7:
> @CAHEK7 @amberhassaan
> 
> Please find the profiling results attached below.
> 
@sgundabo 
Just for a reference - what king of gpu did you use to get those results?
And I guess we need some bigger tensors.

src/kernels/MIOpenDropoutHIP.cpp
amberhassaan: @sgundabo : when 90% of the code is the same, we should look to remove duplication. Sure, slight more effort to refactor and read, but much better in the long term as you are not maintaining two copies of the code.

src/kernels/MIOpenDropoutHIP.cpp
amberhassaan: @CAHEK7 : My original thought was that compilation is expensive and if we compile all the variants of a kernel (in terms of data types etc) then we can amortize the cost somewhat. What you wrote means that we need to enhance the code object extraction so that we can pick specific kernel blobs out of the bigger binary blob and put those in the kernel cache.

src/kernels/MIOpenDropoutHIP.cpp
amberhassaan: I still don't get it. How are `T` and `F` different? It seems like `T` is being passed as the tensor's data type?

sgundabo: So here is an excerpt which should help explain things.
`size_t RD_BLCK = /* (in_len[4] % 4 == 0) ? 4 : */ (in_len[2] % 2 == 0) ? 2 : 1;
const std::string READ_DAT_TYPE = RD_BLCK == 1 ? data_type : data_type + std::to_string(RD_BLCK);`
RD_BLCK can be of size 1,2 or 4. That will make READ_DAT_TYPE float/float2/float4 and this choice is based on the input tensor length.

test/gtest/dropout_ocl_hip.cpp
sgundabo: This code was taken from src/ocl/dropooutocl.cpp. I just wanted my modifications to be contained within the gtest so I did not modify the original source to make the function visible in my gtest.

sgundabo:
> @CAHEK7 @amberhassaan
> 
> Please find the profiling results attached below.
> 
> [DropoutPerf.csv](https://github.com/user-attachments/files/16257288/DropoutPerf.csv) DropoutPerf_FP32 DropoutPerf_FP16

These profiling results are from gfx90a.

sgundabo:
> > @CAHEK7 @amberhassaan
> > Please find the profiling results attached below.
> 
> @sgundabo Just for a reference - what king of gpu did you use to get those results? And I guess we need some bigger tensors.

These results are from a gfx90a.

src/kernels/MIOpenDropoutHIP.cpp
CAHEK7: @amberhassaan, right now there is no way do distinguish two binary blobs except a network config. Then it tries to call some particular kernel (using its function name). If we want to use one blob for a multiple types, we just have to use the same network config for blob matching and different kernel names for different types. It is possible, but it requires some changes in the host calling code.
This particular PR is not relate to that sort of optimization or improvement and it is focused only on OCL->HIP transition with minimal changes and we can't fit aaaall the improvements into the one PR. I would suggest to create follow-up ticket for that task, because it requires some research and better understanding of the library usecases - it's not obvious what would be better for the user - longer compilation and bigger blob size for all the types or faster compilation and smaller blobs for each actually used type.

I have no ideas what could be the difference, can we ever notice it and all that stuff, so lets limit the changes by the kernel only.

src/kernels/MIOpenDropoutHIP.cpp
CAHEK7: @amberhassaan 
As far as I can see for this particular kernel, it is used twice - in forward and backward kernels. And we are going to get rid of separated backward kernel.

As for the overall code duplication across the algorithms, I'm afraid that all the algorithms have some small tweaks and differences, and indexing approaches, so it can be hard to generalize a helper for it.

I would suggest to create another issue to evaluate possibility for the helper function across all the algorithms and most probably it must be done when we rewrite all the OCL kernels.

test/gtest/dropout_ocl_hip.cpp
```diff
@@ -0,0 +1,1068 @@
...
+            maxTotalSize = 256; // 128MB L3
+        }
+        else if(miopen::StartsWith(handle.GetDeviceName(), "gfx103"))
+        {
+            maxTotalSize = 128; // 128MB L3
+        }
+        else
+        {
+            maxTotalSize = 4; // 2MB L2, default case.
+        }
```
CAHEK7: Cache size is in megabytes, not elements.
So proper `maxTotalSize` should be like this: `maxTotalSize *= 1024ull * 1024ull;` in megabytes or `maxTotalSize *= 1024ull * 1024ull / sizeof(DataType);` in elements.

Also I would recommend making more clear comment:
```
maxTotalSize = 16; // twice than 8MB L2
```

sgundabo:
### PerfData
[DropoutPerf_large_5.csv](https://github.com/user-attachments/files/16386909/DropoutPerf_large_5.csv)
[DropoutPerf_large_4.csv](https://github.com/user-attachments/files/16386919/DropoutPerf_large_4.csv)
[DropoutPerf_large_3.csv](https://github.com/user-attachments/files/16386922/DropoutPerf_large_3.csv)
[DropoutPerf_large_2.csv](https://github.com/user-attachments/files/16386923/DropoutPerf_large_2.csv)
[DropoutPerf_large_1.csv](https://github.com/user-attachments/files/16386924/DropoutPerf_large_1.csv)

HW tested: gfx90a

### FP32 Perf
DropoutPerf_large_FP32

### FP16 Perf
DropoutPerf_large_FP16

src/kernels/MIOpenDropoutHIP.cpp
amberhassaan: I agree. It's just that I've seen this pattern in Moreh PRs recently and was wondering if there's a better abstraction that can be built.

src/kernels/MIOpenDropoutHIP.cpp
atamazov: > If we want to use one blob for a multiple types, we just have to use the same network config for blob matching

@CAHEK7 No. Blob is identified by filename and build options. The change of network config is not required.

atamazov: > Might be useful to replace these defines with kernels of different types.

@amberhassaan I suspect this does not worth it and recommend evaluating the effect first.

test/gtest/dropout_ocl_hip.cpp
atamazov: But what is needed to do right away is narrowing the applicability, correct?

sgundabo:
**Raw Perf data with detailed kernel information**
[DropoutPerf_FP16.zip](https://github.com/user-attachments/files/16434930/DropoutPerf_FP16.zip)
[DropoutPerf_FP32.zip](https://github.com/user-attachments/files/16434931/DropoutPerf_FP32.zip)

HW tested: gfx90a

**FP32 Perf**
DropoutPerfRaw_FP32

**FP16 Perf**
DropoutPerfRaw_FP16

test/gtest/dropout_ocl_hip.cpp
CAHEK7: > But what is needed to do right away is narrowing the applicability, correct?

OCL version uses 32bit indexes, it doesn't narrow anything.

And for sure, but we don't need `size_t` for small tensors. We can use 64bit indexing for the big ones, but it's a separated task.

CAHEK7 (approved):

sgundabo:
**RawPerf Data**
[DropoutPerf_smalltensors.zip](https://github.com/user-attachments/files/16463730/DropoutPerf_smalltensors.zip)

**Perf FP16**
DropoutPerf_smalltensors_FP16

**Perf FP32**
DropoutPerf_smalltensors_FP32

sgundabo:
**FP32 Perf Analysis**

**Boxplot dropout0.50 mask0**
Boxplot_min_exec_time_ratio

**Boxplot all dropouts and mask combinations**
Boxplot_all_combinations

src/kernels/MIOpenDropoutHIP.cpp
CAHEK7: I think we can do it later.

CAHEK7:
@junliume could you merge this one, as there are few other PRs depend on it.

junliume:
@sgundabo @CAHEK7 please feel free to ping me on any other channels. Also marking the urgency level of the PR would help for visibility. Thanks!

junliume (approved):

PR #3097, opened 2024-07-02
Backward MHA test using C++ API

amberhassaan:
Provides a C++ Graph API test for backward MHA. Does not execute the graph yet because of graph engine being still in development. A follow up PR will enable graph execution.


test/gtest/graphapi_mha_bwd.cpp
```diff
@@ -0,0 +1,371 @@
...
+            else if(k == "dO")
+            {
+                v.init(std::move(dO.mTensor));
+            }
+            else if(k == "DSCL_dO")
+            {
+                v.init(dO.mDescale);
+            }
+            else if(k == "SCL_O" || k == "SCL_S" || k == "DSCL_O" || k == "DSCL_S" ||
+                    k == "DSCL_S" || k == "ATN_SCL" || k == "SCL_dS" || k == "DSCL_dS" ||
```
junliume: `k == "DSCL_S"` seems to be duplicated, maybe a typo?

Vsevolod1983 (approved):
It needs to add transpose reshape nodes though in the upcoming PRs

scerzh (approved):
LGTM, we can fit the missing transpose nodes in my PR, which will come soon.

test/gtest/graphapi_mha_bwd.cpp
junliume: My last comment was "pending" likely due to outage:
> k == "DSCL_S" seems to be duplicated, maybe a typo?

junliume: k == "DSCL_S" seems to be duplicated, maybe a typo?

junliume:
@amberhassaan could you help to resolve the conflicts?

test/gtest/gtest_common.hpp
```diff
@@ -171,6 +171,15 @@ void invoke_with_params(Check&& check)
...
 
+/// repalcement of gtest's FAIL() because FAIL() returns void and messes up the
+/// return type deduction
+#define MIOPEN_LOG_FATAL(...)                                 \
+    do                                                        \
+    {                                                         \
+        MIOPEN_LOG(miopen::LoggingLevel::Fatal, __VA_ARGS__); \
+        std::abort();                                         \
+    } while(false);
+
```
CAHEK7:
```suggestion
#define MIOPEN_FRIENDLY_FAIL(MSG)                             \
    do                                                        \
    {                                                         \
        [&]() { FAIL() << MSG;}();                            \
    } while(false);
```
I don't think that `std::abort` is a good solution. Gtest's `FAIL()` does not abort and switches to the next test, but in this case it messes whole gtest functionality. I would say the same about C `assert`.

BTW, what is the difference between `assert` and `MIOPEN_LOG_FATAL`? Shall `assert` be replaced?

test/gtest/graphapi_mha_cpp_common.hpp
```diff
@@ -0,0 +1,444 @@
...
+            {
+                mCpuTensor = TensFlt{dims};
+            }
+            else if(dt == miopenInt64)
+            {
+                mCpuTensor = TensI64{dims};
+            }
+            else
+            {
+                MIOPEN_LOG_FATAL("Unsupported data type for tensor" << dt);
```
CAHEK7:
```suggestion
                FAIL() << "Unsupported data type for tensor" << dt;
```
Constructor does not care about return value.

test/gtest/graphapi_mha_cpp_common.hpp
```diff
@@ -0,0 +1,444 @@
...
+            mGraphBuilder->addNode(mAlloc.allocate(gr::OperationRngBuilder{}
+                                                       .setRng(rng_desc)
+                                                       .setSeed(in_tensors[0])
+                                                       .setOffset(in_tensors[1])
+                                                       .setOutput(out_tensors[0])
+                                                       .build()));
+        }
+        else
+        {
+            MIOPEN_LOG_FATAL("Unknown graph node type: " << name);
```
CAHEK7:
```suggestion
            FAIL() << "Unknown graph node type: " << name;
```
`addNode` returns `void`

test/gtest/graphapi_mha_bwd.cpp
```diff
@@ -0,0 +1,371 @@
...
+        std::vector<int64_t> nhs1  = {n, h, s, 1};
+        std::vector<int64_t> all1s = {1, 1, 1, 1};
+
+        MAKE_TENSOR_F(Q, nhsd, false);
+        MAKE_TENSOR_F(K, nhsd, false);
+        MAKE_TENSOR_F(V, nhsd, false);
+        MAKE_TENSOR_F(dO, nhsd, false);
+        MAKE_TENSOR_F(O, nhsd, false);
+
+        MAKE_TENSOR_F(T_MM_0, nhss, true);
```
CAHEK7, line 52: [R] If we can't live without that wrapper, I'd rather moved it to something like `graphapi_common.hpp` or stick to the raw API.
Seems it's kind of widely used case.

CAHEK7 (changes requested):
I'm good if that `MIOPEN_FATAL` request can be done in a follow-up PR, but since there are some upcoming changes for the merge conflicts, it would be better to fix it here.

junliume:
@amberhassaan can we follow up on this PR? Thanks!

test/gtest/gtest_common.hpp
amberhassaan: Good point. I did not know how `FAIL()` was different from `abort()`. We can replace asserts with `MIOPEN_FRIENDLY_FAIL()` I believe.

test/gtest/graphapi_mha_bwd.cpp
amberhassaan: It saves typing in these tests, so I will move it up in the future when the need arises.

test/gtest/graphapi_mha_bwd.cpp
amberhassaan: Fixed in next commit.

test/gtest/graphapi_mha_cpp_common.hpp
amberhassaan: FYI: Well, the constructor does care, as in I get a build error due to return value. I didn't bother investigating it further and just replaced it  with `MIOPEN_FRIENDLY_FAIL`.

CAHEK7 (approved):

amberhassaan:
@junliume : please merge asap. I'll be building on top of this work (modifying the same files) to enable MHA backward testing.

junliume:
@amberhassaan please feel free to ping me on any other channels :)
Also mark this PR as "blocker" or "high“ in urgency would also help.

Restarted CI, once passed we can surely merge asap. Thanks!

junliume (approved):

PR #3099, opened 2024-07-04
Fix #3076 - issues related to in_offset overflow in Col2Im (2D and 3D) and Im2Col (2D only)

atamazov:
Intended to resolve #3076. Fixes some issues in `GemmBwdRest`, `GemmFwdRest`, `GemmWrwUniversal`. 

- Switches 2D and 3D Col2Im kernels to _**unsigned**_ index computations math.
  - As a by-product, this resolves point 2 of https://github.com/ROCm/MIOpen/pull/1966/files#r1106928625
- Adds 64-bit index mode to 2D Col2Im
  - [Informative] Note that I do not see any perf drop related to switching to 64-bit index computations, so maybe we can simply drop 32-bit stuff and simplify the code (actually I see very small gain))). However right now let's keep 32-bit stuff to minimize changes (who knows)?
  - Passes in_offset as ulong instead of int to 2D Im2Col
- There is no find-db and KDB related impact because:
  - Modified kernels were renamed, so the updated library won't load existing kernels from persistent cache. Therefore, regeneration of KDBs is not required (but still desired to avoid performance hit, albeit very small).
  - Applicability of the solvers remain the same.

## 🌀 Examples of resolved issues:

Reproducible on 16GB cards.

- `./bin/MIOpenDriver convfp16 -n 16 -c 400 -H 224 -W 224 -k 1 -y 20 -x 20 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -F 2 -i 1 -V 1 -S GemmBwdRest`
  - ch_offset_max = 6855840000 (6.38 G)
  - ORG: Backward Convolution Data FAILED: 0.0681179 > 0.0082
  - FIX: Backward Convolution Data Verifies OK on GPU reference (0.000153848 < 0.0082)
- `./bin/MIOpenDriver convfp16 -n 3000 -c 1 -H 1024 -W 1024 -k 1 -y 3 -x 3 -F 1 -V 1 -i 1 -F 1 -V 1 -S GemmFwdRest`
  - in_offset_max = 3144679424 (2.92871 G)
  - ORG: Forward Convolution FAILED: 0.264809 > 0.0082
  - FIX: Forward Convolution Verifies OK on GPU reference (2.5259e-06 < 0.0082)
- `./bin/MIOpenDriver convfp16 -n 101 -c 1000 -H 160 -W 160 -k 1 -y 3 -x 3 -F 4 -S GemmWrwUniversal -i 1 -V 1`
[...]

PR #3100, opened 2024-07-04
turn journaling off 

bghimireamd:
* This is a stand by PR just in case if we need new KDB.
* PR #3081 was failing because I did not turned journaling off. This PR fixes that issue.


bghimireamd:
no needed since issue is addressed by #3099

bghimireamd:
no needed since issue is addressed by #3099

PR #2388, opened 2023-09-12
Fusion Find

DrizztDoUrden:
This PR adds find functionality from convolution code to fusion.

Progress:
1. Untangle find-related code from convolution: done
2. Reuse core find-related code: done.
3. Add find-db support: done except for find modes.

Future work:
1. Find modes support: this is extremely cheap to add after find db and allows to use hybrid mode and such. Can be done in a this PR if needed, but I would rather split it out.

Some notes:
It is impossible to completely unify the code between API call and solver method among primitives due to API difference outside of find 2.0, which was designed with that in mind.
This would be optimized a bit after #2146 gets merged due to the changes to how find works there.

PR #3105, opened 2024-07-10
Added missing weight sorting as WTI calculation in fusion

DrizztDoUrden:
This was missing since #2388. Reimplemented as WTI to use in the fallback. Numbers are preserved mostly as were before #2388 with some changes due to how WTI is used: they have been normalized to range of (0;1] and 0 values replaced by -2.

Thanks to @atamazov for noticing.


src/fusion.cpp
```diff
@@ -998,15 +998,22 @@ miopenStatus_t FusionPlanDescriptor::Compile(Handle& handle)
...
+                    const auto id  = solver::Id(solver.SolverDbId());
+                    const auto wti = solver.GetWti(ctx, fusion_problem);
+                    // Assume WTI == 1.0 (100%) is 10 ms.
+                    // Return negative values as is, avoid DIV/0.
+                    const auto time = wti <= 0.0f ? wti : (10.f / wti);
+                    sols.push_back({time, 0, id.Value(), miopenConvolutionAlgoDirect});
                 });
             }
 
             // override the normal find with immed mode with env var
```
averinevg, line 1010: This comment is confusing

src/solver/conv_bin_winoRxS_fused.cpp
```diff
@@ -275,7 +269,16 @@ ConvSolution ConvBinWinogradRxSFused::GetSolution(const FusionContext& context,
...
-
+float ConvBinWinogradRxSFused::GetWti(const FusionContext&, const FusionDescription& problem) const
+{
+    const auto conv_problem = problem.GetConvProblem(0, miopen::conv::Direction::Forward);
+    const auto x            = conv_problem.GetWeightsWidth();
+    const auto y            = conv_problem.GetWeightsHeight();
+    if(x == 3 && y == 3)
+        return 1.f;
+    else
+        return 5.f / 100.f;
```
averinevg: Magic numbers

src/solver/conv_ocl_dir2Dfwd_fused.cpp
```diff
@@ -228,6 +223,16 @@ ConvOclDirectFwdFused::GetSolution(const FusionContext& context,
...
 }
 
+float ConvOclDirectFwdFused::GetWti(const FusionContext&, const FusionDescription& problem) const
+{
+    const auto& desc = *problem.fusion_plan_desc;
+    const int bn_idx = GetOpIdx(desc.op_map, miopenFusionOpBatchNormInference);
+    if(bn_idx != -1)
+        return -2;
+    else
+        return 10.0f / 100.f;
```
averinevg: Magic numbers

src/solver/conv_winoRxS_fused.cpp
```diff
@@ -417,7 +409,18 @@ ConvSolution ConvBinWinogradRxSf2x3g1Fused::GetSolution(const FusionContext& con
...
+                                            const FusionDescription& problem) const
+{
+    const auto conv_problem = problem.GetConvProblem(0, miopen::conv::Direction::Forward);
+    const auto x            = conv_problem.GetWeightsWidth();
+    const auto y            = conv_problem.GetWeightsHeight();
 
+    if(x == 3 && y == 3)
+        return 1.f;
+    else
+        return 5.f / 100.f;
```
averinevg: Magic numbers

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
```diff
@@ -208,6 +207,11 @@ ConvBiasActivAsm1x1U::GetSolution(const FusionContext& context,
     return sol;
 }
 
+float ConvBiasActivAsm1x1U::GetWti(const FusionContext&, const FusionDescription&) const
+{
+    return 50.0f / 100.f;
```
averinevg: Magic numbers

src/fusion.cpp
DrizztDoUrden: That comments are copied from convolution.

src/solver/conv_ocl_dir2Dfwd_fused.cpp
DrizztDoUrden: They are copied from before #2388. Division is to normalize them to the range of acceptable WTI values.

src/fusion.cpp
averinevg: Yes, but `env::enabled(MIOPEN_DEBUG_FORCE_IMMED_MODE_FALLBACK)` is not commented out there. U2Y

src/solver/conv_ocl_dir2Dfwd_fused.cpp
averinevg: This function should probably be removed. The `ConvOclDirectFwd` does not have it.

src/solver/batchnorm/backward_per_activation_fused.cpp
```diff
@@ -233,6 +233,11 @@ ConvSolution BnBwdTrgActivationFused::GetSolution(const FusionContext& context,
     return result;
 }
 
+float BnBwdTrgActivationFused::GetWti(const FusionContext&, const FusionDescription&) const
+{
+    return -2;
+}
```
averinevg: This function can be removed, because it returns the default value

src/solver/batchnorm/forward_inference_fused.cpp
```diff
@@ -170,6 +170,11 @@ ConvSolution BnFwdInferActivationFused::GetSolution(const FusionContext& /*conte
     return result;
 }
 
+float BnFwdInferActivationFused::GetWti(const FusionContext&, const FusionDescription&) const
+{
+    return -2;
+}
```
averinevg: This function can be removed, because it returns the default value

src/solver/batchnorm/forward_per_activation_fused.cpp
```diff
@@ -242,6 +242,11 @@ ConvSolution BnFwdTrgActivationFused::GetSolution(const FusionContext& context,
     return result;
 }
 
+float BnFwdTrgActivationFused::GetWti(const FusionContext&, const FusionDescription&) const
+{
+    return -2;
+}
```
averinevg: This function can be removed, because it returns the default value

src/solver/conv_ck_igemm_fwd_bias_activ_fused.cpp
```diff
@@ -494,6 +494,11 @@ ConvSolution ConvCKIgemmFwdBiasActivFused::GetSolution(
 #endif
 }
 
+float ConvCKIgemmFwdBiasActivFused::GetWti(const FusionContext&, const FusionDescription&) const
+{
+    return -2;
+}
```
averinevg: This function can be removed, because it returns the default value

src/solver/conv_ck_igemm_fwd_bias_res_add_activ_fused.cpp
```diff
@@ -483,6 +483,12 @@ ConvSolution ConvCKIgemmFwdBiasResAddActivFused::GetSolution(
 #endif
 }
 
+float ConvCKIgemmFwdBiasResAddActivFused::GetWti(const FusionContext&,
+                                                 const FusionDescription&) const
+{
+    return -2;
+}
```
averinevg: This function can be removed, because it returns the default value

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
averinevg: This function should probably be removed. The `ConvAsm1x1U` does not have it.

src/solver/conv_bin_winoRxS_fused.cpp
averinevg: This function should probably be removed. The `ConvBinWinogradRxS` does not have it.

src/solver/conv_winoRxS_fused.cpp
averinevg: The `ConvBinWinogradRxSf2x3g1` has proper WTI calculation.

src/fusion.cpp
DrizztDoUrden: It's not immediate mode per se, so I have decided to not include that part. I can uncomment, but it is not very useful right now anyway.

src/solver/batchnorm/backward_per_activation_fused.cpp
DrizztDoUrden: Makes sense. Will remove default value ones.

src/solver/conv_bin_winoRxS_fused.cpp
DrizztDoUrden: Fusion is somewhat different and this is how they were selected before #2388. I would rather restore it to how it worked. Later it can be changed to any desired behavior. Otherwise performance would be different from what is expected (to better or worse). ConvWinoFuryRxSFused is exception due to someone explicitly deciding to include the implementation and it actually returning -2. Maybe I should also remove that, but it is somewhat irrelevant.

src/include/miopen/solver.hpp
```diff
@@ -108,6 +108,8 @@ struct SolverBase
     /// run-time parameters.
     virtual bool IsDynamic() const { return false; }
 
+    static constexpr float wti_cant_be_computed = -2;
```
averinevg:
```suggestion
    static constexpr float wti_unknown = -2.0f;
```

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
```diff
@@ -208,6 +207,15 @@ ConvBiasActivAsm1x1U::GetSolution(const FusionContext& context,
     return sol;
 }
 
+float ConvBiasActivAsm1x1U::GetWti(const FusionContext&, const FusionDescription&) const
+{
+    // Negative values mean rough estimation of time
+    // Solvers with positive values are considered first priority
+    // If none found solvers with negative are used, -2.0f meaning a completely unknown value
+    // So lower absolute values mean higher priority
+    return -.5f;
```
atamazov: [Q] Are you pretty sure that ConvBiasActivAsm1x1U is better than all other applicable ones that return wti_unknown?

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
```diff
@@ -208,6 +207,15 @@ ConvBiasActivAsm1x1U::GetSolution(const FusionContext& context,
     return sol;
 }
 
+float ConvBiasActivAsm1x1U::GetWti(const FusionContext&, const FusionDescription&) const
+{
+    // Negative values mean rough estimation of time
+    // Solvers with positive values are considered first priority
+    // If none found solvers with negative are used, -2.0f meaning a completely unknown value
+    // So lower absolute values mean higher priority
```
atamazov: Use `/// \anchor <LABEL>` to mark comment

src/solver/conv_bin_winoRxS_fused.cpp
```diff
@@ -275,7 +269,21 @@ ConvSolution ConvBinWinogradRxSFused::GetSolution(const FusionContext& context,
     };
     return result;
 }
+float ConvBinWinogradRxSFused::GetWti(const FusionContext&, const FusionDescription& problem) const
+{
+    // Negative values mean rough estimation of time
+    // Solvers with positive values are considered first priority
+    // If none found solvers with negative are used, -2.0f meaning a completely unknown value
+    // So lower absolute values mean higher priority
```
atamazov: Replace `with /// \ref <LABEL>` to avoid duplication of comment.

src/solver/conv_bin_winoRxS_fused.cpp
```diff
@@ -275,7 +269,21 @@ ConvSolution ConvBinWinogradRxSFused::GetSolution(const FusionContext& context,
...
 
+    const auto conv_problem = problem.GetConvProblem(0, miopen::conv::Direction::Forward);
+    const auto x            = conv_problem.GetWeightsWidth();
+    const auto y            = conv_problem.GetWeightsHeight();
+    if(x == 3 && y == 3)
+        return -.01f;
+    else
+        return -.95f;
```
atamazov:
```suggestion
        return wti_unknown / 200.f
    else
        return wti_unknown / 1.9f
```

We may wish to end up with a single place where unknown WTIs for all solvers are defined altogether (possibly as macros or global constexprs).

src/include/miopen/solver.hpp
```diff
@@ -108,6 +108,8 @@ struct SolverBase
     /// run-time parameters.
     virtual bool IsDynamic() const { return false; }
 
+    static constexpr float wti_unknown = -2;
```
atamazov: [notice] google C++ style recommends camelCase with prefix "k", i.e. `kWtiUnknown` (which is very like a kind of hungarian notation which I personally hate). I would prefer WTI_UNKNOWN for this specific case. Decide yourself, no feedback required.

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
atamazov: [Quality] Return `wti_unknown / 4` here. This provides a symbolic connection from this -0.5 to -2.0.

src/solver/conv_ocl_dir2Dfwd_fused.cpp
```diff
@@ -228,6 +223,21 @@ ConvOclDirectFwdFused::GetSolution(const FusionContext& context,
     return result;
 }
 
+float ConvOclDirectFwdFused::GetWti(const FusionContext&, const FusionDescription& problem) const
+{
+    // Negative values mean rough estimation of time
+    // Solvers with positive values are considered first priority
+    // If none found solvers with negative are used, -2.0f meaning a completely unknown value
+    // So lower absolute values mean higher priority
```
atamazov: Replace `with /// \ref <LABEL>` to avoid duplication of comment.

src/solver/conv_ocl_dir2Dfwd_fused.cpp
```diff
@@ -228,6 +223,21 @@ ConvOclDirectFwdFused::GetSolution(const FusionContext& context,
...
+
+    const auto& desc = *problem.fusion_plan_desc;
+    const int bn_idx = GetOpIdx(desc.op_map, miopenFusionOpBatchNormInference);
+    if(bn_idx != -1)
+        return wti_unknown;
+    else
+        return -.9f;
```
atamazov: You know what I mean ;)

src/solver/conv_winoRxS_fused.cpp
```diff
@@ -417,7 +409,12 @@ ConvSolution ConvBinWinogradRxSf2x3g1Fused::GetSolution(const FusionContext& con
     };
     return result;
 }
-
+float ConvBinWinogradRxSf2x3g1Fused::GetWti(const FusionContext& ctx,
+                                            const FusionDescription& problem) const
+{
+    const auto conv_problem = problem.GetConvProblem(0, miopen::conv::Direction::Forward);
+    return conv::ConvBinWinogradRxSf2x3g1().GetWti(ctx, conv_problem);
```
atamazov, line 416: [notice] Maybe multiply with `1.001f` just to notice some small overhead?

src/include/miopen/solver.hpp
atamazov: [notice] Introducing a symbol here logically leads us to introduce symbols in other similar places.

atamazov (changes requested):
Good, but I have to ask for some quality-related changes.

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
DrizztDoUrden: No, but they had these priorities before #2388

DrizztDoUrden: To be fair, I think `unknown / 4` is confusing :)

src/include/miopen/solver.hpp
DrizztDoUrden: Caps is generally reserved for macros to indicate caution required when interacting with them. But I change it to any style if you want.

src/solver/conv_winoRxS_fused.cpp
DrizztDoUrden: I don't think it would be useful because WTI is not meant to be comparable between primitives. I think multiplication by 1.001 would mean that the rest of the kernel is .1% less optimal than non-conv part of its competitors.

src/include/miopen/solver.hpp
atamazov: Up to you, my comment is just a notice

src/solver/conv_asm_1x1u_bias_activ_fused.cpp
atamazov: Let's invent some other name. E.g. `wti_unknown_or_worst` or something else.

averinevg (approved):
LGTM

DrizztDoUrden:
@junliume @JehandadKhan ping

CAHEK7 (approved):

junliume (approved):

PR #3106, opened 2024-07-10
BN FWD INFER OCL to HIP

sgundabo:
This PR focuses on converting the Batch Norm Forward Inference Spatial and Per Activation kernels from OpenCL to HIP. This conversion is a part of the broader initiative to translate all OpenCL kernels within MIOpen, as the OpenCL backend has been deprecated.

**Ensuring correctness:**
The PR includes a GTEST that compares the output of the OpenCL kernel with the HIP implementation. The test cases are derived from the existing batch norm forward inference kernel GTEST.

**Ensuring GPU Performance parity:**
The GTEST also measures the minimum, maximum, mean, median, and standard deviation of the kernel execution time across five runs and records the data in a CSV file. This data is used to create graphs that illustrate the average performance improvement of the HIP implementation over OpenCL. An average performance gain greater than one is considered favorable.
TODO: Collect perf metrics on a wider variety of test cases on a gfx90a, and ensure parity.

**Ensuring Host side Performance parity:**
As the OpenCL backend support is deprecated in MIOpen, the assumption is that this decision was made while being aware of the compilation overhead of HIP kernels over OpenCL.


src/kernels/MIOpenBatchNormFwdInferSpatialHIP.cpp
```diff
@@ -0,0 +1,88 @@
...
+#define _FLOAT float
+#define _FLOAT_PREC float
+#endif
+
+
+template <typename T>
+T __device__ __forceinline__ mad(T a, T b, T c)
+{
+    return a * b + c;
+}
```
CAHEK7: There is a set of `fma` functions supported by HIP, you can find some of them here:
https://rocmdocs.amd.com/projects/HIP/en/develop/reference/kernel_language.html#math-functions

I'm not sure that the documentation is precise, but most probably you can use `fma` directly for all the types.

src/kernels/MIOpenBatchNormFwdInferSpatialHIP.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    _FLOAT_PREC mean, variance, invVariance;
+    _FLOAT_PREC inhat;
+    _FLOAT_PREC pscale, pbias;
+
+    mean        = *(estimatedMean + xgid);
+    variance    = *(estimatedVariance + xgid);
+    pscale      = *(scale + xgid);
+    pbias       = *(bias + xgid);
+    invVariance = rsqrt(fabs(variance + epsilon));
```
CAHEK7:
```suggestion
    const _FLOAT_PREC mean        = *(estimatedMean + xgid);
    const _FLOAT_PREC variance    = *(estimatedVariance + xgid);
    const _FLOAT_PREC pscale      = *(scale + xgid);
    const _FLOAT_PREC pbias       = *(bias + xgid);
    const _FLOAT_PREC invVariance = static_cast<_FLOAT_PREC>(rsqrt(fabs(variance + epsilon)));
```
`index` and `inhat` should be defined inside the loop.
Or even use `const auto` instead of `const _FLOAT_PREC`.

src/kernels/MIOpenBatchNormFwdInferSpatialHIP.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    for(unsigned int idx = ygid; idx < imageDims; idx += gridDim.y * blockDim.y)
+    {
+        for(unsigned int n = 0; n < batchSize; n++)
+        {
+            index      = (n * batchStride) + (xgid * imageDims) + idx;
+            inhat      = ((_FLOAT_PREC)(*(in + index)) - mean) * invVariance;
```
CAHEK7:
```suggestion
            inhat      = (static_cast<_FLOAT_PREC>(*(in + index)) - mean) * invVariance;
```
C-style cast is not a good practice in C++ code. Under the hood it is some sort of a combo of `reinterpret_cast`+`const_cast` the first one can break a lot of optimizations and do strange things and the second one is usually bad.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+            kernel_name += "SpatialEst";
+        }
+    }
+    else
+    {
+
+        if(use_hip)
+        {
+
+            kernel_file += "PerActHIP.cpp";
```
CAHEK7, line 111: Seems `MIOpenBatchNormFwdInferPerActHIP.cpp` has been forgotten.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+                                 ConstData_t x,
+                                 const miopen::TensorDescriptor& yDesc,
+                                 Data_t y,
+                                 const miopen::TensorDescriptor& bnScaleBiasMeanVarDesc,
+                                 ConstData_t bnScale,
+                                 ConstData_t bnBias,
+                                 ConstData_t estimatedMean,
+                                 ConstData_t estimatedVariance,
+                                 double epsilon,
+                                 bool use_hip = false)
```
CAHEK7: Usually default parameters is a bad practice, especially for the newly added functions.
Default parameter should be really rarely used like you almost never want to pass it explicitly. But it may break further refactoring of this function since default parameters must always be the last ones and it makes readability worse because of its implicit nature.

Though it can be a part of refactoring when you add some new parameter and don't want to rewrite old code right now.

Later in the code you've always passed it.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+{
+protected:
+    void SetUp() override
+    {
+        std::tie(bn_config, tensor_layout) = GetParam();
+        bn_infer_test_data.SetUpImpl(bn_config, tensor_layout);
+
+    }
+
+    void RunTestGPU(bool hip_en = false)
```
CAHEK7: Avoid using default parameters.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+    if(xDesc.GetType() == miopenHalf && bnScaleBiasMeanVarDesc.GetType() == miopenHalf)
+    {
+        bfp16parm = true;
+        bfp32parm = false;
+    }
+    else if(xDesc.GetType() == miopenHalf && bnScaleBiasMeanVarDesc.GetType() == miopenFloat)
+    {
+        bfpmixparm = true;
+        bfp32parm  = false;
+    }
```
CAHEK7:
```suggestion
    bool bfpmixparm = xDesc.GetType() != bnScaleBiasMeanVarDesc.GetType();
    bool bfp16parm  = xDesc.GetType() == miopenHalf;
    bool bfp32parm  = xDesc.GetType() == miopenFloat;
    assert(bfp16parm || bfp32parm);
    assert((bfp32parm  && !bfpmixparm) || bfp16parm);
```

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+    int n, c, h, w;
+    std::tie(n, c, h, w) = miopen::tien<4>(xDesc.GetLengths());
```
CAHEK7, line 65:
```suggestion
    auto [n, c, h, w] = miopen::tien<4>(xDesc.GetLengths());
```
C++17 has a special structure binding syntax. It works perfect with simple values like `ints` or `floats`.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+    EXPECT_FALSE(miopen::range_zero(output)) << "HIP data is all zeros";
+    EXPECT_FALSE(miopen::find_idx(output, miopen::not_finite) >= 0)
+        << "Non finite number found in the HIP data";
+    EXPECT_TRUE(miopen::range_distance(ref_out) == miopen::range_distance(output));
+    auto error = miopen::rms_range(ref_out, output);
+    EXPECT_FALSE(miopen::find_idx(ref_out, miopen::not_finite) >= 0)
+        << "Non finite number found in the OCL data";
+
+    EXPECT_TRUE(error < threshold)
+        << "Error beyond tolerance Error:" << error << ",  Threshold: " << threshold;
```
CAHEK7:
```suggestion
    EXPECT_LT(error, threshold);
```
Also check the possible gtest assertion predicates:
https://google.github.io/googletest/reference/assertions.html

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+template <typename T>
+void BNTensorCompare(const tensor<T>& output,
+                     const tensor<T>& ref_out,
+                     const double threshold = std::numeric_limits<T>::epsilon())
+{
+    EXPECT_FALSE(miopen::range_zero(ref_out)) << "OCL data is all zeros";
+    EXPECT_FALSE(miopen::range_zero(output)) << "HIP data is all zeros";
+    EXPECT_FALSE(miopen::find_idx(output, miopen::not_finite) >= 0)
+        << "Non finite number found in the HIP data";
+    EXPECT_TRUE(miopen::range_distance(ref_out) == miopen::range_distance(output));
```
CAHEK7:
```suggestion
    EXPECT_EQ(miopen::range_distance(ref_out), miopen::range_distance(output));
```
But since tensor size does not depend on the algorithm and controlled by the test itself, there is no point to compare and verify those values. They will always match.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+    RunTestGPU(true);
+
+    // Compare the outputs.
+    Verify();
+};
+
+INSTANTIATE_TEST_SUITE_P(BatchNormFwdInferTestSet,
+                         BatchNormFwdInferTestFloat,
+                         testing::Combine(testing::ValuesIn(Network1<BNTestCase>()),
+                                          testing::ValuesIn({miopenTensorNCHW})));
```
CAHEK7:
```suggestion
                                          testing::Values(miopenTensorNCHW)));
```
Probably it's a better way to pass single value, or even multiple ones like `testing::Values(miopenTensorNCHW, miopenTensorNHWC)`.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+        bn_infer_test_data;
+
+    miopenTensorLayout_t tensor_layout;
+};
+
+namespace BatchNormFwdInfer {
+
+struct BatchNormFwdInferTestFloat : BatchNormFwdInferTest<float, float, float, float, float>
+{
+};
```
CAHEK7: `half` datatype is not covered.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+                                            bn_infer_test_data.output.data.size());
+    }
+
+    void ComputeCPU() { test::ComputeCPUBNInference(bn_infer_test_data); }
+
+    void Verify() { BNTensorCompare<YDataType>(bn_infer_test_data.output, bn_infer_test_data.ref_out, 4e-3); }
+
+
+    BNTestCase bn_config;
+    bool test_skipped = false;
```
CAHEK7: It is unused.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+#include <gtest/gtest.h>
+#include <miopen/env.hpp>
+#include <miopen/miopen.h>
+#include <miopen/kernel_build_params.hpp>
+
+#include <miopen/batchnorm/problem_description.hpp>
+
+#include <miopen/miopen.h>
+#include <gtest/gtest.h>
+#include <miopen/solver/ck_utility_common.hpp>
```
CAHEK7: I don't think that ck utilities are needed. Could you double check and include only those headers which are really used.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,274 @@
...
+    std::string kernel_name = "MIOpenBatchNormFwdInfer";
+
+    std::string params = use_hip ? build_params.GenerateFor(miopen::kbp::OpenCL{})
+                                 : build_params.GenerateFor(miopen::kbp::HIP{});
+
+    if(bn_mode == miopenBNSpatial)
+    {
+        if(use_hip)
+        {
+
```
CAHEK7: Extra empty line. And there are few more empty lines.

CAHEK7 (dismissed):
The main concern is missed `half` data type test. Other comments are mostly C++ related and not very critical.

Also clang-format should be applied to the files. And can you take a look into this draft https://gist.github.com/CAHEK7/9ed75ca55acf85f92efd89eb2446a94d and adopt it.

test/gtest/bn_fwd_infer_ocl_hip.cpp
sgundabo: GetLengths() returns size_t, but I expect n,c,h,w to be ints. So, sticking to doing whatever I was doing before.

test/gtest/bn_fwd_infer_ocl_hip.cpp
sgundabo: Fixed, added the Per Activation kernel too.

test/gtest/bn_fwd_infer_ocl_hip.cpp
sgundabo: Removed tensor size check

src/kernels/MIOpenBatchNormFwdInferSpatialHIP.cpp
```diff
@@ -0,0 +1,98 @@
...
+                                         const _FLOAT_PREC* __restrict estimatedMean,
+                                         const _FLOAT_PREC* __restrict estimatedVariance,
+                                         const _FLOAT_PREC* __restrict scale,
+                                         const _FLOAT_PREC* __restrict bias,
+                                         double epsilon,
+                                         unsigned int batchSize,
+                                         unsigned int imageDims,
+                                         unsigned int batchStride)
+{
+    bn_fwd_infer_sparial<_FLOAT, _FLOAT_PREC>(in,
```
CAHEK7:
```suggestion
    bn_fwd_infer_spatial<_FLOAT, _FLOAT_PREC>(in,
```
Tiny typo)

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,322 @@
...
+{
+    EXPECT_FALSE(miopen::range_zero(ref_out)) << "OCL data is all zeros";
+    EXPECT_FALSE(miopen::range_zero(output)) << "HIP data is all zeros";
+    EXPECT_FALSE(miopen::find_idx(output, miopen::not_finite) >= 0)
+        << "Non finite number found in the HIP data";
+    auto error = miopen::rms_range(ref_out, output);
+    EXPECT_FALSE(miopen::find_idx(ref_out, miopen::not_finite) >= 0)
+        << "Non finite number found in the OCL data";
+    EXPECT_LT(error, threshold) << "Error beyond tolerance Error:" << error
+                                << ",  Threshold: " << threshold;
```
CAHEK7: `EXPECT_LT` prints that kind of message:
```
double threshold = 1.;
double error = 2.;
EXPECT_LT(error, threshold);
-----------------------
test.cpp:42: Failure
Expected: (error) < (threshold), actual: 2 vs 1
```
`"Error beyond tolerance"` message is fine, but printing the values is not required.

That's the main difference between `EXPECT_TRUE` and `EXPCET_LT`, the first one does not print the values:
```
double threshold = 1.;
double error = 2.;
EXPECT_TRUE(error < threshold);
-----------------------
/home/user/miopen/test/gtest/gpu_mha_forward.cpp:143: Failure
Value of: error < threshold
  Actual: false
Expected: true
```

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+        output_ref = handle.Read<YDataType>(bn_infer_test_data.out_dev,
+                                            bn_infer_test_data.output.data.size());
+    }
+
+    void ComputeCPU() { test::ComputeCPUBNInference(bn_infer_test_data); }
+
+    void Verify()
+    {
+        BNTensorCompare<YDataType>(bn_infer_test_data.output, bn_infer_test_data.ref_out, 4e-3);
```
CAHEK7: `4e-3` seems too high threshold, ideally they should be bit-exact.
Btw, `BNTensorCompare` has `threshold` defaulted.

test/gtest/bn_fwd_infer_ocl_hip.cpp
```diff
@@ -0,0 +1,322 @@
...
+                     const tensor<T>& ref_out,
+                     const double threshold = std::numeric_limits<T>::epsilon())
+{
+    EXPECT_FALSE(miopen::range_zero(ref_out)) << "OCL data is all zeros";
+    EXPECT_FALSE(miopen::range_zero(output)) << "HIP data is all zeros";
+    EXPECT_FALSE(miopen::find_idx(output, miopen::not_finite) >= 0)
+        << "Non finite number found in the HIP data";
+    auto error = miopen::rms_range(ref_out, output);
+    EXPECT_FALSE(miopen::find_idx(ref_out, miopen::not_finite) >= 0)
+        << "Non finite number found in the OCL data";
```
CAHEK7: I've just realized, that `rms` error is a bad metric for our case.
This is an elementwise operation, it doesn't perform any accumulation and does not depend on the order of operations. It means it must be ideally bit-exact or has a minimal-possible difference.
I think in this particular case `miopen::max_diff` is more suitable and a threshold must be as minimal as possible, down to 0.0.

CAHEK7 (commented):
Everything looks fine, except a few tiny typos.
Only test and perf data are required to get it merged.
Could you adjust verification algorithm and thresholds and attach perf measurements data?

test/gtest/bn_fwd_infer_ocl_hip.cpp
sgundabo: Makes sense. Replaced with max_diff with a zero threshold.

test/gtest/bn_fwd_infer_ocl_hip.cpp
sgundabo: Oh gotcha, that is good to know.

atamazov:
@sgundabo PR description is missing some important info.

Please provide:
- explanation of what is done and _why_
- how this have been tested for correctness
- the same Q about GPU time/performance
  - is it guaranteed that the new kernels have the same or better perf than the old ones? If yes, then _how_ this is guaranteed?
- the same Q about host-side overhead.
  - we know that HIP compilation time is typically 10x longer than OCL. We must ensure that OCL->HIP transition does not lead to degradation of "initial iteration time".

/cc @CAHEK7 @junliume

CAHEK7:
> the same Q about host-side overhead.
> * we know that HIP compilation time is typically 10x longer than OCL. We must ensure that OCL->HIP transition does not lead to degradation of "initial iteration time".

We can't do anything with it, except deprecating HIP and get back to OCL.

junliume:
@sgundabo could you move this PR to ready for review?

CAHEK7:
@junliume we are waiting for the perf metrics

sgundabo:
**Raw Perf Data**
[BNFwdInferRawPerfData.zip](https://github.com/user-attachments/files/16440356/BNFwdInferRawPerfData.zip)

HW info: gfx90a

**FP32 Perf**
BNFwdInferRawPerfData_FP32

**FP16 Perf**
BNFwdInferRawPerfData_FP16

CAHEK7 (approved):

CAHEK7:
It can be safely merged since it does not affect production code. I just don't want to lose this PR.

junliume:
Restarted CI for sanity checks

BradPepersAMD:

PR #3112, opened 2024-07-11
Add tests for ConvBiasResAddActivation forward using Graph API

BrianHarrisonAMD:
This PR adds tests for performing ConvBiasResAddActivation fwd via the graphAPI.

Note: Since we don't have an engine yet that can perform fused convolutions, the execution, and verification are left commented out for now. In a follow up PR once we add the engine, we should be able to uncomment the execution and verification code to verify results.

### DIFF
--- a/test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
+++ b/test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
@@ -1,0 +1,394 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include <gtest/gtest.h>
+#include <miopen/miopen.h>
+#include <miopen/env.hpp>
+
+#include <miopen/graphapi/convolution.hpp>
+#include <miopen/graphapi/execution_plan.hpp>
+#include <miopen/graphapi/matmul.hpp>
+#include <miopen/graphapi/opgraph.hpp>
+#include <miopen/graphapi/pointwise.hpp>
+#include <miopen/graphapi/rng.hpp>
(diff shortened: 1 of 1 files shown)


junliume:
@apwojcik could you help to take a look at the Windows CI?
>  ERROR: Could not find a version that satisfies the requirement setuptools>=40.8.0 (from versions: none)

junliume (approved):

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+        // EXPECT_FALSE(miopen::find_idx(output, miopen::not_finite) >= 0)
+        //     << "Non finite number found in the GPU data";
+
+        // EXPECT_TRUE(error < threshold)
+        //     << "Error beyond tolerance Error:" << error << ",  Threshold: " << threshold;
+    }
+};
+
+struct GPU_ConvBiasResAddActivation_fwd_FP16 : GPU_ConvBiasResAddActivation_fwd<half_float::half>
```
amberhassaan, line 381: lines 381-<end> can be put in a macro to instantiate tests for multiple data types. See `test/gtest/group_conv.hpp`. See https://github.com/ROCm/MIOpen/blob/develop/test/gtest/group_conv.hpp#L571-L586

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+#include "conv3d_test_case.hpp"
+
+namespace gr = miopen::graphapi;
+
+namespace conv_graph_api_test {
+
+bool TestIsApplicable() { return true; }
+
+std::vector<Conv3DTestCase> ConvTestConfigs()
```
amberhassaan, line 50: Follow up task: we should update `Conv3DTestCase` to use structs for spatial dims, padding, dilation, strides etc. See `test/gtest/group_conv.hpp` line 47.

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
```diff
@@ -0,0 +1,395 @@
...
+                                          .setX(x)
+                                          .setB(b)
+                                          .setAlpha1(alpha1)
+                                          .setAlpha2(alpha2)
+                                          .setY(y)
+                                          .build());
+    }
+
+    template <typename DataT>
+    static std::vector<int64_t> Convert(const std::vector<DataT>& value)
```
amberhassaan, line 95: We can remove and revise this test after we merge the `TensorDescriptor` inheritance PR #3119

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+        TensorData& operator=(const TensorData&) = delete;
+        TensorData& operator=(TensorData&&) = default;
+
+        ~TensorData() = default;
+    };
+
+    struct GraphTensorAllocator
+    {
+        gr::AutoDeleteAllocator mAlloc;
```
amberhassaan, line 138: I would have put them in the main test as members but I won't insist on changing it. This is also OK.

amberhassaan (commented):
some comments to address as a followup.

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
BrianHarrisonAMD: Yea I think these can be cleaned up a bit now that we have the TensorDescriptor inheritance.
I think I still needed the convert for the convolution builder though, but ill double check.

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
BrianHarrisonAMD: I was trying to avoid class members on the main test to ensure that the tests are fully isolated.
I know it's possible to reset the state between tests, but I like trying to keep the tests stateless if possible. 

I am fine if we prefer using class members though, and can update to be more consistent.

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
BrianHarrisonAMD: Ah good point.

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
BrianHarrisonAMD: Double checked, and I still need this Convert utility method to go from std::vector<int> that our ConvolutionDescriptor uses, and std::vector<int64_t> that the builder uses.

PR #3114, opened 2024-07-12
[Fixes CI] Fixed tidy hang

DrizztDoUrden:
clang-tidy of versions earlier than 18.0.0 has a bug that causes hangs for hours in some cases with `bugprone-unchecked-optional-access` enabled. Since #3095 it has manifested and led to 3.5 hours static checks stage runs.

Issue in llvm github https://github.com/llvm/llvm-project/issues/59492

I could mess around with changes in #3095, but it would likely manifest again later and is annoying to deal with if one is unfamiliar with the context.

I have tested locally, it reproduces with tidy 17.0.6, doesn't reproduce with 18.0.0, changes fix it for 17.0.6.

### DIFF
--- a/cmake/ClangTidy.cmake
+++ b/cmake/ClangTidy.cmake
@@ -70,6 +70,10 @@ else()
 endif()
 
 set(EXTRA_CHECKS)
+# There is a bug in tidy that hangs it in some cases when it encounters optional access
+# It can spend 3.5h+ and timeout the CI
+# It is fixed in 18.0.0 or worked around by disabling this check: bugprone-unchecked-optional-access
+# https://github.com/llvm/llvm-project/issues/59492
 if (CLANG_TIDY_VERSION VERSION_LESS "18.0.0")
     list(APPEND EXTRA_CHECKS -bugprone-unchecked-optional-access)
 endif()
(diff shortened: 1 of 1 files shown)


shurale-nkn (approved):
LGTM

atamazov (approved):
👏

cmake/ClangTidy.cmake
```diff
@@ -69,6 +69,11 @@ else()
     message( STATUS "Clang tidy found: ${CLANG_TIDY_VERSION}")
 endif()
 
+set(EXTRA_CHECKS)
+if (CLANG_TIDY_VERSION VERSION_LESS "18.0.0")
+    list(APPEND EXTRA_CHECKS -bugprone-unchecked-optional-access)
```
averinevg, line 78: Let's comment on why this is necessary

DrizztDoUrden: Did it now

DrizztDoUrden:
It didn't pass all the tests, but it also doesn't change anything not related to tidy and that step is now <30min, so I ll mark it as passing.

DrizztDoUrden:
Many thanks to Kamil, he told me about the issue, which I didn't encounter locally due to having tidy 18 xD

CAHEK7 (approved):

averinevg (approved):

junliume (approved):

PR #3116, opened 2024-07-15
ASM kernels: remove v2 metadata

averinevg:
This PR removes code object v2 metadata support from ASM kernels.


averinevg:
@atamazov Please review

atamazov:
> @atamazov Please review

@averinevg I would prefer keeping this v2 code  (albeit unused) until we add support for v4 or v5. Why: existing code works as a hint -- it denotes all places where conditional assembly might be needed to support some new MD format. What do you think?

averinevg:
@atamazov Sounds reasonable. We'll probably need to add metadata for PAL runtime, then this code will come in handy

PR #3117, opened 2024-07-15
[WA] Remove exception checking tests for unsupported hardware

BrianHarrisonAMD:
Remove hipBLASLt tests that check exceptions are thrown on unsupported hardware. 

Issue is that on 0.8.0+ of hipBLASLt it will result in a segfault when trying to run on unsupported hardware instead of providing an error status. 

For verification, I ran this updated version and verified it no longer results in a segmentation fault since all tests are skipped on unsupported hardware. 

Note: In the future there will be API added to hipBLASLt for querying supported devices, and will add a note to future work to check that as part of the validity checks.


bghimireamd:
LGTM

bghimireamd (approved):

junliume (approved):

PR #3119, opened 2024-07-16
Update GraphAPI Tensor to inherit from TensorDescriptor

BrianHarrisonAMD:
Changes:
- Update GraphAPI Tensor to inherit from TensorDescriptor.
- Remove Dims, Strides, and Type from GraphAPI Tensor. 
    - Instead use base methods / types.
- Update areas to use the correct methods / types.
- Add tests for checking tensor layout is found from strides correctly.

### DIFF
--- a/src/graphapi/find_engine.cpp
+++ b/src/graphapi/find_engine.cpp
@@ -276,7 +276,7 @@ public:
 
         for(auto& [k, v] : *tensor_map)
         {
-            s = miopenSetProblemTensorDescriptor(mha_prob, v.mEnumId, &(v.mTensDesc));
+            s = miopenSetProblemTensorDescriptor(mha_prob, v.mEnumId, v.mGraphTensor);
             MIOPEN_THROW_IF(s != miopenStatusSuccess,
                             "failed while setting tensor descriptor for mha fwd");
         }
--- a/src/graphapi/find_engine.cpp
+++ b/src/graphapi/find_engine.cpp
@@ -81,7 +81,8 @@ class MHA_Fwd_F8_Pattern : public GraphPatternMatcher
     {
 
         assert(attn_scale);
-        std::vector<std::size_t> all1s = {1LL, 1LL, 1LL, 1LL};
+        std::vector<std::size_t> all1s = {
+            std::size_t{1}, std::size_t{1}, std::size_t{1}, std::size_t{1}};
 
         auto tensor_map = std::make_shared<TensorInfoMap>();
 
--- a/src/graphapi/find_engine.cpp
+++ b/src/graphapi/find_engine.cpp
@@ -81,7 +81,7 @@ class MHA_Fwd_F8_Pattern : public GraphPatternMatcher
     {
 
         assert(attn_scale);
-        std::vector<int64_t> all1s = {1LL, 1LL, 1LL, 1LL};
+        std::vector<std::size_t> all1s = {1LL, 1LL, 1LL, 1LL};
 
         auto tensor_map = std::make_shared<TensorInfoMap>();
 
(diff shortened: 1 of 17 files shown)


scerzh (commented):
Graph API prefers `int64_t` rather than `std::size_t`. The questions is where a "border", an implicit or explicit conversion, between them should lie in our implementation. There are 3 places (the closest to C interface first):

- `BackendTensorDescriptor` class
- `TensorBuilder` class
- `Tensor` class

You chose the closest place (`BackendTensorDescriptor`) which is an excelent solution, but there is a consideration.

We have tests that utilize C interface and tests that utilize Builder classes; these test may share the common parameters (ex. tensor parameters) to be passed via GTest's `TEST_P` or `TEST_F` macros. I propose to move this "border" to `TensorBuilder` class to make sure such tests can share the common parameters correctly.

@amberhassaan what do you think about `Builders`?

amberhassaan:
> Graph API prefers `int64_t` rather than `std::size_t`. The questions is where a "border", an implicit or explicit conversion, between them should lie in our implementation. There are 3 places (the closest to C interface first):
> 
> * `BackendTensorDescriptor` class
> * `TensorBuilder` class
> * `Tensor` class
> 
> You chose the closest place (`BackendTensorDescriptor`) which is an excelent solution, but there is a consideration.
> 
> We have tests that utilize C interface and tests that utilize Builder classes; these test may share the common parameters (ex. tensor parameters) to be passed via GTest's `TEST_P` or `TEST_F` macros. I propose to move this "border" to `TensorBuilder` class to make sure such tests can share the common parameters correctly.
> 
> @amberhassaan what do you think about `Builders`?

I'm going with the assumption that `TensorBuilder` and `Tensor` will both work with `size_t` and we update the tests accordingly.  Which means `int64_t` stops at the C interface. I don't think we call both the `TensorBuilder` and `C` interface with the same param types (but I haven't checked thoroughly). 

There is one point in favor of keeping `int64_t` for TensorBuilder and that would be if we want to be fully compatible with `cudnn_frontend` API which also uses `int64_t` for dims and strides. But then our `TensorDescriptor` is not like that, so we're better off staying with `size_t` which logically is more consistent with our codebase.

BrianHarrisonAMD:
> Graph API prefers `int64_t` rather than `std::size_t`. The questions is where a "border", an implicit or explicit conversion, between them should lie in our implementation. There are 3 places (the closest to C interface first):
> 
> * `BackendTensorDescriptor` class
> * `TensorBuilder` class
> * `Tensor` class
> 
> You chose the closest place (`BackendTensorDescriptor`) which is an excelent solution, but there is a consideration.
> 
> We have tests that utilize C interface and tests that utilize Builder classes; these test may share the common parameters (ex. tensor parameters) to be passed via GTest's `TEST_P` or `TEST_F` macros. I propose to move this "border" to `TensorBuilder` class to make sure such tests can share the common parameters correctly.
> 
> @amberhassaan what do you think about `Builders`?

Ah interesting point.
I like the idea of moving it to the builder.

src/graphapi/find_engine.cpp
```diff
@@ -81,7 +81,7 @@ class MHA_Fwd_F8_Pattern : public GraphPatternMatcher
     {
 
         assert(attn_scale);
-        std::vector<int64_t> all1s = {1LL, 1LL, 1LL, 1LL};
+        std::vector<std::size_t> all1s = {1LL, 1LL, 1LL, 1LL};
```
amberhassaan: `ull` for `size_t` literals.

BrianHarrisonAMD:
> Graph API prefers `int64_t` rather than `std::size_t`. The questions is where a "border", an implicit or explicit conversion, between them should lie in our implementation. There are 3 places (the closest to C interface first):
> 
> * `BackendTensorDescriptor` class
> * `TensorBuilder` class
> * `Tensor` class
> 
> You chose the closest place (`BackendTensorDescriptor`) which is an excelent solution, but there is a consideration.
> 
> We have tests that utilize C interface and tests that utilize Builder classes; these test may share the common parameters (ex. tensor parameters) to be passed via GTest's `TEST_P` or `TEST_F` macros. I propose to move this "border" to `TensorBuilder` class to make sure such tests can share the common parameters correctly.
> 
> @amberhassaan what do you think about `Builders`?

Ah interesting point.
I like the idea of putting it in the TensorBuilder to keep the types consistent for the builder interface, but I am fine with either direction.

src/tensor.cpp
```diff
@@ -193,13 +205,68 @@ TensorDescriptor::TensorDescriptor(miopenDataType_t t,
...
+
 // Main private constructor
 TensorDescriptor::TensorDescriptor(miopenDataType_t t,
                                    miopenTensorLayout_t layout_in,
                                    const std::vector<std::size_t>& lens_in,
                                    const std::vector<std::size_t>& strides_in,
                                    bool use_strides)
     : lens(lens_in), type(t), tensorLayout(layout_in)
+{
+    CheckArguments(t, layout_in, lens, strides_in, use_strides);
```
amberhassaan: two constructors are similar, so we can pick common code into an `init` function. May be `CheckArgsAndInit` and you can use `std::forward` for strides since they can be l-value or r-value.

amberhassaan (commented):
initial review.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: @amberhassaan what about the following since we don't have access to C++ 23 uz?

`std::vector<std::size_t> all1s = {std::size_t{1}, std::size_t{1}, std::size_t{1}, std::size_t{1}};`

Or do we prefer 1ULL?

src/tensor.cpp
BrianHarrisonAMD: What about sharing like this? 

```cpp
TensorDescriptor::TensorDescriptor(miopenDataType_t t,
                                   miopenTensorLayout_t layout_in,
                                   const std::vector<std::size_t>& lens_in,
                                   const std::vector<std::size_t>& strides_in,
                                   bool use_strides)
    : lens(lens_in), strides(use_strides ? strides_in : std::vector<std::size_t>()), type(t), tensorLayout(layout_in)
{
    CheckArgsAndInit(use_strides);
}

TensorDescriptor::TensorDescriptor(miopenDataType_t t,
                                   miopenTensorLayout_t layout_in,
                                   std::vector<std::size_t>&& lens_in,
                                   std::vector<std::size_t>&& strides_in,
                                   bool use_strides)
    : lens(std::move(lens_in)), strides(use_strides ? std::move(strides_in) : std::vector<std::size_t>()), type(t), tensorLayout(layout_in)
{
    CheckArgsAndInit(use_strides);
}
```

amberhassaan (approved):
Looks good.

scerzh (approved):
LGTM

junliume (approved):

PR #3120, opened 2024-07-17
Refactor BnCKFwdInference::GetSolution for NHWC

xinlipn:
1. Rename  RunCKSolution to InitInvokerFactoryBnCKFwdInferenceNHWC to differentiate the upcoming the new API InitInvokerFactoryBnCKFwdInferenceNCHW

2. Move common code to implicitgemm_ck_util.hpp


src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -45,6 +50,142 @@ struct ProblemDescription;
 
 namespace solver {
 #if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+namespace batchnorm {
```
amberhassaan: this code should be moved back to a header used by `batchnorm` solvers only if there is common code between forward and backward batchnorm CKArgs etc., then create a separate header otherwise, move back to cpp files. 

Structure the `CKArgs` class similar to how convolution's CKArgs classes are structured where there's a method called `MakeArumentPointer`.

xinlipn: @amberhassaan ,  I have reverted the changes in implicitgemm_ck_util.hpp. Also refactored BnCKFwdInference::GetSolution

src/solver/batchnorm/backward_ck.cpp
src/solver/conv/conv_hip_implicit_gemm_grouped_fwd_xdlops.cpp

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -126,52 +126,61 @@ template <typename XDataType,
...
           typename BiasDataType,
           typename MeanVarDataType>
-static void RunCKSolution(const Handle& handle,
-                          const AnyInvokeParams& primitive_parameters,
-                          const miopen::batchnorm::ProblemDescription& problem)
+ConvSolution InvokerFactoryMakerNHWC(const miopen::batchnorm::ProblemDescription& bn_problem)
 {
-    const auto& args = CKArgsBNormFwd{problem};
+    ConvSolution result;
+    result.invoker_factory = [bn_problem](const std::vector<Kernel>& kernels) {
```
amberhassaan: Do not capture `problem_description` by value as it's a rather large object. Create variables above line 132 for what you need from `problem_description` and capture those variables by values.

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -209,43 +218,63 @@ bool BnCKFwdInference::IsApplicable(
...
-    ConvSolution result;
-    result.invoker_factory = [=](const std::vector<Kernel>& kernels) {
-        std::ignore = kernels;
-        return [=](const Handle& handle, const AnyInvokeParams& primitive_parameters) {
-            switch(bn_problem.GetXDesc().GetType())
+    return MakeAnyInvokerFactory(
+        bn_problem,
+        [&](auto data_type_val) {
+            using T = decltype(data_type_val);
+            if constexpr(std::is_same_v<T, F16>)
```
amberhassaan: This `if/else-if` logic can be simplified as follows: 
```
using AccTy = std::conditional_t<std::is_same_v<T, F64>, 
                               T, // pick this if true
                               F32>; // pick this if false
InvokerFactoryNHWC<T, T, AccT, T, T, AccT>(bn_problem); 
```

In fact InvokerFactory coudl also be changed to take just two template type parameters `T` and `AccT`.

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -126,52 +126,60 @@ template <typename XDataType,
...
+    result.invoker_factory = [args         = CKArgsBNormFwd{bn_problem},
+                              kernel_index = CheckCKApplicability<XDataType,
+                                                                  YDataType,
+                                                                  AccDataType,
+                                                                  ScaleDataType,
+                                                                  BiasDataType,
+                                                                  MeanVarDataType>(bn_problem)](
+                                 const std::vector<Kernel>& kernels) {
+        std::ignore = kernels;
+        return [&](const Handle& handle, const AnyInvokeParams& primitive_parameters) {
```
CAHEK7: @atamazov @DrizztDoUrden can we guarantee that `invoker_factory` context exists longer that its generated lambda?
I'm asking about this case:
```C++
result.invoker_factory = [=](...)
{
    retunr [&](...){...};
};
```
versus that one:
```C++
result.invoker_factory = [=](...)
{
    retunr [=](...){...};
};
```

If we can guarantee that, we can always capture by reference in the second lambda and save some memory and avoid extra copy operations. If we can't, we must always capture by value.

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -126,52 +126,60 @@ template <typename XDataType,
...
+    ConvSolution result;
+    result.invoker_factory = [args         = CKArgsBNormFwd{bn_problem},
+                              kernel_index = CheckCKApplicability<XDataType,
+                                                                  YDataType,
+                                                                  AccDataType,
+                                                                  ScaleDataType,
+                                                                  BiasDataType,
+                                                                  MeanVarDataType>(bn_problem)](
+                                 const std::vector<Kernel>& kernels) {
+        std::ignore = kernels;
```
CAHEK7:
```suggestion
                                 const std::vector<Kernel>& /* kernels */) {
```

src/solver/batchnorm/forward_inference_ck.cpp
DrizztDoUrden: No, it **almost always** exists for a shorter period. Capturing by a reference here is an error, both lambdas should always capture by copy or move, never referencing.

DrizztDoUrden:
```suggestion
        return [=, args = std::move(args)](const Handle& handle, const AnyInvokeParams& primitive_parameters) {
```

DrizztDoUrden (changes requested):

src/solver/batchnorm/forward_inference_ck.cpp
xinlipn: @CAHEK7 , thanks for the comments. Code updated. Could you mark this as resolved?

src/solver/batchnorm/forward_inference_ck.cpp
xinlipn: @DrizztDoUrden , thanks for the comments. Code updated. Could you mark this as resolved?

atamazov: @CAHEK7 see 
- #216

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -126,52 +126,60 @@ template <typename XDataType,
...
+    ConvSolution result;
+    result.invoker_factory = [args         = CKArgsBNormFwd{bn_problem},
+                              kernel_index = CheckCKApplicability<XDataType,
+                                                                  YDataType,
+                                                                  AccDataType,
+                                                                  ScaleDataType,
+                                                                  BiasDataType,
+                                                                  MeanVarDataType>(bn_problem)](
+                                 const std::vector<Kernel>& /*kernels*/) mutable {
+        return [=, args = std::move(args)](const Handle& handle,
```
amberhassaan: don't use `=` by itself, rather name each variable captured by value.

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -126,52 +126,60 @@ template <typename XDataType,
...
-        {args.xyStrides,
-         args.aligned_scaleBiasMeanVarStrides,
-         args.aligned_scaleBiasMeanVarStrides,
-         args.aligned_scaleBiasMeanVarStrides,
-         args.aligned_scaleBiasMeanVarStrides},
-        {args.xyStrides},
-        {params.x, params.estimatedMean, params.estimatedVariance, params.bnScale, params.bnBias},
-        {params.y},
-        Normalize{params.epsilon});
+            auto argument_ptr = bn_ptr->MakeArgumentPointer(args.xyLengths,
```
amberhassaan: add a method to `BnArgs` called `MakeArgPtr` and you can then pick the fields from inside that class. Better design IMO for composability and hiding details.

amberhassaan (commented):
some minor refactoring needed.

src/solver/batchnorm/forward_inference_ck.cpp
xinlipn: @amberhassaan done

src/solver/batchnorm/forward_inference_ck.cpp
xinlipn: @amberhassaan done

DrizztDoUrden (approved):

iq136boy (approved):

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -126,52 +145,47 @@ template <typename XDataType,
...
+                              kernel_index = CheckCKApplicability<XDataType,
+                                                                  YDataType,
+                                                                  AccDataType,
+                                                                  ScaleDataType,
+                                                                  BiasDataType,
+                                                                  MeanVarDataType>(bn_problem)](
+                                 const std::vector<Kernel>& /*kernels*/) mutable {
+        return [args = std::move(args), kernel_index = kernel_index](
+                   const Handle& handle, const AnyInvokeParams& primitive_parameters) {
+            using DeviceOp = ck::tensor_operation::device::DeviceElementwise<
```
amberhassaan: Sorry, I missed this earlier. Please move lines (161-171), i.e. computation of `bn_ptr` outside and capture `bn_ptr` by `move` semantics. See the `InitInvokerFactory` for convolution solvers.

xinlipn: @amberhassaan , changes have been checked in. Thanks

amberhassaan (approved):
This looks OK now. @junliume : ready to merge when it passes CI.

junliume (approved):

PR #3121, opened 2024-07-17
Changed Fusion to have a separate find mode env variable and default to fast

DrizztDoUrden:
The motivation: our original reasoning to add find into `Compile` was to allow tuning. Hybrid mode actually ignores fallback and goes straight to find, which is slower than it was before the find (what is now in fallback). So unless user explicitly wants to tune, including using Hybrid mode, we should go the fastest way possible.

Also fixed it to only register as found1_0 invoker that would be actually be used later.

### DIFF
--- a/src/find_controls.cpp
+++ b/src/find_controls.cpp
@@ -39,6 +39,8 @@
 #include <ostream>
 #include <cstdlib>
 #include <cstring>
+#include <string_view>
+#include <optional>
 
 MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_FIND_ENFORCE)
 MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_DEBUG_FIND_ONLY_SOLVER)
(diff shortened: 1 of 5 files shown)


averinevg (approved):
LGTM

src/find_controls.cpp
```diff
@@ -225,26 +229,34 @@ FindMode::Values GetFindModeValueImpl2()
...
-}
-
-FindMode::Values GetFindModeValue()
-{
-    static const FindMode::Values val = GetFindModeValueImpl();
+    static const FindMode::Values val = [&]() {
+        auto rv = GetFindModeValueImpl2(variable).value_or(defaultValue);
+        MIOPEN_LOG_NQI(variable.GetName() << " = " << rv);
+        return rv;
+    }();
```
CAHEK7, line 243: It seems we have to forbid any runtime environmental values updates.
Looks like this code caches env variable state one more time and thus makes any runtime update impossible.

I'm not against this particular change, especially it has been always here, I'd like to say that we have to focus on our tests and testing infrastructure and properly fix it asap. (though, there are some possible workarounds for this env var problem even for a single-binary test)

@junliume @JehandadKhan

CAHEK7 (approved):

src/find_controls.cpp
junliume: Yes we will need to log this as part of the test infra renovations.

junliume (approved):

PR #3123, opened 2024-07-18
[bug][tests] Fix broken tests

averinevg:
Currently, a number of tests do nothing. Most tests using `test_drive()`, all tests using `invoke_with_params()` and all tests to which the default data type is not passed are broken. This PR fixes the launch of tests. These tests have not been verified to work correctly and are beyond the scope of this PR.


CAHEK7 (approved):

junliume (approved):

test/driver.hpp
```diff
@@ -1376,45 +1353,50 @@ void test_drive_impl(std::string program_name, std::vector<std::string> as)
...
         {
-            test_drive_impl<Driver<double>>(argv[0], std::move(as));
-            break;
+            test_drive_impl<Driver<double>>(name, std::move(as));
+            return;
         }
     }
+
+    // default datatype
+    test_drive_impl<Driver<float>>(name, std::move(as));
```
atamazov, line 1399: Remove default datatype and fail when datatype is not set. Fix tests.

PR #3126, opened 2024-07-21
Fix GCC8 linking issues with stdc++fs

BrianHarrisonAMD:
stdc++fs linking was missing for libMIOpen.so, (but present for tests, and other locations), which was leading to filesystem segfault issues when running libMIOpen.so built with gcc8 on environments with a newer gcc/libstdc++.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -768,6 +768,10 @@ rocm_set_soversion(MIOpen ${MIOpen_SOVERSION})
 
 clang_tidy_check(MIOpen)
 
+if(HAS_LIB_STD_FILESYSTEM)
+    target_link_libraries(MIOpen PRIVATE stdc++fs)
+endif()
+
 find_package(zstd)
 if(zstd_FOUND)
     target_link_libraries(MIOpen PRIVATE $<IF:$<TARGET_EXISTS:zstd::libzstd_shared>,zstd::libzstd_shared,zstd::libzstd_static>)


src/CMakeLists.txt
```diff
@@ -768,6 +768,10 @@ rocm_set_soversion(MIOpen ${MIOpen_SOVERSION})
 
 clang_tidy_check(MIOpen)
 
+if(HAS_LIB_STD_FILESYSTEM)
+    target_link_libraries(MIOpen PRIVATE stdc++fs)
```
junliume, line 772: It seems to be only necessary at some legacy Linux distros and GCC 8?

junliume (approved):

src/CMakeLists.txt
amberhassaan: Yes, that's right. We could put conditionals on the compiler being `GCC` and version being < 8.

amberhassaan: actually, we can't do that because our compiler is `clang++`. Rather this conditional has to go on something else, e.g., a version of `libstdc++`, which I don't know how to detect, so `HAS_LIB_STD_FILESYSTEM` is a good enough choice till then because it should only exist on older systems (I hope).

BrianHarrisonAMD: Here's some extra details:
- https://bugs.launchpad.net/ubuntu/+source/gcc-8/+bug/1824721

I think this will be fine on newer systems, and we are utilizing the solution from the original PR:
- https://github.com/ROCm/MIOpen/pull/2690

atamazov:
@junliume Just in case - this PR looks good to me.

PR #3127, opened 2024-07-22
fix bnorm args log info

fsx950223:


junliume:
@fsx950223 what problem is this "M" vs. "m" causing?

fsx950223:
> @fsx950223 what problem is this "M" vs. "m" causing?

image

CAHEK7 (approved):

junliume (approved):

PR #3131, opened 2024-07-23
Fixed redundant find on failed fusion fallback

DrizztDoUrden:
After #3095 if fallback fails it tries to do a redundant find. It wastes some time because in fusion fallback actually checks all the solvers on its own.


averinevg (approved):
LGTM

junliume (approved):

PR #3151, opened 2024-07-24
Batchnorm Fused Inference OpenCL to HIP 

sgundabo:
This PR focuses on converting the Batch Norm Fused Inference kernel from OpenCL to HIP. This conversion is a part of the broader initiative to translate all OpenCL kernels within MIOpen, as the OpenCL backend has been deprecated.

**Ensuring correctness:**
The PR includes a GTEST that compares the output of the OpenCL kernel with the HIP implementation. The test cases are derived from the existing batch norm forward inference kernel GTEST. Enabling the performance test FLAG should test the kernel on a wider range of tensor sizes that are calculated based on the architecture.

**Ensuring GPU Performance parity:**
The GTEST also measures the minimum, maximum, mean, median, and standard deviation of the kernel execution time across five runs and records the data in a CSV file. This data is used to create graphs that illustrate the average performance improvement of the HIP implementation over OpenCL. An average performance gain greater than one is considered favorable. The OpenCL FP16 kernel is broken, hence only the correctness of the FP16 implementation is verified.

TODO: Collect perf metrics a gfx90a, and ensure parity.

**Ensuring Host side Performance parity:**
As the OpenCL backend support is deprecated in MIOpen, the assumption is that this decision was made while being aware of the compilation overhead of HIP kernels over OpenCL.


src/kernels/activation_functions.hpp
```diff
@@ -0,0 +1,546 @@
...
+#define MIOPEN_NEURON_RELU 3         // max(0, x)
+#define MIOPEN_NEURON_SOFTRELU 4     // log(1 + e^x)   // bonomial normal log likelihood
+#define MIOPEN_NEURON_ABS 5          // abs(x)
+#define MIOPEN_NEURON_POWER 6        // (alpha + beta * x )^gamma
+#define MIOPEN_NEURON_CLIPPED_RELU 7 // min(alpha, max(0, x))
+#define MIOPEN_NEURON_LEAKY_RELU 8   // alpha * x | x <= 0; x | x > 0
+#define MIOPEN_NEURON_ELU 9          // alpha * (e^x - 1) | x <= 0; x | x > 0
+#define MIOPEN_NEURON_TOTAL 10
+
+#define UNUSED __attribute__((__unused__))
```
CAHEK7: I would use standardized C++ attribute `[[maybe_unused]]`.
https://en.cppreference.com/w/cpp/language/attributes/maybe_unused
Or even explicitly remove parameters names like
```C++
__forceinline__ __device__ void ActivationFunction_PassThru(const uint n,
                                                            FP_TYPE_PREC* res,
                                                            const FP_TYPE_PREC* data,
                                                            const FP_TYPE_PREC /*gamma*/,
                                                            const FP_TYPE_PREC /*beta*/,
                                                            const FP_TYPE_PREC /*alpha*/)
```

src/kernels/activation_functions.hpp
```diff
@@ -0,0 +1,546 @@
...
+                                                   FP_TYPE_PREC* res,
+                                                   const FP_TYPE_PREC* data,
+                                                   const FP_TYPE_PREC gamma,
+                                                   const FP_TYPE_PREC beta,
+                                                   const FP_TYPE_PREC alpha)
+{
+    if constexpr(MIOPEN_NRN_OP_ID == MIOPEN_NEURON_PASTHRU)
+    {
+        ActivationFunction_PassThru(n, res, data, gamma, beta, alpha);
+    }
```
CAHEK7: You can do such a trick in C++:
```C++
template<typename T, size_t N>
__forceinline__ __device__ void ActivationFunction_ELU(T (& __restrict__ res)[N],
                                                       const T (& __restrict__ data)[N],
                                                       UNUSED const T gamma,
                                                       UNUSED const T beta,
                                                       const T alpha)
{
    for(size_t i = 0; i < N; ++i)
    {
        res[i] = (data[i] > 0) ? data[i] : (alpha * (exp(data[i]) - static_cast<FP_TYPE_PREC>(1)));
    }
}

template<typename T, size_t N>
__forceinline__ __device__ void ActivationFunction(T (& __restrict__ res)[N],
                                                   const T (& __restrict__ data)[N],
                                                   const T gamma,
                                                   const T beta,
                                                   const T alpha)
{
    static_assert(N<=4);
    if constexpr(MIOPEN_NRN_OP_ID == MIOPEN_NEURON_PASTHRU)
    {
        ActivationFunction_PassThru(res, data, gamma, beta, alpha);
    }
......
```

A calling this function will be like this:
```C++
ActivationFunction(actRes, bnRes, gamma, beta, alpha);
```
Without explicitly passing `MIOPEN_READ_UNIT` and with guarantees that `actRes` and `bnRes` are equal and there will be no out of bound access ever.

CAHEK7 (commented):
Some minor recommendations.

sgundabo:
**Perf Raw Data gfx90a**
[BatchNormInferFusedInfo.zip](https://github.com/user-attachments/files/16431290/BatchNormInferFusedInfo.zip)

**Perf FP32**
BatchNormInferFusedExtraInfo_FP32

src/kernels/activation_functions.hpp
sgundabo: Very interesting! Good to know, thanks!

CAHEK7 (approved):

CAHEK7:
It can be safely merged since it does not affect production code. I just don't want to lose this PR.

CAHEK7:
That PR is also important, since it provides centralized definition for a set of activation functions (for example for #3247 where it has got a local definition for sigmoid https://github.com/ROCm/MIOpen/pull/3247/files#diff-2a117e014b2a1c04feb3ede9723a78a8d11d656b4e9631fa57ab1d7c58df55d6)
@junliume

junliume (approved):

PR #3154, opened 2024-07-25
[WORKAROUND] Allow to use boost:filesystem instead of std::filesystem, [BUGFIX] check_cxx_linker_flag

atamazov:
- 🟠 Allows to use `boost:filesystem` instead of `std::filesystem` via CMake flag.
  - This is to workaround runtime issues (SIGSEGV) related to misconfigured/buggy std::filesystem on some host configurations (e.g. CentOS 8 with manually installed gcc 10).
  - Usage: `cmake ... -DMIOPEN_WORKAROUND_USE_BOOST_FILESYSTEM=On ...`
- By-products
  - 🔴 [BUGFIX][CMake] Fixed issue in check_cxx_linker_flag, see https://github.com/ROCm/MIOpen/commit/4549e017af33763b2751c316214399030e954f11#r144783324
  - [NFC][CMake] Improved logging in `check_cxx_linker_flag`
  - [NFC] Some formatting for readability

Locally tested (make check with debug configuration, make analyze etc.)

----
- https://github.com/ROCm/MIOpen/labels/workaround
- https://github.com/ROCm/MIOpen/labels/non-miopen-bug
- https://github.com/ROCm/MIOpen/labels/urgency_blocker
- Proposed reviewers:
  - @BrianHarrisonAMD 
  - @apwojcik 
  - @junliume
- Optional reviewers:
  - @DrizztDoUrden
  - @CAHEK7 
  - @shurale-nkn 
  - @averinevg

### DIFF
--- a/cmake/CheckCXXLinkerFlag.cmake
+++ b/cmake/CheckCXXLinkerFlag.cmake
@@ -54,7 +54,17 @@ set(check_cxx_linker_flag_patterns
 include (CheckCXXSourceCompiles)
 
 function(check_cxx_linker_flag _flag _var)
+    # TODO: Replace STATUS with CHECK_START/PASS/FAIL
+    # when cmake_minimum_required becomes >= 3.17
+    message(STATUS "Performing Test ${_var}")
     set (_source "int main() { return 0; }")
+    set(CMAKE_REQUIRED_QUIET TRUE)
     check_cxx_source_compiles("${_source}" _result ${check_cxx_linker_flag_patterns})
+    set(CMAKE_REQUIRED_QUIET FALSE)
     set(${_var} "${_result}" PARENT_SCOPE)
+    if(${_result})
+        message(STATUS "Performing Test ${_var} - Success")
+    else()
+        message(STATUS "Performing Test ${_var} - Failed")
+    endif()
 endfunction()
--- a/cmake/CheckCXXLinkerFlag.cmake
+++ b/cmake/CheckCXXLinkerFlag.cmake
@@ -54,17 +54,8 @@ set(check_cxx_linker_flag_patterns
 include (CheckCXXSourceCompiles)
 
 function(check_cxx_linker_flag _flag _var)
-    # TODO: Replace STATUS with CHECK_START/PASS/FAIL
-    # when cmake_minimum_required becomes >= 3.17
-    message(STATUS "Performing Test ${_var}")
     set (_source "int main() { return 0; }")
-    set(CMAKE_REQUIRED_QUIET TRUE)
-    check_cxx_source_compiles("${_source}" _result ${check_cxx_linker_flag_patterns})
-    set(CMAKE_REQUIRED_QUIET FALSE)
-    set(${_var} "${_result}" PARENT_SCOPE)
-    if(${_result})
-        message(STATUS "Performing Test ${_var} - Success")
-    else()
-        message(STATUS "Performing Test ${_var} - Failed")
-    endif()
+    set(CMAKE_REQUIRED_LINK_OPTIONS "${_flag}")
+    check_cxx_source_compiles("${_source}" "_${_var}" ${check_cxx_linker_flag_patterns})
+    set(${_var} "${_${_var}}" PARENT_SCOPE)
 endfunction()
--- a/src/load_file.cpp
+++ b/src/load_file.cpp
@@ -37,7 +37,11 @@ namespace miopen {
 
 std::vector<char> LoadFile(const fs::path& path)
 {
+#if MIOPEN_WORKAROUND_USE_BOOST_FILESYSTEM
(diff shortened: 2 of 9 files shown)


atamazov:
@BrianHarrisonAMD @apwojcik @junliume @JehandadKhan @DrizztDoUrden @CAHEK7 @shurale-nkn @averinevg Ready for reviewing, CI testing & merging.

BrianHarrisonAMD:
Is there still more issues with std::filesystem leading to segfaults after the #3126 fix?

I was under the impression that the issue only came up when using GCC8, and not including stdc++fs, but seems like there must be some other way to have issues come up if you were hitting issues with CentOS 8 & GCC10.

atamazov:
@BrianHarrisonAMD Yes, @junliume knows more details.

atamazov:
https://github.com/ROCm/MIOpen/labels/TESTING_CI_PASSED except Windows, I guess.

apwojcik (approved):

cmake/CheckCXXLinkerFlag.cmake
```diff
@@ -54,7 +54,17 @@ set(check_cxx_linker_flag_patterns
...
     set (_source "int main() { return 0; }")
+    set(CMAKE_REQUIRED_QUIET TRUE)
     check_cxx_source_compiles("${_source}" _result ${check_cxx_linker_flag_patterns})
+    set(CMAKE_REQUIRED_QUIET FALSE)
     set(${_var} "${_result}" PARENT_SCOPE)
+    if(${_result})
+        message(STATUS "Performing Test ${_var} - Success")
+    else()
+        message(STATUS "Performing Test ${_var} - Failed")
+    endif()
```
BrianHarrisonAMD: Just curious, why did we start suppressing status messages during the check now?

BrianHarrisonAMD:
LGTM

BrianHarrisonAMD (approved):

cmake/CheckCXXLinkerFlag.cmake
```diff
@@ -54,7 +54,17 @@ set(check_cxx_linker_flag_patterns
 include (CheckCXXSourceCompiles)
 
 function(check_cxx_linker_flag _flag _var)
```
atamazov, line 56: 🐛 see https://github.com/ROCm/MIOpen/commit/4549e017af33763b2751c316214399030e954f11#r144783324

cmake/CheckCXXLinkerFlag.cmake
atamazov: @BrianHarrisonAMD because otherwise the output looks like this:
```bash
-- Performing Test HAS_LIB_STD_FILESYSTEM
-- Performing Test _result
-- Performing Test _result - Success
-- Performing Test HAS_LIB_STD_FILESYSTEM - Success
```
But it seems like I've found some more elegant solution, please expect update.

UPD: Done.

cmake/CheckCXXLinkerFlag.cmake
atamazov: @BrianHarrisonAMD I've found it thanks to your previous comment 🤝

atamazov:
@junliume @BrianHarrisonAMD @apwojcik @JehandadKhan @DrizztDoUrden @CAHEK7 @shurale-nkn @averinevg Ready for (hopefully last) reviewing round.

Contains a bugfix now https://github.com/ROCm/MIOpen/labels/bug.

I highly recommend re-running CI.

cmake/CheckCXXLinkerFlag.cmake
atamazov: Fixed now.

src/load_file.cpp
```diff
@@ -37,7 +37,11 @@ namespace miopen {
 
 std::vector<char> LoadFile(const fs::path& path)
 {
+#if MIOPEN_WORKAROUND_USE_BOOST_FILESYSTEM
```
DrizztDoUrden, line 40: I would suggest declaring some alias via using in the fs header so that workaround macro check would not be propagated to other files.

DrizztDoUrden (approved):
LGTM, just a minor suggestion

src/load_file.cpp
atamazov: @DrizztDoUrden Thanks. Of course, I considered this, but decided to leave it as is because
- The alias would only be used once
- Introduces another level of indirection
- The diff would be one line longer

atamazov:
@junliume It seems like the problem with CentOS8/gcc10 is not a blocker anymore? Even if so, I would recommend merging this soon because this PR contains a bugfix in check_cxx_linker_flag. Or, If we do not need `boost:filesystem`, then I can prepare the PR that contains only bugfixes and improvements.

junliume (approved):

PR #3157, opened 2024-07-26
Fixed Find 2.0 fusion ignoring miopenSetFindOptionTuning

DrizztDoUrden:
The value in find options was completely ignored.


CAHEK7 (approved):
Could you just describe why we shouldn't use `FusionContext` to get the contexts for `FindFusion`?

DrizztDoUrden:
Because it is not needed and passed as ExecutionContext anyway. In general, deriving things from ExecutionContext is not very useful and is kind of legacy.
FusionContext is especially not useful:
https://github.com/ROCm/MIOpen/blob/develop/src/include/miopen/fusion/context.hpp

junliume:
@DrizztDoUrden could you follow up on this PR? The build has some issues:
```
[2024-07-31T16:09:30.687Z] /home/jenkins/workspace/able-f2-tuning-option-for-fusion/src/fusion.cpp:1100:30: error: no member named 'do_search' in 'miopen::FindOptions'

[2024-07-31T16:09:30.687Z]  1100 |     ctx.do_search = options->do_search;

[2024-07-31T16:09:30.687Z]       |                     ~~~~~~~~~^
```

junliume (approved):

PR #3159, opened 2024-07-26
Implemented find modes and fallback for FusionPlanDescriptor::Compile

DrizztDoUrden:
Changes are squashed #3095, #3114, #3105, #3121, #3131 with some fixes to make it build.
This significantly speeds up fusion when no find is expected.


junliume:
@DrizztDoUrden hip-tidy is actually failing after this change, why do we need it in the staging branch for ROCm 6.2?

junliume (approved):

PR #3161, opened 2024-07-26
Add pattern and executor for ConvBiasResAddActiv GraphAPI

BrianHarrisonAMD:
This change adds the pattern and executor for ConvBiasResAddActiv GraphAPI, and enables previously disabled tests to verify functionality.

Changes:
- Added Executor, pattern, and hooked them up inside find_engine
- Updated test cases to run pattern matching, executor, and verify output
- Tested with existing test cases
- Updated GraphApiVariantPack tests to allow nullptr for workspace


src/graphapi/find_engine.cpp
```diff
@@ -35,12 +35,101 @@
...
+        assert(activ);
+
+        bool isBiasNode1 = IsBiasNode(add1);
+        bool isBiasNode2 = IsBiasNode(add2);
+
+        if(!isBiasNode1 && !isBiasNode2)
+        {
+            MIOPEN_THROW(miopenStatusBadParm,
+                         "no bias node provided for graph matching ConvBiasResAddActive pattern");
+        }
```
BrianHarrisonAMD: Could be swapped to MIOPEN_THROW_IF

src/graphapi/conv_bias_res_add_activ_forward_executor.cpp
```diff
@@ -0,0 +1,131 @@
...
+    auto convertToFloat = [](auto&& arg) { return Convert(arg); };
+    // The virtual tensor for add is the result of the convolution, and combining the alpha1's allow
+    // users to specify it through the graphAPI properly.
+    float alpha1 =
+        mConvOp->getAlpha() * (xIsVirtual ? std::visit(convertToFloat, mAddOp->getAlpha1())
+                                          : std::visit(convertToFloat, mAddOp->getAlpha2()));
+
+    // The non-virtual tensor for add should be alpha 2.
+    float alpha2 =
+        xIsVirtual ? std::get<float>(mAddOp->getAlpha2()) : std::get<float>(mAddOp->getAlpha1());
```
BrianHarrisonAMD: Looking at this again, I think it would make more sense to move some of this into the GraphPatternMatcher, and instead of passing the operations, I can pass the values I need out of them.  

For example, instead of passing the addOp, and needing to find which tensor to use for z, I could do that in the GraphPatternMatcher, and then make the executor's constructor explicit. I think that fits the design better, and fits closer to the other executor.

src/graphapi/conv_bias_res_add_activ_forward_executor.cpp
```diff
@@ -0,0 +1,93 @@
...
+            Convert(conv.getDilations()),
+            Convert(conv.getPostPaddings()),
+            groupCount};
+}
+
+void ConvBiasResAddActivForwardExecutor::execute(miopenHandle_t handle, const VariantPack& vpk)
+{
+    std::size_t in_c  = mXTensor->GetLengths()[1];
+    std::size_t wei_c = mWTensor->GetLengths()[1];
+    int groupCount    = in_c / wei_c;
```
BrianHarrisonAMD: I think we should probably be checking for this in the pattern matcher, and throw if they aren't evenly divisible.

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,148 @@
...
+        assert(add2);
+
+        bool isBiasNode1 = IsBiasNode(add1);
+        bool isBiasNode2 = IsBiasNode(add2);
+
+        if(!isBiasNode1 && !isBiasNode2)
+        {
+            MIOPEN_THROW(miopenStatusBadParm,
+                         "no bias node provided for graph matching ConvBiasResAddActive pattern");
+        }
```
BrianHarrisonAMD: Come to think of it, should this check actually be part of the matches check above?

I think that might actually make more sense, since we don't actually have a bias node for pattern checking, and this pattern requires a bias node that needs to be checked based off the provided tensor descriptions.

For example, this pattern matcher could currently match to a graph with the same nodes, but want to perform 2 adds instead of a bias. It would then throw when asking for the engines since there is no bias node, and that seems incorrect.

(Looking for feedback from reviewers on this)

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,148 @@
...
+                "invalid weight tensor provided for graph matching ConvBiasResAddActive pattern");
+        }
+        else if(in_c % wei_c != std::size_t{0})
+        {
+            MIOPEN_THROW(miopenStatusBadParm,
+                         "invalid group count from input and weight tensor for graph matching "
+                         "ConvBiasResAddActive pattern");
+        }
+
+        int groupCount = in_c / wei_c;
```
BrianHarrisonAMD, line 163: Should these checks be part of the matches check?
Not sure if matches implies that it should run without issue, or just if the pattern matches. 

In this case, I think it makes sense not to be part of the matches check since this implies invalid tensors were passed in for the ConvolutionOp. 

(Looking for feedback from reviewers on this)

src/graphapi/find_engine.cpp
BrianHarrisonAMD: Also, need to add tests that check for these cases once it's determined how we want to handle it.

Should make sure it's throwing as expected, or not matching as expected once we know how to handle these cases.

src/graphapi/conv_bias_res_add_activ_forward_executor.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <miopen/fusion.hpp>
+#include <miopen/visit_float.hpp>
+
+namespace miopen {
+
+namespace graphapi {
+
+static std::vector<int> Convert(const std::vector<int64_t>& values)
+{
+    return {values.begin(), values.end()};
```
amberhassaan: we should use a `static_cast` here.

amberhassaan: May be put an assert that value is less than max int.

src/graphapi/find_engine.cpp
amberhassaan: some of these checks are in the high-level convolution API but then fused convolutions don't seem to go through that API and hence bypass those checks. I don't think we need to check extensively here and the solver should do more checks of the nature of valid params. Here we primarily check the graph (nodes and tensors). This is my initial opinion subject to change.

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,148 @@
...
+        {
+            MIOPEN_THROW(miopenStatusBadParm,
+                         "no bias node provided for graph matching ConvBiasResAddActive pattern");
+        }
+
+        auto* activ = dynamic_cast<OperationPointwise*>(
+            graph.findOutNeighByName(add2, "OP_POINTWISE:RELU_FWD"));
+        assert(activ);
+
+        auto [add, bias] = isBiasNode1 ? std::make_pair(add2, add1) : std::make_pair(add1, add2);
```
amberhassaan: I don't know if there's flexibility about the position of Bias node. Usually Bias follows Conv directly.

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,148 @@
...
+
+        bool addXIsVirtual = add->getX()->isVirtual();
```
amberhassaan: What's this logic doing? When would `X` be virtual?

amberhassaan (approved):
Some comments but looks OK overall.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: I tried to make it flexible as I wasn't sure if we wanted to be too strict with which input is the virtual one.

For example, if the graph specified that B is the virtual input instead of X, then I think it should be able to handle that still since order doesn't matter for add.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: I was wondering about that.
I figured since they were the same node type, and that they are both adds, then it would be more flexible this way, but I could swap it to enforce bias after conv if that's better.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: In this specific case, since I need to calculate groupCount for the convolution descriptor , I think I need to check and throw here before calculating the groupCount, but I like the general opinion of how we should handle verification. That will minimize duplicating existing validity checks, and seems like a good approach to me.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: Decided to move it since I think it's better not to match if the bias node isn't there.

src/include/miopen/graphapi/variant_pack.hpp
```diff
@@ -156,12 +157,8 @@ class VariantPackBuilder
     }
     VariantPackBuilder& setWorkspace(void* workspace) &
     {
-        if(workspace == nullptr)
-        {
-            MIOPEN_THROW(miopenStatusBadParm);
-        }
-
         mVariantPack.mWorkspace = workspace;
+        mWorkspaceSet           = true;
```
randyspauldingamd, line 161: Just to verify intention--now the user _must_ call this function, but it's ok if they set it to NULL.

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,163 @@
...
+        assert(addXIsVirtual || add->getB()->isVirtual());
+
+        auto convertToFloat = [](auto&& arg) { return static_cast<float>(arg); };
+        // The virtual tensor for the add is the result of the convolution, and combining the
+        // alpha1's allow for users to specify the alpha on either, or both nodes, and have it be
+        // correct.
+        float alpha1 =
+            conv->getAlpha() * (addXIsVirtual ? std::visit(convertToFloat, add->getAlpha1())
```
randyspauldingamd: This swaps alpha1 and alpha2 when X is concrete (i.e. B is virtual). Just a wild guess, but it seems counterintuitive. Also, I find this code to be (subjectively ofc) easier to read:
```suggestion
            auto alpha1 = std::visit(convertToFloat, add->getAlpha1();
            auto alpha2 = std::visit(convertToFloat, add->getAlpha2();
            if(addXIsVirtual) std::swap(alpha1, alpha2);
```

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,163 @@
...
+                                     : std::visit(convertToFloat, add->getAlpha1());
+
+        auto* zTensor = addXIsVirtual ? add->getB() : add->getX();
+
+        bool biasXIsVirtual = bias->getX()->isVirtual();
+        assert(biasXIsVirtual || bias->getB()->isVirtual());
+
+        auto* biasTensor = biasXIsVirtual ? bias->getB() : bias->getX();
+
+        std::shared_ptr<GraphPatternExecutor> exec = ConvBiasResAddActivForwardExecutor::make(
```
randyspauldingamd, line 186: It's sure tempting to add a wrapper class that encapsulates all this selection. Something like:
```suggestion
class OperationPointwise_OneVirtual   // TODO: better name
{ // forgive the pseudocode and partial code--hopefully it'll do for a quick suggestion
public:
  Tensor *add, *bias;
  float alpha1, alpha2;
  OperationPointwise_OneVirtual(OpGraph* graph_ptr) 
{
  // do sanity checking along the way
  add = // identify the 'add' node, identify its concrete Tensor, and store
  bias = // identify the 'bias' node, identify its concrete Tensor, and store
  // select and store alpha1 and alpha2
}
  float getAlpha1() { return alpha1; }
  float getAlpha2() { return alpha2; }
};
```
Then this function becomes, starting at 146: 
```suggestion
OperationPointwise_OneVirtual oneVirtual{graph_ptr};
std::shared_ptr<GraphPatternExecutor> exec = ConvBiasResAddActivForwardExecutor::make(
            conv->getX(),
            conv->getW(),
            conv->getConvolution(),
            groupCount,
            oneVirtual.add,
            oneVirtual.bias,
            activ->getY(),
            oneVirtual.getAlpha1(),
            oneVirtual.getAalpha2(),
            std::visit(convertToFloat, activ->getAlpha1()));
```

src/CMakeLists.txt
```diff
@@ -142,6 +142,7 @@ set( MIOpen_Source
     graphapi/rng.cpp
     graphapi/tensor.cpp
     graphapi/variant_pack.cpp
+    graphapi/conv_bias_res_add_activ_forward_executor.cpp
```
randyspauldingamd: Please keep filenames sorted alphabetically. Although I don't know the convention we use for underscore, since it's ascii code falls between the capitals and lowercase letters :)

src/graphapi/find_engine.cpp
amberhassaan: ok, I see. Good to keep the flexibility.

src/CMakeLists.txt
amberhassaan: agree. Sorting should be maintained.

src/graphapi/find_engine.cpp
CAHEK7: I guess one you've found a list of applicable engines, you have to go through that list and check `isApplicable` method.
I'm a bit out of the context and probably it's not possible at this step, but we can't return non-applicable solver\engine in any case.
(I "unresolved" this case just to make it visible)

src/CMakeLists.txt
CAHEK7: Does anybody know any linter for CMAKE?

src/graphapi/conv_bias_res_add_activ_forward_executor.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <miopen/errors.hpp>
+#include <miopen/graphapi/conv_bias_res_add_activ_forward_executor.hpp>
+#include <miopen/fusion.hpp>
+#include <miopen/visit_float.hpp>
+
+namespace miopen {
+
+namespace graphapi {
+
+static std::vector<int> Convert(const std::vector<int64_t>& values)
```
CAHEK7: I would recommend keeping these functions inside an unnamed namespace, since it covers more cases that just a `static`. It's my personal preference and recommendation, but since we don't have any particular style guide, we can do whatever we want.

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,163 @@
...
+        for(auto& length : lengthsToCheck)
+        {
+            if(length > size_t{1})
+            {
+                notOneOrZeroCount++;
+            }
+        }
+
+        // Assume bias node when at most one dimension is larger than 1.
+        return notOneOrZeroCount <= 1;
```
CAHEK7: It looks like https://en.cppreference.com/w/cpp/algorithm/count

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,163 @@
...
+    }
+
+    static bool hasBiasNode(const OpGraph& graph)
+    {
+        auto* conv = dynamic_cast<OperationConvolutionForward*>(
+            graph.findOutNeighByName(graph.getSourceNode(), "OP_CONVOLUTION_FORWARD"));
+        assert(conv);
```
CAHEK7: We can do `dynamic_cast` to a reference and it will automatically throw an exception.
But probably we've always known that all the `dynamic_cast`'s are castable and they have never returned `nullptr`.

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,163 @@
...
+    static std::unique_ptr<GraphPatternMatcher> Make()
+    {
+        return std::make_unique<ConvBiasResAddActive_Fwd_Pattern>();
+    }
+
+    std::string_view name() const final
+    {
+        static const char* n = "convbiasresaddactivation_fwd";
+        return n;
+    }
```
CAHEK7: Some sort of nitpicking:
```suggestion
    std::string_view name() const final
    {
        static const std::string_view n{"convbiasresaddactivation_fwd"};
        return n;
    }
```
It's identical even on Og, but slightly better on O0.
And it expresses the intension clearly - you don't need to "construct" `string_view` from the `const char*`

src/graphapi/find_engine.cpp
```diff
@@ -36,12 +36,163 @@
...
+            return false;
+        }
+
+        return hasBiasNode(*graph_ptr);
+    }
+
+    std::vector<Engine> getEngines(OpGraph* graph_ptr) const override
+    {
+        assert(graph_ptr);
+        assert(matches(graph_ptr));
```
CAHEK7, line 141: Is it an assertion of a fatal error which should be properly reported even in release build?

src/include/miopen/graphapi/conv_bias_res_add_activ_forward_executor.hpp
```diff
@@ -0,0 +1,109 @@
...
+          mYTensor(yTensor),
+          mAlpha1(alpha1),
+          mAlpha2(alpha2),
+          mActivationAlpha(activationAlpha)
+    {
+    }
+
+    void execute(miopenHandle_t handle, const VariantPack& vpk) final;
+
+    size_t getWorkspaceSize() const final { return size_t{0}; }
```
CAHEK7, line 80: Can it depend on a particular matched engine?

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
```diff
@@ -308,86 +309,183 @@ class GPU_ConvBiasResAddActivation_fwd
...
+                                                                    .build())));
+
+        auto& handle   = get_handle();
+        auto handlePtr = static_cast<miopenHandle_t>(&handle);
+        graphBuilder.setHandle(handlePtr);
+        gr::OpGraph graph;
+        ASSERT_NO_THROW(graph = std::move(graphBuilder).build());
+        std::vector<gr::Engine> engines;
+        ASSERT_THROW(engines = gr::findEngines(&graph), miopen::Exception);
+    }
```
CAHEK7, line 469: Can this test be a part of the common testsuite? Like
```C++
if (input_c % wei_c != 0)
    ASSERT_THROW(engines = gr::findEngines(&graph), miopen::Exception);
    SUCCEED();
    return;
else
    ASSERT_NO_THROW(engines = gr::findEngines(&graph), miopen::Exception);

...the rest of the test...
```
It eliminates extra initialization code.

CAHEK7 (commented):
There are a few questions and recommendations.
I don't block this PR, it can be fixed later, or right now, or even can be ignored.

junliume (approved):

junliume:
@CAHEK7 last round call for reviews, if needed @BrianHarrisonAMD please deal with the review opinions in a follow up PR. Thanks!

src/include/miopen/graphapi/variant_pack.hpp
BrianHarrisonAMD: Since workspace was defaulting to nullptr before, I believe they always had to set it, but now it allows them to set a null value.

src/CMakeLists.txt
BrianHarrisonAMD: Nice catch!

Will do.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: Ah good point.
I can remove this.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: I think in this case it will throw if it doesn't match, (since the code below will fail to find, and explode), and we don't want to pay the runtime cost of checking matches again. I was following how the other pattern matcher was set up, and I think that is the reason anyways.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: I think in this case, it is "applicable", but these exceptions would indicate improperly set up convolution operations. That's why I decided not to move these into the matches check anyways. Basically, if this were to throw, then I would expect any pattern matcher / executor would eventually throw since it's a malformed request.

src/graphapi/find_engine.cpp
BrianHarrisonAMD: Yea that's a good point. There is a lot of virtual checks sprinkled around the pattern checker code, and adding a quick utility class to manage that would be cleaner. 

Thanks!

src/include/miopen/graphapi/conv_bias_res_add_activ_forward_executor.hpp
BrianHarrisonAMD: Eventually probably, but with this current implementation, it will always call the fusion method which doesn't (currently) use workspace.

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
BrianHarrisonAMD: That's a good point, ill try to combine these.

Thanks!

BrianHarrisonAMD: Oh right I remember now, the reason I can't combine these is that it will throw earlier if I try to do the same flow with improperly formed arguments.

I think If I were to try and combine them, then the initialization would diverge quite a bit, and think it's better to keep it like this.

BrianHarrisonAMD:
Code review comments should be addressed now, and just waiting for CI pass. 

Thanks for all the feedback!

CAHEK7 (approved):

junliume:
Updated by resolving conflicts, will merge when CI passes again. Thanks!

junliume (approved):

PR #3170, opened 2024-07-29
Use the cmake variable HALF_INCLUDE_DIR to find half.hpp

trixirt:
On Fedora half.hpp is installed in the system /usr/include dir The use assumes it is installed in half/half.hpp and looks for this path in the cmake configury.  Instead of assuming the path has a half/, pass the found half.hpp path as an include argument -I ${HALF_INCLUDE_DIR}

convert all uses of half/half.hpp with a sed script for f in `find . -type f -name '*.hpp' -o -name '*.cpp' `; do
    sed -i -e 's@#include <half/half.hpp>@#include <half.hpp>@' $f
done

Fixes this issue that was incorrectly closed
https://github.com/ROCm/MIOpen/issues/2672


junliume:
@trixirt In order to make this change complete, we need to update the submodule MIFin too: https://github.com/search?q=repo%3AROCm%2FMIFin%20half&type=code

Actually you can go into ${MIOpen}/fin and pull the submodule, then make necessary changes there.

trixirt:
There are other instances in the ROCm stack that need this fix.
Can you explain why this change to MIFin would need to happen as a prerequisite ?
MIFin project does not have a Readme, what is it used for ?

trixirt:
See here
https://github.com/ROCm/MIFin/pull/104

shurale-nkn:
@atamazov

atamazov:
@junliume @trixirt @shurale-nkn I suspect this is overkill, because ROCm does not officially support Fedora, https://rocm.docs.amd.com/en/latest/compatibility/compatibility-matrix.html.

> On Fedora half.hpp is installed in the system /usr/include

> There are other instances in the ROCm stack that need this fix...

Maybe suggesting the simplest workaround for the Fedora users is enough: put a copy of half.hpp into /usr/include/half/. It does not require any modifications in MIOpen, Fin and other components of the ROCm stack.

Otherwise, no objections.

CMakeLists.txt
```diff
@@ -470,8 +470,9 @@ endif()
 add_definitions(-DBOOST_ALL_NO_LIB=1)
 find_package(Boost REQUIRED COMPONENTS ${BOOST_COMPONENTS})
 
-find_path(HALF_INCLUDE_DIR half/half.hpp)
+find_path(HALF_INCLUDE_DIR half.hpp HINTS ${CMAKE_INSTALL_PREFIX}/include ${CMAKE_INSTALL_PREFIX}/include/half )
 message(STATUS "HALF_INCLUDE_DIR: ${HALF_INCLUDE_DIR}")
+string(APPEND CMAKE_CXX_FLAGS " -I${HALF_INCLUDE_DIR}")
```
junliume: @trixirt @atamazov I am not sure if I like this. I hope that we do not need to modify `CMAKE_CXX_FLAGS` in such a way even though we are only appending to it.

trixirt:
> @junliume @trixirt @shurale-nkn I suspect this is overkill, because ROCm does not officially support Fedora, https://rocm.docs.amd.com/en/latest/compatibility/compatibility-matrix.html.
> 
> > On Fedora half.hpp is installed in the system /usr/include
> 
> > There are other instances in the ROCm stack that need this fix...
> 
> Maybe suggesting the simplest workaround for the Fedora users is enough: put a copy of half.hpp into /usr/include/half/. It does not require any modifications in MIOpen, Fin and other components of the ROCm stack.
> 
> Otherwise, no objections.

This thrashes the location of the header and causes confusion for the users of the header.  The AMD version of half itself does this, the older version installed it to opt/rocm/include, the newer opt/rocm/include/half.

trixirt:
Removed the global change to cxx flags and am using target_include_directories similar to MIFin
Changed signoff to new email account.

BradPepersAMD:

PR #3173, opened 2024-07-30
Add cmake MIOPEN_PARALLEL{COMPILE,LINK}_JOBS

trixirt:
Copied from LLVM cmake variables LLVM_PARALLEL{COMPILE,LINK}_JOBS, adds the option for the ninja generator of splitting the jobs pool between the compile and link jobs.  This allows the user to fine tune the jobs and can help with trashing memory on linking by limiting the number of linkers while keeping the number of compilers high.


shurale-nkn:
LLVM used similar code to override the default generator values for its specific needs.  In our project, we don't have such requirements. 
I don't see the need to add two new variables to the project that will essentially act as proxies for generator values. This will only create an unnecessary layer between the user and the generator.

BradPepersAMD:

PR #3175, opened 2024-07-31
[gtest] Unit test for ConvWinoFuryRxS

averinevg:
This PR adds unit test for `ConvWinoFuryRxS`.

Additional changes:

- Old `test_drive()`-based test has been removed
- Test builds have been speed up by about 3x by adding a common static library
- Several fixes in `ConvFwdSolverTestBase`, `ConvFwdSolverTest`, `ConvBwdSolverTest` (critical) and `ConvWrwSolverTest` (critical)
- Added a general framework for convolution solver unit test
- Fixed BatchNorm gtests on failure
- Added solver device applicability test


averinevg:
@atamazov Please review

src/include/miopen/solver.hpp
```diff
@@ -157,6 +157,9 @@ struct SolverMixin : SolverBase
     virtual float GetWti(const Context&, const Problem&) const { return wti_approximate_worst; };
     virtual size_t GetWorkspaceSize(const Context&, const Problem&) const { return 0; };
 
+    /// Returns "default" solution
+    virtual ConvSolution GetDefaultSolution(const Context&, const Problem&) const = 0;
```
atamazov: [Q] AFAICS this is necessary only for testing. Previously we had some "complicated" template functions, but these were not visible from outside the tests. Now we have something visible everywhere but intended only for tests (which is not good I suspect but maybe I am wrong). What are the merits of this change, maybe I just unable to see them?

Maybe we can re-use it in the library instead of constructs like 
```C++
...s.GetSolution(context, problem, s.GetDefaultPerformanceConfig(context, problem))...
```

But in this case I would recommend removing of `GetDefaultSolution` from non-tunable solvers.

test/gtest/solver_bwd.hpp
```diff
@@ -66,12 +80,11 @@ struct ConvBwdSolverTest
...
         }();
 
-        // const auto network_config = problem.BuildConfKey();
-
         if(!solv.IsApplicable(ctx, problem))
         {
-            test_skipped = true;
-            GTEST_SKIP() << solv.SolverDbId() << ": Not Applicable for this problem" << conv_config;
+            // Do not put GTEST_SKIP here! An inappropriate config should cause the test to fail,
+            // not skip. Otherwise, such testing is pointless.
```
atamazov:
```suggestion
            // The usage of non-applicable config should be considered as a bug in the test.
```

test/gtest/solver_bwd.hpp
atamazov, line 1: [Q] Is this file intended to use only in the solver "unit" (or "smoke") tests? And only for convolution solvers?

test/gtest/solver_bwd.hpp
```diff
@@ -28,32 +28,46 @@
...
 struct ConvBwdSolverTest
     : public ::testing::TestWithParam<std::tuple<miopenConvFwdAlgorithm_t, ConvTestCaseBase>>
 {
+    void SolverBwd(const miopen::solver::conv::ConvSolverBase& solv)
+    {
+        // SetUpTest() and TearDownTest() are moved here so that if a test is skipped, time is not
+        // wasted.
+        this->SetUpTest();
+        this->RunSolver(solv);
+        this->TearDownTest();
```
atamazov: Please comment out why `ASSERT_*` shouldn't be used in the _body_ of this function (after SetUpTest and before TearDownTest). Or, better, follow this: https://github.com/ROCm/MIOpen/pull/3175#discussion_r1703291094

test/gtest/conv_test_base.hpp
```diff
@@ -174,26 +172,27 @@ struct ConvFwdSolverTestBase
...
+        ASSERT_TRUE(error < threshold)
             << "Error beyond tolerance Error:" << error << ",  Threshold: " << threshold;
     }
 
     miopen::ConvolutionDescriptor conv_desc;
     tensor<T> input;
     tensor<T> weights;
     tensor<T> output;
-    tensor<T> ref_out;
+    tensor<Tref> ref_out;
```
atamazov, line 194: 👍

test/gtest/gtest_common.hpp
```diff
@@ -114,6 +114,28 @@ bool IsTestSupportedForDevMask()
     return res;
 }
 
+inline Gpu GetDevGpuType()
```
atamazov: The idea of existing `IsTestSupportedForDevMask` it to mimic behavior of CTest (it supports the same defaults). You can convert existing `*_DISABLED` and `*_ENABLED` from there to `enabled<Gpu::...>` and `disabled<Gpu::...>`, and use the latter with `IsTestSupportedForDevMask`. Why this approach can't be used in your code?

atamazov: I think you can reuse this function in `IsTestSupportedForDevMask` like this:
```C++
return (mask & GetDevGpuType()) != 0;
```
instead of stuff at lines 96-114.

atamazov: [optimization] Rename to `GetDevGpuType_impl` and add 
```C++
Gpu GetDevGpuType()
{
    static const auto rv = GetDevGpuType_impl();
    return rv;
}
```
or use lambda. This would avoid repeating of string comparisons.

test/gtest/solver_bwd.hpp
```diff
@@ -28,32 +28,46 @@
...
 #include "../workspace.hpp"
 
 template <typename T = float, typename Tref = float, bool use_cpu_ref = false>
 struct ConvBwdSolverTest
     : public ::testing::TestWithParam<std::tuple<miopenConvFwdAlgorithm_t, ConvTestCaseBase>>
 {
+    void SolverBwd(const miopen::solver::conv::ConvSolverBase& solv)
+    {
+        // SetUpTest() and TearDownTest() are moved here so that if a test is skipped, time is not
+        // wasted.
```
atamazov: ...But actually I highly recommend keeping the functionality of `SetUp` and `TeadDown`, BUT skipping the test right in SetUp, see http://google.github.io/googletest/reference/testing.html#GTEST_SKIP

test/gtest/unit_solver_ConvWinoFuryRxS.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+INSTANTIATE_TEST_SUITE_P(Unit,
+                         GPU_UnitTestConvSolver_fwd_fp16,
+                         testing::Combine(testing::Values(miopenConvolutionFwdAlgoWinograd),
+                                          testing::ValuesIn(GetConvTestCases()),
+                                          testing::Values(miopenTensorNCHW)));
+
+INSTANTIATE_TEST_SUITE_P(Unit,
+                         GPU_UnitTestConvSolver_bwd_fp16,
+                         testing::Combine(testing::Values(miopenConvolutionFwdAlgoWinograd),
```
atamazov: This looks suspicious:
```
... GPU_UnitTestConvSolver_bwd_fp16 ... miopenConvolutionFwdAlgoWinograd
                           ^^^                           ^^^
```

atamazov (changes requested):
We have something to discuss.

src/include/miopen/solver.hpp
averinevg: The idea is to get a `default` solution using a base class, without templates and all these over-engineered helper functions, regardless of whether solver is tunable or not.
There are a lot of places in the library where the solution is obtained in this way, and of course we can use this function there.

test/gtest/solver_bwd.hpp
averinevg: We should keep `GTEST_SKIP` mentioned here, because it is a fundamental point. We have a bunch of tests that are simply skipped with inapplicable configs.

test/gtest/solver_bwd.hpp
averinevg: [A] Yes, it is.

src/include/miopen/solver.hpp
atamazov: [resolved]

test/gtest/solver_bwd.hpp
atamazov: @averinevg Let's combine both comments.

test/gtest/solver_bwd.hpp
atamazov: [resolved] Let's use more appropriate filenames in some of you future PRs.

test/gtest/solver_bwd.hpp
atamazov: [resolved]

CAHEK7: @atamazov I didn't get why it is resolved.

test/gtest/conv_test_base.hpp
```diff
@@ -174,26 +172,27 @@ struct ConvFwdSolverTestBase
...
         double threshold       = std::numeric_limits<T>::epsilon() * tolerance;
         auto error             = miopen::rms_range(ref_out, output);
 
-        EXPECT_FALSE(miopen::find_idx(ref_out, miopen::not_finite) >= 0)
+        ASSERT_FALSE(miopen::find_idx(ref_out, miopen::not_finite) >= 0)
             << "Non finite number found in the CPU data";
 
-        EXPECT_TRUE(error < threshold)
+        ASSERT_TRUE(error < threshold)
             << "Error beyond tolerance Error:" << error << ",  Threshold: " << threshold;
```
CAHEK7: I'm not quite sure that we have to stop verification after the first encountered numeric error. Probably I'd like to see multiple failed assertions with extra information what went wrong.

Btw, since you've already touched the assertions, I'd recommend using appropriate assertion forms, like
```diff
- ASSERT_TRUE(miopen::range_distance(ref_out) == miopen::range_distance(output));
+ ASSERT_EQ(miopen::range_distance(ref_out), miopen::range_distance(output));
```

```diff
- ASSERT_FALSE(miopen::find_idx(ref_out, miopen::not_finite) >= 0)
+ ASSERT_LT(miopen::find_idx(ref_out, miopen::not_finite), 0)
```

```diff
- ASSERT_TRUE(error < threshold)
- << "Error beyond tolerance Error:" << error << ",  Threshold: " << threshold;
+ ASSERT_LT(error, threshold) << "Error beyond tolerance";
```
`ASSERT_LT` prints its values, while `ASSERT_TRUE` does not.

(the same recommendation is for the other solvers)

test/gtest/solver_bwd.hpp
```diff
@@ -175,6 +181,6 @@ struct ConvBwdSolverTest
     miopen::Allocator::ManageDataPtr wei_dev;
     miopen::Allocator::ManageDataPtr out_dev;
     Workspace wspace{};
-    miopenConvFwdAlgorithm_t algo = miopenConvolutionFwdAlgoDirect;
-    bool test_skipped             = false;
+    miopenConvAlgorithm_t algo = miopenConvolutionAlgoDirect;
+    bool test_skipped          = false;
```
CAHEK7: Is it ever used?
At the first glance you've already removed the case where we skip it in the middle of the execution, so I'd suggest removing `test_skipped` at all.
Also I'd recommend moving skipping logic as close to `SetUp`\`TestSetUp` as possible, because those `SetUp` routines do a lot of memory copy and allocations and tensor initializations which will be thrown away later.

(the same recommendation is for the other solvers)

test/gtest/unit_conv_solver.cpp
```diff
@@ -23,34 +23,15 @@
...
+bool UnitTestConvSolver::CheckTestSupportedForDevice(Gpu supported_gpus)
 {
-    auto solution = GetSolutionImpl(miopen::rank<1>{}, s, ctx, problem);
-    return solution;
+    auto gpu = GetDevGpuType();
+    if(static_cast<int>(gpu) & static_cast<int>(supported_gpus))
+    {
+        return true;
+    }
+    return false;
```
CAHEK7: I'd suggest to pass `supported_gpu` into helper and reduce number of static casts in the tests.
```suggestion
    return IsTestSupportedByDevice(supported_gpus);
```

CAHEK7 (commented):
There are a few non-critical recommendations.

test/gtest/solver_bwd.hpp
atamazov: @CAHEK7 because, as I've discussed today with @averinevg, https://github.com/ROCm/MIOpen/pull/3175#discussion_r1703291094 will be implemented.

test/gtest/conv_test_base.hpp
atamazov: Yeah, it's good to see all failures when something is wrong.

CAHEK7 (approved):

test/gtest/conv_test_base.hpp
atamazov: @CAHEK7 Is it resolved? I still see many replacements of EXPECTs by ASSERTs in this PR.

junliume:
@averinevg we have `Dbsync gfx90a` test issue and please inform @cderb about tuning updates which need to come together with this PR. Thanks!

averinevg:
> @averinevg we have `Dbsync gfx90a` test issue and please inform @cderb about tuning updates which need to come together with this PR. Thanks!

@junliume There is no need to update databases

test/gtest/db_sync.cpp
```diff
@@ -778,12 +782,20 @@ struct TestHandle : Handle
...
+// happened, and the test broke. I disabled that part (since it doesn't work as intended anyway) to
+// keep its behavior the same.
+#if 0
     std::size_t GetMaxComputeUnits() const
     {
         if(num_cu == 0)
             return Handle::GetMaxComputeUnits();
         return num_cu;
     }
+#endif
```
averinevg, line 797: @JehandadKhan  @cderb Please take a look at this. This code has been added in https://github.com/ROCm/MIOpen/pull/2407.

cderb: This should have enabled testing of other skews on the same arch, though it appears that the override to the handle method here did not apply to GetDbBasename. So as in the note, this did not allow for the substitution of cu count. Good catch.

cderb (approved):
LGTM

junliume (approved):

Issue #2828, opened 2024-03-21
Reordering dimensions using miopenTransformTensor is broken


t4c1:
Attaching two reproducers:
 - The first one creates 2-D tensors with `miopenSetTensorDescriptor`. Attempting to change the order of dimensions copies the data, but it is not reordered.
 - The second one creates 4-D tensors with `miopenSetNdTensorDescriptorWithLayout` and attempts to change order from NCHW to NHWC. In this case no data seems to be copied to the second tensor.
 
 Reproducer 1:
```
#define  __HIP_PLATFORM_AMD__
#include <miopen/miopen.h>

#include <vector>
#include <iostream>

int main(int argc, char* argv[]) {
    miopenHandle_t handle;

    miopenCreate(&handle);

    auto dt = miopenDataType_t::miopenFloat;
    auto dt_size = sizeof(float);

    constexpr int ndims = 2;
    int src_dims[ndims]={2, 2};
    int dst_dims[ndims]={2, 2};
    int src_strides[ndims] = {1, 2};
    int dst_strides[ndims] = {2, 1};

    constexpr int n_tensors = 2;
    miopenTensorDescriptor_t tensor_descs[n_tensors] = {};
    miopenCreateTensorDescriptor(&tensor_descs[0]);
    miopenCreateTensorDescriptor(&tensor_descs[1]);
    miopenSetTensorDescriptor(tensor_descs[0], dt, ndims, src_dims, src_strides);
    miopenSetTensorDescriptor(tensor_descs[1], dt, ndims, dst_dims, dst_strides);

    const float alpha = 1.f, beta = 0.f;
    bool is_training = true;

    size_t ws_size_miopen = 0;
    size_t x_size_bytes = 0;

    void * src_dev;
    void * dst_dev;
    hipMalloc(&src_dev, src_dims[0] * src_dims[1] * dt_size);
    hipMalloc(&dst_dev, dst_dims[0] * dst_dims[1] * dt_size);
    std::vector<float> src = {1,2,3,4};
    hipMemcpy(src_dev, src.data(), 4 * dt_size, hipMemcpyDefault);

    miopenTransformTensor(handle, &alpha, tensor_descs[0], src_dev, &beta, tensor_descs[1], dst_dev);

    hipDeviceSynchronize();

    std::vector<float> dst(4);
    hipMemcpy(dst.data(), dst_dev, 4 * dt_size, hipMemcpyDefault);
    std::cout << "actual result: ";
[...]

PR #3184, opened 2024-08-06
[Fix][TransformTensor] Ignore output buffer when BETA=0

atamazov:
The primitive produces invalid results when BETA=0 and output buffer contains junk (NaNs). This PR fixes the issue. 

By-products:
- The fix eliminates reads of output buffer when BETA=0, so some perf gain is expected.
- Small perf gain of kernels for cases when ALPHA=1.
- Removal of some unused code.

Related issue:
- #2828

----
- https://github.com/ROCm/MIOpen/labels/urgency_blocker
- Proposed reviewers:
  - @JehandadKhan 
  - @junliume
  - @t4c1
  - @ShanoToni
- Optional reviewers:
  - @xdgarrido


CAHEK7 (approved):
It would be nice to have some perf measurements.

atamazov:
@CAHEK7 Maybe later (I do not have perf tests for this primitive on hand). This PR is about correctness and I am quite sure that is doesn't lead to perf degradations (hope you sure too). Also I suspect that the most of time on GPU is spent for address/offset computations, so the expected perf gain is pretty small.

src/kernels/MIOpenSubTensorOpWithTransformKernel.cl
```diff
@@ -100,13 +100,42 @@
...
+    do                                         \
+    {                                          \
+        (dst) = (src)*alpha;                   \
+        (void)beta;                            \
+    } while(0)
+#elif MIOPEN_ALPHA_IS_ONE
+#define SUBTENSOR_OP_WITH_ALPHA_BETA(dst, src) \
+    do                                         \
+    {                                          \
+        (dst) = (src) + (dst)*beta;            \
```
atamazov: @junliume @CAHEK7 Maybe
```suggestion
        (dst) = mad((dst), beta, (src));       \
```
what do you think? This should produce one inst instead of two.

junliume (approved):

Issue #2860, opened 2024-03-29
NHWC miopenOpTensor Add incorrect results


ShanoToni:
While trying to do a NHWC conv bias I encountered incorrect results some investigation showed that the miopenOpTensor Add operation produced incorrect results for NHWC format(Works correctly for NCHW).
The used MIOpen library is part of the rocm-5.6.1 package.
I could not find specification regarding the supported formats for miopenOpTensor.
Is this a known/expected limitation?
Additionally looking for similar issues this #2001 seemed related and pointed NHWC not working correctly and a reorder of the tensor data would be the workaround.
#2828 shows reorders not to be working. 
What would the solution be? Is there a workaround to this?
I have attached a reproducer showing this behavior:
```cpp
#define __HIP_PLATFORM_AMD__
#include <vector>

#include <iostream>
#include <miopen/miopen.h>

int main(int argc, char *argv[]) {
  miopenHandle_t handle;

  miopenCreate(&handle);

  auto dt = miopenDataType_t::miopenFloat;
  auto dt_size = sizeof(float);

  miopenTensorDescriptor_t desc_a = {};
  miopenTensorDescriptor_t desc_b = {};
  miopenTensorDescriptor_t desc_bias = {};

  miopenCreateTensorDescriptor(&desc_a);
  miopenCreateTensorDescriptor(&desc_b);
  miopenCreateTensorDescriptor(&desc_bias);

  int a_dims[4] = {1, 2, 1, 5};      // nchw
  int a_strides[4] = {10, 1, 10, 2}; // nchw -> nhwc
  int b_strides[4] = {10, 5, 5, 1};  // nchw -> nchw

  int bias_dims[4] = {1, 2, 1, 1};    // nchw
  int bias_strides[4] = {2, 1, 2, 2}; // nchw

  int ndims = 4;

  miopenSetTensorDescriptor(desc_a, dt, ndims, a_dims, a_strides);
  miopenSetTensorDescriptor(desc_b, dt, ndims, a_dims, b_strides);
  miopenSetTensorDescriptor(desc_bias, dt, ndims, bias_dims, bias_strides);

  size_t ws_size_miopen = 0;
  size_t a_size_bytes = 0;

  void *a_dev;
  void *b_dev;
  void *bias_dev;
  void *out_dev;

  auto a_size = a_dims[0] * a_dims[1] * a_dims[2] * a_dims[3] * dt_size;
  auto bias_size =
[...]

PR #3190, opened 2024-08-09
[DRAFT] Fix op tensor nhwc

atamazov:
WIP for #2860


atamazov:
@CAHEK7 @averinevg For early preview

CAHEK7:
Tensor operations are layout agnostic, the most important limitation is all the tensors must have the same layout. Another limitation from the current implementation - all the kernels, even the generic one, do not support strides in the last dimension, so even if the layout is not default, but the last dimension is tightly packed, we can use most of the kernels (in the other words, we must check that the last dimension stride is 1 and skip any layout checks).
Or if the layout is transposed using strides, we should rearrange lengths and strides for all the tensors to make tightly packed dimension the last.
Layout information is required only for keeping all the tensors in the same layout.

Technically we even can omit this check `if(lay_a != lay_b)` if we can normalize tensor B layout and still have the last non-singular dimension's stride equals to 1.
For example, for NCHW tensor A we can apply any NXXX tensor B if it has the following dimensions: N111. Because there is only 1 scalar value applied to the whole CHW instance.

atamazov:
🛠️ Detection of "mismatched tensor layouts" cannot help to resolve the problem. The PR will be reworked.

CAHEK7:
This PR decided to be closed, because there is another approach to fix that problem.

PR #3181, opened 2024-08-06
Update docker ROCm version for CI

junliume:
Update CK commit hash to its latest `amd-develop` for staging

PR #3191, opened 2024-08-09
Update CK without new base ROCm renew

junliume:
Another form of #3181 without updating base ROCm version (and the potential issues come with it)


carlushuang (approved):
LGTM

Issue #3192, opened 2024-08-10
[CI][Failure] shared_ptr_base.h:199:9: runtime error: member call on address which does not point to an object of type 'std::_Sp_counted_base<>'


junliume:
Another byproduct of #3181

[LastTest.log](https://github.com/user-attachments/files/16567842/LastTest.log)

The error message:
```
/usr/lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/shared_ptr_base.h:199:9: runtime error: member call on address 0x00000b9e6590 which does not point to an object of type 'std::_Sp_counted_base<>'
0x00000b9e6590: note: object has invalid vptr
 00 00 00 00  d8 c0 dd 8e 53 7f 00 00  00 00 00 00 02 00 00 00  d9 01 00 00 00 00 00 00  30 d5 ac 10
              ^~~~~~~~~~~~~~~~~~~~~~~
              invalid vptr
    #0 0x7f53852a1bc7  (/data/MIOpen/build/lib/libMIOpen.so.1+0x29a5ebc7)
    #1 0x7f538e2c87cb  (/data/MIOpen/build/lib/libMIOpen.so.1+0x32a857cb)
    #2 0x7f530f9c2d9e  (/lib/x86_64-linux-gnu/libc.so.6+0x45d9e) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
    #3 0x7f530f9c25c8  (/lib/x86_64-linux-gnu/libc.so.6+0x455c8) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
    #4 0x7f530f9c260f  (/lib/x86_64-linux-gnu/libc.so.6+0x4560f) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
    #5 0x7f530f9a6d96  (/lib/x86_64-linux-gnu/libc.so.6+0x29d96) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
    #6 0x7f530f9a6e3f  (/lib/x86_64-linux-gnu/libc.so.6+0x29e3f) (BuildId: 490fef8403240c91833978d494d39e537409b92e)
    #7 0x249004  (/data/MIOpen/build/bin/test_find_db+0x249004)
```

**[How to reproduce]**:

cmake command:
```
[...]

PR #3198, opened 2024-08-13
Suppress MLIR ubsan errors

BrianHarrisonAMD:
Add suppressing ubsan errors for the file shared_ptr_base.h to workaround errors when upgrading to ROCm 6.2 in the base docker. 

See relevant PR #3181, and Issue #3192 for additional details.

The issue can be tracked down to simply creating the MLIR handle after upgrading to the latest ROCm 6.2.
Since we froze MLIR to an older release, suppressing this warning is our only real option.

Note: 
- Params used when constructing the MLIR handle to replicate, "--x2 1 --operation conv2d --kernel_id 0 --num_cu 104 --arch amdgcn-amd-amdhsa:gfx90a:sramecc+:xnack- --groupsize 1 --fil_layout GNCHW --fil_type fp32 --in_layout NGCHW --out_layout NGCHW --in_type fp32 --out_type fp32 --batchsize 100 --in_channels 25 --out_channels 300 --in_h 32 --in_w 32 --out_h 30 --out_w 30 --fil_h 3 --fil_w 3 --dilation_h 1 --dilation_w 1 --conv_stride_h 1 --conv_stride_w 1 --padding_h 0 --padding_w 0 --kernel_name mlir_gen_igemm_conv2d_v4r4_fwd_xdlops0". Simply creating a handle with those command line arguments will result in the vptr issue when the test application is exiting after upgrading to ROCm 6.2. 
- ROCm 6.2 also updated the version of clang.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ def cmake_build(Map conf=[:]){
 
     def compiler = conf.get("compiler","/opt/rocm/llvm/bin/clang++")
     def make_targets = conf.get("make_targets","check")
-    def debug_flags = "-g -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=undefined -Wno-option-ignored " + conf.get("extradebugflags", "")
+    def debug_flags = "-g -fno-omit-frame-pointer -fsanitize=undefined -fno-sanitize-recover=undefined -Wno-option-ignored -fsanitize-ignorelist=$(pwd)/../UBSanSuppress.supp"  + conf.get("extradebugflags", "")
     def build_envs = "CTEST_PARALLEL_LEVEL=4 " + conf.get("build_env","")
     def prefixpath = conf.get("prefixpath","/opt/rocm")
     def mlir_args = " -DMIOPEN_USE_MLIR=" + conf.get("mlir_build", "ON")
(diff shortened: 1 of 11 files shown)


junliume (approved):
Let's keep track of this WA and minimize the fix later:
> shared_ptr_base.h

atamazov:
@junliume maybe we can simply disable MLIR in the failing stage as a W/A?

BrianHarrisonAMD:
I think that could also be a reasonable workaround, but not sure which way we prefer.
Both aren't ideal hehe.

amberhassaan:
I slightly prefer @atamazov 's suggestion because currently we're saying all best are off with `shared_ptr_base` which could mean that certain memory problems related to `shared_ptr` sneak through, while disabling MLIR means that we aren't sure of MLIR's safety so that's more contained in a way to MLIR as opposed to every use of `std::shared_ptr`.

BrianHarrisonAMD:
I can update this to instead turn off MLIR for the debug steps.

@junliume that good with you?

DrizztDoUrden:
I think disabling MLIR for the sanitized steps as Artem suggested is less harmful because it is at least somewhat tested anyway and we would really not want to miss some shared_ptr misuse in the library if that would happen.

junliume:
> I can update this to instead turn off MLIR for the debug steps.
> 
> @junliume that good with you?

@BrianHarrisonAMD yes let's do it by only disabling checks when MLIR is in use. Thanks!

BrianHarrisonAMD:
Sounds good!

Ill close and make a new PR with disabling MLIR for the sanitize steps.

Thanks for the feedback everyone!

PR #3203, opened 2024-08-19
fix bn layout network config caching issue

bghimireamd:
pytorch's batch norm was failing because incorrect layout (NCHW was being cached when NHWC). This PR is to fix the issue.

### DIFF
--- a/src/include/miopen/batchnorm/problem_description.hpp
+++ b/src/include/miopen/batchnorm/problem_description.hpp
@@ -73,11 +73,11 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
           resultsave(resultsave_),
           resultrunning(resultrunning_)
     {
-        in_layout  = xDesc.GetLayout(xDesc.GetLengths().size() == 4 ? "NCHW" : "NCDHW");
-        out_layout = yOrDyDesc.GetLayout(yOrDyDesc.GetLengths().size() == 4 ? "NCHW" : "NCDHW");
+        in_layout  = ComputeInLayout();
+        out_layout = ComputeOutLayout();
     }
 
-    // Forward
+    // Forward Inference
     ProblemDescription(miopenBatchNormMode_t bn_mode_,
                        const TensorDescriptor& xDesc_,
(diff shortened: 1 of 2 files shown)


aserio:
@junliume @DrizztDoUrden, can we get this PR reviewed quickly? It is blocking a feature that needs to be merged ASAP.

src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -222,10 +222,47 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
     std::string in_layout  = "NCHW";
     std::string out_layout = "NCHW";
     std::string din_layout = "NCHW";
+    int spatial_dim        = 2;
```
averinevg: Looks like the number of dimensions is now hardcoded. How will  3d cases be handled?

src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -222,10 +222,47 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
...
+    {
+        if(spatial_dim == 2)
+        {
+            return dxDesc.GetLayout("NCHW");
+        }
+        else
+        {
+            return dxDesc.GetLayout("NCDHW");
+        }
+    }
```
averinevg:
```suggestion
    std::string ComputeLayout(const TensorDescriptor& td) const
    {
        if(spatial_dim == 2)
        {
            return td.GetLayout("NCHW");
        }
        else
        {
            return td.GetLayout("NCDHW");
        }
    }

    std::string ComputeInLayout() const
    {
        return ComputeLayout(xDesc);
    }

    std::string ComputeOutLayout() const
    {
        return ComputeLayout(yOrDyDesc);
    }

    std::string ComputeDinLayout() const
    {
        return ComputeLayout(dxDesc);
    }
```

averinevg (changes requested):

averinevg (approved):
LGTM

junliume (approved):

PR #3212, opened 2024-08-22
Changed all perf-db creations to be lazy

DrizztDoUrden:
This makes find or similar invocations to not construct any db instance when no tunable solver is applicable or in similar cases. Basically it removes redundant prefetches. After #3078 this can also delay any potential waiting for the db being prefetched, which can reduce expected delays even further.


src/include/miopen/mlo_internal.hpp
```diff
@@ -167,6 +167,9 @@ size_t setBotDfDescFromMLDesc(int spatial_dims, TTo& to, const TensorDescriptor&
     return SetDescFromMLDesc(spatial_dims, to, tensor, &TTo::setBotDfDescr);
 }
 
+MIOPEN_INTERNALS_EXPORT auto MakeConvDbGetter(const ExecutionContext& ctx)
+    -> std::function<PerformanceDb&()>;
+
```
averinevg: [R] I think we shouldn't add new things into this file. This file has confusing name, also it probably has a license problem.

src/include/miopen/any_solver.hpp
```diff
@@ -94,7 +95,18 @@ struct AnySolver
     {
         assert(ptr_value != nullptr);
         return ptr_value->FindSolution(ctx, problem, db, invoke_ctx, perf_cfg);
-    };
+    }
+
+    ConvSolution FindSolution(const ExecutionContext& ctx,
```
averinevg, line 115: Let's add `#define FIN_ANY_SOLVER_FIND_SOLUTION_COMPAT 1` and wrap old `FindSolution()` with `#if/#endif`.
This code should be removed after the `Fin` is fixed.

averinevg (changes requested):

shurale-nkn (approved):
LGTM

averinevg (approved):
LGTM

PR #3213, opened 2024-08-22
TensorDescriptor refactoring & enhancements

averinevg:
TensorDescriptor:

- IsPossibleLayout() for non-packed tensors has been fixed
- Added IsPossibleLayout4D5D()
- Proper GetLayout_t() calculation
- Added GetLayoutEnum()
- Added check for descriptor with layout and strides
- Replaced GetLayoutStr() with LayoutEnumToStr()
- Added unit tests
- Some refactoring

test_tensor_api:

- Added miopenSetTensorDescriptorV2()

conv::ProblemDescription:

- Layout calculation has been fixed
- Added some unit tests
- Some refactoring


src/tensor.cpp
```diff
@@ -111,6 +115,21 @@ void ReorderVector(std::vector<size_t>& lens, const std::initializer_list<size_t
     lens = std::move(out_lens);
 }
 
+std::optional<miopenTensorLayout_t> GetDefaultLayout(unsigned num_dims)
+{
+    // clang-format off
```
CAHEK7: Does it mean that we have to adjust clang-format rules. All of the `switch-case` statements in this PR have `clang-format off`.

I would strongly recommend to remove all the `clang-format off` from the switch-cases.

src/include/miopen/tensor.hpp
```diff
@@ -241,76 +245,38 @@ struct MIOPEN_INTERNALS_EXPORT TensorDescriptor : miopenTensorDescriptor
...
                      std::vector<std::size_t>&& strides_in,
                      bool use_strides);
 
     void CheckArgsAndInit(bool use_strides);
 
-    void SetStrideNd(const std::string& layout);
-    void LensReorder(const std::string& layout);
+    void SetStrides();
+    void VectLensReorder();
+    void VectLensRecalc();
```
CAHEK7: Those two methods are used exactly in one context, moreover it's unsafe to use it somewhere else, because it assumes some implicit behavior.
I'd prefer removing them from a global class context.

src/tensor.cpp
```diff
@@ -272,58 +295,105 @@ void TensorDescriptor::CheckArgsAndInit(bool use_strides)
...
+        break;
     }
-    else
+    // clang-format on
+}
+
+void TensorDescriptor::VectLensRecalc()
+{
+    // clang-format off
+    // NOLINTNEXTLINE (bugprone-unchecked-optional-access)
```
CAHEK7: That's unsafe indeed.
The function implies some external checks and specific context and can't be called in a common case.

src/include/miopen/tensor.hpp
```diff
@@ -204,9 +204,13 @@ struct MIOPEN_INTERNALS_EXPORT TensorDescriptor : miopenTensorDescriptor
     unsigned GetNumDims() const;
 
     miopenDataType_t GetType() const;
+    // clang-format off
```
CAHEK7, line 207: Is it necessary? We should either live we current format or modify `.clang-format`.
Disabled format should be justified.

src/include/miopen/tensor.hpp
```diff
@@ -323,7 +289,17 @@ struct MIOPEN_INTERNALS_EXPORT TensorDescriptor : miopenTensorDescriptor
...
+
+    // For GetLayoutEnum()
+    mutable std::optional<miopenTensorLayout_t> cached_layout_enum;
+    mutable bool cached_layout_enum_calculated = false;
+
+    // For GetLayout_str()
+    mutable std::string cached_layout_str;
+
+    // For GetLayout
+    mutable std::vector<int64_t> cached_permutation;
```
CAHEK7, line 299: I want to clarify: `mutable` is for a lazy evaluation, am I correct?

src/include/miopen/tensor.hpp
```diff
@@ -241,76 +245,38 @@ struct MIOPEN_INTERNALS_EXPORT TensorDescriptor : miopenTensorDescriptor
...
-    {
-        std::vector<std::int64_t> result(lens.size());
-        std::iota(result.begin(), result.end(), 0);
-        std::stable_sort(result.begin(), result.end(), by(std::greater<>{}, [&](auto x) {
-                             return std::make_tuple(strides[x], lens[x]);
-                         }));
-        return result;
-    }
+    static std::vector<int64_t> find_permutation(const std::vector<std::size_t>& lens,
+                                                 const std::vector<std::size_t>& strides);
```
CAHEK7, line 254: [R] It can be a local lambda, like for the other `cached_` variables.

CAHEK7 (commented):
```C++
    void VectLensReorder();
    void VectLensRecalc();
```
question is kind of important and I would like to discuss it. (It doesn't block merging, but probably will be changed in following PRs)
Clang-format looks very strange, it doesn't seem that exception in each `switch-case` looks reasonable.
The rest is optional.

src/include/miopen/tensor.hpp
averinevg: You are right

src/include/miopen/tensor.hpp
averinevg: It's used in one other place outside of this class.

src/tensor.cpp
averinevg: Currently clang-format produced weird output for such statements. I'll see what I can do about it.

averinevg: I'll fix this later as there will be a lot of changes not related to this PR.

src/include/miopen/tensor.hpp
averinevg: I'll fix this later as there will be a lot of changes not related to this PR.

averinevg: With default formatting the code becomes unreadable. It is not possible to control attributes with clang-tidy-12 which we are using. We could update it to clang-tidy-15 on Ubuntu 22.04 which is currently the minimum supported version. This will allow us to control more options (excluding this one, unfortunately).

CC @junliume

src/include/miopen/tensor.hpp
averinevg: I have no objections, will fix

src/tensor.cpp
averinevg: Will fix

src/tensor.cpp
```diff
@@ -111,6 +158,92 @@ void ReorderVector(std::vector<size_t>& lens, const std::initializer_list<size_t
     lens = std::move(out_lens);
 }
 
+// Relevant for NCHWc and CHWNc
+void VectLensReorder(miopenTensorLayout_t layout, std::vector<size_t>& lens)
```
CAHEK7, line 162: It could be just a lambda right inside the function where it is required.

CAHEK7 (approved):

src/tensor.cpp
junliume: It could be a follow up enhancement later.

junliume (approved):

PR #3215, opened 2024-08-22
GenericSearch Patience

cderb:
Adds environment variable MIOPEN_TUNING_PATIENCE which will allow the user to set the maximum number of performance configurations GenericSearch will iterate through without improvement before quitting.

### DIFF
--- a/src/generic_search.cpp
+++ b/src/generic_search.cpp
@@ -61,11 +61,6 @@ std::chrono::milliseconds GetTuningTimeMax()
     return std::chrono::milliseconds{env::value(MIOPEN_TUNING_TIME_MS_MAX)};
 }
 
-std::size_t GetTuningPatience()
-{
-    return env::value(MIOPEN_TUNING_PATIENCE);
-}
-
 std::size_t GetTuningThreadsMax() { return env::value(MIOPEN_COMPILE_PARALLEL_LEVEL); }
 
 } // namespace solver
--- a/src/generic_search.cpp
+++ b/src/generic_search.cpp
@@ -61,6 +61,11 @@ std::chrono::milliseconds GetTuningTimeMax()
     return std::chrono::milliseconds{env::value(MIOPEN_TUNING_TIME_MS_MAX)};
 }
 
+std::size_t GetTuningPatience()
+{
+    return env::value(MIOPEN_TUNING_PATIENCE);
+}
+
 std::size_t GetTuningThreadsMax() { return env::value(MIOPEN_COMPILE_PARALLEL_LEVEL); }
 
 } // namespace solver
--- a/src/include/miopen/generic_search.hpp
+++ b/src/include/miopen/generic_search.hpp
@@ -398,6 +398,7 @@ auto GenericSearch(const Solver s,
     std::shuffle(all_configs.begin(), all_configs.end(), rng);
     std::size_t n_runs_total = std::min(all_configs.size(), GetTuningIterationsMax());
     all_configs.resize(n_runs_total);
+    std::size_t patience = GetTuningPatience();
 
     if(all_configs.empty())
     {
@@ -443,11 +444,16 @@ auto GenericSearch(const Solver s,
     if(!env::enabled(MIOPEN_DEBUG_COMPILE_ONLY))
     {
         size_t n_current       = 0;
+        size_t last_imprv      = 0;
         auto threads_remaining = total_threads;
         while(true)
         {
             if(n_current >= n_runs_total)
                 break;
+            if(last_imprv >= patience)
+                break;
+
+            last_imprv++;
             MIOPEN_LOG_I2("Waiting for item in queue");
             const auto kinder     = solution_queue.pop();
             auto current_config   = std::get<0>(kinder);
@@ -541,6 +547,7 @@ auto GenericSearch(const Solver s,
                             best_config = current_config;
                             best_time   = elapsed_time;
                             n_best      = n_current;
+                            last_imprv  = 0;
                         }
                         else
                         {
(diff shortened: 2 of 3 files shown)


src/generic_search.cpp
```diff
@@ -61,6 +61,11 @@ std::chrono::milliseconds GetTuningTimeMax()
     return std::chrono::milliseconds{env::value(MIOPEN_TUNING_TIME_MS_MAX)};
 }
 
+std::size_t GetTuningPatience()
+{
+    return env::value(MIOPEN_TUNING_PATIENCE);
+}
+
```
CAHEK7: I'm not quite sure that we need this function. Since we need this "Patience" parameter right inside the one particular method of the one particular class, we can use the environmental variable just in-place, like
```diff
- std::size_t patience = GetTuningPatience();
+ std::size_t patience = env::value(MIOPEN_TUNING_PATIENCE);
```

cderb: I'm ok with changing this.

junliume: @cderb while you adapt that change, please run a clang-format to clear the current CI issue.

CAHEK7 (approved):
The only concern I have is the randomness of tuning, like for the case
```
Alg0:
  tune0
  tune1
  tune2
Alg1:
  tune0
  tune1
```
we've always got statistically bad results, because the cases are grouped by the algorithm.
While for the following case:
```
Alg0:
  tune0
Alg1:
  tune1
Alg0:
  tune2
Alg1:
  tune0
Alg0:
  tune1
```
the result will be much better, because it will not be biased by a single row of bad cases from a single algorithm.

cderb:
@CAHEK7 GenericSearch is randomizing performance config order for the solver: https://github.com/ROCm/MIOpen/blob/b6e2e7d4342a9d3b44e307dac61562eee8a2070a/src/include/miopen/generic_search.hpp#L393-L398
So each call to generic search would look like:
``` 
Alg0
  tune2
  tune0
  tune1
```
 
Would you propose a change to GenericSearch where the solvers are evaluated/tested in an interleaved fashion? This would require something like a re-fashioning of the function to take a list of solvers instead.

averinevg:
@cderb What is the difference between MIOPEN_DEBUG_TUNING_ITERATIONS_MAX and newly added MIOPEN_TUNING_PATIENCE?

CAHEK7:
> @CAHEK7 GenericSearch is randomizing performance config order for the solver:
> 
> https://github.com/ROCm/MIOpen/blob/b6e2e7d4342a9d3b44e307dac61562eee8a2070a/src/include/miopen/generic_search.hpp#L393-L398
> 
> 
> So each call to generic search would look like:
> ```
> Alg0
>   tune2
>   tune0
>   tune1
> ```
> 
> Would you propose a change to GenericSearch where the solvers are evaluated/tested in an interleaved fashion? This would require something like a re-fashioning of the function to take a list of solvers instead.

Random shuffle inside one algorithm is better that no shuffle at all, but if the algorithm has A LOT of tuning cases and is very stable for that particular problem, then we may stick with this algorithm. But probably it is intended. But probably it's an intension to stop tuning if we've found such algorithm.

cderb:
> @cderb What is the difference between MIOPEN_DEBUG_TUNING_ITERATIONS_MAX and newly added MIOPEN_TUNING_PATIENCE?

MIOPEN_DEBUG_TUNING_ITERATIONS_MAX is a hard cap on the number of tuning iterations, MIOPEN_TUNING_PATIENCE is a cap on the # of iterations without improvement. If performance improves, the termination count resets.

cderb:
> Random shuffle inside one algorithm is better that no shuffle at all, but if the algorithm has A LOT of tuning cases and is very stable for that particular problem, then we may stick with this algorithm. But probably it is intended. But probably it's an intension to stop tuning if we've found such algorithm.

This shuffle is meant to break up the configs within the algorithm. To facilitate the random sampling of that algorithm while it is being tuned. This would make it so similar configs would have less spatial proximity. So if an env like MIOPEN_TUNING_PATIENCE or MIOPEN_DEBUG_TUNING_ITERATIONS_MAX is set it is more likely a wider range of configs are sampled.

junliume (approved):

atamazov:
@cderb @CAHEK7 @averinevg @junliume ~Unfortunately, the effectiveness of `MIOPEN_TUNING_PATIENCE` depends on distribution of fast PerfConfigs within virtual container (which, in turn, may depend on the Problem).~

The potential issue is using the number of iterations as a limit. It puts ASM kernels at a disadvantage compared to OCL kernels, and puts OCL kernels at a disadvantage compared to HIP kernels. For example building 100 asm kernels would take ~5 sec, while building 100 HIP kernels may take 5 minutes or more. The number-of-iterations limit that works for HIP will not do any good with ASM but may unnecessarily affect ASM performance.

[Notice] That's why `MIOPEN_DEBUG_TUNING_ITERATIONS_MAX`, as it name suggests, is intended for debugging/testing purposes only.

[Recommendation] Rename `MIOPEN_TUNING_PATIENCE` to `MIOPEN_TUNING_PATIENCE_ITERATIONS_MAX`, or, better, replace it with `MIOPEN_TUNING_PATIENCE_TIME_MS_MAX`.

To me, the most promising approach is:
- Using randomly reordered PerfConfigs at generic level together with `MIOPEN_TUNING_TIME_MS_MAX`.
- We can also try combinations with `MIOPEN_TUNING_PATIENCE_TIME_MS_MAX`.

atamazov (commented):
@cderb 🟢 LGTM but please see https://github.com/ROCm/MIOpen/pull/3215#issuecomment-2325131275

atamazov:
> To me, the most promising approach is:
> 
> * Using randomly reordered PerfConfigs at generic level together with `MIOPEN_TUNING_TIME_MS_MAX`.
> * We can also implement `MIOPEN_TUNING_PATIENCE_TIME_MS_MAX` and try different combinations.

...and AFAICS we already have the first item.

PR #3202, opened 2024-08-16
[Tests]All tests in the same test suite must use the same test fixture class

xinlipn:
miopen_gtest failed on Navi3x, MI200 etc with the following message:

[ ERROR ] /opt/rocm/cget/build/tmp-c4f1a42bcee536579a1248d25169f834201170c187fc6d80e8c510a7d28f6f48/googletest-1.14.0/googletest/src/gtest.cc:2788:: /home/bharriso/Source/MIOpen_WIP/MIOpen/test/gtest/conv_3d.cpp:117: Attempted redefinition of test suite GPU_conv3d_FP32.
All tests in the same test suite must use the same test fixture
class.  However, in test suite GPU_conv3d_FP32, you tried
to define a test using a fixture class different from the one
used earlier. This can happen if the two fixture classes are
from different namespaces and have the same name. You should
probably rename one of the classes to put the tests into different
test suites.
Aborted (core dumped)

Ticket:
https://ontrack-internal.amd.com/projects/LWPMIOPEN/issues/LWPMIOPEN-1023

Log with fix
[namespacefix.log](https://github.com/user-attachments/files/16640211/namespacefix.log)

Issue #3133, opened 2024-07-23
Rename gtests according to the new scheme


CAHEK7:
Naming scheme is proposed [on the wiki](https://github.com/ROCm/MIOpen/wiki/GTest-development)
It blocks #3134

Vsevolod1983:
GTests were renamed quite a few time ago....closing the ticket.

PR #3218, opened 2024-08-24
Add gtest name checker

### LINKED ISSUE #3148: Add regexp check for gtests naming
Parse single binary gtest output and check that all the tests match [the naming scheme](https://github.com/ROCm/MIOpen/wiki/GTest-development).
Verifies https://github.com/ROCm/MIOpen/issues/3133

CAHEK7:
Closes #3148 
Right now it does not force test naming, since most of the tests do not fit into scheme.
Enforcing this check will be a separated PR when #3133 will be done.

Byproduct is a regression test for #3202

@atamazov @Vsevolod1983 @xinlipn

### DIFF
--- a/test/gtest/check_names.py
+++ b/test/gtest/check_names.py
@@ -29,12 +29,17 @@
 import re
 import sys
 import argparse
+from collections import defaultdict
+import logging
+
+logging.basicConfig(level=logging.WARNING)
+logger = logging.getLogger("GTest name checker")
 
 """regexp based on https://github.com/ROCm/MIOpen/wiki/GTest-development#naming"""
-re_prefix = re.compile(r"^((Smoke)|(Full)|(Perf)|(Unit.*))")
-re_hw = re.compile(r"^((CPU)|(GPU))")
+re_prefix = re.compile(r"^((Smoke)|(Full)|(Perf)|(Unit.*))$")
+re_hw = re.compile(r"^((CPU)|(GPU))$")
 re_datatype = re.compile(
-    r"^((FP((8)|(16)|(32)|(64)))|(BFP((8)|(16)))|(I((8)|(32)))|(NONE))"
+    r"^((FP((8)|(16)|(32)|(64)))|(BFP((8)|(16)))|(I((8)|(32)))|(NONE))\.$"
 )
 
 
@@ -54,12 +59,9 @@ def parse_args():
     return args
 
 
-def main():
-    """Main function"""
-    args = parse_args()
-    mismatched_prefix = []
-    mismatched_hw = []
-    mismatched_datatype = []
+def parse_tests(args):
+
+    mismatches = defaultdict(str)
 
     with open(args.list) as fp:
         for line in fp.readlines()[2:]:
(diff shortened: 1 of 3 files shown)


test/gtest/check_names.py
junliume, line 1: Is there a requirement on Python version? If so we can modify CMakeList file to be sure. Hence we need to build MIOpen on some legacy distributions.

junliume (approved):

test/gtest/check_names.py
```diff
@@ -0,0 +1,118 @@
...
+"""gtest name linter"""
+import re
+import sys
+import argparse
+
+"""regexp based on https://github.com/ROCm/MIOpen/wiki/GTest-development#naming"""
+re_prefix = re.compile(r"^((Smoke)|(Full)|(Perf)|(Unit.*))")
+re_hw = re.compile(r"^((CPU)|(GPU))")
+re_datatype = re.compile(
+    r"^((FP((8)|(16)|(32)|(64)))|(BFP((8)|(16)))|(I((8)|(32)))|(NONE))"
```
Vsevolod1983: I16 and I64 are missing

CAHEK7:
@junliume @reidkwja @cderb @JehandadKhan does anybody know why CI did not start automatically?

test/gtest/check_names.py
CAHEK7: It is intentional. As far as I can see, potential I16 and I64 tests should not exist.
I64 test is a part of external library testing targeted to that fact that the library does not support I64. We shouldn't test external libraries so thorough.
I16 tests are partially type agnostic, they are actually should belong to FP16/BF16 testsuite. Layout transpose does not care about the type semantic, only about the size of the type and it should be just an `int32_t` for int and float, `int16_t` for half and bfloat16, `int8_t` for int, fp8 and bpf8.

I'll reorganize the tests later, when I'll try to enforce naming scheme.

test/gtest/check_names.py
CAHEK7: Any python3 version should be fine, but probably even python2 will work. It doesn't use anything version-specific.

test/gtest/check_names.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def main():
+    """Main function"""
+    args = parse_args()
+    mismatched_prefix = []
+    mismatched_hw = []
+    mismatched_datatype = []
+
+    with open(args.list) as fp:
```
reidkwja, line 64: It might be a good idea to wrap this into a function then call the function into main like you did with parse_args()

test/gtest/check_names.py
```diff
@@ -0,0 +1,118 @@
...
+            datatype = re.search(re_datatype, name[-1])
+            if not prefix:
+                mismatched_prefix.append(line)
+            if not hw:
+                mismatched_hw.append(line)
+            if not datatype:
+                mismatched_datatype.append(line)
+
+        if mismatched_prefix:
+            print(
```
reidkwja, line 91: Its better to use pythons standard logger and exception handling instead of print

`
import logging
import mylib
logger = logging.getLogger(__name__)
logger.info('Started')
logger.warning('mismatch')

except ValueError:
    logging.error("Please enter a valid integer.")
`

test/gtest/check_names.py
```diff
@@ -0,0 +1,118 @@
...
+            hw = re.search(re_hw, name[0])
+            datatype = re.search(re_datatype, name[-1])
+            if not prefix:
+                mismatched_prefix.append(line)
+            if not hw:
+                mismatched_hw.append(line)
+            if not datatype:
+                mismatched_datatype.append(line)
+
+        if mismatched_prefix:
```
reidkwja: It's better to use python's standard logger and exception handling instead of print

something like this:

```
import logging
logger = logging.getLogger(name)
logger.info('Started')
logger.warning('mismatch')

except ValueError:
    logging.error("Please enter a valid integer.")
```

test/gtest/check_names.py
reidkwja, line 66: It might be a good idea to wrap this into a function then call the it into main like you did with parse_args()

reidkwja (dismissed):

test/gtest/check_names.py
CAHEK7: I added `logging`, but exceptions are not applicable here, because regexp match/search does not throw any.

test/gtest/check_names.py
CAHEK7: Done

test/gtest/check_names.py
CAHEK7: I checked pyhton2 - it works, so any pthon3 versions should be compatible too.

amberhassaan (approved):
my only thought is that we should create helper functions to generate the `test name` in the correct format. 

e.g., 
```
TEST_P(MakeTestName(X, Y, Z), MakeSuiteName(...))
```

CAHEK7:
> my only thought is that we should create helper functions to generate the `test name` in the correct format.
> 
> e.g.,
> 
> ```
> TEST_P(MakeTestName(X, Y, Z), MakeSuiteName(...))
> ```

Unfortunately it's not about the name in `TEST_P`, it's about class name as well. I'm afraid that C++ code will be much uglier if we introduce one more macro level over the current `TEST_P` - `MakeTestName` must reside in 3 or more places, it will break code navigation (because for IDE point of view there is no class names or even names), and we can't enforce naming scheme there (without some special dark macro magic or even the magic won't help).

I feel like the outcome does not worth the effort,

atamazov (approved):
🟢 Good idea!

CAHEK7:
@junliume Packaging stage passed and worked as expected, any other stages are not affected - it's just a python script for a newly added make target for a specific packaging stage.

PR #3220, opened 2024-08-27
[gTests]Fix GTEST_SKIP kills all gtests in miopen_gtest

xinlipn:
miopen_gtest wouldn't start on e.g. Navi3x as below. This is because GTEST_SKIP() is invoked in a global environment setup. The fix is to skip the test in TEST_P(). Log with the fix is attached at the bottom.

root@aus-navi3x-09:/MIOpen/build/bin# ./miopen_gtest
Running main() from /opt/rocm/cget/build/tmp-c4f1a42bcee536579a1248d25169f834201170c187fc6d80e8c510a7d28f6f48/googletest-1.14.0/googletest/src/gtest_main                                                                                                          .cc
/MIOpen/test/gtest/api_convbiasactiv.cpp:189: Skipped
 Skipping fusion test on unsupported ASIC

PRNG seed: 12345678
[==========] Running 8939 tests from 394 test suites.
 Skipping fusion test on unsupported ASIC

[----------] Global test environment tear-down
[==========] 8939 tests from 394 test suites ran. (0 ms total)
[  PASSED  ] 8939 tests.

  YOU HAVE 133 DISABLED TESTS
  
  Log with the fix on Navi3x
[navi3x.log](https://github.com/user-attachments/files/16757739/navi3x.log)

### DIFF
--- a/test/gtest/api_convbiasactiv.cpp
+++ b/test/gtest/api_convbiasactiv.cpp
@@ -36,6 +36,7 @@
 
 #include "tensor_util.hpp"
 #include "get_handle.hpp"
+#include "gtest_common.hpp"
 
 struct CBATestCase
 {
@@ -64,6 +65,18 @@ struct CBATestCase
     }
 };
 
+bool IsTestSupportedForDevice()
+{
+#if WORKAROUND_ISSUE_2212
+    using namespace miopen::debug;
+    using e_mask = enabled<Gpu::gfx900, Gpu::gfx906, Gpu::gfx908, Gpu::gfx90A, Gpu::gfx103X>;
+    using d_mask = disabled<Gpu::gfx110X, Gpu::gfx94X>;
+    return ::IsTestSupportedForDevMask<d_mask, e_mask>();
+#else
+    return true;
+#endif
+}
+
 GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(GPU_ConvBiasActivFwd_FP32);
 struct GPU_ConvBiasActivFwd_FP32
     : public ::testing::TestWithParam<std::tuple<miopenConvFwdAlgorithm_t, CBATestCase>>
@@ -71,6 +84,10 @@ struct GPU_ConvBiasActivFwd_FP32
 protected:
     void SetUp() override
     {
+        if(!IsTestSupportedForDevice())
+        {
+            GTEST_SKIP();
+        }
         std::tie(algo, cba_config) = GetParam();
         const double double_zero   = 0.0f;
         input   = tensor<float>{cba_config.N, cba_config.C, cba_config.H, cba_config.W};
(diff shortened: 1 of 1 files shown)


test/gtest/api_convbiasactiv.cpp
```diff
@@ -150,46 +150,47 @@ struct GPU_ConvBiasActivFwd_FP32
...
-                                                               &output.desc,
-                                                               out_dev.get());
-    EXPECT_EQ(status, miopenStatusSuccess);
-}
-
-void GatherCBATestCases(std::vector<CBATestCase>& cba_test_cases)
 {
     const auto dev_name = get_handle().GetDeviceName();
 #if WORKAROUND_ISSUE_2212
     if(!miopen::StartsWith(dev_name, "gfx11") && !miopen::StartsWith(dev_name, "gfx94"))
```
BrianHarrisonAMD: Can we move the skip to the start of SetUp(), and use the device flags for skipping instead?

```cpp
void SetUp() override
{
    if(!IsTestSupportedForDevice())
    {
         GTEST_SKIP();
    }

   // remaining Setup
}



bool IsTestSupportedForDevice()
{
#if WORKAROUND_ISSUE_2212
    using namespace miopen::debug;
    using e_mask = enabled<Gpu::gfx900, Gpu::gfx906, Gpu::gfx908, Gpu::gfx90A, Gpu::gfx103X>;
    using d_mask = disabled<Gpu::gfx110X, Gpu::gfx94X>;
   return ::IsTestSupportedForDevMask<d_mask, e_mask>();
#else
    return true;
#endif
}
```

test/gtest/api_convbiasactiv.cpp
```diff
@@ -150,46 +150,47 @@ struct GPU_ConvBiasActivFwd_FP32
...
 
+void GatherCBATestCases(std::vector<CBATestCase>& cba_test_cases)
+{
+    cba_test_cases.push_back(CBATestCase{
+        16, 128, 16, 16, 128, 3, 3, 0, 0, 1, 1, 1, 1, miopenActivationRELU, miopenConvolution});
+}
+
 // Extra layer of indirection introduced since GTEST_SKIP() cannot be called from non-void function.
 std::vector<CBATestCase> GetTestValues()
 {
```
BrianHarrisonAMD: I think we could combine these into a single method that returns the testcases, and I think the comment isn't applicable anymore with this fix.

```cpp
std::vector<CBATestCase> GetTestValues()
{
    return {{ 16, 128, 16, 16, 128, 3, 3, 0, 0, 1, 1, 1, 1, miopenActivationRELU, miopenConvolution }};
}
```

BrianHarrisonAMD:
Couple minor comments, but looks good!

Interesting find!
I had no idea it worked that way hehe.

BrianHarrisonAMD (approved):

test/gtest/api_convbiasactiv.cpp
CAHEK7: It's a part of overall gtest improvements https://github.com/ROCm/MIOpen/wiki/GTest-development#early-skip
Probably we should add some tickets here https://github.com/ROCm/MIOpen/issues/3140

xinlipn: > Can we move the skip to the start of SetUp(), and use the device flags for skipping instead?
> 
> ```c++
> void SetUp() override
> {
>     if(!IsTestSupportedForDevice())
>     {
>          GTEST_SKIP();
>     }
> 
>    // remaining Setup
> }
> 
> 
> 
> bool IsTestSupportedForDevice()
> {
> #if WORKAROUND_ISSUE_2212
>     using namespace miopen::debug;
>     using e_mask = enabled<Gpu::gfx900, Gpu::gfx906, Gpu::gfx908, Gpu::gfx90A, Gpu::gfx103X>;
>     using d_mask = disabled<Gpu::gfx110X, Gpu::gfx94X>;
>    return ::IsTestSupportedForDevMask<d_mask, e_mask>();
> #else
>     return true;
> #endif
> }
> ```

@BrianHarrisonAMD , thanks for the comments. Code updated

test/gtest/api_convbiasactiv.cpp
xinlipn: > I think we could combine these into a single method that returns the testcases, and I think the comment isn't applicable anymore with this fix.
> 
> ```c++
> std::vector<CBATestCase> GetTestValues()
> {
>     return {{ 16, 128, 16, 16, 128, 3, 3, 0, 0, 1, 1, 1, 1, miopenActivationRELU, miopenConvolution }};
> }
> ```

 @BrianHarrisonAMD , thanks for the comments. Cleaned up code

junliume (approved):

atamazov (approved):
🟢 LGTM

PR #3223, opened 2024-08-28
Gtest header patch for pre-c++17 compatibility

cderb:
patch gtest header for compatibility with pre-c++17

Some docker packages are failing to build make check with reference to std::filesystem
like compute-artifactory.amd.com:5000/rocm-plus-docker/compute-rocm-dkms-no-npi-hipclang:14644-sles-stg2

This removes reference to std::filesystem while preserving intent.

### DIFF
--- a/test/gtest/perf_helper.hpp
+++ b/test/gtest/perf_helper.hpp
@@ -110,7 +110,7 @@ struct PerfHelper
         }
 
         // If the file was just created (i.e., its size is 0), write the header.
-        if(fs::file_size(filename) == 0)
+        if(miopen::fs::file_size(filename) == 0)
         {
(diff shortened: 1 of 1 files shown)


test/gtest/perf_helper.hpp
```diff
@@ -110,7 +110,8 @@ struct PerfHelper
         }
 
         // If the file was just created (i.e., its size is 0), write the header.
-        if(std::filesystem::file_size(filename) == 0)
```
junliume, line 113: No objections but why not use `fs::file_size` since we have `#include <miopen/filesystem.hpp>`

cderb: Fair point. I'll update.

CAHEK7 (approved):

junliume (approved):

PR #3234, opened 2024-09-04
Enable NCHW/NHWC and NCDHW/NDHWC layout  in batch norm driver command

bghimireamd:
- Enable NCHW/NHWC layout from driver command for batch norm. 
- Moved `GpumemTensor` to `driver/driver.hpp`
- Stopped using old and slow `miopenBNFwdTrainPerActivationRunHost`, `miopenBNFwdTrainSpatialRunHost`, `miopenBNFwdInferPerActivationRunHost`, `miopenBNFwdInferSpatialRunHost`, `miopenBNBwdPerActivationRunHost` and `miopenBNBwdSpatialRunHost` since they only handled NCHW layout. Started using layout agnostic batch norm host function from `test/fusionHost.hpp`

eg: 

`./bin/MIOpenDriver bnorm -n 64 -c 1024 -H 14 -W 14 -m 1 --forw 1 -b 0 -s 0 -r 1`  _// defaults to NCHW_

`./bin/MIOpenDriver bnorm -n 64 -c 1024 -H 14 -W 14 -m 1 --forw 1 -b 0 -s 0 -r 1 -L NCHW`
`./bin/MIOpenDriver bnorm -n 64 -c 1024 -H 14 -W 14 -m 1 --forw 1 -b 0 -s 0 -r 1 -L NHWC`
`./bin/MIOpenDriver bnorm -n 64 -c 1024 -H 14 -W 14 -m 1 --forw 1 -b 0 -s 0 -r 1 --layout NHWC`

`bnorm, bnormfp16, bnormfp16fp32, bnormbfp16fp32,`

3D (we do not have 3D solvers we convert 3D to 2D tensor and feed it to 2D solvers. Once we have 3D batch norm solvers we can remove the conversion from 3D to 2D tensor)

3D example
`./bin/MIOpenDriver bnorm -n 8 -c 8 -H 12 -W 12 -D 12 -m 1 --forw 0 -b 1 -s 1 -r 1 --layout NDHWC`


alexandraBara (commented):
-L/--layout is quite generic. Which layout does it refer to? in/out/fil? 
In convolution we do this: --fil_layout NCHW --in_layout NCHW --out_layout NCHW - which is nice bc its specific. I think BN should follow the same pattern.
Maybe there will be a time where in/fil layout in BN will be different, so we need to keep that in mind

driver/bn_driver.hpp
```diff
@@ -60,19 +63,157 @@
...
+            ///
+            /// However we do not care about precision in gpualloc mode, because validation
+            /// is not used. Therefore, range (0,1] is fine.
+            return gpumemrand::gen_0_1(static_cast<Tgpu*>(GetDevicePtr()), sz);
+        }
+        return dev->ToGPU(q, GetVectorData());
+    }
+
+    template <typename T>
+    status_t AllocOnDevice(stream, context_t ctx, const size_t sz, std::vector<T>&)
```
CAHEK7: That's suspicious:
https://github.com/ROCm/MIOpen/blob/7e161f53d1e6e9b4ecf0fdbf59bb452b4f749c09/driver/conv_driver.hpp#L338
I'm not sure that we want another copy.

driver/bn_driver.hpp
```diff
@@ -60,19 +63,157 @@
...
+    }
+
+    auto GetDevicePtr() -> auto { return dev->GetMem(); }
+};
+//========================
+
 //#define BN_RUNFOR_PROFILER
 
-template <typename Tgpu, typename Tref, typename Tmix = Tgpu>
+template <typename Tgpu, typename Acc, typename Tref, typename Tmix = Tgpu>
```
CAHEK7: It doesn't look like layout related, could you clarify that change?

CAHEK7 (dismissed):
Some minor suggestions

driver/bn_driver.hpp
bghimireamd: reverted

driver/bn_driver.hpp
bghimireamd: should I bring class `GpumemTensor` from `conv_driver.hpp` into another file `driver_common.hpp` and let both `conv_driver.hpp` and `bn_driver.hpp` use it.

CAHEK7: Yes, and somewhen, in the future, when you will see any raw or even managed gpu pointers in the driver\tests, you can recommend to use that class.

bghimireamd: For now I moved to `driver.hpp`

driver/dm_bnorm.cpp
```diff
@@ -31,7 +31,11 @@ static Driver* makeDriver(const std::string& base_arg)
     if(base_arg == "bnorm")
         return new BatchNormDriver<float, double>();
     if(base_arg == "bnormfp16")
+        return new BatchNormDriver<float16, double, float16>();
+    if(base_arg == "bnormfp16fp32")
```
alexandraBara, line 35: "bnormfp16fp32" ? is this a typo?

alexandraBara (dismissed):

driver/driver.hpp
```diff
@@ -157,6 +160,140 @@ struct GPUMem
 #endif
 };
 
+template <typename Tgpu>
+class GpumemTensor
```
alexandraBara, line 164: Can this class be exposed in include/miopen? So we can include it in our Fin code? Otherwise, I think we are reinventing the wheel here: https://github.com/ROCm/MIFin/blob/develop/src/include/tensor.hpp

alexandraBara: @CAHEK7  thoughts?

driver/dm_bnorm.cpp
bghimireamd: Unfortunately, no. We have two different batch normalization kernels: OCL and CK. Both use different data types for their tensors. See the image below

image

For example, dScale is fp16 in the OCL kernel and fp32 in the CK kernel. Thus, bnormfp16fp32 is used for CK, while bnormfp16 is used for the OCL kernel's tensor initialization. This is an unsatisfactory solution for now, until we can develop a proper final API.

driver/driver.hpp
CAHEK7: It should be in the next PR. Lets separate unrelated changes.

src/driver_arguments.cpp
```diff
@@ -66,6 +66,10 @@ void BnDataType(std::stringstream& ss, const miopen::TensorDescriptor& desc)
     {
         ss << "bnormfp16";
     }
+    if(desc.GetType() == miopenBFloat16)
+    {
+        ss << "bnormbfp16";
+    }
```
BrianHarrisonAMD, line 76: Not sure we have enough details here to know, but do we need to handle "bnormfp16fp32" here as well?

And should this be "bnormbfp16fp32" instead?

bghimireamd: you are correct! There is not enough details to find out where the run was "bnormfp16" or "bnormfp16fp32" so I defaulted to "bnormfp16"

include/miopen/miopen.h
```diff
@@ -2738,6 +2738,28 @@ miopenBatchNormalizationForwardTraining(miopenHandle_t handle,
                                         void* resultSaveMean,
                                         void* resultSaveInvVariance);
 
+MIOPEN_EXPORT miopenStatus_t
+miopenBatchNormalizationForwardTraining_V2(miopenHandle_t handle,
```
bghimireamd, line 2781: since C does not support function overloading had to create `miopenBatchNormalizationForwardTraining_V2` for `miopenBatchNormalizationForwardTraining`

junliume:
@bghimireamd the FIN submodule needs to be updated from the CI results. Thanks!

alexandraBara (approved):
LGTM

driver/bn_driver.hpp
```diff
@@ -317,13 +344,92 @@ std::vector<int> BatchNormDriver<Tgpu, Tref, Tmix>::GetInputTensorLengthsFromCmd
...
+template <typename Tgpu, typename Tref, typename Tmix>
+int BatchNormDriver<Tgpu, Tref, Tmix>::ChkLayout_ShortName()
+{
+    // check for short name of layout type
+    if(inflags.FindShortName("layout") == 'L')
+    {
+        // do noting
+        // found valid short names
+        return 0;
+    }
```
BrianHarrisonAMD: Just a nitpick, but surprised this returns an int, and 0 is valid. 
I would expect it to return a bool, and true means valid.

driver/bn_driver.hpp
```diff
@@ -317,13 +344,92 @@ std::vector<int> BatchNormDriver<Tgpu, Tref, Tmix>::GetInputTensorLengthsFromCmd
...
+           (layout_value.compare("NCDHW") == 0) || (layout_value.compare("NDHWC") == 0))
+        {
+            // do nothing,Values are matching as defined in Lib.
+        }
+        else
+        {
+            std::cerr << "Invalid Layout Parameter Value - " << layout_value << std::endl;
+            exit(EXIT_FAILURE);
+        }
+    }
```
BrianHarrisonAMD: Could collapse this down into a single if / else if block to simplify:


```cpp
if((ChkLayout_ShortName()))
{
    std::cerr << " Invalid Layout Short Name = " << ChkLayout_ShortName() << std::endl;
    exit(EXIT_FAILURE);
}
else if((layout_value.compare("NCHW") != 0) && (layout_value.compare("NHWC") != 0) &&
           (layout_value.compare("NCDHW") != 0) && (layout_value.compare("NDHWC") != 0)
{
     std::cerr << "Invalid Layout Parameter Value - " << layout_value << std::endl;
     exit(EXIT_FAILURE);
}
```

src/batch_norm_api.cpp
```diff
@@ -88,27 +90,156 @@ miopenBatchNormalizationForwardInference(miopenHandle_t handle,
...
                         bn_mode,
                         xDesc,
                         x,
                         yDesc,
                         y,
-                        bnScaleBiasMeanVarDesc,
+                        scaleDesc,
+                        (BiasDesc == nullptr) ? scaleDesc : BiasDesc,
+                        (estMeanDesc == nullptr) ? scaleDesc : estMeanDesc,
+                        (estVarianceDesc == nullptr) ? scaleDesc : estVarianceDesc,
```
BrianHarrisonAMD: If we change the original API calls to instead call _V2 with these set, then we could make these descriptors required for the _V2 API, and not need these checks.

It's not a huge deal, but I think doing that would simplify the code a bit.

(If you decide to do this change, then the same also applies for training & backwards)

src/solver/batchnorm/backward_ck.cpp
```diff
@@ -201,7 +201,8 @@ bool BnCKBwdBackward::IsApplicable(
         return false;
     if(bn_problem.GetDirection() != miopen::batchnorm::Direction::Backward)
         return false;
-
+    if(!bn_problem.Is2D())
+        return false;
```
BrianHarrisonAMD, line 205: Should this have always been there?

I would have thought this was needed before these changes as well.

test/gtest/bn_bwd.cpp
```diff
@@ -26,46 +26,123 @@
...
+TEST_P(GPU_BN_V2_BwdNHWC_FP64, BnV2BwdCKDouble) {}
 
-TEST_P(GPU_BNBwd_BFP16, BnBwdCKBFloat16) {}
-TEST_P(GPU_BNBwd_FP64, BnBwdCKDouble) {}
+// // fp16
+// INSTANTIATE_TEST_SUITE_P(Smoke,
+//                          GPU_BN_V1_BwdNCHW_FP16,
+//                          testing::Combine(testing::ValuesIn(NetworkSmall<BNTestCase>()),
+//                                           testing::Values(miopenTensorNCHW),
+//                                           testing::ValuesIn({testBNAPIV1})));
```
BrianHarrisonAMD, line 103: Any reason why this one is commented out?

BrianHarrisonAMD (approved):
LGTM.

Left some suggestions, but don't think they are urgent.

driver/bn_driver.hpp
bghimireamd: Yes! I borrowed the code from conv_driver.hpp and did not bothered to change. I will make this `bool` since that makes more sense here.

src/batch_norm_api.cpp
bghimireamd: good call

src/solver/batchnorm/backward_ck.cpp
bghimireamd: yes! current CK batch norm only supports 2D.

test/gtest/bn_bwd.cpp
bghimireamd: I commented this since the ocl batch norm solver NCHW is not working. I think we need to fix that but that is separate task. I have commented for now so that we don't forget this issue.

cderb (approved):

junliume (approved):

PR #3235, opened 2024-09-04
clang-format-13

averinevg:
Update to clang-format-13


junliume:
> Update to clang-format-13

@averinevg why version 13? In another discussion thread I thought it was mentioned that we need to update to the default version in Ubuntu 22.04?
Update: e.g. why not clang-format-15? ref: https://github.com/ROCm/MIOpen/pull/3213#discussion_r1742364740

averinevg:
> > Update to clang-format-13
> 
> @averinevg why version 13? In another discussion thread I thought it was mentioned that we need to update to the default version in Ubuntu 22.04? Update: e.g. why not clang-format-15? ref: [#3213 (comment)](https://github.com/ROCm/MIOpen/pull/3213#discussion_r1742364740)

@junliume If we update to clang-format-15 right away, there will be too many changes, I tried. Let's move step by step.

Default version in Ubuntu 22.04 is clang-format-14, but it is possible to install clang-format-15 OOTB.

CAHEK7 (approved):

junliume:
@averinevg  @CAHEK7  Let's place this PR after next staging :) It involves lots of changed files, and CK needs to change too. and we want to check what the rest of ROCm is using.

PR #3239, opened 2024-09-06
Fix ConvBiasResAddActivation_fwd tests failing on Navi

BrianHarrisonAMD:
The graphapi tests for ConvBiasResAddActiv need to be skipped on Navi.
Since these use the same backend as the fused solution, I moved the check to a shared place, and updated to use the new IsTestSupportedForDev check.

### DIFF
--- a/test/gtest/fused_conv_bias_res_add_activ.cpp
+++ b/test/gtest/fused_conv_bias_res_add_activ.cpp
@@ -24,11 +24,9 @@
  *
  *******************************************************************************/
 #include <gtest/gtest.h>
+#include <gtest/gtest_ck_common.hpp>
 #include <miopen/miopen.h>
 #include <miopen/env.hpp>
-#if MIOPEN_USE_COMPOSABLEKERNEL
-#include <miopen/solver/ck_utility_common.hpp>
-#endif
 
 #include "tensor_util.hpp"
 #include "get_handle.hpp"
@@ -38,28 +36,16 @@
 MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_TEST_ALL)
 MIOPEN_DECLARE_ENV_VAR_STR(MIOPEN_TEST_FLOAT_ARG)
 
-#if MIOPEN_USE_COMPOSABLEKERNEL
-#define WORAROUND_ISSUE_2533 1
-#endif
-
 namespace conv_bias_act_res_add_fwd {
 
 bool TestIsApplicable()
 {
-#if MIOPEN_USE_COMPOSABLEKERNEL
     const auto float_arg = env::value(MIOPEN_TEST_FLOAT_ARG);
-    return
-#if WORAROUND_ISSUE_2533
-        miopen::solver::ck_utility::is_ck_whitelist(get_handle().GetDeviceName()) //
-#else
     /// \todo Check against specific ASCIs.
-#endif
-        && (float_arg == "--half"           // So far only test for fp16 is implemented.
-            || float_arg.empty())           // Empty when gtest is run without parameters.
-        && !env::disabled(MIOPEN_TEST_ALL); // Not disabled when gtest is run without parameters.
-#else
-    return false;
-#endif
+    return ::IsDeviceSupportedForCK() &&
+           (float_arg == "--half"              // So far only test for fp16 is implemented.
+            || float_arg.empty())              // Empty when gtest is run without parameters.
+           && !env::disabled(MIOPEN_TEST_ALL); // Not disabled when gtest is run without parameters.
 }
 
 std::vector<Conv3DTestCase> ConvTestConfigs()
--- a/test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
+++ b/test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
@@ -45,7 +45,10 @@
 namespace gr = miopen::graphapi;
 namespace conv_graph_api_test {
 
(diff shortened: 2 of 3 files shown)


junliume (approved):

test/gtest/gtest_ck_common.hpp
```diff
@@ -0,0 +1,48 @@
...
+// MI100 : gfx908
+// MI200 : gfx90a
+// MI300 : gfx940, gfx941, gfx942
+static inline bool IsDeviceSupportedForCK()
+{
+#if MIOPEN_USE_COMPOSABLEKERNEL
+#if WORAROUND_ISSUE_2533
+    using e_mask = enabled<Gpu::gfx908, Gpu::gfx90A, Gpu::gfx94X>;
+    using d_mask = disabled<Gpu::gfx103X, Gpu::gfx110X, Gpu::gfx120X>;
+    return ::IsTestSupportedForDevMask<d_mask, e_mask>();
```
averinevg:
```suggestion
    return IsTestSupportedByDevice(Gpu::gfx908 | Gpu::gfx90A | Gpu::gfx94X);
```

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
```diff
@@ -45,13 +46,7 @@
...
 namespace conv_graph_api_test {
 
-bool IsTestSupportedForDevice()
-{
-    using e_mask = enabled<Gpu::gfx94X, Gpu::gfx103X, Gpu::gfx110X>;
-    // gfx120X is not enabled due to WORKAROUND_SWDEV_479810
-    using d_mask = disabled<Gpu::None>;
-    return ::IsTestSupportedForDevMask<d_mask, e_mask>();
-}
+bool IsTestSupportedForDevice() { return ::IsDeviceSupportedForCK(); }
```
averinevg: How do we know that `graphapi` use only `CK`? There should be no dependency here.

test/gtest/fused_conv_bias_res_add_activ.cpp
```diff
@@ -38,28 +36,16 @@
...
-#else
     /// \todo Check against specific ASCIs.
-#endif
-        && (float_arg == "--half"           // So far only test for fp16 is implemented.
-            || float_arg.empty())           // Empty when gtest is run without parameters.
-        && !env::disabled(MIOPEN_TEST_ALL); // Not disabled when gtest is run without parameters.
-#else
-    return false;
-#endif
+    return ::IsDeviceSupportedForCK() &&
```
averinevg, line 45: We have fusions for Winograd solvers too.

averinevg (dismissed):

test/gtest/gtest_ck_common.hpp
CAHEK7: Let me clarify this situation a bit.
`e_mask\d_mask` were introduced specifically to make CTest porting easy - old CTest infrastructure has those CMake flags and I wanted to implement exactly the same functionality for code equality verification - you can compare 1:1 CTest and GTest device lists. Under the hood it has some defaults, disable mask excludes some gpus from default mask and enable mask includes new gpus. The key functionality is a default mask under the hood.

For the new tests I guess it's better to feed a list of supported devices, instead of enabled\disabled masks.

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
BrianHarrisonAMD: The graphAPI implementation in this case is using the fused CK solver.
I agree that it's not really clear, but not really sure how else we can avoid this issue. 
Once we add a generic graph solver, then this shouldn't be an issue, but since we have no generic fallback, I think this is how it needs to work for now.

test/gtest/gtest_ck_common.hpp
BrianHarrisonAMD: Ah I see. 
Sounds good!

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
BrianHarrisonAMD: I think we still need this check either way for the tests, but maybe it's better to have it fail the matches check for the pattern as well on unsupported devices?

```cpp
bool matches(const OpGraph* graph_ptr) const final
    {
        assert(graph_ptr);

        if(!isIsomorphic(*graph_ptr, getPatternGraph()))
        {
            return false;
        }

        return hasBiasNode(*graph_ptr);
    }
```

At least then it'll fail earlier, and in the future if we add additional patterns that match (like a generic one), then it will only match a valid one.

If we go that route, then I guess we would want to rename the executor to indicate it's CK specific, and then in the matches we can check if CK is supported since we only have a CK executor available. Later on if we add additional executors for this pattern, then we can remove these checks.

averinevg: My point is that there should be a check for devices that work with this test at the moment. I am not aware of current graphapi implementation, but this is a general test, it should not have any backend bindings.

BrianHarrisonAMD: Oh you mean you would prefer to have just the checks inside the test, and not have the shared file with fused?

I combined it since they both have the same requirements, but could split it back out if you think that's better.

averinevg: Yes, inside _this_ test. Also let's keep that file for CK.

BrianHarrisonAMD: Do these latest changes match what you were expecting?

PR #3240, opened 2024-09-06
Add Mha Bias to Find 2.0 for forward pass

Vsevolod1983:
Adding bias to find 2/0 in order to support new patterns for Mha Forward


include/miopen/miopen.h
```diff
@@ -5407,33 +5407,34 @@ typedef enum
     miopenTensorMhaAmaxDK             = 33,
     miopenTensorMhaAmaxDV             = 34,
     miopenTensorMhaAmaxDS             = 35,
+    miopenTensorMhaBias               = 36,
```
junliume, line 5410: Need to be extra careful when miopen.h is revised. 
add @BrianHarrisonAMD for review

Vsevolod1983: At some point it was requested to have all values which are not under MIOPEN_BETA_API  to come first, and then have all values which are under MIOPEN_BETA_API define. This approach requires renumbering though.

BrianHarrisonAMD: I don't think we can change the enum values, or else we will break ABI.
I think in this case we would need to retain the old numbering, and set `miopenTensorMhaBias = 61`

Vsevolod1983: https://github.com/ROCm/MIOpen/pull/2776#discussion_r1525479166

It was discussed here with @atamazov

Vsevolod1983: Also, let me request opinion from @DrizztDoUrden as an other Find 2.0 contributor.

BrianHarrisonAMD: Hmm interesting.
If we aren't guaranteeing ABI for the beta API, then I suppose this is fine.

atamazov: @BrianHarrisonAMD @junliume Exactly.

atamazov (approved):
🟢 LGTM!

BrianHarrisonAMD (dismissed):
LGTM

Edit:

After reviewing MIOPEN_BETA_API status, I think we will need to discuss this a bit more to see how we want to handle this and ABI going forward. Don't think this discussion needs to block this PR, as this issue isn't really related to this specific PR, but is just something we need to figure out generally.

include/miopen/miopen.h
BrianHarrisonAMD: I took a look, and it's a bit strange, since I think we always have the betaAPI enabled now. 

```
# Build library with Beta APIs
add_definitions("-DMIOPEN_BETA_API=1")
```

It also looks like we need the betaAPI enabled for compilation to work.
We no longer have guards around using several things that depend on MIOPEN_BETA_API. 
Probably should discuss this a bit more.

CAHEK7: > It also looks like we need the betaAPI enabled for compilation to work.
> We no longer have guards around using several things that depend on MIOPEN_BETA_API.

Typical. We have the same problem with `MIOPEN_USE_COMPOSABLEKERNEL` - we break it quite often.
It may sound scary, but we have to build at least 2 configurations - with all the flags disabled and all the flags enabled, ideally it has to be `2^Nflags` configurations.

BrianHarrisonAMD (approved):

junliume (approved):

PR #3249, opened 2024-09-11
[Bug] missing split_k from end of kernel_id in ck wrw

cderb:
This to correct the stoi error found in:
https://ontrack-internal.amd.com/browse/SWDEV-479172


junliume (approved):

cderb:
@junliume 
implicitgemm_ck_util.hpp:IsCKArgsSupported attempts to parse out split_k from the end of the kernel_id which is supplied to it by the solver.
PerformanceConfigHipImplicitGemmGroupWrwXdlops sets kernel_id in a number of locations but usually appends a '+split_k' onto the end of the string where split_k is an integer.
The bug seems to come from the line in this file which is setting kernel_id, where the last appended string appears to be a kernel name and not the split_k argument. Which is causing a failure in IsCKArgsSupported when the function attempts to parse this string.
Waiting for input from @msaudulhassan to confirm this is the correct fix.

PR #3253, opened 2024-09-13
[testing] Speed up test_activation, convert to GTest

averinevg:
Now it takes 100 minutes (on test PC with a 16-core CPU) to complete `test_activation --all`. Of this time, GPU calculations take only 2 minutes, including data copy operations. Thus, there is huge potential for optimization. The specified test has been completely rewritten and now runs in less than 9 minutes with the same dataset, an hour and a half less.


test/network_data.hpp
```diff
@@ -43,6 +43,9 @@ inline int pick_batch_size(int x, int y)
         return x / y;
 }
 
+// Reduce tests execution time
+#define MIOPEN_TESTS_GET_INPUTS_ENABLE_HUGE_TENSORS 1
+
```
CAHEK7, line 48: We shall think about making that parameter runtime, actually switch between only small tensors for `smoke` tests and only big tensors for `full` tests.

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+    }
+};
+
+class ActivationRELU
+{
+public:
+    template <class T, class Tparam>
+    static T Forward(Tparam alpha, Tparam beta, Tparam gamma, T x)
+    {
+        return (x > static_cast<Tparam>(0)) ? x : static_cast<T>(0);
```
CAHEK7, line 109:
```suggestion
        return (x > static_cast<T>(0)) ? x : static_cast<T>(0);
```
At least `x` is a `T`, not `Tparam` and casting `x` to `Tparam` does not make much sense for the comparison to 0.

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+    template <class T, class Tparam>
+    static T Backward(Tparam alpha, Tparam beta, Tparam gamma, T dy, T x, T y)
+    {
+        return (x > static_cast<Tparam>(0)) ? dy : static_cast<T>(0);
```
CAHEK7, line 115: The same as for the Forward case.

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+    template <class T, class Tparam>
+    static Tparam Forward(Tparam alpha, Tparam beta, Tparam gamma, T x)
+    {
+        return std::abs(x);
+    }
+
+    template <class T, class Tparam>
+    static Tparam Backward(Tparam alpha, Tparam beta, Tparam gamma, T dy, T x, T y)
+    {
+        return dy * static_cast<Tparam>((x > static_cast<Tparam>(0)) ? 1 : -1);
```
CAHEK7, line 149: The same as `ActivationRELU` case

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+    }
+};
+
+class ActivationCLIPPEDRELU
+{
+public:
+    template <class T, class Tparam>
+    static Tparam Forward(Tparam alpha, Tparam beta, Tparam gamma, T x)
+    {
+        return std::min(alpha, std::max(static_cast<Tparam>(0), static_cast<Tparam>(x)));
```
CAHEK7:
```suggestion
        return std::clamp(static_cast<Tparam>(x), static_cast<Tparam>(0), alpha);
```

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+    template <class T, class Tparam>
+    static T Backward(Tparam alpha, Tparam beta, Tparam gamma, T dy, T x, T y)
+    {
+        Tparam x_native = x;
+        return (x_native > 0 && x_native <= alpha) ? dy : static_cast<T>(0);
```
CAHEK7, line 184:
```suggestion
        return (x > 0 &&  static_cast<Tparam>(x) <= alpha) ? dy : static_cast<T>(0);
```

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+    static Tparam Forward(Tparam alpha, Tparam beta, Tparam gamma, T x)
+    {
+        Tparam x_native = x;
+        return (x_native > 0) ? x_native : x_native * alpha;
+    }
+
+    template <class T, class Tparam>
+    static Tparam Backward(Tparam alpha, Tparam beta, Tparam gamma, T dy, T x, T y)
+    {
+        return dy * ((x > static_cast<Tparam>(0)) ? 1 : alpha);
```
CAHEK7, line 201: Could we be more consistent?
Either  `(x_native > static_cast<Tparam>(0))` or `x > 0`?

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+        dx.data[i] = A::Backward(alpha, beta, gamma, dy.data[i], x.data[i], y.data[i]);
+}
+
+template <class A, class Tparam, class T>
+void CpuActivationForwardPackedMultiThread(
+    Tparam alpha, Tparam beta, Tparam gamma, const tensor<T>& x, tensor<T>& y)
+{
+    const auto num_items                    = x.data.size();
+    const std::size_t max_num_items_per_job = 16 * 1024 * 1024;
+    const std::size_t num_jobs = (num_items + max_num_items_per_job - 1) / max_num_items_per_job;
```
CAHEK7: https://github.com/ROCm/MIOpen/issues/2748

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+    const auto num_threads     = CpuActivationGetNumThreads(num_jobs);
+    if(num_threads == 1)
+    {
+        CpuActivationForwardPackedSingleThread<A, Tparam>(alpha, beta, gamma, x, y);
+        return;
+    }
+
+    const std::size_t max_num_jobs_per_thread  = (num_jobs + num_threads - 1) / num_threads;
```
CAHEK7: https://github.com/ROCm/MIOpen/issues/2748
Can you make at least a function for this `Ceil` integer division?

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+    const std::size_t max_num_items_per_thread = max_num_items_per_job * max_num_jobs_per_thread;
+    const std::size_t remainder                = num_items % max_num_items_per_thread;
+    const auto num_async_threads               = num_threads - 1;
+
+    auto func_async = [&, max_num_items_per_thread, alpha, beta, gamma](unsigned thread_num) {
+        const auto offset = max_num_items_per_thread * thread_num;
+        const auto end    = offset + max_num_items_per_thread;
+        for(std::size_t i = offset; i < end; i++)
+            y.data[i] = A::Forward(alpha, beta, gamma, x.data[i]);
+    };
```
CAHEK7: It's quite identical to `CpuActivationForwardPackedSingleThread`, I mean it can be generalized.

test/gtest/activ_common.hpp
```diff
@@ -0,0 +1,532 @@
...
+                                 tensor<T>& dx)
+{
+    CpuActivationBackwardPackedMultiThread<A, Tparam>(alpha, beta, gamma, y, dy, x, dx);
+}
+
+template <class A, class Tparam, class T>
+void CpuActivationForwardNonPacked(
+    Tparam alpha, Tparam beta, Tparam gamma, const tensor<T>& x, tensor<T>& y)
+{
+    throw std::runtime_error("CpuActivationForwardNonPacked is not implemented yet");
```
CAHEK7: It doesn't seem that the initial algorithm had this limitation.
I'm not sure we can just skip non-packed tests.

Especially if we remove previous test completely.

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+            DebugPrintTensors(DEBUG_PRINT_X_TENSOR | DEBUG_PRINT_Y_TENSOR_GPU |
+                                  DEBUG_PRINT_dY_TENSOR | DEBUG_PRINT_dX_TENSOR_CPU |
+                                  DEBUG_PRINT_dX_TENSOR_GPU,
+                              x_tensor,
+                              y_tensor_cpu,
+                              y_tensor_gpu,
+                              dy_tensor,
+                              dx_tensor_cpu,
+                              dx_tensor_gpu,
+                              100);
```
CAHEK7: Do we really want to spam this into CI logs?
Can it be at least under `MIOPEN_LOG_LEVEL`?

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+            NonPackedTensorWarning();
+        }
+
+        auto print_tensor = [=](auto tensor, const char* name) {
+            std::cout << name << "[";
+            const std::size_t len = std::min(max_len, tensor.size());
+            for(std::size_t i = 0; i < len; i++)
+                std::cout << tensor[i] << ",";
+            std::cout << "]" << std::endl;
+        };
```
CAHEK7: I suppose it should print formatted tensor instead of flattened.
But actually I don't think that it's necessary at all.
By the way, there is a https://github.com/ROCm/MIOpen/blob/61d9825a5eaec632a727e30d4ea881f28c38626f/test/tensor_util.hpp#L122
Can we use a single solution?

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+                              dx_tensor_gpu,
+                              100);
+            GTEST_FAIL();
+        }
+
+        if(!verify_bwd_ready.get())
+        {
+            DebugPrintTensors(DEBUG_PRINT_X_TENSOR | DEBUG_PRINT_Y_TENSOR_GPU |
+                                  DEBUG_PRINT_dY_TENSOR | DEBUG_PRINT_dX_TENSOR_CPU |
+                                  DEBUG_PRINT_dX_TENSOR_GPU,
```
CAHEK7: Do we really need that kind of flexibility? It's already bound to the algorithm, and it can be just a `bool forward` parameter.

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+        if(print_mask & DEBUG_PRINT_dX_TENSOR_GPU)
+            print_tensor(dx_gpu.data, "dX_TENSOR_GPU");
+    }
+
+    static void NonPackedTensorWarning()
+    {
+        std::cout << "WARNING: Non-packed tensor. Some modifications are needed to achieve optimal "
+                     "performance."
+                  << std::endl;
+    }
```
CAHEK7, line 184: Either we support non-packed tensors, or we don't. What can we do with such kind of warning in CI?
Btw, it will print identical warnings for each non-packed tensors during initialization and in case of failed verification and does not print any information of which tensor is not packed.

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+                       const tensor<T>& x,
+                       tensor<T>& y,
+                       tensor<T>& dx,
+                       const tensor<T>& dy,
+                       std::promise<void>& forward_ready)
+    {
+        RunCpuForward(desc, x, y);
+        forward_ready.set_value();
+        RunCpuBackward(desc, y, dy, x, dx);
+    }
```
CAHEK7:
```suggestion
    static void RunCpu(const miopen::ActivationDescriptor& desc,
                       const tensor<T>& x,
                       tensor<T>& y,
                       tensor<T>& dx,
                       const tensor<T>& dy,
                       std::promise<void>& forward_ready)
    {
        miopen::tests::CpuActivationForward(
            desc.GetMode(), desc.GetAlpha(), desc.GetBeta(), desc.GetGamma(), x, y);
        forward_ready.set_value();
        miopen::tests::CpuActivationBackward(
            desc.GetMode(), desc.GetAlpha(), desc.GetBeta(), desc.GetGamma(), y, dy, x, dx);
    }
```
All those functions are already `Forward` and `Backward`. Why do we need extra wrappers for them?

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+    {
+        auto handle = miopen::Handle{};
+
+        auto x_dev = handle.Write(x.data);
+        auto y_dev = handle.Write(y.data);
+
+        RunGpuForward(handle, desc, x.desc, x_dev.get(), y.desc, y_dev.get());
+
+        y.data = handle.Read<T>(y_dev, y.data.size());
+        handle.Finish();
```
CAHEK7: `handle.Read`/`handle.ReadTo` are synchronous. That explicit synchronization is not required, even for "just in case". Asynchronous version of this function will break whole library.

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+                       y_dev.get(),
+                       dy.desc,
+                       dy_dev.get(),
+                       x.desc,
+                       x_dev.get(),
+                       dx.desc,
+                       dx_dev.get());
+
+        dx.data = handle.Read<T>(dx_dev, dx.data.size());
+        handle.Finish();
```
CAHEK7: `handle.Read`/`handle.ReadTo` are synchronous. That explicit synchronization is not required, even for "just in case". Asynchronous version of this function will break whole library.

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+        float alpha = 1.0f;
+        float beta  = 0.0f;
+
+        miopenStatus_t status;
+        status =
+            desc.Backward(handle, &alpha, y_desc, y, dy_desc, dy, x_desc, x, &beta, dx_desc, dx);
+
+        if(status != miopenStatusSuccess)
+            throw std::runtime_error("Backward failed");
+    }
```
CAHEK7: Why do we need such wrappers?

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+        GenData(tensor, gen_x_value);
+    }
+
+    static void GenData_dy(tensor<T>& tensor)
+    {
+        auto gen_dy_value = [](auto& v) {
+            v = prng::gen_A_to_B(static_cast<T>(-0.5), static_cast<T>(0.5));
+        };
+        GenData(tensor, gen_dy_value);
+    }
```
CAHEK7, line 217: I'm not insisting, but I'd prefer keeping the data ranges and generators the same as in the previous test.

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+    template <class G>
+    static void GenData(tensor<T>& tensor, G g)
+    {
+        if(!tensor.desc.IsPacked())
+            NonPackedTensorWarning();
+
+        std::for_each(tensor.begin(), tensor.end(), g);
+    }
```
CAHEK7, line 201: That function is only needed for `NonPackedTensorWarning`, but that warning appears multiple times in multiple places and there isn't much sense to have it here.

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+        gpu_ready.wait();
+        cpu_ready.wait();
+
+        auto verify_bwd_ready = std::async(std::launch::async, [&]() {
+            return VerifyDataAsync(dx_tensor_gpu, dx_tensor_cpu, false);
+        });
+
+        verify_fwd_ready.wait();
+        verify_bwd_ready.wait();
```
CAHEK7:
```suggestion
```
`future.get()` waits for the result.

test/gtest/unit_ActivationDescriptor.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+        auto x_data_ready      = std::async(std::launch::async, [&]() { GenData_x(x_tensor); });
+        auto y_data_gpu_ready  = std::async(std::launch::async, [&]() { ZeroData(y_tensor_gpu); });
+        auto dx_data_gpu_ready = std::async(std::launch::async, [&]() { ZeroData(dx_tensor_gpu); });
+        auto dy_data_ready     = std::async(std::launch::async, [&]() { GenData_dy(dy_tensor); });
+
+        x_data_ready.wait();
+        y_data_gpu_ready.wait();
+        dx_data_gpu_ready.wait();
+        dy_data_ready.wait();
```
CAHEK7, line 110: If you pass it right into `RunCpu`/`RunGpu` you can postpone `dx`/`dy` wait till backward pass. (just an idea for already complex async code).

CAHEK7 (changes requested):
That's a great improvement, but it seems logging and code quality can be improved a bit.
1) make static cast in `namespace activ_func {` more uniform and logic, Right now it feels a bit chaotic.
2) it feels like th4ere is too many wrappers over a single function, could you clarify why do we need them, or remove. (also some code generalization may reduce copypaste and number of wrappers)
3) warnings (I doubt that they are necessary at all) and debug print pollutes the logs on CI
4) there are some unnecessary synchronizations

The last concert - it seems that it removes the tests for non-packed tensors and significantly changes data ranges and generation algorithms. That may affect the test coverage.

CAHEK7 (approved):

junliume (approved):

PR #3259, opened 2024-09-18
Updated reduceDims for NCHW and NHWC layouts for Backward CK

xinlipn:
Use reduceDims for switching NCHW and NHWC layouts with CK


bghimireamd:
I think we need to enable NCHW in [IsApplicable ](https://github.com/ROCm/MIOpen/blob/3141df3622070883827860ec7f42bade13454d0f/src/solver/batchnorm/backward_ck.cpp#L210C4-L211C22)of the solver too. Same goes for other solvers too.

xinlipn:
> I think we need to enable NCHW in [IsApplicable ](https://github.com/ROCm/MIOpen/blob/3141df3622070883827860ec7f42bade13454d0f/src/solver/batchnorm/backward_ck.cpp#L210C4-L211C22)of the solver too. Same goes for other solvers too.

@bghimireamd , done for this PR and the other two as well PR3227 and 3260

averinevg:

PR #3260, opened 2024-09-18
Updated reduceDims for NCHW and NHWC layouts forward training

xinlipn:
Update reduceDims for switching NCHW and NHWC layouts with CK


src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -76,6 +79,14 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
         SetSpatialDims();
         in_layout  = ComputeInLayout();
         out_layout = ComputeOutLayout();
+        if(IsLayoutNHWC())
+        { // NHWC
+            reduceDims = {0, 1, 2};
```
bghimireamd: can we do this directly in directly in [CKArgsBNormFwdTraining ](https://github.com/ROCm/MIOpen/blob/f65afdd36d8435c5b4c10e290f04a2f0282ca91d/src/solver/batchnorm/forward_training_ck.cpp#L90). Since `reduceDims` is more of CK thing and brining this to `problem_description` probably not good. Also, I think this reduceDims is for `miopenBNSpatial`. We also need to determine `reduceDims` based on bn mode (`miopenBNSpatial` or `miopenBNPerActivation`). Since we need to handle both case.

xinlipn: > can we do this directly in directly in [CKArgsBNormFwdTraining ](https://github.com/ROCm/MIOpen/blob/f65afdd36d8435c5b4c10e290f04a2f0282ca91d/src/solver/batchnorm/forward_training_ck.cpp#L90). Since `reduceDims` is more of CK thing and brining this to `problem_description` probably not good. Also, I think this reduceDims is for `miopenBNSpatial`. We also need to determine `reduceDims` based on bn mode (`miopenBNSpatial` or `miopenBNPerActivation`). Since we need to handle both case.

@bghimireamd , I agree and that putting reduceDims inside problem_description is not the best design. The original design was to put reduceDims as an input argument, which may affect APIs in e.g. src/include/miopen/solver/implicitgemm_ck_util.hpp. Will revert the reduceDims related changes in problem_description and it maybe more clear that way with the new changes

xinlipn: > can we do this directly in directly in [CKArgsBNormFwdTraining ](https://github.com/ROCm/MIOpen/blob/f65afdd36d8435c5b4c10e290f04a2f0282ca91d/src/solver/batchnorm/forward_training_ck.cpp#L90). Since `reduceDims` is more of CK thing and brining this to `problem_description` probably not good. Also, I think this reduceDims is for `miopenBNSpatial`. We also need to determine `reduceDims` based on bn mode (`miopenBNSpatial` or `miopenBNPerActivation`). Since we need to handle both case.

@bghimireamd, reverted the changes in problem_description. Feedback about reduceDims values with regard to bn mode is a bit mixed.

averinevg:

PR #3267, opened 2024-09-23
[Static CK] Remove unused files

averinevg:
PR removes unused `Static Composable Kernel` files


CAHEK7 (approved):

CAHEK7:
I'd like to ask how did you check that it's unusable?

averinevg:
> I'd like to ask how did you check that it's unusable?

@CAHEK7 I guess `Static Composable Kernel` was the first attempt to make what we now call the `CK`. These sources are used in a number of `ConvHipImplicitGemm*` solvers. None of them use the kernel that was removed. The rest of the files are just deprecated headers that are not included anywhere. These files were added 4 years ago and have remained unchanged since then, it is unlikely that anyone will use them.

junliume (approved):

PR #3268, opened 2024-09-23
[TEST] enable gtest filter

### LINKED ISSUE #3132: Add custom cmake test wrapper which can get gtest_parallel and run a single-binary test with custom regexp
In order to switch to a new test invocation scheme, described [on the wiki](https://github.com/ROCm/MIOpen/wiki/GTest-development), the current [cmake test wrapper](https://github.com/ROCm/MIOpen/blob/97ec5080860fd432b37e426bd3726cf6c8634b7a/test/gtest/CMakeLists.txt#L19) should be updated with the following functionality:
1) generate custom regexp based on the input test configuration:
    * MIOPEN_TEST_ALL maps to `Full` tests if enabled and to `Smoke` otherwise
    * MIOPEN_TEST_FLOAT_ARG maps to a proper test datatype suffix
    * DBSYNC maps into a CPU testsuite together with the other host-based tests, actually should be removed and a new pure CPU CI stage must be added
    * CODECOV_TEST seems to be abandoned, but actually must be removed and incorporated into `Full` testsuite
    * Explicit MLIR and COMPOSABLEKERNELS must be included into the naming scheme [on the wiki](https://github.com/ROCm/MIOpen/wiki/GTest-development) and properly handled.
    * MIOPEN_TEST_DEEPBENCH, MIOPEN_TEST_CONV, MIOPEN_TEST_WITH_MIOPENDRIVER and MIOPEN_DEBUG_TUNING_ITERATIONS_MAX should be removed, since nobody ever sets them
2) add an ability to run the tests based on custom regexp and without passing any environmental values to the tests (implies MIOPEN_TEST_DISCRETE is off and using a single test binary)
3) add an ability to download [gtest_parallel](https://github.com/google/gtest-parallel) and use it to run custom regexp tests in parallel

It blocks #3134

CAHEK7:
Partially fixes #3132
More filter options to come.


test/gtest/CMakeLists.txt
```diff
@@ -58,10 +58,29 @@ function(add_gtest TEST_NAME TEST_CPP)
     "ENVIRONMENT;MIOPEN_DEBUG_TUNING_ITERATIONS_MAX=${MIOPEN_DEBUG_TUNING_ITERATIONS_MAX};"
     "ENVIRONMENT;MIOPEN_TEST_WITH_MIOPENDRIVER=${MIOPEN_TEST_WITH_MIOPENDRIVER};")
 
+    set(MIOPEN_TEST_PREFIX)
+
+    if (MIOPEN_TEST_ALL)
+      set(MIOPEN_TEST_PREFIX Full)
+    else()
+      set(MIOPEN_TEST_PREFIX Smoke)
+    endif()
```
BrianHarrisonAMD: I think MIOPEN_TEST_ALL used to run all tests (including the smoke tests), but now it will only test things that aren't marked as smoke tests. 

At least when I look at the tests that were previously smoke tests, it looks like they just didn't check the MIOPEN_TEST_ALL env variable, and would be included in both runs.

CAHEK7: It depends on a set of nodes.
If we run "TEST_ALL" and "not TEST_ALL" stages on the same set of nodes, then it doesn't make any sense at all.
If "TEST_ALL" uses wider set of nodes (or "not TEST_ALL" stages choose smaller available subset of the same nodes), probably we can reconsider and include smoke tests back into "ALL" for now, but in the future we should consider proper job splitting.

The problem is - we are running the same set of tests on the same nodes multiple times.

test/CMakeLists.txt
```diff
@@ -211,18 +211,22 @@ if(MIOPEN_TEST_WITH_MIOPENDRIVER)
     add_dependencies(check MIOpenDriver)
 endif()
 
+
+set(MIOPEN_TEST_SUFFIX)
```
averinevg:
```suggestion
set(MIOPEN_GTEST_SUFFIX)
```

test/gtest/CMakeLists.txt
```diff
@@ -58,10 +58,29 @@ function(add_gtest TEST_NAME TEST_CPP)
     "ENVIRONMENT;MIOPEN_DEBUG_TUNING_ITERATIONS_MAX=${MIOPEN_DEBUG_TUNING_ITERATIONS_MAX};"
     "ENVIRONMENT;MIOPEN_TEST_WITH_MIOPENDRIVER=${MIOPEN_TEST_WITH_MIOPENDRIVER};")
 
+    set(MIOPEN_TEST_PREFIX)
```
averinevg:
```suggestion
    set(MIOPEN_GTEST_PREFIX)
```

test/gtest/CMakeLists.txt
```diff
@@ -58,10 +58,29 @@ function(add_gtest TEST_NAME TEST_CPP)
...
+
+    if (MIOPEN_TEST_ALL)
+      set(MIOPEN_TEST_PREFIX "*")
+    else()
+      set(MIOPEN_TEST_PREFIX "Smoke*")
```
averinevg: The asterisk should not be part of the prefix unless it is empty.

test/gtest/CMakeLists.txt
```diff
@@ -58,10 +58,29 @@ function(add_gtest TEST_NAME TEST_CPP)
...
+    endif()
+
+    set(MIOPEN_TEST_FILTER "${MIOPEN_TEST_PREFIX}${MIOPEN_TEST_SUFFIX}*")
```
averinevg:
```suggestion
    set(MIOPEN_GTEST_FILTER "${MIOPEN_TEST_PREFIX}${MIOPEN_TEST_SUFFIX}*")
```

averinevg (changes requested):

test/gtest/CMakeLists.txt
BrianHarrisonAMD: Yea I agree, and I think we should have a discussion about the smoke stage.
I think it would be fine to remove running smoke during the full stages, and make them exclusive sets, but just wanted to make sure it was an intentional change that was discussed.

averinevg: Previously we had a set of binaries that were run from cmake. They checked the "--all" option. If it was there, full tests were run, if not - smoke tests. This option is set if the MIOPEN_TEST_ALL is set to 1. So previously we would run either **full tests** or **smoke tests** (Many tests that were added when switching to Google Test are not very well made; they literally check nothing, including these variables) So it needs to be returned to how it was made originally.

test/CMakeLists.txt
CAHEK7: At some point there will be only gtests for the tests. It means that currently having the other tests is just a coincidence and everything related to gtests should be related to simple tests and I don't think that explicit `gtest` makes much sense here.
But actually I don't care.

test/gtest/CMakeLists.txt
CAHEK7: It doesn't matter how the tests treat `--all`/`MIOPEN_TEST_ALL`, the only thing matters is the set of nodes.
If the set of nodes is not the same, it just means that old tests were badly designed.
If the set of nodes is the same, it means the option can be mutually exclusive. That's it.

Mutually exclusive option requires some effort and guarantees from our current CI pipelines. That's out of the scope of this PR.

CAHEK7:
@averinevg @BrianHarrisonAMD can you take a look?
@junliume it looks like this PR is a bit critical, CI finished less than 4 hours.

averinevg (approved):
LGTM

BrianHarrisonAMD (approved):
LGTM.

The FP16 filter also selects BFP16, but I think we can handle that in a follow on PR when we add more sophisticated filters and move forward with this.

junliume (approved):

PR #3277, opened 2024-09-27
Add ck mha fp16 solver

BrianHarrisonAMD:
Purpose of review:
- Integrating CK FA V2 FWD inference solution into MIOpen.

Note:
- Adding tests as a follow up PR.
- This is being merged into an integration branch, and won't be merged into develop until tests are completed and merged as well.
- This solver's functionality is limited to FWD inference for now.


src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -0,0 +1,265 @@
...
+#if MIOPEN_USE_COMPOSABLEKERNEL
+    if(!problem.IsForward())
+    {
+        return false;
+    }
+
+    if(!StartsWith(context.GetStream().GetDeviceName(), "gfx94"))
+    {
+        return false;
+    }
```
BrianHarrisonAMD: Going to double check this.
I know we had this restriction for FP8, and we are only generating the MHA instance for gfx94X right now, but this might be only an FP8 limitation and we should change this / the generation to include all supported devices.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -0,0 +1,265 @@
...
+    fmha_args.hdim_q         = hdim_q;
+    fmha_args.hdim_v         = hdim_v;
+    fmha_args.nhead_q        = nhead_q;
+    fmha_args.nhead_k        = nhead_k;
+    fmha_args.stride_q       = (i_perm ? hdim_q : nhead * hdim_q);
+    fmha_args.stride_k       = (i_perm ? hdim_q : nhead_k * hdim_q);
+    fmha_args.stride_v       = (i_perm ? shape_seqlen_k : nhead_k * shape_seqlen_k);
+    fmha_args.batch_stride_q = (nhead * shape_seqlen_q * hdim_q);
+    fmha_args.batch_stride_k = (nhead_k * shape_seqlen_k * hdim_q);
+    fmha_args.batch_stride_v = (nhead_k * hdim_v * shape_seqlen_k);
```
BrianHarrisonAMD: I think these stride values need to come from the tensorDescriptors, and not calculated actually.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -0,0 +1,265 @@
...
+    fmha_args.seqlen_k_ptr = nullptr;
+
+    fmha_args.scale_s           = scale_s;
+    fmha_args.scale_p           = scale_p;
+    fmha_args.scale_o           = scale_o;
+    fmha_args.stride_bias       = shape_seqlen_k;
+    fmha_args.stride_o          = (o_perm ? hdim_v : nhead * hdim_v);
+    fmha_args.nhead_stride_bias = 0;
+    fmha_args.nhead_stride_lse  = shape_seqlen_q;
+    fmha_args.nhead_stride_o    = o_perm ? shape_seqlen_q * hdim_v : hdim_v;
```
BrianHarrisonAMD: Same here, need these to come from tensorDescriptors.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -0,0 +1,265 @@
...
+    fmha_args.window_size_left  = 0;
+    fmha_args.window_size_right = 0;
+    fmha_args.mask_type         = static_cast<ck_tile::index_t>(fmha_traits.mask_type);
+
+    fmha_args.s_randval = false;
+    // Since we aren't storing the random values these will be unused for now.
+    fmha_args.stride_randval       = max_seqlen_k;
+    fmha_args.nhead_stride_randval = (shape_seqlen_q * max_seqlen_k);
+    fmha_args.batch_stride_randval = (nhead * shape_seqlen_q * max_seqlen_k);
```
BrianHarrisonAMD: Same here, but these aren't used for fwd inference.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -0,0 +1,265 @@
...
+                          hipMemcpyKind::hipMemcpyDeviceToHost);
+                hipMemcpy(&seed,
+                          dataFwd.dropoutSeedData,
+                          sizeof(uint64_t),
+                          hipMemcpyKind::hipMemcpyDeviceToHost);
+                hipMemcpy(&offset,
+                          dataFwd.dropoutOffsetData,
+                          sizeof(uint64_t),
+                          hipMemcpyKind::hipMemcpyDeviceToHost);
+            }
```
BrianHarrisonAMD, line 254: We have this because there is an API difference between MIOpen and CK.
We are taking in 3 single value tensors for probability, seed, and offset, but CK uses CPU values for their API.

For this current version this is how it needs to work, but I think we need to discuss aligning these to remove the need for this readback.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_FA_CK_V2_FWD)
+
+namespace miopen {
+
+namespace solver {
+
+namespace mha {
+
+static std::string Convert(miopenDataType_t dataType)
```
junliume, line 47: @BrianHarrisonAMD some CI issues:
```
error: unused function 'Convert' [-Werror,-Wunused-function]
```

BrianHarrisonAMD: Ill add #ifdef guards around it for compiling without CK.

src/solution.cpp
```diff
@@ -394,47 +394,12 @@ void Solution::RunImpl(Handle& handle,
...
-        (*invoker)(handle, invoke_ctx);
-        return;
-    }
-
-    auto ctx = ExecutionContext{&handle};
-
-    const auto mha_solution = GetSolver() == mhaForward.SolverDbId()
-                                  ? mhaForward.GetSolution(ctx, problem_description)
-                                  : mhaBackward.GetSolution(ctx, problem_description);
+    const auto algo    = AlgorithmName{"MHA"};
```
Vsevolod1983: In Problem.cpp file we have this name as "Mha". Here it's "MHA" with all capitals. Should we stick with the same name,  "Mha" ?

Vsevolod1983: Also, are you sure about the correctness of such a code replacement in general?
In RunImpl you try to find a solution again instead of using the solution which we already found on FindSolution phase. It may happen that ExecutePrimitive will execute not the solution we want to execute, because it will execute the first isApplicable solution.

Vsevolod1983 (changes requested):

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
BrianHarrisonAMD: Checked, and it should be supported.

Adding lib generation for gfx90a as well:
https://github.com/ROCm/composable_kernel/pull/1540

src/solution.cpp
BrianHarrisonAMD: Oh good point, ill fix that.

Vsevolod1983: My another comment has gone, probably I removed it accidentally.
Please consider the correctness of such a replacement in general. On RunImpl step we already have a solver we want to work with. We should not try to find it again in ExecutePromitive call. It may lead to wrong solver execution

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
CAHEK7: That's kind of usual practice, rocblas does the same for scaling - it allows gpu pointer, but actually copies it back and pass as a kernel parameter.

CAHEK7 (approved):

CAHEK7:
It would be nice to have some tests as well.

src/solution.cpp
BrianHarrisonAMD: Ill take a look.

Thanks for the heads up!

BrianHarrisonAMD:
> It would be nice to have some tests as well.

Absolutely! 

Working on those now, and this will live in an integration branch until those are ready.

Vsevolod1983 (changes requested):
changes requested related to ExecutePrimitive function call

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
BrianHarrisonAMD: Updated, and swapped hash to match the updated CK version.
Need to use amd-develop hash once CK develop has been promoted.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
BrianHarrisonAMD: Thanks, should be fixed now!

src/solver.cpp
```diff
@@ -679,6 +679,8 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
     Register(registry, ++id, Primitive::ReLU, prelu::MultiWeightsBackward{}.SolverDbId());
     Register(registry, ++id, Primitive::ReLU, prelu::SingleWeightBackward{}.SolverDbId());
 
+    Register(registry, ++id, Primitive::Mha, mha::MhaCKFlashAttentionV2Forward{}.SolverDbId());
+
     // IMPORTANT: New solvers should be added to the end of the function!
```
averinevg: Several IDs were added here recently, but the git doesn't show the conflict. Probably the empty line before the comment should be removed. It might also be necessary to add a note that we shouldn't leave an empty line before this comment, as otherwise Git won't detect the conflict.

BrianHarrisonAMD: Ill update and take a look.
Thanks for the heads up!

Pretty surprising that it would not detect a conflict since the same lines should have changed regardless of the newline.
I can update the comment if that ends up being the case though.

BrianHarrisonAMD: Yea weird, it didn't flag it as a conflict, but did (luckily) resolve it properly in this case. 
Ill update the comment as well.

requirements.txt
```diff
@@ -7,5 +7,5 @@ nlohmann/json@v3.11.2 -DJSON_MultipleHeaders=ON -DJSON_BuildTests=Off
 ROCm/FunctionalPlus@v0.2.18-p0
 ROCm/eigen@3.4.0
 ROCm/frugally-deep@9683d557eb672ee2304f80f6682c51242d748a50
-ROCm/composable_kernel@9c0811f39a2262dbe4d71b81898187951c1e11ba -DCMAKE_BUILD_TYPE=Release -DINSTANCES_ONLY=ON
+ROCm/composable_kernel@294cb823142a815170cf1faa63e01a431a557a04 -DCMAKE_BUILD_TYPE=Release -DINSTANCES_ONLY=ON
```
BrianHarrisonAMD, line 10: Note, this is pointing to a dev commit where I added generating the device_mha_operations lib for gfx90a as well.
This will need to be updated to point to an amd-develop commit instead once it's been promoted.

Since this is living in a integration branch it should be fine for now, but we shouldn't merge the integration into develop until this is resolved.

src/solution.cpp
BrianHarrisonAMD: Yea you are correct, this needs to use the solverId, and can't make a new container.

Fixing it!

BrianHarrisonAMD: Alright should be updated to be correct now @Vsevolod1983. 
If you get a chance could you take another look at the updated version?

Thanks again for the heads up!

Vsevolod1983 (approved):

PR #3282, opened 2024-09-30
[TEST] forbid GPU tests without explicit datatypes

CAHEK7:
There is only one test which mixed different datatypes and thus had `_NONE` datatype suffix.

All `_NONE` tests must be CPU only unless we have significant reason to add GPU type agnostic tests. 

It's a part of CPU-only stage for CI.

### DIFF
--- a/test/gtest/kernel_tuning_net.cpp
+++ b/test/gtest/kernel_tuning_net.cpp
@@ -10,15 +10,18 @@ struct KernelTuningNetTestCase : AIModelTestCase
     std::string arch;
 };
 
-std::vector<KernelTuningNetTestCase> GetConvAsm1x1UTestCases()
+std::vector<KernelTuningNetTestCase> GetConvAsm1x1UTestCases_FP32()
 {
     return {{{{1, 512, 192, 288, {56, 56}, {1, 1}, {0, 0}, {1, 1}, {1, 1}},
               miopen::conv::Direction::BackwardData,
               miopenFloat,
               miopenTensorNCHW},
              "1,16,1,64,2,2,1,4",
-             "gfx908"},
-            {{{1, 256, 2048, 512, {7, 7}, {1, 1}, {0, 0}, {1, 1}, {1, 1}},
+             "gfx908"}};
+}
+std::vector<KernelTuningNetTestCase> GetConvAsm1x1UTestCases_FP16()
+{
+    return {{{{1, 256, 2048, 512, {7, 7}, {1, 1}, {0, 0}, {1, 1}, {1, 1}},
               miopen::conv::Direction::Forward,
               miopenHalf,
               miopenTensorNCHW},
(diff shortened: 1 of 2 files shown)


test/gtest/kernel_tuning_net.cpp
```diff
@@ -10,71 +10,83 @@ struct KernelTuningNetTestCase : AIModelTestCase
...
+std::vector<KernelTuningNetTestCase> GetConvHipIgemmGroupFwdXdlopsTestCases_FP32()
+{
+    return {{{{16, 256, 2016, 192, {7, 7}, {1, 1}, {0, 0}, {1, 1}, {1, 1}},
+              miopen::conv::Direction::Forward,
+              miopenHalf,
+              miopenTensorNHWC},
+             "DeviceGroupedConvFwdMultipleABD_Xdl_CShuffle<64, 64, 64, 32, Filter1x1Stride1Pad0, "
+             "32, 32, 2, 2, 1, 1, 1, 1, 1>",
+             "gfx942"}};
 }
```
BrianHarrisonAMD: Are the data types here flipped?

Looks like FP16 has miopenFloat, and FP32 has miopenHalf.

test/gtest/kernel_tuning_net.cpp
```diff
@@ -86,131 +98,157 @@ std::vector<KernelTuningNetTestCase> GetConvHipIgemmGroupWrwXdlopsTestCases()
...
+                                                              conv_desc,
+                                                              test_case.direction)
+                           : miopen::conv::ProblemDescription(output_desc,
+                                                              weights_tensor_desc,
+                                                              input_tensor_desc,
+                                                              conv_desc,
+                                                              test_case.direction);
+
+        Solver perf_config;
+        ASSERT_TRUE(perf_config.IsModelApplicable(ctx, problem));
```
BrianHarrisonAMD, line 145: Looks like we used to skip if the model wasn't applicable, and now we just assert for it.
Is the intent that we only create these when the models are applicable, and it should fail if it doesn't match?

test/gtest/kernel_tuning_net.cpp
CAHEK7: :facepalm:

test/gtest/kernel_tuning_net.cpp
CAHEK7: It does not make sense to skip it. I expected this question)
Let's imagine that it's not applicable, skipped and... we don't have any tests to check AI tuning anymore. Silently.

junliume (approved):

test/gtest/kernel_tuning_net.cpp
BrianHarrisonAMD: Makes sense to me, but wanted to confirm!

BrianHarrisonAMD (commented):
Lgtm

PR #3283, opened 2024-09-30
Fix driver being leaked in MIOpenDriver

BrianHarrisonAMD:
This change will clean-up the driver implementation before exiting MIOpenDriver.

### DIFF
--- a/driver/main.cpp
+++ b/driver/main.cpp
@@ -52,10 +52,10 @@ int main(int argc, char* argv[])
         std::cout << " " << argv[i];
     std::cout << std::endl;
 
-    Driver* drv = nullptr;
+    std::shared_ptr<Driver> drv;
     for(auto f : rdm::GetRegistry())
     {
-        drv = f(base_arg);
+        drv = std::shared_ptr<Driver>(f(base_arg));
         if(drv != nullptr)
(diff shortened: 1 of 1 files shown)


CAHEK7 (approved):

bghimireamd (approved):

driver/main.cpp
```diff
@@ -52,10 +52,10 @@ int main(int argc, char* argv[])
         std::cout << " " << argv[i];
     std::cout << std::endl;
 
-    Driver* drv = nullptr;
+    std::shared_ptr<Driver> drv;
     for(auto f : rdm::GetRegistry())
     {
-        drv = f(base_arg);
+        drv = std::shared_ptr<Driver>(f(base_arg));
```
CAHEK7:
```suggestion
        drv.reset(f(base_arg));
```
Or at least `make_shared`. But I think `.reset` is the most suitable.

BrianHarrisonAMD: You can't actually use make_shared in this case since Driver is pure virtual, but reset makes sense to me.
Ill swap it.

JehandadKhan (approved):

junliume (approved):

PR #3285, opened 2024-10-01
Fixed incorrect transpose in find 2.0

DrizztDoUrden:
Moved the MakeTransposed() call out of FindSolutionsImpl() to remove some duplicated logic and effectively apply previous transpose fix to the rest of the function.


cderb (approved):
LGTM

junliume (approved):

PR #3294, opened 2024-10-03
move tensor_holder.hpp and gpuMemTensor.hpp from test to src

bghimireamd:
This PR expose `tensor_holder.hpp` and `gpuMemTensor.hpp`  headers outside of the current module. This change allows third-party libraries to directly include and utilize the MIOpen's tensor holder functionality that help easy data initialization and transfer between cpu and gpu.


averinevg (changes requested):
> This PR expose `tensor_holder.hpp` and `gpuMemTensor.hpp` headers outside of the current module.

We shouldn't do that. This code is not used within the library itself and will only pollute it. We should not mix test and library code.

> This change allows third-party libraries to directly include and utilize the MIOpen's tensor holder functionality that help easy data initialization and transfer between cpu and gpu.

These are our internal helpers. Third-party libraries should have their own stuff.

CAHEK7:
We need to discuss it one more time.

bghimireamd:
PR no longer needed

PR #3298, opened 2024-10-04
Update kdb install instructions

cderb:
update kdb apt install instructions and script

### DIFF
--- a/docs/install/install.rst
+++ b/docs/install/install.rst
@@ -65,10 +65,9 @@ To install the kernels package for your GPU architecture, use the following comm
 
 .. code:: shell
 
-  apt-get install miopenkernels-<arch>-<num cu>
+  apt-get install miopen-hip-<arch>kdb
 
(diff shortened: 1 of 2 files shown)


docs/install/install.rst
```diff
@@ -65,10 +65,9 @@ To install the kernels package for your GPU architecture, use the following comm
 
 .. code:: shell
 
-  apt-get install miopenkernels-<arch>-<num cu>
+  apt-get install miopen-hip-<arch>kdb
```
amd-jnovotny, line 68: Should we add "sudo" to the command? Would apt be preferrable to apt-get?

amd-jnovotny (approved):
Docs changes look good.

docs/install/install.rst
cderb: Currently there is some mixing in the docs for using apt vs sudo apt. I don't really mind either, but some uniformity would be nice. sudo is elevating the user so IMO not the main point we're trying to get across, ex in docker you wouldn't need it. I had just avoided addressing the issue here.

amd-jnovotny: Sounds good. We can look at standardization at some later time.

junliume (approved):

PR #3302, opened 2024-10-07
Update CK commit hash for staging

junliume:

PR #3304, opened 2024-10-08
Add tests for ck mha fp16 solver

BrianHarrisonAMD:
Purpose of review:

- Add tests for CK FA V2 FWD inference solver.
- Follow up to #3277

Note:
- Merging these changes into the integration branch to allow for limited diff, but will follow up with a PR for merging the integration branch into develop once complete. 
- Testing and solver are limited to FWD inference for now.

### DIFF
--- a/test/gtest/graphapi_capi_mha_forward.cpp
+++ b/test/gtest/graphapi_capi_mha_forward.cpp
@@ -296,7 +296,49 @@ class GPU_MhaForward_FP8 : public MhaForwardTest<float8>
     }
 };
 
+class GPU_MhaForward_FP16 : public MhaForwardTest<half_float::half>
+{
+    void RunCPUverify(miopen::Handle& handle) override
+    {
+        auto softmaxRef  = tensor<float>{m_testN, m_testH, m_testS, m_testS};
+        auto oDescRef    = tensor<half_float::half>{m_testN, m_testH, m_testS, m_testD};
+        auto mDescRef    = tensor<float>{m_testN, m_testH, m_testS, 1};
+        auto zInvDescRef = tensor<float>{m_testN, m_testH, m_testS, 1};
+
+        auto lookup = [this](const int64_t id) -> TensorVariant& {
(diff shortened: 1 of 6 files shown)


BrianHarrisonAMD:
Note #3302 updates requirement.txt CK hash to the version needed for the integration branch. 
Ill wait until #3302 is merged to develop before merging the integration branch to develop in order to align these properly (it'll probably be merged first anyways, but mentioning just in case).

test/gtest/graphapi_capi_mha_forward.cpp
```diff
@@ -296,7 +296,59 @@ class GPU_MhaForward_FP8 : public MhaForwardTest<float8>
...
+
+        auto oRes     = GetResult<half_float::half>(miopenTensorMhaO, handle);
+        double oError = miopen::rms_range(oDescRef, oRes);
+
+        if(m_bernulliProbability > 0.0f)
+        {
+            // Due to GPU version using a different dropout generator we will compare to CPU without
+            // dropout and verify that dropout causes a large difference when comparing results.
+            EXPECT_GT(oError, errorThreshold);
+        }
```
CAHEK7, line 342: That's only check which does make some sense with the current approach, but the ideal case is specifically prepared data which will allow us to check statistical properties of the dropout - tensor V must have identity matrixes, so it won't affect the data after dropout and we can say that approximately half of the elements are zero for `m_bernulliProbability = 0.5`.

It's just a side note of how it can be actually checked.

CAHEK7 (approved):

test/gtest/graphapi_capi_mha_forward.cpp
BrianHarrisonAMD: Oh interesting idea.
CK also allows us to capture the dropout data.
I was thinking it would make sense to capture, and use that for verification, but I figured we could add that as part of enabling the additional optional params.

junliume (approved):

PR #3308, opened 2024-10-10
Merge CK fwd mha FP16 solver

BrianHarrisonAMD:
This PR will merge the changes from #3304 & #3277 into develop.


requirements.txt
```diff
@@ -7,5 +7,5 @@ nlohmann/json@v3.11.2 -DJSON_MultipleHeaders=ON -DJSON_BuildTests=Off
 ROCm/FunctionalPlus@v0.2.18-p0
 ROCm/eigen@3.4.0
 ROCm/frugally-deep@9683d557eb672ee2304f80f6682c51242d748a50
-ROCm/composable_kernel@9c0811f39a2262dbe4d71b81898187951c1e11ba -DCMAKE_BUILD_TYPE=Release -DINSTANCES_ONLY=ON
+ROCm/composable_kernel@294cb823142a815170cf1faa63e01a431a557a04 -DCMAKE_BUILD_TYPE=Release -DINSTANCES_ONLY=ON
```
BrianHarrisonAMD: Waiting on #3302 to update the hash

BrianHarrisonAMD:
Looking into the CI issues for external CI / Windows.

BrianHarrisonAMD:
Windows issue was due to the wrong define guard being used.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -0,0 +1,268 @@
...
+            }
+
+            fmha_runtime_traits.has_dropout = false;
+            float probability               = 0;
+
+            // TODO : Change API to take in probability value as host side value instead of device
+            // pointer to match CK API. Calling a blocking hipMemcpy will cause issues with stream,
+            // and isn't async.
+
+            fmha_runtime_args.p_drop = probability;
```
BrianHarrisonAMD, line 241: Disabled probability for now with this PR. 
We can't do a blocking readback as it will use the wrong stream, and the async version won't allow us to set the value properly for the kernel invocation.

I think the best way to handle this is to take in a host value for probability in the future which will allow us to match CK  without a need to readback.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -0,0 +1,268 @@
...
+        return false;
+    }
+
+    auto deviceName = context.GetStream().GetDeviceName();
+    if(!StartsWith(deviceName, "gfx94") && deviceName != "gfx90a")
+    {
+        return false;
+    }
+
+    auto& descsFwd            = problem.GetDescsForward();
```
bghimireamd: nitpick `const`

test/gtest/mha_helper.hpp
```diff
@@ -423,6 +423,33 @@ void MultiHeadAttentionfp8(const tensor<T>& q_val,
     ScaleMult(atten_heads_fp32, o_scale, multi_head_attention_fp8);
 }
 
+template <typename T>
+void MultiHeadAttentionfp16(const tensor<T>& q_val,
```
bghimireamd: could we rename this to  MultiHeadFwdAttentionfp16. May be we need to do same for fp8 and fp32 version too.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
BrianHarrisonAMD: I like adding it as well to be explicit (and will), but in this case they are equivalent.

bghimireamd:
LGTM

bghimireamd (approved):

junliume (approved):

BrianHarrisonAMD:
Restarting CI after code review comments.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
CAHEK7: 1) `hipMemcpy` should not be used directly, actually non-async version must net be used ever.
2) `himMemcpyAsync` with proper current stream from the handle can be used, but should not be used directly.
3) `Handle` must provide all the functions to readback the value in a stream-safe manner, but right now it does not, since it uses `hipMemcpy` under the hood.

A general rule is to hide all direct hip API calls in the `Handle`, but we violate this rule for some new specific functionality like events, of some other hip API functions. We can't wrap them all, but we've lost flexibility and portability and actually that rule can be revisit since OpenCL backend has gone and now everything works through hip runtime.

Partially I agree that an external API is broken and we must pass those values through host pointers/values, but the latest changes in the graph API actually support both ways and all the graph-api tensors have flag is the tensor devise or host, so it means eventually we have to support readback or explicitly check that flag.

CAHEK7 (approved):

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
BrianHarrisonAMD: Oh good point about hiding the call behind handle as well.
That's likely something we will need to decide on in the future.

I think if it were host -> device, then it's fine, and `hipMemcpyAsync` would work, but device -> host requires synchronization, and in this case it's device -> host -> device, which ideally should be avoided. If it was just device -> host, then we could use callbacks, or events to synchronize properly without issue.

junliume:
Thanks @CAHEK7 for the comments. I am trying to move this info to a post in discussion and mark it as part of "MIOpen Developers' Guide" :)
```
hipMemcpy should not be used directly, actually non-async version must not be used ever.
himMemcpyAsync with proper current stream from the handle can be used, but should not be used directly.
```

Issue #3135, opened 2024-07-23
Add base class for gtests


CAHEK7:
Gtests are using random generator and setting internal library state based on environmental variables. It may work for a case `single binary-single test`, but will not work for a `single binary-all the tests` case since the tests will interfere each other and the behavior will be heavily dependent on the test order.

We need a base class for gtests, which will cover resetting prng at the beginning of the tests (trivial, by adding `prng::reset_seed();` in the constructor) and handling consistent environmental variable state but saving and restoring full state of the updated env variables (including `isSet`/`isUset` state).

reidkwja:
currently addressed in #3321

PR #3321, opened 2024-10-17
Adding base class for gtests

reidkwja:
Addressing issue #3135 with a random seed generator + environmental variables saving.

### DIFF
--- a/test/gtest/test_base.hpp
+++ b/test/gtest/test_base.hpp
@@ -1,0 +1,89 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


test/gtest/test_base.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+        save_env_vars();
+    }
+
+    void TearDown() override { restore_env_vars(); }
+
+private:
+    void save_env_vars()
+    {
+        const char* env_vars[] = {"CMAKE_CURRENT_BINARY_DIR",
```
bghimireamd: could we just use `std::vector<std::string>` ?

Vsevolod1983:
Btw, we are going to get rid of using environment variables https://github.com/ROCm/MIOpen/pull/3329

CAHEK7 (changes requested):
Could you provide an example of how it can be used?
For example for `test/gtest/smoke_solver_ConvBinWinograd3x3U.cpp` or `test/gtest/hipblaslt_gemm.cpp`.

test/gtest/smoke_solver_ConvBinWinograd3x3U.cpp
```diff
@@ -64,12 +64,31 @@ bool IsTestSupportedForDevice()
...
+    }
 };
 
+// Parameterized test implementation
 TEST_P(GPU_Conv2dDefault_FP32, FloatTest_smoke_solver_ConvBinWinograd3x3U)
 {
+    // Set environment variables dynamically
+    for(const auto& [var, val] : GPU_Conv2dDefault_FP32::get_env_values())
+    {
+        miopen::env::setEnvironmentVariable(var, val);
```
CAHEK7: I'm afraid it doesn't work for cached env variables.

CAHEK7 (changes requested):
It seems it does not build. Probably something went wrong with proper gtest base class.
Also it would be nice to have less destructive changes, ideally with minimal changes in the tests.

test/gtest/smoke_solver_ConvBinWinograd3x3U.cpp
reidkwja: Yes, I have been trying different ways to integrate it without changing gtest_common.hpp - which will require changing large numbers of tests. 

I am still looking into ways to do so with seamless integration into gtests...

BrianHarrisonAMD:
There is a couple different build errors showing up in the CI that will need to be resolved.

PR #3326, opened 2024-10-21
Enable tuning in Batch norm CK solver

bghimireamd:
This PR only enabling tuning in CK solver.


src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -267,6 +302,43 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
...
+
+    std::string GetDirectionStr() const
+    {
+        std::string s;
+
+        switch(direction)
+        {
+        case Direction::ForwardInference: s = "F2b0"; break;
+        case Direction::ForwardTraining: s = "F1b0"; break;
+        case Direction::Backward: s = "F0b1"; break;
```
DrizztDoUrden: I would suggest a naming scheme like "Inf", "Trn", "Bwd" so that people looking in the db wouldn't have to look it up in the code

src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -267,6 +302,43 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
...
+        default: assert(false);
```
DrizztDoUrden: I'd rather `MIOPEN_THROW` here, because it would be an error that should fail the CI no matter what

src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -267,6 +302,43 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
...
+
+    std::string GetModeStr() const
+    {
+        std::string s;
+
+        switch(bn_mode)
+        {
+        case miopenBNPerActivation: s = "0"; break;
+        case miopenBNSpatial: s = "1"; break;
+        default: assert(false);
```
DrizztDoUrden: Ditto

src/include/miopen/batchnorm/problem_description.hpp
DrizztDoUrden: Also you don't have to store them in temporary, can just do it like this
```suggestion
        case Direction::ForwardInference: return "F2b0";
        case Direction::ForwardTraining: return "F1b0";
        case Direction::Backward: return "F0b1";
```

src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -267,6 +302,43 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
...
+    }
+
+    std::string GetModeStr() const
+    {
+        std::string s;
+
+        switch(bn_mode)
+        {
+        case miopenBNPerActivation: s = "0"; break;
+        case miopenBNSpatial: s = "1"; break;
```
DrizztDoUrden: Same suggestion like above:
```suggestion
        case miopenBNPerActivation: return "0";
        case miopenBNSpatial: return "1";
```

src/ocl/batchnormocl.cpp
```diff
@@ -250,8 +250,7 @@ void BatchNormForwardInference(Handle& handle,
         }();
 
         const auto algo    = AlgorithmName{"miopenBatchNormalizationForwardInference"};
-        const auto solvers = solver::SolverContainer<solver::batchnorm::BnFwdInference,
-                                                     solver::batchnorm::BnCKFwdInference>{};
+        const auto solvers = solver::SolverContainer<solver::batchnorm::BnCKFwdInference>{};
```
DrizztDoUrden: Just a question: are you sure you want to remove non-ck solver?

src/solver/batchnorm/backward_ck.cpp
```diff
@@ -184,66 +166,404 @@ static ConvSolution MakeAnyInvokerFactory(const miopen::batchnorm::ProblemDescri
...
+{
+#if !MIOPEN_BACKEND_HIP || !MIOPEN_USE_COMPOSABLEKERNEL
+    std::ignore = problem_desc;
+    return false;
+#else
+    if(this->valid_kernels.empty())
+    {
+        this->HeuristicInit(problem_desc);
+        assert(!valid_kernels.empty());
+        return true;
```
DrizztDoUrden: I would suggest either throwing or returning false when condition in the assert is false

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -112,6 +113,171 @@ struct CKArgsBNormFwd
...
+                                                    {nullptr, nullptr, nullptr, nullptr, nullptr},
+                                                    {nullptr},
+                                                    Normalize{0.0});
+        if(it->IsSupportedArgument(argument_ptr.get()))
+        {
+            valid_kernels.push_back(it->GetTypeString());
+        }
+    }
+
+    assert(!valid_kernels.empty());
```
DrizztDoUrden: this should throw instead, otherwise we segfault two lines bellow

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -112,6 +113,171 @@ struct CKArgsBNormFwd
...
+          typename BiasDataType,
+          typename MeanVarDataType>
+bool PerformanceConfigBnCKFwdInference::CheckIsSupportCKArgs(
+    const miopen::batchnorm::ProblemDescription& problem) const
+{
+    const auto& args = CKArgsBNormFwd{problem};
+    const auto bn_fwd_ptrs =
+        DeviceOpBnFwdInfPtrs<XDataType, YDataType, ScaleDataType, BiasDataType, MeanVarDataType>::
+            GetInstances();
+    assert(!bn_fwd_ptrs.empty());
```
DrizztDoUrden: should be a throw

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -112,6 +113,171 @@ struct CKArgsBNormFwd
...
+    const miopen::batchnorm::ProblemDescription& problem_desc)
+{
+#if !MIOPEN_BACKEND_HIP || !MIOPEN_USE_COMPOSABLEKERNEL
+    std::ignore = problem_desc;
+    return false;
+#else
+    if(this->valid_kernels.empty())
+    {
+        this->HeuristicInit(problem_desc);
+        assert(!valid_kernels.empty());
```
DrizztDoUrden: should either throw or return false on failure

src/solver/batchnorm/forward_training_ck.cpp
```diff
@@ -174,68 +158,397 @@ static ConvSolution MakeAnyInvokerFactory(const miopen::batchnorm::ProblemDescri
...
+    const miopen::batchnorm::ProblemDescription& problem_desc)
+{
+    const auto& args       = CKArgsBNormFwdTraining{problem_desc};
+    const auto bn_fwd_ptrs = DeviceOpBNFwdTrainingPtrs<XDataType,
+                                                       YDataType,
+                                                       AccDataType,
+                                                       ScaleDataType,
+                                                       BiasDataType,
+                                                       MeanVarDataType>::GetInstances();
+    assert(!bn_fwd_ptrs.empty());
```
DrizztDoUrden: should throw

src/solver/batchnorm/forward_training_ck.cpp
```diff
@@ -174,68 +158,397 @@ static ConvSolution MakeAnyInvokerFactory(const miopen::batchnorm::ProblemDescri
...
+    const miopen::batchnorm::ProblemDescription& problem) const
+{
+    const auto& args       = CKArgsBNormFwdTraining{problem};
+    const auto bn_fwd_ptrs = DeviceOpBNFwdTrainingPtrs<XDataType,
+                                                       YDataType,
+                                                       AccDataType,
+                                                       ScaleDataType,
+                                                       BiasDataType,
+                                                       MeanVarDataType>::GetInstances();
+    assert(!bn_fwd_ptrs.empty());
```
DrizztDoUrden: should throw

src/solver/batchnorm/forward_training_ck.cpp
```diff
@@ -174,68 +158,397 @@ static ConvSolution MakeAnyInvokerFactory(const miopen::batchnorm::ProblemDescri
...
+static int CheckCKApplicability(const miopen::batchnorm::ProblemDescription& problem)
+{
+    const auto& args       = CKArgsBNormFwdTraining{problem};
+    const auto bn_fwd_ptrs = DeviceOpBNFwdTrainingPtrs<XDataType,
+                                                       YDataType,
+                                                       AccDataType,
+                                                       ScaleDataType,
+                                                       BiasDataType,
+                                                       MeanVarDataType>::GetInstances();
+    assert(!bn_fwd_ptrs.empty());
```
DrizztDoUrden: should throw

src/solver/batchnorm/forward_training_ck.cpp
```diff
@@ -174,68 +158,397 @@ static ConvSolution MakeAnyInvokerFactory(const miopen::batchnorm::ProblemDescri
...
+{
+    ConvSolution result;
+    auto bn_fwd_ptrs = DeviceOpBNFwdTrainingPtrs<XDataType,
+                                                 YDataType,
+                                                 AccDataType,
+                                                 ScaleDataType,
+                                                 BiasDataType,
+                                                 MeanVarDataType>::GetInstances();
+
+    assert(config.index >= 0 && !bn_fwd_ptrs.empty() && config.index < bn_fwd_ptrs.size());
```
DrizztDoUrden: This should probably throw. But if the asserts above are changed to throws, assert here is probably manageable.

DrizztDoUrden (changes requested):

DrizztDoUrden:
IsApplicable/GetSolution in general should throw instead of asserting for several reasons:
1. Asserts are compiled out when building in release mode. As a consequence, such code can segfault or abort otherwise if assert contains some precondition of the following code. For example:
```C++
assert(i < vec.size());
auto x = vec[i];
```
With `i >= vec.size()` in debug it would produce a meaningful error, while release build would just segfault at the second line.
2. Uses of solver methods are wrapped in `try-catch` blocks, which allows library to skip a faulty problem-solver pair and try next one. Both assert and segfault would terminate entire program.
Other than that, ideally, we want perf config to only contain a single case rather than every possible one and an index. They are copied around and in general interpreted as values in `GenericSearch`. But if the amount of cases is not great, it shouldn't be a problem.

src/ocl/batchnormocl.cpp
bghimireamd: will undo this :)

DrizztDoUrden (approved):
lgtm

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
```diff
@@ -240,7 +240,8 @@ MhaCKFlashAttentionV2Forward::GetSolution([[maybe_unused]] const ExecutionContex
 
             fmha_runtime_args.p_drop = probability;
             fmha_runtime_args.drop_seed_offset =
-                std::make_pair(dataFwd.dropoutSeedData, dataFwd.dropoutOffsetData);
+                std::make_pair(reinterpret_cast<uint64_t>(dataFwd.dropoutSeedData),
```
bghimireamd: @CAHEK7 had to this to silence the compiler.

CAHEK7: I'd like to double check - are you going to convert gpu pointer to integer? Is the code using `fmha_runtime_args.drop_seed_offset` aware that it's actually gpu pointer?
For me that `reinterpret_cast` doesn't look like a correct solution.

src/include/miopen/batchnorm/problem_description.hpp
```diff
@@ -267,6 +306,39 @@ struct MIOPEN_INTERNALS_EXPORT ProblemDescription : ProblemDescriptionBase, Prob
...
+
+    std::string GetModeStr() const
+    {
+        switch(bn_mode)
+        {
+        case miopenBNPerActivation: return "0";
+        case miopenBNSpatial: return "1";
+        default: MIOPEN_THROW(miopenStatusInvalidValue, "Wrong Batchnorm Direction provided");
+        }
+    }
```
CAHEK7: [Q]
Why those are semantically identical functions written in different ways?
(I mean `GetModeStr` and `GetDirectionStr`)

src/solver/batchnorm/backward_ck.cpp
```diff
@@ -161,89 +198,307 @@ template <typename XDataType,
...
+    {
+        if(bn_bwd_ptrs[i]->GetTypeString() == this->kernel_id)
+        {
+            break;
+        }
+    }
+    if(i == valid_kernels.size())
+    {
+        return false;
+    }
```
CAHEK7: `std::find`/`std::find_if`

src/solver/batchnorm/backward_ck.cpp
```diff
@@ -161,89 +198,307 @@ template <typename XDataType,
...
+}
+
+template <typename XDataType,
+          typename DxDataType,
+          typename DyDataType,
+          typename AccDataType,
+          typename ScaleDataType,
+          typename DscaleDbiasDataType,
+          typename MeanVarDataType>
+static int CheckCKApplicability(const miopen::batchnorm::ProblemDescription& problem)
```
CAHEK7: I have doubts that `int` is appropriate return type.
It looks like it must return `bool`. That function is used in `return (F() != -1);` context wich can be perfectly expressed using `bool`.

Probably I missed something, but for example all the convolutions use `bool`.

src/solver/batchnorm/backward_ck.cpp
```diff
@@ -161,89 +198,307 @@ template <typename XDataType,
...
+                                                    nullptr,
+                                                    nullptr,
+                                                    nullptr);
+        if(it->IsSupportedArgument(argument_ptr.get()))
+        {
+            return count;
+        }
+        count++;
+    }
+    return -1;
```
CAHEK7: `std::find_if` + `return it != bn_bwd_ptrs.end();`

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
bghimireamd: yes! its a GPU pointer but I think for now we have disabled the dropout https://github.com/ROCm/MIOpen/blob/develop/src/solver/mha/mha_ck_fa_v2_solver_forward.cpp#L234

src/solver/batchnorm/forward_inference_ck.cpp
```diff
@@ -140,11 +144,140 @@ static int CheckCKApplicability(const miopen::batchnorm::ProblemDescription& pro
...
+    {
+        if(bn_fwd_ptrs[i]->GetTypeString() == this->kernel_id)
+        {
+            break;
+        }
+    }
+    if(i == valid_kernels.size())
+    {
+        return false;
+    }
```
CAHEK7: `std::find`/`std::find_if`

src/solver/batchnorm/backward_ck.cpp
```diff
@@ -161,89 +198,307 @@ template <typename XDataType,
...
-                                 miopen::batchnorm::BwdInvokeParams>(bn_problem, kernel_id);
+#if !MIOPEN_BACKEND_HIP || !MIOPEN_USE_COMPOSABLEKERNEL
+    std::ignore = problem_desc;
+    return false;
+#else
+    if(this->valid_kernels.empty())
+    {
+        this->HeuristicInit(problem_desc);
+        if(valid_kernels.empty())
+            MIOPEN_THROW(miopenStatusInternalError, "BnCKBwdBackward valid_kernels empty");
```
CAHEK7: I think we should have a trust chain somewhere.
Currently `HeuristicInit` guaranteed to return non-empty `valid_kernels` or it will throw otherwise.
It's an absolutely inexpensive check, but doing this after each call in the call-tree does not make a lot of sense for me.

src/solver/batchnorm/backward_ck.cpp
```diff
@@ -161,89 +198,307 @@ template <typename XDataType,
...
+                                                            nullptr,
+                                                            nullptr,
+                                                            nullptr,
+                                                            nullptr,
+                                                            0.0,
+                                                            PassThroughOp{},
+                                                            nullptr,
+                                                            nullptr,
+                                                            nullptr);
+    return bn_bwd_ptrs[i]->IsSupportedArgument(argument_ptr.get());
```
CAHEK7: Haven't we checked this when we were adding the instances into valid kernel list?
`this->kernel_id` seems to be from a `valid_kernels`.

src/solver/batchnorm/backward_ck.cpp
```diff
@@ -161,89 +198,307 @@ template <typename XDataType,
...
+    const auto& args       = CKArgsBNormBwd{problem};
+    const auto bn_bwd_ptrs = DeviceOpBNBwdPtrs<XDataType,
+                                               DxDataType,
+                                               DyDataType,
+                                               AccDataType,
+                                               ScaleDataType,
+                                               DscaleDbiasDataType,
+                                               MeanVarDataType>::GetInstances();
+    if(bn_bwd_ptrs.empty())
+        MIOPEN_THROW(miopenStatusInternalError, "BnCKBwdBackward kernel empty");
```
CAHEK7: It throws in `isApplicable`, is it intended?
I mean is it really serious internal error (like bn_bwd_ptrs must never have been empty), or it just mean that CK is not applicable for some reason?

src/include/miopen/batchnorm/solvers.hpp
```diff
@@ -132,34 +137,210 @@ struct BnFwdInference final : BatchnormSolver
...
+        : index(idx), kernel_id(kernl_id)
+    {
+    }
+    PerformanceConfigBnCKFwdInference() : PerformanceConfigBnCKFwdInference(0, "") {}
+    PerformanceConfigBnCKFwdInference(bool) : PerformanceConfigBnCKFwdInference(0, "") {}
+    MIOPEN_INTERNALS_EXPORT void
+    HeuristicInit(const miopen::batchnorm::ProblemDescription& problem_desc);
+    MIOPEN_INTERNALS_EXPORT bool
+    SetNextValue(const miopen::batchnorm::ProblemDescription& problem_desc);
+    MIOPEN_INTERNALS_EXPORT bool IsValidValue() const;
```
CAHEK7, line 155: It seems unused.

Also what kind of value it validates?

CAHEK7 (commented):
It looks some improvements can be done here. But it will work even without any changes.

The only thing which really concerns me is `reinterpret_cast` in `src/solver/mha/mha_ck_fa_v2_solver_forward.cpp` - it doesn't look like a correct code.

src/solver/mha/mha_ck_fa_v2_solver_forward.cpp
CAHEK7: It would be better to make it explicitly 0 and put the same comment as 
https://github.com/ROCm/MIOpen/blob/0c24160d7368f1194cf24cea713af82ead4275c2/src/solver/mha/mha_ck_fa_v2_solver_forward.cpp#L237-L239

src/include/miopen/batchnorm/solvers.hpp
DrizztDoUrden: ~Unless something changed, it is required by `GenericSearch`. Idea is that search gets initial config and then steps through them, skipping ones that are not expected to produce a working kernel, which is reported by this function.~ I think this particular method is used to validate what has been read from perf-db

junliume:
@bghimireamd Windows builds are failing, the error messages look like something caused by this PR, could you take a look?

CAHEK7 (approved):

junliume (approved):

PR #3329, opened 2024-10-22
[GTests refactoring] Removing usage of  environment variables for DISCRETE tests

Vsevolod1983:
What has been removed:

1. MIOPEN_TEST_MLIR  - Variable is replaced with negative_filter logic for gtests_discovery call.

2. MIOPEN_TEST_COMPOSABLEKERNEL - Variable is replaced with usage of C++ define, which we already have.

3. CODECOV_TEST   - Variable is replaced with negative_filter logic for gtests_discovery call.

4. MIOPEN_TEST_DBSYNC   - Variable is replaced with negative_filter logic for gtests_discovery call.

5. MIOPEN_TEST_CONV - this variable is a bit strange and its purposes is unclear. Just not to break anything I safely replaced this variable with with negative_filter logic for gtests_discovery call

6. MIOPEN_TEST_DEEPBENCH - - Variable is replaced with negative_filter logic for gtests_discovery call.

7. MIOPEN_DEBUG_TUNING_ITERATIONS_MAX - just removed because a value which is being passed does not used inside the tests

8. MIOPEN_TEST_WITH_MIOPENDRIVER - Variable is replaced with usage of C++ define MIOPEN_BUILD_DRIVER

9. Also this PR contains minor cleanup.

Sometimes, In order to replace variable with negative_filter logic it was necessary to change test names a little bit.
For example, If we want to exclude Code Coverage related tests from being executed we use "*CodecovTest*" filter for this, so all tests that are related to Code Coverage should contain "CodecovTest" substring inside their names.


CAHEK7 (approved):

junliume (approved):

test/gtest/CMakeLists.txt
```diff
@@ -46,25 +59,38 @@ function(add_gtest TEST_NAME TEST_CPP)
...
+        add_gtest_negative_filter("*DeepBench*")
+    endif()      
+
+    if (NOT CODECOV_TEST)
+        add_gtest_negative_filter("*CodecovTest*")
+    endif()      
+
+    if (NOT MIOPEN_TEST_CONV)
+        add_gtest_negative_filter("*MIOpenTestConv*")
+    endif()
```
averinevg, line 91: @Vsevolod1983 @CAHEK7 @junliume We should add this to the wiki.

PR #3330, opened 2024-10-23
[tuning] Interface for Fin

averinevg:
This PR adds interface for `Fin`


src/include/miopen/fin/fin_interface.hpp
```diff
@@ -0,0 +1,247 @@
...
+
+// Convolution solver
+class ConvSolver : public SolverMixin<miopen::ExecutionContext, miopen::conv::ProblemDescription>
+{
+public:
+    std::string GetAlgo(miopen::conv::Direction dir) const;
+
+protected:
+    ConvSolver(const miopen::solver::SolverBase* solver_base,
+               uint64_t solver_id,
```
DrizztDoUrden, line 136: [Suggestion] I would recommend using `solver::Id` instead of numeric ids. It would be hassle free to create this from either number or string.

DrizztDoUrden (approved):

junliume:
@averinevg please resolve the conflicts and a few hip tidy issue. Thanks.

alexandraBara (approved):

test/gtest/unit_FinInterface.cpp
```diff
@@ -0,0 +1,620 @@
...
+    return params;
+}
+
+template <class SolverInfo>
+const auto& GetSolversInfo();
+
+template <>
+const auto& GetSolversInfo<ConvSolverInfo>()
+{
+    static const std::unordered_map<std::string, ConvSolverInfo> solver_info = {
```
randyspauldingamd, line 208: Wow, what a list! Is this generated, or maintained by hand? There are likely to be many additions coming with MX data types. Seems like it wouldn't be too hard to import it from a file that would be easier for external tools to interact with (e.g. json, csv).

In either case, it would be nice to have placeholders for the missing ordinals. That way we can know at a glance if an algo was deleted or otherwise expected to be missing. Otherwise, it looks like something may have been missed.

averinevg: This list is filled manually. It contains all the solvers for the specified primitives currently available in the library. This is needed to check that the interface provides correct data. However, we don't need such a large list; a dozen solvers would be sufficient. There is also no need to update it when new solvers are added.

src/fin/fin_interface.cpp
```diff
@@ -0,0 +1,592 @@
...
+    if(!sbase->IsTunable())
+    {
+        SetObjectNonTunable();
+        return;
+    }
+
+    switch(id)
+    {
+    case 1: SetObject<miopen::solver::conv::ConvAsm3x3U>(); break;
+    case 2: SetObject<miopen::solver::conv::ConvAsm1x1U>(); break;
```
CAHEK7, line 286: Since it's handcrafted, I think we need some sort of instruction how to keep it synchronized (Here and in the solver counterpart, probably).

src/fin/fin_interface.cpp
```diff
@@ -0,0 +1,592 @@
...
+    // New tunable solver should be added here
+    default: MIOPEN_THROW(miopenStatusInternalError, "Unknown solver ID");
+    }
+}
+
+template <>
+AnySolver<miopen::ExecutionContext, miopen::batchnorm::ProblemDescription>::AnySolver(
+    const miopen::solver::SolverBase* solver_base, uint64_t id)
+    : sbase(solver_base)
+{
```
CAHEK7, line 351: Probably it worth to keep each solver-specific part in a separate file. Otherwise, it can become really huge.
Like
```
fin_interface.cpp
fin_conv.cpp
fin_batchnorm.cpp
...etc...
```
It's just a recommendation. Or we can even keep it somewhere around the other solver-related files, like `solver/conv/fin_conv.cpp`

CAHEK7 (approved):

src/fin/fin_interface.cpp
averinevg: I will move the creation of the auxiliary class into the constructor of the interface solver, so the check will occur automatically. When adding a new tunable solver without making changes to the interface, an exception will be raised, and the CI will fail.

src/fin/fin_interface.cpp
averinevg: Good point. Right now, it's not an issue, but it's worth thinking about how to ensure it doesn't turn into a mess in the future.

alexandraBara (approved):
LGTM

src/include/miopen/batchnorm/solvers.hpp
```diff
@@ -133,6 +133,7 @@ struct BnFwdInference final : BatchnormSolver
 
     bool IsApplicable(const ExecutionContext& context,
                       const miopen::batchnorm::ProblemDescription& problem) const override;
+    bool IsDynamic() const override { return true; }
```
bghimireamd, line 136: @averinevg Why did we marked `BnFwdInference` as dynamic?

averinevg: I got it from [here](https://github.com/ROCm/MIOpen/blob/037eb6dc436b1cf405fe37f7a2304f4dcf5c29d0/src/include/miopen/solver.hpp#L76). Is there something wrong with this?

PR #3343, opened 2024-10-28
General db update for gfx942

cderb:
General db update for tracking configurations on gfx942.


junliume (approved):

junliume:
@cderb there are strange CI issues, we cannot merge until these are resolved.

cderb:
Issued discovered: MIOpen isn't reading sqlite3 .db file from system for perf results. Will need to correct MIOpen pdb lookup behavior, or modify tuna export script to produce db.txt files

cderb:
> Issued discovered: MIOpen isn't reading sqlite3 .db file from system for perf results. Will need to correct MIOpen pdb lookup behavior, or modify tuna export script to produce db.txt files

Perceived issue wasn't an issue. cmake triggered sqlite2txt function is properly translating the db files. But now we are exporting db.txt files anyway.

PR #3346, opened 2024-10-30
TensorOps kernels refactoring

novakovicdj:
This is draft PR for refactoring tensor ops kernels to solver structure, so far only Op1dTensorGeneric kernel is switched


src/tensor/problem_description.cpp
```diff
@@ -0,0 +1,67 @@
...
+ *******************************************************************************/
+
+#include <miopen/tensor/problem_description.hpp>
+#include <miopen/names.hpp>
+
+namespace miopen {
+
+namespace tensor {
+
+NetworkConfig ProblemDescription::MakeNetworkConfig() const
```
Vsevolod1983: Several questions about PD (Problem Description) network config:
1. We have alphas and beta in PD  Should we add them in NetworkConfig ? Or probably we need remove alphas from PD at all  and add only beta to NetworkConfig?
2.  Should we add tensor sizes to NetworkConfig ?
@CAHEK7  your thoughts ?

CAHEK7: 1 and 2: each problem descriptor should provide a unique network config, so sizes, strides, alphas, betas, etc. must be a part of the network config. Regarding to alphas and betas - most probably the values are not important, usually the only thing is important is it zero or not.

1: parameters from the problem description and from the invoke params should not duplicate each other.
problem description represents static problem parameters, may affect compilation and performance cannot be changed in runtime, while invoke params represents runtime parameters which can be changed for a particular problem (for example data pointers). In this particular case, we should check which parameter can be a pure runtime parameter (alphas is a good candidate, while beta=0 or beta!=0 may affect compilation) and move them outside of the problem description.

src/include/miopen/tensor/solvers.hpp
```diff
@@ -0,0 +1,94 @@
...
+    ConvSolution GetSolution(const ExecutionContext& context,
+                             const miopen::tensor::ProblemDescription& problem) const override;
+
+    std::size_t GetWorkspaceSize(const ExecutionContext& context,
+                                 const miopen::tensor::ProblemDescription& problem) const override;
+
+    bool MayNeedWorkspace() const override { return false; }
+};
+
+} // namespace tensor
```
CAHEK7: I think it would be better to rename it into `OpTensor` or `TensorOp` or something less generic than just a `tensor`.
The algorithm represents tensor operations, not tensors.

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+    auto alens       = aTensorDesc.GetLengths();
+    auto blens       = bTensorDesc.GetLengths();
+    auto asize       = alens.size();
+
+    if(asize == 1)
+    {
+        return true;
+    }
+    // add support for this later
+    // if(asize == 2 && ((blens[0] == 1 && blens[1] == 1) || (blens[0] > 1 && blens[1] > 1)))
```
CAHEK7: This condition covers implicit 2d cases which can be mapped into 1d only for packed tensors, but there are explicit 2d cases which are actual 1d, for example  A:`1xC` and B:`1xC` - it's a prefect 1d case for any strides and current condition does not cover it and does not take strides into account.

src/tensor/problem_description.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+        bool is_lite = clens[0] == 1 && blens[0] == 1 && alens[0] == 1 &&
+                       (blens[1] == clens[1] || blens[1] == 1) && blens[2] == clens[2];
+
+        if(lite_applicable && is_lite)
+        {
+            ss << "-" << std::to_string(RD_BLCK) << "x" << std::to_string(local_threads) << "x"
+               << std::to_string(grp_sz) << std::to_string(local_threads2)
+               << std::to_string(grp_sz2);
+        }
```
CAHEK7: There is no need for those kinds of calculations. All those parameters are derivatives of the sizes and strides, and they are implementation specific.
Previously it was kind of hardcoded for a group of kernels, but now problem descriptor and its network config are generic description of the requested algorithm, and they should contain minimum generic set of parameters which are unique for each particular problem.

Shortly: strides for all tensors and sizes for A and B tensors are required, while local/global threadcount, groups, etc. are not.

src/include/miopen/tensor/problem_description.hpp
```diff
@@ -0,0 +1,152 @@
...
+struct NetworkConfig;
+
+namespace tensor {
+
+struct ProblemDescription : ProblemDescriptionBase
+{
+    ProblemDescription(const miopenTensorOp_t tensorOp_,
+                       const void* alpha0_,
+                       const void* alpha1_,
+                       const void* beta_,
```
CAHEK7: Check this conversation
https://github.com/ROCm/MIOpen/pull/3346/files#r1824480257

Probably `alpha0/1` must not be a part of the PD, ideally `beta` as well, but right now it has to be there..

src/include/miopen/tensor/problem_description.hpp
```diff
@@ -0,0 +1,152 @@
...
+            MIOPEN_THROW(miopenStatusBadParm, "Alpha1 value is nullptr");
+        }
+        if(beta_ == nullptr)
+        {
+            MIOPEN_THROW(miopenStatusBadParm, "Beta value is nullptr");
+        }
+
+        alpha0 = alpha0_;
+        alpha1 = alpha1_;
+        beta   = beta_;
```
CAHEK7: I'd rather kept those parameters as values, not pointers.
For example, check how it's done for convolutions.

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    if(aTensorDesc.AllDimsFitIntoInt())
+    {
+        build_params.Define("DIM_TYPE", "uint32_t");
+    }
+    else
+    {
+        build_params.Define("DIM_TYPE", "uint64_t");
+    }
```
CAHEK7: Could you make a header and put common helpers there? I guess it's identical for all the tensorops solvers.
Think about how to generalize most of the stuff and avoid lots of copy-pasted code.

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+}
+
+ConvSolution Op1dTensorGeneric::GetSolution(const ExecutionContext& context,
+                                            const miopen::tensor::ProblemDescription& problem) const
+{
+    auto result = ConvSolution{miopenStatusSuccess};
+
+    auto aTensorDesc = problem.GetATensorDesc();
+    auto bTensorDesc = problem.GetBTensorDesc();
+    auto cTensorDesc = problem.GetCTensorDesc();
```
CAHEK7: It actually copies descriptors one more time, it's better to use explicit `const auto&`.

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+                auto miopen_alpha0 = as_float(*(static_cast<const float*>(params.alpha0)));
+                auto miopen_alpha1 = as_float(*(static_cast<const float*>(params.alpha1)));
+                auto miopen_beta   = as_float(*(static_cast<const float*>(params.beta)));
+
+                auto blens = params.bTensorDesc.GetLengths();
+                auto clens = params.cTensorDesc.GetLengths();
+
+                auto astrides = params.aTensorDesc.GetStrides();
+                auto bstrides = params.bTensorDesc.GetStrides();
+                auto cstrides = params.cTensorDesc.GetStrides();
```
CAHEK7: Pure `auto` copies the vectors, it's better to use `const auto&` in that case.

src/solver/tensor/Op2dTensorLite.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+    auto asize = alens.size();
+
+    size_t local_threads = 256;
+    int max_num_wg       = 4096;
+
+    // for naive tensor ops
+    size_t RD_BLCK    = (clens[2] % 4 == 0) ? 4 : (clens[2] % 2 == 0) ? 2 : 1;
+    size_t total_work = std::max(clens[2] / RD_BLCK, size_t(1));
+    size_t grp_sz     = (total_work + local_threads - 1) / local_threads;
```
CAHEK7: It looks common for `isApplicable` and `getSolution` and probably for the other solvers.

CAHEK7 (changes requested):

src/include/miopen/tensor/invoke_params.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+    TensorDescriptor bTensorDesc;
+    ConstData_t BTensor;
+
+    TensorDescriptor cTensorDesc;
+    Data_t CTensor;
+
+    size_t Aoffset;
+    size_t Boffset;
+    size_t Coffset;
```
DrizztDoUrden: Do we need to handle this internally? IIRC it should be possible to externally pass any subtensor via changing pointer+descriptor. If so this is a duplicated functionality

src/include/miopen/tensor/invoke_params.hpp
```diff
@@ -0,0 +1,97 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2023 Advanced Micro Devices, Inc.
```
DrizztDoUrden: 2024

src/include/miopen/tensor/problem_description.hpp
```diff
@@ -0,0 +1,152 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2023 Advanced Micro Devices, Inc.
```
DrizztDoUrden: 2024

src/include/miopen/tensor/problem_description.hpp
DrizztDoUrden: Would a bool marking if alpha0/... has a "default" value meaning no additional work required suffice?

src/include/miopen/tensor/problem_description.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+        if(blens.size() != clens.size())
+        {
+            MIOPEN_THROW("Number of dims in B and C Tensors do not match: " +
+                         std::to_string(blens.size()) + ", " + std::to_string(clens.size()));
+        }
+
+        if(!nonStandardSquash)
+        {
+            for(std::size_t i = 0; i < clens.size(); i++)
```
DrizztDoUrden: https://en.cppreference.com/w/cpp/algorithm/all_any_none_of
this would make it a one-liner with clear intent

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+namespace tensor {
+
+bool Op1dTensorGeneric::IsApplicable(const ExecutionContext& context,
+                                     const miopen::tensor::ProblemDescription& problem) const
+{
+    auto aTensorDesc = problem.GetATensorDesc();
+    auto bTensorDesc = problem.GetBTensorDesc();
+    auto alens       = aTensorDesc.GetLengths();
+    auto blens       = bTensorDesc.GetLengths();
```
DrizztDoUrden: At least some of these should be references (probably const ones)

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+    kernel.comp_options = build_params.GenerateFor(
+        kbp::HIP{}); // GetDataTypeKBP(aTensorDesc.GetType()).GenerateFor(kbp::HIP{});
+    kernel.kernel_file = "MIOpenTensorKernelsHip.cpp";
+    kernel.kernel_name = "Op1dTensorGeneric";
+
+    for(uint32_t i = 0; i <= 2; i++)
+    {
+        kernel.l_wk.push_back(vld[i]);
+        kernel.g_wk.push_back(vgd[i]);
+    }
```
DrizztDoUrden:
```suggestion
    using std::begin, std::end;
    kernel.l_wk.insert(end(kernel.l_wk), begin(vld), end(vld));
    kernel.g_wk.insert(end(kernel.g_wk), begin(vgd), end(vgd));
```

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    size_t local_threads = 256;
+    size_t max_num_wg    = 4096;
+
+    auto num_wg           = std::clamp(clens[0] / local_threads, size_t(1), size_t(max_num_wg));
+    num_wg                = num_wg > max_num_wg ? max_num_wg : num_wg;
+    size_t global_threads = num_wg * local_threads;
+
+    const std::vector<size_t> vld{local_threads, 1, 1};
+    const std::vector<size_t> vgd{global_threads, 1, 1};
```
DrizztDoUrden: Looks like std::array<size_t, 3> would suffice here

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+        kernel.l_wk.push_back(vld[i]);
+        kernel.g_wk.push_back(vgd[i]);
+    }
+
+    result.invoker_factory = [=](const std::vector<Kernel> kernels) {
+        return [=](const Handle& handle_, const AnyInvokeParams& raw_params) {
+            decltype(auto) kernel = handle_.Run(kernels.front());
+            decltype(auto) params = raw_params.CastTo<miopen::tensor::InvokeParams>();
+
+            visit_float(bTensorDesc.GetType(), [&](auto as_float) {
```
DrizztDoUrden:
```suggestion
            visit_float(params.bTensorDesc.GetType(), [&](auto as_float) {
```

src/solver/tensor/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,195 @@
...
+    kernel.kernel_file = "MIOpenTensorKernelsHip.cpp";
+    kernel.kernel_name = "Op1dTensorGeneric";
+
+    for(uint32_t i = 0; i <= 2; i++)
+    {
+        kernel.l_wk.push_back(vld[i]);
+        kernel.g_wk.push_back(vgd[i]);
+    }
+
+    result.invoker_factory = [=](const std::vector<Kernel> kernels) {
```
DrizztDoUrden:
```suggestion
    result.invoker_factory = [](const std::vector<Kernel> kernels) {
```

src/include/miopen/tensor_ops.hpp
```diff
@@ -189,6 +189,22 @@ MIOPEN_INTERNALS_EXPORT void OpTensor(const Handle& handle,
                                       size_t Coffset         = 0,
                                       bool nonStandardSquash = false);
 
+MIOPEN_INTERNALS_EXPORT void OpTensorNew(Handle& handle,
```
DrizztDoUrden: I would suggest a renaming it to OpTensor2
name OpTensorNew would not age well

DrizztDoUrden (changes requested):

src/include/miopen/tensor/problem_description.hpp
novakovicdj: what is your idea on how to use any of those functions for this? I think that it is hard to use those with two iterators etc. so maybe this old implementation is better

src/solver/tensorOp/tensor_op_helpers.hpp
```diff
@@ -0,0 +1,217 @@
...
+#include <miopen/datatype.hpp>
+
+namespace miopen {
+
+namespace solver {
+
+namespace tensorOp {
+
+inline void GetCommonParams(KernelBuildParameters& build_params,
+                            miopen::tensorOp::ProblemDescription problem,
```
CAHEK7:
```suggestion
                            const miopen::tensorOp::ProblemDescription& problem,
```

src/solver/tensorOp/tensor_op_helpers.hpp
```diff
@@ -0,0 +1,217 @@
...
+    }
+}
+
+inline void
+GetRDBLCKandREADTYPE(size_t len, miopenDataType_t type, size_t& RD_BLCK, std::string& READ_TYPE)
+{
+    RD_BLCK                     = (len % 4 == 0) ? 4 : (len % 2 == 0) ? 2 : 1;
+    const std::string data_type = GetDataType(type);
+    READ_TYPE                   = (RD_BLCK == 1) ? data_type : data_type + std::to_string(RD_BLCK);
+}
```
CAHEK7:
```suggestion
inline std::tuple<size_t, std::string>
GetRDBLCKandREADTYPE(size_t len, miopenDataType_t type)
{
    const std::string data_type = GetDataType(type);
    size_t RD_BLCK = (len % 4 == 0) ? 4 : (len % 2 == 0) ? 2 : 1;
    return std::make_tuple(RD_BLCK, (RD_BLCK == 1) ? data_type : data_type + std::to_string(RD_BLCK));
}
```
Check this out:
https://godbolt.org/z/M8rzv5z6T

src/solver/tensorOp/tensor_op_helpers.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+inline void GetBitmapAndWgInfo(const std::vector<size_t>& blens,
+                               const std::vector<size_t>& clens,
+                               int& num_wg,
+                               int& work_per_wg,
+                               unsigned int& bitmap)
```
CAHEK7:
```suggestion
inline std::tuple<int, int, unsigned int>
GetBitmapAndWgInfo(const std::vector<size_t>& blens,
                   const std::vector<size_t>& clens)
```

src/solver/tensorOp/tensor_op_helpers.hpp
```diff
@@ -0,0 +1,217 @@
...
+        }
+        else
+        {
+            work_per_wg *= clens[i];
+        }
+    }
+}
+
+inline void
+IsBitmapLeadingOnes(unsigned int bitmap, int n_size, int first_not_one, bool& leading_ones)
```
CAHEK7:
```suggestion
inline bool
IsBitmapLeadingOnes(unsigned int bitmap, int n_size, int first_not_one)
```

src/solver/tensorOp/tensor_op_helpers.hpp
```diff
@@ -0,0 +1,217 @@
...
+}
+
+inline void Get4dParams(const miopen::tensorOp::ProblemDescription& problem,
+                        bool is4dLite,
+                        int& num_wg_orig,
+                        int& work_per_wg,
+                        int& incr_wg,
+                        unsigned int& bitmap,
+                        size_t& local_threads,
+                        size_t& global_threads)
```
CAHEK7: I'm not sure about this one, but probably it should also return a tuple.
The only problem is a structure binding capturing in lambdas, that won't work in C++17:
```C++
auto [foo, bar] = GetFooBar(val);
auto useFooBar = [=]()
{
    printf("%d, %d", foo, bar);
};
```
You can use `std::tie`, but it's ugly.

src/tensorOp/problem_description.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+        if(i != (alens.size() - 1))
+        {
+            alens_str += "x";
+            blens_str += "x";
+            astrides_str += "x";
+            bstrides_str += "x";
+            cstrides_str += "x";
+        }
+    }
```
CAHEK7:
```suggestion
    auto printDims = [&ss] (const auto& dim)
    {
        for(uint32_t i = 0; i < alens.size(); i++)
        {
            ss << alens[i];
            if(i != (alens.size() - 1))
            {
                ss << "x";
            }
        }
        ss << "-";
    }

    ss << std::to_string(aTensorDesc.GetType()) << "-" << std::to_string(tensorOp) << "-"
    printDims(alens);
    printDims(blens);
    ...etc...
    ss << std::to_string((beta == 0));
```

src/tensorOp/problem_description.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    ss << std::to_string(aTensorDesc.GetType()) << "-" << std::to_string(tensorOp) << "-"
+       << alens_str << "-" << blens_str << "-" << astrides_str << "-" << bstrides_str << "-"
+       << cstrides_str << "-" << std::to_string((beta == 0));
```
CAHEK7: `std::to_string((beta == 0));` -> `(float_equal(beta, 0.0f) ? "1" : "0")`

src/solver/tensorOp/Op1dTensorGeneric.cpp
```diff
@@ -0,0 +1,179 @@
...
+bool Op1dTensorGeneric::IsApplicable(const ExecutionContext& context,
+                                     const miopen::tensorOp::ProblemDescription& problem) const
+{
+    const auto& aTensorDesc = problem.GetATensorDesc();
+    // const auto& bTensorDesc = problem.GetBTensorDesc();
+    const auto& alens = aTensorDesc.GetLengths();
+    // const auto& blens       = bTensorDesc.GetLengths();
+    auto asize       = alens.size();
+
+    if(GetDataType(aTensorDesc.GetType()) == "double")
```
CAHEK7: String comparison does not make much sense.
```suggestion
    if(aTensorDesc.GetType() == miopenDouble)
```
It should be fixed in all the solvers

src/solver/tensorOp/Op2dTensorLite.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+    if(GetDataType(aTensorDesc.GetType()) == "double")
+    {
+        return false;
+    }
+
+    if(asize < 3)
+    {
+        return false;
+    }
```
CAHEK7: It can be applicable in 2d case, it forces it explicitly in this statement `clens[0] == 1 && blens[0] == 1 && alens[0] == 1`.
But it introduces some complications later.

Probably that `isApplicable` tuning can be done in a separate PR.

~Also it's not a `asize < 3`, it must be `asize != 3`. That should be fixed in this PR, because if `asize == 4`, it'll break everything in `GetSolution`~

src/solver/tensorOp/Op2dTensorSquash.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    if(GetDataType(aTensorDesc.GetType()) == "double")
+    {
+        return false;
+    }
+
+    if(asize < 3)
+    {
+        return false;
+    }
```
CAHEK7: See comments for the `2dLite`.

src/solver/tensorOp/Op4dTensorLite.cpp
```diff
@@ -0,0 +1,175 @@
...
+    int work_per_wg     = 0;
+    unsigned int bitmap = 0;
+
+    GetBitmapAndWgInfo(blens, clens, num_wg, work_per_wg, bitmap);
+
+    // quick fix for btensor = <1, 1, 1, 1>
+    if(bTensorDesc.GetElementSize() == 1)
+        bitmap = 4;
+
+    auto fwd_conv_bias = bitmap == (1 << 2) ? 1 : 0;
```
CAHEK7: That must be `bool`

src/solver/tensorOp/Op4dTensorLite.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    bool packed_equal_tensor =
+        packed_tensor && (bTensorDesc.GetElementSize() == cTensorDesc.GetElementSize());
+
+    if(GetDataType(aTensorDesc.GetType()) == "double")
+    {
+        return false;
+    }
+
+    if(asize == 4 && fwd_conv_bias == 0 && packed_equal_tensor)
```
CAHEK7: I think `asize == 4` should be the first condition in that function, before all those `bitmap` and other calculations.

src/solver/tensorOp/OpTensorFwdBias.cpp
```diff
@@ -0,0 +1,215 @@
...
+        bitmap = 4;
+
+    auto fwd_conv_bias = bitmap == (1 << 2) ? 1 : 0;
+
+    if(GetDataType(aTensorDesc.GetType()) == "double")
+    {
+        return false;
+    }
+
+    if(asize == 4 && fwd_conv_bias != 0)
```
CAHEK7: See `4dLite` comments.

CAHEK7 (changes requested):

src/solver/tensorOp/Op2dTensorLite.cpp
novakovicdj: I think that it is better to keep `isApplicable` straightforward and same applications as they are in `tensorocl.cpp` file. That additional tuning can be done later in separate PR, I can also open an issue for that to be completed later.

shurale-nkn:
Please provide a comparison of the average only CPU time (new solver vs old api) measurements for 100 calls with same problem  and the costs associated with the first call of the unique problem configuration.

novakovicdj:
> Please provide a comparison of the average only CPU time (new solver vs old api) measurements for 100 calls with same problem and the costs associated with the first call of the unique problem configuration.

Here is a comparison of average host time between old and new structure
| Kernel | | New structure [ms] | Old structure [ms] | diff [ms] |
| --- | --- | --- | --- | --- |
| Op1dTensorGeneric | first run | 279.3786 | 291.3806 | -12.002 |
| | other 100 runs | 0.2908 | 0.2549 | 0.0359 |
| Op2dTensorGeneric | first run | 281.8186 | 283.4622 | -1.6436 |
| | other 100 runs | 0.356 | 0.2432 | 0.1128 |
| Op2dTensorLite | first run | 634.2228 | 662.2278 | -28.005 |
| | other 100 runs | 0.335 | 0.2308 | 0.1042 |
| Op2dTensorSquash | first run | 668.978 | 699.9932 | -31.0152 |
| | other 100 runs | 0.3481 | 0.2548 | 0.0933 |
| Op3dTensorGeneric | first run | 642.1512 | 656.3394 | -14.1882 |
| | other 100 runs | 0.2659 | 0.2485 | 0.0174 |
| OpTensorFwdBias | first run | 636.6204 | 654.8222 | -18.2018 |
| | other 100 runs | 0.3351 | 0.2321 | 0.103 |
| OpTensorFwdBiasGeneric | first run | 636.4756 | 662.4915 | -26.0159 |
| | other 100 runs | 0.3498 | 0.2434 | 0.1064 |
| OpTensorLeadingOnes | first run | 644.8348 | 666.8713 | -22.0365 |
| | other 100 runs | 0.3466 | 0.2755 | 0.0711 |
| OpTensorLeadingOnesGeneric | first run | 648.6535 | 669.6379 | -20.9844 |
| | other 100 runs | 0.3552 | 0.2569 | 0.0983 |
| Op4dTensorLite | first run | 641.4747 | 664.4976 | -23.0229 |
| | other 100 runs | 0.33 | 0.2206 | 0.1094 |
| Op4dTensorGeneric | first run | 650.7638 | 670.8961 | -20.1323 |
| | other 100 runs | 0.3563 | 0.2456 | 0.1107 |
| Op5dTensorGeneric | first run | 655.6774 | 685.431 | -29.7536 |
| | other 100 runs | 0.3745 | 0.2437 | 0.1308 |

New structure is faster on average for 20ms for first runs and it is slower for 0.1ms for other 100 calls or 0.001ms per call

shurale-nkn:
> > Please provide a comparison of the average only CPU time (new solver vs old api) measurements for 100 calls with same problem and the costs associated with the first call of the unique problem configuration.
> 
> Here is a comparison of average host time between old and new structure
> 
> Kernel		New structure [ms]	Old structure [ms]	diff [ms]
> Op1dTensorGeneric	first run	279.3786	291.3806	-12.002
> other 100 runs	0.2908	0.2549	0.0359
> Op2dTensorGeneric	first run	281.8186	283.4622	-1.6436
> other 100 runs	0.356	0.2432	0.1128
> Op2dTensorLite	first run	634.2228	662.2278	-28.005
> other 100 runs	0.335	0.2308	0.1042
> Op2dTensorSquash	first run	668.978	699.9932	-31.0152
> other 100 runs	0.3481	0.2548	0.0933
> Op3dTensorGeneric	first run	642.1512	656.3394	-14.1882
> other 100 runs	0.2659	0.2485	0.0174
> OpTensorFwdBias	first run	636.6204	654.8222	-18.2018
> other 100 runs	0.3351	0.2321	0.103
> OpTensorFwdBiasGeneric	first run	636.4756	662.4915	-26.0159
> other 100 runs	0.3498	0.2434	0.1064
> OpTensorLeadingOnes	first run	644.8348	666.8713	-22.0365
> other 100 runs	0.3466	0.2755	0.0711
> OpTensorLeadingOnesGeneric	first run	648.6535	669.6379	-20.9844
> other 100 runs	0.3552	0.2569	0.0983
> Op4dTensorLite	first run	641.4747	664.4976	-23.0229
> other 100 runs	0.33	0.2206	0.1094
> Op4dTensorGeneric	first run	650.7638	670.8961	-20.1323
> other 100 runs	0.3563	0.2456	0.1107
> Op5dTensorGeneric	first run	655.6774	685.431	-29.7536
> other 100 runs	0.3745	0.2437	0.1308
> New structure is faster on average for 20ms for first runs and it is slower for 0.1ms for other 100 calls or 0.001ms per call

The results are very strange; we need to obtain the experiment protocol. How was the program executed, and what was used for measurement?
so far, according to the table, each subsequent launch is on average 30% slower

novakovicdj:
> > > Please provide a comparison of the average only CPU time (new solver vs old api) measurements for 100 calls with same problem and the costs associated with the first call of the unique problem configuration.
> > 
> > 
> > Here is a comparison of average host time between old and new structure
> > Kernel		New structure [ms]	Old structure [ms]	diff [ms]
> > Op1dTensorGeneric	first run	279.3786	291.3806	-12.002
> > other 100 runs	0.2908	0.2549	0.0359
> > Op2dTensorGeneric	first run	281.8186	283.4622	-1.6436
> > other 100 runs	0.356	0.2432	0.1128
> > Op2dTensorLite	first run	634.2228	662.2278	-28.005
> > other 100 runs	0.335	0.2308	0.1042
> > Op2dTensorSquash	first run	668.978	699.9932	-31.0152
> > other 100 runs	0.3481	0.2548	0.0933
> > Op3dTensorGeneric	first run	642.1512	656.3394	-14.1882
> > other 100 runs	0.2659	0.2485	0.0174
> > OpTensorFwdBias	first run	636.6204	654.8222	-18.2018
> > other 100 runs	0.3351	0.2321	0.103
> > OpTensorFwdBiasGeneric	first run	636.4756	662.4915	-26.0159
> > other 100 runs	0.3498	0.2434	0.1064
> > OpTensorLeadingOnes	first run	644.8348	666.8713	-22.0365
> > other 100 runs	0.3466	0.2755	0.0711
> > OpTensorLeadingOnesGeneric	first run	648.6535	669.6379	-20.9844
> > other 100 runs	0.3552	0.2569	0.0983
> > Op4dTensorLite	first run	641.4747	664.4976	-23.0229
> > other 100 runs	0.33	0.2206	0.1094
> > Op4dTensorGeneric	first run	650.7638	670.8961	-20.1323
> > other 100 runs	0.3563	0.2456	0.1107
> > Op5dTensorGeneric	first run	655.6774	685.431	-29.7536
> > other 100 runs	0.3745	0.2437	0.1308
> > New structure is faster on average for 20ms for first runs and it is slower for 0.1ms for other 100 calls or 0.001ms per call
> 
> The results are very strange; we need to obtain the experiment protocol. How was the program executed, and what was used for measurement? so far, according to the table, each subsequent launch is on average 30% slower

I have talked with @CAHEK7 and he suggested to remove compilation time from first run time and to enable kernel profiling so here are new results that I have got
note. diff for first run is (new_total_time - new_comp_time) - (old_total_time - old_comp_time)

Kernel | run | New structure [ms] (total) | New structure [ms] (compilation) | Old structure [ms] (total) | Old structure [ms] (compilation) |diff [ms] |
-- | -- | -- | -- | -- | -- | -- |
Op1dTensorGeneric | First | 279.0176 | 278.813 | 289.2538 | 289.0891 | 0.0399
  | Other 100 | 2.3191 |   | 2.5519 |   | -0.2328
Op2dTensorGeneric | First | 282.0158 | 281.8159 | 285.6689 | 285.4783 | 0.0093
  | Other 100 | 2.318 |   | 2.3533 |   | -0.0353
Op2dTensorLite | First | 645.8057 | 645.6093 | 662.8484 | 662.6658 | 0.0138
  | Other 100 | 2.3686 |   | 2.275 |   | 0.0936
Op2dTensorSquash | First | 680.7002 | 680.5031 | 709.6122 | 709.4446 | 0.0295
  | Other 100 | 2.5681 |   | 2.2783 |   | 0.2898
Op3dTensorGeneric | First | 641.8974 | 641.701 | 695.5571 | 695.3875 | 0.0268
  | Other 100 | 2.471 |   | 2.1212 |   | 0.3498
OpTensorFwdBias | First | 640.256 | 640.062 | 680.8951 | 680.709 | 0.0079
  | Other 100 | 2.3217 |   | 2.4506 |   | -0.1289
OpTensorFwdBiasGeneric | First | 671.2975 | 671.0999 | 686.9962 | 686.8118 | 0.0132
  | Other 100 | 2.5107 |   | 2.3875 |   | 0.1232
OpTensorLeadingOnes | First | 676.697 | 676.5015 | 660.4297 | 660.2519 | 0.0177
  | Other 100 | 2.5548 |   | 2.4725 |   | 0.0823
OpTensorLeadingOnesGen | First | 656.037 | 655.834 | 677.3558 | 677.18 | 0.0272
  | Other 100 | 2.7894 |   | 2.4013 |   | 0.3881
Op4dTensorLite | First | 662.9597 | 662.7599 | 667.7071 | 667.5309 | 0.0236
  | Other 100 | 2.5528 |   | 2.2425 |   | 0.3103
Op4dTensorGeneric | First | 659.8318 | 659.6297 | 673.9867 | 673.8067 | 0.0221
  | Other 100 | 2.3756 |   | 2.3735 |   | 0.0021
Op5dTensorGeneric | First | 654.421 | 654.2273 | 693.2979 | 693.1147 | 0.0105
  | Other 100 | 2.2762 |   | 2.187 |   | 0.0892


</body>

</html>


This is part of code that I changed in tensor_ops test for this performance testing

`       handle.EnableProfiling(true);

        auto start = std::chrono::steady_clock::now();
        miopen::OpTensor2(handle,
                          // miopenTensorOpAdd,
                          // miopenTensorOpMax,
                          // miopenTensorOpMin,
                          miopenTensorOpMul,
                          &alpha0,
                          a.desc,
                          a_dev.get(),
                          &alpha1,
                          b.desc,
                          b_dev.get(),
                          &beta,
                          c.desc,
                          c_dev.get(),
                          Aoffset,
                          Boffset,
                          Coffset,
                          false); // it does not verify non-standard behaviour
        auto end = std::chrono::steady_clock::now();
        auto elapsed_init =
            std::chrono::duration_cast<std::chrono::duration<float, std::milli>>(end - start)
                .count() -
            handle.GetKernelTime();

        printf("First time: %f ms\n", elapsed_init);

        float elapsed_other_full = 0;

        for(int i = 0; i < 100; i++)
        {
            start = std::chrono::steady_clock::now();
            miopen::OpTensor2(handle,
                              // miopenTensorOpAdd,
                              // miopenTensorOpMax,
                              // miopenTensorOpMin,
                              miopenTensorOpMul,
                              &alpha0,
                              a.desc,
                              a_dev.get(),
                              &alpha1,
                              b.desc,
                              b_dev.get(),
                              &beta,
                              c.desc,
                              c_dev.get(),
                              Aoffset,
                              Boffset,
                              Coffset,
                              false); // it does not verify non-standard behaviour
            end = std::chrono::steady_clock::now();
            auto elapsed_other =
                std::chrono::duration_cast<std::chrono::duration<float, std::milli>>(end - start)
                    .count() -
                handle.GetKernelTime();
            elapsed_other_full += elapsed_other;
        }

        printf("Other times total: %f ms\nOther times average: %f ms\n",
               elapsed_other_full,
               elapsed_other_full / 100);`
               
I was switching between OpTensor and OpTensor2 to run tests for old and new structure, also only test case for one kernel was running each time.
I will run more tests and try to figure out why host times for some kernels are slower than for the others.

src/solver/tensorOp/Op5dTensorGeneric.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+std::size_t Op5dTensorGeneric::GetWorkspaceSize(
+    [[maybe_unused]] const ExecutionContext& context,
+    [[maybe_unused]] const miopen::tensorOp::ProblemDescription& problem) const
+{
+    return 0;
+}
+
+ConvSolution
+Op5dTensorGeneric::GetSolution(const ExecutionContext& context,
```
CAHEK7:
```suggestion
Op5dTensorGeneric::GetSolution(const ExecutionContext&,
```
hip-tidy complains

src/tensorOp/problem_description.cpp
```diff
@@ -0,0 +1,73 @@
...
+NetworkConfig ProblemDescription::MakeNetworkConfig() const
+{
+    std::ostringstream ss;
+
+    auto alens = aTensorDesc.GetLengths();
+    auto blens = bTensorDesc.GetLengths();
+
+    auto astrides = aTensorDesc.GetStrides();
+    auto bstrides = bTensorDesc.GetStrides();
+    auto cstrides = cTensorDesc.GetStrides();
```
CAHEK7:
```suggestion
    const auto& alens = aTensorDesc.GetLengths();
    const auto& blens = bTensorDesc.GetLengths();

    const auto& astrides = aTensorDesc.GetStrides();
    const auto& bstrides = bTensorDesc.GetStrides();
    const auto& cstrides = cTensorDesc.GetStrides();
```

src/include/miopen/tensorOp/problem_description.hpp
```diff
@@ -0,0 +1,132 @@
...
+            MIOPEN_THROW("A and C Tensors do not match");
+        }
+
+        if(bTensorDesc.GetType() != cTensorDesc.GetType())
+        {
+            MIOPEN_THROW("Datatypes for B and C tensors do not match !");
+        }
+
+        auto blens = bTensorDesc.GetLengths();
+        auto clens = cTensorDesc.GetLengths();
```
CAHEK7:
```suggestion
        const auto& blens = bTensorDesc.GetLengths();
        const auto& clens = cTensorDesc.GetLengths();
```

src/include/miopen/tensorOp/problem_description.hpp
```diff
@@ -0,0 +1,132 @@
...
+    NetworkConfig MakeNetworkConfig() const override;
+
+private:
+    const miopenTensorOp_t tensorOp;
+
+    float beta;
+
+    const TensorDescriptor& aTensorDesc;
+    const TensorDescriptor& bTensorDesc;
+    const TensorDescriptor& cTensorDesc;
```
CAHEK7: We should keep tensor descriptors alive
```suggestion
    TensorDescriptor aTensorDesc;
    TensorDescriptor bTensorDesc;
    TensorDescriptor cTensorDesc;
```

src/include/miopen/tensorOp/invoke_params.hpp
```diff
@@ -0,0 +1,97 @@
...
+    ConstData_t BTensor;
+
+    TensorDescriptor cTensorDesc;
+    Data_t CTensor;
+
+    size_t Aoffset;
+    size_t Boffset;
+    size_t Coffset;
+
+    bool nonStandardSquash;
```
CAHEK7: You don't need all the deleted parameters on runtime.
```suggestion
    ConstData_t ATensor;
    ConstData_t BTensor;
    Data_t CTensor;

    size_t Aoffset;
    size_t Boffset;
    size_t Coffset;
```

`nonStandardSquash` is not required at all.

Those parameters can be captured from `GetSolution`, so we won't copy tensor descriptors back and forth.
```C++
                const auto& blens = params.bTensorDesc.GetLengths();
                const auto& clens = params.cTensorDesc.GetLengths();

                const auto& astrides = params.aTensorDesc.GetStrides();
                const auto& bstrides = params.bTensorDesc.GetStrides();
                const auto& cstrides = params.cTensorDesc.GetStrides();
```

src/include/miopen/tensor/problem_description.hpp
DrizztDoUrden: my bad, didn't notice 1 letter difference
you can probably do something like
```
constexpr auto comparator = [](auto&& c, auto&& b){ return b == 1 || b == c; };
auto const diff = std::mismatch(std::begin(clens), std::end(clens), std::begin(blens), comparator);
if (diff.first != std::end(clens))
    MIOPEN_THROW(...);
```

src/include/miopen/tensor/invoke_params.hpp
CAHEK7: I think that the main point is the pointer is `void *` and actual type is an `miopen_Type_t` enum. That's why you can't just add them without special helpers.

novakovicdj:
I did some profiling to compare old and new structure and saw that creation of network_config is slower than before, this is more visible for bigger dimension tensors and it is consequence of the new format of network_config. Network_config creation for 5d tensors is around 4 times slower than in the old structure and around 3 times slower compared to 1d tensor network_config in the new structure. Because of all of that I switched to using string and got speed up of around 2.2 times compared to using stream.

After that I run 500 iterations of old and new structure for all tensor kernels and got the result that the new version is faster for 0.0005ms on average, which is around 20% faster than old structure.

CAHEK7 (approved):

src/kernels/MIOpenTensorKernels.cl
```diff
@@ -23,24 +23,7 @@
...
-#endif
-#endif
-#if MIOPEN_USE_FP32 == 1
-#define _FLOAT float
-#ifndef FLT_MAX
-#define MAX_VAL 3.402823466e+38F /* max value */
-#else
-#define MAX_VAL FLT_MAX
-#endif
-#endif
```
BrianHarrisonAMD, line 43: I never saw `MIOPEN_USE_FP16` or `MIOPEN_USE_FP32` getting set in tensorocl.cpp.
Was this wrong before, and we actually needed to have the fp16 extension enabled, but it was missed?

BrianHarrisonAMD:
I am not seeing a lot of testing coverage for OpTensor before the changes.
The only tests I can see that specifically run OpTensor seem to be [these](https://github.com/ROCm/MIOpen/blob/d16aed9b2651cb41981efdcc094189685403995e/test/tensor_ops.cpp), and it looks like they aren't covering all ops. 

Would it be possible to add new tests to the gtest suite to ensure correctness for the new solvers being added?

src/CMakeLists.txt
```diff
@@ -337,11 +337,22 @@ set( MIOpen_Source
     solver/softmarginloss/forward_softmarginloss.cpp
     solver/softmax/attn_softmax.cpp
     solver/softmax/softmax.cpp
+    solver/tensorOp/Op1dTensorGeneric.cpp
+    solver/tensorOp/Op2dTensorGeneric.cpp
+    solver/tensorOp/Op2dTensorLite.cpp
+    solver/tensorOp/Op2dTensorSquash.cpp
+    solver/tensorOp/Op3dTensorGeneric.cpp
+    solver/tensorOp/OpTensorFwdBias.cpp
```
randyspauldingamd: Could you alphabetize these?

src/include/miopen/rnn/solvers.hpp
```diff
@@ -171,9 +171,9 @@ class RNNForwardDataModularAlgo : RNNModuleAlgoBase
     // base API
     void PrepareWriteBuffers(const Handle& handle, const runtimeArgsFwd& runtimeArgs) const;
 
-    void PropX(const Handle& handle, const runtimeArgsFwd& runtimeArgs) const;
+    void PropX(Handle& handle, const runtimeArgsFwd& runtimeArgs) const;
```
randyspauldingamd: Removing the 'const' from some of these signatures (but not all) makes it look hackish. It also changes the contract, which may or may not be important. Is it feasible to work through and fix the issues that necessitated dropping the const?

src/solver/tensorOp/Op2dTensorGeneric.cpp
```diff
@@ -0,0 +1,185 @@
...
+    std::array<size_t, 2> bstrides;
+    std::array<size_t, 2> cstrides;
+    std::tie(astrides[0], astrides[1]) = miopen::tien<2>(aTensorDesc.GetStrides());
+    std::tie(bstrides[0], bstrides[1]) = miopen::tien<2>(bTensorDesc.GetStrides());
+    std::tie(cstrides[0], cstrides[1]) = miopen::tien<2>(cTensorDesc.GetStrides());
+
+    miopenDataType_t data_type = bTensorDesc.GetType();
+    bool fit_into_int          = aTensorDesc.AllDimsFitIntoInt();
+
+    size_t local_threads = 32;
```
randyspauldingamd, line 87: Why so small? (is this intended?)

src/solver/tensorOp/tensor_op_helpers.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+    // Special case for adding tensors in place
+    size_t global_threads =
+        (static_cast<int>(leading_ones) == 1 && (d - 1) == 3) ? num_wg : num_wg * local_threads;
+    global_threads = (global_threads < local_threads) ? local_threads : global_threads;
+
+    if(is4dLite)
+    {
+        // for naive tensor ops
+        const std::string data_type = GetDataType(bTensorDesc.GetType());
```
randyspauldingamd: Effectively unused

src/solver/tensorOp/tensor_op_helpers.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+        size_t TENS_LEN = cTensorDesc.GetElementSize();
+        size_t RD_BLCK  = (TENS_LEN % 4 == 0) ? 4 : (TENS_LEN % 2 == 0) ? 2 : 1;
+        const std::string READ_TYPE =
```
randyspauldingamd: Unused--is something missing?

randyspauldingamd (changes requested):

src/kernels/MIOpenTensorKernels.cl
novakovicdj: If you go into ```GetDataTypeKernelParams``` and then into ```GetDataTypeKBP``` you will see that it sets all of these in there but we do not use anything from here except this ```#pragma``` to support half type for OCL kernels. Based on that I do not define any of those macros (```MIOPEN_USE_*```) in solver structure and just keep this part needed for half type

src/solver/tensorOp/Op2dTensorGeneric.cpp
novakovicdj: It is intended yes, I tested new ```Op2dTensorGeneric``` kernel with different number of ```local_threads``` to improve ```B Nx1``` case and this was the best solution (there is not noticeable impact on other cases)

novakovicdj:
> I am not seeing a lot of testing coverage for OpTensor before the changes. The only tests I can see that specifically run OpTensor seem to be [these](https://github.com/ROCm/MIOpen/blob/d16aed9b2651cb41981efdcc094189685403995e/test/tensor_ops.cpp), and it looks like they aren't covering all ops.
> 
> Would it be possible to add new tests to the gtest suite to ensure correctness for the new solvers being added?

Do you want me to add new tests as part of this PR or to add them after this one is merged?

src/include/miopen/rnn/solvers.hpp
novakovicdj: Currently `ExecutePrimitive` function does not support `const Handle&` so we need to pass non-const and therefore remove const at some places. 
I tried changing it but it just pulls one issue after another. I would have needed to change too much code just for it to work. @DrizztDoUrden is more familiar with that part of the code and will try to make some changes to make it possible for `ExecutePrimitive` function to accepts `const Handle&` and then we can return to these places and return back the const

CAHEK7: > Removing the 'const' from some of these signatures (but not all) makes it look hackish. It also changes the contract, which may or may not be important. Is it feasible to work through and fix the issues that necessitated dropping the const?

Good catch, but https://github.com/ROCm/MIOpen/pull/2671#discussion_r1450533100 https://github.com/ROCm/MIOpen/pull/2671#discussion_r1450534663 https://github.com/ROCm/MIOpen/pull/2671#discussion_r1452345703 https://github.com/ROCm/MIOpen/pull/1276#discussion_r1455119631

That why sometimes it's good practice to quick check almost all the PRs to be aware about the internals of the library.
Finally I think that https://github.com/ROCm/MIOpen/pull/1276#discussion_r1455119631 is gonna be fixed soon, and I wish I've got you support in https://github.com/ROCm/MIOpen/pull/2671, probably in that case the problem got fixed few month earlier.

CAHEK7: I think it can be done in follow-up PR when this and https://github.com/ROCm/MIOpen/pull/3402 have been merged.

src/kernels/MIOpenTensorKernels.cl
BrianHarrisonAMD: Ah

BrianHarrisonAMD: I think it might be worth still only enabling that for FP16 kernels.
I don't think it's a big deal, but not sure why we would change it either.

BrianHarrisonAMD: Ah after discussing I see why this is preferrable.
I am fine with removing the defines entirely, and leaving this to make it compatible.

novakovicdj:
> I am not seeing a lot of testing coverage for OpTensor before the changes. The only tests I can see that specifically run OpTensor seem to be [these](https://github.com/ROCm/MIOpen/blob/d16aed9b2651cb41981efdcc094189685403995e/test/tensor_ops.cpp), and it looks like they aren't covering all ops.
> 
> Would it be possible to add new tests to the gtest suite to ensure correctness for the new solvers being added?

Current test for tensorOp is covering all solvers except for Op2dTensorSquash but I did some changes and tested it locally and it worked fine. There is a plan to switch this test to gtest and then those improvements of testing tensorOps will be implemented.

As a part of this PR I will add some unit tests for Problem Descriptor, so please do not merge this yet

CAHEK7 (approved):

src/solver/tensorOp/tensor_op_helpers.hpp
CAHEK7: I'm curious why either compilers or clang-tidy does not complain about it. Usually, they do..

CAHEK7:
@randyspauldingamd @BrianHarrisonAMD I guess it's a final review round.

BrianHarrisonAMD:
I think we need to coordinate these changes with #3402 to ensure const is correct after both are merged.
Or I guess we could decide to do a follow-up.

randyspauldingamd (approved):

randyspauldingamd:
> I think we need to coordinate these changes with #3402 to ensure const is correct after both are merged. Or I guess we could decide to do a follow-up.

If you were asking for feedback, I'd be fine with a follow-up (in a timely fashion). @novakovicdj, are you going to be joining our scrum anytime soon? If not, perhaps we could ask @DrizztDoUrden to add a ticket and coordinate with you.

BrianHarrisonAMD:
Yea I think it depends on what @DrizztDoUrden would like to do.
I think we either merge this first, and fix it in #3402, or merge both, and fix in a separate follow up.

CAHEK7:
> I think we need to coordinate these changes with #3402 to ensure const is correct after both are merged. Or I guess we could decide to do a follow-up.

It can be fixed later. It's quite big and require extra effort for the maintenance.

CAHEK7:
> > I think we need to coordinate these changes with #3402 to ensure const is correct after both are merged. Or I guess we could decide to do a follow-up.
> 
> If you were asking for feedback, I'd be fine with a follow-up (in a timely fashion). @novakovicdj, are you going to be joining our scrum anytime soon? If not, perhaps we could ask @DrizztDoUrden to add a ticket and coordinate with you.

Probably not in a short-term perspective, there is some bureaucracy involved.

BrianHarrisonAMD:
Greetings @novakovicdj!

Can you update this branch with develop and resolve the conflicts?

CAHEK7:
Hi @BrianHarrisonAMD @BradPepersAMD,
I'm not sure about the latest merge policies and are they applicable to Djordje, but probably we have to merge it manually.
Housekeeping that big PR can be painful.

BrianHarrisonAMD:
Ill kick off another CI run, and we can merge once it passes.

BrianHarrisonAMD:
Looks like it failed, but seems unrelated.
Restarted that stage.

junliume (approved):

DrizztDoUrden (approved):
lgtm

DrizztDoUrden (approved):
lgtm

CAHEK7:
@BrianHarrisonAMD @BradPepersAMD @junliume I hope it will be merged soon.  Or do we still have constantly failing develop?
I'm a bit confused about the fact that the reviewed and approved PR is hanging there for eternity.

src/tensor.cpp
```diff
@@ -865,6 +876,1200 @@ void from_json(const nlohmann::json& j, TensorDescriptor& descriptor)
...
+            });
+
+            break;
+        }
+        default: assert(false);
+        }
+    }
+}
+
+void OpTensor(Handle& handle,
```
shurale-nkn:
```suggestion
void OpTensor(const Handle& handle,
```

All other tensor functions works with const handle. Just like the old version of the OpTensor function.
The reason as I can see this function 

https://github.com/ROCm/MIOpen/blob/370e378a984cb9d6cd870d1b8fd7a346d4325a08/src/include/miopen/find_solution.hpp#L470-L477

Which mistakenly non-const , but that's not necessary.

https://github.com/ROCm/MIOpen/blob/370e378a984cb9d6cd870d1b8fd7a346d4325a08/src/include/miopen/find_solution.hpp#L441-L447

shurale-nkn: And this should be resolved in https://github.com/ROCm/MIOpen/pull/3402.
which automatically resolves the merge conflict.

BrianHarrisonAMD:
Still seems to not be passing, but the test failures look unrelated.

@BradPepersAMD  thoughts on merging this?

randyspauldingamd:
> Still seems to not be passing, but the test failures look unrelated.
> 
> @BradPepersAMD thoughts on merging this?

There are big time bombs associated with leaving the const correctness unfinished. I'd say at a bare minimum we need to see an epic and scoped ticket(s) prior to merging to be sure that's handled in a timely manner.

shurale-nkn (changes requested):

BrianHarrisonAMD:
> > Still seems to not be passing, but the test failures look unrelated.
> > @BradPepersAMD thoughts on merging this?
> 
> There are big time bombs associated with leaving the const correctness unfinished. I'd say at a bare minimum we need to see an epic and scoped ticket(s) prior to merging to be sure that's handled in a timely manner.

I think originally the hope was to merge this before #3402 was merged, and update it there, but with the delays in merging this, I think we need to fix it with this PR now. There is also merge conflicts that will need to be resolved since we merged #3402. 

@novakovicdj when you get a chance, can you resolve, and we will start another round of CI.

novakovicdj:
@BrianHarrisonAMD @BradPepersAMD Do you know if these tensorOps should support `double` data type, @CAHEK7 asked me to skip them in `isApplicable` function but now after this PR https://github.com/ROCm/MIOpen/pull/3424 there is whole set of tests with double type that are failing?

BrianHarrisonAMD:
Restarted CI again.
I think the node issues are resolved now.

PR #3351, opened 2024-11-01
Add bfloat16 support to OCL bn kernels

BrianHarrisonAMD:
Draft of updating OCL BN to support bf16


src/kernels/MIOpenBatchNormFwdTrainSpatial.cl
```diff
@@ -145,8 +145,8 @@ MIOpenBatchNormFwdTrainSpatial(const __global _FLOAT* __restrict in,
         variance = 0;
     }
     invVariance = rsqrt(variance + (_FLOAT_ACCUM)epsilon);
-    pvscale     = (_FLOAT_ACCUM)lcl_scale;
-    pvbias      = (_FLOAT_ACCUM)lcl_bias;
+    pvscale     = FLOAT2ACCUM(lcl_scale);
+    pvbias      = FLOAT2ACCUM(lcl_bias);
```
bpepers-me: this seems off.  lcl_scale and lcl_bias are actually _FLOAT_PREC, not FLOAT.  so technically a FLOATPREC2ACCUM?

BrianHarrisonAMD: Oh in that case we don't need the change.

Nice catch!

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -815,8 +815,8 @@ MIOpenBatchNormBwdSpatialDScaleDBias(const __global _FLOAT* x_in,
     {
         unsigned int betaindex  = cidx + ygrp_sz * ygrp_id + 6;
         unsigned int gammaindex = cidx + ygrp_sz * ygrp_id + 4;
-        buff[gammaindex]        = (_FLOAT_PREC)dscale;
-        buff[betaindex]         = (_FLOAT_PREC)dbias;
+        buff[gammaindex]        = FLOAT2FLOATPREC(dscale);
+        buff[betaindex]         = FLOAT2FLOATPREC(dbias);
```
bpepers-me, line 819: this is odd since buff is _FLOAT*.  but then a bunch of the functions around here don't seem right  like the ones that only use _FLOAT so they better never be used for bf16.

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -860,8 +860,8 @@ MIOpenBatchNormBwdSpatialFinalDScaleDBias(__global _FLOAT* buff,
 
     if(ygid == 0)
     {
-        delta_scale[xgid] = (_FLOAT_PREC)ds;
-        delta_bias[xgid]  = (_FLOAT_PREC)db;
+        delta_scale[xgid] = FLOAT2FLOATPREC(ds);
+        delta_bias[xgid]  = FLOAT2FLOATPREC(db);
```
bpepers-me, line 864: same here

src/kernels/MIOpenBatchNormActivFwdTrainSpatial.cl
```diff
@@ -109,21 +109,21 @@ MIOpenBatchNormActivFwdTrainSpatial(float INHW,
 
     if(lid < MIO_BN_SEGMENT)
     {
-#if MIOPEN_USE_FP16 == 0
+#if (MIOPEN_USE_FP16 == 0) || (MIOPEN_USE_BF16 == 1)
```
BrianHarrisonAMD: This define isn't for mixed, and we shouldn't do this for BF16 since it might cause problems.

BrianHarrisonAMD: Oh it's fine actually since the define doesn't exit, but ill remove it.

src/kernels/MIOpenBatchNormBwdPerAct.cl
```diff
@@ -77,8 +77,8 @@ __kernel void MIOpenBatchNormBwdPerActivationSaved(const __global _FLOAT* in,
         {
             // per (x-dims) channel load a block of data into LDS
             index  = in_nstride * n + adjIndex;
-            xhat   = (_FLOAT_PREC)(in[index] - mean) * invVar;
-            dyelem = (_FLOAT_PREC)(dy_in[index]);
+            xhat   = (FLOAT2FLOATPREC(in[index]) - mean) * invVar;
```
BrianHarrisonAMD, line 80: This is different now, but I think it was wrong before, and correct now.

src/kernels/MIOpenBatchNormActivFwdTrainSpatial.cl
```diff
@@ -109,21 +109,21 @@ MIOpenBatchNormActivFwdTrainSpatial(float INHW,
 
     if(lid < MIO_BN_SEGMENT)
     {
-#if MIOPEN_USE_FP16 == 0
+#if (MIOPEN_USE_FP16 == 0))
```
CAHEK7: extra bracket

src/kernels/batchnorm_functions.h
```diff
@@ -54,6 +59,31 @@
...
+#define _FLOAT ushort
+#ifdef _FLOAT_PREC
+#undef _FLOAT_PREC
+#endif
+#define _FLOAT_PREC float
+
+#ifdef EPSILON
+#undef EPSILON
+#endif
+#define EPSILON (_FLOAT)0.00781250
```
CAHEK7: `EPSILON` expands to `(ushort)0.00781250` which means 0

src/kernels/batchnorm_functions.h
```diff
@@ -54,6 +59,31 @@
 
 #endif
 
+#if MIOPEN_USE_BFPMIX == 1
+#define _FLOAT ushort
```
CAHEK7, line 63: It must not be `ushort`, `ushort` is just a container type holding 2 unsigned bytes.
Casting to that type is dangerous.

CAHEK7 (changes requested):

src/kernels/MIOpenBatchNormBwdSpatial.cl
BrianHarrisonAMD: Yea these look incorrect for BF16, and just generally, but I think these don't get called for mixed.
Only issue is it might have compilation errors that were okay before let me check how it behaved for FP16.

src/kernels/batchnorm_functions.h
BrianHarrisonAMD: I think that's what we want in this case.
Anywhere casting (and not using the macro to convert properly), or doing math in that type is wrong, and all places doing mixed (which is what we are supporting right now), will cast to FP32.

src/kernels/batchnorm_functions.h
BrianHarrisonAMD: Oh interesting.
I don't think it's actually used in this case.
Ill swap it to _FLOAT_PREC & match FP32 for now, but looks like we pass in epsilon and don't use this define.

src/kernels/MIOpenBatchNormBwdPerAct.cl
```diff
@@ -89,11 +89,11 @@ __kernel void MIOpenBatchNormBwdPerActivationSaved(const __global _FLOAT* in,
         for(int n = 0; n < N; n++)
         {
             index         = in_nstride * n + adjIndex;
-            xhat          = (_FLOAT_PREC)(in[index] - mean) * invVar;
+            xhat          = (FLOAT2FLOATPREC(in[index]) - mean) * invVar;
             tmp1          = mad(xhat, dxhathat, dxhat);
-            tmp2          = mad((_FLOAT_PREC)N, dy_in[index] * pvt_scale, -tmp1);
+            tmp2          = mad((_FLOATPREC)N, FLOAT2FLOATPREC(dy_in[index]) * pvt_scale, -tmp1);
```
BrianHarrisonAMD: This is the wrong cast.

src/kernels/batchnorm_functions.h
bpepers-me: yes ushort is what we want.  it is passed in as ushort and nothing should nakedly be using _FLOAT for any math operations or assuming it can naturally convert to _FLOAT_ACCUM, _FLOAT_PREC, or even just float/int/...  any use of an argument or variable that is _FLOAT need to use the FLOAT2ACCUM or FLOAT2FLOATPREC and this gives you back a value you can actually do math with.

junliume (approved):

bpepers-me (approved):

src/kernels/MIOpenBatchNormActivFwdTrainSpatial.cl
CAHEK7: Fixed in #3358

src/batchnorm/problem_description.cpp
```diff
@@ -36,6 +36,109 @@ namespace miopen {
...
 
+bool is_fp16_or_bfp16(miopenDataType_t type)
+{
+    return ((type == miopenHalf) || (type == miopenBFloat16));
+}
+
+bool is_fp32_or_fp64(miopenDataType_t type)
+{
+    return ((type == miopenFloat) || (type == miopenDouble));
+}
```
averinevg, line 47: Please don't use snake case for function names

src/solver/batchnorm/forward_spatial_single.cpp
```diff
@@ -40,31 +40,32 @@ namespace solver {
 namespace batchnorm {
 
 bool BnFwdTrainingSpatialSingle::IsApplicable(
-    const ExecutionContext&, const miopen::batchnorm::ProblemDescription& problem) const
+    const ExecutionContext&, const miopen::batchnorm::ProblemDescription& bn_problem) const
 {
-    if(problem.GetDirection() != miopen::batchnorm::Direction::ForwardTraining ||
-       problem.GetMode() != miopenBNSpatial)
+    if(bn_problem.GetDirection() != miopen::batchnorm::Direction::ForwardTraining ||
+       bn_problem.GetMode() != miopenBNSpatial)
```
averinevg, line 46: This renaming was not superfluous. I believe that the name `problem` in this context is sufficient. Inside _all_ solvers this structure is called exactly this way, and renaming it will only bring chaos.

src/batchnorm/problem_description.cpp
```diff
@@ -36,6 +36,109 @@ namespace miopen {
...
+             bn_problem.GetBnSMean().GetType() == miopenFloat &&
+             bn_problem.GetBnSVar().GetType() == miopenFloat) ||
+            // case 1 : fp32 or fp64
+            (is_fp32_or_fp64(bn_problem.GetXDesc().GetType()) &&
+             is_fp32_or_fp64(bn_problem.GetDXDesc().GetType()) &&
+             is_fp32_or_fp64(bn_problem.GetBnScale().GetType()) &&
+             is_fp32_or_fp64(bn_problem.GetBnBias().GetType()) &&
+             is_fp32_or_fp64(bn_problem.GetBnSMean().GetType()) &&
+             is_fp32_or_fp64(bn_problem.GetBnSVar().GetType())));
+}
```
averinevg, line 140: These functions should not be here. ProblemDescription is a universal structure, not tied to a backend, third-party library, etc. Move them to the appropriate files.

averinevg (commented):
Post-merge review

@BrianHarrisonAMD Please create post-merge fix.

CC @junliume @bpepers-me

PR #3354, opened 2024-11-02
Disable backwards weight bfloat16 tests outside gfx94X

BrianHarrisonAMD:
Disable bfloat16 backwards weight conv tests outside gfx94X


bpepers-me (approved):
Looks good!

PR #3355, opened 2024-11-02
Bg/fix bn solvers is applicable

bghimireamd:
Fix Gtest for ock and ck kernel for batch norm
fix driver cmd


src/fusion.cpp
```diff
@@ -389,15 +389,18 @@ std::string LogCmdBnormFusion(const miopenFusionPlanDescriptor_t fusePlanDesc, i
...
+        //                                 nullptr,
+        //                                 nullptr,
+        //                                 nullptr,
+        //                                 bn_op->mode,
+        //                                 nullptr,
+        //                                 nullptr,
+        //                                 nullptr,
+        //                                 nullptr,
+        //                                 miopen::debug::BatchNormDirection_t::ForwardInference,
+        //                                 false);
```
BrianHarrisonAMD: Should this be removed? 

Not really sure what makes sense here tbh.

bghimireamd: ok will remove this. It a fusion and driver command does not make sense to me. Since, fusion itself is a driver command.

PR #3359, opened 2024-11-03
fix negative cache results

BrianHarrisonAMD:
In some cases the loop would exit early and that would result in negative entries that could cause problems with caching.


cderb (approved):

iq136boy (approved):

bpepers-me (approved):
good to go!

PR #3360, opened 2024-11-03
Fix double accumulation

BrianHarrisonAMD:
Fix accumulation issue if it finishes fast enough to still average


cderb (approved):
that'll work

bpepers-me (approved):
good to go!

iq136boy (approved):

PR #3362, opened 2024-11-04
Temporarily disable igemm wrw solvers for bfloat16 unless device is gfx92

BrianHarrisonAMD:
Temporarily disable solvers that are failing on MI200 to prevent test failures.


bghimireamd (approved):

CAHEK7 (approved):

iq136boy (approved):
LGTM

averinevg:
@BrianHarrisonAMD @CAHEK7 @bghimireamd @iq136boy @junliume @bpepers-me We must wrap temporary changes using `#if` and `#endif` directives. If we don't, these changes will remain here forever. It is impossible to remember everything, and the sheet with notes can be lost, and the file with notes can be deleted, so we usually use the following macros:

- For guthub issues: `WORKAROUND_ISSUE_XXXX`
- For SWDEV tickets: `WORKAROUND_SWDEV_XXXXXX`

Look for examples in the library code. Also there must also be an issue or ticket, the number of which is specified in the macro.

PR #3363, opened 2024-11-04
Update db_sync to work with db.txt

cderb:


test/gtest/db_sync.cpp
```diff
@@ -338,6 +338,7 @@ void GetPerfDbVals(const fs::path& filename,
                    std::unordered_map<std::string, std::string>& vals,
                    std::string& select_query)
 {
+#if MIOPEN_ENABLE_SQLITE && MIOPEN_USE_SQLITE_PERFDB
```
CAHEK7, line 341: I'm just curious, can `MIOPEN_USE_SQLITE_PERFDB` imply `MIOPEN_ENABLE_SQLITE`?
I think the opposite is barely possible and just a single condition is a bit simpler.

CAHEK7 (approved):

test/gtest/db_sync.cpp
cderb: That may be the case. I'm mimicking other places in the src that have this switch. So perhaps this could be uniformly updated.

junliume (approved):

PR #3369, opened 2024-11-05
[MI300][Tuning] BF16 matching for FP16 configs

cderb:
gfx942 update tuned bf16 configs to match fp16 range


BrianHarrisonAMD (approved):
Looks good to me!

PR #3373, opened 2024-11-06
SWDEV-485839: Replace asm version of readfirstlane with its builtin version

Vsevolod1983:
SWDEV-485839: According to comments and suggestions in this ticket I implemented the possible fix.
In the log we can see that Rocm 6.3 is used for testing which is not released yet. Current version is 6.2.2.
So, the possible reason of this issue is the transition from one Rocm to another.


CAHEK7 (approved):

CAHEK7:
@junliume  @bpepers-me it blocks 6.3

junliume (approved):

PR #3376, opened 2024-11-07
[gtest] move cpu-based ctest to gtest

CAHEK7:
It rewrites small CPU-based tests from CTest to gtest (only perfdb\sqlite_perfdb are left) and enables `NONE` datatype by default for all the stages. There is no urgent need for a pure CPU stage, since all the pure CPU tests take about 4 seconds to complete.

Side not for reviewers - I'd recommend doing it commit-by-commit, since each commit is a separated test rewriting.

The biggest change is tensor_test. It's not an ideal implementation, but I think now it's a bit better.
I fixed few bugs, generalized test fixtures and reduce amount of code more than 3 times, while kept all the previous cases and even added few more.
I'm open for ideas how it can be improved more.


Vsevolod1983 (approved):

iq136boy (approved):

junliume (approved):

test/gtest/sequences.cpp
```diff
@@ -0,0 +1,207 @@
...
+ *
+ *******************************************************************************/
+#include <miopen/sequences.hpp>
+#include <gtest/gtest.h>
+
+namespace Seq = miopen::seq;
+
+TEST(CPU_sequence_NONE, SimpleListTest)
+{
+    using TestSequence = Seq::Sequence<int, 1, 2, 4, 5>;
```
randyspauldingamd, line 33: Future improvements (nothing mentioned here needs to hold up this PR):

The documentation in sequences.hpp is quite confused. It explicitly states that values must be unique, but then gives a long example of a problem that only exists with duplicate values. The former and latter may apply to different subsets of the types defined here, but no effort is made to clarify.
...Actually, looking through the rest of that file, there are many more problems than just that...too much to talk about here.

Sequence must be tested with negative values and out-of-numerical-order values. Sign and numerical order aren't important, so we need to distinguish the behavior from other possibilities such as "magnitude is greater", "return the next value greater than the current", etc.

Add tests for ValidateValues().

Add tests for find().

The current ctor is broken since it uses assert (does not check for duplicates in Release build).
Related to that, add a test that enforces throwing on duplicated values.

I've found more issues, but I need to get back to my tickets. We can discuss tomorrow.

test/gtest/sequences.cpp
```diff
@@ -0,0 +1,207 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+#include <miopen/sequences.hpp>
+#include <gtest/gtest.h>
+
+namespace Seq = miopen::seq;
+
+TEST(CPU_sequence_NONE, SimpleListTest)
```
randyspauldingamd, line 31: Suggest, improves brevity and maintainability (also applies to most other tests):
```
#define SimpleListTest_VALUES 1, 2, -1, 4, 5
TEST(CPU_sequence_NONE, SimpleListTest)
{
    Seq::Sequence<int, SimpleListTest_VALUES> test_seq;
    auto expected = std::vector<int>{SimpleListTest_VALUES};

    auto it = test_seq.begin();
    for(auto v : expected)  EXPECT_EQ(*(it++), v);
    EXPECT_EQ(it++, test_seq.end());  // may actually want explicit tests that define and enforce behavior of traveling past the ends of the sequence
}
```
We should also talk about other possibilities, for example:
- preincrement vs. postincrement
- decrement (probably not intended to function, but it does)
- other behaviors in comparison with STL containers

test/gtest/sequences.cpp
CAHEK7: Agree.
This particular PR is focused on the existing tests.

PR #3378, opened 2024-11-11
Fix the fwd convolution Nan output issue 

iq136boy:
Fix the fwd convolution Nan output issue


cderb (approved):

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -759,7 +759,8 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
             /// \todo: Will need SetTensor() to properly zero out non-packed tensors
             /// Note: Need to clear buffer memory for BWD data since all values may not be set.
             if(output_tr_inst.GetConvOperandTag() == internal::ConvOperandTag::Weights ||
-               output_tr_inst.GetConvOperandTag() == internal::ConvOperandTag::Input)
+               output_tr_inst.GetConvOperandTag() == internal::ConvOperandTag::Input ||
+               output_tr_inst.GetConvOperandTag() == internal::ConvOperandTag::Output)
```
bpepers-me: Are there other operands besides these three?  If not, can we just remove the if statement at all and just always zero out the buffer and say in a comment that it's needed in all cases?

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -926,7 +927,8 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
                 auto invoker_ptr     = sh_conv_ptr->MakeInvokerPointer();
                 HipEventProfiler pfr(handle);
 
-                // Zero out the buffer for BWDs data since it won't always write all output values.
+                // Zero out the buffer for output data since it won't always write all output
+                // values.
                 if constexpr(std::is_same_v<CastType, miopen::conv::DataInvokeParams>)
                 {
                     ZeroOutTensor(handle, data_ctx.tensors.outDesc, data_ctx.tensors.out);
```
bpepers-me, line 929: Should we also add a unit test that would result in that NaN to see that it's fixed with this change?

CAHEK7 (approved):
Is it related to https://github.com/ROCm/MIOpen/pull/3377 ?

iq136boy:
> Is it related to #3377 ?

Yes. The same issue but for fwd solver.

junliume (approved):

junliume:
Ping me when the CI pipelines have passed, and we will merge this PR asap

PR #3379, opened 2024-11-13
Update CK commit hash for staging

junliume:

PR #3392, opened 2024-11-16
Enable ck's grouped convolution fwd instances

linsun12:
This PR is for enabling ck's grouped convolution fwd instances inside MIOpen.
Changes are minor and the goal is to
1. change the output datatype to int8 for consistency between input datatype and output datatype. The int32 as output datatype for convint8 was legacy code from previous customer request and is no longer used.
2. Enable int8 datatype for naive_conv solver as reference solver for validating output
3. ~~commit hash for composable kernel is switched to ck's latest develop.~~ commit hash set by #3379 
4. enable int8 unit tests for group_conv2d_fwd


BrianHarrisonAMD:
Greetings @linsun12!

Heads up, there was a formatting error on these changes.
I fixed it to allow CI to run, but here is the script I run in case you need it in the future: 
 
```shell
#!/bin/bash
find . -iname '*.h' \
    -o -iname '*.hpp' \
    -o -iname '*.cpp' \
    -o -iname '*.h.in' \
    -o -iname '*.hpp.in' \
    -o -iname '*.cpp.in' \
    -o -iname '*.cl' \
| xargs -n 1 -P $(nproc) -I{} -t clang-format-12 -style=file {} -i 2>/dev/null
```

requirements.txt
```diff
@@ -7,5 +7,5 @@ nlohmann/json@v3.11.2 -DJSON_MultipleHeaders=ON -DJSON_BuildTests=Off
 ROCm/FunctionalPlus@v0.2.18-p0
 ROCm/eigen@3.4.0
 ROCm/frugally-deep@9683d557eb672ee2304f80f6682c51242d748a50
-ROCm/composable_kernel@4b798833c4cc614bbf7f7fb43c5ee81d12602f05 -DCMAKE_BUILD_TYPE=Release
+ROCm/composable_kernel@0c9012fb70bcd2750ff0d5b8c23e4bc6f5937709 -DCMAKE_BUILD_TYPE=Release
```
BrianHarrisonAMD: We should coordinate this with #3379.
I think Jun's update should be new enough to have the changes you need.

BrianHarrisonAMD:
I think we should also add tests to check the new instances are being selected & integrated properly.

As a starting point, it looks like there might already be some relevant tests that are commented out in [test/gtest/group_conv2d_fwd.cpp](https://github.com/ROCm/MIOpen/blob/eef83ebea6b99e7d98bc2032fbadbf2d4b6db055/test/gtest/group_conv2d_fwd.cpp#L36) .

linsun12:
> Greetings @linsun12!
> 
> Heads up, there was a formatting error on these changes. I fixed it to allow CI to run, but here is the script I run in case you need it in the future:
> 
> ```shell
> #!/bin/bash
> find . -iname '*.h' \
>     -o -iname '*.hpp' \
>     -o -iname '*.cpp' \
>     -o -iname '*.h.in' \
>     -o -iname '*.hpp.in' \
>     -o -iname '*.cpp.in' \
>     -o -iname '*.cl' \
> | grep -v -E '(build/)|(install/)' \
> | xargs -n 1 -P $(nproc) -I{} -t clang-format-12 -style=file {} -i 2>/dev/null
> ```

@BrianHarrisonAMD Thank you for the information!

requirements.txt
linsun12: Yes Jun's update is a newer hash, will drop this hash change waiting for that PR to finish first.

linsun12: commit hash is set to latest version from develop

iq136boy (approved):

BrianHarrisonAMD (approved):
LGTM!

PR #3393, opened 2024-11-17
Forbid inefficient TensorDescriptor initialization

CAHEK7:
Initializing TensorDescriptor from `std::vector<int>` is very inefficient due to extra checks and multiple intermediate vector, since internally `std::vector<size_t>` is used.

Changed all the initializations to the native `size_t`, removed constructors with `std::vector<int>` and added workarounds for a legacy descriptors initializations with `int`'s.

It increased performance for the current RNN implementation for a few percents.


src/tensor.cpp
```diff
@@ -88,28 +88,20 @@ std::optional<miopenTensorLayout_t> GetDefaultLayout(unsigned num_dims)
 }
 
 template <class T>
-bool CheckLengths(const std::vector<T>& lens, T maxval = 0)
+bool CheckLengthsValues(const std::vector<T>& lens)
 {
-    if(lens.empty())
-        return false;
```
averinevg, line 94: We should check that the vector is not empty.

src/tensor.cpp
```diff
@@ -88,28 +88,20 @@ std::optional<miopenTensorLayout_t> GetDefaultLayout(unsigned num_dims)
...
-    if(maxval)
-    {
-        if(!std::all_of(lens.cbegin(), lens.cend(), [maxval](T x) { return x <= maxval; }))
-            return false;
-    }
-    return true;
+    return std::all_of(lens.cbegin(), lens.cend(), [](T x) {
+        return x > 0 &&
+               static_cast<size_t>(x) <= static_cast<size_t>(std::numeric_limits<int64_t>::max());
```
averinevg, line 95: Is the compiler "smart" enough to throw away the second part of the condition for the int data type?

averinevg (commented):
LGTM, although there are a couple of minor comments.

src/tensor.cpp
CAHEK7: Sure. Except for -O0 case.

src/tensor.cpp
CAHEK7: No, not here, we check non-empty vector later, otherwise we do it multiple times.
This function is dedicated for values sanity check, that's it.

averinevg: I'm fine with it. If this check is present in the constructor, then everything is okay.

averinevg (approved):

CAHEK7:
Majority of tests heavily rely on `vector<int>`

randyspauldingamd (approved):
I like it. A lot. I'd like it even more if we abandoned using std::vector directly and create objects for lengths and strides and wrap a lot of the conversions and checks into them.

averinevg:
@randyspauldingamd I believe this is now your PR :smile:. Does it pass the CI?

CAHEK7:
> @randyspauldingamd I believe this is now your PR 😄. Does it pass the CI?

It doesn't.
The tests, especially old ctests, heavily rely on that `int` and that's why I've put the comment here https://github.com/ROCm/MIOpen/pull/3424#discussion_r1869914175 - we must not bring the old legacy code into the GTests and there is no excuses (@Vsevolod1983, @BrianHarrisonAMD, could you please check this for any upcoming PRs?)

Fixing old the CTests is a pain in the `everywhere`. First of all, the template code in the test driver should be adapted (hard but manageable), next - mostly all the ctests should be modified (easy but lots of changes). Good news is - I've already updated various test data configs to be compatible with both `int` and `size_t` here https://github.com/ROCm/MIOpen/pull/3416/files#diff-1daaa3cdf3d99199e44b03aea4f80477db07637a62713338f21dd1eec71e8fcd
But probably we can get rid of the old tests soon, so the problem disappeared itself.

randyspauldingamd:
I just had to open my big mouth...lol. Yeah, I can take this one. I expect I'll get a few of those ctest->gtest conversion tasks after the new year anyway :)

BradPepersAMD:

PR #3399, opened 2024-11-20
[BN] Enable NHWC in OCL

bghimireamd:
* Enable NHWC for Batch norm forward infer
* Initialize the driver and gtest with similar range of values. 
* added BN activation gtest
* Bug fixes on NCHW


src/solver/batchnorm/forward_spatial_multiple.cpp
```diff
@@ -40,9 +40,42 @@ namespace solver {
...
+    bool bfp32parm          = xDesc.GetType() == miopenFloat;
+    bool bfpmixparm = (xDesc.GetType() == miopenHalf || xDesc.GetType() == miopenBFloat16) &&
+                      problem.GetBnScale().GetType() == miopenFloat;
+
+    // NCHW is Applicable for variant = 2 only
+    if((!(n < 3) &&
+        !((in_nhw < 33554432 && in_cstride > 1024) ||
+          ((n >= 256) && (in_cstride > 60) && bfpmixparm) || ((in_cstride > 512) && bfpmixparm)) &&
+        !(in_cstride <= 512)) ||
+       !((n > 768) && (in_cstride > 150) && bfp32parm))
```
CAHEK7: It's barely readable and probably redundant condition.
May I ask you to do the math and simplify it? (at least replace `!(n < 3)` with `(n >= 3)`, but there are more simplifications possible)

src/solver/batchnorm/backward_spatial_multiple.cpp
```diff
@@ -38,9 +38,34 @@ namespace solver {
...
+bool BNBwdIsCaseVariant2(const miopen::batchnorm::ProblemDescription& problem)
+{
+    int n, c, h, w;
+    std::tie(n, c, h, w) = tien<4>(problem.GetXDesc().GetLengths());
+
+    unsigned int in_cstride = h * w;
+    unsigned int in_nhw     = n * in_cstride;
+
+    if(!(in_nhw < (32 * 1024 * 1024) && in_cstride > 1024) &&
+       !(in_nhw < (32 * 1024 * 1024) && in_cstride > 512) && !(in_cstride <= 512))
```
CAHEK7: It's not that complex as the condition from src/solver/batchnorm/forward_spatial_multiple.cpp, but still can be simplified, since it contains redundant statements.

src/solver/batchnorm/backward_spatial_multiple.cpp
```diff
@@ -38,9 +38,34 @@ namespace solver {
 
 namespace batchnorm {
 
+bool BNBwdIsCaseVariant2(const miopen::batchnorm::ProblemDescription& problem)
+{
+    int n, c, h, w;
```
CAHEK7:
```suggestion
    size_t n, c, h, w;
```

src/solver/batchnorm/backward_spatial_multiple.cpp
```diff
@@ -38,9 +38,34 @@ namespace solver {
...
+    std::tie(n, c, h, w) = tien<4>(problem.GetXDesc().GetLengths());
+
+    unsigned int in_cstride = h * w;
+    unsigned int in_nhw     = n * in_cstride;
```
CAHEK7:
```suggestion
    size_t in_cstride = h * w;
    size_t in_nhw     = n * in_cstride;
```

src/solver/batchnorm/backward_spatial_multiple.cpp
```diff
@@ -38,9 +38,34 @@ namespace solver {
 
 namespace batchnorm {
 
+bool BNBwdIsCaseVariant2(const miopen::batchnorm::ProblemDescription& problem)
+{
+    int n, c, h, w;
+    std::tie(n, c, h, w) = tien<4>(problem.GetXDesc().GetLengths());
+
+    unsigned int in_cstride = h * w;
```
CAHEK7: But actually you can avoid all the prevois `int`-related comment by the following code:
```suggestion
   auto [n, c, h, w] = tien<4>(problem.GetXDesc().GetLengths());
   auto in_cstride = problem.GetXDesc().GetStrides()[1];
```

src/solver/batchnorm/forward_spatial_multiple.cpp
```diff
@@ -40,9 +40,42 @@ namespace solver {
 
 namespace batchnorm {
 
+bool BNFwdTrainIsCaseVariant2(const miopen::batchnorm::ProblemDescription& problem)
+{
+    const auto& xDesc = problem.GetXDesc();
+    int n, c, h, w;
+    std::tie(n, c, h, w)    = tien<4>(xDesc.GetLengths());
+    unsigned int in_cstride = h * w;
+    unsigned int in_nhw     = n * in_cstride;
```
CAHEK7: See the comments from src/solver/batchnorm/backward_spatial_multiple.cpp

CAHEK7 (changes requested):

src/solver/batchnorm/backward_spatial_multiple.cpp
```diff
@@ -38,9 +38,34 @@ namespace solver {
...
+    unsigned int in_nhw     = n * in_cstride;
+
+    if(!(in_nhw < (32 * 1024 * 1024) && in_cstride > 1024) &&
+       !(in_nhw < (32 * 1024 * 1024) && in_cstride > 512) && !(in_cstride <= 512))
+    {
+        return true;
+    }
+    else
+        return false;
+}
```
CAHEK7:
```suggestion
    return !(in_nhw < (32 * 1024 * 1024) && in_cstride > 1024) &&
       !(in_nhw < (32 * 1024 * 1024) && in_cstride > 512) && !(in_cstride <= 512);
```

src/solver/batchnorm/backward_spatial_multiple.cpp
CAHEK7: One example of the transformations:
```
X == in_nhw < (32 * 1024 * 1024)
Y == in_cstride > 1024
Z == in_cstride > 512

!(X & Y) & !(X & Z) & !(!Z)
(!X | !Y) & (!X | !Z) & Z
(!X | !Y) & (!X & Z | !Z & Z) // !Z & Z -> false
(!X | !Y) & !X & Z
!X & Z | !Y & !X & Z
!X & Z & (1 | !Y) // (1 | !Y) -> true
!X & Z
```
So it basically means  `return (in_nhw >= (32 * 1024 * 1024)) && (in_cstride > 512);`
Which probably can be simplified more, since `in_nhw` is `n * in_cstride` and we know for sure that `in_cstride` must be greater than 512.

If there are any doubts about those transformations, here is a proof (tested in excel, lol):
Using that fact that if Y is true, then Z must always be true we can even exclude few cases:
|X|Y|Z|old|new|result|
|-|-|-|-|-|-|
|0|0|0|FALSE|FALSE|TRUE|
|1|0|0|FALSE|FALSE|TRUE|
|0|0|1|TRUE|TRUE|TRUE|
|1|0|1|FALSE|FALSE|TRUE|
|0|1|1|TRUE|TRUE|TRUE|
|1|1|1|FALSE|FALSE|TRUE|

driver/bn_driver.hpp
```diff
@@ -195,10 +195,10 @@ int BatchNormDriver<TInput, Tref, TAcc, TScaleBias, TOut>::GetandSetData()
...
 
     in.AllocOnHost(tensor<TInput>{bn_layout, in_len});
-    for(size_t i = 0; i < in.GetVector().size(); i++)
-    {
-        in.GetVector()[i] = prng::gen_canonical<TInput>();
-    }
+    // 0.0 to 2.0 (since unsigned)
+    in.GetTensor().generate([](auto...) {
+        return prng::gen_descreet_unsigned<TInput>(2e-3 /*scale*/, 1000 /*range*/);
+    });
```
BrianHarrisonAMD: Why unsigned?

driver/bn_driver.hpp
```diff
@@ -208,20 +208,25 @@ int BatchNormDriver<TInput, Tref, TAcc, TScaleBias, TOut>::GetandSetData()
...
-            bias.GetVector()[i]  = prng::gen_canonical<TInput>();
-        }
+        // -2.0 to 2.0
+        scale.GetTensor().generate([](auto...) {
+            return prng::gen_descreet_uniform_sign<TScaleBias>(2e-3 /*scale*/, 1000 /*range*/);
+        });
+        bias.GetTensor().generate([](auto...) {
+            return prng::gen_descreet_uniform_sign<TScaleBias>(2e-3 /*scale*/, 1000 /*range*/);
+        });
     }
```
BrianHarrisonAMD: We re-use the same lambda in quite a few places.
Could we declare it once, and re-use it?

driver/bn_driver.hpp
```diff
@@ -208,20 +208,25 @@ int BatchNormDriver<TInput, Tref, TAcc, TScaleBias, TOut>::GetandSetData()
...
     {
         estMean.AllocOnHost(tensor<TAcc>{bn_layout, derivedBnDesc.GetLengths()});
         estVariance.AllocOnHost(tensor<TAcc>{bn_layout, derivedBnDesc.GetLengths()});
 
-        auto gen_value_emean = [](auto...) { return prng::gen_descreet_unsigned<TAcc>(1e-2, 100); };
-        estMean.InitHostData(estMean.GetTensor().desc.GetElementSize(), true, gen_value_emean);
+        // 0.0 to 1.0
+        estMean.InitHostData(estMean.GetTensor().desc.GetElementSize(), true, [](auto...) {
+            return prng::gen_descreet_uniform_sign<TAcc>(2e-3 /*scale*/, 1000 /*range*/);
+        });
```
BrianHarrisonAMD: Isn't this also -2 to 2?

driver/bn_driver.hpp
bghimireamd: ok will do

src/batch_norm.cpp
```diff
@@ -109,14 +104,11 @@ void profileSequence(const Handle& handle, unsigned char select, float* ctime)
     case 0:
         if(handle.IsProfilingEnabled())
         {
-            *ctime = 0.;
-            handle.ResetKernelTime();
             ktime  = handle.GetKernelTime();
             *ctime = ktime;
```
BrianHarrisonAMD, line 108: Huh, this would be always 0 before since it reset before getting the kernel time right?

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -638,93 +638,122 @@ MIOpenBatchNormBwdSpatial(const __global _FLOAT* __restrict x_in,
...
+                   ygrp_sz);
 #else
-    commitID = 64;
     local _FLOAT_ACCUM lcl_data_x[MIO_BN_LDSGCN_SIZE];
     local _FLOAT_ACCUM lcl_data_y[MIO_BN_LDSGCN_SIZE];
-    gcn_reduce2(&mean, &variance, FLOAT2ACCUM(INHW), lcl_data_x, lcl_data_y, lid);
+    gcn_reduce2(&mean, &variance, INHW, lcl_data_x, lcl_data_y, ylid);
 #endif
 
-    barrier(CLK_LOCAL_MEM_FENCE | CLK_GLOBAL_MEM_FENCE);
```
BrianHarrisonAMD, line 680: Just to confirm, we believe this isn't needed since the reduce already handles synchronization correctly right?
I think that is the case, but also always a bit worried seeing a barrier removed.

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -638,93 +638,122 @@ MIOpenBatchNormBwdSpatial(const __global _FLOAT* __restrict x_in,
...
+                   lcl_data_x + xlid * ygrp_sz,
+                   lcl_data_y + xlid * ygrp_sz,
+                   ylid,
+                   ygrp_sz);
 #else
-    commitID = 64;
     local _FLOAT_ACCUM lcl_data_x[MIO_BN_LDSGCN_SIZE];
     local _FLOAT_ACCUM lcl_data_y[MIO_BN_LDSGCN_SIZE];
-    gcn_reduce2(&mean, &variance, FLOAT2ACCUM(INHW), lcl_data_x, lcl_data_y, lid);
+    gcn_reduce2(&mean, &variance, INHW, lcl_data_x, lcl_data_y, ylid);
```
BrianHarrisonAMD, line 684: Does this not need to have the same update as the lds_reduce2_2d above with the change to data pointers etc?

src/kernels/batchnorm_functions.h
```diff
@@ -208,6 +208,62 @@
...
+                         unsigned int ystride,
+                         unsigned int xgrp_sz,
+                         unsigned int xgrp_id,
+                         unsigned int xlid,
+                         unsigned int xstride)
+{
+#if MIOPEN_USE_FPMIX || MIOPEN_USE_BFPMIX
+#if MIO_LAYOUT_NHWC
+    unsigned int index = (ygroupoffset + vindex * 2 + xlid % 2) * ystride +
+                         (xgrp_sz * xgrp_id + xlid / 2 * 2) * xstride;
```
BrianHarrisonAMD: ``` xlid / 2 * 2 ``` 
is the intent here to offset by evens?

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -638,93 +638,122 @@ MIOpenBatchNormBwdSpatial(const __global _FLOAT* __restrict x_in,
...
+            mean, meanvarbuff, 2, ygrp_sz * yoffset, ystride, xgrp_sz, xgrp_id, xlid, xstride);
+        storeToStash(invVariance,
+                     meanvarbuff,
+                     3,
+                     ygrp_sz * yoffset,
+                     ystride,
+                     xgrp_sz,
+                     xgrp_id,
+                     xlid,
+                     xstride);
```
BrianHarrisonAMD: Why is vindex 2 & 3? 
I was expecting 0 & 1.

driver/bn_driver.hpp
bghimireamd: good point. will switch to `gen_descreet_signed`

src/kernels/MIOpenBatchNormBwdSpatial.cl
bghimireamd: during storing 0 and 1 is for `MIOpenBatchNormBwdSpatialMeanVariance` and 2 & 3 is for `MIOpenBatchNormBwdSpatialFinalMeanVariance`

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -733,205 +762,228 @@ MIOpenBatchNormBwdSpatialMeanVariance(const __global _FLOAT* __restrict in,
...
+                   lcl_data_y + xlid * ygrp_sz,
+                   ylid,
+                   ygrp_sz);
 #else
-    local _FLOAT_ACCUM lcl_data_x2[MIO_BN_LDSGCN_SIZE];
-    local _FLOAT_ACCUM lcl_data_y2[MIO_BN_LDSGCN_SIZE];
+    local _FLOAT_ACCUM lcl_data_x[MIO_BN_LDSGCN_SIZE];
+    local _FLOAT_ACCUM lcl_data_y[MIO_BN_LDSGCN_SIZE];
     gcn_reduce2(&dscale, &dbias, (_FLOAT_ACCUM)1.0, lcl_data_x2, lcl_data_y2, ylid);
 #endif
```
BrianHarrisonAMD: I think these need similar changes as the lds_reduce2_2d branch above, and also it's still using lcl_data_x2 & lcl_data_y2, but it was renamed to lcl_data_x & lcl_data_y.

BrianHarrisonAMD: We need to test with MIOPEN_USE_AMDGCN true, since it'll still be used on certain devices.

src/kernels/MIOpenBatchNormBwdSpatial.cl
BrianHarrisonAMD: Oh looks like it goes up to 5 actually.
Don't we need to increase our index math to account for the fact that we are jumping by 5 elements each time, and to ensure it doesn't overlap?

I guess right now it might be that it's storing these values in the first 12 positions (or 6 if it was FP32) of the meanvarbuff offset by the ygrp_sz, and the assumption is that there will always been enough distance to store that much?

bghimireamd: There should be always space for variant 2 since N* C * H * W value will be large enough to hold at least 12 elements.

BrianHarrisonAMD: Also, looks like we are duplicating the mean & variance for all groups in a loop, but shouldn't it be the same as before where it only writes it once for each group when the commitID == lid? 

Seems like we are duplicating effort here, but I might be missing something that explains why we need this change.

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -638,93 +638,122 @@ MIOpenBatchNormBwdSpatial(const __global _FLOAT* __restrict x_in,
...
+    unsigned int ystride = MIO_LAYOUT_NHWC ? MIO_BN_C : 1;
+
+    if(xgid >= MIO_BN_C)
+        return;
+
+    _FLOAT_PREC variance = (_FLOAT_PREC)0.;
+    _FLOAT_PREC mean     = (_FLOAT_PREC)0.;
+    _FLOAT_PREC invVariance;
+
+    for(unsigned int yoffset = ylid; yoffset < MIO_BN_NGRPS; yoffset += ygrp_sz)
```
BrianHarrisonAMD, line 662: This seems off to me.
Before we were lopping over every group, and then calculating an offset & index for that group.

Now we are incrementing yoffset by the y group size, but then comparing it against the # of groups for terminating the loop.
Is that correct?

src/kernels/MIOpenBatchNormBwdSpatial.cl
BrianHarrisonAMD: Looking at these defines, I think we need to track if this path gets used. 
It's actually dependent on quite a few defines, and not sure what conditions would use the gcn path.

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -638,93 +638,122 @@ MIOpenBatchNormBwdSpatial(const __global _FLOAT* __restrict x_in,
...
+    unsigned int xgrp_id = get_group_id(0);
+    unsigned int xgid    = get_global_id(0);
+    unsigned int xgrp_sz = get_local_size(0);
+    unsigned int ygrp_sz = get_local_size(1);
+
+    unsigned int xstride = MIO_LAYOUT_NHWC ? 1 : MIO_BN_HW;
+    unsigned int ystride = MIO_LAYOUT_NHWC ? MIO_BN_C : 1;
+
+    if(xgid >= MIO_BN_C)
+        return;
```
BrianHarrisonAMD, line 656: Since this is variant 2, is the assumption that C will always be >1?

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -638,93 +638,122 @@ MIOpenBatchNormBwdSpatial(const __global _FLOAT* __restrict x_in,
...
+#if !MIOPEN_USE_AMDGCN || MIO_BN_GRP1 > 1
+    local _FLOAT_ACCUM lcl_data_x[MIO_BN_LDS_SIZE];
+    local _FLOAT_ACCUM lcl_data_y[MIO_BN_LDS_SIZE];
+    lds_reduce2_2d(&mean,
+                   &variance,
+                   (_FLOAT_ACCUM)1.0,
+                   lcl_data_x + xlid * ygrp_sz,
+                   lcl_data_y + xlid * ygrp_sz,
+                   ylid,
+                   ygrp_sz);
```
BrianHarrisonAMD: Similar to above, I think gcn needs the same type of updates, and / or we need to review if gcn is a branch that can be hit. 

If we can't use the gcn path, then I would prefer removing it, and only having the LDS path which is tested and updated.
If we can use the gcn path, then I think we need to update it with similar changes right?

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -638,93 +638,122 @@ MIOpenBatchNormBwdSpatial(const __global _FLOAT* __restrict x_in,
...
         }
     }
 
-#if !MIOPEN_USE_AMDGCN
-    local _FLOAT_ACCUM lcl_data_x[MIO_BN_NGRPS];
-    local _FLOAT_ACCUM lcl_data_y[MIO_BN_NGRPS];
-    lds_reduce2(&mean, &variance, (_FLOAT_ACCUM)1.0, lcl_data_x, lcl_data_y, ylid);
+#if !MIOPEN_USE_AMDGCN || MIO_BN_GRP1 > 1
+    local _FLOAT_ACCUM lcl_data_x[MIO_BN_LDS_SIZE];
+    local _FLOAT_ACCUM lcl_data_y[MIO_BN_LDS_SIZE];
```
BrianHarrisonAMD: Why do we need to change the lcl data from # of groups to the LDS size?

BrianHarrisonAMD: Actually, was this wrong before?

src/kernels/MIOpenBatchNormBwdSpatial.cl
```diff
@@ -733,205 +762,228 @@ MIOpenBatchNormBwdSpatialMeanVariance(const __global _FLOAT* __restrict in,
...
 #endif
 
-    if(ygid == 0)
+    if(ylid == 0)
     {
-        delta_scale[xgid] = FLOAT2FLOATPREC(ds);
-        delta_bias[xgid]  = FLOAT2FLOATPREC(db);
+        delta_scale[xgid] = dscale;
+        delta_bias[xgid]  = dbias;
     }
```
BrianHarrisonAMD, line 910: Why did this change to the localId instead of the globalId for writing out the values?
Seems like it should be the localId to me, but a bit surprised to see this changed.

src/kernels/MIOpenBatchNormFwdTrainSpatial.cl
```diff
@@ -516,40 +524,47 @@ MIOpenBatchNormFwdTrainSpatialFinalMeanVariance(
...
+    unsigned int ygrp_sz    = get_local_size(1);
+    unsigned int xgrp_id    = get_group_id(0);
+    unsigned int xstride    = MIO_LAYOUT_NHWC ? 1 : MIO_BN_HW;
+    unsigned int ystride    = MIO_LAYOUT_NHWC ? MIO_BN_C : 1;
+    unsigned int commitID   = 0;
+
+    if(xgid >= MIO_BN_C)
+        return;
+
+    for(unsigned int yoffset = ylid; yoffset < MIO_BN_NGRPS; yoffset += ygrp_sz)
```
BrianHarrisonAMD, line 544: Here is another spot I think we need to check the loop conditions.

src/kernels/MIOpenBatchNormFwdTrainSpatial.cl
```diff
@@ -559,14 +574,23 @@ MIOpenBatchNormFwdTrainSpatialFinalMeanVariance(
...
+        storeToStash(invVariance,
+                     meanvarbuff,
+                     3,
+                     ygrp_sz * yoffset,
+                     ystride,
+                     xgrp_sz,
+                     xgrp_id,
+                     xlid,
+                     xstride);
     }
```
BrianHarrisonAMD: Also similar to above.
Is the loop condition correct, and why do we have to store the mean & variance for each group now?

src/kernels/MIOpenBatchNormFwdTrainSpatial.cl
```diff
@@ -559,14 +574,23 @@ MIOpenBatchNormFwdTrainSpatialFinalMeanVariance(
...
 
     // Save mean and calculate and save running mean
-    unsigned int ygid = get_global_id(1);
```
BrianHarrisonAMD, line 569: Below there is code saving results assuming xgid == to channel.
Is that a correct assumption for both layouts? 

The x stride makes me think it is, but just want to double check.

src/kernels/batchnorm_functions.h
```diff
@@ -208,6 +208,62 @@
...
+                                 unsigned int ystride,
+                                 unsigned int xgrp_sz,
+                                 unsigned int xgrp_id,
+                                 unsigned int xlid,
+                                 unsigned int xstride)
+{
+#if MIOPEN_USE_FPMIX || MIOPEN_USE_BFPMIX
+    // 2 _FLOAT values are used to store 1 _FLOAT_PREC value.
+#if MIO_LAYOUT_NHWC
+    // xgrp_sz values are stpit in two parts: even threads use 2 values at even rows, odd threads -
```
BrianHarrisonAMD: typo on the comment stpit -> split

src/kernels/batchnorm_functions.h
```diff
@@ -208,6 +208,62 @@
...
+#if(MIO_BN_VARIANT == 2)
+inline _FLOAT_PREC loadFromStash(const __global _FLOAT* stash,
+                                 unsigned int vindex,
+                                 unsigned int ygroupoffset,
+                                 unsigned int ystride,
+                                 unsigned int xgrp_sz,
+                                 unsigned int xgrp_id,
+                                 unsigned int xlid,
+                                 unsigned int xstride)
+{
```
BrianHarrisonAMD: I think we could unify the load and store index calculation into a single method that is shared.

src/kernels/reduction_functions.h
```diff
@@ -59,6 +59,30 @@ regLDSreduce(_FLOAT_ACCUM* value, local _FLOAT_ACCUM* data, uint localID, _FLOAT
...
+                                  _FLOAT_ACCUM scale,
+                                  local _FLOAT_ACCUM* lcl_data_x,
+                                  local _FLOAT_ACCUM* lcl_data_y,
+                                  unsigned int lid,
+                                  unsigned int size)
+{
+    lcl_data_x[lid] = (_FLOAT_ACCUM)*x;
+    lcl_data_y[lid] = (_FLOAT_ACCUM)*y;
+    barrier(CLK_LOCAL_MEM_FENCE);
+    for(unsigned int red = (size >> 1); red > 0; red >>= 1)
```
BrianHarrisonAMD, line 73: Everywhere this is called size == ygrp_sz, and the lcl_data_x / lcl_data_y is initialized to be MIO_BN_LDS_SIZE.

Is the idea that MIO_BN_LDS_SIZE is at least as big as ygrp_sz, and we need this since there is now cases where ygrp_sz is smaller than MIO_BN_LDS_SIZE?

src/solver/batchnorm/backward_spatial_multiple.cpp
```diff
@@ -51,17 +51,50 @@ bool BnBwdTrainingSpatialMultiple::IsApplicable(
...
+    {
+        if(!((in_nhw >= static_cast<size_t>(32 * 1024 * 1024) || in_cstride <= 1024) &&
+             in_cstride > 512))
+            return false;
+
+        unsigned int ylocalsize = 1024;
+        unsigned int last_ylocalsize =
+            in_cstride % ylocalsize == 0 ? ylocalsize : in_cstride % ylocalsize;
+        if(last_ylocalsize < stash_values * (problem.GetXDesc().GetType() == miopenFloat ? 1 : 2))
+            return false;
```
BrianHarrisonAMD, line 94: Just a nitpick, but it looks like we are checking size compared to stash_values in two different ways here, and it would be nice to check the same way between branches.

src/kernels/MIOpenBatchNormFwdTrainSpatial.cl
```diff
@@ -584,18 +608,24 @@ MIOpenBatchNormFwdTrainSpatialMeanVariance(const __global _FLOAT* __restrict in,
...
+    unsigned int xlid    = get_local_id(0);
+    unsigned int ylid    = get_local_id(1);
+    unsigned int xgrp_id = get_group_id(0);
+    unsigned int ygrp_id = get_group_id(1);
+    unsigned int xgrp_sz = get_local_size(0);
     unsigned int ygrp_sz = get_local_size(1);
     unsigned int index;
-    unsigned int cidx      = xgid * MIO_BN_HW;
-    unsigned int meanindex = cidx + ygrp_sz * ygrp_id;
-    unsigned int varindex  = meanindex + 2;
```
BrianHarrisonAMD, line 595: varIndex was always +2 regardless of precision before it looks like.

That's surprising, we will need to make sure everywhere using mvbuff is swapped to the new logic that handles it different depending on the datatype.

src/solver/batchnorm/backward_spatial_multiple.cpp
```diff
@@ -51,17 +51,50 @@ bool BnBwdTrainingSpatialMultiple::IsApplicable(
...
+        if(problem.GetXDesc().GetType() == miopenFloat)
+        {
+            if(last_ylocalsize < stash_values)
+                return false;
+        }
+        else
+        {
+            // Even threads use 2 values at even rows, odd threads - at odd rows.
+            if(c % 2 != 0 || last_ylocalsize < stash_values * 2)
+                return false;
```
BrianHarrisonAMD, line 81: For mixed NHWC multiple can't handle a c that isn't a multiple of 2?

src/solver/batchnorm/backward_spatial_single.cpp
```diff
@@ -40,8 +40,11 @@ namespace solver {
 namespace batchnorm {
 
 bool BnBwdTrainingSpatialSingle::IsApplicable(
-    const ExecutionContext&, const miopen::batchnorm::ProblemDescription& bn_problem) const
+    const ExecutionContext& context, const miopen::batchnorm::ProblemDescription& bn_problem) const
 {
+    if(BnBwdTrainingSpatialMultiple{}.IsApplicable(context, bn_problem))
+        return false;
+
```
BrianHarrisonAMD, line 47: Did we want to start preferring multiple over single generally?
We used to do the opposite for fwd training at least.

src/solver/batchnorm/backward_spatial_single.cpp
```diff
@@ -61,14 +64,7 @@ bool BnBwdTrainingSpatialSingle::IsApplicable(
...
 
-    int n, c, h, w;
-    std::tie(n, c, h, w) = tien<4>(bn_problem.GetXDesc().GetLengths());
-
-    unsigned int in_cstride = h * w;
-    unsigned int in_nhw     = n * in_cstride;
-
-    return (in_cstride > 1024 && in_nhw < (32 * 1024 * 1024)) ||
-           (in_cstride > 512 && in_nhw < (32 * 1024 * 1024)) || in_cstride <= 512;
+    return true;
```
BrianHarrisonAMD, line 67: Do we not still need these checks?

I thought there was limits to what the single version could support for size.

src/solver/batchnorm/forward_spatial_multiple.cpp
```diff
@@ -50,7 +48,54 @@ bool BnFwdTrainingSpatialMultiple::IsApplicable(
...
+        if(problem.GetXDesc().GetType() == miopenFloat)
+        {
+            if(last_ylocalsize < stash_values)
+                return false;
+        }
+        else
+        {
+            // Even threads use 2 values at even rows, odd threads - at odd rows.
+            if(c % 2 != 0 || last_ylocalsize < stash_values * 2)
+                return false;
```
BrianHarrisonAMD, line 77: Training also can't handle c % 2 != 0 NHWC for mixed precision?

src/solver/batchnorm/forward_spatial_multiple.cpp
```diff
@@ -208,9 +213,11 @@ ConvSolution BnFwdTrainingSpatialMultiple::GetSolution(
         result.construction_params.push_back(copy);
 
         copy.kernel_name = kernel.kernel_name + "FinalMeanVariance";
+        copy.g_wk[1]     = kernel.l_wk[1];
```
BrianHarrisonAMD, line 222: Why does FinalMeanVariance change the y grid size to match the localsize?

BrianHarrisonAMD: Ah just a single group for FinalMeanVariance?

src/solver/batchnorm/forward_spatial_single.cpp
```diff
@@ -40,51 +40,18 @@ namespace solver {
...
-          ((in_cstride > 512) && bfpmixparm) ||
-          in_cstride <= 512))
-        return false;
-    // clang-format on
-
-    if((n > 768) && (in_cstride > 150) && bfp32parm)
-    {
-        return false;
-    }
-
```
BrianHarrisonAMD, line 87: I think we still need some of these checks.
I thought single has limits to the sizes it can handle.

src/solver/batchnorm/forward_spatial_single.cpp
```diff
@@ -40,51 +40,18 @@ namespace solver {
 namespace batchnorm {
 
 bool BnFwdTrainingSpatialSingle::IsApplicable(
-    const ExecutionContext&, const miopen::batchnorm::ProblemDescription& bn_problem) const
+    const ExecutionContext& context, const miopen::batchnorm::ProblemDescription& bn_problem) const
 {
+    if(BnFwdTrainingSpatialMultiple{}.IsApplicable(context, bn_problem))
+        return false;
```
BrianHarrisonAMD, line 46: Swapping to prefer multiple over single as well here?

test/gtest/bn_test_data.hpp
```diff
@@ -189,12 +199,13 @@ struct BNTestData
 
     void InitTensorsWithRandValue()
     {
-        input.generate(
-            [](auto...) { return prng::gen_descreet_uniform_sign<XDataType>(1e-2, 100); });
+        // 0.0 to 2.0 (since unsigned)
+        input.generate([](auto...) {
+            return prng::gen_descreet_unsigned<XDataType>(2e-3 /*scale*/, 1000 /*range*/);
+        });
```
BrianHarrisonAMD: Similar to driver, any reason this needs to be unsigned?

test/gtest/bn_test_data.hpp
```diff
@@ -253,15 +265,17 @@ struct BNInferTestData : public BNTestData<XDataType, YDataType, TConfig>
...
+            [](auto...) { return prng::gen_descreet_uniform_sign<BiasDataType>(2e-3, 1000); });
+        estMean.generate(
+            [](auto...) { return prng::gen_descreet_uniform_sign<MeanVarDataType>(2e-3, 1000); });
+
+        // 0.0 to 2.0
         auto gen_var = [](auto...) {
-            return static_cast<MeanVarDataType>(1e-2 * (prng::gen_0_to_B(100) + 1));
+            return static_cast<MeanVarDataType>(2e-3 * (prng::gen_0_to_B(1000) + 1));
         };
         estVariance.generate(gen_var);
```
BrianHarrisonAMD: Same here, why swapping values around?

Probably best to match driver behavior.

src/solver/batchnorm/backward_spatial_single.cpp
ex-rzr: The issue with the original order is that Single and Multiple need to check exactly the same conditions (and this conditions must ), i.e. Single must know when Multiple can be applied. That's why Single has `return (in_cstride > 1024 && in_nhw < (32 * 1024 * 1024)) ||...` - this is conditions that Multiple also checks.

src/solver/batchnorm/backward_spatial_multiple.cpp
ex-rzr: Yes, because to store one 32-bit intermediate values we use 16-bit values in C dim (for NHWC).
Theoretically, we can store 16-bit halves in different rows (i.e. in HW dim), but that makes indexing even more complicated.

So currently we store as
```
L0,H0,L2,H2,L4,H4...
L1,H1,L3,H3,L5,H5...
...
```
where L and H are low and high 16-bit parts of one 32-bit value.
But we also can store as
```
L0,L1,L2,L3...
H0,H1,H2,H3...
```
There are pros and cons for this approach:
* C can be odd;
* but we need to do two 16-bit loads/stores for each 32-bit value

src/solver/batchnorm/backward_spatial_single.cpp
ex-rzr: These checks are for Multiple, correct me if I'm mistaken.

src/kernels/reduction_functions.h
ex-rzr: MIO_BN_LDS_SIZE == ygrp_sz * xgrp_sz always.

Originally, Multiple supported only NCHW with xlocalsize = 1 and ylocalsize = 1024, the naive support of NHWC also used xlocalsize = 1, but this means that all memory accesses are with huge strides, not coalesced (about 10x slower in most cases).
To optimize it we changed workgroup size to shapes like xlocalsize = 64 and ylocalsize = 16. Hence each workgroup needs to do xlocalsize reductions, not 1, because each workgroups processes xlocalsize channels.

There are bank conflicts, hard to say if they affect performance (they can be fixed with padding).

src/batch_norm.cpp
ex-rzr: Yes, that's why the first kernel was ignored in the total duration

src/kernels/batchnorm_functions.h
ex-rzr: Good idea

src/kernels/batchnorm_functions.h
ex-rzr: We need two 16-bit positions to store one 32-bit value. To make it less restrictive on values of C (and C % xlocalsize, i.e. the size of the last workgroup) , we store even and odd values in different rows (HW dim).

src/solver/batchnorm/forward_spatial_multiple.cpp
ex-rzr: Yes, previously the Final kernels run for the full size, basically all groups read and calculated exactly the same final values, BUT they also wrote the finals to their own part of stash (the finals can be stored once because eventually one group would overwrite it with out values as stash == out).

The current approach is to use a single group in y dim (there are still many groups x dim, i.e. for channels, as they are independent). BUT we still need to write the final for each workgroup for the same reason.

This kernel usually takes a small part or the overall duration, but the new approach is faster.

src/kernels/MIOpenBatchNormFwdTrainSpatial.cl
ex-rzr: Yes, x - channels, y - HW

src/kernels/MIOpenBatchNormFwdTrainSpatial.cl
ex-rzr: The indices for stashed values were strange:
for bwd they were: (0,2) ; (1,3); (4;6). Why 6, not 5? I don't know.
We changed them to 
(0,1) ; (2,3) ; (4,5) for bwd
(0,1) ; (2,3) for fwd

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: For fp16, fp32 supports C = 1 and odd C in general

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: I'm not even sure that the old loop was entirely correct.
Let's assume we have H*W = 10000, NCHW
ygrp_sz = 1024
yngrps = 10
```
for(int gn = 0; gn < yngrps; gn++) // gn = [0..10)
  offset = gn * ygrp_sz + lid; // offset = gn*1024+[0, 1024)
  if(offset < yngrps) // the first iteration is enough to process all yngrps
```
So this loop does more iterations than needed.

Now we iterate over all MIO_BN_NGRPS (remember that in the old kernel yngrps = MIO_BN_NGRPS, in the new kernel yngrps = 1), the loop does as many iterations as needed.

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: `gcn_reduce2` calculates reduction over the whole workgroup, it can't be used with more than 1 channel per workgroup (1 channel is currently used with NCHW, or  NHWC, fp32 and C = 1).

On RDNA, MIO_BN_GRP1 = 1 (1 channel per workgroup) the new `lds_reduce2_2d` == the old `lds_reduce2`.

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: The kernel does this (correct me if I'm wrong)
1. reads values from global memory `meanvarbuff`
2. reduces them, with synchronization required for local memory
3. writes the results to global memory `meanvarbuff`

I don't see any ways that allows any thread to reach (3) without finishing all reads in (1) and synchronization in (2).

Also, I haven't checked (I don't know how to dump ISA for the kernels in MIOpen), but I suspect that `CLK_GLOBAL_MEM_FENCE` may be extremely slow on MI300.

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: Actually, I think with the new kernel (that is run with one workgroup in y dim) we can reuse 0, 1 again, because there is no conflict between workgroups as it was before, when multiple groups read and calculated the same values.
That means that we need 4 stashed values, not 6 for bwd.

Let me check to be 100% sure.

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: Yes, this was incorrect. I spent quite a lot of time to understand what was wrong.

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: Oh, good catch!

The condition must use `MIO_BN_GRP0`, so `lds_reduce2_2d` was always used.

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: The new Final kernels run with one workgroup in y dim, ygid == ylid

src/solver/batchnorm/forward_spatial_single.cpp
EnricoDeg: The checks are in multiple and then single calls `BnFwdTrainingSpatialMultiple{}.IsApplicable(...)`. In this way, we have all the checks in one place.

src/solver/batchnorm/forward_spatial_single.cpp
EnricoDeg: The previous logic should not be changed. The difference is that all the checks are now in multiple instead of repeating them in single and multiple.

src/solver/batchnorm/forward_spatial_multiple.cpp
EnricoDeg: Yes, the reason is the same as for bwd. Because to store one 32-bit intermediate value, we use 16-bit values in C for NHWC.

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: Yes, it's possible now. I added this change to our branch, there are some other changes, it needs to be merged to this one.

fwd now uses 2 values, bwd - 4.

src/kernels/MIOpenBatchNormBwdSpatial.cl
ex-rzr: Fixed in our branch

src/kernels/batchnorm_functions.h
ex-rzr: Fixed in our branch

src/kernels/batchnorm_functions.h
ex-rzr: Fixed in our branch

src/kernels/MIOpenBatchNormBwdSpatial.cl
BrianHarrisonAMD: Ah yea the if condition makes it similar. 
It seems odd to me to compare offset to the number of groups, but it was checking the same condition before basically.

src/kernels/MIOpenBatchNormBwdSpatial.cl
BrianHarrisonAMD: From comment above I think we just need to verify the GCN path.

src/solver/batchnorm/backward_spatial_single.cpp
BrianHarrisonAMD: Do you mean these restrictions were only for multiple?
Or do you mean that this check is for the multiple kernel solution (variant 2)? 

I think we will need to review, but this file is checking the single solver applicability, and I thought there was some limits to what it could support.

src/solver/batchnorm/backward_spatial_multiple.cpp
BrianHarrisonAMD: Does that mean that we have a gap in coverage for FP16 / BF16 mixed if C == 1? 

I thought there was conditions where the single solver couldn't be used due to size limits, and multiple had to be used. 
If that's the case, then I think at a certain size we wouldn't have a solver that could be used when C == 1.

src/solver/batchnorm/backward_spatial_single.cpp
BrianHarrisonAMD: I think we will need to get @bghimireamd to review since he had changed the IsApplicable quite a bit, and I thought single / multiple had different conditions for support. 

That being said, it does need to prefer one over the other.
If we think multiple (variant 2) is better to prefer, then that seems fine.

src/solver/batchnorm/forward_spatial_single.cpp
BrianHarrisonAMD: I think it'll use multiple if it can, but if it can't, then it falls back to single due to this check:

```C++
if(BnFwdTrainingSpatialMultiple{}.IsApplicable(context, bn_problem))
        return false;
```

If the coverage is identical between single and multiple, then I think this is an issue.
It'll fail the check for multiple, and then potentially allow single afterwards to run.

test/random.hpp
```diff
@@ -40,5 +40,26 @@ inline T gen_descreet_unsigned(double scale, int32_t range)
...
+        // uniform sign give balance of both negative and positive values
+        return prng::gen_descreet_uniform_sign<T>(scale_arg, range_arg);
+    };
+}
+
+template <typename T, typename ScaleT, typename RangeT>
+auto uniform_unsigned_initializer(ScaleT scale_arg, RangeT range_arg)
+{
+    return [=](auto&&...) -> T {
+        // uniform sign give balance of both negative and positive values
```
bghimireamd: will remove this comment.

BrianHarrisonAMD (approved):
LGTM

PR #3400, opened 2024-11-21
Added a .gitignore with standard content

BradPepersAMD:
Use the one from CK since it seems a close match to our needs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,68 @@
+# Compiled Object files
+*.slo
+*.lo
+*.o
+*.obj
+
+# Precompiled Headers
+*.gch
+*.pch
+*.ipch
+
+# Compiled Dynamic libraries
+*.so
+*.dylib
+*.dll
+
+# Fortran module files
+*.mod
+
+# Compiled Static libraries
+*.lai
+*.la
+*.a
+*.lib
+
+# Executables
+*.exe
+*.out
+*.app
+
+# vim tags
+tags
+.tags
+.*.swp
+
+# Editors
+.vscode
+
+# build-in-source directory
+build*
+
+# emacs temporary/backup files
+.\#*
+\#*\#
+*~
+
+# GDB temporary files
+.gdb_history
+install.dir*
+
+# documentation artifacts
+_build/
+_images/
+_static/
+_templates/
+_toc.yml
+_doxygen/
+
+# JetBrains IDE
+.idea/
+cmake-build*/
+build*/
+
+# Python virtualenv
+.venv/
+
+# Python cache
+__pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,6 @@ tags
 .vscode
 
 # build-in-source directory
-build*
 
 # emacs temporary/backup files
 .\#*


BrianHarrisonAMD (approved):
LGTM

averinevg:
Why do we need this file? It's usually useful for developers who automatically commit all their changes. I am against this practice, as every developer should manually review everything they've written, and only then manually add each file before committing. The repository should not contain junk.

CAHEK7:
I love `git add .`

.gitignore
```diff
@@ -0,0 +1,68 @@
...
+*.pch
+*.ipch
+
+# Compiled Dynamic libraries
+*.so
+*.dylib
+*.dll
+
+# Fortran module files
+*.mod
```
CAHEK7, line 18: We don't have any fortran sources

.gitignore
```diff
@@ -0,0 +1,68 @@
+# Compiled Object files
+*.slo
+*.lo
+*.o
+*.obj
```
CAHEK7, line 5: We don't build in-tree, encouraging this sound strange, but accpetable.

.gitignore
```diff
@@ -0,0 +1,68 @@
...
+
+# Precompiled Headers
+*.gch
+*.pch
+*.ipch
+
+# Compiled Dynamic libraries
+*.so
+*.dylib
+*.dll
```
CAHEK7, line 15: We don't build binaries in-tree.

.gitignore
```diff
@@ -0,0 +1,68 @@
+# Compiled Object files
+*.slo
+*.lo
+*.o
+*.obj
+
+# Precompiled Headers
+*.gch
+*.pch
+*.ipch
```
CAHEK7, line 10: That may be useful, but we don't use gcc (clang can create precompiled headers everywhere) and we dont use precompiled headers.

.gitignore
```diff
@@ -0,0 +1,68 @@
...
+# Compiled Static libraries
+*.lai
+*.la
+*.a
+*.lib
+
+# Executables
+*.exe
+*.out
+*.app
```
CAHEK7, line 29: We don't build in-tree.

.gitignore
```diff
@@ -0,0 +1,68 @@
...
+.gdb_history
+install.dir*
+
+# documentation artifacts
+_build/
+_images/
+_static/
+_templates/
+_toc.yml
+_doxygen/
```
CAHEK7, line 56: I'm not sure that it is relevant to our project.

CAHEK7 (approved):

.gitignore
```diff
@@ -0,0 +1,68 @@
...
+
+# JetBrains IDE
+.idea/
+cmake-build*/
+build*/
+
+# Python virtualenv
+.venv/
```
averinevg, line 64: I don't think we need this. We don't have scripts that create the `.venv` automatically, and the path to it can be anything.

.gitignore
```diff
@@ -0,0 +1,68 @@
...
+_images/
+_static/
+_templates/
+_toc.yml
+_doxygen/
+
+# JetBrains IDE
+.idea/
+cmake-build*/
+build*/
```
averinevg, line 61: All editors should be in one place, not scattered throughout the file. BTW, does anyone use JetBrains IDE?

.gitignore
```diff
@@ -0,0 +1,68 @@
...
+
+# Executables
+*.exe
+*.out
+*.app
+
+# vim tags
+tags
+.tags
+.*.swp
```
averinevg, line 34: I think that in their time, Vim and Emacs were truly amazing tools, but do we really need to have them here?

averinevg (dismissed):

.gitignore
CAHEK7: `.swp` might be useful, that happens from time to time for me, especially I ssh session got accidently disconnected.

.gitignore
```diff
@@ -0,0 +1,68 @@
...
+
+# Editors
+.vscode
+
+# build-in-source directory
+build*
```
CAHEK7: ~Simple `build*` ignores too much of everything, like if we want to add script `build-all.py` or whatever.~
```suggestion
```
~I didn't want do request any changes, because it's actually not important and we can live with any kind of `.gitignore` file except of `*`.~

Nevermind, that line should actually be deleted, since it is covered by the line 62 (`build*/`) and without strange implications of `build*`.

.gitignore
BradPepersAMD: yep.  this is a stock file covering common issues across different types of projects.

.gitignore
BradPepersAMD: yes but stock list of things you wouldn't want to include and none of them will clash with any file you would create and want to include in versioning.

.gitignore
BradPepersAMD: Well I still use emacs sometimes!  But never vim!!!

PR #3402, opened 2024-11-22
Changed all references and pointers to Handle to be const

DrizztDoUrden:
This allows things like ExecutePrimitive to be called without non-const Handle


CAHEK7 (approved):

src/adam.cpp
```diff
@@ -35,7 +35,7 @@
 
 namespace miopen {
 
-miopenStatus_t Adam(Handle& handle,
+miopenStatus_t Adam(Handle const& handle,
```
BradPepersAMD: If you look at the other parameters, they are all defined using "const type&" rather than "type const&".  Can we please keep this consistent and use "const type&" in all locations.  Hopefully this can just be a quick find/replace on "Handle const&" to "const Handle&".

src/include/miopen/execution_context.hpp
```diff
@@ -98,11 +98,11 @@ struct MIOPEN_INTERNALS_EXPORT ExecutionContext
     bool use_dynamic_solutions_only = false;
     bool is_for_generic_search      = false;
 
-    inline Handle& GetStream() const { return *stream; }
-    inline void SetStream(Handle* stream_) { stream = stream_; }
+    inline Handle const& GetStream() const { return *stream; }
+    inline void SetStream(Handle const* stream_) { stream = stream_; }
```
BradPepersAMD: same with the order on the pointer ones please.

src/include/miopen/mlo_internal.hpp
```diff
@@ -274,7 +274,7 @@ struct mlo_construct_base
     /*
      * set library stream
      */
-    inline void setStream(miopen::Handle* stream) { _ctx.SetStream(stream); }
+    inline void setStream(miopen::Handle const* stream) { _ctx.SetStream(stream); }
```
BradPepersAMD: slightly different one to catch where const is at

BradPepersAMD (dismissed):
Love to see const correctness improved!  Need to follow same declaration syntax used for const on other references.

BrianHarrisonAMD (approved):
Looks good to me!

Love to see it!

shurale-nkn (approved):
LGTM

shurale-nkn (approved):
LGTM

PR #3403, opened 2024-11-23
Fix Nan issue in convolution asm Kernel with NCHW layout

iq136boy:
Fix Nan issue in convolution asm Kernel with NCHW layout by clearing the buffer.

### DIFF
--- a/src/conv/invokers/impl_gemm_dynamic.cpp
+++ b/src/conv/invokers/impl_gemm_dynamic.cpp
@@ -179,8 +179,6 @@ InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ProblemDescri
         is_gemm_not_empty.emplace_back(gemm_k_gid > 0);
     }
     bool need_set_zero = true;
-    // if(y < stride_h || x < stride_w || dilation_h != 1 || dilation_w != 1)
-    //    need_set_zero = true;
 
     return [=](const std::vector<Kernel>& kernels) {
         const auto kernel = kernels[0];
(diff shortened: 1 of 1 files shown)


src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -178,9 +178,9 @@ InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ProblemDescri
         const int gemm_k_gid = k * y_dot_slice_gid[gemm_id] * x_dot_slice_gid[gemm_id];
         is_gemm_not_empty.emplace_back(gemm_k_gid > 0);
     }
-    bool need_set_zero = false;
-    if(y < stride_h || x < stride_w || dilation_h != 1 || dilation_w != 1)
-        need_set_zero = true;
+    bool need_set_zero = true;
```
randyspauldingamd, line 181: It's unfortunate to remove this micro-opt altogether. What if we add it to ProblemDescription with a default value=true. Then the caller can choose when it is safe to skip the zeroing (which was probably the preferred logic since the beginning).

Otherwise, just delete the variables entirely and remove the conditionals from the invokers.

randyspauldingamd (changes requested):

src/conv/invokers/impl_gemm_dynamic.cpp
BradPepersAMD: I feel that effort is outside the scope of this fix since it will greatly expand the effort on something that is critical to get a fix for quickly.  If there is a desire to clean up tech debt here, make it a separate ticket please.

BrianHarrisonAMD: I am tracking some follow up work for these NaN issues that we could add this task to. 

I do think we should remove the commented out code though.

xinlipn: > I am tracking some follow up work for these NaN issues that we could add this task to.
> 
> I do think we should remove the commented out code though.

HI @BrianHarrisonAMD , cleaned up commented code

randyspauldingamd (approved):

BradPepersAMD (approved):
I think more cleanup needs to happen with this and consideration across other kernels on this clearing need but I want to get this fix in ASAP to resolve more of these numeric issues so will merge this now as is.

PR #3406, opened 2024-11-27
RNN Dynamic algo single stream

shurale-nkn:
Added new algo type (miopenRNNroundedDynamic) for LSTM. Should produce less unique requests to rocblass. 
Prerequisite for multi-stream version.
Prerequisite for native support of the padded_seq layout.


src/include/miopen/rnn/tmp_buffer_utils.hpp
```diff
@@ -64,6 +64,16 @@ OutputIt exclusive_scan_wa(InputIt first, InputIt last, OutputIt d_first, T init
 
 } // namespace WA_RHEL
 
+inline Data_t moveDataPtrByte(Data_t ptr, size_t byteOffset)
+{
+    return static_cast<Data_t>(reinterpret_cast<char*>(ptr) + byteOffset);
```
CAHEK7: If `Data_t` is `void*` then `static_cast` is enough.

src/rnn/Solutions/bwd_s_stream.cpp
```diff
@@ -134,5 +134,94 @@ void RNNModularSingleStreamBWD::ComputeBWD(Handle& handle,
...
+                dhy,
+                dhx,
+                cx,
+                dcy,
+                dcx,
+                real_dx,
+                temp_dx,
+                w,
+                workSpace,
+                reserveSpace] = runtimeArgsExt;
```
CAHEK7:
```suggestion
    const auto runtimeArgsExt = rnnAlgoModules.createRuntimeArgsExt(realRuntimeArgs);
    const auto& [real_dy,
                temp_dy,
                dhy,
                dhx,
                cx,
                dcy,
                dcx,
                real_dx,
                temp_dx,
                w,
                workSpace,
                reserveSpace] = runtimeArgsExt;
```
I'm not sure that the full copy of everything is needed. Reference seems reasonable in that case.
https://godbolt.org/z/afG9rsTP9

src/include/miopen/rnn/algorithms/default_algo_utils.hpp
```diff
@@ -0,0 +1,724 @@
...
+    // RNNModuleAlgoBase(RNNModuleAlgoBase const&) = default;
+
+    RNNModuleAlgoBase(const GeneralLstmRedBuffer& rb_layout,
+                      const GeneralLstmTempBuffer& workspace_info,
+                      const WeightsBufferDescriptor& weights_layout,
+                      const HiddenBuffersDescriptor& hidden_hxcx_info,
+                      const IOBufferDescriptor& x_info,
+                      const IOBufferDescriptor& y_info,
+                      const RNNDescriptor& rnn_desc,
+                      const BatchController& batch_controller,
```
CAHEK7, line 164:
```suggestion
    RNNModuleAlgoBase(GeneralLstmRedBuffer&& rb_layout,
                      GeneralLstmTempBuffer&& workspace_info,
                      WeightsBufferDescriptor&& weights_layout,
                      HiddenBuffersDescriptor&& hidden_hxcx_info,
                      IOBufferDescriptor&& x_info,
                      IOBufferDescriptor&& y_info,
                      RNNDescriptor&& rnn_desc,
                      BatchController&& batch_controller,
```
And move everything. There is a bunch of vectors under the hood of descripotrs.
But probably it's not critical, if all the vectors will be replaced with inline vector (which is coming soon).

CAHEK7 (approved):
There are some minor recommendations.
Unfortunately I can't go deep into the code, since I'm not familiar with the algorithm itself.

src/include/miopen/rnn/algorithms/default_algo_utils.hpp
shurale-nkn: well, yes and no)  all objects are not standard data descriptors. Part of them is array, part need to convert to array, part should be vector just because stored dynamic data size.
RNNDescriptor only ref, because the lifetime of a class cannot be longer than the lifetime of the object it lives in.

src/rnn/Solutions/bwd_s_stream.cpp
shurale-nkn: Good point

shurale-nkn: But this is not only a copy, the main compiler optimization to save data in register (and read only once from stack) if we have multiple usage in this function, in the moment it cost's us extra push and pop (from stack) operation per used register. But in reference case this is read from stack each data for each function.

CAHEK7: It depends, it does not require to be on stack - it's not volatile and it's up to compiler to decide. Reference can be a pointer, but actually it's just an alias to some entity and it can perfectly live on registers, if compiler decides that it worth living there.
If you take a look into an example - the copy is a real copy and it resides on stack.

1) `rnnAlgoModules.createRuntimeArgsExt` has always returned on stack.
2) `RuntimeArgsExt` is quite huge - 12 poninter
3) You need a readonly access
4) x86-64 has 16 GPRs, but actually ~14 or less
5) following code is quite complex, has some function calls (more GPRs wasted on class), lots of read-write local variables, etc.

I would not expect even the half of those variables can be places on registers. If it would be a tight loop and a small struct, then may be.
Btw, you can disassemble and compare 2 cases, it's rather simple.

P.S. I'm not insisting on that change, I'm just saying that the assumption may be far from reality.

shurale-nkn:
Hi @junliume @BrianHarrisonAMD @BradPepersAMD, when we can merge this ? Thanks

BradPepersAMD (commented):
I'm not seeing any unit tests that seem to really validate the new RNN algorithm from a correctness perspective.  Is there something I'm missing?

shurale-nkn:
> I'm not seeing any unit tests that seem to really validate the new RNN algorithm from a correctness perspective. Is there something I'm missing?

For the user, the result of this algorithm is identical to the standard one. Therefore, the general test logic extended by testing with this algorithm is sufficient. Partially, this was the reason for adding a separate type.

junliume (approved):

PR #3407, opened 2024-11-28
[Bug] Fix issue with WRW heuristics instance selection

BrianHarrisonAMD:
After the fix for #3249 WRW heuristics would fail to parse split_k properly, and this would lead to using the default instance in all cases. This change will prevent issues where the split_k value is missing or malformed, and also allow for valid cases to parse.

### DIFF
--- a/src/include/miopen/solver/implicitgemm_ck_util.hpp
+++ b/src/include/miopen/solver/implicitgemm_ck_util.hpp
@@ -140,7 +140,7 @@ bool IsCKArgsSupported(const ProblemDescriptionType& problem, const std::string&
                      std::is_same_v<DeviceOpType, conv::DeviceOpGWrwPtrs<ck::bhalf_t>>)
         {
             auto pos = kernel_id.find_last_of('+');
-            if(pos == string::npos)
+            if(pos == std::string::npos)
             {
                 MIOPEN_LOG_I2("Unable to parse split_k from kernel_id for wrw: " << kernel_id);
                 return false;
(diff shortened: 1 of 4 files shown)


test/gtest/unit_implicitgemm_ck_util.cpp
```diff
@@ -0,0 +1,162 @@
...
+};
+} // namespace unit_implicitgemm_ck_util_test
+
+using namespace unit_implicitgemm_ck_util_test;
+
+struct CPU_UnitTestImplicitGemmCKUtil_NONE : CKArgParsingTest<StubbedWRWCKArgs, StubbedDeviceOps>
+{
+};
+
+// TEST_F(CPU_UnitTestImplicitGemmCKUtilWRW_NONE, TestParsing) { TestParsing(); }
```
BrianHarrisonAMD: whoops ill remove

cderb (approved):
lgtm

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -137,10 +138,28 @@ bool IsCKArgsSupported(const ProblemDescriptionType& problem, const std::string&
...
+            {
+                split_k = std::stoi(kernel_id.substr(pos + 1));
+            }
+            catch(std::exception& e)
+            {
+                MIOPEN_LOG_WE("Unable to parse split_k from kernel_id for wrw: "
+                              << kernel_id << " : " << e.what());
+                return false;
+            }
+
```
CAHEK7, line 162: Doesn't is complain "set but not used variable" on `split_k`?

CAHEK7 (approved):

CAHEK7:
Probably it's better to use `Mock` instead of `Stubbed`, at least it's kind of widely used test terminology for fake instances of anything.

src/include/miopen/solver/implicitgemm_ck_util.hpp
BrianHarrisonAMD: It doesn't actually, but that is a bit surprising.

BrianHarrisonAMD:
> Probably it's better to use `Mock` instead of `Stubbed`, at least it's kind of widely used test terminology for fake instances of anything.

Good point!
 I should do this with google mock, and the name should be Mock.

PR #3412, opened 2024-11-28
Remove cout logging

BrianHarrisonAMD:
Remove print statements that should be only logging messages.


BradPepersAMD (approved):
obvious win

PR #3388, opened 2024-11-15
Revert "Fixed incorrect transpose in find 2.0"

DrizztDoUrden:
Reverts ROCm/MIOpen#3285

PR #3413, opened 2024-11-29
Produced solutions now contain original problem

DrizztDoUrden:
Also added a transposed test case for each direction for find 1.0 and find 2.0 tests

### DIFF
--- a/src/include/miopen/problem.hpp
+++ b/src/include/miopen/problem.hpp
@@ -186,7 +186,8 @@ private:
                                             const FindOptions& options,
                                             std::size_t max_solutions,
                                             const Buffers& buffers,
-                                            const ConvolutionDescriptor& conv_desc) const;
+                                            const ConvolutionDescriptor& conv_desc,
+                                            Problem const& problem) const;
 
     std::vector<Solution> FindSolutionsImpl(Handle& handle,
                                             const FindOptions& options,
--- a/src/include/miopen/problem.hpp
+++ b/src/include/miopen/problem.hpp
@@ -187,7 +187,7 @@ private:
                                             std::size_t max_solutions,
                                             const Buffers& buffers,
                                             const ConvolutionDescriptor& conv_desc,
-                                            Problem const& problem) const;
+                                            const Problem& problem) const;
 
     std::vector<Solution> FindSolutionsImpl(Handle& handle,
                                             const FindOptions& options,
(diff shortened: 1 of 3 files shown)


src/include/miopen/problem.hpp
```diff
@@ -186,7 +186,8 @@ struct Problem
                                             const FindOptions& options,
                                             std::size_t max_solutions,
                                             const Buffers& buffers,
-                                            const ConvolutionDescriptor& conv_desc) const;
+                                            const ConvolutionDescriptor& conv_desc,
+                                            Problem const& problem) const;
```
BrianHarrisonAMD: Not a major issue, but can we keep the const location consistent?

src/problem.cpp
```diff
@@ -181,9 +181,10 @@ Problem::FindSolutions(Handle& handle, const FindOptions& options, std::size_t m
             [&](const ConvolutionDescriptor& op_desc) {
                 if(op_desc.mode == miopenTranspose)
                     return MakeTransposed().FindSolutionsImpl(
-                        handle, options, max_solutions, buffers, op_desc);
+                        handle, options, max_solutions, buffers, op_desc, *this);
```
BrianHarrisonAMD, line 184: I can see why the original change started causing issues since we would be giving back the transposed problem with the solution, and that can cause issues later on.

However, for the reverted version from #3388 why was it causing issues / fixed others?
These methods are const, and it seems like FindSolutionsImpl must be changing state despite being const.

src/problem.cpp
```diff
@@ -464,7 +465,8 @@ std::vector<Solution> Problem::FindSolutionsImpl(Handle& handle,
                                                  const FindOptions& options,
                                                  std::size_t max_solutions,
                                                  const Buffers& buffers,
-                                                 const ConvolutionDescriptor& conv_desc) const
+                                                 const ConvolutionDescriptor& conv_desc,
+                                                 Problem const& original) const
```
BrianHarrisonAMD: Same here with const location.

src/include/miopen/problem.hpp
DrizztDoUrden: Sorry, a habit and was writing in a hurry
Sure, will change

src/problem.cpp
DrizztDoUrden: MakeTransposed() produces a pre-transposed miopen::Problem, which has different direction and swapped descriptors compared to *this

IIRC, reverted version actually has several problems: ValidateGroupCount is called with y descriptor instead of x and something is swapped incorrectly, but I do not remember what exactly

Current approach combines actually correct behavior of FindSolutionImpl from 3285 (the bug there was storing wrong problem) with original problem stored in the solution, thus RunSolution also having correct behavior.

BrianHarrisonAMD: From the revert PR it looks like the swap was happening, but I could believe that it wasn't working as expected, which lead to knock on issues which this version will resolve.

BradPepersAMD (approved):

PR #3415, opened 2024-12-01
Codegen integration

arai713:


test/gtest/codegen_group_conv.hpp
```diff
@@ -0,0 +1,567 @@
...
+        input.generate(gen_value_in);
+        output.generate(gen_value_out);
+        std::fill(weights.begin(), weights.end(), T{0});
+    }
+
+    void verify(const tensor<T>& computed)
+    {
+        EXPECT_FALSE(miopen::range_zero(ref)) << "Cpu data is all zeros";
+        EXPECT_FALSE(miopen::range_zero(computed)) << "Gpu data is all zeros";
+        EXPECT_TRUE(miopen::range_distance(ref) == miopen::range_distance(computed));
```
CAHEK7:
```suggestion
        EXPECT_EQ(miopen::range_distance(ref), miopen::range_distance(computed));
```

test/gtest/codegen_group_conv.hpp
```diff
@@ -0,0 +1,567 @@
...
+        {
+            threshold *= std::numeric_limits<T>::epsilon();
+        }
+        else
+        {
+            threshold = 3.0e-3;
+        }
+        auto error = miopen::rms_range(ref, computed);
+
+        EXPECT_FALSE(miopen::find_idx(ref, miopen::not_finite) >= 0)
```
CAHEK7:
```suggestion
        EXPECT_LT(miopen::find_idx(ref, miopen::not_finite), 0)
```

test/gtest/codegen_group_conv.hpp
```diff
@@ -0,0 +1,567 @@
...
+            << "Non finite number found in the reference output";
+
+        EXPECT_TRUE(error <= threshold)
+            << "Error beyond tolerance Error:" << error << ",  Threshold: " << threshold;
```
CAHEK7:
```suggestion
        EXPECT_LE(error, threshold);
```
It will print something like:
```
Mismatch, error must be less than threshold :
error = 100500
threshold = 10
```

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
```diff
@@ -0,0 +1,314 @@
...
+#endif
+    return false;
+}
+
+ConvSolution ConvHipImplicitGemmGroupFwdXdlopsCodegen::GetSolution(
+    [[maybe_unused]] const ExecutionContext& ctx,
+    [[maybe_unused]] const ProblemDescription& problem) const
+{
+    auto x = CKArgs(problem);
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
```
CAHEK7:
```suggestion
#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
    auto x = CKArgs(problem);
```

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
```diff
@@ -0,0 +1,314 @@
...
+            auto kernel = handle_.AddKernel("tmp",
+                                            "tmp",
+                                            "cg_main.cpp",
+                                            kernel_info.kernel_name,
+                                            kernel_info.l_wk,
+                                            kernel_info.g_wk,
+                                            kernel_info.comp_options,
+                                            0,
+                                            src);
+
```
CAHEK7: Usually you should add the kernels outside of this lambda and later just use added kernels.
Like here: https://github.com/ROCm/MIOpen/blob/6c4a0b3d0a462fdbdbeebb0ccadcf27f64bfa930/src/solver/softmax/softmax.cpp#L336

The list of added kernels passed through `const std::vector<Kernel>& kernels` into that lambda.

src/solver.cpp
```diff
@@ -678,6 +678,11 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
              fusion::ConvWinoFuryRxSFused<2, 3>{}.SolverDbId(),
              miopenConvolutionAlgoWinograd);
 
+    RegisterWithSolver(registry,
+                       ++id,
+                       conv::ConvHipImplicitGemmGroupFwdXdlopsCodegen{},
+                       miopenConvolutionAlgoImplicitGEMM);
+
```
CAHEK7: It should be added at the end of the list, otherwise it will affect all the others solver ids.

test/gtest/codegen_group_conv.hpp
```diff
@@ -0,0 +1,567 @@
...
+
+        auto ctx = miopen::ExecutionContext{};
+
+        ctx.SetStream(&handle);
+
+        if(!solv.IsApplicable(ctx, problem))
+        {
+            test_skipped = true;
+            GTEST_SKIP() << solv.SolverDbId() << "Not Applicable for this problem" << conv_config;
+        }
```
CAHEK7: Basically this condition should never happen.
It's not quite good from the testing perspective - the whole test can be silently disabled and nobody knew about it.
And from the performance point of view - there is a lot of job done in SetUp, sometimes tens just to get if throw right here.

So it's better to remove `test_skipped` completely and provide only those configs which are 100% compatible with the solver.

test/gtest/codegen_group_conv.hpp
```diff
@@ -0,0 +1,567 @@
...
+        return {0.0f, 3.3f};
+    }
+    else
+    {
+        return {0.0f};
+    }
+}
+
+#define DEFINE_GROUP_CONV_TEST(ndim, type, naming_type, dir, cg)                            \
+    struct GPU_GroupConv##ndim##D_##dir##_##naming_type##_##cg                              \
```
CAHEK7: `type` (`FP16`) must be the last in the test name. Please take a look at https://github.com/ROCm/MIOpen/wiki/GTest-development

CAHEK7: Also you may use kind of `lightweight` version of a new type declaration:
```C++
using GPU_GroupConv##ndim##D_##dir##_##naming_type##_##cg = GroupConvTestFix<ndim, type, Direction::dir, cg>;
```
You don't need extra brackets)

CAHEK7 (changes requested):
It's just a quick review with some simple things and common pitfalls.
A haven't looked deep into the code yet.

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
```diff
@@ -0,0 +1,294 @@
...
+    }
+    if(!problem.Is2d())
+    {
+        return false;
+    }
+    if(!(problem.IsLayoutNHWC() || problem.IsLayoutDefault()))
+    {
+        return false;
+    }
+    return true;
```
bghimireamd, line 189: you will need [ck_white_list ](https://github.com/ROCm/MIOpen/blob/develop/src/solver/conv_ck_igemm_fwd_bias_res_add_activ_fused.cpp#L424) since CK kernel runs only in some hardware.

test/gtest/codegen_group_conv2d_fwd.cpp
```diff
@@ -0,0 +1,33 @@
...
+ *
+ *******************************************************************************/
+#include <gtest/gtest.h>
+
+//#include "group_conv.hpp"
+#include "codegen_group_conv.hpp"
+
+using namespace codegen_group_conv;
+
+DEFINE_CG_GROUP_CONV2D_TEST(half, FP16, Forward);
```
bghimireamd, line 33: only fp16 for now?

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
```diff
@@ -0,0 +1,294 @@
...
+    }
+    if(!problem.IsDirectionForward())
+    {
+        return false;
+    }
+    if(!problem.Is2d())
+    {
+        return false;
+    }
+    if(!(problem.IsLayoutNHWC() || problem.IsLayoutDefault()))
```
bghimireamd, line 181: only nhwc is supported or both nhwc and nchw are supported? Just curious

CMakeLists.txt
```diff
@@ -328,6 +328,8 @@ add_compile_definitions($<$<COMPILE_LANGUAGE:CXX>:HIP_COMPILER_FLAGS=${HIP_COMPI
 # HIP
 if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "HIPNOGPU")
     if(MIOPEN_USE_COMPOSABLEKERNEL)
+	      include(/usr/local/lib/cmake/composable_kernel_host/ck_host_targets.cmake)
+	      find_package(ck_host)
```
bghimireamd: just curious why is ck host needed?

arai713: ck_host is the package that has all the codegen methods in it

test/gtest/codegen_group_conv2d_fwd.cpp
arai713: Yeah, we wanted to get just one example running, then extend with other data types or add new examples after everything looks okay.

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
arai713: This example only supports nhwc

bghimireamd: so can we just do `if(!(problem.IsLayoutNHWC())`

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
bghimireamd: Just realized your [solver currently only servers gfx908](https://github.com/ROCm/MIOpen/blob/f8e3ec06367e1da2e27953fe68bc417793a15c41/src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp#L234) so I think you might need to do if(ctx.GetStream().GetDeviceName() == "gfx908") instead of is_ck_whitelist. CI will fail in other devices.

arai713: That is just the architecture selected, but this works on all the architectures listed in the whitelist, so I think ck_white_list is fine.

requirements.txt
```diff
@@ -7,5 +7,5 @@ nlohmann/json@v3.11.2 -DJSON_MultipleHeaders=ON -DJSON_BuildTests=Off
 ROCm/FunctionalPlus@v0.2.18-p0
 ROCm/eigen@3.4.0
 ROCm/frugally-deep@9683d557eb672ee2304f80f6682c51242d748a50
-ROCm/composable_kernel@fb948120d2d674607e70d0e7587dcb249c0e74c7 -DCMAKE_BUILD_TYPE=Release
+ROCm/composable_kernel@5574422afe4a357e1e5be60cc58ab42645553b93 -DCMAKE_BUILD_TYPE=Release
```
BrianHarrisonAMD: We will want to update this commit to match with CK next time it's promoted to staging I think, and same in the docker file.

CMakeLists.txt
```diff
@@ -328,6 +328,8 @@ add_compile_definitions($<$<COMPILE_LANGUAGE:CXX>:HIP_COMPILER_FLAGS=${HIP_COMPI
 # HIP
 if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "HIPNOGPU")
     if(MIOPEN_USE_COMPOSABLEKERNEL)
+	      include(/usr/local/lib/cmake/composable_kernel_host/ck_host_targets.cmake)
```
BrianHarrisonAMD: Why is ck_host installed into a different location than the rest of the libs?

Also, do we need a separate include call here, or does find_package cover it since we have /usr/local/lib added to the prefix path?

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
```diff
@@ -0,0 +1,298 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2023 Advanced Micro Devices, Inc.
```
BrianHarrisonAMD: Date

src/include/miopen/conv/solvers.hpp
```diff
@@ -4488,6 +4488,26 @@ struct ConvHipImplicitGemmGroupFwdXdlops final
     bool CheckCKApplicability(const miopen::conv::ProblemDescription&) const;
 };
 
+struct ConvHipImplicitGemmGroupFwdXdlopsCodegen final : ConvSolver
+{
+    // TODO: update this fcn
+    const std::string& SolverDbId() const override
+    {
+        return GetSolverDbId<ConvHipImplicitGemmGroupFwdXdlopsCodegen>();
+    }
```
BrianHarrisonAMD: What need to be done here?

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
```diff
@@ -0,0 +1,298 @@
...
+    std::string prologue = "";
+
+    auto x = CKArgs(problem);
+
+    const auto workspace_req = GetWorkspaceSize(ctx, problem);
+
+    auto soln         = ConvSolution{miopenStatusSuccess};
+    soln.workspace_sz = workspace_req;
+
+    auto solution = x.prob.GetSolutions("gfx908", prologue, epilogue);
```
BrianHarrisonAMD, line 234: Won't this need to match the current device's arch?

I think right now it'll be limited to gfx908 instances.

src/CMakeLists.txt
```diff
@@ -851,10 +852,11 @@ endfunction()
 
 target_include_directories(MIOpen PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/include>
+    /usr/local/lib/composable_kernel_host/include
```
BrianHarrisonAMD: This would need to be restricted to only when we are using CK.

Also, if ck_host was in the same location as other libs, we probably don't need this right?

src/include/miopen/conv/solvers.hpp
arai713: This was an old comment, there's no need to change that function

src/solver/codegen_conv_hip_implicit_gemm_2d_grouped_fwd_xdlops.cpp
arai713: This is supported on all devices from the ck whitelist in MIOpen. That function will check whether the given architecture is supported or not. I have been testing on gfx908 and gfx90a.

requirements.txt
arai713: The PR from CK has been merged into develop. Before it gets merged into staging, we'll need to make one last check to make sure that version of CK is still hipRTC compatible. Then I can change the commit hash here to match

CMakeLists.txt
arai713: The ck_host lib is installed into a different location due to conflicting struct names with MIGraphX. We moved it from /opt/rocm/lib/ so they don't inadvertently access the ck_host lib if they didn't mean to.

src/CMakeLists.txt
arai713: ck_host is in another location to prevent potential clashes due to similar structure names with MIGraphX

BradPepersAMD:

PR #3416, opened 2024-12-01
Move tensor_set/scale/cast/copy to gtest

CAHEK7:
This PR moves tensor_set/scale/cast/copy to gtest framework.
It introduces fast and flexible cpu verification routines for unary/binary/ternary operations (@Vsevolod1983 please pay attention to this, it's very useful for ternary tensor_ops conversion)
Those routines are up to 4x faster for unary operations, 3xfaster for binary and almost 2x faster for ternary operations (probably more, it becomes better for bigger tensors, I checked only those sizes what are in the testsuites).

It introduces more `size_t` friendly configs for #3393 (@Vsevolod1983 please take a look)

It also fixes few obvious bugs in the Cast operations.


CAHEK7:
This PR blocks #3424

src/ocl/tensorocl.cpp
```diff
@@ -2135,8 +2135,10 @@ void CastTensor(const Handle& handle,
         MIOPEN_THROW(miopenStatusBadParm, "Tensor dimension sizes unsupported.");
     }
 
+    auto miopen_alpha = *(static_cast<const float*>(alpha));
```
Vsevolod1983, line 2138: Is it possible alpha to be nullptr here ? Or I missed this check ?

test/conv2d_bias.cpp
```diff
@@ -35,7 +35,7 @@ struct conv2d_bias_driver : public conv_bias_driver<T>
                    tensor_elem_gen_checkboard_sign{}(is...);
         };
 
-        this->add(this->output, "output", this->get_tensor(get_inputs, gen_value));
+        this->add(this->output, "output", this->get_tensor(get_inputs<int>, gen_value));
```
Vsevolod1983, line 38: Does this change relate to the purpose of the given PR?

src/ocl/tensorocl.cpp
BrianHarrisonAMD: It looks like it's set everywhere it's used, but also surprising we pass it as a pointer. 
It seems like it could be a float in this case.

test/gtest/binary_tensor_ops.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+    static void SetUpTestSuite()
+    {
+        static constexpr auto dstType = miopen_type<DstType>{};
+        static constexpr auto srcType = miopen_type<SrcType>{};
+
+        uint64_t dstMaxValue = dstType == miopenHalf ? 5 : (dstType == miopenInt8 ? 126 : 32767);
+        uint64_t srcMaxValue = srcType == miopenHalf ? 5 : (srcType == miopenInt8 ? 126 : 32767);
+
+        uint64_t maxValue = std::min(dstMaxValue, srcMaxValue);
```
BrianHarrisonAMD, line 55: For the old tests, it looks like they had different max values being used depending on the mode.

Here are the ones for tensor_copy.cpp for example:
```cpp
               uint64_t max_value = miopen_type<T>{} == miopenHalf   ? 5
                             : miopen_type<T>{} == miopenInt8 ? 127
                                                              : 17;
```

I think it would make sense to keep the max values the same as before for the different tests. 
That being said, it also doesn't seem like it's a huge difference, and is probably okay without it.
Sort of curious why they were like that before.

test/gtest/binary_tensor_ops.cpp
```diff
@@ -0,0 +1,295 @@
...
+        {
+            operate_over_subtensor(
+                [alpha](auto& dst, auto src) { dst = static_cast<float>(src) * alpha; },
+                dstSuperCpu,
+                srcSuperCpu,
+                dstDesc,
+                srcDesc,
+                dstOffset,
+                srcOffset);
+        }
```
BrianHarrisonAMD, line 150: Looks like we gained coverage here.
I think the old tests only checked the clamp case.

Nice!

src/ocl/tensorocl.cpp
```diff
@@ -2146,7 +2148,9 @@ void CastTensor(const Handle& handle,
 
         const std::vector<std::size_t>& lens = srcDesc_flat.GetLengths();
 
-        std::string network_config = "cast " + std::to_string(dstDesc_flat.GetType());
+        // TODO: make proper network config
+        std::string network_config = "cast " + std::to_string(srcDesc_flat.GetType()) +
+                                     std::to_string(dstDesc_flat.GetType());
```
BrianHarrisonAMD, line 2153: Was this wrong before?

Looks like it used to have a caching key based off the dst type only, but uses the src type as part of the compilation params.

test/gtest/binary_tensor_ops.cpp
```diff
@@ -0,0 +1,295 @@
...
+    INSTANTIATE_TEST_SUITE_P(Full,                                                             \
+                             GPU_binaryTensorOps_copy_##TEST_TYPE,                             \
+                             testing::Combine(testing::ValuesIn(get_sub_tensor<size_t>()),     \
+                                              testing::ValuesIn(get_tensor_offsets()),         \
+                                              testing::Values(0.0f),                           \
+                                              testing::Values(false)));
+
+X_INSTANTIATE_COPY(FP32, float);
+X_INSTANTIATE_COPY(FP16, float16);
+X_INSTANTIATE_COPY(BFP16, bfloat16);
```
BrianHarrisonAMD, line 293: Why is there no I8 tests for Copy?

BrianHarrisonAMD (approved):
Looks good to me.
I think this expands the coverage in our CI, and my comments are not blocking.

test/conv2d_bias.cpp
CAHEK7: Yep, it won't compile because it can't inference a type even if there is default one.

test/gtest/binary_tensor_ops.cpp
CAHEK7: The old test did `int` -> `dstType` conversion, so it was focused on dst type only.
Now it is `srcType` -> `dstType` conversion and we must fit into both.
If you are care about `17` vs `32768` - for me if does make sense because the old values were just random numbers and for the operations themselves it doesn't matter, especially for `Copy`. I took it from the `Cast` which is kind of sensitive to the input data and I see not point to change it for `Copy` which can take actually any random garbage.

BrianHarrisonAMD: Yea I agree, and think it's fine to leave it as is.

src/ocl/tensorocl.cpp
CAHEK7: Before - it did not use `srcType` for the network config and failed if you do call `int` -> `float` and then `float` -> `int` conversion (nobody tested mixed calls ever and we had that problem in the other algorithms).
In the both cases it called `int` -> `float` because it has been added first. Now it calls appropriate kernels

src/ocl/tensorocl.cpp
CAHEK7: I just moved it earlier because previously it was even worse - it did CopyTensor for any alpha and since I extended the coverage, I've found out the case and fixed it.
You are right, but I have no time for another review round) Probably I can make some subsequent PR related to that nullptr check (and I8 Copy case).

> It seems like it could be a float in this case.

For all the types except `double` it must be float and for the `double` it must be double, but MIOpen barely supports doubles, almost everywhere... Anyway, it can be done later.

Btw, that's why I asked to add explicit `double` check for all the `isApplicable` methods here https://github.com/ROCm/MIOpen/pull/3346/files#diff-1af35db437d55c5d90c9186b7db9736706d96549c0f20ea3156113db8fdc16e4

src/ocl/tensorocl.cpp
CAHEK7: Btw - that's one of the "gtest single binary" problems - we may stuck withing the first added kernel.

test/gtest/binary_tensor_ops.cpp
CAHEK7: It is forgotten.

src/ocl/tensorocl.cpp
CAHEK7: Proper fix is to include all the compile-time and performance critical parameters into the network config.
In the first case it will choose a properly compiled kernel and in the second one it will choose the most performant one (if there are any). I haven't analyzed it deep and just fixed that part with has been found we the wider coverage.

Vsevolod1983 (approved):

PR #3419, opened 2024-12-03
InlineVector implementation

novakovicdj:
This is PR for initial implementation of `InlineVector` class from issue https://github.com/ROCm/MIOpen/issues/3408, also added some tests for this.

I will do additional performance testing with tensorOp solver structure to check if, and how much, host time is reduced using `InlineVector` instead of `std::vector` for lens/strides in `TensorDescriptor`


src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,251 @@
...
+    const T* begin() const noexcept { return _data.begin(); }
+
+    T* end() noexcept { return (_data.begin() + real_size); }
+
+    const T* end() const noexcept { return (_data.begin() + real_size); }
+
+    // Constant iterator
+    const T* cbegin() const noexcept { return begin(); }
+
+    const T* cend() const noexcept { return end(); }
```
CAHEK7: MSVC is strict about iterators types, please replace `T*` with proper iterators.
You can add typedefs for pointer/reference/iterator and reuse them from `std::array`.
```c++
using storage_type = std::array<T, N>;
using iterator = storage_type::iterator;
...etc...
private:
storage_type _data;
```

CAHEK7 (changes requested):

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,251 @@
...
+#include <miopen/errors.hpp>
+
+namespace miopen {
+
+template <typename T, std::size_t N>
+class InlineVector
+{
+public:
+    using value_type = T;
+    static_assert(std::is_scalar_v<T>, "Input data size is bigger than InlineVector's capacity");
```
DrizztDoUrden: [Suggestion] Strictly speaking, message does not reflect the check done.  I would suggest replacing the message with something like "Inline vector has no support for non-scalar types yet"

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,251 @@
...
+
+    // Default constructor
+    InlineVector() = default;
+
+    // Copy and move constructor
+    InlineVector(const InlineVector& inline_vec)     = default;
+    InlineVector(InlineVector&& inline_vec) noexcept = default;
+
+    InlineVector(std::initializer_list<T> __data) : real_size(__data.size())
```
DrizztDoUrden: Names starting with double underscore or underscore followed by a capital letter are reserved by the implementation. E.g. 17.4.3.1.2 of 2003 standard. I think, most of our code has members with no underscore and arguments having a trailing one when needed.

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,251 @@
...
+
+    std::reverse_iterator<const T*> crend() const noexcept
+    {
+        return std::reverse_iterator<const T*>(cbegin());
+    }
+
+    // Element access
+    T& operator[](std::size_t n) noexcept { return _data[n]; }
+
+    const T& operator[](std::size_t n) const noexcept { return _data[n]; }
```
DrizztDoUrden: I'd add at least an assert with size validation in them
Also a suggestion: add subscripts that accept std::integral_constant<size_t, n> and something like ""_icst user-defined literal, that allows making a compile-time check and most of the code where we want to use this has constant indices anyway.

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,251 @@
...
+    void push_back(const T&& e)
+    {
+        if(real_size == N)
+        {
+            MIOPEN_THROW("InlineVector already full");
+        }
+        _data[real_size++] = std::move(e);
+    }
+
+    // Create element and add it to the back
```
DrizztDoUrden: `emplace_back` is not very useful if you limit data types to scalars

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    // Add element to the back
+    void push_back(const T& e)
+    {
+        if(real_size == N)
+        {
+            MIOPEN_THROW("InlineVector already full");
+        }
+        _data[real_size++] = e;
```
DrizztDoUrden: ~~I would suggest using pre-increment in push_back/emplace_back. It wouldn't change much in terms of how it works except for data races, but the intent would be more clear.~~

DrizztDoUrden (changes requested):

src/include/miopen/inline_vector.hpp
CAHEK7: It mimics `std::` implementation, assert seems a good tradeoff.
As for the compile-time checks - they are possible for the capacity only.

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,261 @@
...
+    {
+        if(real_size > N)
+        {
+            MIOPEN_THROW("Input data size is bigger than InlineVector's capacity");
+        }
+
+        std::copy(data.begin(), data.end(), storage.begin());
+    }
+
+    template <typename _InputIterator, typename = std::_RequireInputIter<_InputIterator>>
```
CAHEK7: `_RequireInputIter` is gcc/clang specific.
I guess that part can be omitted at all:
`template <typename InputIterator>`

BrianHarrisonAMD:
After develop fix, looks like there is this hip tidy error:

[2024-12-03T18:18:41.307Z] test/gtest/inline_vector_basic_ops.cpp:54:1: warning: syntax error [syntaxError]
[2024-12-03T18:18:41.307Z] TEST(CPU_InlineVectorFindIfAndDistance_NONE, Test)
[2024-12-03T18:18:41.307Z] ^

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,261 @@
...
+    {
+        if(real_size > N)
+        {
+            MIOPEN_THROW("Input data size is bigger than InlineVector's capacity");
+        }
+
+        std::copy(data.begin(), data.end(), storage.begin());
+    }
+
+    template <typename _InputIterator>
```
DrizztDoUrden: _InputIterator is reserved by the standard (underscore + uppercase), you can just name it InputIterator or Iterator

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,261 @@
...
+    {
+        if(real_size == N)
+        {
+            MIOPEN_THROW("InlineVector already full");
+        }
+        storage[real_size++] = std::move(e);
+    }
+
+    // Create element and add it to the back
+    template <typename... _Args>
```
DrizztDoUrden: _Args is a name reserved by the standard (underscore + uppercase)

DrizztDoUrden (changes requested):

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+    // Constant reverse iterators
+    const_reverse_iterator crbegin() const noexcept { return const_reverse_iterator(end()); }
+
+    const_reverse_iterator crend() const noexcept { return const_reverse_iterator(begin()); }
+
+    // Element access
+    reference operator[](std::size_t n) noexcept
+    {
+        assert(n < N);
```
DrizztDoUrden: runtime assert should be with real_size

src/include/miopen/inline_vector.hpp
```diff
@@ -0,0 +1,257 @@
...
+        return storage[n];
+    }
+
+    const_reference operator[](std::size_t n) const noexcept
+    {
+        assert(n < N);
```
DrizztDoUrden: ditto

DrizztDoUrden (changes requested):

src/include/miopen/inline_vector.hpp
novakovicdj: I misunderstood you, will change that

src/include/miopen/inline_vector.hpp
novakovicdj: I will remove it and leave comment for it to be implemented when adding support for other data types.

CAHEK7 (approved):

CAHEK7:
@BrianHarrisonAMD @BradPepersAMD @junliume we need this PR for host-side performance improvements (especially around rnn/lstm networks, which is, as I remember, quite critical)
Also it may help to improve testing time a bit.

DrizztDoUrden (approved):

BrianHarrisonAMD:
Starting CI for this.

PR #3423, opened 2024-12-03
Fix hip-tidy build

CAHEK7:


BrianHarrisonAMD (approved):
Looks good to me.
We will need to figure out why this was passing in some tidy checks.

randyspauldingamd (commented):
get 'er done!

PR #3424, opened 2024-12-04
[GTests] Convert TensorOps CTest to GTest

Vsevolod1983:
Test is refactored. 
Three types are supported: float, double and half. For example, I8 and BF16 are not supported even in original CTest.

Test is renamed to ternary_tensor_ops as was requested by @CAHEK7


test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+    miopenTensorOpAdd, miopenTensorOpMul, miopenTensorOpMin, miopenTensorOpMax};
+
+struct TestCase
+{
+    std::vector<int> tensorlens_ac;
+    std::vector<int> tensorlens_b;
+    std::vector<int64_t> offsets;
+    std::vector<int> stride_a;
+    std::vector<int> stride_b;
+    std::vector<int> stride_c;
```
CAHEK7: All of the `int`s have to be `size_t` (except `int64_t`)

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+                                                       {1, 16, 8},
+                                                       {1, 16, 1},
+                                                       {20, 1, 1},
+                                                       {16, 8},
+                                                       {16, 1},
+                                                       {1, 8},
+                                                       {8},
+                                                       {1}};
+
+static std::vector<std::vector<int64_t>> offsetsArr = {
```
CAHEK7: `static` - > `namespace {}`
`int` -> `size_t`

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+struct TensorOpsCommon : public testing::TestWithParam<TestCase>
+{
+    void SetUp() override
+    {
+        prng::reset_seed();
+
+        CreateTensors();
+
+        tensor<T> tensorGPU = std::move(CalculateOnGPU());
+        tensor<T> tensorCPU = std::move(CalculateOnCPU());
```
CAHEK7: 1) `std::move` is not required because of mandatory RVO. RVO guarantees zero cost in-place object creation without even copying or moving anything. 
2) No need to mimic old approach, since it's kind of weird and incredibly inefficient. See #3416

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+                         &testCase.alphabeta[2],
+                         tensorC.desc,
+                         c_dev.get(),
+                         testCase.offsets[0],
+                         testCase.offsets[1],
+                         testCase.offsets[2],
+                         false); // it does not verify non-standard behaviour
+
+        auto r = tensorC;
+        r.data = handle.Read<T>(c_dev, r.data.size());
```
CAHEK7: That's inefficient, you copied `tensorC` and override it right away with `handle.Read`, which... doubles the memory requirements by creating new vector.

See #3416 how it can be done in more efficient way.

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+        else if(testCase.operation == miopenTensorOpMul)
+        {
+            return CalculateOnCPUDataOp([alpha1, alpha2, beta](auto& C, auto A, auto B) {
+                C = A * alpha1 * B * alpha2 + C * beta;
+            });
+        }
+        else if(testCase.operation == miopenTensorOpMin)
+        {
+            return CalculateOnCPUDataOp([alpha1, alpha2, beta](auto& C, auto A, auto B) {
+                C = ((A * alpha1) < B * alpha2 ? A * alpha1 : B * alpha2) + C * beta;
```
CAHEK7:
```suggestion
                C = std::min(A * alpha1, B * alpha2) + C * beta;
```

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+            });
+        }
+        else
+        {
+            return CalculateOnCPUDataOp([alpha1, alpha2, beta](auto& C, auto A, auto B) {
+                C = ((A * alpha1) > B * alpha2 ? A * alpha1 : B * alpha2) + C * beta;
```
CAHEK7:
```suggestion
                C = std::max(A * alpha1, B * alpha2) + C * beta;
```

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    template <typename DataOp>
+    tensor<T> CalculateOnCPUDataOp(DataOp&& dataOp)
+    {
+        const TestCase& testCase = GetParam();
+
+        auto r     = tensorC;
+        auto clens = r.desc.GetLengths();
+        auto blens = tensorB.desc.GetLengths();
```
CAHEK7: It is unused.

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+            });
+        }
+    }
+
+    template <typename DataOp>
+    tensor<T> CalculateOnCPUDataOp(DataOp&& dataOp)
+    {
+        const TestCase& testCase = GetParam();
+
+        auto r     = tensorC;
```
CAHEK7: With high probability that copy is not required, but some cheches are needed to make it true.

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+                           const std::vector<int>& strides,
+                           int offset,
+                           bool isPacked)
+    {
+        uint64_t max_value = miopen_type<T>{} == miopenHalf ? 5 : 17;
+
+        if(!isPacked)
+        {
+            std::vector<int> real_strides(strides.begin() + (strides.size() - lens.size()),
+                                          strides.end());
```
CAHEK7: `std::Vector<size_t>`, see #3393 for the explanation. There is absolutely no point to continue using `int`.

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+        double threshold = std::numeric_limits<T>::epsilon() * tolerance;
+        double error     = miopen::rms_range(tensorCPU.data, tensorGPU.data);
+
+        ASSERT_LE(error, threshold)
+            << "TensorOp: " << testCase.operation << std::endl
+            << "A tensor: " << tensorA.desc.ToString() << std::endl
+            << "B tensor: " << tensorB.desc.ToString() << std::endl
+            << "IsPacked: " << testCase.packed << std::endl
+            << "Offsets: " << testCase.offsets[0] << "," << testCase.offsets[1] << ","
+            << testCase.offsets[2] << std::endl;
```
CAHEK7, line 242: Usually failed test prints its testcase automatically.

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+private:
+    tensor<T> tensorA;
+    tensor<T> tensorB;
+    tensor<T> tensorC;
+};
+
+struct GPU_TensorOps_FP32 : public TensorOpsCommon<float>
+{
+};
```
CAHEK7:
```suggestion
using GPU_TensorOps_FP32 = TensorOpsCommon<float>;
```

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+        }
+    }
+
+    return true;
+}
+
+template <typename T>
+void AddTestCases(std::vector<TestCase>& testCases,
+                  const std::vector<int> tensorALens,
+                  const std::vector<int> tensorBLens)
```
CAHEK7:
```suggestion
                  const std::vector<int>& tensorALens,
                  const std::vector<int>& tensorBLens)
```

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+                                                       {1, 8},
+                                                       {8},
+                                                       {1}};
+
+static std::vector<std::vector<int64_t>> offsetsArr = {
+    {0, 0, 0}, {64, 32, 16}, {32, 16, 32}, {32, 16, 32}};
+
+static std::vector<std::vector<float>> alphabetaArr = {{1, 1, 0}, {-1, 1, 1}, {1.0, 0.5, 0.3}};
+
+static std::vector<std::vector<int>> stridesArr = {{8 * 16 * 20 * 16, 8 * 16 * 20, 8 * 16, 8, 1}};
```
CAHEK7: There is no point of having `vector` of `vectors`. It can be just a vector. There is no point to mimic old inefficient approach.

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+                continue;
+            if(!checkStride(tensorBLens, stride_b))
+                continue;
+            if(!checkStride(tensorALens, stride_c))
+                continue;
+
+            for(const auto& alphabeta : alphabetaArr)
+                for(const auto& operation : operationArr)
+                {
+                    TestCase& testCase = testCases.emplace_back();
```
CAHEK7, line 331: It's better to properly `emplace` everything, passing all the parameters into the `emplace` method.

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+                    testCase.stride_b      = stride_b;
+                    testCase.stride_c      = stride_c;
+                }
+        }
+}
+
+template <typename T>
+inline auto GetCases()
+{
+    static std::vector<TestCase> testCases;
```
CAHEK7:
```suggestion
    static const std::vector<TestCase> testCases = GenCases(); //or use lambda
```

CAHEK7: It has to be type-agnostic - now it creates 3 independent test lists.

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+        {
+            if(!checkTensorsCompatibility(tensorALens, tensorBLens))
+            {
+                continue;
+            }
+
+            AddTestCases<T>(testCases, tensorALens, tensorBLens);
+        }
+
+    return testing::ValuesIn(testCases);
```
CAHEK7: Probably it even should be a `static const auto cases = testing::Values(GenCases());`

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+TEST_P(GPU_TensorOps_FP32, TestFloat) {}
+
+TEST_P(GPU_TensorOps_FP16, TestFloat16) {}
+
+TEST_P(GPU_TensorOps_FP64, TestDouble) {}
+
+INSTANTIATE_TEST_SUITE_P(Smoke, GPU_TensorOps_FP32, GetCases<float>());
+INSTANTIATE_TEST_SUITE_P(Smoke, GPU_TensorOps_FP64, GetCases<double>());
+INSTANTIATE_TEST_SUITE_P(Smoke, GPU_TensorOps_FP16, GetCases<half_float::half>());
```
CAHEK7: It's not `Smoke`, it is `Full`, it has thousands of cases.
Smoke tests must have 1-2 cases and that's it.

I would recommend to change the way how the test cases are generated and do not use ALL the operations for the full tests, since the main point of the full tests with all the different offsets, sizes, etc is to check tensor traversal and broadcasting, while the operation itself it's not important. (as usual, see #3416)

You can have all the operations in the smoke tests, but with limited number of different tensors geometry.

CAHEK7 (changes requested):

test/gtest/ternary_tensor_ops.cpp
Vsevolod1983: It prints, but it is not readable becauseits  in a binary format.

CAHEK7: It can have an `operator<<`.

CAHEK7 (approved):

test/gtest/ternary_tensor_ops.cpp
```diff
@@ -0,0 +1,379 @@
...
+    void SetUp() override
+    {
+        prng::reset_seed();
+
+        CreateTensors();
+
+        std::vector<T> tensorGPUData = CalculateOnGPU();
+        std::vector<T> tensorCPUData = CalculateOnCPU();
+
+        CompareResults(tensorGPUData, tensorCPUData);
```
scerzh: This is the body of the test, it's better to place it into a private function and call it from TEST_P(){}

scerzh (commented):
LGTM but a small comment.

test/gtest/ternary_tensor_ops.cpp
Vsevolod1983: yeah this is a good catch)

scerzh (approved):
LGTM

bghimireamd:
LGTM

bghimireamd (approved):

PR #3425, opened 2024-12-05
Enabled the flexible split_k support in ck-based 3D convolution backward weight solver.

iq136boy:
In this P:
we enabled the flexible split_k support in ck-based convolution backward weight solver. The solver is able to try all the possible combination of CK instances and split_k value to find the one with best performance in exhaustive search mode.
In addition, the ck-based f16f8f16 conv solvers are deprecated since there are no instances support for those solvers from CK side.  
The preliminary manually tuning shows significantly performance uplift (10x more for most of the testing 3D wrw configs).
Next:
@cderb will take the branche and update the perf DB and do the tuning work.


src/CMakeLists.txt
```diff
@@ -266,9 +266,6 @@ set( MIOpen_Source
     solver/conv/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
     solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
     solver/conv/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
-    solver/conv/conv_hip_implicit_gemm_f16f8f16_fwd_xdlops.cpp
-    solver/conv/conv_hip_implicit_gemm_f16f8f16_bwd_xdlops.cpp
-    solver/conv/conv_hip_implicit_gemm_f16f8f16_wrw_xdlops.cpp
```
BrianHarrisonAMD, line 271: Do we need to remove these with this PR?

Might be better to separate this out into a different PR unless it's needed for this work.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -125,6 +193,29 @@ std::vector<std::string> FillValidKernelsIDs(const ProblemDescriptionType& probl
...
+           std::is_same_v<DeviceOpType, conv::DeviceOpGBwdWeightBilinearPtrs<float>> ||
+           std::is_same_v<DeviceOpType, conv::DeviceOpGBwdWeightBilinearPtrs<int8_t>> ||
+           std::is_same_v<DeviceOpType, conv::DeviceOpGBwdWeightBilinearPtrs<ck::bhalf_t>> ||
+           std::is_same_v<DeviceOpType, conv::DeviceOpGBwdWeightScalePtrs<ck::half_t>> ||
+           std::is_same_v<DeviceOpType, conv::DeviceOpGBwdWeightScalePtrs<float>> ||
+           std::is_same_v<DeviceOpType, conv::DeviceOpGBwdWeightScalePtrs<int8_t>> ||
+           std::is_same_v<DeviceOpType, conv::DeviceOpGBwdWeightScalePtrs<ck::bhalf_t>>;
+}
+#endif
+
```
BrianHarrisonAMD, line 218: Might be less error prone to have the internals of the function have the define checks like it's done for "IsCKArgsSupported", and just return false otherwise. 

Right now it wouldn't matter since this method is only called in places already checking these flags, but if it were used elsewhere we would need to add extra guards to ensure the function is only called when it's defined.

src/solver.cpp
```diff
@@ -635,18 +635,6 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
...
-                       conv::ConvHipImplicitGemmF16F8F16FwdXdlops{},
-                       miopenConvolutionAlgoImplicitGEMM);
-    RegisterWithSolver(registry,
-                       ++id,
-                       conv::ConvHipImplicitGemmF16F8F16BwdXdlops{},
-                       miopenConvolutionAlgoImplicitGEMM);
-    RegisterWithSolver(registry,
-                       ++id,
-                       conv::ConvHipImplicitGemmF16F8F16WrwXdlops{},
-                       miopenConvolutionAlgoImplicitGEMM);
```
BrianHarrisonAMD, line 649: What impact does the solver ID have on our caches?
Might need to double check if this will cause problems since it'll shift the ID for all the solvers below.

BrianHarrisonAMD: Yea I thought I remembered seeing this:

    // When solver gets removed its registration line should be replaced with ++id to keep
    // backwards compatibility. New solvers should only be added to the end of list unless it is
    // intended to reuse an id of a removed solver.
    
If we remove these in this PR, then I think we need to keep the ID's still incrementing with a note.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -36,6 +36,8 @@
 #if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
 #include <ck/utility/data_type.hpp>
 #include <ck/library/tensor_operation_instance/gpu/grouped_convolution_backward_weight.hpp>
+#include <ck/library/tensor_operation_instance/gpu/grouped_convolution_backward_weight_bilinear.hpp>
+#include <ck/library/tensor_operation_instance/gpu/grouped_convolution_backward_weight_scale.hpp>
```
BrianHarrisonAMD, line 40: Probably fine for now, but some of these common utilities are starting to be implementation specific, and likely should be moved to the specific solvers, or a file that is specific for that group of solvers.

On the other hand, if we are planning on adding split-k support to all solvers eventually, then this might be unified again and this special logic can be removed.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
```diff
@@ -337,7 +289,7 @@ struct CKArgs
     int Z;
     miopenAlphaBetaCase_t alpha_beta_case;
     miopenDataType_t data_type;
-    ck::index_t split_k = 2;
+    // ck::index_t split_k_fix = 2;
```
BrianHarrisonAMD: Commented out code to remove.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
```diff
@@ -302,16 +240,30 @@ struct CKArgs
...
     template <typename ConvPtr>
     bool IsSupportedBy(const ConvPtr& conv_ptr) const
     {
-        auto arg_ptr = MakeArgPtr(conv_ptr, nullptr, nullptr, nullptr, 1.0f, 0.0f);
-        if(CKWrwRequireWorkspace(G, C, K, data_type, alpha_beta_case))
+        auto arg_ptr = MakeArgPtr(conv_ptr, nullptr, nullptr, nullptr, 1.0f, 0.0f, 1);
+        // Creat dummy workspace to pass the ck IsSupportedArgument check.
+
+        int dummy_var = 1;
+        conv_ptr->SetWorkSpacePointer(arg_ptr.get(), &dummy_var);
```
BrianHarrisonAMD, line 256: Don't need to fix with this review, but I am wondering if when we run there is a path that doesn't do this, and that's why we get the warning messages from CK.

src/solver.cpp
iq136boy: @BrianHarrisonAMD  oh, yes. The ++id needs to be placed. I will fix it.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
iq136boy: OK. I will remove it.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
iq136boy: Yea. Might need to check with CK if the condition is correct. But it will not affect the correct and performance.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
```diff
@@ -433,20 +387,43 @@ void PerformanceConfigHipImplicitGemm3DGroupWrwXdlops::HeuristicInit(
...
+        case miopenBFloat16: Init<ck::bhalf_t>(problem); break;
+        case miopenInt64:
+        case miopenInt32:
+        case miopenFloat8:
+        case miopenBFloat8:
+        case miopenDouble: break;
+        }
         assert(!valid_kernels.empty());
         return true;
     }
```
BrianHarrisonAMD: Should it still call HeuristicInit, but then return false if valid_kernels is empty? 
Or maybe just return false if the valid_kernels are empty?

I think it shouldn't return true if it's empty since it'll keep looping on SetNextValue, but also not sure why HeuristicInit was removed here.

src/CMakeLists.txt
iq136boy: Yes. Otherwise, we need to add the split_k stuff in the gemm_f16f8f16_wrw_xdlops.cpp solver too, which is meaningless since these solvers are totally useless now because the underlying CK instances were deleted from CK already a while ago.

BrianHarrisonAMD: Oh that's interesting.

Alright sounds good.

BrianHarrisonAMD (approved):
I think this looks good, and we can start tuning.

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: Yes. I agree. It is a kind of awkward but we need it because now CK only enable split_k support in the 2d and 3d wrw solvers. If finally we have general split-k support, then we can remove the special logic.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
iq136boy: Yes. Basically the code section I added here does exactly the same in heuristicInit. And you are right, instead of the assertion check, it needs to return false if the valid_kenel is still empty after the initialization.

cderb:
Added db update.

cderb (approved):

BrianHarrisonAMD:
Pushed a fix for broken tests, waiting on CI

PR #3428, opened 2024-12-06
[BUG] Fix memory leak in GraphExecutorFind20

scerzh:
Two issues were fixed:

* A raw `miopenSolution_t` descriptor was used in `GraphExecutorFind20` without calling `miopenDestroySolution()` at the end of `GraphExecutorFind20`'s lifetime.

* A vector of `miopenSolution_t` descriptors was used in an unsafe way causing all of them to leak upon an exception in `getEngines()`.

Just calling `miopenDestroySolution()` isn't enough to organize a proper memory management as it produces dangling pointers if copies were made and duplicating `miopenSolution_t` isn't provided by the lib. `miopenSolution_t` descriptor was replaced with a `Solution`.

For the second issue (in `getEngines()`) ~an `std::unique_ptr` with a custom deleter was utilized~ a `scope_exit` template class was introduced.


Vsevolod1983 (approved):

averinevg (approved):
LGTM

PR #3430, opened 2024-12-10
[GRAPHAPI] Implement MIOPEN_BACKEND_EXECUTION_PLAN_DESCRIPTOR Serialization

### LINKED ISSUE #2843: [MHA] Implement MIOPEN_BACKEND_OPERATIONGRAPH_DESCRIPTOR in Graph API

### LINKED ISSUE #2846: [MHA] Implement MIOPEN_BACKEND_EXECUTION_PLAN_DESCRIPTOR in Graph API
Implement MIOPEN_BACKEND_EXECUTION_PLAN_DESCRIPTOR

- [x] #2922
- [x] #2925
- [x] MIOPEN_BACKEND_EXECUTION_PLAN_DESCRIPTOR Serialization

scerzh:
closes #2846

and depends on #3428 which had not been merged before this PR was created

The PR consists of two parts:

* Serialization of `GraphExecutorFind20`
* Serialization of `ConvBiasResAddActivForwardExecutor`


DrizztDoUrden (approved):

src/graphapi/engine.cpp
```diff
@@ -36,6 +43,179 @@ GraphPatternExecutor::~GraphPatternExecutor() = default;
 
 size_t GraphExecutorFind20::getWorkspaceSize() const { return mSolution.GetWorkspaceSize(); }
 
+namespace {
+
+const std::string_view base64Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                                     "abcdefghijklmnopqrstuvwxyz"
+                                     "0123456789+/";
+
+std::string base64Encode(const char* data, size_t length)
```
Vsevolod1983: Should we move base64 encode/decode to utils of any kind? Looks like it might be commonly used routine...

Vsevolod1983 (approved):

src/graphapi/engine.cpp
CAHEK7: Why are the pointer and the size unbundled here? Can we pass `vector<char>`
Especially if it's inside an anonymous namespace and for a single specific case in that PR.

But probably for a common case, like Vsevolod mentioned, it's worth to have those fields separated.

UPD: you just wanted more convenient `length` usage.

src/graphapi/engine.cpp
```diff
@@ -36,6 +43,179 @@ GraphPatternExecutor::~GraphPatternExecutor() = default;
...
+        while(i++ < 3)
+        {
+            encodedString += '=';
+        }
+    }
+
+    return encodedString;
+}
+
+std::string base64Decode(const std::string& encodedString)
```
CAHEK7: Returning binary blob as std::string may have some problems on Windows, I guess @apwojcik has already fixed a few cases and replaced them with a std::vector<char>.

But probably it was related to a serialization of such strings via `ofstream`. Could you double check it?

scerzh: Agree. `std::vector` would be better here, regardless of whether `std::string` has issues with Windows. Even better `std::vector<uint8_t>`, though requires a `reinterpret_cast` for `miopenLoadSolution` and `miopenSaveSolution`

src/graphapi/engine.cpp
scerzh: Among possible options:
```C++
std::string base64Encode(const uint8_t* begin, const uint8_t* end)
std::string base64Encode(const uint8_t* begin, size_t size)
std::string base64Encode(const std::vector<uint8_t>& data)
```
I chose that one, but probably better to provide all of them.

PR #3431, opened 2024-12-11
[Conv] add ck bfp16 for fwd and bwd

bghimireamd:


BradPepersAMD (commented):
Looks good and I think important to ensure we have BF16 support on CK kernels.  I'd like someone else with more knowledge of what is involved in enabling this to approve as well before this is merged.

BradPepersAMD:
What about WRW?  Is it already enabled for BF16?

BrianHarrisonAMD:
We only have a grouped WRW solver, and it has BF16 enabled already.

JonathanLichtnerAMD:
This was submitted via PR https://github.com/ROCm/MIOpen/pull/3521

PR #3432, opened 2024-12-11
[GTest] For CPU_Dump_* tests try to write temporary file into temp folder instead of into current directory

Vsevolod1983:
Partial fix for SWDEV-488434 ( Should fix CPU_Dump_* failures)


scerzh (approved):
LGTM

test/gtest/dumpTensorTest.cpp
```diff
@@ -142,26 +142,27 @@ void testDumpWithNan(const std::string& test_file_name)
     fs::remove(test_file_name);
 }
 
-TEST(CPU_Dump_FP32, testDump) { testDump<float>(test_file_name_prefix + "float.bin"); }
-
-TEST(CPU_Dump_NAN_FP32, testDump)
+namespace {
+std::string GetFullFileName(const std::string& filename)
```
CAHEK7, line 146:
```suggestion
std::string GetFullFileName(const std::string_view& filename)
```

CAHEK7 (approved):

PR #3433, opened 2024-12-11
Perf test enabling

alexandraBara:


.jenkins/perf_test
```diff
@@ -0,0 +1,731 @@
...
+            }
+        }
+        /*stage("Packages") {
+            parallel {
+                stage("HIP Package") {
+                    agent{ node { label 'nogpu'} }
+                    steps{
+                        script {
+                        echo "Packages Stage"
+                        def utils = load "vars/utils.groovy"
```
alexandraBara, line 52: @cderb can I do this somewhere else, do i dont have to call load before any call to utils.something?

cderb: You should be able to include a reference to the utils.groovy at the head of the file:
https://github.com/ROCm/MITuna/blob/8beff11d8a8b890dc8080591192f4732d3d84d99/.jenkins/perf_eval#L1-L8

PR #3437, opened 2024-12-13
Fix kernel cache threading

BrianHarrisonAMD:
Fix kernel cache issue when running tuning.
Due to CompileAgents running during tuning there is the main thread, and the compile agent thread accessing the kernel cache.
Adding locking to the kernel cache prevents the critical section.


DrizztDoUrden (approved):

BradPepersAMD (approved):
LGTM

PR #3440, opened 2024-12-17
Fix incorrect casting of ASM solvers when using bf16 datatype

BrianHarrisonAMD:
With the #3403 fix to clear output buffers, there was a bug introduced that was incorrectly casting tensors. 

Going to add tests to catch this when I get a chance.

### DIFF
--- a/src/conv/invokers/impl_gemm_dynamic.cpp
+++ b/src/conv/invokers/impl_gemm_dynamic.cpp
@@ -178,7 +178,6 @@ InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ProblemDescri
         const int gemm_k_gid = k * y_dot_slice_gid[gemm_id] * x_dot_slice_gid[gemm_id];
         is_gemm_not_empty.emplace_back(gemm_k_gid > 0);
     }
-    bool need_set_zero = true;
 
     return [=](const std::vector<Kernel>& kernels) {
         const auto kernel = kernels[0];
@@ -186,14 +185,13 @@ InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ProblemDescri
             decltype(auto) data_ctx = primitive_parameters.CastTo<conv::DataInvokeParams>();
             const auto& tensors     = data_ctx.tensors;
             float elapsed           = 0;
-            if(need_set_zero)
-            {
-                float zero = 0.f;
-                SetTensor(handle, tensors.outDesc, tensors.out, &zero);
 
-                if(handle.IsProfilingEnabled())
-                    elapsed += handle.GetKernelTime();
-            }
+            float zero = 0.f;
+            SetTensor(handle, tensors.outDesc, tensors.out, &zero);
+
+            if(handle.IsProfilingEnabled())
+                elapsed += handle.GetKernelTime();
+
             for(int gemm_id = 0; gemm_id < num_of_gemms; gemm_id++)
             {
                 if(is_gemm_not_empty[gemm_id])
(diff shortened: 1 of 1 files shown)


src/conv/invokers/impl_gemm_dynamic.cpp
```diff
@@ -178,22 +178,20 @@ InvokerFactory MakeImplGemmDynamicBackwardDataInvokerFactory(const ProblemDescri
         const int gemm_k_gid = k * y_dot_slice_gid[gemm_id] * x_dot_slice_gid[gemm_id];
         is_gemm_not_empty.emplace_back(gemm_k_gid > 0);
     }
-    bool need_set_zero = true;
```
randyspauldingamd, line 181: `SetTensor` is rather heavyweight even after the kernel has been built, so we don't want to delete `need_set_zero`. Setting it true was just a temporary kludge since they didn't have time to work out the logic at the time. There's supposed to be a ticket to fix it. Here's the kludge: https://github.com/ROCm/MIOpen/pull/3403  @iq136boy @xinlipn

randyspauldingamd (changes requested):

src/conv/invokers/impl_gemm_dynamic.cpp
BrianHarrisonAMD: I can wrap it in a workaround define like we do in other cases to keep the original if we are planning on coming back to this.

I also think I am going to swap to [this way of clearing](https://github.com/ROCm/MIOpen/blob/develop/src/include/miopen/solver/implicitgemm_ck_util.hpp#L744C1-L764C2) instead since I saw hipMemsetAsync giving better performance when it was an option.

randyspauldingamd: Sounds good!

I would've suggested ZeroOutTensor, but capturing the zero-set kernel time is desirable. If we have a canned solution for capturing async execution time, I don't know about it.

In this case it would be best to modify the kernel(s) to ensure every pixel is touched. Those sort of ops are often free due to being shadowed by global I/O latency.

BrianHarrisonAMD: We have the [HipEventProfiler](https://github.com/ROCm/MIOpen/blob/9b4cf56919c85aae26a3eb9206b16dd9506c171b/src/include/miopen/hipoc_kernel.hpp#L50-L58) RAII wrapper that can do this.

I think ideally we would use this everywhere, and remove the automatic accumulation logic that happens in the handle, but that is a large change.

randyspauldingamd: Oh, interesting. I'm not sure it's real useful as-is--we would need explicit control of when `ResetKernelTime` is called and it needs to not block in the dtor, but it's a good starting point.

BrianHarrisonAMD: HipEventProfiler will make a start event when constructed if profiling is enabled (nothing otherwise), and then a stop event when destructed if profiling is enabled, and then accumulate the difference between start / stop into the profiling result on the handle. Any ResetKernelTime calls in between should be removed, but also will not impact the overall result, as the start / stop events will capture the full profiling time within that scope, reset the profiling time, and then set the result to the difference between start / stop.

I think this isn't perfect, since the handle is also already capturing profile timing and synchronizing itself when it needs to, but HipEventProfiler ensures we capture all the kernel time used within a scope, and also prevents issues with the manual accumulation we do right now. I think It would be better if we only had the start / stop events being synchronized on, and used to capture the profiling results, but that is too big of a change for right now. 

Is there an issue with the above that I am missing? 
I think as is, HipEventProfiler is safe & correct for capturing the profiling results.

PR #3444, opened 2024-12-18
[gTest] Add more test coverage for Batchnorm Forward Training

xinlipn:
More test cases added for better coverage. Here's the summary
1. N C H W: 768 1 14 14
Covers: 
forward_spatial_single.cpp: variant == 0, variant == 1
[768 1 14 14.txt](https://github.com/user-attachments/files/18177130/768.1.14.14.txt)

2. N C H W: 768 1 23 23
Covers: 
forward_spatial_single.cpp: variant == 1 and IsApplicable()
forward_spatial_multiple.cpp: variant == 2 (1st)
[768 1 23 23.txt](https://github.com/user-attachments/files/18177132/768.1.23.23.txt)

3. N C H W: 832 1 14 14
Covers: 
forward_spatial_single.cpp: variant == 1 and IsApplicable()
forward_spatial_multiple.cpp: variant == 0  then variant == 2 (2nd)
[832 1 14 14.txt](https://github.com/user-attachments/files/18177135/832.1.14.14.txt)

4. N C H W: 832 1 28 28
Covers: 
forward_spatial_single.cpp: variant == 1 and IsApplicable()
forward_spatial_multiple.cpp: variant ==  2 (1st)  then variant == 2 (2nd)
[832 1 28 28.txt](https://github.com/user-attachments/files/18177146/832.1.28.28.txt)

A few questions/observations:
1. Should NetworkLarge be used for GPU_BN_FWD_Train_Large_FP64?
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/test/gtest/bn_fwd_train.cpp#L147

2. This blurb in IsApplicable 
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/src/solver/batchnorm/forward_spatial_single.cpp#L75-L80
is the same as in GetSolution
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/src/solver/batchnorm/forward_spatial_single.cpp#L159-L161
Which means this snippet will never be run
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/src/solver/batchnorm/forward_spatial_single.cpp#L169-L179

3. Likewise, these two lines are identical 
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/src/solver/batchnorm/forward_spatial_single.cpp#L83
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/src/solver/batchnorm/forward_spatial_single.cpp#L182

Due to 2 and 3 above, variant will never be set to 2 in forward_spatial_single.cpp

4. When this condition is met in forward_spatial_single.cpp, IsApplicable returns false
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/src/solver/batchnorm/forward_spatial_single.cpp#L75-L80
And forward_spatial_multiple.cpp will be called. In its GetSolution, this snippet will always be skipped due to the fact it's false as above
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/src/solver/batchnorm/forward_spatial_multiple.cpp#L123-L132

Therefore, variant will never be 0 or 1 for forward_spatial_multiple.cpp

Please note the cases that do not pass IsApplicable() in forward_spatial_single.cpp will be handled in IsApplicable() in forward_spatial_multiple.cpp
https://github.com/ROCm/MIOpen/blob/310192d5f6ede7a788199365477efe5fe3dd5ffd/src/solver/batchnorm/forward_spatial_multiple.cpp#L53


reidkwja (approved):

test/gtest/bn_fwd_train.cpp
```diff
@@ -144,7 +144,7 @@ INSTANTIATE_TEST_SUITE_P(Smoke,
 
 INSTANTIATE_TEST_SUITE_P(Smoke,
                          GPU_BN_FWD_Train_Large_FP64,
-                         testing::Combine(testing::ValuesIn(NetworkSmall<BNTestCase>()),
+                         testing::Combine(testing::ValuesIn(NetworkLarge<BNTestCase>()),
```
CAHEK7, line 147: Shall it be a new `Full` instance? Let's keep `Smoke` tests relatively small, that's the point.

xinlipn: Hi @CAHEK7  could you clarify what you mean by "Full" instance? For each data type, there's 4 sets of test data for NetworkSmall and 32 for NetworkLarge.

bghimireamd: This is not a new Full instance. Its just there was a typo before. It should be calling NetworkLarge for testing AP2. For AP1 we only test small bn network (4 test).

bghimireamd (approved):

test/gtest/bn_fwd_train.cpp
CAHEK7: I mean `Full` tests and `Smoke` tests - `Smoke` is a small and fast subset which checks only a basic functionality and must not have any long-running tests or large configs.
The main point of `Smoke` tests and a quick sanity check before the long runs.

That's why I'm kind of recommend to move those tests into the `Full` tests, since it uses large configs.

PR #3446, opened 2024-12-19
[GTests] Rename non-unique test names (TEST, TEST_P, TEST_F, TYPED_TEST). Part 1.

Vsevolod1983:
When we use single test binary we should keep our test names unique (First parameter in TEST* macros).  Otherwise, we get a lot of failures, which are not related to tests themselves.  Non unique names create a conflict when get combined into one executable.

As an example of this extra failures from the ticket: SWDEV-488434
[  FAILED  ] Full/GPU_UnitTestConvSolverBwd_FP16.ConvWinoFuryRxSf2x3/0, where GetParam() = (4-byte object <FF-FF FF-FF>, 0, (x:{2,64,128,128,128} w:{32,64,3,3,3} pad:{1,1,1} stride:{1,1,1} dilation:{1,1,1} type_x:0 type_w:0 type_y:0))

We can see that "ConvWinoFuryRxSf2x3" is never used in "Full" test suites. So we have reports about the failures in tests which even don't exist!

There will be series of PRs with renaming for easier reviewing.

### DIFF
--- a/test/gtest/conv_igemm_mlir_bwd_wrw.cpp
+++ b/test/gtest/conv_igemm_mlir_bwd_wrw.cpp
@@ -80,18 +80,18 @@ bool IsTestSupportedForDevice()
 
 } // namespace
 
-class GPU_Conv2dDefaultMLIRTest_FP32 : public FloatTestCase<std::vector<TestCase>>
+class GPU_Conv2dIGemmBwdWrwMLIRTest_FP32 : public FloatTestCase<std::vector<TestCase>>
 {
 };
-class GPU_Conv2dDefaultMLIRTest_FP16 : public HalfTestCase<std::vector<TestCase>>
+class GPU_Conv2dIGemmBwdWrwMLIRTest_FP16 : public HalfTestCase<std::vector<TestCase>>
 {
 };
 
-TEST_P(GPU_Conv2dDefaultMLIRTest_FP32, FloatTest_conv_igemm_mlir_bwd_wrw)
+TEST_P(GPU_Conv2dIGemmBwdWrwMLIRTest_FP32, FloatTest_conv_igemm_mlir_bwd_wrw)
 {
     if(IsTestSupportedForDevice())
     {
-        invoke_with_params<conv2d_driver, GPU_Conv2dDefaultMLIRTest_FP32>(db_check);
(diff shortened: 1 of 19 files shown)


scerzh (approved):

junliume (approved):

averinevg (changes requested):
Great job :smile:. Since not too many changes have been made yet, it’s worth making some adjustments to the naming. I think we should stick to the following naming scheme: **UnitTestConvSolver**+**_SolverName_**+**_Direction_**. The **UnitTestConvSolver** part is necessary so that we can use a filter to run only tests for convolution solvers. The second part is just the full name of the solver. If the solver's name contains the words 'Conv' or 'Solver', they can be excluded from the name. The third part is the direction. If the solver’s name ends with a direction, this part can be omitted. Full name for the `ConvOclDirectFwd` will end with two directions (part of the solver’s name and the actual direction) for some tests, since this solver also works with the backward direction. What is in this pull request is generally good. All of the above applies only to tests starting with `unit_conv_solver_*`.

Vsevolod1983:
@averinevg 
Actually, I have the second branch with a lot of stuff already renamed.
Schema that I used: "UnitTestConv" + Name + "Solver" + Direction
Can we consider it for our naming ?

Vsevolod1983:
> Great job 😄. Since not too many changes have been made yet, it’s worth making some adjustments to the naming. I think we should stick to the following naming scheme: **UnitTestConvSolver**+**_SolverName_**+**_Direction_**. The **UnitTestConvSolver** part is necessary so that we can use a filter to run only tests for convolution solvers. The second part is just the full name of the solver. If the solver's name contains the words 'Conv' or 'Solver', they can be excluded from the name. The third part is the direction. If the solver’s name ends with a direction, this part can be omitted. Full name for the `ConvOclDirectFwd` will end with two directions (part of the solver’s name and the actual direction) for some tests, since this solver also works with the backward direction. What is in this pull request is generally good. All of the above applies only to tests starting with `unit_conv_solver_*`.

Ok, I'll rename

Vsevolod1983:
@averinevg 
Done

averinevg (approved):
LGTM

BradPepersAMD (approved):
Looks good!

PR #3449, opened 2024-12-25
[GTests] Rename non-unique test names (TEST, TEST_P, TEST_F, TYPED_TEST). Part 2.

Vsevolod1983:
When we use single test binary we should keep our test names unique (First parameter in TEST* macros). Otherwise, we get a lot of failures, which are not related to tests themselves. Non unique names create a conflict when get combined into one executable.

As an example of this extra failures from the ticket: SWDEV-488434
[ FAILED ] Full/GPU_UnitTestConvSolverBwd_FP16.ConvWinoFuryRxSf2x3/0, where GetParam() = (4-byte object , 0, (x:{2,64,128,128,128} w:{32,64,3,3,3} pad:{1,1,1} stride:{1,1,1} dilation:{1,1,1} type_x:0 type_w:0 type_y:0))

We can see that "ConvWinoFuryRxSf2x3" is never used in "Full" test suites. So we have reports about the failures in tests which even don't exist!

There will be series of PRs with renaming for easier reviewing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -114,7 +114,7 @@ DEBIAN_FRONTEND=noninteractive apt-get purge -y --allow-unauthenticated \
     miopen-hip
 
 # TODO: it should be able to automatically get commit hash from requirements.txt
-ARG CK_COMMIT=05e8a4c88e6f236a72ca6e33154c5ae711a4dace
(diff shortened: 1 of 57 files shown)


scerzh (approved):

averinevg (approved):
LGTM

PR #3460, opened 2025-01-03
Fix tuning timing results

BrianHarrisonAMD:
Tuning timing results for CK solvers were not timing the same as other solvers which lead to differences in results. 
This is due to timing based off a scoped start/stop event which includes the CPU timing, and also includes the other start/stop events synchronizing which gives improper timing results (leading to higher overall runtime compared to other solvers).

There is also an additional issue where only the fastest solver within a family of solvers would record it's timing values. 
The combination of these issues made it so the MISA solvers were being selected, and hiding the other potential CK solvers.


DrizztDoUrden (approved):

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -485,6 +485,8 @@ class TransposeInstance
 
     void ZeroOutBuffer(const Handle& handle)
     {
+        WorkAroundHipEventProfiler pfr(handle);
+
```
BrianHarrisonAMD: This probably should just use the HipEventProfiler since it's not related to the reason the work around was added, (and should stay if we fix the underlying issue).

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -749,6 +751,8 @@ ZeroOutTensor(const Handle& handle, const TensorDescriptor& tensorDesc, Data_t t
     // Use faster clear if possible.
     if(tensorDesc.IsPacked())
     {
+        WorkAroundHipEventProfiler pfr(handle);
+
```
BrianHarrisonAMD: Same here.

iq136boy (approved):

PR #3461, opened 2025-01-03
Fix tensor casting happening in all cases for BF16 & FP16 WRW

BrianHarrisonAMD:
Tensor casting was happening in all cases, but should only be happening when a split-k kernel was being used.

### DIFF
--- a/test/gtest/unit_conv_solver.hpp
+++ b/test/gtest/unit_conv_solver.hpp
@@ -95,11 +95,13 @@ struct UnitTestConvSolverParams
     void UseCpuRef();
     void EnableDeprecatedSolvers();
     void Tunable(std::size_t iterations_max);
+    void DisableXnack();
 
     Gpu supported_devs;
     bool use_cpu_ref;
     bool enable_deprecated_solvers;
     bool tunable;
+    bool disable_xnack;
     std::size_t tuning_iterations_max;
 };
 
--- a/test/gtest/unit_conv_solver.hpp
+++ b/test/gtest/unit_conv_solver.hpp
@@ -95,7 +95,7 @@ struct UnitTestConvSolverParams
     void UseCpuRef();
     void EnableDeprecatedSolvers();
     void Tunable(std::size_t iterations_max);
-    void DisableXnack();
+    void CheckXnackDisabled();
 
     Gpu supported_devs;
     bool use_cpu_ref;
--- a/test/gtest/unit_conv_solver_ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC.cpp
+++ b/test/gtest/unit_conv_solver_ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC.cpp
@@ -84,34 +84,34 @@ auto GetFullTestParams(miopenDataType_t datatype)
 
 } // namespace
 
-using GPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP16 =
+using GPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP16 =
     GPU_UnitTestConvSolverBwd_FP16;
-using GPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_BFP16 =
+using GPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_BFP16 =
     GPU_UnitTestConvSolverBwd_BFP16;
-using GPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP32 =
+using GPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP32 =
     GPU_UnitTestConvSolverBwd_FP32;
-using CPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCDevApplicabilityBwd_NONE =
+using CPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCDevApplicabilityBwd_NONE =
     CPU_UnitTestConvSolverDevApplicabilityBwd_NONE;
 
-TEST_P(GPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP16,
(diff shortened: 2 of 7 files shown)


test/gtest/unit_conv_solver_ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC.cpp
averinevg, line 1: It would be great to find all the unit tests for this solver, add them to this file (or leave them out if there’s already a test for a specific functionality here), and then remove the old test. This was my idea when I was developing these tests – all unit tests for a specific solver should be in one file. This applies to the other solvers in this pull request as well.

test/gtest/unit_conv_solver_ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+auto GetSmokeTestParams(miopenDataType_t datatype)
+{
+    Gpu supportedDevices = Gpu::gfx90A | Gpu::gfx94X;
+    if(datatype != miopenBFloat16)
+    {
+        supportedDevices = supportedDevices | Gpu::gfx908;
+    }
+    auto testParams = miopen::unit_tests::UnitTestConvSolverParams(supportedDevices);
+    testParams.Tunable(1);
```
averinevg, line 50: We can probably set this parameter to 5 and use single function.
```suggestion
    testParams.Tunable(5);
```

test/gtest/unit_conv_solver_ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    return std::vector{
+        // clang-format off
+        TestCase{{16, 5, 225, 225}, {64, 5, 3, 3}, {0, 0}, {1, 1}, {1, 1}, datatype},
+        TestCase{{16, 576, 1, 1}, {576, 576, 1, 1}, {0, 0}, {1, 1}, {1, 1}, datatype},
+        TestCase{{16, 2048, 8, 32}, {4096, 2048, 1, 1}, {0, 0}, {1, 1}, {1, 1}, datatype},
+        TestCase{{16, 2048, 8, 32}, {2048, 2048, 1, 1}, {0, 0}, {1, 1}, {1, 1}, datatype},
+        TestCase{{16, 2048, 48, 32}, {2048, 2048, 3, 1}, {0, 0}, {3, 1}, {1, 1}, datatype},
+        TestCase{{16, 2048, 1, 512}, {2048, 2048, 1, 2}, {0, 0}, {1, 2}, {1, 1}, datatype},
+        TestCase{{16, 2048, 1, 30}, {576, 2048, 1, 1}, {0, 0}, {1, 1}, {1, 1}, datatype},
```
averinevg, line 70: Here, we should specify why these test cases are needed (if they are regression tests, a reference to the regression is required; if they are network layers, specify for which network).

test/gtest/unit_conv_solver_ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC.cpp
```diff
@@ -0,0 +1,164 @@
...
+} // namespace
+
+using GPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP16 =
+    GPU_UnitTestConvSolverBwd_FP16;
+using GPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_BFP16 =
+    GPU_UnitTestConvSolverBwd_BFP16;
+using GPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP32 =
+    GPU_UnitTestConvSolverBwd_FP32;
+using CPU_UnitTestConvSolverConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWCDevApplicabilityBwd_NONE =
+    CPU_UnitTestConvSolverDevApplicabilityBwd_NONE;
```
averinevg: More information about naming is [here](https://github.com/ROCm/MIOpen/pull/3446#pullrequestreview-2520943560).
```suggestion
using GPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP16 =
    GPU_UnitTestConvSolverBwd_FP16;
using GPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_BFP16 =
    GPU_UnitTestConvSolverBwd_BFP16;
using GPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP32 =
    GPU_UnitTestConvSolverBwd_FP32;
using CPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCDevApplicabilityBwd_NONE =
    CPU_UnitTestConvSolverDevApplicabilityBwd_NONE;
```

averinevg (changes requested):

test/gtest/unit_conv_solver_ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC.cpp
BrianHarrisonAMD: In this case I need the full tests to do an exhaustive tuning to find the instance that was causing issues. 
I split it up since I didn't want the smoke tests checking all instances. 

If there is a better way to control instance selection to force a replication, then I would be interested though!

test/gtest/unit_conv_solver_ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC.cpp
BrianHarrisonAMD: Discussed, and made a follow up ticket for now.

test/gtest/unit_conv_solver.hpp
```diff
@@ -95,11 +95,13 @@ struct UnitTestConvSolverParams
     void UseCpuRef();
     void EnableDeprecatedSolvers();
     void Tunable(std::size_t iterations_max);
+    void DisableXnack();
```
averinevg: I think it would be better to name it `CheckXnackDisabled()`, as here we are not turning anything off, but simply checking.

averinevg (approved):
LGTM

test/gtest/unit_conv_solver.hpp
BrianHarrisonAMD: Ah very good point. Ill change that.

BradPepersAMD (approved):

PR #3469, opened 2025-01-13
Catch exception for PredictSolver in GetSolutionsFallback

cderb:
Allows for continued execution if heuristics fails to find the metadata file.


src/ocl/convolutionocl.cpp
```diff
@@ -610,7 +610,8 @@ ConvolutionDescriptor::GetSolutionsFallback(const ExecutionContext& ctx,
         }
         catch(const miopen::Exception& ex)
         {
-            MIOPEN_LOG_I2("[Warning] Caught exception: (" << ex.what() << "), passing empty solver vector");
+            MIOPEN_LOG_I2("[Warning] Caught exception: (" << ex.what()
+                                                          << "), passing empty solver vector");
```
BrianHarrisonAMD, line 614: With this change in place, do the individual solvers no longer have issues with trying to load the model files / failing?

BrianHarrisonAMD:
I imagine it's very difficult, but is it possible for us to add testing coverage for these changes?

src/ocl/convolutionocl.cpp
cderb: This catches the load error for the missing metadata file when heuristics is invoked. Which was only occurring during this fallback solutions function.

BrianHarrisonAMD:
Looks good to me, but I think it would be worth checking if it's possible to add tests to ensure we maintain this behavior.

BrianHarrisonAMD (approved):

PR #3470, opened 2025-01-14
[RNN] Dynamic algo optimization

shurale-nkn:
changing the algorithm for decomposing seq_len into several pow 2 numbers.

average % of total diff at rnnt_ootb less than 2%.
( TotalTime=fwd+bwd+wrw; diff%=(new_TotalTime - old_TotalTime)/old_TotalTime)

image
image

image

image
image


DrizztDoUrden (approved):
LGTM

PR #3476, opened 2025-01-22
Move miopenSetConvolutionFindMode to the public API

BrianHarrisonAMD:
Moving miopenSetConvolutionFindMode to the public API, and updating MIOpen version to 3.4.

This change is to enable tuning through API calls by setting the FindMode to normal, and turning on exhaustive search.
This matches the behavior when you run miopenFindConvolutionForwardAlgorithm, miopenFindConvolutionBackwardDataAlgorithm, or miopenFindConvolutionBackwardWeightsAlgorithm with MIOPEN_FIND_MODE=1  MIOPEN_FIND_ENFORCE=3.


include/miopen/miopen.h
```diff
@@ -645,6 +645,36 @@ typedef enum
...
+ * utilize the Immediate mode fallback. If Start-up times are expected to be faster, but worse GPU
+ * performance.
+ *
+ * * Hybrid: Checks the Find-db for an entry. If there is a hit, use that entry. If there is a miss,
+ * use the existing Find machinery. Slower start-up times than Fast Find, but no GPU performance
+ * drop.
+ *
+ * * Dynamic Hybrid: Checks the Find-db for an entry. If there is a hit, uses that entry. If there
+ * is a miss, uses the existing Find machinery with skipping non-dynamic kernels, thus saving
+ * compilation time.slow-compiling kernels. Faster start-up times than Hybrid Find, but GPU
```
jeffdaily: Is that a typo?

jeffdaily (approved):

include/miopen/miopen.h
BrianHarrisonAMD: Yea this all looks slightly off.
I'll fix it up.

Thanks!

include/miopen/miopen.h
```diff
@@ -645,6 +645,32 @@ typedef enum
...
+ * miopenFindConvolutionBackwardWeightsAlgorithm().
+ */
+typedef enum
+{
+    miopenConvolutionFindModeNormal =
+        1, /*!< Full Find mode call, which will benchmark all the solvers and return a list. >*/
+    miopenConvolutionFindModeFast =
+        2, /*!< Checks the Find-db for an entry. If there is a hit, use that entry. If there is a
+              miss, utilize the Immediate mode fallback. Start-up times are expected to be faster,
+              but worse GPU performance. >*/
```
amd-jnovotny:
```suggestion
              but with worse GPU performance. >*/
```

include/miopen/miopen.h
```diff
@@ -645,6 +645,32 @@ typedef enum
...
+    miopenConvolutionFindModeHybrid =
+        3, /*!< Checks the Find-db for an entry. If there is a hit, use that entry. If there is a
+              miss, use the existing Find machinery. Slower start-up times than Fast Find, but
+              better GPU performance. >*/
+    // miopenConvolutionFindModeReserved_4 = 4, /*!< Reserved - do not use */
+    miopenConvolutionFindModeDynamicHybrid =
+        5, /*!< Checks the Find-db for an entry. If there is a hit, uses that entry. If there is a
+              miss, uses the existing Find machinery with skipping non-dynamic kernels, thus saving
+              compilation time. Faster start-up times than Hybrid Find, but GPU performance may be a
```
amd-jnovotny:
```suggestion
              compilation time. Faster start-up times than Hybrid Find, but GPU performance might be a
```

amd-jnovotny (approved):
Looks good. I've made a couple of very minor editorial suggestions for the API documentation.

DrizztDoUrden (approved):

cderb (approved):
lgtm

include/miopen/miopen.h
```diff
@@ -1169,6 +1195,31 @@ MIOPEN_EXPORT miopenStatus_t miopenGetConvolutionAttribute(miopenConvolutionDesc
...
                                                            int* value);
 
+/*! @brief Sets the Find Mode attribute in the convolution descriptor.
+ *
+ * The subsequent calls of miopenFindConvolutionForwardAlgorithm(),
+ * miopenFindConvolutionBackwardDataAlgorithm(), or miopenFindConvolutionBackwardWeightsAlgorithm()
+ * invoked with convDesc, will follow the findMode set by this call.
+ *
+ * Note that the default Find Mode is overriden by the MIOPEN_FIND_MODE environment variable,
+ * if it is set. If unset, the default is as specified by miopenConvolutionFindModeDefault.
```
randyspauldingamd, line 1205: I don't see `miopenConvolutionFindModeDefault` defined anywhere. It's only assigned by CMake via MIOPEN_DEFAULT_FIND_MODE, which is super kludgy design at best.

Also, `FindMode` in `find_controls.hpp` duplicates these. We should assign the values there using the values in miopenConvolutionFindMode_t.

It's obviously too late to fix these things here, but we should capture them in a ticket.

test/gtest/conv_api_find_mode.cpp
```diff
@@ -0,0 +1,57 @@
...
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <gtest/gtest.h>
+#include <miopen/miopen.h>
+
+#define MIOPEN_CHECK_RET(val) ASSERT_EQ(val, miopenStatusSuccess)
+
+class GPU_ConvFindModeAPI_FP32 : public ::testing::Test
```
randyspauldingamd, line 32: I don't *think* this requires GPU.. I could be wrong, but if correct I believe 'CPU_ConvFindModeAPI_NONE' is more in spirit with the gtest design doc.

test/gtest/conv_api_find_mode.cpp
```diff
@@ -0,0 +1,57 @@
...
+        conv_descr, miopenConvolutionFindMode_t::miopenConvolutionFindModeNormal));
+    MIOPEN_CHECK_RET(miopenGetConvolutionFindMode(conv_descr, &findMode));
+    EXPECT_EQ(findMode, miopenConvolutionFindMode_t::miopenConvolutionFindModeNormal);
+
+    ASSERT_EQ(miopenGetConvolutionFindMode(conv_descr, nullptr), miopenStatusBadParm);
+    ASSERT_EQ(miopenGetConvolutionFindMode(nullptr, &findMode), miopenStatusBadParm);
+    ASSERT_EQ(miopenSetConvolutionFindMode(
+                  nullptr, miopenConvolutionFindMode_t::miopenConvolutionFindModeNormal),
+              miopenStatusBadParm);
+
```
randyspauldingamd, line 55: What if you try to set it to a garbage value? The C API won't care, but it will propagate to the enum class miopen::debug::FindMode (find_controls.hpp) which is undefined behavior.

Sorry I don't have any suggestions--but I did want to point it out :)

randyspauldingamd (commented):
Never too late, right?

test/gtest/conv_api_find_mode.cpp
BrianHarrisonAMD: Yea totally right, ill make a follow up PR.

test/gtest/conv_api_find_mode.cpp
BrianHarrisonAMD: Since there is a FindMode struct wrapping the enum, I believe it'll behave as if it was set to normal, but that's an interesting point. 

I don't think we typically check enum values for correctness in our API, but we probably should.

PR #3479, opened 2025-01-22
Convert softmax from CTest to GTest

Vsevolod1983:
CTest to GTest conversion for Softmax
soft_max.cpp file is detected as renamed (moved), so probably will not be easy to review.


test/gtest/soft_max.cpp
```diff
@@ -55,34 +44,138 @@ T logaddexp(T x, T y, T neg_inf)
...
+        dinput = tensor<T>{test_case.in_dim}.generate(tensor_elem_gen_integer{max_value});
+
+        std::vector<T> tensorCpuDataBackward = GetBackwardCpu();
+        std::vector<T> tensorGpuDataBackward = GetBackwardGpu();
+
+        // check backward results
+        CompareResults(tensorGpuDataBackward, tensorCpuDataBackward);
+    }
+
+    std::vector<T> GetForwardCpu() const
```
bghimireamd, line 175: we also have [mloSoftmaxForwardRunHost ](https://github.com/ROCm/MIOpen/blob/8be8d80c3ab6567fa0947cb59ee6943358557d93/driver/mloSoftmaxHost.hpp#L49) and [mloSoftmaxBackwardRunHost ](https://github.com/ROCm/MIOpen/blob/8be8d80c3ab6567fa0947cb59ee6943358557d93/driver/mloSoftmaxHost.hpp#L243)too.

bghimireamd: Since this one is parallel may be we can move this [mloSoftmaxForwardRunHost ](https://github.com/ROCm/MIOpen/blob/8be8d80c3ab6567fa0947cb59ee6943358557d93/driver/mloSoftmaxHost.hpp#L49)and [mloSoftmaxBackwardRunHost](https://github.com/ROCm/MIOpen/blob/8be8d80c3ab6567fa0947cb59ee6943358557d93/driver/mloSoftmaxHost.hpp#L243) to avoid duplications.

Vsevolod1983: hi, @bghimireamd 
In C-test I did not see calls of these functions. Scope of my refactoring was to replace c-test with g-test.
If I misunderstand something, lets discuss.

Vsevolod1983: Or do you mean that we need to enhance test coverage for softmax? If yes I would prefer to do it in separate PR/ticket.

bghimireamd: yeah separate PR/ticket make more sense here :)

bghimireamd (approved):
LGTM

test/gtest/soft_max.cpp
```diff
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright (c) 2017 Advanced Micro Devices, Inc.
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
```
BrianHarrisonAMD: 2025 now hehe

test/gtest/soft_max.cpp
```diff
@@ -330,154 +401,97 @@ struct verify_backward_sofmax
...
-        /// \todo Resolve this workaround. Regular failures on Radeon VII, ROCm 3.3:
-        /// --float --input-dim 1 1 8 8 --algorithm 0 --mode 1 --scales 1 0 --tolerance 8000
-        /// FAILED: -nan
-        in_dim_set.erase({1, 1, 8, 8});
-        in_dim_set.erase({1, 1, 14, 14});
-        in_dim_set.erase({1, 1, 27, 27});
-        in_dim_set.erase({1, 32, 7, 7});
-        in_dim_set.erase({1, 32, 8, 8});
+        double threshold = std::numeric_limits<T>::epsilon() * tolerance;
+        double error     = miopen::rms_range(tensorCPUData, tensorGPUData);
```
BrianHarrisonAMD: This seems like a pretty large tolerance.
For float it might not be as large, but for half I think this would be very substantial.
Is it possible to reduce this, and still have the tests pass?

Vsevolod1983: Decreased for half to 80

BrianHarrisonAMD (approved):

PR #3481, opened 2025-01-23
Update ci docker to use ROCm 6.3.1 as base image

BrianHarrisonAMD:
Updating CI docker to a newer ROCm version.


bghimireamd (approved):

Dockerfile
```diff
@@ -123,7 +123,7 @@ RUN wget -O ck.tar.gz https://www.github.com/ROCm/composable_kernel/archive/${CK
     -D CMAKE_PREFIX_PATH=/opt/rocm \
     -D CMAKE_CXX_COMPILER_LAUNCHER="${COMPILER_LAUNCHER}" \
     -D CMAKE_BUILD_TYPE=Release \
-    -D GPU_ARCHS="gfx908;gfx90a;gfx942;gfx1100" \
+    -D GPU_ARCHS="gfx908;gfx90a;gfx942;gfx1100;gfx1101;gfx1102;gfx1103;gfx1200;gfx1201" \
```
cderb, line 126: Should this be filled by the GPU_ARCHS arg from line 92?

cderb (approved):

Dockerfile
BrianHarrisonAMD: Yea, but Jun set it up like this since he was having issues with formatting when reusing the ARG. 
I think we need to clean the Dockerfile generally since it has quite a few duplications.

PR #3482, opened 2025-01-23
Follow up changes for FindMode API

BrianHarrisonAMD:
@randyspauldingamd had feedback that was missed in the first PR #3476. 
This is a follow-up to address that feedback.

Changes:
- Updated internal enums to use external values to ensure they are in sync.
- Removed build variable, and replaced it with a enum value defined in the public API to make it clear what default is used.
- Added checking for invalid enum values.
- Added tests to cover new checks.
- Fixed naming of tests.


cderb (approved):

jeffdaily (approved):

amd-jnovotny (approved):
No documentation changes

test/gtest/conv_api_find_mode.cpp
```diff
@@ -29,29 +29,57 @@
...
                   nullptr, miopenConvolutionFindMode_t::miopenConvolutionFindModeNormal),
               miopenStatusBadParm);
 
+    ASSERT_EQ(miopenSetConvolutionFindMode(conv_descr, static_cast<miopenConvolutionFindMode_t>(0)),
+              miopenStatusBadParm);
+    ASSERT_EQ(
+        miopenSetConvolutionFindMode(conv_descr, static_cast<miopenConvolutionFindMode_t>(100)),
+        miopenStatusBadParm);
+    ASSERT_EQ(miopenSetConvolutionFindMode(conv_descr, static_cast<miopenConvolutionFindMode_t>(4)),
+              miopenStatusBadParm);
```
randyspauldingamd: Wups, I didn't notice these ASSERT_EQ's the first time. Do we need to assert? It's usually best to let the test continue as long as possible to maximize feedback.

Funny, because I did notice lines above such as `MIOPEN_CHECK_RET(miopenGetConvolutionFindMode(conv_descr, &findMode));` that don't need to assert. However, I decided not to comment on those to keep the complexity down.

BrianHarrisonAMD: Ah good point.
I think the createConv one does need to Assert since the rest cannot work if it fails, but the others can be Expects.

randyspauldingamd (approved):

PR #3487, opened 2025-02-03
Use state for disabling deterministic instead of environment variable

BrianHarrisonAMD:
The implicit GEMM solvers from CK were checking the environment variable instead of the attribute flag which resulted in CK solvers still being applicable when deterministic was enabled through the API.

This change will swap to using the attribute flag which will still be set when the environment flags are enabled, but also take into account what is set through the API.

After discussing with @bartekxk I believe only backwards weights (WRW) needs to be disabled for the CK grouped conv solvers when deterministic is set.


bghimireamd (approved):
LGTM

iq136boy (approved):
LGTM

PR #3502, opened 2025-02-12
Replacing vector with InlineVector in TensorDescriptor

novakovicdj:
This PR is replacing `std::vector` with `InlineVector` in `TensorDescriptor` for lens/strides. `InlineVector` https://github.com/ROCm/MIOpen/pull/3419

I did some performance testing to compare `InlineVector` with `std::vector`(initialization, elements access etc). The results are that `InlineVector` is ~6% faster than `std::vector`. For example initialization with 5 elements for `InlineVector` takes around 20.25ns, and initialization with 5 elements for `std::vector` takes around 21.85ns.

Based on these results I made changes in the code. I tried to make only necessary changes and to reduce overhead as much as possible (especially in the `src/`), also some changes needed to be made in `InlineVector` itself to be able to complete this change.


driver/CBAInferFusion_driver.hpp
```diff
@@ -106,9 +106,9 @@ class CBAInferFusionDriver : public Driver
     InputFlags& GetInputFlags() override { return inflags; }
 
     int GetandSetData() override;
-    std::vector<int> GetInputTensorLengthsFromCmdLine();
-    std::vector<int> GetOutputTensorLengths();
-    std::vector<int> GetWeightTensorLengthsFromCmdLine();
+    miopen::InlineVector<int, 5> GetInputTensorLengthsFromCmdLine();
```
CAHEK7: I think, since we've already created or own type, we can create a bit more convenient shortcut, like
```c++
template<typename T>
using LenStrides = InlineVector<T, 5>
```

src/include/miopen/inline_vector.hpp
```diff
@@ -66,21 +96,44 @@ class InlineVector
...
+        }
+        else
+        {
+            real_size = std::distance(first, last);
+            if(real_size > N)
+            {
+                MIOPEN_THROW("Input data size is bigger than InlineVector's capacity");
+            }
+            std::copy(first, last, storage.begin());
+        }
```
CAHEK7:
```suggestion
    template <typename InputIterator, typename std::enable_if_t<!std::is_integral_v<InputIterator>, bool> = true>
    InlineVector(InputIterator first, InputIterator last)
    {
        real_size = std::distance(first, last);
        if(real_size > N)
        {
            MIOPEN_THROW("Input data size is bigger than InlineVector's capacity");
        }
        std::copy(first, last, storage.begin());
    }
```

src/include/miopen/inline_vector.hpp
```diff
@@ -56,7 +58,35 @@ class InlineVector
...
+
+    InlineVector(size_type&& cnt, T&& val) : real_size(cnt)
+    {
+        if(real_size > N)
+        {
+            MIOPEN_THROW("Input data size is bigger than InlineVector's capacity");
+        }
+
+        std::fill(storage.begin(), storage.begin() + cnt, val);
+    }
```
CAHEK7:
```suggestion
    InlineVector(const size_typ cnt, const T& val = T()) : real_size(cnt)
    {
        if(real_size > N)
        {
            MIOPEN_THROW("Input data size is bigger than InlineVector's capacity");
        }

        std::fill_n(storage.begin(), cnt, val);
    }
```
A bit more detailed explanation:

It seems that
```C++
    InlineVector(size_type sz) : real_size(sz)
    {
        if(real_size > N)
        {
            MIOPEN_THROW("Input data size is bigger than InlineVector's capacity");
        }
    }
```
May not behave as expected, because when you are constructing `std::vector` like `std::vector<int> foo(5)` it must initialize `foo` with `0` by standard, but you case of `InlineVector` it remains uninitialized.

Also I see not points for rvalue-ref `InlineVector(size_type&& cnt, T&& val)`, btw passing `size_type` by ref or rvalue-ref is not a good idea - most of the time it will work, but it may prevent some optimizations.

src/include/miopen/inline_vector.hpp
```diff
@@ -66,21 +96,44 @@ class InlineVector
...
 
     // Copy/move operator
     InlineVector& operator=(const InlineVector& inline_vec) = default;
     InlineVector& operator=(InlineVector&& inline_vec) noexcept = default;
 
+    // Compare operators
+    bool operator==(const InlineVector& riv) const { return storage == riv.storage; }
+    bool operator!=(const InlineVector& riv) const { return storage != riv.storage; }
+    bool operator<(const InlineVector& riv) const { return storage < riv.storage; }
+    bool operator>(const InlineVector& riv) const { return storage > riv.storage; }
```
CAHEK7: It must take `real_size` into account, since the rest of the array may not be initialized.
So you have to implement those operators instead of reusing it from `std::array`.

src/include/miopen/inline_vector.hpp
```diff
@@ -246,6 +347,27 @@ class InlineVector
...
+std::ostream& operator<<(std::ostream& os, const InlineVector<T, N>& iv)
+{
+    if(iv.empty())
+    {
+        os << "{}";
+    }
+    else
+    {
+        os << "{";
+        for(int i = 0; i < iv.size() - 1; i++)
```
CAHEK7:
```suggestion
        for(size_t i = 0; i < iv.size() - 1; i++)
```
It emits `singned-unsigned comparison` warning.

src/include/miopen/inline_vector.hpp
```diff
@@ -202,6 +255,54 @@ class InlineVector
...
+            MIOPEN_THROW("Cannot erase data at this position");
+        }
+        real_size -= 1;
+        for(int i = idx; i < real_size; i++)
+        {
+            storage[i] = storage[i + 1];
+        }
+        return iterator(data() + idx);
+    }
+
```
CAHEK7, line 289: Why do we need this?

CAHEK7 (changes requested):

src/include/miopen/inline_vector.hpp
novakovicdj: Method erase is used in `driver/kthvalue_driver.hpp` file in `int KthvalueDriver<TIO>::GetandSetData()` function, so I wanted to keep it as it is and provide needed method in `InlineVector` same as in `std::vector`. Same situation is with insert method

CAHEK7 (approved):
I don't have any significant objection.
The only recommendation is to double all the `auto var =` parts of the code to be sure that `auto` is correctly used and shouldn't be replaced with `const auto& var =`.
We would like to avoid extra copying and pure `auto` is kind of error prone.

BradPepersAMD:

PR #3504, opened 2025-02-12
[RNN][gtest] GRU test fix

shurale-nkn:
Updated initial values in input buffers.
deepbench test size reduction: too large for verification in a reasonable time.


DrizztDoUrden (approved):
LGTM

PR #3507, opened 2025-02-13
Set mFinalized to true in BackendEngineDescriptor contructor

### LINKED ISSUE #3491: [Bug?] BackendEngineDescriptor is not marked as finalized when it is created from Engine.
https://github.com/ROCm/MIOpen/blob/cd9f3c2c6b5be362a7eba4010841aec60e87fb05/src/include/miopen/graphapi/engine.hpp#L203-L206

It seems like `mFinalized = true;` should be in the constructor body.

sohaibnd:
The constructor is called as part of BackendEngineHeurDescriptor::finalize() to store the Engine results from findEngines(), so it should already be finalized.


sohaibnd:
Mentioned by @lshqqytiger in issue https://github.com/ROCm/MIOpen/issues/3491

scerzh (approved):
A nice find! Thank you!

BrianHarrisonAMD (approved):

PR #3515, opened 2025-02-14
Swap to using median value with outliers removed for deciding on best solution to run

BrianHarrisonAMD:
These changes are intended to help make solution and instance selection more consistent.

Here are the issues I was seeing:
- There was no warm-up run for the invoker during instance tuning, and as a result, the first sample could have a much higher runtime than the average. 
- Using averages for comparing results caused outliers to have a big impact on the instance selection, and solver selection.

To resolve these issues:

- I have added a warm-up run to the instance tuning. This will ensure we measure all instances in a similar way.
- I have swapped from using averages to instead use the median value with outliers removed via modified Z.

### DIFF
--- a/src/conv/solver_finders.cpp
+++ b/src/conv/solver_finders.cpp
@@ -292,7 +292,7 @@ static std::vector<Solution> EvaluateInvokers(const Handle& handle,
             {
                 // Remove outliers that are more than 2 positive modified z-score's away, and get
                 // the mean.
-                elapsed = miopen::RemoveHighOutliersAndGetMean(samples, 2.0f);
+                elapsed = miopen::removeHighOutliersAndGetMean(samples, 2.0f);
             }
             else
             {
--- a/src/conv/solver_finders.cpp
+++ b/src/conv/solver_finders.cpp
@@ -291,9 +291,9 @@ static std::vector<Solution> EvaluateInvokers(const Handle& handle,
 
             if(samples.size() > 0)
             {
-                // Remove outliers that are more than 2 modified z-score's away, and get the median
-                // result.
-                elapsed = miopen::RemoveOutliersAndGetMedian(samples, 2.0f);
+                // Remove outliers that are more than 2 positive modified z-score's away, and get
+                // the mean.
+                elapsed = miopen::RemoveHighOutliersAndGetMean(samples, 2.0f);
             }
             else
             {
--- a/src/conv/solver_finders.cpp
+++ b/src/conv/solver_finders.cpp
@@ -278,7 +278,6 @@ static std::vector<Solution> EvaluateInvokers(const Handle& handle,
                 // don't include warm-up run in our samples.
                 if(i > 0)
                 {
-
                     samples.push_back(handle.GetKernelTime());
                 }
                 else
--- a/src/conv/solver_finders.cpp
+++ b/src/conv/solver_finders.cpp
@@ -274,15 +274,16 @@ static std::vector<Solution> EvaluateInvokers(const Handle& handle,
             while(i < N_RUNS_MAX && elapsed < TIME_MS_MAX)
             {
                 invoker(handle, invoke_ctx);
+
+                // don't include warm-up run in our samples.
                 if(i > 0)
                 {
-                    // don't include warm-up run in our measurements
+
                     samples.push_back(handle.GetKernelTime());
                 }
                 else
                 {
-                    // Use first_elapsed if we go past max time since we won't have any other
-                    // samples.
+                    // Keep first run just in case we go over the limit, and have no samples.
                     first_elapsed = handle.GetKernelTime();
                 }
                 ++i;
--- a/src/conv/solver_finders.cpp
+++ b/src/conv/solver_finders.cpp
@@ -32,6 +32,7 @@
 #include <miopen/perf_field.hpp>
 #include <miopen/conv/problem_description.hpp>
 #include <miopen/solution.hpp>
+#include <miopen/utility/modified_z.hpp>
 
 MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_CONV_GEMM)
 MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_CONV_DIRECT)
@@ -230,6 +231,7 @@ static std::vector<Solution> EvaluateInvokers(const Handle& handle,
     auto best         = std::numeric_limits<float>::max();
     auto best_invoker = Invoker{};
     auto ret          = std::vector<Solution>{};
+    std::vector<float> samples;
 
     for(const auto& sol : solutions)
     {
@@ -261,31 +263,40 @@ static std::vector<Solution> EvaluateInvokers(const Handle& handle,
 
         try
         {
-            // Run invoker max 6 times, with ~5 sec time limit.
+            // Run invoker max 8 times, with ~5 sec time limit.
             using elapsed_t                 = decltype(handle.GetKernelTime());
             constexpr elapsed_t TIME_MS_MAX = 5000.0;
             constexpr int N_RUNS_MAX        = 8;
-            constexpr int N_RUNS_DISCARD    = 3;
             auto elapsed                    = static_cast<elapsed_t>(0);
             auto first_elapsed              = static_cast<elapsed_t>(0);
             int i                           = 0;
+            samples.clear();
             while(i < N_RUNS_MAX && elapsed < TIME_MS_MAX)
             {
                 invoker(handle, invoke_ctx);
-                elapsed += handle.GetKernelTime();
-                if(i == (N_RUNS_DISCARD - 1))
-                    first_elapsed = elapsed;
+                if(i > 0)
+                {
+                    // don't include warm-up run in our measurements
+                    samples.push_back(handle.GetKernelTime());
+                }
+                else
+                {
+                    // Use first_elapsed if we go past max time since we won't have any other
+                    // samples.
+                    first_elapsed = handle.GetKernelTime();
+                }
                 ++i;
             }
-            // If the execution time was not too long,
-            // then the 1st run is not counted (assume it's warm-up):
-            if(i > N_RUNS_DISCARD)
+
+            if(samples.size() > 0)
             {
-                elapsed = (elapsed - first_elapsed) / static_cast<elapsed_t>(i - N_RUNS_DISCARD);
+                // Remove outliers that are more than 2 modified z-score's away, and get the median
+                // result.
+                elapsed = miopen::RemoveOutliersAndGetMedian(samples, 2.0f);
             }
-            else if(i > 0)
+            else
             {
-                elapsed /= i;
+                elapsed = first_elapsed;
             }
 
             MIOPEN_THROW_IF(elapsed <= 0, "Invalid elapsed time detected in EvaluateInvokers");
--- a/src/include/miopen/generic_search.hpp
+++ b/src/include/miopen/generic_search.hpp
@@ -546,7 +546,7 @@ auto GenericSearch(const Solver s,
 
                         // Remove outliers that are more than 2 positive modified z-score's away,
                         // and get the mean.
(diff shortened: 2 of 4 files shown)


src/conv/solver_finders.cpp
```diff
@@ -261,31 +263,41 @@ static std::vector<Solution> EvaluateInvokers(const Handle& handle,
...
             {
                 invoker(handle, invoke_ctx);
-                elapsed += handle.GetKernelTime();
-                if(i == (N_RUNS_DISCARD - 1))
-                    first_elapsed = elapsed;
+
+                // don't include warm-up run in our samples.
+                if(i > 0)
+                {
+
```
cderb: extra newline

cderb (approved):
LGTM

src/include/miopen/utility/modified_z.hpp
```diff
@@ -0,0 +1,117 @@
...
+#pragma once
+
+#include <vector>
+#include <algorithm>
+#include <miopen/errors.hpp>
+
+namespace miopen {
+
+template <typename T>
+T MedianOfSortedData(const std::vector<T>& sortedData)
```
scerzh: Some time ago we decided to stick to camelCase for function and variable names. However, I've noticed that we've recently stopped following this rule. Do you think we should abandon it?

scerzh (approved):
LGTM

src/include/miopen/utility/modified_z.hpp
BrianHarrisonAMD: Oh I am fine swapping it.
I just copy the style I see in existing files, and I happened to have one open that wasn't camelCase.

src/conv/solver_finders.cpp
```diff
@@ -261,31 +263,41 @@ static std::vector<Solution> EvaluateInvokers(const Handle& handle,
...
+            samples.clear();
             while(i < N_RUNS_MAX && elapsed < TIME_MS_MAX)
             {
                 invoker(handle, invoke_ctx);
-                elapsed += handle.GetKernelTime();
-                if(i == (N_RUNS_DISCARD - 1))
-                    first_elapsed = elapsed;
+
+                // don't include warm-up run in our samples.
+                if(i > 0)
```
CAHEK7, line 279: Sometimes even the second run may be longer than the rest of the batch. That's why it was "N_RUNS_DISCARD=3". I mean hardcoding it to "1" may not be a perfect solution.
That has been done for a reason.

By the way, why dot just gather all the samples and cut the first (or whatever number of "warmups") on postprocessing, since you've already gathering most of the samples. In that case you don't even need to handle a special 'first_elapsed' variable.

src/include/miopen/utility/modified_z.hpp
```diff
@@ -0,0 +1,117 @@
...
+T RemoveOutliersAndGetMedian(std::vector<T>& data, T z_threshold)
+{
+    std::sort(data.begin(), data.end());
+
+    std::vector<T> modZScores = ModifiedZScores(data);
+    std::vector<T> filteredData;
+
+    for(size_t i = 0; i < data.size(); ++i)
+    {
+        if(std::abs(modZScores[i]) <= z_threshold)
```
randyspauldingamd: This isn't what we want--it should be more robust in terms of not discarding valid samples. Sorry, I should have been more clear when I told you it was a one-tailed distribution:

To estimate typical real-world gpu performance, we only reject _slow_ samples (large positive z-scores). The rationale is that some infrequent operation on the system caused a large stochastic delay during that iteration. This happens all the time. On the other hand, gpu's don't tend to run _faster_ than normal, so we assume ALL negative z-scores are statistical in nature and retain them regardless of magnitude.

Could you change it to
```
if(modZScores[i] > z_threshold)
```
and it'd be nice to change the function name to reflect this. Perhaps 'removeHighTailOutliersAndGetMedian'

There _are_ of course possible outliers on the fast side--for example, perhaps there was a significant boost in clock rate during only one iteration. This would not be an appropriate sample to use for _typical_ performance. However, in the interest of simplicity it seems safe to ignore those cases.

src/include/miopen/utility/modified_z.hpp
```diff
@@ -0,0 +1,117 @@
...
+        std::transform(sortedData.begin(),
+                       sortedData.end(),
+                       std::back_inserter(modZScores),
+                       [&](auto& value) { return 0.6745 * (value - median) / mad; });
+
+        return modZScores;
+    }
+}
+
+template <typename T>
```
randyspauldingamd, line 115: Many of these functions will behave strangely if they're ever called with integral types or low-precision floats. We could:
A. ignore this, i.e. trust the user to never use them with problematic data
B. use enable_if to prevent this
C. cast inside the functions so floats are used when appropriate
I'm fine with letting you make the call--just pointing out some options.

src/include/miopen/generic_search.hpp
```diff
@@ -528,21 +543,25 @@ auto GenericSearch(const Solver s,
...
+                        // median result.
+                        elapsed_time = miopen::RemoveOutliersAndGetMedian(samples, 2.0f);
                         if(elapsed_time < best_time)
                         {
                             MIOPEN_LOG_I('#' << n_current << '/' << n_failed << '/' << n_runs_total
                                              << ' ' << elapsed_time << " < " << best_time << ' '
                                              << current_config);
                             best_config = current_config;
                             best_time   = elapsed_time;
+                            worst_time  = samples.back();
```
randyspauldingamd: It's not obvious that 'samples' _has_ been sorted at this point. The function name is already long...maybe add a blurb to the comment above?

src/include/miopen/generic_search.hpp
```diff
@@ -528,21 +543,25 @@ auto GenericSearch(const Solver s,
                     if(ret == 0)
                     {
                         is_passed = true;
-                        elapsed_time /= N_RUNS;
+
+                        // Remove outliers that are more than 2 modified z-score's away, and get the
+                        // median result.
+                        elapsed_time = miopen::RemoveOutliersAndGetMedian(samples, 2.0f);
```
randyspauldingamd: We should continue using the mean as the accepted value. MED/MAD is just for outlier rejection. At this point we've tossed out the 'bad' values, so every remaining sample is equally important.

randyspauldingamd (changes requested):
Use the mean for post-outlier rejection and don't reject fast samples.

src/include/miopen/generic_search.hpp
```diff
@@ -503,11 +517,11 @@ auto GenericSearch(const Solver s,
...
                 // Smooth the jitter of measurements:
-                // If the 1st probe is NOT too bad (measured time <= 1.10 * best known time),
-                // then re-run it 9 times more and compute average time,
-                // and decide using average of all 10 attempts vs. the best.
+                // If the 1st probe is NOT too bad (measured time <= 1.10 * worst sample of the best
+                // config), then gather 9 more samples, and remove outliers. Use the median value
+                // with outliers removed for calculating best config.
                 constexpr int N_RUNS = 10;
-                if(elapsed_time / best_time < 1.10f)
+                if(elapsed_time / worst_time < 1.10f)
```
randyspauldingamd, line 524: Relating the full test to the worst sample is a very nice change. Although it could increase search time in cases when there was one 'large-ish' sample that was almost-but-not-quite rejected. I can think of ways to test for this, but I think it's low priority and should be left as possible future work.

src/include/miopen/utility/modified_z.hpp
BrianHarrisonAMD: Ah yea that's a good point ill swap it.

BrianHarrisonAMD: I think I want it to be:

```C++
if(modZScores[i] <= z_threshold)
```

Since it's only adding to the filteredData if it's less than or equal to the z_threshold we are removing.
That way we only filter positive z_scores higher than the threshold.

randyspauldingamd: Ahh of course, nice catch!

randyspauldingamd (approved):
lgtm

src/include/miopen/utility/modified_z.hpp
```diff
@@ -0,0 +1,127 @@
...
+    T median = (size % 2 == 0) ? (sortedData[size / 2 - 1] + sortedData[size / 2]) / 2.0
+                               : sortedData[size / 2];
+
+    return median;
+}
+
+template <typename T>
+T median(std::vector<T>& data)
+{
+    std::sort(data.begin(), data.end());
```
CAHEK7, line 65: Actually you need an `nth_element`. But it seems you rely on the sorted results later (which is kind of dangerous, because somebody will replace `st::sort` to `std::nth_element` sooner or later and silently break the code relied on sorting)

src/include/miopen/utility/modified_z.hpp
CAHEK7: D. `static_assert` instead of B

src/include/miopen/utility/modified_z.hpp
```diff
@@ -0,0 +1,127 @@
...
+    std::vector<T> modZScores = modifiedZScores(data);
+    std::vector<T> filteredData;
+
+    for(size_t i = 0; i < data.size(); ++i)
+    {
+        if(modZScores[i] <= z_threshold)
+        {
+            filteredData.push_back(data[i]);
+        }
+    }
```
CAHEK7, line 130: `std::copy_if`

CAHEK7 (commented):
There are some minor suggestions.

src/conv/solver_finders.cpp
BrianHarrisonAMD: We used to discard the first 3, but outliers can happen at any point.
Now it will always discard the first sample since it's the warm-up (and very often outside the expected range), but for the rest the z-score is used to remove outliers.

src/include/miopen/utility/modified_z.hpp
BrianHarrisonAMD: Yea we need it sorted eventually, but that's a good point ill leave a comment.
Luckily tests would catch it if someone changed it.

src/include/miopen/utility/modified_z.hpp
BrianHarrisonAMD: I had tried that originally, but since I am checking one array (modZScores[i]), and copying the value from the other (data[i]), it seemed worse than the for loop.

src/include/miopen/utility/modified_z.hpp
randyspauldingamd: It seems like enable_if would be preferred since it'll reduce compile time. It does look uglier though. Are there other nuances involved?

test/gtest/unit_modified_z.cpp
```diff
@@ -0,0 +1,185 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2024 Advanced Micro Devices, Inc.
```
BrianHarrisonAMD: 2025

src/include/miopen/utility/modified_z.hpp
BrianHarrisonAMD: I prefer static_assert since it makes it very clear what the issues are when it fails.
That's why I went with that suggestion anyways.

src/include/miopen/utility/modified_z.hpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+template <typename T>
+T removeHighOutliersAndGetMean(std::vector<T>& data, T z_threshold)
+{
+    static_assert(std::is_floating_point_v<T>);
+    std::sort(data.begin(), data.end());
+
+    std::vector<T> modZScores = modifiedZScores(data);
```
BrianHarrisonAMD, line 121: This will be an issue if modZScores is empty due to a MAD of 0.

src/include/miopen/utility/modified_z.hpp
CAHEK7: I'm not sure that we need it sorted - as far as I can see we need min, max and median. But since we are dealing with small arrays, it doesn't matter.

PR #3521, opened 2025-02-18
Add unit tests for implicit gemm bwd/fwd xdlops solver

JonathanLichtnerAMD:
Add bfp16 and unit tests for the CK implicit gemm fwd and bwd xdlops solvers

* enable bfp16 for implicit gemm fwd and bwd xdlops solvers

* add unit tests for implicit gemm fwd and bwd xdlops solvers

* add int8 handling to the gtest unit conv solver

---------

    Co-authored-by: bibek <108366729+bghimireamd@users.noreply.github.com>


bghimireamd:
I attempted to locate where the CPU and GPU results are compared. I checked [RunTest](https://github.com/ROCm/MIOpen/blob/develop/test/gtest/unit_conv_solver.hpp#L130), but I didn't find any code performing this comparison.

averinevg:
> I attempted to locate where the CPU and GPU results are compared. I checked [RunTest](https://github.com/ROCm/MIOpen/blob/develop/test/gtest/unit_conv_solver.hpp#L130), but I didn't find any code performing this comparison.

@bghimireamd https://github.com/ROCm/MIOpen/blob/25c25532bad2d104f0f8cec59fa753910573abb5/test/gtest/unit_conv_solver.cpp#L284

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+auto GetTestParams(miopenDataType_t datatype)
+{
+    Gpu supportedDevices = Gpu::gfx908 | Gpu::gfx94X;
+    if(datatype != miopenHalf)
+    {
+        supportedDevices = supportedDevices | Gpu::gfx90A;
+    }
+    auto params = miopen::unit_tests::UnitTestConvSolverParams(supportedDevices);
+    params.Tunable(1);
```
averinevg: In order to check tuning we need several iterations
```suggestion
    params.Tunable(5);
```

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+auto GetConvTestCases(miopenDataType_t datatype)
+{
+    using TestCase = miopen::unit_tests::ConvTestCase;
+
+    return std::vector{
+        // clang-format off
+        TestCase{{datatype, miopenTensorNHWC, {1, 128, 24, 48}},
+                 {datatype, miopenTensorNHWC, {384, 128, 1, 1}},
+                 datatype, {{0, 0}, {1, 1}, {1, 1}}},
```
averinevg: Since it is a smoke test then smaller config can be used in order to spend less time on the test.

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+using GPU_UnitTestConvSolverImplicitGemmBwdXdlops_FP16  = GPU_UnitTestConvSolverBwd_FP16;
+using GPU_UnitTestConvSolverImplicitGemmBwdXdlops_BFP16 = GPU_UnitTestConvSolverBwd_BFP16;
+using GPU_UnitTestConvSolverImplicitGemmBwdXdlops_FP32  = GPU_UnitTestConvSolverBwd_FP32;
+using CPU_UnitTestConvSolverImplicitGemmBwdXdlopsDevApplicability_FP16 =
+    CPU_UnitTestConvSolverDevApplicabilityBwd_NONE;
+using CPU_UnitTestConvSolverImplicitGemmBwdXdlopsDevApplicability_BFP16 =
+    CPU_UnitTestConvSolverDevApplicabilityBwd_NONE;
+using CPU_UnitTestConvSolverImplicitGemmBwdXdlopsDevApplicability_FP32 =
+    CPU_UnitTestConvSolverDevApplicabilityBwd_NONE;
```
averinevg: This test is necessary to check the solver with different gpus (using mock handle). If the solver applicability does not depend on the data type, there is no need to test with all supported data types. One is enough.

test/gtest/unit_conv_solver_ConvHipImplicitGemmFwdXdlops.cpp
```diff
@@ -0,0 +1,121 @@
...
+    return params;
+}
+
+} // namespace
+
+using GPU_UnitTestConvSolverImplicitGemmFwdXdlops_I8    = GPU_UnitTestConvSolverFwd_I8;
+using GPU_UnitTestConvSolverImplicitGemmFwdXdlops_FP16  = GPU_UnitTestConvSolverFwd_FP16;
+using GPU_UnitTestConvSolverImplicitGemmFwdXdlops_BFP16 = GPU_UnitTestConvSolverFwd_BFP16;
+using GPU_UnitTestConvSolverImplicitGemmFwdXdlops_FP32  = GPU_UnitTestConvSolverFwd_FP32;
+using CPU_UnitTestConvSolverImplicitGemmFwdXdlops_NONE =
```
averinevg:
```suggestion
using CPU_UnitTestConvSolverImplicitGemmFwdXdlopsDevApplicability_NONE =
```

averinevg (changes requested):

test/gtest/unit_conv_solver.cpp
```diff
@@ -699,6 +699,9 @@ void RunSolver(const miopen::solver::conv::ConvSolverInterface& solver,
         case miopenBFloat16:
             RunSolver<bfloat16, bfloat16>(solver, params, direction, conv_config, algo);
             return;
+        case miopenInt8:
+            RunSolver<int8_t, int8_t>(solver, params, direction, conv_config, algo);
+            return;
```
BrianHarrisonAMD, line 742: Bit surprising this is in this PR.

Any reason in particular it was added with these changes?

JonathanLichtnerAMD: Yes, because I added tests for all the data types supported by the implicit gemm fwd xdlops solver in the next commit:

```
+                         GPU_UnitTestConvSolverImplicitGemmFwdXdlops_I8,
+                         testing::Combine(testing::Values(GetTestParams()),
+                                          testing::Values(miopenConvolutionAlgoImplicitGEMM),
+                                          testing::ValuesIn(GetConvTestCases(miopenInt8))));
```

I assumed we wanted to test all the existing data types and not just for bfp16.  I can update the comment in that commit to mention that it will be used in a future commit, or alternatively, if we don't care about int8, then I can remove this change as well as the int8 test.

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
BrianHarrisonAMD: I think it would be good to add a separate set of FullTestCases that cover the edge cases / different conv settings.

Specifically, having a few cases with padding, strides > 1, and dilation > 1 would be good.

@averinevg thoughts on this?
When I look at the existing unit_conv_solver's I can see some where we have shapes we are testing related to specific defects, but having a general small set that can ensure basic functionality is covered seems useful as a starting point.

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
```diff
@@ -0,0 +1,135 @@
...
+    };
+}
+
+auto GetTestParams(miopenDataType_t datatype)
+{
+    Gpu supportedDevices = Gpu::gfx908 | Gpu::gfx94X;
+    if(datatype != miopenHalf)
+    {
+        supportedDevices = supportedDevices | Gpu::gfx90A;
+    }
```
BrianHarrisonAMD: Should this be here?
It looks like gfx90A should be supported for half.

test/gtest/unit_conv_solver.cpp
BrianHarrisonAMD: Ah makes sense.
All good!

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: The fp16 tests will fail without this because of the following lines in the IsApplicable() method in conv_hip_implicit_gemm_bwd_data_xdlops.cpp:
``` 
    if(arch == "gfx90a" && problem.IsGfx90aFp16altRequired())
        return false;
```
It did seem odd that gfx908 supports this but not gfx90a...

BrianHarrisonAMD: Oh weird the grouped version doesn't have the same check I believe.
Might be worth checking with CK why that is.

BrianHarrisonAMD: https://rocm.docs.amd.com/projects/MIOpen/en/latest/conceptual/MI200-alt-implementation.html

BrianHarrisonAMD: Okay got some clarity here.
I think we actually might need to have a similar check in the grouped conv.

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: Change made.  Thanks!

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: That's a good point.  I've reduced these as much as CK would allow me.

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: Done

test/gtest/unit_conv_solver_ConvHipImplicitGemmFwdXdlops.cpp
JonathanLichtnerAMD: Done

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: I've added some "basic" full tests (padding, strides > 1, dilation > 1, etc).  Please take a look and let me know if these look okay or if I should add more tests.

Note that the last commit is currently WIP since CK didn't seem to like stride > 1 for the bwd convolution...

averinevg: I think we should discuss here what we will refer to as smoke and full tests. We could even later add this to the wiki.

In my opinion, a smoke test is a test which is used to ensure that the solver correctly selects the kernel, and that this kernel is compiled, executed, and produces correct computations. If the solver has a complex kernel selection logic, we should test all possible paths. If constants are passed to the kernel during compilation, and their values depend on input parameters, I suggest considering this as one kernel, regardless of the fact that we have multiple binaries generated from the same source code. We should use minimal necessary configurations to perform the testing in the shortest possible time.

Full tests are needed to ensure that the calculations are performed correctly for specific configurations. I would consider all regression tests as full tests.

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
averinevg: We have `MIOPEN_DEBUG_CONVOLUTION_ATTRIB_FP16_ALT_IMPL` for that. I'm currently working on adding this to the unit tests. I plan to finish it today.

averinevg:
@JonathanLichtnerAMD Please do not use force-push, because after it, it's unclear what was changed, and we have to review everything again.

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: @averinevg: From a practical usage point-of-view, what is the difference between smoke tests and full tests.  From what I can see we run both the smoke and full tests in the github jenkins pipeline, so it looks like we use the smoke tests to give us early failures if a PR is obviously broken rather than running all the tests, which might give the developer early warning that something is wrong and would minimize pipeline resources.  The developer might also just run the smoke tests, and then rely on the jenkins pipeline to run all the tests since the tests take *so long* to run.

Or are there other distinctions in how we use smoke and full tests?  I think if we want smoke tests to be fast, and ensure that functionality most likely works, then what you mention above makes sense to me.

On that note, are you okay with the full tests that I've added?

averinevg: @JonathanLichtnerAMD Yes, you're right. As for the tests themselves, I have no objections.

JonathanLichtnerAMD:
> @JonathanLichtnerAMD Please do not use force-push, because after it, it's unclear what was changed, and we have to review everything again.

Sigh, okay I will start doing this.  I come from a development practice where we kept our commit history clean so that each commit built on the other commits, which meant that reviews _should_ be easier.  It also meant that the history would be easy to read and follow after the merge (although that's moot since we squash our commits on the merge).

However, github does a very poor job of handling this style of development, so I will start adding commits for each change....

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
```diff
@@ -0,0 +1,151 @@
...
+    using TestCase = miopen::unit_tests::ConvTestCase;
+
+    return std::vector{
+        // clang-format off
+        TestCase{{datatype, miopenTensorNHWC, {1, 32, 8, 8}},
+                 {datatype, miopenTensorNHWC, {32, 32, 3, 3}},
+                 datatype, {{1, 1}, {1, 1}, {1, 1}}}, // non-zero padding
+        //TestCase{{datatype, miopenTensorNHWC, {1, 32, 8, 8}},
+        //         {datatype, miopenTensorNHWC, {32, 32, 3, 3}},
+        //         datatype, {{0, 0}, {2, 2}, {1, 1}}}, // stride > 1
```
BrianHarrisonAMD: Should this be uncommented?

JonathanLichtnerAMD: Unfortunately the test fails if the stride > 1, so I was going to look into that a bit more today (CK doesn't seem to like a stride > 1 in this case).  I will either update this, or remove this test...

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: PS @averinevg and @BrianHarrisonAMD: is it okay if I resolve this conversation, or is there a pending action for me here?  Thanks!

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: I found the CK IsSupportedArgument() method (in device_conv2d_bwd_data_xdl_nhwc_kyxc_nhwk.hpp) for this solver and traced it, and CK indicated the following reasons for rejecting these particular parameters:
1. invalid Gridwise GEMM size, and/or
2. invalid ConvolutionBackwardDataSpecialization::Filter1x1Stride1Pad0 specialization

I think it should be fine to just delete the above commented lines...

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: @BrianHarrisonAMD and @averinevg: can this discussion be resolved?  Thanks

BrianHarrisonAMD: yep

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
BrianHarrisonAMD: yep

BrianHarrisonAMD (approved):
LGTM

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
averinevg: Not exactly. I added a couple of lines to handle this in my [PR](https://github.com/ROCm/MIOpen/pull/3539). I think we need to use it here as well.
https://github.com/ROCm/MIOpen/blob/5562f621f6abb0f8bb411d53f709c788b5c4cad1/test/gtest/unit_conv_solver.hpp#L99

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: Hi @bartekxk, I just wanted to double check that CK is exhibiting the correct behaviour for this convolution, i.e., that it is really not supported for the given arguments (it seems to rule out a lot of convolutions if stride is greater than one).  For example, the following convolution does not find any ConvHipImplicitGemmBwdXdlops kernels for the given arguments:

```
MIOPEN_DEBUG_DISABLE_FIND_DB=1  MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvHipImplicitGemmBwdXdlops ./bin/MIOpenDriver convbfp16 -n 1 -c 32 -H 8 -W 8 -k 32 -y 3 -x 3 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -S 129

...

Backward Data Conv solutions available: 2
- id: 155 algo: 5, time: 500 ms, ws: 0, name: ConvHipImplicitGemmGroupBwdXdlops 
- id: 86 algo: 1, time: 1000 ms, ws: 0, name: ConvDirectNaiveConvBwd
```

but it does work work if stride is 1:

```
MIOPEN_DEBUG_DISABLE_FIND_DB=1  MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvHipImplicitGemmBwdXdlops ./bin/MIOpenDriver convbfp16 -n 1 -c 32 -H 8 -W 8 -k 32 -y 3 -x 3 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -S 129

...

Backward Data Conv solutions available: 3
- id: 129 algo: 5, time: 500 ms, ws: 0, name: ConvHipImplicitGemmBwdXdlops
- id: 155 algo: 5, time: 500 ms, ws: 0, name: ConvHipImplicitGemmGroupBwdXdlops
- id: 86 algo: 1, time: 1000 ms, ws: 0, name: ConvDirectNaiveConvBwd
```

It also works with stride > 1 if we use larger values for C and k, e.g.:
```
MIOPEN_DEBUG_DISABLE_FIND_DB=1  MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvHipImplicitGemmBwdXdlops ./bin/MIOpenDriver convbfp16 -n 1 -c 128 -H 24 -W 48 -k 384 -y 3 -x 3 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -S 129
```

but not for something like
```
MIOPEN_DEBUG_DISABLE_FIND_DB=1  MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvHipImplicitGemmBwdXdlops ./bin/MIOpenDriver convbfp16 -n 1 -c 32 -H 24 -W 48 -k 32 -y 3 -x 3 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -S 129
```

Thanks!

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: Thank you, @averinevg.  I will make these changes once your changes are merged

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
bartekxk: Hello @JonathanLichtnerAMD , I ran following command in CK 
`python3 ../script/convert_miopen_driver_to_profiler.py ./bin/MIOpenDriver convbfp16 -n 1 -c 32 -H 24 -W 48 -k 32 -y 3 -x 3 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 -in_layout NHWC`
And I can find a lot of instances for this case in CK

averinevg (approved):

averinevg:
@JonathanLichtnerAMD This [smoke test](https://github.com/ROCm/MIOpen/blob/41a3b435279cd52ccb42a1f9827a714c10b2b48f/test/CMakeLists.txt#L835) needs to be removed as it is no longer needed.

https://github.com/ROCm/MIOpen/blob/41a3b435279cd52ccb42a1f9827a714c10b2b48f/test/CMakeLists.txt#L835

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: Thanks @bartekxk.  Were any of these instances for ConvHipImplicitGemmBwdXdlops?  The concern was that we were not getting any kernels from this solver....  But maybe I was running an old instance of CK (maybe)?

JonathanLichtnerAMD:
@averinevg The custom smoke test "add_custom_test(smoke_solver_ConvHipImplicitGemmFwdXdlops ..." has been removed.

averinevg (approved):

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
JonathanLichtnerAMD: I will resolve this conversation since it sounds like this is not an issue.  We can track this via a separate item if you think there might be an issue here.  Thanks!

PR #3543, opened 2025-02-24
gfx950 changes, squashed

randyspauldingamd:
Adding gfx950 support
Improving support for NPI repo
Renamed [b]float8->[b]float8_fnuz, retained existing support
- note on types: this PR is rename-only. In most cases, usage is not under consideration.


CMakeLists.txt
```diff
@@ -634,20 +634,24 @@ endforeach()
...
+            list(APPEND LFS_MISSING_FILES ${__fname})
+        else()
+            unpack_db(${FILE_NAME})
+            if( NOT ENABLE_ASAN_PACKAGING )
+                if( NOT MIOPEN_TEST_DBSYNC )
+                    set(__component_name COMPONENT ${COMPONENT_NAME})
+                endif()
+                rocm_install(FILES ${KERNELS_BINARY_DIR}/${__fname}
+                            DESTINATION ${DATABASE_INSTALL_DIR}
+                            ${__component_name})
```
averinevg, line 650: Seems unrelated to gfx950

driver/driver.hpp
```diff
@@ -44,9 +44,9 @@
 using half         = half_float::half;
 using hip_bfloat16 = bfloat16;
 #include <hip_float8.hpp>
-using float16 = half_float::half;
-using float8  = miopen_f8::hip_f8<miopen_f8::hip_f8_type::fp8>;
-using bfloat8 = miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8>;
+using float16      = half_float::half;
+using float8_fnuz  = miopen_f8::hip_f8<miopen_f8::hip_f8_type::fp8>;
+using bfloat8_fnuz = miopen_f8::hip_f8<miopen_f8::hip_f8_type::bf8>;
```
averinevg, line 49: Why do we need this?

src/solver/conv/conv_winoRxS.cpp
```diff
@@ -681,7 +681,7 @@ static bool IsApplicableBase(const ExecutionContext& ctx, const ProblemDescripti
     if(problem.IsFp16() &&
        !(StartsWith(name, "gfx906") || StartsWith(name, "gfx908") || StartsWith(name, "gfx90a") ||
          StartsWith(name, "gfx94") || StartsWith(name, "gfx1011") || StartsWith(name, "gfx1012") ||
-         StartsWith(name, "gfx103") || StartsWith(name, "gfx11")))
+         StartsWith(name, "gfx103") || StartsWith(name, "gfx11") || StartsWith(name, "gfx95")))
```
averinevg: I think this needs to be sorted

CMakeLists.txt
randyspauldingamd: This is from the internal repo (NPI), which doesn't have enough space for the database files. Now if *.bz2 are missing, it's a warning instead of a failure. This is now also the preferred mechanism for the public repo.

driver/driver.hpp
randyspauldingamd: This is just changing the name of the type, and the purpose of the code is beyond the scope of this PR. It would be great to clean up unneeded code, so please do add a ticket to the backlog if you'd like to see this investigated.

averinevg: If the purpose of this code is beyond the scope of this PR, then why do these changes take up a significant part of it?

averinevg (changes requested):
~~It is necessary to remove everything from this PR that is not related to adding the new gpu support~~

cderb (approved):

randyspauldingamd:
> It is necessary to remove everything from this PR that is not related to adding the new gpu support

@averinevg please read the PR and revise your review. Thanks!

src/solver/conv/conv_winoRxS.cpp
randyspauldingamd: Will do!

driver/driver.hpp
randyspauldingamd: The old hardware used a nonstandard definition for [b]float8. The new hardware corrects this, and no longer uses those definitions. We've been instructed to change the existing names, adding the _fnuz suffix. There will be a followup PR which adds the new types, but reuses the original [b]float8 names.

include/miopen/miopen.h
```diff
@@ -371,17 +371,12 @@ typedef enum
...
-// miopenReserved1 = 7,
-// miopenReserved2 = 8,
-#endif
-    miopenInt64 = 9,
+    miopenBFloat16 = 5,     /*!< 16-bit binary floating point (8-bit exponent, 7-bit fraction)
+                               (Partially supported) */
+    miopenDouble       = 6, /*!< 64-bit floating point (Partially supported) */
+    miopenFloat8_fnuz  = 7,
+    miopenBFloat8_fnuz = 8,
+    miopenInt64        = 9,
```
averinevg, line 379: Should we keep MIOPEN_BETA_API here?

randyspauldingamd: These are the new permanent names, so no, not on these.

BradPepersAMD: I'd like to stop using it until we actually decide and discuss what it does.  It looks like it should hide some changes as they go through beta testing.  But I think it's always enabled so it is doing nothing right now.

averinevg (approved):

include/miopen/miopen.h
CAHEK7: @BradPepersAMD 
It actually marks parts of the library as "unstable" allowing as to do API-breaking changes without significant consequences.
That's the main point.

But, as far as I remember, there is no procedure for a transition into a stable state, so those `MIOPEN_BETA` parts may exist for ages.

PR #3545, opened 2025-02-25
run single test case for bn

bghimireamd:
Tests cases :[  [(NCHW, spatial), (NHWC, spatial), (NCHW, per active), (NHWC, per active)]](https://github.com/ROCm/MIOpen/blob/develop/test/gtest/bn_fwd_train.cpp#L163-L165) for batch norm were running in parallel. Running tests in parallel was causing large tensor case fail with insufficient memory error. This PR now make those batch norm tests run sequentially from cmake's `gtest_discover_tests`.


shurale-nkn (approved):
LGTM if the solution fits within the time constraints of the CI

BrianHarrisonAMD (approved):
LGTM assuming the CI is okay with it.

PR #3547, opened 2025-02-25
promote latest CK version

illsilin:
update the CK hash to the tip of amd-develop branch.


AviralGoelAMD:
ignore my commits, the first one replaces the hash with dummy value and the second one reverts that replacement to original value

PR #3553, opened 2025-02-26
[BN] Serial run large tensors test cases

bghimireamd:
Follow up PR for https://github.com/ROCm/MIOpen/pull/3545 Since that solution ended up taking huge CI time. In this PR we sperate the large tensor and create a separate file, effectively creating separate binary. Then to this binary that only does large tensor test we serialize the run via CMake's `gtest_discover_tests`


test/gtest/bn_test_data.hpp
```diff
@@ -119,19 +122,35 @@ inline std::vector<BN2DTestCase> Network2DLarge()
...
+        {128, 256, 16, 16, miopen::batchnorm::Direction::Backward, 0, 1},
+    };
+}
 
-        };
-    // clang-format on
+template <>
+inline std::vector<BN3DTestCase> Network3DSerialCase()
+{
+    return {{2, 2048, 16, 128, 128, miopen::batchnorm::Direction::Backward, 0, 1}};
```
shurale-nkn, line 153: 2\*2048\*16\*128\*128=1073741824=1\*10^9

test/gtest/bn_bwd_serial_run.cpp
```diff
@@ -0,0 +1,213 @@
...
+                         TestNameGenerator<BN2DTestCase>());
+
+INSTANTIATE_TEST_SUITE_P(Smoke,
+                         GPU_BNBWDSerialRun3D_FP64,
+                         testing::Combine(testing::ValuesIn(Network3DSerialCase<BN3DTestCase>()),
+                                          testing::ValuesIn({miopenTensorNCHW, miopenTensorNHWC}),
+                                          testing::ValuesIn({miopenBNSpatial,
+                                                             miopenBNPerActivation}),
+                                          testing::ValuesIn({testBNAPIV2})),
+                         TestNameGenerator<BN3DTestCase>());
```
BrianHarrisonAMD, line 213: Can we move these to full if they are going to run serial?

Seems like long running tests shouldn't be in smoke stage.

test/gtest/bn_test_data.hpp
```diff
@@ -119,19 +122,35 @@ inline std::vector<BN2DTestCase> Network2DLarge()
...
+inline std::vector<BN2DTestCase> Network2DSerialCase()
+{
+    return {
+        {64, 2048, 7, 7, miopen::batchnorm::Direction::Backward, 0, 1},
+        {64, 2048, 17, 17, miopen::batchnorm::Direction::Backward, 0, 1},
         // edge cases
         {69328, 1, 22, 22, miopen::batchnorm::Direction::ForwardTraining, 1, 1},
         {69328, 1, 13, 79, miopen::batchnorm::Direction::ForwardTraining, 1, 1},
+        {128, 256, 14, 14, miopen::batchnorm::Direction::Backward, 0, 1},
+        {128, 256, 16, 16, miopen::batchnorm::Direction::Backward, 0, 1},
```
shurale-nkn, line 146: 69328\*1\*22\*22=33554752=3.3\*10^7
128\*256\*14\*14=6422528=6.4\*10^6
they are all quite small compared to 10^9. 
Recommendation to remove from serial in next PR.

shurale-nkn (approved):

test/gtest/bn_test_data.hpp
BrianHarrisonAMD: There is also some in the Network large above that are bigger I think. 
I think we should decide on a cut-off size, and split it consistently.

BrianHarrisonAMD (approved):

test/gtest/bn_bwd_serial_run.cpp
shurale-nkn: I agree that it would be better to move them. But if a solution is needed right here and now, in my opinion, it’s not critical.
The biggest consumers are (2, 2048, 16, 128, 128)
image

test/gtest/bn_test_data.hpp
bghimireamd: When I listed all the shapes size by descending order 

1. Shape: (2, 2048, 16, 128, 128) --> Size: 1.07e+09
2. Shape: (69328, 1, 13, 79) --> Size: 7.12e+07
3. Shape: (64, 1024, 1024) --> Size: 6.71e+07
4. Shape: (64, 256, 56, 56) --> Size: 5.14e+07

It seem like I can cut off 1e+9 since (2, 2048, 16, 128, 128)  was the tensor size that caused the error in CI

PR #3556, opened 2025-02-27
Remove phantom example text from API documentation

amd-jnovotny:
This fixes the issues where irrelevant example links leading to the same miopenFuseProblems description. The root cause of the problem is a bug with the "@example" keyword, which caused the duplicate entries. It converted it to "@details" to treat it as body text in the API. The text still displays as part of the miopenFuseProblems documentation, but no longer repeats elsewhere.


bghimireamd (approved):

amd-jnovotny:
@bghimireamd : Thanks for the approval. I don't have authorization to merge so someone on your team will have to do so.

PR #3558, opened 2025-02-27
use 32 threads to build ck

bghimireamd:
Do not merge this. Just testing if lowering the number of thread fix the OOM issue during CK build


bghimireamd:
32 threads took longer then 21 hrs and still did not complete build. Closing this PR since build time is not feasible.

PR #3564, opened 2025-02-28
[BN] Finalize batch norm OpenCL kernel optimization

ex-rzr:
Vectorization:

 - fwd spatial multiple
 - bwd spatial multiple
 - fwd spatial inference
 - fwd inference per activation

Stashing methods:

 - add stashing in N dimension when `last_ylocalsize < stash_values`
 - add stashing in N dimension in groups of 2 bytes for mixed precision when `c % 2 != 0`

Kernel dispatch spatial multiple:

 - extract common
 - decrease `max_localsize` until the number of workgroups is greater than 80% of the available CUs

Fixes:

 - fix failures due to unaligned accesses in mixed precision NCHW when `H * W` is odd (like `-H 225 -W 225`)


bghimireamd (approved):
LGTM

BrianHarrisonAMD (approved):

PR #3565, opened 2025-02-28
Dockerfile/Jenkins Improvements.

adickin-amd:
- Remove duplicated GPU_ARCHS 
- Read CK_COMMIT from requirements.txt rather than having duplicate shas in both files
- refactor dockerfile to be more streamlined, reducing RUN commands to save space.  Additionally remove unneeded components to save more space in the final image.
- add video/render group ids to docker opts when launching container on jenkins to attempt to fix GPU not found issue.


vars/utils.groovy
```diff
@@ -207,7 +222,7 @@ def getDockerImage(Map conf=[:])
         freckle = sh(script: 'git rev-parse --short HEAD', returnStdout: true).trim()
     }
 
-    def dockerArgs = "--build-arg BUILDKIT_INLINE_CACHE=1 --build-arg PREFIX=${prefixpath} --build-arg GPU_ARCHS='\"${gpu_arch}\"' --build-arg USE_MLIR='${mlir_build}' --build-arg INSTALL_MIOPEN=${install_miopen} --build-arg FRECKLE=${freckle}"
+    def dockerArgs = "--build-arg BUILDKIT_INLINE_CACHE=1 --build-arg PREFIX=${prefixpath} --build-arg GPU_ARCHS='\"${gpu_arch}\"' --build-arg CK_BUILD_THREADS=${num_threads_for_ck_build} --build-arg USE_MLIR='${mlir_build}' --build-arg INSTALL_MIOPEN=${install_miopen} --build-arg FRECKLE=${freckle}"
```
adickin-amd: Need to either move the Ck build thread calculation into the dockerfile or add it after we have computed the container name.  Doing it here causes two different names when 64 vs 32 is selected which triggers two builds of an identical container.

Dockerfile
```diff
@@ -1,155 +1,147 @@
...
+    usermod -a -G render,video root
 
-# purge existing composable kernel installed with ROCm
-# hence cannot use autoremove since it will remove more components
-# even purge will remove some other components which is not ideal
-RUN apt-get update && \
-DEBIAN_FRONTEND=noninteractive apt-get purge -y --allow-unauthenticated \
-    composablekernel-dev \
-    miopen-hip
+# Make sure /opt/rcom is in the paths
```
BrianHarrisonAMD, line 108: nit: rcom -> rocm

BrianHarrisonAMD:
Looks good to me assuming it builds successfully.
Waiting on CI confirmation.

Dockerfile
adickin-amd: Will address with sccache changes.  This was preexisting and will require a full container rebuild.

vars/utils.groovy
```diff
@@ -293,13 +296,12 @@ def buildHipClangJob(Map conf=[:]){
...
+            if (lfs_pull) {
+                sh "git lfs pull --exclude="
+            }
             withDockerContainer(image: image, args: dockerOpts + ' -v=/var/jenkins/:/var/jenkins') {
                 timeout(time: 420, unit:'MINUTES')
                 {
-                    if (lfs_pull) {
-                        sh "git lfs pull --exclude="
-                    }
-
```
BrianHarrisonAMD: Wonder if this was done in the container before since we could guarantee git lfs was installed.
This will cause issues if it's missing on the machines.

adickin-amd: LFS is part of the default install.

BrianHarrisonAMD (approved):
Looks good to me.

I think we should test without the root user changes & git-lfs moved to see if it still resolves the issues we were seeing, but I am good either way depending on the results of the test.

Jenkinsfile
```diff
@@ -254,7 +254,7 @@ pipeline {
                     options {
                         retry(2)
                     }
-                    agent{ label rocmnode("gfx90a") }
+                    agent{ label rocmnode("gfx90a_disabled") }
```
adickin-amd: note all changes to this file will be reverted before merging.  Using this to target certain CI machines.

adickin-amd:
> Looks good to me.
> 
> I think we should test without the root user changes & git-lfs moved to see if it still resolves the issues we were seeing, but I am good either way depending on the results of the test.

Works without root and lfs changes on the selected machines.  Will revert my jenkinsfile changes and once a seconds review is done we should be good to merge.  This should not need another full day of CI runs.

adickin-amd:
Run 11 fully passed CI.

Dockerfile
```diff
@@ -1,155 +1,147 @@
...
-RUN pip3 install -r /doc-requirements.txt
-
-# Composable Kernel requires this version cmake
-RUN pip3 install --upgrade cmake==3.27.5
+    fi && \
+    ccache -s 
 
-#install miopen
-ARG INSTALL_MIOPEN=OFF
-ARG FRECKLE=0
```
cderb, line 152: It would be best to keep these down here to prevent a total docker rebuild when miopen version 
 or installation changes. FRECKLE is a commit hash used to trigger a rebuild for miopen.

cderb (changes requested):

Dockerfile
adickin-amd: Do you know if those args are intended more for developers building their own container or for the CI?(I only see it used as part of a perf test job, )  Based on how we compute the container name, any change to any build arg for the docker container will result in a complete rebuild.   i can put it lower but it will only affect local development.

cderb: It is mainly for the performance test CI. If the process building the docker has access to the build history then it significantly speeds up rebuilding to have the arg further down. This should be enforceable in the CI if the machine building the docker is set consistently.

cderb: Will address docker miopen install with a separate dockerfile in a new ticket.

cderb (approved):

PR #3566, opened 2025-02-28
[BN] run large size tensor test cases serially

bghimireamd:
Follow up of https://github.com/ROCm/MIOpen/pull/3553. In this PR we run 3D test case (The test case that caused the memory insufficient issue) serially.

When I listed all the shape's sizes in batch norm test case by descending order

Shape: (2, 2048, 16, 128, 128) --> Size: 1.07e+09
Shape: (69328, 1, 13, 79) --> Size: 7.12e+07
Shape: (64, 1024, 1024) --> Size: 6.71e+07
Shape: (64, 256, 56, 56) --> Size: 5.14e+07

Shape: (2, 2048, 16, 128, 128) --> Size: 1.07e+09
For now any test case with tensor size greater then 1e09 need to be run serially.


BrianHarrisonAMD (approved):
LGTM

PR #3572, opened 2025-03-04
Improvement of environment variables usage

### LINKED ISSUE #2674: Remove OS environmental variables from the tests
https://github.com/ROCm/MIOpen/pull/2655 showed that there are few tests which uses OS functions to access environmental variables.
That may not work as expected and all the environmental variables access should be converted into direct functions like in the #2553 

Check env.hpp for the list of available direct access functions.

### LINKED ISSUE #3037: Remove `setEnvironmentVariable` from the tests and use `env::update(VAR, value)` instead.
Based on https://github.com/ROCm/MIOpen/pull/2837#pullrequestreview-2107431371
`setEnvironmentVariable` should not be used in the tests, since it's an overkill and we don't need to update OS environment to get our internal state updated.  `setEnvironmentVariable` was misused on early stage of gtest transition, it was before we've got special functions for proper internal state updating.
Now `env::update(VAR, value)` does exactly what we need.

### LINKED ISSUE #3134: Remove environment variables from gtest
Remove environment variables from gtest (`MIOPEN_TEST_ALL`, `MIOPEN_TEST_FLOAT_ARG`, etc).
This must be done simultaneously for all the tests and this issue switches old ctest scheme to the new one.

It must be done after all the tests have been renamed in #3133 and new test wrapper has been implemented in #3132

averinevg:
This PR contains multiple changes related to the usage of environment variables. An interface for interacting with the library environment variables has been added. Also several long-standing issues have been fixed.

Fixes https://github.com/ROCm/MIOpen/issues/2674
Fixes https://github.com/ROCm/MIOpen/issues/3037
Fixes https://github.com/ROCm/MIOpen/issues/3134


src/env_debug.cpp
```diff
@@ -0,0 +1,204 @@
...
+                const auto res =
+                    std::from_chars(value.data(), value.data() + value.size(), ullvalue);
+                if(res.ec == std::errc::invalid_argument ||
+                   res.ec == std::errc::result_out_of_range)
+                {
+                    MIOPEN_THROW(miopenStatusInvalidValue,
+                                 "Invalid value for env variable: " + value);
+                }
+                miopen::env::update(var, ullvalue);
+            }
```
BrianHarrisonAMD: The previous code looked like it was doing an integral check before:

```
 if constexpr(std::is_integral_v<remove_cvref_t<U>> && !is_same_v<U, bool>)
 {
    setEnvironmentVariable(name_, std::to_string(value));
 }
```

Is this not needed / supported anymore?

src/include/miopen/env_debug.hpp
```diff
@@ -23,18 +23,26 @@
  * SOFTWARE.
  *
  *******************************************************************************/
-
 #pragma once
 
-#include <miopen/env.hpp>
+#ifdef MIOPEN_BUILD_TESTING
```
BrianHarrisonAMD, line 28: Why did this need to move from the test folder?

It looks like it's only used for testing.

src/env_debug.cpp
```diff
@@ -0,0 +1,204 @@
...
+    {MIOPEN_DEBUG_FIND_ONLY_SOLVER.GetName(), MIOPEN_DEBUG_FIND_ONLY_SOLVER},
+    {MIOPEN_DEBUG_TUNING_ITERATIONS_MAX.GetName(), MIOPEN_DEBUG_TUNING_ITERATIONS_MAX},
+    {MIOPEN_DEBUG_WORKAROUND_ISSUE_2492.GetName(), MIOPEN_DEBUG_WORKAROUND_ISSUE_2492},
+    {MIOPEN_ENABLE_LOGGING_CMD.GetName(), MIOPEN_ENABLE_LOGGING_CMD},
+    {MIOPEN_ENABLE_LOGGING_ELAPSED_TIME.GetName(), MIOPEN_ENABLE_LOGGING_ELAPSED_TIME},
+    {MIOPEN_FIND_ENFORCE.GetName(), MIOPEN_FIND_ENFORCE},
+    {MIOPEN_FIND_MODE.GetName(), MIOPEN_FIND_MODE},
+    {MIOPEN_LOG_LEVEL.GetName(), MIOPEN_LOG_LEVEL},
+    // clang-format on
+};
```
BrianHarrisonAMD: Does this need to match the env vars that are used in tests?

src/env_debug.cpp
averinevg: That code was more versatile; I'll add it here as well.

scerzh (approved):

test/gtest/conv_igemm_dynamic_xdlops_nhwc_nchw.cpp
```diff
@@ -28,21 +28,20 @@
...
-    env::update(MIOPEN_DEBUG_FIND_ONLY_SOLVER,
-                "ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC;"
-                "ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC;"
-                "ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC");
+    lib_env::update(MIOPEN_FIND_MODE, "normal");
+    lib_env::update(MIOPEN_DEBUG_FIND_ONLY_SOLVER,
+                    "ConvAsmImplicitGemmGTCDynamicFwdXdlopsNHWC;"
+                    "ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC;"
+                    "ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC");
 }
```
BrianHarrisonAMD, line 45: I think tests that set environment variables like this will still be an issue for the single gtest binary runs right?

I think we would need to use the scoped setters to ensure the env settings get reset between runs to fix the single gtest binary.

Is that right, or did I miss some magic that is making it work?

BrianHarrisonAMD (approved):

test/gtest/conv_igemm_dynamic_xdlops_nhwc_nchw.cpp
averinevg: Currently, we set these variables where they affect the test itself. This is the only reason why it works. Scoped setters are definitely needed, they are present in the new unit tests.

src/env_debug.cpp
averinevg: Yes. If an attempt is made to modify an unknown variable, the exception occurs.

PR #3583, opened 2025-03-06
fix bn tuning issue with CK batchnorm solvers

bghimireamd:
~~* enabled CK solver~~
* Fixed tuning issue with CK batch norm
* batch norm driver iter default to 10 (similar to conv). Before it was 1.


src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -313,11 +313,10 @@ ConvSolution InitAnyInvokerFactory(const ProblemDescriptionType& problem,
...
                 auto invoker_ptr     = sh_conv_ptr->MakeInvokerPointer();
 
-                const auto enable_profiling = handle.IsProfilingEnabled();
-                float elapsed_time =
-                    invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), enable_profiling});
-                if(enable_profiling)
+                invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), false});
+                if(handle.IsProfilingEnabled())
                 {
+                    float elapsed_time = handle.GetKernelTime();
```
BrianHarrisonAMD: Since we are disabling the timing from CK we will need to track the kernel time ourself.

I think we need to add something like:

```
{
           WorkAroundHipEventProfiler prf(handle);
           invoker_ptr->Run(argument_ptr.get(), {handle.GetStream(), false});
}

BrianHarrisonAMD (approved):
LGTM

PR #3586, opened 2025-03-07
[tests] Unit test for #3572

averinevg:
This PR adds unit test for https://github.com/ROCm/MIOpen/pull/3572


test/gtest/unit_lib_env_var.cpp
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+private:
+    std::string_view name;
+    std::optional<std::string> prev;
+};
+
+} // namespace
+
+TEST(CPU_UnitTestLibEnvVarGetUnknownVariable_NONE, LibEnvVar)
```
Vsevolod1983: Is it possible to use TEST_P and suite instead of plain TEST macro, so we have better control on Smoke/Full runs?

scerzh (approved):

test/gtest/unit_lib_env_var.cpp
averinevg: Yes, it is possible. Already done.

Vsevolod1983 (approved):

PR #3603, opened 2025-03-11
Update CK commit hash for staging

AviralGoelAMD:
This PR updates the CK commit hash in requirements.txt and Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,10 +115,10 @@ RUN echo Building for GPU Archs: ${GPU_ARCHS} && \
     cd composable_kernel-${CK_COMMIT} && \
     mkdir build && cd build && \
     num_threads=$(nproc) && \
-    if [ "$num_threads" -gt 128 ]; then \
-        num_threads=64; \
-    elif [ "$num_threads" -gt 32 ]; then \
+    if [ "$num_threads" -lt 32 ]; then \
         num_threads=$(( num_threads / 2 )); \
+    else \
+        num_threads=32; \
     fi && \
     echo Building CK with ${num_threads} threads && \
     CXX=/opt/rocm/bin/amdclang++ cmake \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -115,10 +115,10 @@ RUN echo Building for GPU Archs: ${GPU_ARCHS} && \
     cd composable_kernel-${CK_COMMIT} && \
     mkdir build && cd build && \
     num_threads=$(nproc) && \
-    if [ "$num_threads" -gt 128 ]; then \
-        num_threads=64; \
-    elif [ "$num_threads" -gt 32 ]; then \
+    if [ "$num_threads" -lt 32 ]; then \
```
afagaj, line 118: It's a bit weird the jump from nproc=31 to nproc=32 (number of build threads jumps from 16 to 32).  

Would it be better to do something like this?
```
-    if [ "$num_threads" -lt 32 ]; then \
+    if [ "$num_threads" -lt 64 ]; then \
```

BrianHarrisonAMD: Yea it will have a big jump, and still use 100% of the cores if you used a machine that had exactly 32 threads. 
Could change it to -lt 64 we use half, and otherwise cap at 32. 
I think that makes more sense.

BrianHarrisonAMD: The docker hash would change though I think we can make that change a separate follow-up to avoid a rebuild hehe.

JonathanLichtnerAMD:
This is probably a comment for the scripts that create this PR (if you haven't changed this already) but the commit message

`Update CK commit hash in requirements.txt and Dockerfile`

should just be 

`Update CK commit hash in requirements.txt`

BrianHarrisonAMD (approved):

JonathanLichtnerAMD (approved):

PR #3604, opened 2025-03-11
Find Mode TrustVerify

cderb:
Additional MIOPEN_FIND_MODE = 6 (TrustVerify)
This mode extends DynamicHybrid

Running with TrustVerify will first attempt to load tuning results from system resources
  If no solution is returned tuning will be triggered
  If a solution is retuned the user find db will be checked for the solution
     If solution is from the user find db it will be used
     If a solution is from the system (db or model), the solution will be evaluated and the new time will be compared to the time reported by the solution
       If the evaluated / reported time is less than the tolerance threshold then the system result is added to the user db and the solution is returned
       If the evaluated / reported time exceeds the tolerance, then tuning will be triggered

This find mode will ensure that all configurations are tuned for the deployment system. Solutions from system resources are verified once and tuned if markedly different from expectation. Results from user dbs are considered reliable and used without further verification.


include/miopen/miopen.h
```diff
@@ -664,6 +664,8 @@ typedef enum
               miss, uses the existing Find machinery with skipping non-dynamic kernels, thus saving
               compilation time. Faster start-up times than Hybrid Find, but GPU performance might be
               a bit worse. >*/
+    miopenConvolutionFindModeTrustVerify =
+        6,
```
BrianHarrisonAMD: Needs docs like the others.
The PR description is probably pretty close to what is needed!

cderb:
The performance of find mode TrustVerify is between the performance of find mode DynamicHybrid and find enforce SearchDbUpdate.
TrustVerify will have the performance of SearchDbUpdate when
- There is no system db or model solution
- The solution returned from system is a degree slower than reported by the solution itself based on a 1 time evaluation
TrustVerify will have the performance of DynamicHybrid when
- The solution returned from system resource performs as expected based on 1 time evaluation
- Solution is present in user db

First run with TrustVerify is generally slower than DynamicHybrid. 
After the first run, the user find db will be fully populated and user perf db populated with regenerated entries. All following runs should exhibit ideal runtimes.

docs/conceptual/tuningdb.rst
```diff
@@ -29,7 +29,6 @@ Enable this feature using these commands:
 
 .. code:: bash
 
-  export MIOPEN_FIND_MODE=3
```
cderb, line 32: Found that setting find enforce disables find mode inputs. So removing these unnecessary lines.

docs/how-to/find-and-immediate.rst
```diff
@@ -237,6 +237,12 @@ modes by using the ``MIOPEN_FIND_MODE`` environment variable with one of these v
   entry. If there's a FindDb hit, it uses that entry. If there's a miss, it uses the existing find machinery,
   skipping non-dynamic kernels. It offers faster start-up times than hybrid find, but GPU performance
   might decrease.
+* ``TRUST_VERIFY``/``6`` (trust verify find): Checks :doc:`FindDb <../conceptual/finddb>` for an entry.
+  If there's a UserFindDb hit, it uses that entry.
+  If there's a FindDb hit the result is evaluated. If the ratio of evaluated to reported result time is
```
amd-jnovotny:
```suggestion
  If there's a FindDb hit, the result is evaluated. If the ratio of evaluated to reported result time is
```

docs/how-to/find-and-immediate.rst
```diff
@@ -237,6 +237,12 @@ modes by using the ``MIOPEN_FIND_MODE`` environment variable with one of these v
...
+  below the tolerance threshold, the result is used and added to the UserFindDb. Otherwise tuning will be triggered.
+  If there's a miss, tuning will be triggered, skipping non-dynamic kernels.
+  This mode may have slow start-up times, but typically selects the most performant solutions.
```
amd-jnovotny:
```suggestion
  This mode can have slow start-up times but typically selects the most performant solutions.
```

amd-jnovotny (approved):
Looks good. I've made a couple of minor editorial suggestions.

include/miopen/miopen.h
cderb: added doc

src/ocl/convolutionocl.cpp
```diff
@@ -164,12 +277,107 @@ std::vector<Solution> FindConvolution(const ExecutionContext& ctx,
...
-        /// \todo Consider if we need (and want to spend time) for this.
-        const auto id = solver::Id{sol->solution_id};
-        const auto& s = id.GetSolver();
-        CompileSolution(id, ctx, problem);
-        results.push_back({id, sol->time, s.GetWorkspaceSize(ctx, problem)});
+        if(findMode.IsTrustVerify(ctx))
+        {
+            // is user find db record?
+            auto ufdb_sols = miopen::GetSolutions<UserFindDb>(ctx, problem, 1, &invoke_ctx);
+            if(ufdb_sols.empty())
```
BrianHarrisonAMD: Looks like we need to read the DB twice.
I don't think it's a huge deal since it'll be cached, but if we knew which DB the result came from we could avoid the extra read.

src/ocl/convolutionocl.cpp
```diff
@@ -164,12 +277,107 @@ std::vector<Solution> FindConvolution(const ExecutionContext& ctx,
...
+                        EvaluateInvokers(handle,
+                                         conv_sols,
+                                         algo,
+                                         problem.MakeNetworkConfig(),
+                                         invoke_ctx,
+                                         is_optimal,
+                                         false);
+
+                    const float eval_time            = eval_sols.front().GetTime();
+                    constexpr float VERIFY_TOLERANCE = 1.10f;
```
BrianHarrisonAMD: 10% is pretty tight, but maybe we are okay with that?

BrianHarrisonAMD (approved):
LGTM just a couple comments to consider, but I don't think they are blocking.

src/ocl/convolutionocl.cpp
cderb: Correct, we need to know if the entry is from the system or the user db, as there are different actions depending. There doesn't appear to be a return from the GetSolutions function presently which would indicate that.

src/ocl/convolutionocl.cpp
cderb: It's ok to loosen this tolerance. I've also seen much higher degrees of variation even when the selected solver and solver parameters are correct.

BradPepersAMD:
Can we get a list of current issues this should fix or improve?  As well as any situations that this may cause drops compared to existing results?  In particular, I'm thinking of cases on distributed runs where they may be starting with a blank user db on every node.

cderb:
This find mode option will function like MIOPEN_FIND_ENFORCE=3 in the worst case. Worst case is when either there is no entry or the system entry is much slower than advertised.
In the best case the entry is either in the user db or the system entry is acceptable (at which point the system entry becomes a user entry). This would be a simple db recall.
This strategy is best for long running workloads as it guarantees that configurations used are optimal.
Assuming the system entries are perfect this mode will be slower on first run than the current default DYNAMIC_HYBRID. This is due to the overhead of benchmarking the kernels to verify their runtimes.

I had wanted to split the option for tuning individual solvers to MIOPEN_FIND_ENFORCE, but this env overrides MIOPEN_FIND_MODE. So presently TRUST_VERIFY also effectively enforces MIOPEN_FIND_ENFORCE=SEARCH_DB_UPDATE, which can take quite some time. Having another option that forgoes individual solver tuning may give better results for shorter running applications and would cap the runtime closer to find mode DYNAMIC_HYBRID.

src/include/miopen/generic_search.hpp
```diff
@@ -306,6 +307,48 @@ GetAllSolutions(const Solver s, const Context& context_, const Problem& problem)
 std::size_t GetTuningIterationsMax();
 std::chrono::milliseconds GetTuningTimeMax(); // returns the max allowed time in milliseconds
 std::size_t GetTuningThreadsMax();
+std::size_t GetTuningPatience();
+
+template <typename Context>
+std::chrono::milliseconds GetTuningTimeMax(const Context& ctx,
```
cderb, line 313: This is actually a limit on compilation time, not tuning time outright.

include/miopen/miopen.h
```diff
@@ -664,8 +664,10 @@ typedef enum
               miss, uses the existing Find machinery with skipping non-dynamic kernels, thus saving
               compilation time. Faster start-up times than Hybrid Find, but GPU performance might be
               a bit worse. >*/
+    miopenConvolutionFindModeTrustVerify     = 6,
+    miopenConvolutionFindModeTrustVerifyFull = 7,
     miopenConvolutionFindModeDefault =
-        miopenConvolutionFindModeDynamicHybrid /*!< Default FindMode >*/
+        miopenConvolutionFindModeTrustVerify /*!< Default FindMode >*/
```
cderb: This is updating the default find mode.

iq136boy (approved):

src/include/miopen/generic_search.hpp
```diff
@@ -306,6 +307,48 @@ GetAllSolutions(const Solver s, const Context& context_, const Problem& problem)
...
+                                           const miopen::conv::ProblemDescription& problem)
+{
+    const auto& conv     = problem.GetConv();
+    const auto& findMode = conv.findMode;
+    auto tuningMs        = env::value(MIOPEN_TUNING_TIME_MS_MAX);
+    if(findMode.IsTrustVerify(ctx) && !findMode.IsExhaustive(ctx))
+    {
+        if(tuningMs == DEFAULT_MIOPEN_TUNING_TIME_MS_MAX)
+            tuningMs = 1000;
+    }
```
cderb: Setting max compilation time with TrustVerify find mode.

src/include/miopen/generic_search.hpp
```diff
@@ -306,6 +307,48 @@ GetAllSolutions(const Solver s, const Context& context_, const Problem& problem)
...
+std::size_t GetTuningPatience(const Context& ctx, const miopen::conv::ProblemDescription& problem)
+{
+    const auto& conv     = problem.GetConv();
+    const auto& findMode = conv.findMode;
+    auto patience        = env::value(MIOPEN_TUNING_PATIENCE);
+    if(findMode.IsTrustVerify(ctx) && !findMode.IsExhaustive(ctx))
+    {
+        if(patience == DEFAULT_MIOPEN_TUNING_PATIENCE)
+            patience = 6;
+    }
```
cderb: Setting a tuning patience with TrustVerify find mode. Iterations without improvement.

BrianHarrisonAMD:
Feedback:
- I think we need a check, that will not use this mode if the userDB isn't writable / accessible.
- I think we might need some more data, or just pick a less restrictive range for determining to retune.
- Add setting the configurable knobs through API calls instead of just env variables to make it settable for users programatically.
- If / when we move forward with this we need to coordinate with pytorch, and determine how the benchmark flags should behave. 
    -  My proposal is:
         - Default this is on, with a pretty forgiving range for determining the systemDB entry is good, and not too high of a patience setting / max tuning time setting.
         - When the benchmark flag is set in Pytorch, we should reduce the allowed range for accepting a systemDB entry, and then also increase the patience / max tuning time setting.
              - Might even be worth chatting about this, as it might make sense to have benchmark actually invalidate userDB entries as well come to think of it.

JonathanLichtnerAMD (approved):

src/include/miopen/find_controls.hpp
```diff
@@ -152,13 +154,35 @@ class MIOPEN_INTERNALS_EXPORT FindMode
...
+    bool IsTrustVerify(const Context& context) const
+    {
+    //TrustVerify uses user db as groud truth, disable if no user db
+#if MIOPEN_DISABLE_USERDB
+        return false
+#else
+        return (value == Values::TrustVerify || value == Values::TrustVerifyFull) &&
+               IsEnabled(context);
+#endif
+    }
```
cderb: Fall back to dynamic hybrid when user db is disabled.

src/include/miopen/generic_search_controls.hpp
```diff
@@ -47,3 +49,5 @@ MIOPEN_DECLARE_ENV_VAR_UINT64(MIOPEN_COMPILE_PARALLEL_LEVEL,
                               std::thread::hardware_concurrency() / 2)
 #endif
 MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_COMPILE_ONLY)
+
+MIOPEN_DECLARE_ENV_VAR_UINT64(MIOPEN_VERIFY_TOLERANCE_PCT, 15)
```
cderb: Verify tolerance setting moved to env var.

src/ocl/convolutionocl.cpp
```diff
@@ -140,6 +253,120 @@ static void ShrinkToFind10Results(std::vector<Solution>& found)
...
+        AutoEnableProfiling enableProfiling{handle};
+        bool is_optimal = true;
+
+        std::vector<Solution> eval_sols;
+        auto db = MakeConvDbGetter(ctx);
+        // reverse solutions so that EvaluateInvokers registers the fastest solution last
+        auto sol_itr = solutions.rbegin();
+        auto sol_end = solutions.rend();
+        if(!model_result)
+            sol_itr = solutions.rend() - 1;
```
cderb: Made correction to iterator, now correct solution is evaluated and compared to previous db time. Previously the wrong solver was compared, and this was causing extra tunings

src/ocl/convolutionocl.cpp
```diff
@@ -150,26 +377,61 @@ std::vector<Solution> FindConvolution(const ExecutionContext& ctx,
...
+        if(findMode.IsTrustVerify(ctx))
+        {
+            ufdb_sols = miopen::GetSolutions<UserFindDb>(ctx, problem, 1, &invoke_ctx);
+            if(!ufdb_sols.empty())
+                sols = ufdb_sols;
+            else
+                sols = conv.GetSolutions(ctx, problem, 2, &fallback, &invoke_ctx);
+        }
+        else
+            sols = conv.GetSolutions(ctx, problem, 1, &fallback, &invoke_ctx);
```
cderb, line 431: Returning more than 1 solution from GetSolutions can cause warning message for workspace size in hybrid dynamic mode. Separated paths here.

cderb (commented):
@BrianHarrisonAMD commented updates.

src/include/miopen/generic_search.hpp
```diff
@@ -469,7 +512,6 @@ auto GenericSearch(const Solver s,
                 break;
             }
 
-            last_imprv++;
```
BrianHarrisonAMD, line 472: I guess this was wrong before?

BrianHarrisonAMD: I guess it didn't really matter, but was in an odd spot.

cderb: Yes, this was causing an issue, because end signals from the solution_queue were allowing processes to come back and increment last_imprv.

include/miopen/miopen.h
```diff
@@ -665,8 +665,10 @@ typedef enum
...
```
randyspauldingamd: What's the schedule for Unit Testing? It needs to have Unit Tests that exercise every possible path, including output/DB updates. It would be best to have Unit Tests prior to merge, but at a minimum they must be in place before making this the default.

randyspauldingamd (changes requested):
Needs Unit Tests prior to becoming the default mode.

include/miopen/miopen.h
cderb: Looks like the same level of rigor wasn't used to install dynamichybrid as the default. Test will need to be written from scratch.

src/ocl/convolutionocl.cpp
```diff
@@ -140,6 +253,122 @@ static void ShrinkToFind10Results(std::vector<Solution>& found)
     found = std::move(out);
 }
 
+std::vector<Solution> VerifiedFDBSolution(const ExecutionContext& ctx,
```
cderb, line 351: The main change for this find mode is within this function and in the search enforcement on lines 449, 450. @randyspauldingamd the unit testing of VerifiedFDBSolution might be sufficient for this find mode addition. Though this is complicated by the reliance on db paths which are not programmatically settable. Alternatively we might reduce this into smaller functions which lend themselves better to testing.

include/miopen/miopen.h
```diff
@@ -665,6 +665,8 @@ typedef enum
               miss, uses the existing Find machinery with skipping non-dynamic kernels, thus saving
               compilation time. Faster start-up times than Hybrid Find, but GPU performance might be
               a bit worse. >*/
+    miopenConvolutionFindModeTrustVerify     = 6,
+    miopenConvolutionFindModeTrustVerifyFull = 7,
     miopenConvolutionFindModeDefault =
         miopenConvolutionFindModeDynamicHybrid /*!< Default FindMode >*/
```
cderb, line 671: Resetting default mode to dynamic hybrid.
Will consider find mode 6 when user find db is usable in a networked file system.

randyspauldingamd (approved):
LGTM!

PR #3605, opened 2025-03-11
Docker/Jenkins build speed improvements.

adickin-amd:
- Add SCCache to build CK in the Dockerfile.
- Split perf test build out of main Dockerfile into its own.  
   - Perf tests now build based on the main Dockerfile and require much less time building a container.
- Set the depth of git checkout to clone less history of the repository.
   - Note: Investigating this change closer, the original clone might come from Jenkins and not the checkout steps in the pipeline file. 
- Fix the CK build by removing a set of ""s from the arg inputs which causes _'s to get prepended and appended to the cmake command.


adickin-amd:
Full Perf job run done here: http://micimaster.amd.com/blue/organizations/jenkins/MiOpenPerf_adickin/detail/MiOpenPerf_adickin/3/pipeline/

vars/utils.groovy
```diff
@@ -243,13 +239,41 @@ def getDockerImage(Map conf=[:])
             dockerImage.push()
         }
     }
+
+
+    if(params.INSTALL_MIOPEN == 'ON')
+    {
+        def freckle = sh(script: 'git rev-parse --short HEAD', returnStdout: true).trim()
+        dockerArgs = ""
+        dockerArgs = dockerArgs + " --build-arg BASE_DOCKER=${image} --build-arg FRECKLE=${freckle} -f Dockerfile.perftests"
```
cderb: Looks like this can be cleaned up.

adickin-amd: good catch.  Funny way I set that up.

Dockerfile
```diff
@@ -105,57 +119,63 @@ RUN wget https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.
...
 # Composable Kernel installed separated from rbuild to take in values from GPU_ARCHS 
 RUN sed -i '/composable_kernel/d' /requirements.txt
 
 # rbuild is used to trigger build of requirements.txt, dev-requirements.txt
 RUN if [ "$USE_FIN" = "ON" ]; then \
-        rbuild prepare -s fin -d $PREFIX -DGPU_ARCHS="${GPU_ARCHS}" -DCMAKE_CXX_COMPILER_LAUNCHER="${COMPILER_LAUNCHER}"; \
+    rbuild prepare -s fin -d $PREFIX -DGPU_ARCHS="${GPU_ARCHS}"; \
     else \
-        rbuild prepare -s develop -d $PREFIX -DGPU_ARCHS="${GPU_ARCHS}" -DCMAKE_CXX_COMPILER_LAUNCHER="${COMPILER_LAUNCHER}"; \
+    rbuild prepare -s develop -d $PREFIX -DGPU_ARCHS="${GPU_ARCHS}"; \
```
BrianHarrisonAMD, line 175: nit:

Should these be indented?

script/sccache_wrapper.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
BrianHarrisonAMD, line 1: Does this script match CK's version?

vars/utils.groovy
```diff
@@ -243,13 +250,41 @@ def getDockerImage(Map conf=[:])
             dockerImage.push()
         }
     }
+
+
+    if(params.INSTALL_MIOPEN == 'ON')
+    {
+        def freckle = sh(script: 'git rev-parse --short HEAD', returnStdout: true).trim()
+        dockerArgs = " --build-arg BASE_DOCKER=${image} --build-arg FRECKLE=${freckle} -f Dockerfile.perftests"
```
BrianHarrisonAMD, line 260: Do we need freckle anymore?
I thought it was to make the docker image name unique, but now we have the _perfTest image name that will make it unique.

Dockerfile
```diff
@@ -105,57 +119,63 @@ RUN wget https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.
...
+    export COMPILERS_HASH_DIR=/tmp/.sccache && \
+    export SCCACHE_BIN=/usr/local/.cargo/bin/sccache && \
+    export SCCACHE_EXTRAFILES=/tmp/.sccache/rocm_compilers_hash_file && \
+    export SCCACHE_REDIS="redis://$MIOPEN_SCCACHE" && \
+    echo "connect = $MIOPEN_SCCACHE" >> redis-cli.conf && \
+    export SCCACHE_C_CUSTOM_CACHE_BUSTER="${MIOPEN_SCCACHE_CUSTOM_CACHE_BUSTER}" && \
+    echo $SCCACHE_C_CUSTOM_CACHE_BUSTER && \
+    stunnel redis-cli.conf && \
+    export PATH=$PATH:${SCCACHE_INSTALL_LOCATION} && \
+    ./sccache_wrapper.sh --enforce_redis; \
```
BrianHarrisonAMD, line 146: nit:

Should be indent this?

Dockerfile
adickin-amd: ill check, I think vscode auto format took it out.  I can force it back.

script/sccache_wrapper.sh
adickin-amd: Yes, this is directly from CK

adickin-amd: I copied as I was thinking I would eventually setup MIOpen CI with sccache.

vars/utils.groovy
adickin-amd: I think you still want it for perf tests.  The main dockerfile wont change often but if you run perf tests against several commits I think you still need something to make it rebuild.

BrianHarrisonAMD: Oh I see.

cderb: The issue is re-triggering the docker build when the miopen branch is updated. The Dockerfile is not otherwise able to discern updates to the git branch. And the Dockerfile is simply seating the miopen from the project directory. We could force the docker to build without cache, but this is also not exactly right.

BrianHarrisonAMD (approved):

cderb (approved):

PR #3606, opened 2025-03-12
Rjs/test 950

randyspauldingamd:
Fixing test failures on gfx950


BrianHarrisonAMD (approved):

src/solver/conv/conv_multipass_wino3x3WrW.cpp
```diff
@@ -487,7 +487,8 @@ bool ConvWinograd3x3MultipassWrW<WinoDataH, WinoFilterH, WinoDataW, WinoFilterW>
                                                                                        problem)))
         return false;
 
-    if(!(StartsWith(name, "gfx8") || StartsWith(name, "gfx9")) || name == "gfx942")
+    if(!(StartsWith(name, "gfx8") || StartsWith(name, "gfx9")) || name == "gfx942" ||
+       name == "gfx950")
```
averinevg:
```suggestion
    if(!(StartsWith(name, "gfx8") || name == "gfx900" || name == "gfx906"  || name == "gfx908" || name == "gfx90a"))
```

test/gtest/gtest_common.hpp
```diff
@@ -63,9 +63,10 @@ enum class Gpu : int
     gfx908  = 1 << 2,
     gfx90A  = 1 << 3,
     gfx94X  = 1 << 4,
-    gfx103X = 1 << 5,
-    gfx110X = 1 << 6,
-    gfx120X = 1 << 7,
+    gfx95X  = 1 << 5,
```
averinevg:
```suggestion
    gfx950  = 1 << 5,
```

averinevg (changes requested):

src/solver/conv/conv_multipass_wino3x3WrW.cpp
randyspauldingamd: So this never should have been applied to gfx942?

BrianHarrisonAMD: It was avoiding gfx942, and gfx950, but it's a bit confusing.

The change here should be the same, but makes it explicitly avoid ASICs that aren't gfx8*, gfx900, gfx906, gfx908, or gfx90a.

BrianHarrisonAMD (approved):

PR #3610, opened 2025-03-12
MIOpenDriver moving to lib/miopen

arvindcheru:
Proposed Changes:
- Moving MIOpenDriver binary to lib/miopen as part of an effort to move to install folder/file based on type of binary
- MIOpenDriver considering as internal binary proposed to move to lib/<pkgnm>/ folder 
- Updating invocations of this binary to point to new proposed file path.


averinevg (commented):
@arvindcheru Could you please describe in more detail what is wrong at the moment and what we will gain from these changes?

averinevg (approved):

BrianHarrisonAMD (approved):

driver/CMakeLists.txt
```diff
@@ -83,5 +83,5 @@ endif()
 if( NOT ENABLE_ASAN_PACKAGING )
   install(TARGETS MIOpenDriver 
       PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
-      DESTINATION ${CMAKE_INSTALL_BINDIR})
+      DESTINATION ${CMAKE_INSTALL_LIBDIR}/${MIOPEN_INSTALL_DIR})
```
JonathanLichtnerAMD, line 86: I don't think is right that we are moving a binary to a lib directory.  We should find a better location for the MIOpen driver, something that fits better with the Linux Filesystem Standard (see https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard).  Maybe .../rocm/usr/bin?

BrianHarrisonAMD:
I think the miopen_gtest binary also installs to /opt/rocm/bin.

Is that fine, or should that be moved as well?

sudo apt install miopen-hip-clients for example.

driver/CMakeLists.txt
nunnikri: We are following FHS specification itself. https://www.pathname.com/fhs/pub/[fhs-2.3.pdf](https://www.pathname.com/fhs/pub/fhs-2.3.pdf)
Insted of /usr ROCm install everything in /opt/rocm. Inside rocm folder we follow FHS standard for /usr. 
"/usr/lib includes object files, libraries, and internal binaries that are not intended to be executed directly by
users or shell scripts."

So all internal binaries which are not meant to be used by users should be in lib/ folder itself.

JonathanLichtnerAMD: I couldn't access the link above but version 3.0 https://refspecs.linuxfoundation.org/FHS_3.0/[fhs-3.0.pdf](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.pdf) indicates the following for /usr/lib:

> /usr/lib includes object files and libraries. 1
 On some systems, it may also include internal binaries
that are not intended to be executed directly by users or shell scripts. 2

Both lines are footnoted but the implication is that the binaries that are placed in /usr/lib (or /usr/libexec) are for "binaries run by other programs" (see the section title for /use/libexec).  The binaries that are put in /usr/lib or /usr/libexec are run by other programs like a library and *are* not run by users.

The MIOpenDriver is actually documented externally (see https://rocm.docs.amd.com/projects/MIOpen/en/latest/install/install.html#building-the-driver), and could be run by end users.  We run it internally all the time, and I suspect external contributors might also want to run it, so it doesn't really fit the definition of what should go in a "lib" or "libexec" directory.  I'm okay with moving it somewhere else, but it's a binary that gets run by (at least internal) users, so let's put it someplace else other than in a "lib" or "libexec" subpath.  Thanks

nunnikri: My understading was MiOpenDriver was used only internally, in that case lib is the right folder(internally used compiled code). But If its meant to be used by external users then we should retain in bin folder itself and discard this PR

JonathanLichtnerAMD: I asked the team just to make sure my statement above was correct, and @BrianHarrisonAMD mentioned the following:

> I think it's only really used externally if we are trying to troubleshoot something, or the user is trying to tune some specific shapes.  I wouldn't expect an external user to know about it, or use it generally.
>
> We do mention it in the tuning documentation though.

So it's not usually used externally, but it is documented and might be used sometimes.  Is there a location for binaries that are usually not used externally but sometimes might be?

nunnikri: If its there in external documentation, i would leave it in bin folder itself.

arvindcheru: Thank you for all review, and information shared. So based on the input learn that we dont need to move from bin folder, shall we abandon this PR.

pfultz2: If you want to install it in a non-public location, it would be better to use `rocm_install_targets(MIOpenDriver PRIVATE)` from rocm-cmake. This will install the binary in lib/miopen/bin and will adjust the rpaths to find the correct binaries. The exact path becomes a rocm-cmake problem that miopen doesn't need to worry about.

In general, private executables should be for executables that are used by the library, but users should never call it directly(it is common to have a private executable to do hiprtc out-of-process to get parallel compilation like in rocfft and migraphx). However, since the driver is a user facing tool, that is not called by miopen, then it should really be in the bin directly.

arvindcheru: Thank you @pfultz2 and all reviewers for the comments. Will be abandoning this change request. (not changing the current install folder (bin))

PR #3613, opened 2025-03-13
[BN] Initial tuning for OpenCL kernels

EnricoDeg:
Summary:

 - check all applicable variants during tuning and pick the fastest (currently 2 tuning parameters: variant and vector size)
 - improve heuristics of fwd spatial single to use all variants (now very similar to the ones for bwd)
 - merge spatial single and multiple solvers in spatial solver because tuning is done per solver

Notes:

 - setting `MIOPEN_FIND_MODE=1` and `MIOPEN_FIND_ENFORCE=4` enables tuning and all the applicable configurations are run. Then the fastest is stored in a database and after unsetting the env variables, the fastest configuration is picked as default (same method used to  tune CK instances)
 - without running tuning the default configuration is picked based on heuristics


src/solver.cpp
```diff
@@ -562,19 +562,18 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
     Register(registry, ++id, Primitive::Activation, activ::ActivBwdSolver0{}.SolverDbId());
     Register(registry, ++id, Primitive::Activation, activ::ActivBwdSolver1{}.SolverDbId());
 
-    RegisterWithSolver<batchnorm::BnFwdTrainingSpatialSingle>(registry, ++id, Primitive::Batchnorm);
+    // combine BnFwdTrainingSpatialMultiple and BnFwdTrainingSpatialSingle
+    RegisterWithSolver<batchnorm::BnFwdTrainingSpatial>(registry, ++id, Primitive::Batchnorm);
```
averinevg, line 566: Wouldn't it be better to use a new id?

src/solver.cpp
```diff
@@ -562,19 +562,18 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
...
-        registry, ++id, Primitive::Batchnorm);
-
+    ++id; // removed solver BnFwdTrainingSpatialMultiple (it is now part of BnFwdTrainingSpatial)
     RegisterWithSolver<batchnorm::BnFwdTrainingPerActivation>(registry, ++id, Primitive::Batchnorm);
 
-    RegisterWithSolver<batchnorm::BnBwdTrainingSpatialSingle>(registry, ++id, Primitive::Batchnorm);
-    RegisterWithSolver<batchnorm::BnBwdTrainingSpatialMultiple>(
-        registry, ++id, Primitive::Batchnorm);
+    // combine BnBwdTrainingSpatialMultiple and BnBwdTrainingSpatialSingle
+    RegisterWithSolver<batchnorm::BnBwdTrainingSpatial>(registry, ++id, Primitive::Batchnorm);
```
averinevg, line 575: Ditto

src/solver.cpp
EnricoDeg: Does it mean that I should add `++id` there and then add `RegisterWithSolver<batchnorm::BnFwdTrainingSpatial>(registry, ++id, Primitive::Batchnorm);` at the end of the file?

I wanted to avoid having the batch norm solvers ids scattered in the file

averinevg: I would consider the renamed solver with extended applicability as a new one. But in this case, it probably doesn't affect anything.

bghimireamd (approved):

averinevg:
I guess this PR can be closed since https://github.com/ROCm/MIOpen/pull/3642 has been merged.

bghimireamd:
> I guess this PR can be closed since #3642 has been merged.

yes!

PR #3616, opened 2025-03-13
2D and 3D NGCHW forward convolution solvers based on CK instances

iq136boy:
Implement 2d and 3d NGCHW forward convolution solvers base on CK NGCHW layout instances.
The solver utilize the fused CK convolution and layout transform kernels to avoid the tensor layout transform in MIOpen side.


averinevg:
@iq136boy Please add also `unit_conv_solver_*` tests for these solvers.

averinevg (changes requested):
Review comments on https://github.com/ROCm/MIOpen/pull/3617 are also applicable to this PR.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_fwd_nchw_xdlops.cpp
```diff
@@ -0,0 +1,557 @@
...
+namespace miopen {
+namespace solver {
+namespace conv {
+
+using ProblemDescription = miopen::conv::ProblemDescription;
+
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+
+using InLayoutNCHW                         = ck::tensor_layout::convolution::NGCDHW;
+using WeiLayout                            = ck::tensor_layout::convolution::GKZYXC;
```
BrianHarrisonAMD: Layout for the weights needs to be GKCZYX (same as WRW)

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_fwd_nchw_xdlops.cpp
```diff
@@ -0,0 +1,557 @@
...
+        Z               = ProblemInterpreter::GetFilterDepthZ(problem);
+        alpha_beta_case = ProblemInterpreter::GetAlphaBetaCase(problem);
+
+        in_lengths  = {G, N, C, Di, Hi, Wi};
+        out_lengths = {G, N, K, Do, Ho, Wo};
+        wei_lengths = {G, K, C, Z, Y, X};
+
+        in_strides  = {Di * Hi * Wi * C, Di * Hi * Wi * G * C, Di * Hi * Wi, Hi * Wi, Wi, 1};
+        out_strides = {Do * Ho * Wo * K, Do * Ho * Wo * G * K, Do * Ho * Wo, Ho * Wo, Wo, 1};
+        wei_strides = {K * Z * Y * X * C, Z * Y * X * C, 1, Y * X * C, X * C, C};
```
BrianHarrisonAMD: Just reminder to flip lengths and strides once layout changes.

BrianHarrisonAMD: Also, very surprised this isn't causing errors.
Looks like we are assuming the data is in NHWC layout, but it'll come in as NCHW layout from MIOpen side.

BrianHarrisonAMD:
I think we need to double check the verification is working properly for these tests.
It looks like the CK args would treat the MIOpen weights as a different layout than they are, and I am surprised that doesn't cause issues.

iq136boy:
Close and will reopen if needed.

PR #3617, opened 2025-03-13
2D and 3D NGCHW bacward weight convolution solvers based on CK instances

iq136boy:
Implement 2d and 3d NGCHW backward convolution solvers base on CK NGCHW layout instances.


averinevg:
@iq136boy Please add also `unit_conv_solver_*` tests for these solvers.

src/CMakeLists.txt
```diff
@@ -266,6 +266,8 @@ set( MIOpen_Source
     solver/conv/conv_hip_implicit_gemm_3d_grouped_fwd_xdlops.cpp
     solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
     solver/conv/conv_hip_implicit_gemm_3d_grouped_bwd_xdlops.cpp
+    solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
+    solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```
averinevg: This needs to be sorted

src/solver.cpp
```diff
@@ -707,6 +707,14 @@ inline SolverRegistrar::SolverRegistrar(IdRegistryData& registry)
              multimarginloss::MultiMarginLossForward{}.SolverDbId());
 
     Register(registry, ++id, Primitive::Mha, mha::MhaCKFlashAttentionV2Forward{}.SolverDbId());
+    RegisterWithSolver(registry,
```
averinevg, line 710: Let's add one blank line here to separate from mha::MhaCKFlashAttentionV2Forward

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2023 Advanced Micro Devices, Inc.
```
averinevg:
```suggestion
 * Copyright (c) 2025 Advanced Micro Devices, Inc.
```

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,615 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (c) 2023 Advanced Micro Devices, Inc.
```
averinevg:
```suggestion
 * Copyright (c) 2025 Advanced Micro Devices, Inc.
```

test/gtest/group_conv.hpp
```diff
@@ -434,13 +434,15 @@ struct GroupConvTestFix
         {
             DispatchSolver<miopen::solver::conv::ConvHipImplicitGemmGroupFwdXdlops,
                            miopen::solver::conv::ConvHipImplicitGemmGroupBwdXdlops,
-                           miopen::solver::conv::ConvHipImplicitGemmGroupWrwXdlops>();
+                           miopen::solver::conv::ConvHipImplicitGemmGroupWrwXdlops,
+                           miopen::solver::conv::ConvHipImplicitGemmGroupWrwCKNCHWXdlops>();
```
averinevg: I think it is better to implement the test in `unit_conv_solver_*`. That would allow the use of the entire testing infrastructure that is already being used for other solvers.

src/include/miopen/conv/solvers.hpp
```diff
@@ -4903,6 +4903,178 @@ struct ConvHipImplicitGemmGroupWrwXdlops final
     bool CheckCKApplicability(const miopen::conv::ProblemDescription&) const;
 };
 
+struct PerformanceConfigHipImplicitGemmGroupWrwCKNCHWXdlops
+    : PerfConfigBaseCK<PerformanceConfigHipImplicitGemmGroupWrwCKNCHWXdlops>
```
averinevg, line 4899: Is it possible to have a common performance config or at least a base class or template for all these performance configs for CK? We need to somehow generalize this code. At the moment, we already have a bunch of these duplicates. It really seems like all of these are just copied and pasted.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -65,19 +65,40 @@ template <typename DataType>
...
+    ck::tensor_operation::element_wise::PassThrough,
+    ck::tensor_operation::element_wise::PassThrough,
+    ck::tensor_operation::element_wise::PassThrough>;
+template <typename DataType>
+using DeviceOpGWrwNCHWPtrs = ck::tensor_operation::device::instance::DeviceOperationInstanceFactory<
+    DeviceOpGWrwNCHW<DataType>>;
+
+// TODO: need to combine/template/refactor these CK parameters once finalizing
+// whether we keep the CK NHWGC + MIOpen transform solvers along with new CK NGCHW
+// slolvers
```
averinevg, line 86:
```suggestion
// solvers
```

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -1191,5 +1289,41 @@ MakeSolutionGroupConvImplicitGemmXdlops(const miopen::conv::ProblemDescription&
 #endif
 }
 
+template <typename InvokerFactoryMakerNHWC>
+ConvSolution
+MakeSolutionGroupConvImplicitGemmNCHWXdlops(const miopen::conv::ProblemDescription& problem,
+                                            InvokerFactoryMakerNHWC&& invoker_factory_maker_ndhwc)
```
averinevg, line 1296: Maybe `solution_maker_ndhwc`?
```suggestion
                                            InvokerFactoryMakerNHWC&& solution_maker_ndhwc)
```

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+    [[maybe_unused]] const ProblemDescription& problem,
+    [[maybe_unused]] const PerformanceConfigHipImplicitGemm3DGroupWrwCKNCHWXdlops& config) const
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    return MakeSolutionGroupConvImplicitGemmNCHWXdlops(problem, [&](auto data_type_val) {
+        using T = decltype(data_type_val);
+        switch(problem.GetAlphaBetaCase())
+        {
+        case BILINEAR:
+            return InitInvokerFactoryNHWC<DeviceOpGBwdWeightBilinearPtrsNGCDHW<T>,
```
averinevg, line 506: Why InitInvokerFactoryNHWC returns ConvSolution? Probably it should be renamed.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+#include <miopen/conv/solvers.hpp>
+#include <miopen/generic_search.hpp>
+#include <miopen/conv/wrw_invoke_params.hpp>
+#include <miopen/solver/problem_description_interpreter.hpp>
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+#include <miopen/solver/ck_utility_common.hpp>
+#endif
+#include <miopen/solver/implicitgemm_ck_util.hpp>
+#include <miopen/solver/implicitgemm_util.hpp>
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_3D_CONV_IMPLICIT_GEMM_HIP_WRW_XDLOPS)
```
averinevg: This env variable already exists

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,615 @@
...
+#include <miopen/conv/wrw_invoke_params.hpp>
+#include <miopen/solver/problem_description_interpreter.hpp>
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+#include <miopen/solver/ck_utility_common.hpp>
+#include <miopen/conv/heuristics/ai_heuristics.hpp>
+#endif
+#include <miopen/solver/implicitgemm_ck_util.hpp>
+#include <miopen/solver/implicitgemm_util.hpp>
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_GROUP_CONV_IMPLICIT_GEMM_HIP_WRW_XDLOPS)
+MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_GROUP_CONV_IMPLICIT_GEMM_HIP_WRW_XDLOPS_AI_HEUR)
```
averinevg: These env variables already exist

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,615 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <vector>
+#include <cstdint>
+
+#include <miopen/conv/solvers.hpp>
+#include <miopen/generic_search.hpp>
```
averinevg, line 31:
```suggestion
#include <miopen/conv/solvers.hpp>
#include <miopen/env.hpp>
#include <miopen/generic_search.hpp>
```

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,615 @@
...
+        return false;
+    if(!ck_utility::is_ck_whitelist(ctx.GetStream().GetDeviceName()))
+        return false;
+    switch(problem.GetInDataType())
+    {
+    case miopenHalf: return CheckCKApplicability<ck::half_t>(problem);
+    case miopenFloat: return CheckCKApplicability<float>(problem);
+    case miopenInt8: return CheckCKApplicability<int8_t>(problem);
+    case miopenBFloat16:
+        return StartsWith(ctx.GetStream().GetDeviceName(), "gfx94") &&
```
averinevg:
```suggestion
        return ctx.GetStream().GetDeviceName() == "gfx942" &&
```

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
averinevg, line 1: This file is almost an exact copy of [conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp](https://github.com/ROCm/MIOpen/blob/develop/src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_xdlops.cpp). Is it possible to merge these two files into one?

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ *******************************************************************************/
+
+#include <vector>
+#include <cstdint>
+
+#include <miopen/conv/solvers.hpp>
+#include <miopen/generic_search.hpp>
```
averinevg, line 31:
```suggestion
#include <miopen/conv/solvers.hpp>
#include <miopen/env.hpp>
#include <miopen/generic_search.hpp>
```

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+                    int split_k) const
+    {
+        return MakeArgPtr(conv_ptr, tensors.x, tensors.dw, tensors.dy, alpha, beta, split_k);
+    }
+
+    template <typename ConvPtr>
+    bool IsSupportedBy(const ConvPtr& conv_ptr) const
+    {
+        auto arg_ptr = MakeArgPtr(conv_ptr, nullptr, nullptr, nullptr, 1.0f, 0.0f, 1);
+        // Creat dummy workspace to pass the ck IsSupportedArgument check.
```
averinevg, line 225:
```suggestion
        // Create dummy workspace to pass the ck IsSupportedArgument check.
```

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+}
+
+bool ConvHipImplicitGemm3DGroupWrwCKNCHWXdlops::IsApplicable(
+    [[maybe_unused]] const ExecutionContext& ctx,
+    [[maybe_unused]] const ProblemDescription& problem) const
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    if(env::disabled(MIOPEN_DEBUG_3D_CONV_IMPLICIT_GEMM_HIP_WRW_XDLOPS))
+        return false;
+    if(env::enabled(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC))
```
averinevg, line 462:
```suggestion
    if(problem.GetConv().attribute.deterministic)
```

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+    if(problem.HasMixedDataTypes())
+        return false;
+    if(!problem.IsDirectionBackwardWrW())
+        return false;
+    if(!problem.Is3d())
+        return false;
+    if(!problem.IsLayoutDefault())
+        return false;
+    // needed because layout transpose kernel does not support non-packed tensors
+    if(problem.IsLayoutDefault() && problem.HasNonPackedTensors())
```
averinevg: IsLayoutDefault() is already checked

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+    if(!ck_utility::is_ck_whitelist(ctx.GetStream().GetDeviceName()))
+        return false;
+    switch(problem.GetInDataType())
+    {
+    case miopenHalf: return CheckCKApplicability<ck::half_t>(problem);
+    case miopenFloat: return CheckCKApplicability<float>(problem);
+    case miopenInt8: return CheckCKApplicability<int8_t>(problem);
+    case miopenBFloat16:
+        return StartsWith(ctx.GetStream().GetDeviceName(), "gfx94") &&
+               CheckCKApplicability<ck::bhalf_t>(problem);
```
averinevg: Should be applicable to gfx950 as well

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+        return false;
+    if(!ck_utility::is_ck_whitelist(ctx.GetStream().GetDeviceName()))
+        return false;
+    switch(problem.GetInDataType())
+    {
+    case miopenHalf: return CheckCKApplicability<ck::half_t>(problem);
+    case miopenFloat: return CheckCKApplicability<float>(problem);
+    case miopenInt8: return CheckCKApplicability<int8_t>(problem);
+    case miopenBFloat16:
+        return StartsWith(ctx.GetStream().GetDeviceName(), "gfx94") &&
```
averinevg:
```suggestion
        return ctx.GetStream().GetDeviceName() == "gfx942" &&
```

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
averinevg: Should be applicable to gfx950 as well

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
averinevg, line 1: This file is almost an exact copy of [conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp](https://github.com/ROCm/MIOpen/blob/develop/src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp). Is it possible to merge these two files into one?

averinevg (changes requested):

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
iq136boy: @averinevg I would prefer to keep both solvers at the moment. The reason is that we do not want to modify the old CK solvers which has already been tuned and adding heuristics already. If we change them now, then we need to redo all the tuning and heuristic too, which would take a long time and would miss the customer's deadline.  Also, we still not sure about performance about the NEW CK NCHW kernels VS the CK NHWC kernel + MIOpen transform. In the worst case, if the new kernels' overall performance is worse than the ones we have now, then we might need to consider just deprecate the new solvers and still use the current ones. If it is better, then we can consider code refactoring to reduce the duplicate code,  which will be addressed in a separate PR.

iq136boy:
@averinevg thanks taking time for the code review. I will address them and post the update.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -65,19 +65,40 @@ template <typename DataType>
...
+template <typename DataType>
+using DeviceOpGWrwNCHWPtrs = ck::tensor_operation::device::instance::DeviceOperationInstanceFactory<
+    DeviceOpGWrwNCHW<DataType>>;
+
+// TODO: need to combine/template/refactor these CK parameters once finalizing
+// whether we keep the CK NHWGC + MIOpen transform solvers along with new CK NGCHW
+// slolvers
+using InLayoutNDHWGC  = ck::tensor_layout::convolution::NDHWGC;
+using InLayoutNGCDHW  = ck::tensor_layout::convolution::NGCDHW;
+using WeiLayout       = ck::tensor_layout::convolution::GKZYXC;
```
BrianHarrisonAMD, line 89: Should we also update the weight layouts name similar to the others?
Now we have two weight layouts it seems.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -130,6 +151,64 @@ using DeviceOpGBwdWeightScalePtrs =
     ck::tensor_operation::device::instance::DeviceOperationInstanceFactory<
         DeviceOpGBwdWeightScale<DataType>>;
 
+template <typename DataType>
+using DeviceOpGBwdWeightDefaultNGCDHW =
+    ck::tensor_operation::device::DeviceGroupedConvBwdWeight<3,
+                                                             InLayoutNGCDHW,
+                                                             WeiLayout,
```
BrianHarrisonAMD, line 158: Does the weights layout not also change?
I was expecting this one to be GKCZYX (same for other NCHW solvers as well).

src/include/miopen/solver/implicitgemm_ck_util.hpp
BrianHarrisonAMD, line 89: Need to split these once we get an update.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+        data_type       = ProblemInterpreter::GetOutputDataType(problem);
+        alpha_beta_case = ProblemInterpreter::GetAlphaBetaCase(problem);
+
+        in_lengths  = {G, N, C, Di, Hi, Wi};
+        out_lengths = {G, N, K, Do, Ho, Wo};
+        wei_lengths = {G, K, C, Z, Y, X};
+
+        in_strides  = {Di * Hi * Wi * C, Di * Hi * Wi * G * C, Di * Hi * Wi, Hi * Wi, Wi, 1};
+        out_strides = {Do * Ho * Wo * K, Do * Ho * Wo * G * K, Do * Ho * Wo, Ho * Wo, Wo, 1};
+        wei_strides = {K * Z * Y * X * C, Z * Y * X * C, 1, Y * X * C, X * C, C};
```
BrianHarrisonAMD, line 80: The weight lengths & strides will also need to be change once we get update from CK for layout.

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,532 @@
...
+    return GenericSearch(*this, ctx, problem, invoke_ctx);
+}
+
+bool ConvHipImplicitGemm3DGroupWrwCKNCHWXdlops::IsApplicable(
+    [[maybe_unused]] const ExecutionContext& ctx,
+    [[maybe_unused]] const ProblemDescription& problem) const
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    if(env::disabled(MIOPEN_DEBUG_3D_CONV_IMPLICIT_GEMM_HIP_WRW_XDLOPS))
+        return false;
```
BrianHarrisonAMD: Are we planning on re-using this, or separating between solvers?

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
BrianHarrisonAMD: Is this still needed? 

We used to have it since we didn't have a good transpose for non-packed, but does CK support non-packed NCHW?

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_nchw_xdlops.cpp
BrianHarrisonAMD: Not sure if we need this.
It was added to the other solver originally due to a WA, but we should verify if it's still needed for these new solvers.
I believe the old ones were actually work for @cderb when he tried them in a recent update.

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,615 @@
...
+        X               = ProblemInterpreter::GetFilterWidthX(problem);
+        data_type       = ProblemInterpreter::GetOutputDataType(problem);
+        alpha_beta_case = ProblemInterpreter::GetAlphaBetaCase(problem);
+        input           = {G, N, C, Hi, Wi};
+        output          = {G, N, K, Ho, Wo};
+        weight          = {G, K, C, Y, X};
+
+        in_strides  = {Hi * Wi * C, Hi * Wi * G * C, Hi * Wi, Wi, 1};
+        out_strides = {Ho * Wo * K, Ho * Wo * G * K, Ho * Wo, Wo, 1};
+        wei_strides = {K * Y * X * C, Y * X * C, 1, X * C, C};
```
BrianHarrisonAMD: Just a reminder to check weight & weight stride when layout changes.

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,615 @@
...
+        if(valid_kernels[i].find(type) != std::string::npos)
+        {
+            heuristic_indexes.push_back(i);
+            heuristic_kernels[i] = GetKernelAsTokens(valid_kernels[i]);
+        }
+    }
+}
+
+bool PerformanceConfigHipImplicitGemmGroupWrwCKNCHWXdlops::ModelApplyToken(
+    int idx, std::string value, const std::string& arch, const ProblemDescription& problem)
```
BrianHarrisonAMD: Probably fine to leave this for now, but I imagine we will need new heuristics for this new solver.

BrianHarrisonAMD: Or maybe we can leverage the same one as before since the performance params should be close.

Thoughts @cderb ?

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,615 @@
...
+}
+
+bool ConvHipImplicitGemmGroupWrwCKNCHWXdlops::IsApplicable(
+    [[maybe_unused]] const ExecutionContext& ctx,
+    [[maybe_unused]] const ProblemDescription& problem) const
+{
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
+    if(env::disabled(MIOPEN_DEBUG_GROUP_CONV_IMPLICIT_GEMM_HIP_WRW_XDLOPS))
+        return false;
+    if(env::enabled(MIOPEN_DEBUG_CONVOLUTION_DETERMINISTIC))
```
BrianHarrisonAMD: Same comments as 3D.
Env shared, and this should check the conv flag for deterministic instead of the env.

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
```diff
@@ -0,0 +1,615 @@
...
+        return false;
+    if(!problem.IsDirectionBackwardWrW())
+        return false;
+    if(!problem.Is2d())
+        return false;
+    if(!problem.IsLayoutDefault())
+        return false;
+    // needed because layout transpose kernel does not support non-packed tensors
+    if(problem.IsLayoutDefault() && problem.HasNonPackedTensors())
+        return false;
```
BrianHarrisonAMD: This also true for CK transpose?

(can remove the layout check either way).

src/solver/conv/conv_hip_implicit_gemm_grouped_wrw_nchw_xdlops.cpp
BrianHarrisonAMD: Same as 3D, I think we should see if this is needed.

BrianHarrisonAMD:
I think we need to double check the verification is working properly for these tests.
It looks like the CK args would treat the MIOpen weights as a different layout than they are, and I am surprised that doesn't cause issues.

src/include/miopen/solver/implicitgemm_ck_util.hpp
iq136boy: Yes. The weight tensor layout need to NCHW too.

iq136boy:
close and will reopen if needed.

PR #3552, opened 2025-02-26
[tests] Unit tests for ConvMPBidirectWinograd

averinevg:
This PR adds unit tests for `ConvMPBidirectWinograd`

Additional changes:
- WORKAROUND_ISSUE_1146 has been removed for this solver.
- Added `ConvMPBidirectWinograd<2-3>` to the list of Winograd solvers.
- Enable  `ConvMPBidirectWinograd` only for gfx900, gfx906 and gfx908 (so that it doesn't enabled for all subsequent gfx9).

PR #3626, opened 2025-03-14
Fix winograd tolerance for MI100

BrianHarrisonAMD:
Fixing a testing break due to winograd having very tight tolerance ranges after getting added in #3552.

### DIFF
--- a/test/gtest/unit_conv_solver.cpp
+++ b/test/gtest/unit_conv_solver.cpp
@@ -312,10 +312,6 @@ double GetThreshold(miopenConvAlgorithm_t algo, miopen::conv::Direction directio
         {
             tolerance *= 3.0;
         }
-        else if(algo == miopenConvolutionAlgoWinograd)
-        {
-            tolerance *= 10.0;
-        }
     }
 
     double threshold = std::numeric_limits<T>::epsilon() * tolerance;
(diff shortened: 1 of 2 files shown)


iq136boy (approved):

test/gtest/unit_conv_solver.cpp
```diff
@@ -312,6 +312,10 @@ double GetThreshold(miopenConvAlgorithm_t algo, miopen::conv::Direction directio
         {
             tolerance *= 3.0;
         }
+        else if(algo == miopenConvolutionAlgoWinograd)
+        {
```
adickin-amd: does this apply to all GPU arch's or just the m100 series?

adickin-amd (approved):

test/gtest/unit_conv_solver.cpp
```diff
@@ -312,6 +312,10 @@ double GetThreshold(miopenConvAlgorithm_t algo, miopen::conv::Direction directio
...
+            tolerance *= 10.0;
```
randyspauldingamd: I don't think the "just get the test to pass" approach is valid--it would be better to disable the tests. This has been discussed at least twice in the past, but never gotten the proper attention. Let's start working towards a fix.

randyspauldingamd (changes requested):
Let's discuss a proper solution.

test/gtest/unit_conv_solver.cpp
BrianHarrisonAMD: All

shurale-nkn (commented):
I'm ok with test disable. LGTM in advance for this PR, I’m off for the weekend

shurale-nkn (approved):
LGTM in advance

randyspauldingamd (approved):

PR #3635, opened 2025-03-17
Fixed miopenSetProblemTensorDescriptor to produce a correct error instead of silently failing

DrizztDoUrden:
When miopenSetProblemTensorDescriptor is invoked with miopenTensorArgumentId_t values that are already present in the problem it now produces an error code instead of silently ignoring passed values.


BrianHarrisonAMD (approved):

shurale-nkn (approved):
LGTM

PR #3646, opened 2025-03-21
BatchNorm tune API

bghimireamd:
The API version of MIOPEN_FIND_ENFORCE. But for now we only support 
1) None
2) DbUpdate
3) Search
4) SearchDbUpdate
5) DbClean

This PR introduces **TuningPolicy (this is the new thing)**

This allows us to Tune from API. 

`miopenSetTuningPolicy(handle, miopenTuningPolicySearch); // set tuning`
`miopenBatchnorm...(...);`
`miopenSetTuningPolicy(handle, miopenTuningPolicyNone); // unset tuning`

This effect is not limited to bnorm.

Story for smoke test of this feature has been created and assigned. The PR currently only has unit test.


src/include/miopen/handle.hpp
```diff
@@ -84,6 +84,8 @@ using hipblasLt_handle_ptr = MIOPEN_MANAGE_PTR(hipblasLtHandle_t, hipblasLtDestr
 
 struct MIOPEN_EXPORT Handle : miopenHandle
 {
+    miopenTuningPolicy_t tuning_policy;
```
JonathanLichtnerAMD: Should this be private since you have getters and setters for this?

include/miopen/miopen.h
```diff
@@ -8226,6 +8226,39 @@ MIOPEN_EXPORT miopenStatus_t miopenMultiMarginLossForward(miopenHandle_t handle,
...
+/*! @ingroup handle
+ * @enum miopenTuningPolicy_t
+ * Tuning policy for MIOpen Find-related calls.
+ */
+typedef enum
+{
+    miopenTuningPolicyNone     = 1, /* do not enforce anything */
+    miopenTuningPolicyDbUpdate = 2, /* do not load existing entry, do not tune  */
+    miopenTuningPolicySearch   = 3, /* do not load existing entry, tune */
+    miopenTuningPolicyDbClean  = 5, /* remove existing entry, do not tune */
```
JonathanLichtnerAMD: I know you mentioned skipping SEARCH_DB_UPDATE, which corresponds to a 4, and was basically equivalent to miopenTuningPolicySearch.  But doesn't this new api also apply to convolutions, so wouldn't we want to support SEARCH_DB_UPDATE in this api?

src/api/tuning.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <miopen/miopen.h>
+
+#include <miopen/errors.hpp>
+#include <miopen/handle.hpp>
+
+miopenStatus_t miopenSetTuningPolicy(miopenHandle_t handle, miopenTuningPolicy_t newValue)
+{
+    return miopen::try_([&] {
+        if(newValue < miopenTuningPolicyNone || newValue > miopenTuningPolicyDbClean)
```
JonathanLichtnerAMD, line 35: Since this is an enum, the user could actually enter a 4 here.  I assume this would just default to SEARCH_DB_UPDATE behaviour, but I wanted to double check...

JonathanLichtnerAMD (dismissed):

include/miopen/miopen.h
bghimireamd: I made sure tuning policy 4 is invalid in setter

BrianHarrisonAMD: This new API would work with convs right?
It's looks like a handle level tuning policy that will be used if the find loop doesn't have find options set on it. 
I think it's fine to have this as an option, but we probably also want an option that works for BN API calls specifically.

The way convs currently work, you can set the tuning options per API call, and that makes it so you don't have to unset the tuning policy after you are done. This change would require a pattern that is like:

miopenSetTuningPolicy(option)
// some API calls 
miopenSetTuningPolicy(none)

Which could be error prone, and have side-effects that are hard for people to notice.

BrianHarrisonAMD:
There is the new tuning policy here as well:
https://github.com/ROCm/MIOpen/pull/3604

Probably need to coordinate these changes with that as well.

include/miopen/miopen.h
cderb: This is inaccurate if miopenTuningPolicy_t is to mirror MIOPEN_FIND_ENFORCE behavior
DB_UPDATE - load existing entry, do not tune
SEARCH - load existing entry, tune if no entry is found
SEARCH_DB_UPDATE - do not load existing entry, tune

src/include/miopen/find_solution.hpp
```diff
@@ -63,8 +63,15 @@ auto FindSolutionImpl(rank<1>,
...
-    const FindEnforce enforce =
-        options && options->find_enforce ? *options->find_enforce : FindEnforce{};
+    const auto enforce = [&]() {
+        if(options && options->find_enforce)
+            return *options->find_enforce;
+        const auto& handle = context.GetStream();
+        if(handle.GetTuningPolicy() != miopenTuningPolicyNone)
+            return FindEnforce{static_cast<FindEnforceAction>(handle.GetTuningPolicy())};
+        return FindEnforce{};
+    }();
```
cderb: Looks like MIOPEN_FIND_ENFORCE is higher priority than TuningPolicy.

test/gtest/smoke_tuning_policy.cpp
```diff
@@ -0,0 +1,44 @@
...
+    EXPECT_EQ(miopenGetTuningPolicy(&handle, &test_tuning_policy), miopenStatusSuccess);
+    EXPECT_EQ(test_tuning_policy, miopenTuningPolicy_t::miopenTuningPolicyDbUpdate);
+
+    // test by casting
+    EXPECT_EQ(miopenSetTuningPolicy(&handle, static_cast<miopenTuningPolicy_t>(2)),
+              miopenStatusSuccess);
+    EXPECT_EQ(miopenGetTuningPolicy(&handle, &test_tuning_policy), miopenStatusSuccess);
+    EXPECT_EQ(test_tuning_policy, miopenTuningPolicy_t::miopenTuningPolicyDbUpdate);
+
+    miopenStatus_t status = miopenSetTuningPolicy(&handle, static_cast<miopenTuningPolicy_t>(4));
```
randyspauldingamd: Use EXPECT_NO_THROW here since that's a requirement:
```suggestion
    miopenStatus_t status;
    EXPECT_NO_THROW(status = miopenSetTuningPolicy(&handle, static_cast<miopenTuningPolicy_t>(4)););
```

src/api/tuning.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <miopen/errors.hpp>
+#include <miopen/handle.hpp>
+
+miopenStatus_t miopenSetTuningPolicy(miopenHandle_t handle, miopenTuningPolicy_t newValue)
+{
+    return miopen::try_([&] {
+        if(newValue < miopenTuningPolicyNone || newValue > miopenTuningPolicyDbClean ||
+           newValue == 4)
+            MIOPEN_THROW(miopenStatusBadParm,
+                         "miopenSetTuningPolicy called with invalid value of newValue");
```
randyspauldingamd:
```suggestion
            MIOPEN_THROW(miopenStatusBadParm,
                         "miopenSetTuningPolicy called with invalid value of " + std::to_string(newValue));
```

src/find_controls.cpp
```diff
@@ -55,6 +55,14 @@ MIOPEN_EXPORT bool FindEnforceDisable = false;
 
 } // namespace debug
 
+static_assert(FindEnforceAction::None == static_cast<FindEnforceAction>(miopenTuningPolicyNone));
+static_assert(FindEnforceAction::DbUpdate ==
+              static_cast<FindEnforceAction>(miopenTuningPolicyDbUpdate));
+static_assert(FindEnforceAction::Search ==
+              static_cast<FindEnforceAction>(miopenTuningPolicySearch));
+static_assert(FindEnforceAction::DbClean ==
+              static_cast<FindEnforceAction>(miopenTuningPolicyDbClean));
```
randyspauldingamd: As Chris mentioned, these are the checks which have inconsistent verbiage in miopen.h.

randyspauldingamd (changes requested):

src/api/tuning.cpp
bghimireamd: I modified the struct. Now SEARCH_DB_UPDATE case is also handled.

include/miopen/miopen.h
bghimireamd: I modified the struct to mirror MIOPEN_FIND_ENFORCE.

bghimireamd:
> There is the new tuning policy here as well: #3604
> 
> Probably need to coordinate these changes with that as well.

Added comment https://github.com/ROCm/MIOpen/blob/2253ba57bc4c91c8f75686375eabe98f86ff8a44/include/miopen/miopen.h#L8232 saying we have limited support.

src/include/miopen/find_solution.hpp
bghimireamd: TuningPolicy has higher priority

test/gtest/smoke_tuning_policy.cpp
bghimireamd: test no longer valid

src/api/tuning.cpp
bghimireamd: good catch!

src/find_controls.cpp
bghimireamd: updated the struct all together

src/api/tuning.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include <miopen/miopen.h>
+
+#include <miopen/errors.hpp>
+#include <miopen/handle.hpp>
+
+miopenStatus_t miopenSetTuningPolicy(miopenHandle_t handle, miopenTuningPolicy_t newValue)
+{
+    return miopen::try_([&] {
+        if(newValue < miopenTuningPolicyNone || newValue > miopenTuningPolicyDbClean)
```
reidkwja, line 35: Extending on @JonathanLichtnerAMD comment, I think the current logic assumes that the values are consecutive - in which this check might not cover it under the circumstance of new tunning policy additions. It might be better to utilize `switch(newValue)` with explicit cases.

src/api/tuning.cpp
```diff
@@ -0,0 +1,54 @@
...
+            MIOPEN_THROW(miopenStatusBadParm,
+                         "miopenSetTuningPolicy called with invalid value of " +
+                             std::to_string(newValue));
+
+        auto& handle_deref = miopen::deref(handle);
```
reidkwja, line 40: It might be a better approach to explicitly validate `handle != nullptr` at the top. This way the null handle isn't passed into lower levels - from an abstraction standpoint.

src/api/tuning.cpp
bghimireamd: not sure switch statement helps here since its just a setter

src/api/tuning.cpp
bghimireamd: `deref` automatically checks for nullptr

BrianHarrisonAMD (approved):

src/include/miopen/handle.hpp
```diff
@@ -180,10 +180,6 @@ struct MIOPEN_EXPORT Handle : miopenHandle
     std::string GetDeviceName() const;
     virtual const TargetProperties& GetTargetProperties() const;
 
-private:
-    std::string GetDeviceNameImpl() const;
-
-public:
```
BrianHarrisonAMD, line 186: Not a huge deal, but why move this out of private?

test/gtest/smoke_tuning_policy.cpp
```diff
@@ -0,0 +1,40 @@
...
+              miopenStatusSuccess);
+    EXPECT_EQ(miopenGetTuningPolicy(&handle, &test_tuning_policy), miopenStatusSuccess);
+    EXPECT_EQ(test_tuning_policy, miopenTuningPolicy_t::miopenTuningPolicySearchDbUpdate);
+
+    // reset back to none
+    EXPECT_EQ(miopenSetTuningPolicy(&handle, miopenTuningPolicy_t::miopenTuningPolicyNone),
+              miopenStatusSuccess);
+    EXPECT_EQ(miopenGetTuningPolicy(&handle, &test_tuning_policy), miopenStatusSuccess);
+    EXPECT_EQ(test_tuning_policy, miopenTuningPolicy_t::miopenTuningPolicyNone);
+}
```
BrianHarrisonAMD: We should also have tests for invalid values I think (negative, and bigger than max etc).

include/miopen/miopen.h
```diff
@@ -8404,6 +8404,49 @@ MIOPEN_EXPORT miopenStatus_t miopenMultiMarginLossForward(miopenHandle_t handle,
...
+ * Note: TuningPolicy has higher priority over MIOPEN_FIND_ENFORCE.
+ */
+typedef enum
+{
+    miopenTuningPolicyNone     = 1, /* do not enforce anything */
+    miopenTuningPolicyDbUpdate = 2, /* tune and update the db  */
+    miopenTuningPolicySearch =
+        3, /* search db first, if record not found tune but do not update the db*/
+    miopenTuningPolicySearchDbUpdate = 4, /* combination of Search and DbUpdate */
+    miopenTuningPolicyDbClean        = 5, /* remove existing entry, do not tune */
```
cderb: advertised behavior:
https://github.com/ROCm/MIOpen/blob/7e387ac49ab79d8e78aed3a54ec2362745b1183c/docs/conceptual/perfdb.rst?plain=1#L71-L87
implemented behavior:
3 will use db/model results if available and tune + update the user db when results are missing.
4 will ignore all results and tune + update the user db
2 claims to update the db, but in implementation does nothing
  - MIOPEN_FIND_ENFORCE values are referenced only in find_solution.hpp in FindSolutionImpl. A combination of IsSearch + IsDbUpdate will cancel the db load, but only IsSearch will trigger tuning.
https://github.com/ROCm/MIOpen/blob/7e387ac49ab79d8e78aed3a54ec2362745b1183c/src/include/miopen/find_solution.hpp#L81-L85
Perhaps this is a separate issue, but I would like to avoid having 2 expectations of how these policies behave.

src/include/miopen/handle.hpp
bghimireamd: moved it back to private

include/miopen/miopen.h
bghimireamd: I made 2 update db to skip perdb load and do the tuning

src/include/miopen/find_controls.hpp
```diff
@@ -85,8 +85,9 @@ class MIOPEN_INTERNALS_EXPORT FindEnforce
     template <class Context>
     bool IsSearch(const Context& context) const
     {
-        return IsEnabled(context) &&
-               (action == FindEnforceAction::Search || action == FindEnforceAction::SearchDbUpdate);
+        return IsEnabled(context) && (action == FindEnforceAction::Search ||
+                                      action == FindEnforceAction::SearchDbUpdate ||
+                                      action == FindEnforceAction::DbUpdate);
```
bghimireamd, line 90: I added FindEnforceAction::DbUpdate here in Search too since[ this condition](https://github.com/ROCm/MIOpen/blob/develop/src/include/miopen/find_solution.hpp#L81) skips perf db loading.

BrianHarrisonAMD (approved):
Looks good to me, but I think there has been some other changes, and this will have to be updated to include the trust_verify changes from @cderb.

src/include/miopen/find_controls.hpp
cderb: The perf db load skip condition flags on both IsSearch and IsDbUpdate. So there is no need to ambiguate these terms here.

src/include/miopen/find_solution.hpp
cderb: My mistake, I misread.

amd-hsivasun:

PR #3653, opened 2025-03-26
use scope class for env setting

bghimireamd:
* Create Scope env class
* remove static initialization of action in FindEnforce

### DIFF
--- a/src/env_debug.cpp
+++ b/src/env_debug.cpp
@@ -110,6 +110,9 @@ private:
 
         std::optional<std::string> Get() const override
         {
+            if(!var)
+                return std::nullopt;
+
             const auto value = miopen::env::value(var);
 
             if constexpr(is_type_bool<value_type>)
--- a/src/env_debug.cpp
+++ b/src/env_debug.cpp
@@ -110,8 +110,6 @@ private:
 
         std::optional<std::string> Get() const override
         {
-            if(!var)
-                return std::nullopt;
             const auto value = miopen::env::value(var);
 
             if constexpr(is_type_bool<value_type>)
--- a/test/gtest/gtest_common.hpp
+++ b/test/gtest/gtest_common.hpp
@@ -40,18 +40,18 @@
 #include "../lib_env_var.hpp"
 
 template <typename T>
-class EnvScopedSetter
+class ScopedEnvironment
 {
 public:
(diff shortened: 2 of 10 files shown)


src/env_debug.cpp
```diff
@@ -110,8 +110,6 @@ struct LibEnvVar
 
         std::optional<std::string> Get() const override
         {
-            if(!var)
-                return std::nullopt;
```
BrianHarrisonAMD: Do we not need this guard?

BrianHarrisonAMD: I guess it might do this anyways?

test/gtest/gtest_common.hpp
```diff
@@ -39,6 +39,34 @@
...
+
+    ~EnvScopedSetter() { lib_env::update(env_name, prev_val); }
+
+private:
+    lib_env::LibEnvVar env_name;
+    T prev_val;
+
+    void SetValue(T value)
+    {
+        std::cout << env_name.name << std::endl;
```
BrianHarrisonAMD: logging instead of cout I think.

test/gtest/gtest_common.hpp
```diff
@@ -39,6 +39,34 @@
...
+    EnvScopedSetter(const EnvScopedSetter&) = delete;
+    EnvScopedSetter(EnvScopedSetter&&)      = delete;
+    EnvScopedSetter& operator=(const EnvScopedSetter&) = delete;
+    EnvScopedSetter& operator=(EnvScopedSetter&&) = delete;
+
+    ~EnvScopedSetter() { lib_env::update(env_name, prev_val); }
+
+private:
+    lib_env::LibEnvVar env_name;
+    T prev_val;
```
BrianHarrisonAMD, line 68: I think this should be std::optional<T>.
The ~ should check if it's set, and if it's not, then it should clear instead of updating.
Probably should have the changed flag as well, to skip updating it if it's not needed.

BrianHarrisonAMD (approved):
LGTM assuming CI passes.

JonathanLichtnerAMD (approved):

randyspauldingamd:
@bghimireamd has there been any testing of this with the single gtest binary yet?

bghimireamd:
> @bghimireamd has there been any testing of this with the single gtest binary yet?

Just saw this comment. The test did passed for Batch norm.

PR #3654, opened 2025-03-26
[GTests] Fix for SWDEV-522871

averinevg:
This PR partially fixes  SWDEV-522871 (workaround)


BrianHarrisonAMD (approved):

BrianHarrisonAMD:
Seems reasonable for now.

How did we want to handle this eventually?
I think we either remove these, or we have to make the tests only run on the expected platforms I guess.

JonathanLichtnerAMD (approved):

PR #3659, opened 2025-03-28
Skip some tests if MIOpen is built without CK to avoid failing on gfx90a, gfx908 and gfx94x (CK-compatible devices)

Vsevolod1983:
Tests ConvHipImplicitGemmFwdXdlops and ConvHipImplicitGemmFwdXdlops will fail on particular devices (CK-compatible devices) if MIOpen is built without CK.
Let's skip these tests to avoid failing.

### DIFF
--- a/test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
+++ b/test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdXdlops.cpp
@@ -73,8 +73,13 @@ auto GetConvFullTestCases(miopenDataType_t datatype)
 
 auto GetTestParams(miopenDataType_t datatype)
 {
+// If CK is not installed these tests would fail
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
     Gpu supportedDevices = Gpu::gfx908 | Gpu::gfx90A | Gpu::gfx94X;
-    auto params          = miopen::unit_tests::UnitTestConvSolverParams(supportedDevices);
+#else
+    Gpu supportedDevices = Gpu::None;
+#endif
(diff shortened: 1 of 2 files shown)


scerzh:
Should the workaround `WORKAROUND_SWDEV_522871` (from #3654) be removed now?

averinevg:
> Should the workaround `WORKAROUND_SWDEV_522871` (from #3654) be removed now?

No, this solves a different problem

averinevg (commented):
The problem in not caused by the CK being not installed and is not related to running on a specific gpu. The issue is that when the library is built with `-DMIOPEN_USE_COMPOSABLEKERNEL=Off`, the CK is excluded at the compiling stage. As a result, all solvers that use CK become inapplicable, which leads to test failures. The code is correct, but all comments, the title and the description of this PR should be changed.

scerzh:
`IsApplicable()` of a solver should return `false` and the test should check if it is that way

averinevg:
> `IsApplicable()` of a solver should return `false` and the test should check if it is that way

~~Then such a test would be useless. A variety of reasons can cause this function to return false.~~

scerzh:
> Then such a test would be useless. A variety of reasons can cause this function to return false.

Not completely useless, it will detect the case when `MIOPEN_USE_COMPOSABLEKERNEL` is `false` and `IsApplicable()` returns `true`, won't it?

averinevg:
> > Then such a test would be useless. A variety of reasons can cause this function to return false.
> 
> Not completely useless, it will detect the case when `MIOPEN_USE_COMPOSABLEKERNEL` is `false` and `IsApplicable()` returns `true`, won't it?

I got this now. That would be useful. But I think this is out of the scope of this PR.

averinevg (approved):
LGTM

scerzh (approved):
LGTM

BrianHarrisonAMD (approved):

PR #3660, opened 2025-03-28
[BN] BN driver signed input initialization 

bghimireamd:
* made the input tensor of Batch Norm signed


adickin-amd (approved):
Ive tested with fp32, fp16, and bfp16 and the verification now passes

BrianHarrisonAMD (approved):
LGTM

PR #3662, opened 2025-03-31
Vasilii made test_execute_primitive resilient to MIOPEN_FIND_ENFORCE

randyspauldingamd:
Addressing a failure in https://ontrack-internal.amd.com/browse/SWDEV-524093

Fix for CPU_ExecutePrimitive_NONE* tests were sensitive to MIOPEN_FIND_ENFORCE being set in the environment. This was only affecting CQE tests.


BrianHarrisonAMD:
Is this not already fixed by @bghimireamd's change here?

https://github.com/ROCm/MIOpen/pull/3653

If we still have spots setting the FIND_ENFORCE environment during testing without properly clearing it, then we need to find and fix that spot I think. Otherwise, we could hit huge increases to testing runtime, and volatility in the single gtest binary.

randyspauldingamd:
> Is this not already fixed by @bghimireamd's change here?
> 
> #3653
> 
> If we still have spots setting the FIND_ENFORCE environment during testing without properly clearing it, then we need to find and fix that spot I think. Otherwise, we could hit huge increases to testing runtime, and volatility in the single gtest binary.

We're not sure. I had in my notes that I repro'd the failure last week, but I couldn't repro it today. I'm going to do some testing with and without the single binary this evening to learn more, and I've asked Bibek what has been done.

In any case, this is a good fail-safe. Increasing test robustness is one of my primary concerns.

randyspauldingamd:
> Is this not already fixed by @bghimireamd's change here?
> 
> #3653
> 
> If we still have spots setting the FIND_ENFORCE environment during testing without properly clearing it, then we need to find and fix that spot I think. Otherwise, we could hit huge increases to testing runtime, and volatility in the single gtest binary.

I've verified this crash does still exist with PR 3653. Continuing to investigate whether it might occur in CQE.

randyspauldingamd:
Still affecting QA.
Will merge this and create followup tickets to investigate the cause.

randyspauldingamd:
~Promoted a test-local object to use for this and changed name to better reflect the usage.~
Added a ticket for further investigation.

randyspauldingamd:
The ExecutePrimitive test now follows black-box better without requiring modifications to the test itself.

src/include/miopen/env_debug.hpp
```diff
@@ -46,6 +46,29 @@ MIOPEN_INTERNALS_EXPORT void UpdateEnvVariable(std::string_view name, std::strin
 // MT-Unsafe
 MIOPEN_INTERNALS_EXPORT void ClearEnvVariable(std::string_view name);
 
+class EnvVarSanitizer
```
bghimireamd: similar class is here https://github.com/ROCm/MIOpen/blob/develop/test/gtest/gtest_common.hpp#L43. Can we reuse that class?

bghimireamd (dismissed):

src/include/miopen/env_debug.hpp
randyspauldingamd: Sure, I'll add functionality to clear the var to it

test/gtest/gtest_common.hpp
```diff
@@ -65,22 +75,39 @@ class ScopedEnvironment
...
-    void SetValue(T value)
+    bool ClearValue()
     {
-        const auto val = miopen::debug::env::GetEnvVariable(env_name.name);
-        if(val)
+        prev_env = miopen::debug::env::GetEnvVariable(env_name.name);
+
+        if(prev_env)
+        {
+            lib_env::clear(env_name);
```
reidkwja, line 88: It might be a good idea to provide logging/exception if clearing the variable fails.

reidkwja (approved):

test/gtest/gtest_common.hpp
randyspauldingamd: I checked under the hood and I'd say the existing mechanism is sufficient. It already has strict enforcement on valid env names:

https://github.com/ROCm/MIOpen/blob/20ddfa7d44f3ecbaea95515f676e18f61015e416/src/env_debug.cpp#L207

Glad you mentioned that--I hadn't dived into this code before and it seems the env lists may have gotten out of sync.

PR #3666, opened 2025-04-01
Add out-of-bounds memory checking for GPUMem buffers and sub-buffers

JonathanLichtnerAMD:
This PR adds two different detection mechanisms for detecting out-of-bounds GPU buffer
memory accesses:
1. The first adds a new argument to MIOPenDriver, --gpubuffer_check <value>, which can be used to make the GPUMem class allocate GPU buffers that are aligned to 2 MiB alignment boundaries.  Specifying 1 for this option causes the client memory to be allocated at the start of the buffer (to detect OOBs before the buffer's start) and specifying 2 for this option causes the client memory to be aligned at the end of the buffer (to detect OOBs after the buffer's end).  Note that this is currently enabled for convolutions and batchnorm, although it is simple to add it for other problems.  (We might also want to consider turning this checking on by default since this only affects MIOpenDriver usage.)
2. The second method detects OOBs memory accesses for sub-buffers.  This can be enabled via the MIOPEN_DEBUG_CHECK_SUB_BUFFER_OOB_MEMORY_ACCESS environment variable, and uses a technique similar to that used for the GPUMem class (but adapted for sub-buffers).


driver/driver.hpp
```diff
@@ -131,8 +140,33 @@ struct GPUMem
     void* GetMem() { return buf; }
     size_t GetSize() { return sz * data_sz; }
 
+    size_t GetTotalSize(size_t userSize)
+    {
+        constexpr size_t memoryPageSize = 2 * 1024 * 1024;
+
+        auto roundUp = [&](size_t bytes) {
+            return ((bytes + memoryPageSize) / memoryPageSize) * memoryPageSize;
```
cderb: This will count an extra page when bytes is exactly the page size.

cderb (approved):
LGTM

driver/driver.hpp
```diff
@@ -131,8 +140,33 @@ struct GPUMem
     void* GetMem() { return buf; }
     size_t GetSize() { return sz * data_sz; }
 
+    size_t GetTotalSize(size_t userSize)
+    {
+        constexpr size_t memoryPageSize = 2 * 1024 * 1024;
```
averinevg: Why was this value chosen?
I think it is better to change the variable name to avoid confusion.

driver/driver.hpp
```diff
@@ -452,6 +501,36 @@ inline void Driver::InitDataType()
...
+    case 0: return GPUMem::Check::None;
+    case 1: return GPUMem::Check::Front;
+    case 2: return GPUMem::Check::Back;
+    default:
+        std::cerr << "Error: Invalid option " << check
+                  << " used with --gpubuffer_check.  Should be 0 (none), 1 (front), or 2 (back).";
+        exit(EXIT_FAILURE);
+    }
+}
+
```
averinevg: This code should probably be placed in `driver.cpp`

src/hip/handlehip.cpp
```diff
@@ -795,16 +796,123 @@ std::ostream& Handle::Print(std::ostream& os) const
...
+
+    auto roundUp = [&](size_t bytes) {
+        return ((bytes + memoryPageSize) / memoryPageSize) * memoryPageSize;
+    };
+
+    const auto totalSize = roundUp(size);
+    void* mem            = nullptr;
+    auto status          = hipMalloc(&mem, totalSize);
+    if(status != hipSuccess)
+        MIOPEN_THROW_HIP_STATUS(status, "[MIOpenDriver] hipMalloc failed " + std::to_string(size));
```
averinevg: `MIOpenDriver` is misleading here

src/hip/handlehip.cpp
```diff
@@ -65,6 +65,7 @@
 #define WORKAROUND_FAULTY_HIPMEMGETINFO_VEGA_NAVI2X (HIP_PACKAGE_VERSION_FLAT >= 5007000000ULL)
 
 MIOPEN_DECLARE_ENV_VAR_UINT64(MIOPEN_DEVICE_CU)
+MIOPEN_DECLARE_ENV_VAR_UINT64(MIOPEN_DEBUG_CHECK_SUB_BUFFERS)
```
averinevg: From the name it is not entirely clear what check is being performed. It would be good to clarify this it the name and also add it to the documentation. Additionally it should be noted that no actual check is being performed.

driver/driver.hpp
averinevg: When `bytes` is close to `memoryPageSize`, all of this becomes meaningless.

src/hip/handlehip.cpp
JonathanLichtnerAMD: Yes, the check is implicit.  How about MIOPEN_DEBUG_CHECK_SUB_BUFFER_OOB_MEMORY_ACCESS?  It's a bit long but it's more descriptive....

driver/driver.hpp
BrianHarrisonAMD: Originally in my example it allowed some extra room to help figure out how far it was going past the end, but in this version I don't think we need that. 

`((bytes + memoryPageSize - 1) / memoryPageSize) * memoryPageSize;`

could be used to avoid an extra page.

driver/driver.hpp
BrianHarrisonAMD: Originally in my example it allowed some extra room to help figure out how far it was going past the end, but in this version I don't think we need that. 

`((bytes + memoryPageSize - 1) / memoryPageSize) * memoryPageSize;`

could be used to avoid an extra page.

BrianHarrisonAMD: Yea if it's right on the page size, then it's basically a no-op. 
This just gives us a way to verify memory access issues for kernels that can't use ASAN.

driver/driver.hpp
```diff
@@ -131,8 +140,33 @@ struct GPUMem
...
+    size_t GetTotalSize(size_t userSize)
+    {
+        constexpr size_t memoryPageSize = 2 * 1024 * 1024;
+
+        auto roundUp = [&](size_t bytes) {
+            return ((bytes + memoryPageSize) / memoryPageSize) * memoryPageSize;
+        };
+
+        if(check == Check::None)
+            return userSize;
```
BrianHarrisonAMD: Nit: I think having this at the top of the function makes it a bit more clear.

driver/driver.hpp
averinevg: When it equals the page size, we have a one page buffer, but if it's smaller, this will be a problem, as it will simply not enough to compensate memory access issues. I also have concerns about unaligned data access. Wouldn't be better to simply add a page before or after the buffer? This way, we will always have a buffer with a constant size of reserved space.

BrianHarrisonAMD: I think we want the buffer to be aligned with the end or start of a page boundary to force consistently triggering memory faults. Since ASAN doesn't work with ASM kernels, I think we want this as a way to help reproduce memory access issues when they come up.

The alignment issues are interesting. I think we should be also checking for that, and failing gracefully if it's not a valid alignment for the kernel. Unless the expectation is that all the memory is aligned properly, then we should modify this to match that.

How would adding a page before & after help with detection?

JonathanLichtnerAMD: This is fixed.  Thanks for catching this!

driver/driver.hpp
JonathanLichtnerAMD: By using 2 MiB, this causes the GPU to raise a memory fault if the access is across a 2 MiB boundary.  I've updated the name of the lambda and the variable to make this clearer.

driver/driver.hpp
JonathanLichtnerAMD: Done

src/hip/handlehip.cpp
JonathanLichtnerAMD: Fixed.  Thanks for catching this!

src/hip/handlehip.cpp
JonathanLichtnerAMD: I've renamed it and added to the documentation

driver/driver.hpp
JonathanLichtnerAMD: The alignment issue is interesting.  The sub-buffers were padded out to a 256 byte alignment, which this change loses when we put the client buffer at the end of the larger buffer.  We could potentially add this padding back in but this would also possibly hide some OOB access issues (since reads or writes into the padding wouldn't be detected).  We could also add a similar padding for the GPUMem checking if we feel it's warranted.  However, Chris ran Brian's changes on a great number of shapes and I don't think he saw any failures that were due to the alignment.  But maybe we just didn't notice these because the memory fault issues were much more glaring :) 

@averinevg do we know the impact of this alignment.  I suspect this might affect performance, but I am new to GPU memory and kernel debugging.

driver/driver.hpp
JonathanLichtnerAMD: Fixed.   Thanks!

docs/how-to/debug-log.rst
```diff
@@ -319,6 +319,18 @@ settings to help with debugging:
 
 .. _control-parallel-compilation:
 
+Check for GPU sub-buffer out-of-bounds memory accesses
+==========================================================
+Use the ``MIOPEN_DEBUG_CHECK_SUB_BUFFER_OOB_MEMORY_ACCESS`` to help debug potential out-of-bounds (OOBs)
+memory access errors on GPU sub-buffers.  If the environment variable is undefined or set to zero, then no
+sub-buffer out-of-bounds detection is done.  Otherwise:
```
amd-jnovotny:
```suggestion
sub-buffer out-of-bounds detection is performed. To verify the memory accesses, use one of these values:
```

docs/how-to/debug-log.rst
```diff
@@ -319,6 +319,18 @@ settings to help with debugging:
...
+
+* ``MIOPEN_DEBUG_CHECK_SUB_BUFFER_OOB_MEMORY_ACCESS=1``: Check for OOBs before the start of the sub-buffers.
+* ``MIOPEN_DEBUG_CHECK_SUB_BUFFER_OOB_MEMORY_ACCESS=2``: Check for OOBs after the end of the sub-buffers.
+
+Note that this checking is not intended for production use since there is a possible performance hit
+from enabling this functionality.
```
amd-jnovotny:
```suggestion
Note that these checks are not intended for production use because they might cause a performance hit.
```

amd-jnovotny (approved):
Just a couple of suggestions for the docs.

src/hip/handlehip.cpp
averinevg: I meant that this text shouldn't be there

driver/driver.hpp
averinevg: I finally get how this works.

As for alignment, I think aligning to a large power of 2 should be sufficient; however, I believe this might potentially affect performance.

averinevg: Resolved

docs/how-to/debug-log.rst
JonathanLichtnerAMD: I've made these changes.  Thanks!

src/hip/handlehip.cpp
JonathanLichtnerAMD: Yes, I think I missed a few spots :)  I've removed all the MIOpenDriver text in this file now.

averinevg (approved):
LGTM

driver/bn_driver.hpp
```diff
@@ -526,31 +531,36 @@ int BatchNormDriver<TInput, Tref, TAcc, TScaleBias, TOut>::AllocateBuffersAndCop
 #if MIOPEN_BACKEND_OPENCL
     clGetCommandQueueInfo(q, CL_QUEUE_CONTEXT, sizeof(cl_context), &ctx, nullptr);
 #endif
-    status |= in.AllocOnDeviceAndInit(q, ctx, GetTensorSize(&in.GetTensor().desc));
+    status |= in.AllocOnDeviceAndInit(q, ctx, GetTensorSize(&in.GetTensor().desc), buffer_check);
```
randyspauldingamd, line 534: I think it's important to add unit tests for this so we know how successful it is. I started working out how to do that to give you a head start, and haven't quite got it working yet. Let's discuss before it goes in.

randyspauldingamd (changes requested):

driver/bn_driver.hpp
randyspauldingamd: Jonathan will add the Unit Tests in future PRs.

randyspauldingamd (approved):

driver/bn_driver.hpp
JonathanLichtnerAMD: Yes, unit tests for this would be great.  I've created story https://ontrack-internal.amd.com/browse/LWPMIOPEN-1443 to track the work to add unit tests.

PR #3624, opened 2025-03-14
[WORKAROUND] Disable solvers with abandoned buffer intrinsics in HIP 6.4

scerzh:
llvm.amdgcn.buffer.* intrinsics have been abandoned in HIP 6.4

A workaround has been created to disable the solvers which use these intrinsics

The definitions for the intrinsics are located in `src/kernels/static_composable_kernel/include/utility/static_kernel_amd_buffer_addressing.hpp` and wrapped in a template `SetData::Run` in `src/kernels/static_composable_kernel/include/utility/static_kernel_in_memory_operation.hpp` which is included in most of the sources in `src/kernels/static_composable_kernel`

`SetData::Run` is called from the following sources:
```
src/kernels/static_composable_kernel/src/kernel_wrapper/static_kernel_gridwise_convolution_backward_data_implicit_gemm_v1r1_ncdhw_kczyx_nkdhw.cpp
src/kernels/static_composable_kernel/src/kernel_wrapper/static_kernel_gridwise_convolution_backward_data_implicit_gemm_v1r1_nchw_kcyx_nkhw.cpp
src/kernels/static_composable_kernel/src/kernel_wrapper/static_kernel_gridwise_convolution_backward_data_implicit_gemm_v1r1_xdlops_nchw_kcyx_nkhw.cpp
src/kernels/static_composable_kernel/src/kernel_wrapper/static_kernel_gridwise_convolution_backward_data_implicit_gemm_v4r1_ncdhw_kczyx_nkdhw.cpp
src/kernels/static_composable_kernel/src/kernel_wrapper/static_kernel_gridwise_convolution_backward_data_implicit_gemm_v4r1_nchw_kcyx_nkhw.cpp
src/kernels/static_composable_kernel/src/kernel_wrapper/static_kernel_gridwise_convolution_backward_data_implicit_gemm_v4r1_xdlops_nchw_kcyx_nkhw.cpp
src/kernels/static_composable_kernel/src/kernel_wrapper/static_kernel_gridwise_convolution_backward_weights_implicit_gemm_v4r4_ncdhw_kczyx_nkdhw.cpp
src/kernels/static_composable_kernel/src/kernel_wrapper/static_kernel_gridwise_convolution_backward_weights_implicit_gemm_v4r4_nchw_kcyx_nkhw.cpp
[...]

PR #3670, opened 2025-04-07
[FIX] replacement for deprecated intrinsics.

shurale-nkn:
Fix for PR #3624


src/include/miopen/solver/implicitgemm_util.hpp
```diff
@@ -571,6 +569,13 @@ static inline auto get_static_ck_common_compiler_flag(const ExecutionContext& ct
     compiler_flag +=
         std::string(" -DCK_USE_AMD_V_FMAC_F32=") + (is_use_v_fmac_f32(ctx) ? '1' : '0');
 
+    compiler_flag += std::string(" -DCK_USE_AMD_BUFFER_PTR_TYPE=") +
+#if HIP_PACKAGE_VERSION_FLAT >= 6004000000
+                     '1'
+#else
+                     '0'
+#endif
+        ;
```
scerzh:
```suggestion
    compiler_flag += " -DCK_USE_AMD_BUFFER_PTR_TYPE="
#if HIP_PACKAGE_VERSION_FLAT >= 6004000000
                     "1";
#else
                     "0";
#endif
```
The legacy of C will serve

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdDataV1R1Xdlops.cpp
```diff
@@ -35,14 +35,7 @@
 // MIOPEN_DEBUG_CONV_IMPLICIT_GEMM_HIP_BWD_V1R1_XDLOPS=1 to enable it.
 #define WORKAROUND_SWDEV_251757 1
 
-// LLVM buffer intrinsics llvm.amdgcn.buffer.* have been removed in HIP 6.4
-#define WORKAROUND_SWDEV_498660 (HIP_PACKAGE_VERSION_FLAT >= 6004000000)
-
-#if WORKAROUND_SWDEV_498660
-#define SOLVER_NAME DISABLED_ConvHipImplicitGemmBwdDataV1R1Xdlops
-#else
 #define SOLVER_NAME ConvHipImplicitGemmBwdDataV1R1Xdlops
```
scerzh, line 38: Should we get rid of the macro (SOLVER_NAME) altogether?

scerzh (approved):
LGTM

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdDataV1R1Xdlops.cpp
shurale-nkn: I don’t really care either way. But I was glad I didn’t have to read all the code just to figure out the solver’s name.
@averinevg

averinevg: The macro can be safely removed, although it's not necessary. It was originally done this way so that the workaround could be easily removed without having to deal with the code. It's up to you.

averinevg (approved):
LGTM

test/gtest/unit_conv_solver_ConvHipImplicitGemmBwdDataV1R1Xdlops.cpp
averinevg: You can also remove `#ifndef HIP_PACKAGE_VERSION_FLAT`

scerzh (approved):
LGTM

PR #3679, opened 2025-04-15
[WA] disable failing tests at 908

shurale-nkn:
temporary patch. Revert after fix.


src/solver/conv/conv_hip_implicit_gemm_bwd_v1r1.cpp
```diff
@@ -637,6 +637,9 @@ size_t ConvHipImplicitGemmBwdDataV1R1::GetWorkspaceSize(const ExecutionContext&,
 bool ConvHipImplicitGemmBwdDataV1R1::IsApplicable(const ExecutionContext& ctx,
                                                   const ProblemDescription& problem) const
 {
+    if(ctx.GetStream().GetDeviceName() == "gfx908")
+        return false;
```
averinevg, line 641: If this is intended as a workaround, it should be clearly marked using the corresponding macro.

PR #3681, opened 2025-04-15
[tests] Rjs/swdev 527347

randyspauldingamd:
Follow-up to https://github.com/ROCm/MIOpen/pull/3678
Added tolerances for gfx950
Extended Tolerances to set multiple gpus
Added a few basic sanity checks to Tolerances

### DIFF
--- a/test/gtest/unit_conv_solver.cpp
+++ b/test/gtest/unit_conv_solver.cpp
@@ -26,6 +26,7 @@
 
 #include <miopen/conv/data_invoke_params.hpp>
 #include <miopen/conv/wrw_invoke_params.hpp>
+#include <miopen/errors.hpp>
 #include <miopen/generic_search.hpp>
 
 #include "unit_conv_solver.hpp"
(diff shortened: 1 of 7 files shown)


test/gtest/unit_conv_solver.cpp
```diff
@@ -244,11 +245,32 @@ uint64_t Tolerances::GetKey(Gpu gpu, miopenDataType_t type)
 
 void Tolerances::Set(Gpu gpu, miopenDataType_t type, float value)
 {
-    values[GetKey(gpu, type)] = value;
+    if(gpu == Gpu::None)
+        return;
+
+    int igpu = static_cast<int>(gpu);
+    if((igpu & (igpu - 1)) == 0)
```
BrianHarrisonAMD, line 252: heh neat.

test/gtest/unit_conv_solver.cpp
```diff
@@ -244,11 +245,32 @@ uint64_t Tolerances::GetKey(Gpu gpu, miopenDataType_t type)
 
 void Tolerances::Set(Gpu gpu, miopenDataType_t type, float value)
 {
-    values[GetKey(gpu, type)] = value;
+    if(gpu == Gpu::None)
+        return;
```
BrianHarrisonAMD, line 249: Gpu::All is a special case.
Seems like we should handle that as well here.

BrianHarrisonAMD (approved):
Looks good to me.

I think adding a handler for all would be nice as well, but don't think it's blocking.

test/gtest/unit_conv_solver.cpp
randyspauldingamd: It's intended to let you set them all if you like. Hmm, I forgot to add a UT for that.

randyspauldingamd:
> Looks good to me.
> 
> I think adding a handler for all would be nice as well, but don't think it's blocking.

I added the UT for that. Thanks for the suggestion!

test/gtest/unit_conv_solver_tests.cpp
averinevg, line 1: Unit test for a unit test? Then we'll need yet another unit test for this test :smile:

test/gtest/unit_conv_solver.cpp
```diff
@@ -244,11 +245,32 @@ uint64_t Tolerances::GetKey(Gpu gpu, miopenDataType_t type)
...
+
+    int igpu = static_cast<int>(gpu);
```
averinevg, line 251:
```suggestion
    auto igpu = static_cast<std::underlying_type_t<Gpu>>(gpu);
```

test/gtest/unit_conv_solver.cpp
```diff
@@ -244,11 +245,32 @@ uint64_t Tolerances::GetKey(Gpu gpu, miopenDataType_t type)
...
+            }
+        }
+    }
 }
 
 float Tolerances::Get(Gpu gpu, miopenDataType_t type) const
 {
+    MIOPEN_THROW_IF((static_cast<int>(gpu) & (static_cast<int>(gpu) - 1)) != 0,
+                    "cannot call Tolerance::Get with multiple gpus");
+    MIOPEN_THROW_IF(gpu > Gpu::gfxLast, "Tolerance::Get called with invalid gpu");
```
averinevg, line 272: Technically, we can call the function, we just get default value. It might be better to use assertions instead of exceptions here - we don't expect any random data.

test/gtest/unit_conv_solver.cpp
randyspauldingamd: I like it. This need to be staged today so I'll get it next week.

PR #3690, opened 2025-04-22
Enable conv ck partial support on Navi31

iq136boy:
Enable CK convolution solvers on Navi31 (partially support)


BradPepersAMD:
Closing for now since actual coverage for data types is very low

PR #3691, opened 2025-04-22
BN: NHWC optimization

EnricoDeg:
Summary:

 - N parallelization
 - Improve reduction NHWC
 - Add support for vector size 2 and 8
 - Improve parallelization of Final kernels
 - Extend parameter space of tuning (workgroup sizes)
 - Fix stash values for bwd when saved mean and variance are used (reduce limitations)
 - Use heuristics to limit the number of instances (`xlocalsize` and `vectorsize`)

Performance should be compared using tuning in both `develop` branch and this branch (setting `MIOPEN_FIND_MODE=1`  `MIOPEN_FIND_ENFORCE=4`).


BradPepersAMD:
Extensive changes and this is where I hope our investment into deeper unit test coverage pays off since it's pretty difficult to fully validate changes of this extent.  Will wait for a successful CI run to validate before approving and merging.

BradPepersAMD:
I guess the one thing missing, from this PR and our own CI, is validation of what this does to performance, across a wide range of BN shapes and data types, for both OOTB and tuned.  What we need to see is that OOTB performance has not been significantly impacted and of course wanting to see that there is an uplift either OOTB or at the very least with tuning.

EnricoDeg:
> I guess the one thing missing, from this PR and our own CI, is validation of what this does to performance, across a wide range of BN shapes and data types, for both OOTB and tuned. What we need to see is that OOTB performance has not been significantly impacted and of course wanting to see that there is an uplift either OOTB or at the very least with tuning.

This is a good idea. In this specific case, I'm not expecting significant changes with OOTB performance but there should be with tuning.

src/include/miopen/batchnorm/common_spatial.hpp
```diff
@@ -423,26 +556,208 @@ inline void DefaultConfigSpatialMultiple(const miopen::batchnorm::ProblemDescrip
...
     unsigned int in_cstride = h * w;
 
-    // Default configuration for spatial multiple tries to use vectorization
-    // for both NCHW or NHWC
-    size_t vectorsize =
-        problem.IsLayoutNHWC() ? (c % 4 == 0 ? 4 : 1) : (in_cstride % 4 == 0 ? 4 : 1);
-    if(IsSpatialMultipleApplicable(problem, vectorsize, stash_values))
+    size_t xlocalsize_default = 0;
+    size_t ylocalsize_default = 0;
+    size_t vectorsize_default = 4;
```
shurale-nkn, line 561: This line breaks the testing in CI. 
Old code: 
`size_t vectorsize =problem.IsLayoutNHWC() ? (c % 4 == 0 ? 4 : 1) : (in_cstride % 4 == 0 ? 4 : 1);`
The new code will fail in cases where C<4.

EnricoDeg: Interesting. Inside `IsSpatialMultipleApplicable()` I check:
```
        if(c % vectorsize != 0)
        {
            return false;
        }
```
So in that case, it should revert to `vectorsize = 1`. My tests were passing locally but maybe I'm missing sth. Can you provide the MIOpenDriver command of a failing case?

EnricoDeg: I understand it now. The problem is that the configuration is computed even when the vectorsize is not supported, then the sanitizer complains.

I fixed the issue in this PR: https://github.com/ROCm/MIOpen/pull/3727

I'm not computing the configuration in that cases

PR #3693, opened 2025-04-22
Disabled rocblas for fp8 in MIOpen when rocblas version is not in the range

iq136boy:
Disabled rocblas for fp8 in MIOpen

### DIFF
--- a/src/gemm_v2.cpp
+++ b/src/gemm_v2.cpp
@@ -96,15 +96,6 @@ FlagsForRocblasFp32Fp16Call(const miopen::GemmDescriptor& desc) // bool gfx90aFp
 #endif
 }
 
-bool IsFP8Supported(const std::string& device_name)
-{
-#if USE_ROCBLAS_GEMM_EX3
-    return device_name == "gfx942" || miopen::StartsWith(device_name, "gfx95");
-#else
-    return false;
-#endif
-}
-
 #if USE_ROCBLAS_GEMM_EX3
 static inline rocblas_computetype rocBlasComputeType_ex3(const miopen::GemmDescriptor& desc)
 {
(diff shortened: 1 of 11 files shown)


BrianHarrisonAMD:
Looking at mha_common.hpp, it looks like if ex3 is disabled, then it'll silently not perform the GEMM, and give incorrect results:

https://github.com/ROCm/MIOpen/pull/3693/files#diff-e457b11a82e95af506e87c229d051344438a842e787f8f6effcaf2509c243963R182-R226

Might make sense to throw here instead as a temporary workaround.

iq136boy:
> Looking at mha_common.hpp, it looks like if ex3 is disabled, then it'll silently not perform the GEMM, and give incorrect results:
> 
> https://github.com/ROCm/MIOpen/pull/3693/files#diff-e457b11a82e95af506e87c229d051344438a842e787f8f6effcaf2509c243963R182-R226
> 
> Might make sense to throw here instead as a temporary workaround.

@BrianHarrisonAMD Yes. It seems that all the solver tests that called roclblas gemm opertions failed with the workaround. I have disabled it.

BrianHarrisonAMD (approved):

randyspauldingamd (approved):
LGTM

src/gemm_v2.cpp
```diff
@@ -264,6 +268,15 @@ MIOPEN_DECLARE_ENV_VAR_UINT64(MIOPEN_GEMM_ENFORCE_BACKEND)
 
 namespace miopen {
 
+bool IsFP8Supported(const std::string& device_name)
```
randyspauldingamd, line 271: Function does not exist outside of the #if MIOPEN_USE_ROCBLAS above.

randyspauldingamd (dismissed):
~~Missing a fn body~~
NVM, false alarm!

PR #3694, opened 2025-04-23
Rjs/intrinsic mangling

randyspauldingamd:
Disables the following solvers and associated tests on gfx942 and gfx950:
ConvHipImplicitGemmBwdDataV1R1Xdlops
ConvHipImplicitGemmBwdDataV4R1Xdlops
ConvHipImplicitGemmForwardV4R4Xdlops
ConvHipImplicitGemmForwardV4R4Xdlops_Padded_Gemm
ConvHipImplicitGemmForwardV4R5Xdlops
ConvHipImplicitGemmWrwV4R4Xdlops
ConvHipImplicitGemmWrwV4R4Xdlops_Padded_Gemm

Issue was found after disabling the v_mac_f32 kernels:

[ RUN      ] Smoke/GPU_UnitTestConvSolverHipImplicitGemmBwdDataV1R1XdlopsBwd_BFP16.ConvHipImplicitGemmBwdDataV1R1Xdlops/0
Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.raw.ptr.buffer.load.v2i16
ptr @llvm.amdgcn.raw.ptr.buffer.load.v2bf16
Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.raw.ptr.buffer.load.v4i16
ptr @llvm.amdgcn.raw.ptr.buffer.load.v4bf16
Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.raw.ptr.buffer.store.i16
ptr @llvm.amdgcn.raw.ptr.buffer.store.bf16
LLVM ERROR: Broken module found, compilation aborted!
Aborted (core dumped)

[ RUN      ] Smoke/GPU_UnitTestConvSolverHipImplicitGemmBwdDataV4R1XdlopsBwd_BFP16.ConvHipImplicitGemmBwdDataV4R1Xdlops/0
Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.raw.ptr.buffer.load.v4i16
ptr @llvm.amdgcn.raw.ptr.buffer.load.v4bf16
Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.raw.ptr.buffer.store.i16
ptr @llvm.amdgcn.raw.ptr.buffer.store.bf16
LLVM ERROR: Broken module found, compilation aborted!
Aborted (core dumped)

I suspect these have not been seen before because miopen_gtest was segfaulting before the tests were executed. The solvers shall be disabled for 6.5, and the issues will be discussed among the team.


randyspauldingamd:
Implemented as a branch from https://github.com/ROCm/MIOpen/pull/3682

src/solver/conv/conv_hip_implicit_gemm_fwd_v4r1.cpp
```diff
@@ -113,11 +116,14 @@ bool ConvHipImplicitGemmV4R1WrW::IsApplicable(const ExecutionContext& ctx,
         return false;
     if(!problem.IsFp32() && !problem.IsFp16() && !problem.IsBfp16())
         return false;
+    const std::string name = ctx.GetStream().GetDeviceName();
+    if(problem.IsFp32() && (name == "gfx942" || name == "gfx950"))
+        return false;
```
BrianHarrisonAMD: I think this is from the other PR, but why is only FP32 being skipped for gfx942 & gfx950?

src/solver/conv/conv_hip_implicit_gemm_bwd_v1r1.cpp
```diff
@@ -657,10 +657,11 @@ bool ConvHipImplicitGemmBwdDataV1R1::IsApplicable(const ExecutionContext& ctx,
         return false;
     if(!problem.Is2d() && !(problem.Is3d() && problem.IsFp32()))
         return false;
-
     if(!(problem.IsFp32() || problem.IsBfp16()))
         return false;
-
+    const std::string name = ctx.GetStream().GetDeviceName();
+    if(name == "gfx942" || name == "gfx950")
+        return false;
```
BrianHarrisonAMD: Do we know if this is a work around, or are we leaving this? 

I think these are typically disabled anyways due to:

```
if(ThisSolverIsDeprecatedStatic::IsDisabled(ctx))
        return false;
```
And they are re-enabled for tests.

If we are going to leave this, then I think it's fine as is, but if we want to come back and fix it later, then we should wrap these in a workaround define referencing the issue. 

(This would apply to all the changes, but just commenting here)

randyspauldingamd: Weird. Yeah, I'm trying to avoid the clutter of an actual workaround. I think it's easy enough to find them all without it, but it wouldn't take much to persuade me to add it.

src/solver/conv/conv_hip_implicit_gemm_fwd_v4r1.cpp
randyspauldingamd: Correct. Commented on in the other PR.

BrianHarrisonAMD (approved):
Approving assuming the one change on the other PR is merged into this as well to deal with the extra fp32 check.

I think adding the workaround define comes down to if we plan to come back and remove this eventually.

cderb (approved):

JonathanLichtnerAMD (approved):

PR #3695, opened 2025-04-23
Fused batch norm and activation

EnricoDeg:
Summary:

 - Add activation in spatial BN kernels (fwd train and inference and bwd)
 - Extend batch norm spatial solvers to support activation functions
 - Extend batch norm API
 - Add tests
 - Add MIOpenDriver support for fused bnorm+activation


BradPepersAMD (approved):
Ok green CI and looks good to me.  Will merge.

PR #3697, opened 2025-04-25
[Hotfix]delete the buggy mha f8 code section

iq136boy:
delete the buggy mha f8 code section

### DIFF
--- a/src/solver/mha/mha_common.hpp
+++ b/src/solver/mha/mha_common.hpp
@@ -128,6 +128,18 @@ inline void gemm(const Handle& handle,
 
     float beta = 0.0f;
 
+    auto cvtMiopen2Rocblas = [](miopenDataType_t miopen) {
+        switch(miopen)
+        {
+        case miopenFloat: return rocblas_datatype::rocblas_datatype_f32_r;
+#if USE_ROCBLAS_EX3
+        case miopenFloat8_fnuz: return rocblas_datatype::rocblas_datatype_f8_r;
+        case miopenBFloat8_fnuz: return rocblas_datatype::rocblas_datatype_bf8_r;
+#endif
+        default: return rocblas_datatype::rocblas_datatype_invalid;
+        }
+    };
+
     // fp32 x fp32 case
     if(AType == miopenFloat && AType == BType)
     {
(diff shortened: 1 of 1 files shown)


src/solver/mha/mha_common.hpp
```diff
@@ -173,16 +163,6 @@ inline void gemm(const Handle& handle,
...
     }
-    // only bfp8 x fp32, fp32 x bfp8 and [b]fp8 x [b]fp8 combinations are supported
-    else if(cvtMiopen2Rocblas(AType) != rocblas_datatype::rocblas_datatype_invalid             //
-            && cvtMiopen2Rocblas(BType) != rocblas_datatype::rocblas_datatype_invalid          //
-            && (AType == BType || AType == miopenBFloat8_fnuz || BType == miopenBFloat8_fnuz)) //
-    {
-        assert(handle.GetDeviceName() == "gfx942");
-#if USE_ROCBLAS_EX3
-        MIOPEN_THROW("rocblas GEMM operations is not supported!");
-#endif
```
TorreZuk: I think you have this backwards, if USE_ROCBLAS_EX3 is true you have the f8 types and the API to call.
This is !USE_ROCBLAS_EX3 or #else (original PR merged with opposite of what I expected).

BradPepersAMD: ok thanks.  looks like we just removed the check entirely so if it was wrong, and your analysis seems to strongly indicate it was, it's gone now so should be fine?

BrianHarrisonAMD: It's a bit mixed now.

The gemm_v2.cpp code allows using the EX3 API when it's available, and properly prevents using it when it's not there. 

Inside mha_common.hpp, we have just ripped out the functionality entirely.

I think it would make sense to either:
- Fully rip out the usage of ROCBLAS_EX3 from both places.
- Keep the original code, and update it to work properly with the USE_RCOBLAS_EX3 define. 

I think this will resolve the compilation issue, but both files handled it differently.

TorreZuk: You decide what you want to do but inside a #if USE_ROCBLAS_EX3  you can use the types and call _ex3 API as you did in earlier releases.

BradPepersAMD: The original patch did this:
image
which I think was incorrect and we could just restore the original code since it was checking USE_ROCBLAS_EX3 properly and should still work when allowed.  perhaps an else on it to do the log message?

BrianHarrisonAMD: Yes I think we should bring that back, and have an else that throws.

BrianHarrisonAMD: We would also need a define check inside cvtMiopen2Rocblas where it's referencing the types that are removed though.

TorreZuk: If you install rocBLAS etc. from any ROCm 6.x era package you can build and test all functions work just as they did as before.  #if !defined(USE_ROCBLAS_EX3) is where I would expect the new fallback code or exception throwing.  [In mha_common.hpp](https://github.com/ROCm/MIOpen/commit/14722754d23d40058849a41ec0164940756227ed#diff-e457b11a82e95af506e87c229d051344438a842e787f8f6effcaf2509c243963)   deletes code it could have kept calling.

iq136boy: > You decide what you want to do but inside a #if USE_ROCBLAS_EX3 you can use the types and call _ex3 API as you did in earlier releases.

@TorreZuk I think I would make the changes base on your comments. I would bring back the code and add #if USE_ROCBLAS_EX3 properly base on what you have concerned.

iq136boy: > I think you have this backwards, if USE_ROCBLAS_EX3 is true you have the f8 types and the API to call. This is !USE_ROCBLAS_EX3 or #else (original PR merged with opposite of what I expected).

Yes. the logic in the delete section is not correct. should be the opposite. I will fix it.

iq136boy:
@TorreZuk @BrianHarrisonAMD @BradPepersAMD  A fix has been pushed. Please review.

TorreZuk (approved):
That makes more sense to me

src/solver/mha/mha_common.hpp
BradPepersAMD: looks good to me now!

BradPepersAMD (approved):

Issue #2851, opened 2024-03-26
"-mno-xnack" flag is removed by upstream clang


littlewu2508:
Building and testing MIOpen with upstream clang, I got some failures. One is `test_conv2d`:

The output of `AMD_COMGR_SAVE_TEMPS=1 AMD_COMGR_REDIRECT_LOGS=stdout AMD_COMGR_EMIT_VERBOSE_LOGS=1 /run/user/18014/portage/sci-libs/miopen-5.7.1-r1/work/MIOpen-rocm-5.7.1_build/bin/test_conv2d --float --cmode conv --pmode valid --group-count 1 --batch_size 8 --input_channels 32 --output_channels 32 --spatial_dim_elements 28 28 --filter_dims 5 5 --pads_strides_dilations 1 1 1 1 1 1 --trans_output_pads 0 0 --in_layout NCHW --fil_layout NCHW --out_layout NCHW --deterministic 0 --tensor_vect 0 --vector_length 1 --output_type int32 --int8_vectorize 0` is:

[miopen-test_conv2d-comgr.log](https://github.com/ROCm/MIOpen/files/14752148/miopen-test_conv2d-comgr.log)

Original post is in https://github.com/ROCm/MIOpen/issues/990#issuecomment-2012468648 which points out that `-mno-xnack` is not supported by upstream clang. Maybe we should remove the "-target-feature", "-xnack" from `optAsm` rather than adding a `-mno-xnack`

atamazov:
@littlewu2508 Yep, it seems like after some recent changes in the compiler, `-m[no-]<target-feature>` can't be used with `xnack`, and XNACK should be specified in `-mcpu`. Ref: https://llvm.org/docs/AMDGPUUsage.html#id117. I am wondering why `-m[no-]xnack` normally works with latest ROCm.

I'll look into this. But please let us know hot urgent/important is this issue for you.

/cc @junliume @JehandadKhan 

<details><summary><h3>ℹ️ Copy of the discussion thread from #990</h3></summary><hr>

_@littlewu2508:_
The `mno-xnack` is removed in upstream clang [llvm/llvm-project@5cae708](https://github.com/llvm/llvm-project/commit/5cae70800266119bbf319675a175cba9a7f315b1) in 2020. However it seems MIOpen is still using that:

https://github.com/ROCm/MIOpen/blob/f34a90f7699e4af0e021ff0bf4111b1c407bdec1/src/comgr.cpp#L1013

[...]

PR #3699, opened 2025-04-25
Allow targetid for mcpu kernel arguments

cderb:
Resolve issue for generating system kdb #3001 
unblock #2891 resolving #2851


test/gtest/db_sync.cpp
```diff
@@ -520,11 +522,50 @@ void SetupPaths(fs::path& fdb_file_path,
...
+namespace miopen {
+struct TestHandle : Handle
+{
+    TestHandle(size_t _num_cu) : Handle(), num_cu(_num_cu) {}
+
+// Probably, according to the idea of the author of this test, the number of CUs should have been
+// substituted with the value passed to the constructor (which in fact did not happen). After
+// https://github.com/ROCm/MIOpen/pull/3175, the method became virtual, the substitution actually
+// happened, and the test broke. I disabled that part (since it doesn't work as intended anyway) to
+// keep its behavior the same.
```
averinevg: This comment and the `#if/#endif` can be removed. I wrote the comment and disabled the code below, but now it is enabled again, someone probably fixed it.

cderb: I'll remove the comment and #defs.

src/include/miopen/target_properties.hpp
```diff
@@ -29,7 +29,7 @@
 #include <boost/optional.hpp>
 #include <string>
 
-#define WORKAROUND_ISSUE_3001 1
+#define WORKAROUND_ISSUE_3001 0
```
BrianHarrisonAMD: Can this be removed?

BrianHarrisonAMD (dismissed):
LGTM, but probably need to update, and re-run CI.

src/include/miopen/target_properties.hpp
cderb: yes, I think so.

amd-hsivasun:

PR #3702, opened 2025-04-28
Add Winograd Rage Solver

scerzh:
Two asm kernels were integrated: version 4.6 and 4.7

4.6 is selected when filter sizes are `R <= 3` and `S <= 3`
4.7 is selected when `R > 3` and `S > 3`

The kernels were tested only for MI300

`.altmacro` support has been extended in ROCm 6.4 leading to macro arguments are expanded even if they are not preceded by a `\` and causing sometimes compiler errors or incorrect binary output. Every assembler kernel should be verified over both new and old `.altmacro` support. This time it has been done in the following way.

* Compile the source

```bash
/opt/rocm/llvm/bin/clang -x assembler -target amdgcn-amd-amdhsa -mcpu=gfx942 -I. -Wa,-defsym,ROCM_METADATA_VERSION=5 -mcumode -mwavefrontsize64 -c Conv_Winograd_Rage_v4_6_0_fp16_fp32acc_f2x3_stride1.s
```

```bash
/opt/rocm/llvm/bin/clang -x assembler -target amdgcn-amd-amdhsa -mcpu=gfx942 -I. -Wa,-defsym,ROCM_METADATA_VERSION=5 -mcumode -mwavefrontsize64 -c Conv_Winograd_Rage_v4_7_0_fp16_fp32acc_f2x3_stride1.s
```

* Extract sections

```bash
/opt/rocm/llvm/bin/llvm-objdump --full-contents Conv_Winograd_Rage_v4_6_0_fp16_fp32acc_f2x3_stride1.o > Conv_Winograd_Rage_v4_6_0_fp16_fp32acc_f2x3_stride1.txt
```

```bash
/opt/rocm/llvm/bin/llvm-objdump --full-contents Conv_Winograd_Rage_v4_7_0_fp16_fp32acc_f2x3_stride1.o > Conv_Winograd_Rage_v4_7_0_fp16_fp32acc_f2x3_stride1.txt
```

* Perform the above steps for both compiler versions
* Compare the resulted outputs


src/solver/conv/conv_wino_rage_RxS.cpp
```diff
@@ -0,0 +1,282 @@
...
+    /// uses it to find out required CO version (hack).
+    /// \todo Delete when COv2 support is removed.
+    KernelBuildParameters options{
+        {"ROCM_METADATA_VERSION", 5},
+    };
+    kernelInfo.comp_options = options.GenerateFor(kbp::GcnAsm{});
+    kernelInfo.comp_options += std::string(" -mcumode -mwavefrontsize64");
+
+    /// \todo Add case for gfx12 wgp size of 384U
+    uint64_t wgSize = 768U;
```
reidkwja, line 207: Instead of hardcoding, consider querying the wavefront/group size from the device - or at least defining it as named constant. Something like this: `constexpr uint64_t DefaultGfx9WgSize = 768;`

src/solver/conv/conv_wino_rage_RxS.cpp
```diff
@@ -0,0 +1,282 @@
...
+{
+    return static_cast<T>(1) << exp;
+}
+
+constexpr uint64_t maxNGroups = PowOf2<16>() - 1;
+
+bool isShaderConstraintsMetV4_6(const WinoShaderArgs& args)
+{
+    // clang-format off
+    return args.N < PowOf2<16>()
```
reidkwja: Given the long list of &&, it might be a better idea to group related checks in predicates - to enhance readability and future refactoring. 

Something like this: 
```
bool dimsWithin16bit(const WinoShaderArgs& a) {
  return a.N < PowOf2<16>() && a.C < PowOf2<16>() && /* add other similar items */;
}
bool accumWithin31bit(const WinoShaderArgs& a) {
  return (uint64_t(a.N-1)*a.C+1)*a.H*a.W < PowOf2<31>() && /* other items here too*/;
}
return dimsWithin16bit(args) && accumWithin31bit(args) && /* return all other predicates in similar ways */;

```

src/solver/conv/conv_wino_rage_RxS.cpp
shurale-nkn: @reidkwja ?
Looks like you’re mixing things up. That’s not how it works.

test/gtest/unit_conv_solver_ConvWinoRageRxS.cpp
```diff
@@ -0,0 +1,144 @@
...
+auto GetConvTestCases(miopenDataType_t datatype)
+{
+    using TestCase = miopen::unit_tests::ConvTestCase;
+
+    return std::vector{
+        // clang-format off
+        TestCase{{ 1, 16, 1080, 1920}, {16, 16, 3, 3}, {1, 1}, {1, 1}, {1, 1}, datatype},
+        TestCase{{ 1,  8, 1080, 1920}, {16,  8, 3, 3}, {1, 1}, {1, 1}, {1, 1}, datatype},
+        TestCase{{ 1, 16, 1080, 1920}, {16, 16, 1, 1}, {0, 0}, {1, 1}, {1, 1}, datatype},
+        TestCase{{ 1, 16, 1080, 1920}, { 4, 16, 3, 3}, {1, 1}, {1, 1}, {1, 1}, datatype},
```
shurale-nkn: Are these really the necessary sizes? They seem like quite large tensors for a unit test.

src/solver/conv/conv_wino_rage_RxS.cpp
```diff
@@ -0,0 +1,282 @@
...
+    if(!(problem.GetDilationH() == 1 && problem.GetDilationW() == 1))
+        return false;
+
+    WinoShaderArgs args;
+    if(!args.SetConvParams(problem))
+        return false;
+
+    args.n_groups = getNGroups(ctx);
+
+    return isShaderConstraintsMetV4_6(args);
```
shurale-nkn: Xnack now supported?

shurale-nkn (changes requested):

src/solver/conv/conv_wino_rage_RxS.cpp
averinevg: Yea, this needs to be checked for all asm kernels on gfx9xx

test/gtest/unit_conv_solver_ConvWinoRageRxS.cpp
averinevg: Yes, that's too much for smoke tests. These test cases should probably be moved to the full tests. Do they check specific networks or corner cases?

scerzh: Agree, it's too much for a Smoke test.
I tried the common cases for CNNs.

src/solver/conv/conv_wino_rage_RxS.cpp
scerzh: Good catch, thanks!

src/solver/conv/conv_wino_rage_RxS.cpp
scerzh: Done.

src/kernels/winograd/Conv_Winograd_Rage_v4_7_0_fp16_fp32acc_f2x3_stride1.s
```diff
@@ -27,9 +27,7 @@
 
 .if (.amdgcn.gfx_generation_number == 9 && .amdgcn.gfx_generation_minor == 4 && .amdgcn.gfx_generation_stepping == 2)
     KERNEL_PROLOG _fp16_fp32acc_f2x3_stride1
-    .noaltmacro
```
shurale-nkn: is this change verified at rocm >=6.4 ???

src/include/miopen/conv/kernel_interface/winograd_kernel_interface.hpp
```diff
@@ -145,6 +145,62 @@ struct WinoShaderArgsV2
...
+
+    bool paddedSizesFit16bits() const noexcept
+    {
+        // clang-format off
+        return static_cast<int64_t>(pad_h) + H <= PowOf2<16, int64_t>()
+            && static_cast<int64_t>(pad_w) + W <= PowOf2<16, int64_t>()
+            && std::abs(static_cast<int64_t>(pad_h)) + out_h + R <= PowOf2<16, int64_t>()
+            && std::abs(static_cast<int64_t>(pad_w)) + out_w + S <= PowOf2<16, int64_t>();
+        // clang-format on
+    }
```
averinevg, line 203: These functions are not related to the kernel interface. I think this should belong to a specific solver or a group of solvers, and it should also be placed in a .cpp file rather than in a .hpp.

src/kernels/winograd/Conv_Winograd_Rage_v4_6_0_metadata.inc
```diff
@@ -0,0 +1,212 @@
...
+  - .name: \kernel_name
+    .symbol: \kernel_name\().kd
+    .sgpr_count: \sc
+    .vgpr_count: \wc
+    .group_segment_fixed_size: 65536
+    .private_segment_fixed_size: 0
+    .kernarg_segment_size: 232
+    .kernarg_segment_align: 8
+    .reqd_workgroup_size: [ \wg_x, 1, 1 ]
+    .max_flat_workgroup_size: 1024
```
averinevg: Inconsistency between `.reqd_workgroup_size` and `.max_flat_workgroup_size`
https://llvm.org/docs/AMDGPUUsage.html#amdgpu-amdhsa-code-object-kernel-metadata-map-table-v3

src/include/miopen/conv/kernel_interface/winograd_kernel_interface.hpp
scerzh: I put them here to allow compiler to optimize inline calls.

I saw that the restrictions in the recent versions of Fury and Rage kernels and probably in the upcoming Ultra kernel are mostly common and therefore are going to belong to all Winograd solvers. I'm going to reuse these functions across them.

We have similar member functions in ProblemDescription which are reused in a variety of solvers.

src/kernels/winograd/Conv_Winograd_Rage_v4_7_0_fp16_fp32acc_f2x3_stride1.s
scerzh: Yes, I did the verification.

src/kernels/winograd/Conv_Winograd_Rage_v4_6_0_metadata.inc
scerzh: I interpreted the word "consistent" as `.max_flat_workgroup_size >= .reqd_workgroup_size` and found the value of 1024 in the assembler listings I received from Timofei.

Now double checking this.

scerzh: The Assembler team has successfully tested values that meet the condition `.max_flat_workgroup_size >= .reqd_workgroup_size`. The value `1024` was sourced from https://github.com/llvm/llvm-project/blob/llvmorg-19.1.7/llvm/lib/Target/AMDGPU/Utils/AMDGPUBaseInfo.cpp#L987

I'll update it to `\wg_x` to align to our standard shader conventions.

averinevg: I wrote about inconsistency, though it's actually fine. `.max_flat_workgroup_size` is a mandatory field that defines the maximum work-group size, and in our case that's the only size applied. There is no point in setting a value different from `.reqd_workgroup_size`.

src/kernels/winograd/Conv_Winograd_Rage_v4_7_0_fp16_fp32acc_f2x3_stride1.s
scerzh: Added the verification steps to the description.
I also discovered that not all versions of rocm>=6.4 support the new `.altmacro` behavior. Some still behave the old way and it's hard to tell which ones do without trying them out.

src/kernels/winograd/Conv_Winograd_Rage_v4_6_0_metadata.inc
scerzh: Done.

shurale-nkn (approved):
LGTM

averinevg (approved):
LGTM

PR #3715, opened 2025-05-05
Skip FindCore when user specified workspace is smaller than required

xinlipn:
Customer reports seg fault when user specified workspace size, e.g. 1073741824 is smaller than required 5452602112.

1. When workspace is set to be 1073741824, fallback is false, and line 161 will be skipped in default Hybrid mode
https://github.com/ROCm/MIOpen/blob/b0143f43d50168f3e9c91887c5fe52c50d118bd1/src/ocl/convolutionocl.cpp#L157-L161

2. Then this blurb will be called. FindCore() will encounter seg fault with smaller workspace size.
https://github.com/ROCm/MIOpen/blob/b0143f43d50168f3e9c91887c5fe52c50d118bd1/src/ocl/convolutionocl.cpp#L174-L190

Below are the logs before the changes:
Workspace size: 1073741824, with seg fault
[1GB Seg Faultg.txt](https://github.com/user-attachments/files/20032160/1GB.Seg.Faultg.txt)

Workspace size: 5452602112, no seg fault
[5452602112 tuning.txt](https://github.com/user-attachments/files/20032168/5452602112.tuning.txt)

Log after the changes:
Workspace size: 1073741824, warning messages and no seg fault
[1GB No Seg Fault after.txt](https://github.com/user-attachments/files/20038541/1GB.No.Seg.Fault.after.txt)

MIOpen(HIP): Warning [FindConvolution] Skip FindCore possible cause: user specified workspace is smaller than required
MIOpen Error: /MIOpen/src/ocl/convolutionocl.cpp:1132: No suitable algorithm was found to execute the required convolution
MIOpen error: 7 at line 135


src/ocl/convolutionocl.cpp
```diff
@@ -188,6 +189,11 @@ std::vector<Solution> FindConvolution(const ExecutionContext& ctx,
                             force_attach_binary);
         });
     }
+    else
+    {
+        MIOPEN_LOG_W(
+            "Skip FindCore possible cause: user specified workspace is smaller than required");
+    }
```
cderb: If this branch is taken the function will produce an empty result vector. Can we supply a default or naive result.

src/ocl/convolutionocl.cpp
```diff
@@ -171,7 +172,7 @@ std::vector<Solution> FindConvolution(const ExecutionContext& ctx,
         CompileSolution(id, ctx, problem);
         results.push_back({id, sol->time, s.GetWorkspaceSize(ctx, problem)});
     }
-    else
+    else if(!skip_find_core)
```
BrianHarrisonAMD: I don't think we want to skip find_core since we still want to evaluate the solutions that are applicable.

There is already a mechanism inside EvaluteInvokers for skipping the solution if it doesn't have enough workspace:
- https://github.com/ROCm/MIOpen/blob/2c2fd43c57a2ecd55c03e0e0b6f4e2d8f3929cb2/src/conv/solver_finders.cpp#L239-L255

I thought this should be preventing these issues, why is it not working in this case?

BrianHarrisonAMD: Maybe having issues with the full find loop?

https://github.com/ROCm/MIOpen/blob/2c2fd43c57a2ecd55c03e0e0b6f4e2d8f3929cb2/src/include/miopen/generic_search.hpp#L475-L593

src/ocl/convolutionocl.cpp
xinlipn: > If this branch is taken the function will produce an empty result vector. Can we supply a default or naive result.

@cderb , I have updated the PR that when provided workspace is smaller than required. Immediate fallback kernels will be picked, i.e. ConvDirectNaiveConvWrw in this case. Log is as below

[ConvDirectNaiveConvWrw.log](https://github.com/user-attachments/files/20211748/ConvDirectNaiveConvWrw.log)

src/ocl/convolutionocl.cpp
xinlipn: > Maybe having issues with the full find loop?
> 
> https://github.com/ROCm/MIOpen/blob/2c2fd43c57a2ecd55c03e0e0b6f4e2d8f3929cb2/src/include/miopen/generic_search.hpp#L475-L593

@BrianHarrisonAMD , indeed it's the find loop causing the seg fault as this line below. 

https://github.com/ROCm/MIOpen/blob/2c2fd43c57a2ecd55c03e0e0b6f4e2d8f3929cb2/src/include/miopen/generic_search.hpp#L492

xinlipn: @BrianHarrisonAMD , please see the reply to @cderb above. Thanks

src/ocl/convolutionocl.cpp
```diff
@@ -156,10 +157,12 @@ std::vector<Solution> FindConvolution(const ExecutionContext& ctx,
...
         // override the normal find with immed mode with env var
-        if(!sols.empty() && (!(findMode.IsHybrid(ctx) && fallback) ||
-                             env::enabled(MIOPEN_DEBUG_FORCE_IMMED_MODE_FALLBACK)))
+        if(!sols.empty() &&
+           (!(findMode.IsHybrid(ctx) && fallback) ||
+            env::enabled(MIOPEN_DEBUG_FORCE_IMMED_MODE_FALLBACK) || less_workSpaceSize))
             sol = sols.front();
         // In Hybrid Find mode, we use Normal Find instead of Immediate fallback kernels.
+        // Use immediate fallback kernels when provided workspace is smaller than required workspace
     }
```
BrianHarrisonAMD: Won't this still hit FindCore if it's not fast, hybrid mode, or if GetSolutions return is empty?
I.E if you have tuning enabled, but provide < max workspace, then it will still hit issues.

Also, why does GenericSearch get triggered in this case?
I thought it should only happen if tuning was enabled, but if this fixes the issue, then I think findMode must be hybrid or fast, and not exhaustive.

cderb: This becomes:
use solution if it came from the db (ie wasn't a fallback)
- it's not ok to use fallbacks if find mode is hybrid (the default)
- unless immediate mode is forced
- or the workspace size is less than the maximum required by tunable solvers

This will hit FindCore if sols is empty. Which could lead to the workspace error during tuning.
Also solutions returned by the fdb or fallback solutions could require a workspace greater than the environment limitation.
I would suggest having 2 separate checks. 
- In GetSolutions comparing the workspace in miopenConvSolution_t to that in the invoke_ctx, to validate stored solutions against the env limitation. 
- In the tuning loop within FindCore, skipping proposed solutions with workspace > than the size in invoke_ctx.

xinlipn: > Won't this still hit FindCore if it's not fast, hybrid mode, or if GetSolutions return is empty? I.E if you have tuning enabled, but provide < max workspace, then it will still hit issues.

For this case, Find mode is DYNAMIC_HYBRID, GetSolutions returns one solution, i.e. ConvDirectNaiveConvWrw, need to double check if a direct naive kernel is guaranteed in fallback path, i.e. use default/precomputed kernels, not freshly tuned on local machine. If such a kernel is guaranteed, this logic would work.

> Also, why does GenericSearch get triggered in this case? I thought it should only happen if tuning was enabled, but if this fixes the issue, then I think findMode must be hybrid or fast, and not exhaustive.

GenericSearch is only triggered when findMode is hybrid but not fast.  here's why local tuning is triggered

1. 
https://github.com/ROCm/MIOpen/blob/9f638e9ef615bb302b755b50e623085fbd279721/src/ocl/convolutionocl.cpp#L788

calls 

https://github.com/ROCm/MIOpen/blob/9f638e9ef615bb302b755b50e623085fbd279721/src/ocl/convolutionocl.cpp#L708-L711

fdb_record doesn't exist as shown in the log below:

MIOpen(HIP): Info [Prefetch] File is unreadable: "/MIOpen/buildseg/share/miopen/db/gfx90a68.HIP.3_4_0.ufdb.txt"
[1GB Seg Faultg.txt](https://github.com/user-attachments/files/20223490/1GB.Seg.Faultg.txt)

2. Then fallback path is taken, which returns a valid solution in this case
https://github.com/ROCm/MIOpen/blob/9f638e9ef615bb302b755b50e623085fbd279721/src/ocl/convolutionocl.cpp#L796

We can also see after local tuning is done successfully, there will be no seg fault with original code and small workspace size.
[1GB No Seg Fault.txt](https://github.com/user-attachments/files/20223749/1GB.No.Seg.Fault.txt)

BrianHarrisonAMD: I agree with @cderb.

If we aren't already validating the DB results have a valid amount of workspace, then that needs to be added.
Essentially, we need to filter the cached results down to ones that will work with the workspace provided.
We should double check this by caching the results with full workspace, and then running the same conv with reduced workspace. 

We also need to ensure that when we tune we are filtering out solutions that require more workspace than provided.
I think this needs a fix in GenericSearch, since there is still ways to trigger this, and we should never segfault.

xinlipn: > * In GetSolutions comparing the workspace in miopenConvSolution_t to that in the invoke_ctx, to validate stored solutions against the env limitation.

@cderb  GetSolutions is indeed checking workspace size

https://github.com/ROCm/MIOpen/blob/9fa2113160b80e1f1a6bf5f7b21c59feb73ce116/src/ocl/convolutionocl.cpp#L761-L762

xinlipn:
Fixed Smoke/GPU_ConvBiasActivInferFusionCompileStep_FP32.ConvBiasActivAsm1x1UFloat_testCompile failure due to GetWorkspaceSize() not implemented.  

https://github.com/ROCm/MIOpen/blob/9fa2113160b80e1f1a6bf5f7b21c59feb73ce116/src/include/miopen/fusion/fusion_invoke_params.hpp#L200-L204

Assign max value to keep the logic flow in the above mentioned case per discussion in [PR1875](https://github.com/ROCm/MIOpen/pull/1875)

https://github.com/ROCm/MIOpen/blob/270b10d0fcd7ccaba724570df3e06193a4cbe4bc/src/include/miopen/generic_search.hpp#L339

Local test has passed
[GPU_ConvBiasActivInferFusionCompileStep_FP32 passed.txt](https://github.com/user-attachments/files/20275688/GPU_ConvBiasActivInferFusionCompileStep_FP32.passed.txt)

src/include/miopen/generic_search.hpp
```diff
@@ -346,6 +366,9 @@ void CompileAgent(size_t thread_index,
             std::move(current_config), std::move(current_solution), false);
         comp_queue.push(std::move(tup));
     }
+    // Signal to mark thread completion
+    auto tmp = std::make_tuple<PerformanceConfig, ConvSolution, bool>({}, {}, true);
+    comp_queue.push(std::move(tmp));
```
cderb, line 371: This seems like a good catch, I'm surprised generic search wasn't hanging without this.

src/include/miopen/generic_search.hpp
```diff
@@ -330,12 +331,31 @@ void CompileAgent(size_t thread_index,
...
+            MIOPEN_LOG_I2("CompileAgent: Workspace size not implemented, treating as unlimited.");
+        }
+
+        if(provided_size < current_solution.workspace_sz)
+        {
+            MIOPEN_LOG_I2("Thread: " << thread_index
+                                     << " Skipping config due to workspace requirement "
+                                     << current_solution.workspace_sz << " > " << provided_size);
+            continue;
+        }
```
cderb, line 357: I like this check. Though it seems like generic search is insisting that a workspace size be constant across as solver: https://github.com/ROCm/MIOpen/blob/afb4feaac60af16f783140764eb671fb6e05d0a7/src/include/miopen/generic_search.hpp#L477-L484

xinlipn: Not sure if I got the question right, with provided workspace size set to 1073741824,  both default_solution.workspace_sz and current_solution.workspace_sz are 5452602112. Their values are different from user provided workspace size

MIOpen(HIP): Error [GenericSearch] ********default_solution.workspace_sz: 5452602112
MIOpen(HIP): Error [GenericSearch] ********current_solution.workspace_sz: 5452602112

Below is the complete log (without my changes)
[solution.workspace_sz.txt](https://github.com/user-attachments/files/20349629/solution.workspace_sz.txt)

cderb (approved):
LGTM

BrianHarrisonAMD (approved):

PR #3706, opened 2025-04-30
[HOTFIX] Fix .altmacro compile issues

scerzh:
### Root Cause and Plan of Work

See #3490 
We continue fixing the remaining issues related to the extended .altmacro support in the new compiler.

The plan of work:
* Prepare the unit tests for the affected assembler kernels. This is needed to extract compile options and to test the fix later.
* Extract compile options

```bash
MIOPEN_ENABLE_LOGGING=1 MIOPEN_LOG_LEVEL=6 miopen_gtest --gtest_filter=*testname*
```

* Compile the source with the old compiler

```bash
/opt/rocm/llvm/bin/clang -x assembler ... -c source.s
```

* Save the sections

```bash
/opt/rocm/llvm/bin/llvm-objdump --full-contents source.o > compiler_old_src_old.txt
```

* Fix the source
* Run the tests and ensure they pass
* Compile the source with the old compiler with the same options (cmd is the same)

```bash
/opt/rocm/llvm/bin/clang -x assembler ... -c source.s
```

* Save the sections

```bash
/opt/rocm/llvm/bin/llvm-objdump --full-contents source.o > compiler_old_src_new.txt
```

* Compile the source with the new compiler with the same options (cmd is the same)

```bash
/opt/rocm/llvm/bin/clang -x assembler ... -c source.s
```

* Save the sections

```bash
/opt/rocm/llvm/bin/llvm-objdump --full-contents source.o > compiler_new_src_new.txt
```

* Compare the compile output

```bash
sha512sum compiler_{old,new}_src_???.txt
```

The compile commands will be listed below for the fixed sources.

### Command Lines for Compiling

conv1x1u_stride2.s

```bash
[...]

PR #3716, opened 2025-05-05
[BUG][TESTS] Add unit tests for conv asm 1x1u bwdwrw3x3 3x3u 1x1uv2

scerzh:
Adding unit tests for the solvers affected by #3706


src/solver/conv/conv_asm_1x1u.cpp
```diff
@@ -550,7 +550,7 @@ bool ConvAsm1x1U::IsApplicable(const ExecutionContext& ctx, const ProblemDescrip
         return false;
 
     const std::string name = ctx.GetStream().GetDeviceName();
-    if(name.find("gfx9") == std::string::npos)
+    if(!(StartsWith(name, "gfx90")))
```
averinevg, line 553: Should `gfx942` be added here or are the necessary instructions missing?

src/solver/conv/conv_asm_1x1u_stride2.cpp
```diff
@@ -511,7 +511,7 @@ bool ConvAsm1x1UV2::IsApplicable(const ExecutionContext& ctx,
         return false;
 
     const std::string name = ctx.GetStream().GetDeviceName();
-    if(name.find("gfx8") == std::string::npos && name.find("gfx9") == std::string::npos)
+    if(!(StartsWith(name, "gfx8") || StartsWith(name, "gfx90")))
```
averinevg, line 514: It might be a good idea to clean up and remove all the code for `gfx8`. Not in this PR of course.

test/gtest/unit_conv_solver_ConvAsm1x1U.cpp
```diff
@@ -0,0 +1,185 @@
...
+        auto p =
+            miopen::unit_tests::UnitTestConvSolverParams(Gpu::gfx900 | Gpu::gfx906 | Gpu::gfx908);
+        p.Tunable(5);
+        p.SetConvAttrFp16Alt(1);
+        p.CheckXnackDisabled();
+        p.EnableDeprecatedSolvers();
+        return p;
+    }();
+    return params;
+}
```
averinevg: Note: `AttrFp16Alt` works only for `gfx90a`.

test/gtest/unit_conv_solver_ConvAsm1x1U.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+INSTANTIATE_TEST_SUITE_P(Smoke,
+                         CPU_UnitTestConvSolverAsm1x1UDevApplicabilityFwd_FP32,
+                         testing::Combine(testing::Values(GetTestParamsFp32()),
+                                          testing::Values(GetConvTestCases(miopenFloat)[0])));
+
+INSTANTIATE_TEST_SUITE_P(Smoke,
+                         CPU_UnitTestConvSolverAsm1x1UDevApplicabilityBwd_FP32,
+                         testing::Combine(testing::Values(GetTestParamsFp32()),
+                                          testing::Values(GetConvTestCases(miopenFloat)[0])));
```
averinevg: There is no need to run this test for all data types and directions. Although it is a short test and has almost no impact on the overall test time, it doesn't improve test coverage either.

test/gtest/unit_conv_solver_ConvAsm1x1U.cpp
scerzh: Does it mean it should be ignored by any solver on any device other than `gfx90a`?

test/gtest/unit_conv_solver_ConvAsm1x1U.cpp
scerzh: Backward and forward path have different default values for `Fp16Alt` flag and the result of `IsApplicable()` depends on it. But if the flag should be ignored on any device except `gfx90a`, 2/3 of these tests are useless and it's only needed to confirm the solver isn't applicable for `gfx90a` with the flag set.

https://github.com/ROCm/MIOpen/blob/5b9d08fb31e1c1ccd1a784b3c63499bee7c589f7/src/include/miopen/convolution.hpp#L83-L85

scerzh: I've removed unnecessary tests.

test/gtest/unit_conv_solver_ConvAsm1x1U.cpp
scerzh: I've removed unnecessary tests.

BrianHarrisonAMD (approved):
LGTM assuming tests pass.

PR #3712, opened 2025-05-01
Updated docker for dev env to ROCm 6.4 release

BradPepersAMD:
Updated docker to pull the latest ROCm 6.4 release files so that our CI and build env are closer to mainline builds.

PR #3718, opened 2025-05-05
Changes needed to fix build issues with latest environment change

BrianHarrisonAMD:
- Includes #3712  to update the environment
- Includes other warning suppression changes Brad added
- Includes Workaround that disables warnings as errors for kernel compilation (fixes DEV_BUILD testing errors for now)
- Reverted the following commits to get it back to mainline state for intrinsic related changes:
    - #3694 
    - #3682 
    - #3689 
    - #3670


randyspauldingamd:
I've started a build on gfx942. All relevant tests were skipped, as expected. Started the entire test suite.

BradPepersAMD (approved):
I'm good with this since it undoes other things which looks right and is relatively safe and the rest of it is also just basic adjustments to improve the env and adjust to the new compiler version.

PR #3724, opened 2025-05-08
Fix static tests failing due to missing rocBLAS linkage when MIOpen is built as a static lib

BrianHarrisonAMD:
Changes:

- Updated the CTests to have rocBLAS & hipBLASLt linkage if they are enabled.
- Updated the GTests to consistently check the build flags before linking to these additional targets

Reason for change:

- When MIOpen is built as a static lib, the rocBLAS linkage isn't automatically being included for the testing executables. Adding the rocBLAS linkage resolves this issue.


bghimireamd (approved):
LGTM

iq136boy (approved):

PR #3725, opened 2025-05-08
Clean db files to pass db_sync

cderb:
Removed inapplicable entries from find db. Removed stale entries from perf db.


src/db_record.cpp
```diff
@@ -182,7 +183,11 @@ void DbRecord::WriteIdsAndValues(std::ostream& stream) const
 
     const auto pairsJoiner = [](const std::string& sum,
                                 const std::pair<std::string, std::string>& pair) {
-        const auto pair_str = pair.first + ':' + pair.second;
+        std::string first(pair.first);
+        std::string second(pair.second);
+        boost::trim(first);
+        boost::trim(second);
+        const auto pair_str = first + ':' + second;
```
BrianHarrisonAMD: Why was this needed? 

If the keys & values from the map had extra whitespace, then this will change the keys when they get written.
I think if we add this, then the Get/Erase also need to trim.

cderb: Right. The Remove call to the miopen::RamDb was inserting newlines to the fdb file inside of value lists prior to this change. This made some entries unreadable by dissociating the values with their key. There might be a better way to fix this.

src/db_record.cpp
```diff
@@ -138,6 +139,7 @@ bool DbRecord::ParseContents(std::istream& contents)
 
         auto id     = id_and_values.substr(0, id_size);
         auto values = id_and_values.substr(id_size + 1);
+        boost::trim(values);
```
cderb, line 142: @BrianHarrisonAMD moved trim to the data import. Newline character was included from txt file.

BrianHarrisonAMD (approved):

BrianHarrisonAMD:
Looks good assuming we can get a clean CI run!

iq136boy (approved):

JonathanLichtnerAMD (approved):

JonathanLichtnerAMD:
This passed CI after pulling the BN fixes, so I think this can be merged now.

PR #3727, opened 2025-05-09
BN: fix bug from sanitizer

EnricoDeg:
Avoid unnecessary calculations when vectorsize is not supported for a given problem.

This fixes the sanitizer issues in debug mode.


shurale-nkn (approved):
LGTM

iq136boy (approved):

Issue #3708, opened 2025-05-01
kernel_tuning_net gtest is fragile due to violating blackbox


randyspauldingamd:
Test failure observed after updating CK:
```
[2025-05-01T10:36:12.397Z] /home/jenkins/workspace/MLLIBS_MIOpen_PR-3707/test/gtest/kernel_tuning_net.cpp:188: Failure
[2025-05-01T10:36:12.397Z] Expected equality of these values:
[2025-05-01T10:36:12.397Z]   perf_config.ToString()
[2025-05-01T10:36:12.397Z]     Which is: "DeviceGroupedConvBwdDataMultipleD_Xdl_CShuffle_v1<128, 128, 32, 32, 8, 8, Default, 32, 32, 2, 1, 8, 8, 1, 1, BlkGemmPipelineScheduler: Intrawave, BlkGemmPipelineVersion: v1>"
[2025-05-01T10:36:12.397Z]   test_case.expected_config
[2025-05-01T10:36:12.397Z]     Which is: "DeviceGroupedConvBwdDataMultipleD_Xdl_CShuffle_v1<128, 128, 32, 32, 8, 8, Default, 32, 32, 2, 1, 8, 8, 1, 1>"
```

@msaudulhassan reports that this is expected behavior when CK and the ktn models get out of sync. This makes the test inherently fragile, and it should be rewritten to use mocks and canned "gold" model files to achieve blackbox. In other words, a proper test shall test only the code itself, and not involve the actual version of CK nor model files.

The test shall be disabled for now. The priority and timeline are uncertain at this time, so if anyone can suggest an acceptable workaround, please do!

ammallya:
Imported to ROCm/rocm-libraries

PR #3733, opened 2025-05-14
Rework KTN gtest

cderb:
Test no longer compares to hard coded string.
Test applicability of performance config returned by model.
Test solver can be used for shape.
Test invocation of solution.
#3708


test/gtest/kernel_tuning_net.cpp
```diff
@@ -181,11 +160,76 @@ class KernelTuningNetTest : public ::testing::TestWithParam<KernelTuningNetTestC
...
-        Solver perf_config;
+        auto data_type = test_case.data_type;
+        auto in_tensor =
+            GPUMem{0, input_tensor_desc.GetNumBytes() / sizeof(data_type), sizeof(data_type)};
+        auto wt_tensor =
+            GPUMem{0, weights_tensor_desc.GetNumBytes() / sizeof(data_type), sizeof(data_type)};
+        auto out_tensor =
+            GPUMem{0, output_desc.GetNumBytes() / sizeof(data_type), sizeof(data_type)};
+        auto workSpaceSize = conv_desc.GetWorkSpaceSize(ctx, problem);
+        auto workSpace     = GPUMem{0, workSpaceSize / sizeof(data_type), sizeof(data_type)};
```
reidkwja: Because of splitting bytes into elements, some solvers require workspace to be aligned like ConvAsmImplicitGemmGTCDynamicWrwXdlopsNHWC (256 B padding) - or have a multiple of some tile size. 

So it might be a good idea to assert it meets alignment:

`assert((workSpaceSize % required_alignment) == 0);`

test/gtest/kernel_tuning_net.cpp
```diff
@@ -181,11 +160,76 @@ class KernelTuningNetTest : public ::testing::TestWithParam<KernelTuningNetTestC
                                                               conv_desc,
                                                               test_case.direction);
 
-        Solver perf_config;
+        auto data_type = test_case.data_type;
+        auto in_tensor =
```
reidkwja: Becasue data_type is enum (not C++ type), when dividing the total byte_size, it will always be the size of that enum (which is 4 bytes) rather than the size of the underlying data elements intended to be allocated - which could over/under allocate GPU buffer. Is there a better way instead, like an actual element size or so?

BrianHarrisonAMD: There is a utility for this I believe:

https://github.com/ROCm/MIOpen/blob/7e387ac49ab79d8e78aed3a54ec2362745b1183c/src/include/miopen/datatype.hpp#L84-L110

BrianHarrisonAMD:
Looks like build is failing on windows.

```
lld-link: error: undefined symbol: bool __cdecl miopen::conv::IsAlgorithmDisabled(enum miopenConvAlgorithm_t)
--
  | 30811 | 01:18:24 AM | >>> referenced by C:\w\test\gtest\kernel_tuning_net.cpp:216
  | 30812 | 01:18:24 AM | >>>               CMakeFiles/test_kernel_tuning_net.dir/kernel_tuning_net.cpp.obj:(protected: void __cdecl KernelTuningNetTest<struct miopen::solver::conv::PerformanceConfigConvAsm1x1U>::TestParameterPredictionModel(class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>))
  | 30813 | 01:18:24 AM | >>> referenced by C:\w\test\gtest\kernel_tuning_net.cpp:216
more_vert | 30814 | 01:18:24 AM | >>>               CMakeFiles/test_kernel_tuning_net.dir/kernel_tuning_net.cpp.obj:(protected: void __cdecl KernelTuningNetTest<struct miopen::solver::conv::PerformanceConfigHipImplicitGemmGroupFwdXdlops>::TestParameterPredictionModel(class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>))
  | 30815 | 01:18:24 AM | >>> referenced by C:\w\test\gtest\kernel_tuning_net.cpp:216
  | 30816 | 01:18:24 AM | >>>               CMakeFiles/test_kernel_tuning_net.dir/kernel_tuning_net.cpp.obj:(protected: void __cdecl KernelTuningNetTest<struct miopen::solver::conv::PerformanceConfigHipImplicitGemmGroupBwdXdlops>::TestParameterPredictionModel(class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>))
  | 30817 | 01:18:24 AM | >>> referenced 1 more times
  | 30818 | 01:18:24 AM |  
  | 30819 | 01:18:24 AM | lld-link: error: undefined symbol: bool __cdecl miopen::conv::IsEnoughWorkspace(class std::basic_string_view<char, struct std::char_traits<char>>, struct miopen::solver::Id const &, unsigned __int64, struct miopen::AnyInvokeParams const *)
  | 30820 | 01:18:24 AM | >>> referenced by C:\w\test\gtest\kernel_tuning_net.cpp:220
  | 30821 | 01:18:24 AM | >>>               CMakeFiles/test_kernel_tuning_net.dir/kernel_tuning_net.cpp.obj:(protected: void __cdecl KernelTuningNetTest<struct miopen::solver::conv::PerformanceConfigConvAsm1x1U>::TestParameterPredictionModel(class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>))
  | 30822 | 01:18:24 AM | >>> referenced by C:\w\test\gtest\kernel_tuning_net.cpp:220
  | 30823 | 01:18:24 AM | >>>               CMakeFiles/test_kernel_tuning_net.dir/kernel_tuning_net.cpp.obj:(protected: void __cdecl KernelTuningNetTest<struct miopen::solver::conv::PerformanceConfigHipImplicitGemmGroupFwdXdlops>::TestParameterPredictionModel(class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>))
  | 30824 | 01:18:24 AM | >>> referenced by C:\w\test\gtest\kernel_tuning_net.cpp:220
  | 30825 | 01:18:24 AM | >>>               CMakeFiles/test_kernel_tuning_net.dir/kernel_tuning_net.cpp.obj:(protected: void __cdecl KernelTuningNetTest<struct miopen::solver::conv::PerformanceConfigHipImplicitGemmGroupBwdXdlops>::TestParameterPredictionModel(class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>))
  | 30826 | 01:18:24 AM | >>> referenced 1 more times
  | 30827 | 01:18:24 AM | clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

Probably need to add MIOPEN_INTERNALS_EXPORT to some of the functions that are used now in the tests.

test/gtest/kernel_tuning_net.cpp
cderb: This has been resolved I believe.

cderb:
> Looks like build is failing on windows.
> 
> Probably need to add MIOPEN_INTERNALS_EXPORT to some of the functions that are used now in the tests.

I think this has been corrected.

BrianHarrisonAMD (approved):
Edit:

Might be worth getting @msaudulhassan to look at this.
I think the original intent of these tests were to ensure the trained model was matching the expected result once integrated.

We may want to have both types of tests.
Some specific points to ensure the model integration is matching expected predictions.
And another set that ensures a broad set of shapes is giving correct predictions.

cderb:
I'm not a fan of having hard coded expected results in these tests. There would be no way to guarantee the results between commits.
In practice there has been no expectation for the results of these tests, they've simply been updated with whatever is coming out of the model.
I think the valuable portion of this is confirming that the model is producing a viable result for a solver which is intended to be applicable for a set of shapes.

BrianHarrisonAMD (approved):
LGTM

PR #3739, opened 2025-05-21
Remove CK's gemm and reduction libraries from MIOpen

JonathanLichtnerAMD:
This PR removes CK's gemm and reduction libraries from MIOpen.  This change is a step towards moving to CK header-only libraries, and should also allow MIOpen's CK build to be sped up, since it will no longer need to build device_gemm_operations and device_reduction_operations for MIOpen.

This was tested via a subset of the usual CI on the branch, and was also run against _make check_ with the specific CK libraries forcefully removed from the docker container on MI200.  (I'll probably also want to do this on MI100 and MI300.)

Removal of CK's batchnorm will be handled in a separate PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,7 +328,7 @@ add_compile_definitions($<$<COMPILE_LANGUAGE:CXX>:HIP_COMPILER_FLAGS=${HIP_COMPI
 # HIP
 if( MIOPEN_BACKEND STREQUAL "HIP" OR MIOPEN_BACKEND STREQUAL "HIPOC" OR MIOPEN_BACKEND STREQUAL "HIPNOGPU")
     if(MIOPEN_USE_COMPOSABLEKERNEL)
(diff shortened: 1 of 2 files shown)


BrianHarrisonAMD (approved):

BrianHarrisonAMD:
Looks good assuming CI finds no issues.

JonathanLichtnerAMD:
Note that this branch has passed CI on one of Adam's branches http://micimaster.amd.com/job/MLLIBS/job/MIOpen/job/AD%252FReduced_CK_test/

Although we are still waiting for gx94x results...

JonathanLichtnerAMD:
Adam's branch hit a weird timeout on "Fp16 Hip All Install gfx94X" (see http://micimaster.amd.com/blue/organizations/jenkins/MLLIBS%2FMIOpen/detail/AD%2FReduced_CK_test/5/pipeline), but everything else passed so we should be safe to merge this.

PR #3742, opened 2025-05-22
Remove CK's batchnorm usage from MIOpen

JonathanLichtnerAMD:
This PR removes CK's batchnorm from MIOpen.  It wasn't used, and removing it will simplify migrating away from static CK libraries.  If we decide this is needed in the future, then we can re-add it.


BrianHarrisonAMD (approved):
LGTM assuming tests pass etc.

src/fin/fin_interface.cpp
```diff
@@ -356,9 +356,6 @@ AnySolver<miopen::ExecutionContext, miopen::batchnorm::ProblemDescription>::AnyS
     {
     case 113: SetObject<miopen::solver::batchnorm::BnFwdTrainingSpatial>(); break;
     case 117: SetObject<miopen::solver::batchnorm::BnBwdTrainingSpatial>(); break;
-    case 142: SetObject<miopen::solver::batchnorm::BnCKFwdInference>(); break;
-    case 143: SetObject<miopen::solver::batchnorm::BnCKBwdBackward>(); break;
-    case 144: SetObject<miopen::solver::batchnorm::BnCKFwdTraining>(); break;
```
BrianHarrisonAMD, line 361: I am not as familiar with the fin interface for solvers.
I think this is correct / fine, but probably worth double checking with @cderb & @averinevg to make sure.

JonathanLichtnerAMD: I'll add them to the review just to be safe

averinevg (approved):
LGTM

PR #3744, opened 2025-05-26
[HOTFIX] Fix Winograd Rage group convolution

scerzh:
Fix dispatch for group convolutions in Winograd Rage solver

### DIFF
--- a/src/solver/conv/conv_wino_rage_RxS.cpp
+++ b/src/solver/conv/conv_wino_rage_RxS.cpp
@@ -210,7 +210,7 @@ ConvWinoRageRxSCommon<Winodata, Winofilter>::GetSolution(const ExecutionContext&
     kernelInfo.l_wk.push_back(1);
     kernelInfo.l_wk.push_back(1);
 
-    kernelInfo.g_wk.push_back(wgSize * nGroups);
+    kernelInfo.g_wk.push_back(wgSize * nGroups * args.G);
     kernelInfo.g_wk.push_back(1);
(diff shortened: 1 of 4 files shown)


src/solver/conv/conv_wino_rage_RxS.cpp
```diff
@@ -210,7 +210,7 @@ ConvWinoRageRxSCommon<Winodata, Winofilter>::GetSolution(const ExecutionContext&
     kernelInfo.l_wk.push_back(1);
     kernelInfo.l_wk.push_back(1);
 
-    kernelInfo.g_wk.push_back(wgSize * nGroups);
+    kernelInfo.g_wk.push_back(wgSize * nGroups * args.G);
```
BrianHarrisonAMD, line 213: conv_wino_fury_RxS.cpp has the same calculation for g_wk.
Are they different, or do both need to be updated?

conv_winoRxS.cpp seems to match this though.

scerzh: Current version of Fury doesn't support grouped convolutions and doesn't need to be updated together with Rage.

We are planning to upgrade Fury to 4.6 (which supports grouped convolutions) and then we will make them match.

BrianHarrisonAMD (approved):
LGTM assuming tests are passing.

Slimakanzer (approved):
LGTM

averinevg (approved):
LGTM

PR #3751, opened 2025-05-27
Switch to using CK header-only for layernorm

JonathanLichtnerAMD:
This PR migrates MIOpen to use the header-only CK for layernorm.  The motivation for doing this is two-fold:

1. Currently, MIOpen gets its CK kernels from static libraries that CK builds.  In order to speed up build times (with a focus of building MIOpen in TheRock), we have been reducing the number of CK libraries that MIOpen depends upon.  Currently, MIOpen requires the _convolution_ and _other_ libraries, but MIOpen only requires the _other_ library for CK's layernorm kernels.  By removing this last dependency, MIOpen will only need to build the _convolution_ library which should further speed up the CK build time for MIOpen.
2. We are considering migrating MIOpen to use CK header-only for all of its convolutions (see [CK Dependency Strategy for MIOpen in TheRock ](https://amdcloud.sharepoint.com/:w:/r/sites/ComposableKernels/_layouts/15/Doc.aspx?sourcedoc=%7BF91F46C8-7BB5-4845-8430-C9F1C69844C1%7D&file=Strategy%20for%20CK%20in%20MIOpen%20in%20TheRock.docx&wdLOR=c269A5722-4833-4281-920E-E4D3F230D16F&nav=eyJjIjo0MDg5ODA1ODF9&action=default&mobileredirect=true) for details).  Since MIOpen's use of CK's layernorm is small, this can provide a quick and easy prototype of how this migration to header-only could be done.

Note that CK provides examples of header-only and the static library usage in their repo.  For example, for layernorm the example https://github.com/ROCm/composable_kernel/tree/develop/example/27_layernorm2d_fwd provides an example of header-only usage, whereas https://github.com/ROCm/composable_kernel/tree/develop/client_example/05_layernorm provides an example that uses the static CK library, similar to how MIOpen would use it.

This PR changes MIOpen to use CK header-only, but the changes to the MIOpen solvers were minimized by moving the GetInstances() method into MIOpen, so logically MIOpen now gathers the kernels in a manner similar to how CK does it.  (Note that CK still needs to do this for their client examples.)

Anyway, this has passed CI and should now be ready for review but there are still a few items to do:
 - [x] run this on MI100 CI
 - [x] run this on MI300 CI
 - [x] create a CK branch that only builds the _convolution_library for MIOpen to see how the build time changes (*)
 - [x] this has been tested using MIOpenDriver and forcing the CK layernorm kernels to be used, but it might be good to try more shapes

(*) Note that adding the new CK layernorm kernels only increased the MIOpen build time by a few seconds on MI200 with -j 32 (3m54s w/ the changes vs. 3m52s w/o the changes).


src/kernels/ck_header_only/layernorm/fwd/device_layernorm2d_fwd_f16_instance.cpp
```diff
@@ -0,0 +1,27 @@
...
+void add_device_normalization_fwd_rank_2_1_f16_instances(
+    std::vector<std::unique_ptr<DeviceNormalizationFwd<F16, F16, F16, F16, F16, Pass, 2, 1>>>&
+        instances)
+{
+    add_device_operation_instances(instances,
+                                   device_normalization_f16_generic_instance<Pass, 2, 1>{});
+    add_device_operation_instances(instances, device_normalization_f16_instances<Pass, 2, 1>{});
+    add_device_operation_instances(instances,
+                                   device_normalization_splitk_f16_instances<Pass, 2, 1>{});
+}
```
BrianHarrisonAMD: Edit: Hmm maybe not since you need to add all of them, and there might be multiple solvers that need a piece.

BrianHarrisonAMD:
Where is ::GetInstances() declared? 

I thought we would need our own version of something like this:
https://github.com/ROCm/composable_kernel/blob/132bd5b874537f313f871b61398e1999456981ce/library/include/ck/library/tensor_operation_instance/gpu/normalization_fwd.hpp#L73

JonathanLichtnerAMD:
> Where is ::GetInstances() declared?
> 
> I thought we would need our own version of something like this: https://github.com/ROCm/composable_kernel/blob/132bd5b874537f313f871b61398e1999456981ce/library/include/ck/library/tensor_operation_instance/gpu/normalization_fwd.hpp#L73

@BrianHarrisonAMD It's currently using the one you linked, library/include/ck/library/tensor_operation_instance/gpu/normalization_fwd.hpp.  Calling GetInstance now will invoke the functions that were added under src/kernels/ck_header_only/layernorm/fwd rather than the ones from the CK device_other_operations library.

BrianHarrisonAMD:
> > Where is ::GetInstances() declared?
> > I thought we would need our own version of something like this: https://github.com/ROCm/composable_kernel/blob/132bd5b874537f313f871b61398e1999456981ce/library/include/ck/library/tensor_operation_instance/gpu/normalization_fwd.hpp#L73
> 
> @BrianHarrisonAMD It's currently using the one you linked, library/include/ck/library/tensor_operation_instance/gpu/normalization_fwd.hpp. Calling GetInstance now will invoke the functions that were added under src/kernels/ck_header_only/layernorm/fwd rather than the ones from the CK device_other_operations library.

I sort of assumed that would be the answer, but that is very odd hehe.

JonathanLichtnerAMD:
> > > Where is ::GetInstances() declared?
> > > I thought we would need our own version of something like this: https://github.com/ROCm/composable_kernel/blob/132bd5b874537f313f871b61398e1999456981ce/library/include/ck/library/tensor_operation_instance/gpu/normalization_fwd.hpp#L73
> > 
> > 
> > @BrianHarrisonAMD It's currently using the one you linked, library/include/ck/library/tensor_operation_instance/gpu/normalization_fwd.hpp. Calling GetInstance now will invoke the functions that were added under src/kernels/ck_header_only/layernorm/fwd rather than the ones from the CK device_other_operations library.
> 
> I sort of assumed that would be the answer, but that is very odd hehe.

@BrianHarrisonAMD Yes, I created a GetInstances() method inside MIOpen, which allows more fine-grained control over what kernels MIOpen selects (this did end being different than the CK version, since CK also had some groupnorm kernels that MIOpen didn't make use of).

JonathanLichtnerAMD:
Note that I hit some issues building CK inline as part of the MIOpen build, but those went away when I did a git pull....  Anyway, building the CK other/layernorm changes in CK and in MIOpen had basically the same overall time.  With 64 threads on an MI200:
- MIOpen + layernorm + CK-conv-only:  51m6s (real), 2104m2s (user)
- MIOpen + CK-conv-and-layernorm:      51m4s (real), 2169m4s

BradPepersAMD:
Are there any outstanding issues or concerns holding us back from merging this change?

adickin-amd (approved):

BrianHarrisonAMD (approved):
LGTM

PR #3756, opened 2025-05-28
Disable ZeroOutBuffer for CK Grouped Conv Fwd

bartekxk:


randyspauldingamd (approved):
LGTM

BrianHarrisonAMD:
I think we will want to hold off on merging this until we have all the clears only happening under the correct conditions, and we can do extensive testing to ensure we don't encounter additional NaN or accuracy issues.

I believe the plan is to remove all the extra clears from MIOpen, and allow CK to clear when needed instead.

BrianHarrisonAMD:
I think this can be closed and is covered by #3783 now.

Issue #3757, opened 2025-05-28
Refactor tests for reduction solvers


scerzh:
Current GTest implementation was derived from CTest one. It doesn't affect the CI because the tests never fail and therefore useless for the CI. The implementation only prints to std::cerr the detected issues and affects compiler's staging tests which fail upon such std::cerr output.

The current issues printed to std::err demonstrate compiler errors for reduction kernels for BFP16 datatype. CTest implementation contained only F32 tests, and during move to GTest, test cases were added for FP16, BFP16 and I8 datatypes. These new test cases are going to be disabled until this issue is resolved.

* Make tests fail CI on errors
* Investigate BFP16 issue

ammallya:
Imported to ROCm/rocm-libraries

PR #3758, opened 2025-05-28
[GTEST][WA] Disable failing reduction tests

scerzh:
Disable tests that fail staging for compiler, see #3757


averinevg (approved):
LGTM except that I'm not sure if the data type `BHalf` actually exists.

scerzh:
@averinevg 
I suggest the intention was to add tests for `miopenBFloat16`.
Nevertheless, there are also numeric overflows for `miopenHalf` and these tests never fail making them useless for the CI.

averinevg:
BTW I tested reduction with fp32/fp16 today and it works. It makes sence to remove the rest of the tests, since reduction does not support these data types.

scerzh:
> BTW I tested reduction with fp32/fp16 today and it works. It makes sence to remove the rest of the tests, since reduction does not support these data types.

I am confused. These tests showed the following results when I tried them:
* FP32 test passes
* FP16 test breaks on numeric overflows
* BFP16 test breaks on compiling
* I8 test passes

Only FP32 is actually supported by the solver, correct? If so, we need to fix IsApplicable().

PR #3761, opened 2025-05-28
Refactor Jenkinsfile to run tests differently

adickin-amd:
- Smoke steps have been removed.  Full tests make many of the smoke tests redundant and we need to speed up the CI.
- Smoke tests which are not redundant have been moved into the Nightly Tests
- combined package and tidy checks
-


TristanZerrAMD (commented):
Nice work Adam!  

High level question.  In the codebase, I see a lot of explicit usage of the word "Smoke" in how various test are structured (500+ instances).  Do these need to be refactored at all to stay consistent with CI?  For example, there's lots of code like this:  

*// Smoke tests
INSTANTIATE_TEST_SUITE_P(Smoke,
                         GPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP16,
                         testing::Combine(testing::Values(GetSmokeTestParams(miopenHalf)),
                                          testing::Values(miopenConvolutionAlgoImplicitGEMM),
                                          testing::ValuesIn(GetConvSmokeTestCases(miopenHalf))));*

adickin-amd:
> Nice work Adam!
> 
> High level question. In the codebase, I see a lot of explicit usage of the word "Smoke" in how various test are structured (500+ instances). Do these need to be refactored at all to stay consistent with CI? For example, there's lots of code like this:
> 
> _// Smoke tests INSTANTIATE_TEST_SUITE_P(Smoke, GPU_UnitTestConvSolverAsmImplicitGemmGTCDynamicBwdXdlopsNHWCBwd_FP16, testing::Combine(testing::Values(GetSmokeTestParams(miopenHalf)), testing::Values(miopenConvolutionAlgoImplicitGEMM), testing::ValuesIn(GetConvSmokeTestCases(miopenHalf))));_

@TristanZerrAMD I think the answer depends on what we want to do with Smoke going forward.  As a developer you could still use the smoke gtest filter to run a set of tests locally to confirm your changes are good without running the full test suite.  As far as I know Ci was running the smoke tests first since they took less time, this meant that failures would be found sooner in the jenkins pipeline which would use less CI time.  Unfortunately the smoke tests are at least 2+ hours and the Smoke tests get run when we run the full set of tests.  In this PR I have changed it to just run the full tests so we don't chew up the CI as much when all things are green. An alternative is to exclude the smoke tests from the full runs but that will require us to build MIOpen again which just takes up more time.

TristanZerrAMD (approved):

reidkwja (approved):
lgtm - one thing to double check (just to be safe) is 'needs_reboot:false' is appropriately placed for correct stages to potentially avoid endless stall infra wise.

adickin-amd:
> lgtm - one thing to double check (just to be safe) is 'needs_reboot:false' is appropriately placed for correct stages to potentially avoid endless stall infra wise.

All jobs should be the same, all thats changed is the ordering.

PR #3763, opened 2025-05-28
Update tensor check logic 

Cemberk:
There is an issue seen by upstream Transformers unit tests 
There are customer models though not common that use strides with dimensions 1 apart here is some example reproducer code from upstream investigations. This happens on 4D and 5D tensors. This PR updates the tensor constraints to allow this customer condition.

```
import torch
from torch.nn import functional as F
if __name__ == "__main__":
    fix_layout = False
    input = torch.randn(13, 4, 1, 15, 15).to(dtype=torch.float32, device="cuda:0")
    input = torch.as_strided(input, size=(13, 4, 1, 15, 15), stride=(900, 225, 900, 15, 1))
    print(f"{input.shape = }, {input.device = }, {input.dtype = }, {input.stride() = }")
    if fix_layout:
        input = input.reshape(input.shape)
        print(f"{input.shape = }, {input.device = }, {input.dtype = }, {input.stride() = } {input.is_contiguous() = }")
    F.batch_norm(
        input=input,
        running_mean=torch.zeros(4).to("cuda:0"),
        running_var=torch.ones(4).to("cuda:0"),
        weight=torch.ones(4).to("cuda:0"),
        bias=torch.zeros(4).to("cuda:0"),
        training=False,
        momentum=0.1,
        eps=1e-5,
    )
```

Before: Any stride violation → reject tensor
After:  Stride violation with size-1 dimension → allow (safe)
        Stride violation with normal dimensions → reject (unsafe)

Example that now works:
- Shape: [13, 4, 1, 15, 15] 
- Stride: [900, 225, 900, 15, 1]
- layout_strides for "NCHW": [900, 225, 900, 15]
- Violation: 225 < 900 at positions 1,2 (C,H dimensions)
- H dimension (pos 2) has size 1 → SAFE → Allow
- Result: Tensor now accepted as "NCHW" layout

### DIFF
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -684,32 +684,49 @@ bool TensorDescriptor::IsPossibleLayout(const std::string& storage_layout,
     std::vector<std::size_t> layout_strides(base_layout.size());
     std::transform(base_layout.cbegin(), base_layout.cend(), layout_strides.begin(), op);
 
+    // Check monotonic decreasing with conservative PyTorch compatibility fix
     for(unsigned i = 0; i < (layout_strides.size() - 1); i++)
     {
         if(layout_strides[i] < layout_strides[i + 1])
         {
-            // Find which dimensions are involved in this stride violation
-            auto char_at_i = base_layout[i];
-            auto char_at_i_plus_1 = base_layout[i + 1];
+            // Conservative fix: Only allow stride violations when:
+            // 1. This is a 5D tensor 
+            // 2. The violating dimension has size 1
+            // 3. The stride values are actually shared (same value appears elsewhere)
+            bool allow_violation = false;
             
-            auto dim_pos_i = storage_layout.find(char_at_i);
-            auto dim_pos_i_plus_1 = storage_layout.find(char_at_i_plus_1);
-            
-            if(dim_pos_i == std::string::npos || dim_pos_i_plus_1 == std::string::npos)
-                MIOPEN_THROW(miopenStatusInternalError, "wrong layout format");
-            
-            // This is memory-safe because size-1 dimensions don't create aliasing
-            if(lens[dim_pos_i] == 1 || lens[dim_pos_i_plus_1] == 1)
+            if(lens.size() == 5)
             {
-                continue; // Skip this violation - it's safe
+                auto char_at_i = base_layout[i];
+                auto char_at_i_plus_1 = base_layout[i + 1];
+                
+                auto dim_pos_i = storage_layout.find(char_at_i);
+                auto dim_pos_i_plus_1 = storage_layout.find(char_at_i_plus_1);
+                
+                if(dim_pos_i != std::string::npos && dim_pos_i_plus_1 != std::string::npos)
+                {
+                    // Only allow if the larger stride dimension has size 1
+                    if(lens[dim_pos_i_plus_1] == 1)
+                    {
+                        // Verify this is genuine stride sharing (not arbitrary values)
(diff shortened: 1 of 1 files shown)


averinevg:
Hi @Cemberk, the PR is failing in CI at the clang-format check stage. You need to fix the formatting; here's the link to the wiki: [How-to-format-code](https://github.com/ROCm/MIOpen/wiki/How-to-format-code).

averinevg:
@Cemberk, I think your description mixes 4D and 5D tensors - correct me if I'm wrong.

> -  Shape: [13, 4, 1, 15, 15]
> -  Stride: [900, 225, 900, 15, 1]

This is a 5D tensor.
> -  layout_strides for "NCHW": [900, 225, 900, 15]

This is a 4D tensor.
> -  H dimension (pos 2) has size 1 → SAFE → Allow

D dimension?
> -  Result: Tensor now accepted as "NCHW" layout

"NCDHW"?

Also, it is not clear to me why they set such a strange stride. Could you please comment on that?

src/tensor.cpp
```diff
@@ -684,13 +684,44 @@ bool TensorDescriptor::IsPossibleLayout(const std::string& storage_layout,
...
-    // Check monotonic decreasing
+    // Check monotonic decreasing with conservative PyTorch compatibility fix
     for(unsigned i = 0; i < (layout_strides.size() - 1); i++)
     {
         if(layout_strides[i] < layout_strides[i + 1])
-            return false;
+        {
+            bool allow_violation = false;
+    
+            if(lens.size() == 5)
```
averinevg, line 694: Should this work only for 5D tensor?

src/tensor.cpp
```diff
@@ -684,13 +684,44 @@ bool TensorDescriptor::IsPossibleLayout(const std::string& storage_layout,
...
+            {
+                auto char_at_i = base_layout[i];
+                auto char_at_i_plus_1 = base_layout[i + 1];
+    
+                auto dim_pos_i = storage_layout.find(char_at_i);
+                auto dim_pos_i_plus_1 = storage_layout.find(char_at_i_plus_1);
+    
+                if(dim_pos_i != std::string::npos && dim_pos_i_plus_1 != std::string::npos)
```
averinevg: Already checked at line 679

BrianHarrisonAMD:
Okay I looked at the reproducer, and it looks like they are providing incorrect strides, but since the dim is 1 we are basically saying it doesn't matter.

I can understand allowing that, but I think we should have a revised check that skips verifying strides for dims that are 1.

BrianHarrisonAMD:
Ill make a new PR branched off this with my changes for how to handle skipping stride checks when dim == 1.

src/tensor.cpp
```diff
@@ -684,13 +684,42 @@ bool TensorDescriptor::IsPossibleLayout(const std::string& storage_layout,
...
+        {
+            bool allow_violation = false;
+
+            if(lens.size() == 5)
+            {
+                auto char_at_i        = base_layout[i];
+                auto char_at_i_plus_1 = base_layout[i + 1];
+
+                auto dim_pos_i        = storage_layout.find(char_at_i);
+                auto dim_pos_i_plus_1 = storage_layout.find(char_at_i_plus_1);
```
BrianHarrisonAMD, line 700: Edit: actually I think it was fine to access base_layout this way since it was reordered to match base_layout.

BrianHarrisonAMD:
Closing as I think this was addressed with #3771

PR #3765, opened 2025-05-29
Default GTests that use test_drive<> functionality to be disabled

TristanZerrAMD:
Sample output:

```
root@af8a5a3674b8:/home/triszerr/temp/MIOpen/build# ./bin/miopen_gtest --gtest_filter=*gru*
Running main() from /opt/rocm/cget/build/tmp-c4f1a42bcee536579a1248d25169f834201170c187fc6d80e8c510a7d28f6f48/googletest-1.14.0/googletest/src/gtest_main.cc
PRNG seed: 12345678
Note: Google Test filter = *gru*
[==========] Running 2 tests from 2 test suites.
[----------] 1 test from Full/GPU_DeepBenchGRU_FP32
[ RUN      ] Full/GPU_DeepBenchGRU_FP32.FloatTest_deepbench_gru/0
/home/triszerr/temp/MIOpen/test/gtest/deepbench_gru.cpp:45: Skipped
-> GTests making use of test_drive<> functionaly are currently disabled

[  SKIPPED ] Full/GPU_DeepBenchGRU_FP32.FloatTest_deepbench_gru/0 (0 ms)
[----------] 1 test from Full/GPU_DeepBenchGRU_FP32 (0 ms total)

[----------] 1 test from Full/GPU_GRUExtra_FP32
[ RUN      ] Full/GPU_GRUExtra_FP32.FloatTest_gru_extra/0
/home/triszerr/temp/MIOpen/test/gtest/gru_extra.cpp:45: Skipped
-> GTests making use of test_drive<> functionaly are currently disabled

[  SKIPPED ] Full/GPU_GRUExtra_FP32.FloatTest_gru_extra/0 (0 ms)
[----------] 1 test from Full/GPU_GRUExtra_FP32 (0 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 2 test suites ran. (0 ms total)
[  PASSED  ] 0 tests.
[  SKIPPED ] 2 tests, listed below:
[  SKIPPED ] Full/GPU_DeepBenchGRU_FP32.FloatTest_deepbench_gru/0
[  SKIPPED ] Full/GPU_GRUExtra_FP32.FloatTest_gru_extra/0
```


test/gtest/conv_group.cpp
BrianHarrisonAMD, line 1: This file, I think wasn't properly flagging any errors, since it just captures the stderr string, and prints it.
These are the type of files I think we should disable (and I think most are like this).

test/gtest/conv_for_implicit_gemm.cpp
BrianHarrisonAMD, line 1: This file also appears to be capturing the stderr string, and then checking for specific text to flag it.
I don't think we would want to include this in the bulk disable, since it was checking for a specific condition as a failiure.

test/gtest/conv_embed_db.cpp
BrianHarrisonAMD, line 1: This file appears to be checking the stderr stream for a specific fail condition, and probably shouldn't be disabled since it at least does have a condition it's checking for.

TristanZerrAMD:
Thanks @BrianHarrisonAMD.  I went back and did a quick search, and I think you pointed out the only two instances where the code was checking std::err for a specific error condition.  I believe the rest are indeed all forwarding to std::cout.  I reverted those two files, and made the skip message a bit more clear.

randyspauldingamd (commented):
There are two main problems with these:

1) gtest output capture can't be trusted, as discussed in this old-but-still-relevant-I-think thread:
https://stackoverflow.com/questions/3803465/how-to-capture-stdout-stderr-with-googletest

2) Not even the ones that do minimal checking fully replicate the behavior of the original ctests:
https://github.com/ROCm/MIOpen/blob/07ae1851c5eb7f5881f6cd80bd7105efb33ed6ae/test/CMakeLists.txt#L695-L701

I'd say just disable them all to make it clear which need to be revisited. Then start by designing a gtest-aware version of test/driver.hpp. Perhaps it can reuse much of the existing code and we have individual tests pass in an array of predicates for flexibility.

averinevg (commented):
It would be helpful to add TODO comments in the code indicating that skipped tests need to be fixed.

test/gtest/gtest_common.hpp
```diff
@@ -315,6 +315,20 @@ MIOPEN_LIB_ENV_VAR(MIOPEN_DEBUG_CONV_IMPLICIT_GEMM)
 MIOPEN_LIB_ENV_VAR(MIOPEN_LOG_LEVEL)
 MIOPEN_LIB_ENV_VAR(MIOPEN_FIND_ENFORCE)
 
+#define MIOPEN_ENABLE_TEST_DRIVE_WITH_GTEST 0
```
TristanZerrAMD, line 319: @averinevg - I think adding a single TODO type comment would make sense above this line, what do you think?

TristanZerrAMD:
@randyspauldingamd, great feedback, thanks.  

@BrianHarrisonAMD, what are your thoughts on going ahead and just disabling those two files, conv_embed_db and conv_for_implicit_gemm?  I think that's what you are proposing Randy?  Then for a bigger follow up chunk of work, we design a proper gtest-aware version of test/driver.hpp

BrianHarrisonAMD:
> @randyspauldingamd, great feedback, thanks.
> 
> @BrianHarrisonAMD, what are your thoughts on going ahead and just disabling those two files, conv_embed_db and conv_for_implicit_gemm? I think that's what you are proposing Randy? Then for a bigger follow up chunk of work, we design a proper gtest-aware version of test/driver.hpp

Yea I am fine with that.
I don't think these tests have been verified to work properly, and it's better to just disable them since we know many aren't flagging anything now.

randyspauldingamd:
I wrote up a very brief introduction to my idea; link repeated here:
https://github.com/ROCm/MIOpen/issues/3770

BrianHarrisonAMD (approved):

PR #3767, opened 2025-05-30
Enable split_k support for CK backward data solvers (2D)

iq136boy:
Enable split_k support for CK backward data solvers (2D) and loopover the value set of split_k value to find the best split_k and ck kernel instance combination which gives the best performance.


src/solver/conv/conv_hip_implicit_gemm_grouped_bwd_xdlops.cpp
```diff
@@ -453,15 +458,25 @@ bool PerformanceConfigHipImplicitGemmGroupBwdXdlops::SetNextValue(const ProblemD
...
+        if(!NextLinear(0, valid_kernels.size() - 1, index))
+        {
+            kernel_id = valid_kernels[index] + "+" + std::to_string(split_k);
+            break;
+        }
+        // All split_k and index values were iterated
         return false;
+    } while(false);
+#endif
+    return true;
```
BrianHarrisonAMD, line 479: Seems like we could convert the break -> return true;, and then we can remove the do / while loop.
Not a huge deal, but I think it's easier to read that way.

src/solver/conv/conv_hip_implicit_gemm_grouped_bwd_xdlops.cpp
```diff
@@ -367,7 +371,7 @@ bool PerformanceConfigHipImplicitGemmGroupBwdXdlops::RunParameterPredictionModel
            }))
     {
         index     = heuristic_indexes[0];
-        kernel_id = valid_kernels[index];
+        kernel_id = valid_kernels[index] + "+1";
         MIOPEN_LOG_I("Params set by AI: " << ToString());
         return true;
```
BrianHarrisonAMD, line 376: This will require a follow-up for heuristics to predict split_k right?

iq136boy: > This will require a follow-up for heuristics to predict split_k right?

Yes. If I did not give the split_k value, the run will fail. But we do need a prediction value for split_k.

BrianHarrisonAMD (approved):

cderb (approved):

PR #3771, opened 2025-05-30
Allow less restrictive tensor checks

BrianHarrisonAMD:
Based off the changes proposed in #3763, but this is a general check that will allow skipping to the next set of strides when dim == 1.

Note: still need to add more tests to cover these conditions, but opening the PR now to get feedback on changes, and run the CI.


iq136boy (approved):

src/tensor.cpp
```diff
@@ -674,23 +674,46 @@ bool TensorDescriptor::IsPossibleLayout(const std::string& storage_layout,
...
+    std::transform(base_layout.cbegin(), base_layout.cend(), layout_dims.begin(), dims_op);
+
+    unsigned current_index = 0;
+    unsigned next_index    = 1;
+    // Check monotonic decreasing with skipping violations when dim == 1
+    while(next_index < layout_strides.size())
     {
-        if(layout_strides[i] < layout_strides[i + 1])
+        // potential violation detected
+        if(layout_strides[current_index] < layout_strides[next_index])
```
BrianHarrisonAMD: I think it should also skip if the current_index has a dim of 1, and generally should be modified to skip checks when dim == 1.

test/gtest/unit_TensorDescriptor.cpp
```diff
@@ -198,31 +198,60 @@ class TestPossibleLayout4D5D : public ::testing::TestWithParam<TestCasePossibleL
...
-            TestCase{{miopenHalf, miopenTensorNHWC, {1, 1, 1, 1}, {1000, 1, 100, 10}}, {"NHWC"}},
-            TestCase{{miopenHalf, miopenTensorCHWN, {1, 1, 1, 1}, {1, 1000, 100, 10}}, {"CHWN"}},
             TestCase{{miopenHalf, miopenTensorNCHW, {2, 2, 2, 2}, {1000, 100, 10, 1}}, {"NCHW"}},
             TestCase{{miopenHalf, miopenTensorNHWC, {2, 2, 2, 2}, {1000, 1, 100, 10}}, {"NHWC"}},
             TestCase{{miopenHalf, miopenTensorCHWN, {2, 2, 2, 2}, {1, 1000, 100, 10}}, {"CHWN"}},
-            TestCase{{miopenHalf, miopenTensorNCDHW, {1, 1, 1, 1, 1}, {10000, 1000, 100, 10, 1}}, {"NCDHW"}},
-            TestCase{{miopenHalf, miopenTensorNDHWC, {1, 1, 1, 1, 1}, {10000, 1, 1000, 100, 10}}, {"NDHWC"}},
+            TestCase{{miopenHalf, miopenTensorNCHW, {2, 2, 2, 2}, {1000, 100, 10, 1}}, {"NCHW"}},
+            TestCase{{miopenHalf, miopenTensorNHWC, {2, 2, 2, 2}, {1000, 1, 100, 10}}, {"NHWC"}},
+            TestCase{{miopenHalf, miopenTensorCHWN, {2, 2, 2, 2}, {1, 1000, 100, 10}}, {"CHWN"}},
```
BrianHarrisonAMD: All existing checks that had dims of 1, now need to be changed to dims > 1. 

Added extra checks to verify more relaxed layout logic as well.

test/gtest/unit_TensorDescriptor.cpp
```diff
@@ -198,31 +198,58 @@ class TestPossibleLayout4D5D : public ::testing::TestWithParam<TestCasePossibleL
...
             TestCase{{miopenHalf, {2, 2, 2, 2}, {1000, 100, 10, 1}}, {"NCHW"}},
             TestCase{{miopenHalf, {2, 2, 2, 2}, {1000, 1, 100, 10}}, {"NHWC"}},
             TestCase{{miopenHalf, {2, 2, 2, 2}, {1, 1000, 100, 10}}, {"CHWN"}},
             TestCase{{miopenHalf, {2, 2, 2, 2}, {1, 1, 1, 1}}, {"NCHW", "NHWC", "CHWN"}},
-            TestCase{{miopenHalf, {1, 1, 1, 1, 1}, {10000, 1000, 100, 10, 1}}, {"NCDHW"}},
-            TestCase{{miopenHalf, {1, 1, 1, 1, 1}, {10000, 1, 1000, 100, 10}}, {"NDHWC"}},
-            TestCase{{miopenHalf, {1, 1, 1, 1, 1}, {1, 1, 1, 1, 1}}, {"NCDHW", "NDHWC"}},
+            TestCase{{miopenHalf, {2, 2, 2, 2, 2}, {10000, 1000, 100, 10, 1}}, {"NCDHW"}},
+            TestCase{{miopenHalf, {2, 2, 2, 2, 2}, {10000, 1, 1000, 100, 10}}, {"NDHWC"}},
+            TestCase{{miopenHalf, {2, 2, 2, 2, 2}, {1, 1, 1, 1, 1}}, {"NCDHW", "NDHWC"}},
```
randyspauldingamd: Duplicate of line 214. At first glance it doesn't look like it should be valid, although having this test case simply marks it as an intentional way users can save memory for tensors where all values are the same. I don't see a reason to disallow this. We should add the 4D case as well.

test/gtest/unit_TensorDescriptor.cpp
```diff
@@ -198,31 +198,58 @@ class TestPossibleLayout4D5D : public ::testing::TestWithParam<TestCasePossibleL
             TestCase{{miopenHalf, miopenTensorNDHWC, {1, 1, 1, 1, 1}}, {"NCDHW", "NDHWC"}},
             TestCase{{miopenHalf, miopenTensorNDHWC, {2, 2, 2, 2, 2}}, {"NDHWC"}},
 
-            TestCase{{miopenHalf, {1, 1, 1, 1}, {1000, 100, 10, 1}}, {"NCHW"}},
-            TestCase{{miopenHalf, {1, 1, 1, 1}, {1000, 1, 100, 10}}, {"NHWC"}},
-            TestCase{{miopenHalf, {1, 1, 1, 1}, {1, 1000, 100, 10}}, {"CHWN"}},
+            TestCase{{miopenHalf, {2, 2, 2, 2}, {1000, 100, 10, 1}}, {"NCHW"}},
+            TestCase{{miopenHalf, {2, 2, 2, 2}, {1000, 1, 100, 10}}, {"NHWC"}},
+            TestCase{{miopenHalf, {2, 2, 2, 2}, {1, 1000, 100, 10}}, {"CHWN"}},
```
randyspauldingamd: Duplicates of 205-207

test/gtest/unit_TensorDescriptor.cpp
```diff
@@ -198,31 +198,58 @@ class TestPossibleLayout4D5D : public ::testing::TestWithParam<TestCasePossibleL
...
             TestCase{{miopenHalf, {1, 1, 1, 1}, {1, 1, 1, 1}}, {"NCHW", "NHWC", "CHWN"}},
             TestCase{{miopenHalf, {2, 2, 2, 2}, {1000, 100, 10, 1}}, {"NCHW"}},
             TestCase{{miopenHalf, {2, 2, 2, 2}, {1000, 1, 100, 10}}, {"NHWC"}},
             TestCase{{miopenHalf, {2, 2, 2, 2}, {1, 1000, 100, 10}}, {"CHWN"}},
             TestCase{{miopenHalf, {2, 2, 2, 2}, {1, 1, 1, 1}}, {"NCHW", "NHWC", "CHWN"}},
-            TestCase{{miopenHalf, {1, 1, 1, 1, 1}, {10000, 1000, 100, 10, 1}}, {"NCDHW"}},
-            TestCase{{miopenHalf, {1, 1, 1, 1, 1}, {10000, 1, 1000, 100, 10}}, {"NDHWC"}},
-            TestCase{{miopenHalf, {1, 1, 1, 1, 1}, {1, 1, 1, 1, 1}}, {"NCDHW", "NDHWC"}},
+            TestCase{{miopenHalf, {2, 2, 2, 2, 2}, {10000, 1000, 100, 10, 1}}, {"NCDHW"}},
+            TestCase{{miopenHalf, {2, 2, 2, 2, 2}, {10000, 1, 1000, 100, 10}}, {"NDHWC"}},
```
randyspauldingamd: Duplicates of 212-213

test/gtest/unit_TensorDescriptor.cpp
```diff
@@ -198,31 +198,58 @@ class TestPossibleLayout4D5D : public ::testing::TestWithParam<TestCasePossibleL
...
-            TestCase{{miopenHalf, miopenTensorNHWC, {1, 1, 1, 1}, {1000, 1, 100, 10}}, {"NHWC"}},
-            TestCase{{miopenHalf, miopenTensorCHWN, {1, 1, 1, 1}, {1, 1000, 100, 10}}, {"CHWN"}},
             TestCase{{miopenHalf, miopenTensorNCHW, {2, 2, 2, 2}, {1000, 100, 10, 1}}, {"NCHW"}},
             TestCase{{miopenHalf, miopenTensorNHWC, {2, 2, 2, 2}, {1000, 1, 100, 10}}, {"NHWC"}},
             TestCase{{miopenHalf, miopenTensorCHWN, {2, 2, 2, 2}, {1, 1000, 100, 10}}, {"CHWN"}},
-            TestCase{{miopenHalf, miopenTensorNCDHW, {1, 1, 1, 1, 1}, {10000, 1000, 100, 10, 1}}, {"NCDHW"}},
-            TestCase{{miopenHalf, miopenTensorNDHWC, {1, 1, 1, 1, 1}, {10000, 1, 1000, 100, 10}}, {"NDHWC"}},
+            TestCase{{miopenHalf, miopenTensorNCHW, {2, 2, 2, 2}, {1000, 100, 10, 1}}, {"NCHW"}},
+            TestCase{{miopenHalf, miopenTensorNHWC, {2, 2, 2, 2}, {1000, 1, 100, 10}}, {"NHWC"}},
+            TestCase{{miopenHalf, miopenTensorCHWN, {2, 2, 2, 2}, {1, 1000, 100, 10}}, {"CHWN"}},
```
randyspauldingamd: Duplicates of 216-218

test/gtest/unit_TensorDescriptor.cpp
randyspauldingamd: I think we already had all of the 2's. I marked them so you can double-check.

test/gtest/unit_TensorDescriptor.cpp
```diff
@@ -198,31 +198,58 @@ class TestPossibleLayout4D5D : public ::testing::TestWithParam<TestCasePossibleL
...
             TestCase{{miopenHalf, miopenTensorNCDHW, {2, 2, 2, 2, 2}, {10000, 1000, 100, 10, 1}}, {"NCDHW"}},
             TestCase{{miopenHalf, miopenTensorNDHWC, {2, 2, 2, 2, 2}, {10000, 1, 1000, 100, 10}}, {"NDHWC"}},
+
+            // Test cases where invalid strides with dim of 1 are allowed.
+            // (When dim is 1 we can safely ignore stride values.)
+            // Note: Some layouts are equivalent when dim == 1 (NCHW == CHWN when n = 1 for example)
```
randyspauldingamd: Suggested rewording (feel free to continue tweaking as you see fit):
```
// Stride is ignored when the corresponding dimension length is 1. As a result,
// multiple layouts have identical memory footprints. For example, NCHW == CHWN for n = 1.
```

randyspauldingamd (commented):
I would greatly prefer strict checking to help users catch mistakes early on, but I don't think it's necessary to impose it upon them. Forge onward!

test/gtest/unit_TensorDescriptor.cpp
BrianHarrisonAMD: Oh good point, I didn't realize we already had these exact cases with non-1 dims hehe.

Ill remove the dupes and only keep the new checks.

test/gtest/unit_TensorDescriptor.cpp
BrianHarrisonAMD: Yea very clear now that I am looking at it again that these are duped.
Ill remove, and only keep the new ones.

JonathanLichtnerAMD (approved):

test/gtest/unit_TensorDescriptor.cpp
BrianHarrisonAMD: I think we have the 4D case isn't it same as:
` TestCase{{miopenHalf, {2, 2, 2, 2}, {1, 1, 1, 1}}, {"NCHW", "NHWC", "CHWN"}},` ?

test/gtest/graphapi_conv_bias_res_add_activ_fwd.cpp
```diff
@@ -55,9 +55,9 @@ static std::vector<Conv3DTestCase> ConvTestConfigs()
 { // g   n   c   k   image   filter   pad   stride   dilation
     // clang-format off
     return {{1, 1, 4, 4, {14, 11, 1}, {3, 3, 3}, {1, 1, 1}, {1, 1, 1}, {1, 1, 1}, miopenConvolution},
-            {1, 1, 1, 1, {1, 4, 4}, {2, 2, 2}, {1, 1, 1}, {1, 1, 1}, {1, 1, 1}, miopenConvolution},
-            {1, 1, 1, 1, {8, 8, 8}, {2, 2, 2}, {0, 0, 0}, {1, 1, 1}, {1, 1, 1}, miopenConvolution},
-            {1, 1, 1, 1, {8, 8, 8}, {2, 2, 2}, {0, 0, 0}, {2, 2, 2}, {1, 1, 1}, miopenConvolution},
+            {1, 1, 2, 1, {1, 4, 4}, {2, 2, 2}, {1, 1, 1}, {1, 1, 1}, {1, 1, 1}, miopenConvolution},
+            {1, 1, 2, 1, {8, 8, 8}, {2, 2, 2}, {0, 0, 0}, {1, 1, 1}, {1, 1, 1}, miopenConvolution},
+            {1, 1, 2, 1, {8, 8, 8}, {2, 2, 2}, {0, 0, 0}, {2, 2, 2}, {1, 1, 1}, miopenConvolution},
```
BrianHarrisonAMD, line 60: Edit: Looked into this more, and think it might be something odd with the graphAPI path of how the layout gets decided for fusions that leads to it thinking the tensors are NCHW when fusion only supports NHWC.

Going to run the remaining tests to see if there is anything else having issues or if this was isolated to this graphAPI test that we are deprecating and removing eventually.

cderb (approved):

randyspauldingamd (approved):
I guess they aren't going to supply more information in response to Evgenii's questions, so this LGTM unless you want to ping them again.

However, this has made it more clear how the string layouts are a limiting design flaw. What we need for more generic use cases such as this is a set of enums so the solvers can quickly determine whether the memory layout is appropriate for their kernels by checking flag(s). That of course would require a very large refactor. We could mimic it quite easily using the existing API though with little added compute expense, by simply including all valid layouts as a delimited list within the string. For example, for tensor dims {5, 1, 10, 10} the memory layout string would be "|NCHW|NHWC|CHWN|" (plus any others). Notice that I've included delims at the start and end which clearly marks it as a multiple-layout string. If the first character is a delimiter we would require "|\<layout\>|" within the string for a match. We'd also add helper function(s) to handle all this conditional stuff such as:
```
TensorDescriptor::IsCompatibleLayout(const std::string& test_layout, const std::string& possible_layouts){...}
```

BrianHarrisonAMD:
> I guess they aren't going to supply more information in response to Evgenii's questions, so this LGTM unless you want to ping them again.
> 
> However, this has made it more clear how the string layouts are a limiting design flaw. What we need for more generic use cases such as this is a set of enums so the solvers can quickly determine whether the memory layout is appropriate for their kernels by checking flag(s). That of course would require a very large refactor. We could mimic it quite easily using the existing API though with little added compute expense, by simply including all valid layouts as a delimited list within the string. For example, for tensor dims {5, 1, 10, 10} the memory layout string would be "|NCHW|NHWC|CHWN|" (plus any others). Notice that I've included delims at the start and end which clearly marks it as a multiple-layout string. If the first character is a delimiter we would require "|<layout>|" within the string for a match. We'd also add helper function(s) to handle all this conditional stuff such as:
> 
> ```
> TensorDescriptor::IsCompatibleLayout(const std::string& test_layout, const std::string& possible_layouts){...}
> ```

Yea I agree.
I think I wrote something up, but then replaced it once I realized the test breaks were due to overwriting the calculated layout when it wasn't needed.

A full proper fix would account for the fact that multiple layouts might be supported, and we would need solvers to check if their required layout is one of them. This also also can impact the solvers where it supports multiple layouts, but one is faster. 

MISA & CK for example can support either, but if NHWC is allowed, then it should run that path instead.

For now I think this gives us a bit more flexibility, but there is still more room to improve if we have time.

PR #3775, opened 2025-06-03
[Conv] group conv bias active 

bghimireamd:
* add CK's fused grouped conv + bias + activ as solver
*  (2D)NHWC
*  bfp16
* driver cmd : `MIOPEN_DEBUG_DISABLE_FIND_DB=1  MIOPEN_DEBUG_FIND_ONLY_SOLVER=ConvCKIgemmGrpFwdBiasActivFused ./bin/MIOpenDriver CBAInferbfp16 -F 4 -n 1 -c 64 -H 56 -W 56 -k 64 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -g 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC` working


src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -0,0 +1,482 @@
...
+    if(desc.op_map.size() != 3)
+        return false;
+    if(desc.op_map[0]->kind() != miopenFusionOpConvForward)
+        return false;
+    if(desc.op_map[1]->kind() != miopenFusionOpBiasForward)
+        return false;
+    if(desc.op_map[2]->kind() != miopenFusionOpActivForward)
+        return false;
+    const auto& activ_op = dynamic_cast<ActivFwdFusionOpDescriptor&>(*desc.op_map[2]);
+    if(activ_op.activMode != miopenActivationRELU)
```
BrianHarrisonAMD: I think this needs to allow more activation modes (clipped & clamp?)

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -0,0 +1,482 @@
...
+    const auto& desc = *fdesc_problem.fusion_plan_desc;
+    if(desc.op_map.empty())
+    {
+        MIOPEN_THROW(miopenStatusInternalError, "desc.op_map.empty()");
+    }
+    if(desc.op_map.size() != 3)
+        return false;
+    if(desc.op_map[0]->kind() != miopenFusionOpConvForward)
+        return false;
+    if(desc.op_map[1]->kind() != miopenFusionOpBiasForward)
```
BrianHarrisonAMD, line 420: We may need to be flexible on bias here.
We should check if it's easy to allow having no bias.

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -0,0 +1,482 @@
...
+        return false;
+    const std::string arch = ctx.GetStream().GetDeviceName();
+    if(arch != "gfx908" && arch != "gfx90a" && arch != "gfx942")
+        return false;
+    if(!conv_problem.IsLayoutNHWC())
+        return false;
+
+    switch(conv_problem.GetInDataType())
+    {
+    case miopenBFloat16: return CheckCKApplicability<ck::bhalf_t>(conv_problem);
```
BrianHarrisonAMD, line 452: Can we extend support to other dtypes as well?

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -0,0 +1,482 @@
...
+            dynamic_cast<miopen::fusion::BiasOpInvokeParam&>(*data_ctx.op_args.params[1]);
+        assert(&bias_param);
+
+        return MakeArgPtr(op_ptr,
+                          data_ctx.in,
+                          conv_param.weights,
+                          bias_param.bdata,
+                          data_ctx.out,
+                          conv_param.alpha,
+                          conv_param.beta);
```
BrianHarrisonAMD: Don't we need to extract the activation params here and provide to CK?

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -0,0 +1,482 @@
...
+            {bias_strides},
+            output_len,
+            out_strides,
+            strides,
+            dilation,
+            lPadding,
+            rPadding,
+            in_element_op,
+            wei_element_op,
+            out_element_op);
```
BrianHarrisonAMD: Ah I think this needs to be a lambda that uses the activation values for determining how to do the out_element_op.
Right now it forces RELU.

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
bghimireamd: not right now. CK only has instances for bhalf_t.

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
bghimireamd: I think so.

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
BrianHarrisonAMD: Looks like we need a separate solver for this.

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -0,0 +1,504 @@
...
+            wei_element_op,
+            clampOp);
+    }
+
+    template <typename DevOpPtr>
+    auto MakeArgPtr(const DevOpPtr& op_ptr,
+                    const miopen::fusion::FusionInvokeParams& data_ctx) const
+    {
+        const auto& conv_param =
+            dynamic_cast<miopen::fusion::ConvolutionOpInvokeParam&>(*data_ctx.op_args.params[0]);
```
reidkwja, line 200: In an event where we have a `dynamic_cast<ConvolutionOpInvokeParam&>` fail, it might not get to the assert part in a situations where ` params[0] ` is not a ConvolutionOpInvokeParam. Do you think using a pointer here instead, like `dynamic_cast<ConvolutionOpInvokeParam*>` along with the checks/exceptions, would be a better suitable alternative for this PR? 

This is can also be applicable to other parts in the code as well - where `dynamic_cast<>` is used.

bghimireamd: I do have `assert(&conv_param);` . Probably have to address in another PR.

test/gtest/conv_test_base.hpp
```diff
@@ -114,6 +114,19 @@ struct GroupConvTestConfig<2u>
...
+        return {
+            // clang-format off
+            // g   n    C    K    img         filter    pad     stride  dilation
+            {1,   32,   64,  128, {28, 28},   {3, 3},   {0, 1}, {1, 2}, {2, 1}},
+            {32,  16,   32,   64,  {7, 7},    {3, 3},   {1, 1}, {1, 1}, {1, 1}},
+            {1,   16,   32,   64, {16, 16},   {2, 2},   {0, 0}, {3, 3}, {1, 1}},
+            {4,    8,   16,   32, {32, 4},    {3, 1},   {1, 0}, {1, 1}, {1, 1}},
+            // clang-format on
+        };
+    }
```
SamuelReeder, line 128: Small shapes for these tests but should be lots of breadth. Clamp uses full existing test config.

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -0,0 +1,504 @@
...
+
+bool ConvCKIgemmGrpFwdBiasActivFused::IsApplicable(const FusionContext& ctx,
+                                                   const FusionDescription& fdesc_problem) const
+{
+#if !MIOPEN_BACKEND_HIP || !MIOPEN_USE_COMPOSABLEKERNEL
+    std::ignore = ctx;
+    std::ignore = fdesc_problem;
+    return false;
+#else
+    const auto& desc = *fdesc_problem.fusion_plan_desc;
```
SamuelReeder, line 411: Is it guaranteed that `FusionDescription` is not populated with `nullptr` for `fusion_plan_desc`? I'm not fully familiar with the flow.

SamuelReeder (approved):
Assuming my last comment is a non-issue, LGTM.

src/include/miopen/fusion/solvers.hpp
```diff
@@ -185,6 +185,76 @@ struct ConvCKIgemmFwdBiasActivFused final
     bool CheckCKApplicability(const miopen::conv::ProblemDescription&) const;
 };
 
+struct PerformanceConfigConvCKIgemmGrpFwdBiasActivFused
+    : PerfConfigBase<PerformanceConfigConvCKIgemmGrpFwdBiasActivFused>
+{
+    int index;
+    std::string kernel_id;
+    std::vector<std::string> valid_kernels;
+    PerformanceConfigConvCKIgemmGrpFwdBiasActivFused(int idx, std::string kernl_id)
```
adickin-amd, line 194: typo to fix on kernl_id

driver/CBAInferFusion_driver.hpp
```diff
@@ -264,11 +271,91 @@ class CBAInferFusionDriver : public Driver
...
+        }
+    }
+}
+
 template <typename Tgpu, typename Tref>
 int CBAInferFusionDriver<Tgpu, Tref>::ParseCmdLineArgs(int argc, char* argv[])
 {
     inflags.Parse(argc, argv);
 
+    // todo: get spatial dimension from cmd line
```
adickin-amd, line 321: i assume this will be done with the 3d work

driver/CBAInferFusion_driver.hpp
```diff
@@ -679,9 +792,13 @@ int CBAInferFusionDriver<Tgpu, Tref>::AllocateBuffersAndCopy()
         for(int i = 0; i < wei_sz; i++)
         {
 #if(CBA_DEBUG_VALUES == 1)
-            wei[i] = 1.; // prng::gen_canonical<Tgpu>(); // 1.;
+            auto rval   = 1.; // prng::gen_canonical<Tgpu>(); // 1.;
```
adickin-amd, line 795: do we need the commented out code?  is it valuable?

adickin-amd (approved):

driver/CBAInferFusion_driver.hpp
bghimireamd: probably not

bghimireamd:
LGTM

driver/CBAInferFusion_driver.hpp
bghimireamd: yes! https://github.com/ROCm/MIOpen/pull/3802

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
bghimireamd: yeah we can address that latter. May be making const ref could help https://github.com/ROCm/MIOpen/blob/develop/src/include/miopen/fusion/problem_description.hpp#L46

PR #3783, opened 2025-06-06
Add compile time conditional clearing of the output buffers 

adickin-amd:
- grouped 2d/3d conv in CK have clearing built in now, make it so we dont clear as well for grouped conv only.

### DIFF
--- a/cmake/ClangCheck.cmake
+++ b/cmake/ClangCheck.cmake
@@ -1,0 +1,23 @@
+# Copyright © Advanced Micro Devices, Inc., or its affiliates.
+# SPDX-License-Identifier:  MIT
+
+set(CLANG_FORMAT_PRUNE -path "./build" -prune -o -path "./install" -prune -o -path "./fin" -prune -o)
+
+# Note: The clang-format in /opt/rocm produces different results than the one in /usr/bin.  MIOpen
+# formatting is based on the one in /usr/bin so we use that one
+# set(CLANG_FORMAT_BINARY /opt/rocm/llvm/bin/clang-format)
+set(CLANG_FORMAT_BINARY /usr/bin/clang-format-12)
+
+add_custom_target(
+    check_format
+    COMMAND  find . ${CLANG_FORMAT_PRUNE} -regex ".*\\.\\(cpp\\|hpp\\|h.in\\|hpp.in\\|cpp.in\\|cl\\)" -exec ${CLANG_FORMAT_BINARY} --dry-run --Werror {} +
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    VERBATIM
+)
+
+add_custom_target(
+    format
+    COMMAND  find . ${CLANG_FORMAT_PRUNE} -regex ".*\\.\\(cpp\\|hpp\\|h.in\\|hpp.in\\|cpp.in\\|cl\\)" -exec ${CLANG_FORMAT_BINARY} --verbose -i {} +
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    VERBATIM
+)
--- a/driver/conv_driver.hpp
+++ b/driver/conv_driver.hpp
@@ -101,10 +101,10 @@ struct AutoMiopenWarmupMode
         miopen::debug::FindEnforceDisable = true;
         miopen::debug::IsWarmupOngoing    = true;
     }
-    AutoMiopenWarmupMode(const AutoMiopenWarmupMode&) = delete;
-    AutoMiopenWarmupMode(AutoMiopenWarmupMode&&)      = delete;
+    AutoMiopenWarmupMode(const AutoMiopenWarmupMode&)            = delete;
+    AutoMiopenWarmupMode(AutoMiopenWarmupMode&&)                 = delete;
     AutoMiopenWarmupMode& operator=(const AutoMiopenWarmupMode&) = delete;
-    AutoMiopenWarmupMode& operator=(AutoMiopenWarmupMode&&) = delete;
+    AutoMiopenWarmupMode& operator=(AutoMiopenWarmupMode&&)      = delete;
     ~AutoMiopenWarmupMode()
     {
         miopen::debug::LoggingQuiet       = debug_logging_quiet_prev;
@@ -127,10 +127,10 @@ struct AutoPrepareForGpuReference
         miopen::debug::AlwaysEnableConvDirectNaive = true;
         miopen::debug::LoggingQuiet                = true;
     }
-    AutoPrepareForGpuReference(const AutoPrepareForGpuReference&) = delete;
-    AutoPrepareForGpuReference(AutoPrepareForGpuReference&&)      = delete;
+    AutoPrepareForGpuReference(const AutoPrepareForGpuReference&)            = delete;
+    AutoPrepareForGpuReference(AutoPrepareForGpuReference&&)                 = delete;
     AutoPrepareForGpuReference& operator=(const AutoPrepareForGpuReference&) = delete;
-    AutoPrepareForGpuReference& operator=(AutoPrepareForGpuReference&&) = delete;
+    AutoPrepareForGpuReference& operator=(AutoPrepareForGpuReference&&)      = delete;
     ~AutoPrepareForGpuReference()
     {
         miopen::debug::LoggingQuiet                = quiet_prev;
@@ -391,15 +391,16 @@ private:
     miopenConvolutionMode_t mode;
 
     bool is_wrw = true, is_bwd = true, is_fwd = true;
-    bool is_wrw_winograd       = false;
-    bool is_wrw_igemm          = false;
-    bool is_fwd_igemm          = false;
-    bool is_bwd_igemm          = false;
-    bool time_enabled          = false;
-    bool wall_enabled          = false;
-    bool warmup_enabled        = false;
-    bool is_gpualloc           = false;
-    GPUMem::Check buffer_check = GPUMem::Check::None;
+    bool is_wrw_winograd              = false;
+    bool is_wrw_igemm                 = false;
+    bool is_fwd_igemm                 = false;
+    bool is_bwd_igemm                 = false;
+    bool time_enabled                 = false;
+    bool wall_enabled                 = false;
+    bool warmup_enabled               = false;
+    bool is_gpualloc                  = false;
+    bool populate_output_with_garbage = false;
+    GPUMem::Check buffer_check        = GPUMem::Check::None;
 
     int num_iterations = 1;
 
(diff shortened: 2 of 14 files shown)


requirements.txt
```diff
@@ -7,5 +7,5 @@ nlohmann/json@v3.11.2 -DJSON_MultipleHeaders=ON -DJSON_BuildTests=Off
 ROCm/FunctionalPlus@v0.2.18-p0
 ROCm/eigen@3.4.0
 ROCm/frugally-deep@9683d557eb672ee2304f80f6682c51242d748a50
-ROCm/composable_kernel@00247e3c297032a2cbdaae465113648ec1857d3f -DCMAKE_BUILD_TYPE=Release
+ROCm/composable_kernel@f20b6e492dc00e7bee410b5a2a1e17fc89c3e636 -DCMAKE_BUILD_TYPE=Release
```
BrianHarrisonAMD: Before merging make sure we coordinate with CK.
We basically need both changes to go in at the same time hehe.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -1008,11 +1013,14 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
...
                     {
-                        elapsed += handle.GetKernelTime();
+                        ZeroOutTensor(handle, data_ctx.tensors.dwDesc, data_ctx.tensors.dw);
+
+                        if(handle.IsProfilingEnabled())
+                        {
+                            elapsed += handle.GetKernelTime();
+                        }
                     }
                 }
```
BrianHarrisonAMD, line 1024: Just a nit, but the ZeroOutputs check happen before the alpha_beta_case check right?

adickin-amd: ah yea, ill swap

requirements.txt
adickin-amd: Is this right for steps
1. Merge CK PR when were ready
2. Update this PR with new CK sha
3. Let this PR fully build new docker
4. Merge this PR

src/solver/conv/conv_hip_implicit_gemm_3d_grouped_wrw_xdlops.cpp
```diff
@@ -537,18 +537,21 @@ ConvSolution ConvHipImplicitGemm3DGroupWrwXdlops::GetSolution(
...
+                                                 false,
                                                  DeviceOpGBwdWeightScalePtrs<T>,
                                                  CKArgs<T>,
                                                  miopen::conv::WrWInvokeParams>(
                     ctx, problem, config.kernel_id);
             default:
                 return InitInvokerFactoryWrwNCHW<3,
+                                                 false,
                                                  DeviceOpGBwdWeightDefaultPtrs<T>,
                                                  CKArgs<T>,
```
BrianHarrisonAMD, line 556: I imagine yes, but we confirmed 3D also has the automatic clearing added?

adickin-amd: yes, that is my understanding.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -868,12 +869,15 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
...
-            /// Note: Need to clear buffer memory for output since all values may not be set.
-            elapsed = handle.IsProfilingEnabled() ? handle.GetKernelTime() : 0.0f;
-            output_tr_inst.ZeroOutBuffer(handle);
-            if(handle.IsProfilingEnabled())
-                elapsed += handle.GetKernelTime();
+            if constexpr(ZeroOutputs)
+            {
+                /// \todo: Will need SetTensor() to properly zero out non-packed tensors
+                /// Note: Need to clear buffer memory for output since all values may not be set.
+                elapsed = handle.IsProfilingEnabled() ? handle.GetKernelTime() : 0.0f;
```
BrianHarrisonAMD: This needs to happen even if ZeroOutputs isn't set I believe.

The ConvertFrom calls above will accumulate kernel time in the handle, and I think the WorkAroundHipEventProfiler wipes it out.

The way it works now is pretty error prone since we need to keep accumulating kernel time in elapsed manually. 

Probably want to also check that the way timing is calculated before / after this change is the same to catch these unexpected errors.

adickin-amd: ill pull it out of the if

driver/driver.hpp
```diff
@@ -275,6 +277,22 @@ class GpumemTensor
...
+            // In the past we have had some issues with incorrect results due to Nans in the output
+            // buffers.  In order to test we clear output buffers before computing anything you can
+            // init the buffers with NaNs.
+            // Note, we only do this for the gpu buffers, adding the behaviour for the host buffers
+            // causes a crash as the host code doesnt handle NaNs.
+            auto dev_buf_ptr = static_cast<Tgpu*>(dev->buf);
+            for(size_t i = 0; i < dev->sz; ++i)
+            {
+                dev_buf_ptr[i] = std::numeric_limits<Tgpu>::quiet_NaN();
+            }
```
BrianHarrisonAMD: Doesn't this need to be like the inits below that uses a hipMemCpy? 
I don't think we should be writing the device buffer directly.

```
    template <typename T>
    status_t AllocOnDeviceAndInit(stream q,
                                  context_t ctx,
                                  const size_t sz,
                                  std::vector<T>& init,
                                  GPUMem::Check check = GPUMem::Check::None)
    {
        AllocOnDevice(q, ctx, sz, init, check);
        if(is_gpualloc)
        {
            /// \ref gpualloc_random_init
            return gpumemrand::gen_0_1(static_cast<Tgpu*>(GetDevicePtr()), sz);
        }
        return dev->ToGPU(q, init.data());
    }
```

adickin-amd: Depends what we want to do.  Originally I put the nan population into the host buffers that relies on the `ToGPU` call to copy them over.  Our host side code does not correctly handle the Nans so I just made our GPU side buffer have them

adickin-amd: Ill debug the host side buffer to see how much we need to change to make it robust to Nans.

BrianHarrisonAMD: I think that's fine, but for setting the device data I think you still need to use hipMemCpy.

driver/driver.hpp
```diff
@@ -275,6 +277,22 @@ class GpumemTensor
                                   GPUMem::Check check = GPUMem::Check::None)
     {
         AllocOnDevice(q, ctx, sz, check);
+
+        if(init_gpu_output_nan)
+        {
+            // In the past we have had some issues with incorrect results due to Nans in the output
+            // buffers.  In order to test we clear output buffers before computing anything you can
+            // init the buffers with NaNs.
```
JonathanLichtnerAMD: You have a run-on sentence here...

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -867,13 +868,16 @@ ConvSolution InitInvokerFactoryNCHW(const ExecutionContext& ctx,
             input1_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
             input2_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
             output_init_tr_inst.ConvertFrom(handle, kernels, conv_tensors);
-
             /// \todo: Will need SetTensor() to properly zero out non-packed tensors
             /// Note: Need to clear buffer memory for output since all values may not be set.
```
JonathanLichtnerAMD: I think this comment should be moved into the if statement starting at line 875.

I'm also not sure about the \todo comment just above it...

driver/driver.hpp
adickin-amd: What buffer should I copy from when using hipMemCpy?  If I dont want to mess with host memory, I need to allocate another full buffer, populate that with nans, then copy it over dont i?  Its possible Im over complicating this.

src/include/miopen/solver/implicitgemm_ck_util.hpp
adickin-amd: moved the comment.  removed the todo, i think its confusing as well.

src/include/miopen/solver/implicitgemm_ck_util.hpp
adickin-amd: note, I put this back.  If I put the alpha_beta_case inside the constexpr if, we get unused variable warnings.  The alternative is some kind of extra code to void cast alpha_beta_case.

requirements.txt
BrianHarrisonAMD: Yea

driver/driver.hpp
BrianHarrisonAMD: https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___memory.html#gac7441e74affcce4b8b69dba996c5ebc4

I think you just need to use hipMemset in this case to init the buffer.

BrianHarrisonAMD: It's also potentially a bit more complicated than this, as NaN can change based off the data type. 
We would want to ensure we use the proper NaN value for each datatype.

cmake/ClangCheck.cmake
```diff
@@ -0,0 +1,23 @@
+# Copyright © Advanced Micro Devices, Inc., or its affiliates.
+# SPDX-License-Identifier:  MIT
+
+set(CLANG_FORMAT_PRUNE -path "./build" -prune -o -path "./install" -prune -o -path "./fin" -prune -o)
+
+# Note: The clang-format in /opt/rocm produces different results than the one in /usr/bin.  MIOpen
```
adickin-amd, line 6: I kept breaking the formatting so i brought this cmake file over from hipdnn so we can have make check_format and make format targets

src/include/miopen/solver/implicitgemm_ck_util.hpp
JonathanLichtnerAMD: Thanks.  Minor (optional) nit:  should we change "///" to "//"?

driver/driver.hpp
adickin-amd: okay, I think this is now correct.

cmake/ClangCheck.cmake
BrianHarrisonAMD: We add should add that to the documentation for contributing since that is a nicer way to do formatting.

BrianHarrisonAMD (approved):
LGTM

cmake/ClangCheck.cmake
adickin-amd: I will update that Wiki page once this is merged

driver/conv_driver.hpp
```diff
@@ -2317,6 +2340,11 @@ int ConvDriver<Tgpu, Tref>::RunForwardGPUReference()
         std::cout << "gpu reference convolution does not support bias yet" << std::endl;
         return -1;
     }
+    if(init_output_nan)
```
adickin-amd, line 2346: Did we want to have nans possible when doing gpu reference implementation?

SamuelReeder (approved):
LGTM now!

BrianHarrisonAMD (approved):

PR #3786, opened 2025-06-09
[BUG] [CONV] Fix incorrect stride calculation when w=1/h=1 in MISA solvers

SreecharanGundaboluAMD:
**Overview**
This pull request addresses a critical bug in the FWD, WRW and BWD solvers where the stride was calculated incorrectly due to buggy manual flipping of inputs and outputs. 

Can be reproduced using:
`MIOpenDriver conv -n 3 -c 1 -H 1 -W 1 -k 3 -y 3 -x 2 -p 2 -q 1 -u 2 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 0`

**Bugfix Details**
The manual direction flips previously implemented in the FWD, WRW and BWD solvers have been replaced with methods from the `ProblemInterpreter` class which perform flips based on the convolution direction, ensuring consistency and accuracy. The `ProblemInterpreter` class is already utilized in most CK solvers.

**Enhancements**
1.Regression Tests
Added regression tests to relevant unit tests to ensure the correctness of the stride calculation and the overall functionality of the WRW and BWD solvers.
2. Code Cleanup: 
Removed WORKAROUND_ISSUE_2867, as this issue is effectively addressed by the bugfix implemented in this pull request.


shurale-nkn (approved):
LGTM

BradPepersAMD (approved):
Looks like a solid improvement to use standard methods across CK and MISA and add some test cases to catch similar mistakes.

PR #3792, opened 2025-06-10
Move HipTidy step to FullTests

TristanZerrAMD:


adickin-amd (approved):
Looks correct to me

BrianHarrisonAMD:
Do we want this moved to nightly?

I think not running this on PRs will potentially flag a lot of issues in develop instead.

JonathanLichtnerAMD:
> Do we want this moved to nightly?
> 
> I think not running this on PRs will potentially flag a lot of issues in develop instead.

@TristanZerrAMD Is the concern with this step that it takes roughly twice as long as the other steps in this stage (~ 1 hour vs ~30 minutes or so), and so the "Full Tests" are delayed by an extra 30 minutes?  If so, we could move this step to the "Full Tests" stage.  That should save us ~30 minutes CI time.  Maybe the stages really become "Fast Checks" or "Quick Checks" (i.e., 30 minutes or less) and "Full or Slow Tests" (~1+ hour runtimes).

I'd also be okay leaving this in the first stage, since 1 hour isn't so terrible... it might also be worth seeing if this can be sped up, i.e., what is taking most of the time in this check...

TristanZerrAMD:
Hey @BrianHarrisonAMD + @JonathanLichtnerAMD, really good feedback!  Thinking about it more, I agree we should not take these off PR's, it would become a nuisance for people if they merge and then find out later there was a HipTidy change required.  My primary feedback was that I noticed in one of my PR's, it took about 2 hours to get to the stage where tests were running, with 1h and 40min of that time spent in HipTidy.  It would be nice to get runtime test results faster - I believe both of you are suggesting the same thing, which would be to run HipTidy in parallel; I think this would definitely be an improvement.  So do we think moving HipTidy to "Full Tests" makes sense?

TristanZerrAMD:
@BrianHarrisonAMD + @JonathanLichtnerAMD - moved HipTidy to FullTests, will let the CI cycle through, let me know if I missed anything.

pfultz2:
If you switch to rocm-cmake's clang tidy support(which is similar): https://rocm.docs.amd.com/projects/ROCmCMakeBuildTools/en/latest/reference/ROCMClangTidy.html

It supports a cache(with Makefile builds only), which will only rerun tidy on the files that have changed. This can give a massive boost to the tidy runs on PRs. You just need to make sure you share the cache directory across builds. Its not very large cache, just lists of md5 hashes.

In migraphx, we actually run tidy on github's hosted runners(since it doesnt require a gpu), but we had to trim down the dockerfile size for that. GH actions have a nice way to cache a directory and copy it in again on new builds/PRs. The other advantage of using GH action is that it can annotate the PR with the results(using the [`ROCM_ENABLE_GH_ANNOTATIONS`](https://rocm.docs.amd.com/projects/ROCmCMakeBuildTools/en/latest/reference/ROCMAnalyzers.html#variable:ROCM_ENABLE_GH_ANNOTATIONS) option), so you dont have to scroll through an output log. I dont know if there is an easy way to do that with our jenkins CI though:

https://github.com/ROCm/AMDMIGraphX/blob/develop/.github/workflows/ci.yaml#L113

adickin-amd (approved):

JonathanLichtnerAMD (approved):

PR #3795, opened 2025-06-11
Add activation args and values for clamp

BrianHarrisonAMD:
Draft to show changes

### DIFF
--- a/src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
+++ b/src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
@@ -64,7 +64,7 @@ using OutElementOp = ck::tensor_operation::element_wise::AddClamp;
 const auto in_element_op  = InElementOp{};
 const auto wei_element_op = WeiElementOp{};
 
-static OutElementOp GetOutElementOp(const miopen::fusion::ActivationOpInvokeParam& activationOp)
+OutElementOp GetOutElementOp(const miopen::fusion::ActivationOpInvokeParam& activationOp)
 {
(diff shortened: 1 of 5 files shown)


src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -63,12 +63,21 @@ using OutElementOp = ck::tensor_operation::element_wise::AddClamp;
...
 const auto in_element_op  = InElementOp{};
 const auto wei_element_op = WeiElementOp{};
-// const auto out_element_op = OutElementOp{floor, ceil};
-const auto GetOutElementOp = []() {
-    const float floor = 0;
-    const float ceil  = std::numeric_limits<ck::bhalf_t>::max();
-    return OutElementOp{floor, ceil};
-};
+
+static OutElementOp GetOutElementOp(const miopen::fusion::ActivationOpInvokeParam& activationOp)
```
BrianHarrisonAMD: Probably don't need static.

adickin-amd: Is this something that we can put in a more global location so the conv+act solver can use it to?  Not sure if we want to share things like this between solvers?  Im thinking we will want the same params available.

bghimireamd: I am guessing conv+act  could use `GetOutElementOp`

BrianHarrisonAMD: Yea we could put it in a share utility, or even the ck_util.

PR #3800, opened 2025-06-13
Draft of fixes for the 7.0 hipRTC changes

BrianHarrisonAMD:
DO NOT MERGE YET


src/kernels/MIOpenKthvalue.cpp
```diff
@@ -29,6 +29,8 @@
 #include <hip/hip_bfloat16.h>
 #include <hip/hip_runtime.h>
 #endif
+#include "miopen_cstdint.hpp"
+#include <type_traits>
```
BrianHarrisonAMD, line 33: We need to run this on current version, and see if this should be guarded with versions.

PR #3808, opened 2025-06-16
Bart/kernel search log

amd-bartgips:
This PR attempts to change the logging behaviour of MIOpen such that we can extract more data when doing exhaustive tuning.:
* Allows for a env var flag `MIOPEN_TUNING_ALLOW_OUTLIERS` to force a full benchmark run (i.e. 10 iterations in GenericSearch) even when the config seems to not be an improvement based on first probe
* At logging level Info2 we also add a line to show all the samples for downstream processing.
* this is also implemented in [src/conv/solver_finders.cpp](https://github.com/ROCm/MIOpen/compare/bart/kernel_search_log?expand=1#diff-d15575b89a5340eb476292e59f22be216632f89d65d0101fdc7dcd25194e19d5) such that we can extract benchmarking data for solvers that do not have searchable configs.


amd-bartgips:
Some initial comments/questions:
* should we perhaps change the log level for the "samples" statements
* should we refactor the few lines that perform the vector-->string operation into a function (if so where shall we put it?)
* Should we consider a env_var to synchronize/alter the amount of samples gathered at the different places (sometimes 10, sometimes 7, etc)

amd-hsivasun:

PR #3809, opened 2025-06-16
Fix NOCK build break 

BrianHarrisonAMD:


adickin-amd (approved):

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -1076,6 +1076,7 @@ ConvSolution InitInvokerFactoryNHWC(const ExecutionContext&,
     else
     {
         ConvSolution result;
+#if MIOPEN_BACKEND_HIP && MIOPEN_USE_COMPOSABLEKERNEL
```
JonathanLichtnerAMD, line 1079: This is optional since we want a quick fix, but we could move the ifdef above line 996 and do the checking for HIP and CK once instead of duplicating the check

JonathanLichtnerAMD (approved):

PR #3812, opened 2025-06-17
Add 3d support to group conv activ

adickin-amd:


src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -964,7 +976,7 @@ MakeNHWCCKArgPtr(const std::shared_ptr<DeviceOpType>& sh_conv_ptr,
             conv_param.alpha,
             conv_param.beta,
             GetOutElementOp<typename CKArgsType::OutputDataType,
-                            typename CKArgsType::OutputElementOpType>(activ_param));
+                            typename CKArgsType::OutputElementOpType>(*activ_param_ptr));
```
SamuelReeder, line 1008: Should we have error handling incase `activ_param_ptr` remains `nullptr`?

adickin-amd: There are asserts but those are compiled out at runtime.  The cast for activ_param_ptr would throw on a bad cast I think due to the fact that its a reference.

`&dynamic_cast<const miopen::fusion::ActivationOpInvokeParam&>(*data_ctx.op_args.params[2]);`

src/include/miopen/fusion/solvers.hpp
```diff
@@ -249,6 +249,8 @@ struct ConvCKIgemmGrpFwdActivFused final
     GetSolution(const FusionContext& ctx,
                 const FusionDescription& fdesc_problem,
                 const PerformanceConfigConvCKIgemmGrpFwdActivFused& config) const override;
+    bool MayNeedWorkspace() const override { return true; }
+    size_t GetWorkspaceSize(const FusionContext&, const FusionDescription&) const override;
```
BrianHarrisonAMD, line 253: Do we need this if it's not support NCHW layout?

BrianHarrisonAMD: Oh I guess we added it nevermind.

src/include/miopen/solver/implicitgemm_ck_util.hpp
SamuelReeder: What if the `op_args.params.size()` are < 2 though? So it skips the conditionals and then you dereference `nullptr`. Unless that's unreachable.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -947,16 +958,28 @@ MakeNHWCCKArgPtr(const std::shared_ptr<DeviceOpType>& sh_conv_ptr,
...
-        const auto& activ_param =
-            dynamic_cast<miopen::fusion::ActivationOpInvokeParam&>(*data_ctx.op_args.params[2]);
-        assert(&activ_param);
+        if(data_ctx.op_args.params.size() == 2)
+        {
+            activ_param_ptr = &dynamic_cast<const miopen::fusion::ActivationOpInvokeParam&>(
+                *data_ctx.op_args.params[1]);
+            assert(activ_param_ptr);
+        }
+        else if(data_ctx.op_args.params.size() > 2)
```
BrianHarrisonAMD: I think we should make it exactly 2 & 3, and throw if it's not one of those.

src/include/miopen/solver/implicitgemm_ck_util.hpp
```diff
@@ -878,15 +878,26 @@ MakeNCHWCKArgPtr(const CKArgsType& ck_args,
...
-        assert(&activ_param);
-
-        ConstData_t bias_buf = bias_param.bdata;
+        if(data_ctx.op_args.params.size() == 2)
+        {
+            activ_param_ptr = &dynamic_cast<const miopen::fusion::ActivationOpInvokeParam&>(
+                *data_ctx.op_args.params[1]);
+            assert(activ_param_ptr);
+        }
+        else if(data_ctx.op_args.params.size() > 2)
```
BrianHarrisonAMD: I think this should be exactly 2 & 3, and then throw if it mismatches.

src/include/miopen/solver/implicitgemm_ck_util.hpp
BrianHarrisonAMD: Oh yea I guess this is similar to what I said above.
We should throw if the fusion size isn't exactly what we are expecting / supporting I think.

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
```diff
@@ -95,13 +101,66 @@ using DeviceOpGFwdRelu =
...
+                                                                                OutDataT,
+                                                                                InElemOp,
+                                                                                WeiElemOp,
+                                                                                OutElemOp,
+                                                                                ACompType,
+                                                                                BCompType>>
+{
+    static constexpr ck::index_t NDim = NDimSpatial;
+    // using BComputeType                = BCompType;
+    // ..
```
BrianHarrisonAMD: Can remove it looks like.

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
```diff
@@ -144,32 +242,90 @@ struct CKArgs
...
+            std::array<ck::index_t, 2> adjusted_lPadding{};
+            std::array<ck::index_t, 2> adjusted_rPadding{};
+
+            std::copy(
+                filter_stride.begin(), filter_stride.begin() + 2, adjusted_filter_stride.begin());
+            std::copy(filter_dilation.begin(),
+                      filter_dilation.begin() + 2,
+                      adjusted_filter_dilation.begin());
+            std::copy(lPadding.begin(), lPadding.begin() + 2, adjusted_lPadding.begin());
+            std::copy(rPadding.begin(), rPadding.begin() + 2, adjusted_rPadding.begin());
```
BrianHarrisonAMD, line 305: Why does 2D need to adjust the kernel args?

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
```diff
@@ -186,10 +342,11 @@ struct CKArgs
         return MakeArgPtr(op_ptr,
                           data_ctx.in,
                           conv_param.weights,
+                          nullptr,
                           data_ctx.out,
                           conv_param.alpha,
                           conv_param.beta,
-                          GetOutElementOpClamp(activ_param));
```
BrianHarrisonAMD, line 192: Where is this defined?

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
```diff
@@ -186,10 +342,11 @@ struct CKArgs
...
+                          GetOutElementOp<ck::bhalf_t, OutElementOp>(activ_param));
```
BrianHarrisonAMD, line 347: Where is GetOutElementOp defined now?

src/include/miopen/fusion/solvers.hpp
adickin-amd: this does support NCHW

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
adickin-amd: it was refactored into ck_utils. see https://github.com/ROCm/MIOpen/pull/3802

src/include/miopen/solver/implicitgemm_ck_util.hpp
adickin-amd: can do

src/include/miopen/solver/implicitgemm_ck_util.hpp
adickin-amd: we now throw in the else if not 2 or 3

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
adickin-amd: This is based on the CBA PR but its due to how we store the data in CKArgs now.

For example if you look at `in_lens` it used to be defined as `std::array<ck::index_t, 5> in_lens;` and now its `std::array<ck::index_t, 6> in_lens;`.  Since there is now an extra dimension we dont need for 2D, we adjust `adjusted_in_lens` to be the first 5 elements `std::copy(in_lens.begin(), in_lens.begin() + 5, adjusted_in_lens.begin());`.  

This goes for all the ckargs, they all have 1 extra dimension so we cut 6 -> 5 and 3->2 for 2D.

SamuelReeder (approved):
LGTM. Looked at the prelim datatype support too.

BrianHarrisonAMD (approved):

PR #3813, opened 2025-06-17
[P1] Add missing copyright comments 

SamuelReeder:
See [this ticket]( https://ontrack-internal.amd.com/browse/SWDEV-537482). Edited in web since I can't use VPN currently.


randyspauldingamd (approved):
I don't think it matters, but the years are all 2017. That actually may be right haha!

SamuelReeder:
> I don't think it matters, but the years are all 2017. That actually may be right haha!

Good catch! Many are outdated, but changed these to 2025 to be safe.

PR #3817, opened 2025-06-18
Add split_k support to 3d ck bwd conv solver

iq136boy:
Add split_k support to 3d ck bwd conv solver.


BrianHarrisonAMD:
I think with last time we added split_k for 2D we also had to update the tuning DBs since there was cached entries that were no longer valid. 

Not sure if that will also be the case here, but should be checked before merging.

iq136boy:
Temporarily closed for repo migration purpose and will reopen when it is done.

PR #3818, opened 2025-06-18
Fix Segfault on WRW Backwards

adickin-amd:
With a refactor to the InitInvokerFactoryNCHW and InitInvokerFactoryNHWC functions we guarded a few sections in the newly refactored functions with `else if constexpr(std::is_same_v<CastType, miopen::conv::DataInvokeParams>)`.  This causes issues when the `CastType` is `miopen::conv::WrWInvokeParam`.  Looking at the previous code, we did not guard the default function flow.  I think the best fix for now is to remove this guard as we likely do want some default behaviour we can fallback on rather than crashing/throwing.

### DIFF
--- a/src/include/miopen/solver/implicitgemm_ck_util.hpp
+++ b/src/include/miopen/solver/implicitgemm_ck_util.hpp
@@ -838,13 +838,9 @@ ConvTensors GetTensors(const CastType& data_ctx)
 
         return tensors;
     }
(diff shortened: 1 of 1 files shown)


JonathanLichtnerAMD (approved):

JonathanLichtnerAMD:
LGTM assuming tests pass

PR #3820, opened 2025-06-19
Create samples directory with inital grouped conv + bias +activ sample

SamuelReeder:
Finding standalone examples of the MIOpen API is difficult—most code is tucked away in consumer repositories, driver internals, or test suites. This samples directory provides a straightforward, E2E illustration that makes MIOpen more accessible. In particular, this initial example compiles and executes a CBA-fused forward convolution.


samples/cba_fused_infer/sample_cba_fused_fwd.cpp
```diff
@@ -0,0 +1,307 @@
...
+                      : (std::max((wei_spatial_lens[i] - (in_spatial_lens[i] % strides[i])), 0));
+        pads[i] /= 2;
+    }
+
+    // parames for ReLU (alpha=0, beta=0, gamma=0), f(x) = max(0, x)
+    double activ_alpha = 0.0;
+    double activ_beta  = 0.0;
+    double activ_gamma = 0.0;
+
+    std::vector<int> in_len = {in_n, in_c, in_spatial_lens[0], in_spatial_lens[1]};
```
SamuelReeder, line 177: I define my input tensor dimensions with NCHW layout but when I set the descriptor via `miopenSetNdTensorDescriptorWithLayout`, I pass `miopenTensorNHWC` as the layout arg. I could not make it work when trying to define all the tensors in NHWC layout natively.  In the driver code, it seems like tensors are passed as NCHW regardless of the layout enum too. So I'm wondering if tensors are supposed to be passed as NCHW and they're transformed internally?

If not, I was probably computing the dimensions incorrectly in one of my NHWC tensors.

BrianHarrisonAMD: A bit confusing, but tensor dims are always defined in NCHW order.
For a tensor that is NHWC layout, it just has strides such that C's stride == 1 (or the smallest stride value if it's not packed).

BrianHarrisonAMD:
We should coordinate with @amd-jnovotny for doc changes.

samples/README.md
```diff
@@ -0,0 +1,9 @@
+# MIOpen Samples
+
+This directory contains sample code for using [MIOpen](https://github.com/ROCmSoftwarePlatform/MIOpen).
```
amd-jnovotny:
```suggestion
This directory contains sample code for using [MIOpen](https://github.com/ROCm/MIOpen).
```

amd-jnovotny (commented):
This is a good idea. I've added one suggestion and maybe we can link to this page from the MIOpen index page and/or ToC too.

amd-jnovotny (approved):
Looks good.

amd-jnovotny (approved):

PR #3821, opened 2025-06-19
Quick change to enable 3d conv activ in driver

SamuelReeder:


SamuelReeder:
NVM I realized it's already fine.

PR #3822, opened 2025-06-19
Remove File reorganization backward compatibility (MIOpen)

jonatluu:
Summary of proposed changes:
remove backwards compatibility code. Feature has been disabled related code is not needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -501,9 +501,6 @@ message(STATUS "HALF_INCLUDE_DIR: ${HALF_INCLUDE_DIR}")
 
 option( MIOPEN_DEBUG_FIND_DB_CACHING "Use system find-db caching" ON)
 
-# FOR HANDLING ENABLE/DISABLE OPTIONAL BACKWARD COMPATIBILITY for FILE/FOLDER REORG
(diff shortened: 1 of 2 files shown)


JonathanLichtnerAMD:
@jonatluu I think your branch names are causing issues with CI.  It's not running for your branch, and I do not see you branch in jenkins (http://micimaster.amd.com/job/MLLIBS/job/MIOpen/).

Generally, we do not use colons in branch names, and I cannot even view your commits from git.  Please see https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#naming-branches-and-tags for a discussion on what special characters are safe to use in git.

Anyway, please base the PR from the "develop" branch and for branch names we typically use the format <userid>/<descriptive text>, so you could use something like jonatluu/remove-file-reorg

I hope this helps!

JonathanLichtnerAMD:
This passed our internal CI, so this looks good to me: http://micimaster.amd.com/job/MLLIBS/job/MIOpen/view/change-requests/job/PR-3822/

The azure CI failures are unrelated to this change.

JonathanLichtnerAMD (approved):

PR #3825, opened 2025-06-20
BN: Variant 1 spatial fwd train OCL - HIP conversion

EnricoDeg:
Summary:

* `MIOpenBatchNormFwdTranSpatialHIP`

  * Created this file by converting the OpenCL variant 1 implementation to HIP without heavily changing the rational of the original implementation.

* Environment Configuration Handling

  * Introduced a more modern mechanism for loading and storing environment variables into `miopen::batchnorm::config` and `miopen::config`.

* `default_configuration.hpp`

  * Added this header to provide default values for macros that may be used without being explicitly defined by the user.

* `batchnorm_functions.hpp`

  * Converted all functions to HIP.
  * Replaced some macros with `__forceinline__` template functions for better readability and maintainability.
  * Note: some of these functions are not used by the `MIOpenBatchNormFwdTranSpatialHIP` variant 1 and will require future testing.

* `reduction_functions.hpp`

  * Converted only the functions necessary for `MIOpenBatchNormFwdTranSpatialHIP` variant 1.
  * Remaining functions are yet to be converted.

* `vector_type.hpp`

  * Vector types using `__attribute__((ext_vector_type(4)))`


src/kernels/activation_functions.hpp
```diff
@@ -23,10 +23,24 @@
  * SOFTWARE.
  *
  *******************************************************************************/
+
+// TODO: if this header is only used for batch norm, it should also use the type defined
+// in configurations.hpp. We should move these common parts into a separate header,
+// maybe call it `batchnorm_common.hpp`.
```
BradPepersAMD: a quick find shows this is used in MIOpenRNNHiddenStatusUpdate.cl and MIOpenNeuron.cl outside of batch norm so this comment isn't needed.

src/kernels/configuration.hpp
```diff
@@ -0,0 +1,297 @@
...
+{
+    unknown,
+    gfx103x,
+    gfx110x,
+    gfx120x,
+};
+
+namespace detail {
+
+// TODO: why this is here, becasue before c++ 20, double is not supported to be template parameter
```
BradPepersAMD, line 124: we just moved to C++ 20 and was merged yesterday so perhaps no longer needed?

BrianHarrisonAMD (dismissed):
I didn't do a close review of the kernel code, but the changes seem generally fine assuming we are happy with the performance of the kernels. 

We should run tests across ASICs, and do a perf sweep before merging if we haven't already.

bghimireamd:
do we have gtest for this kernel?

EnricoDeg:
> do we have gtest for this kernel?

The existing bnorm fwd tests which use variant 1 are now running the hip version

src/kernels/configuration.hpp
EnricoDeg: In develop I see hip kernels still compiled with c++17

It should be here: https://github.com/ROCm/MIOpen/blob/d904d7021f6c4e35abd5e3d6ce3cdd8a5984732a/src/hip/hip_build_utils.cpp#L135

or am I looking in the wrong place?

src/kernels/activation_functions.hpp
EnricoDeg: Removed

amd-hsivasun:

PR #3828, opened 2025-06-21
docs/sphinx clean up

ghost:
MIOpen/README.md
be ensured that 'doxygen' is installed

docs\sphinx\requirements.txt
The contents of this file should be simplified to:
  rocm-docs-core
  sphinx
  doxysphinx
(without specifying exact versions of packages)
This will give more flexibility for building documentation from the sources

docs/install/build-source.rst
'clang-format-10' is better replace to 'clang-format'
('clang-format' is a symbolic link pointing to the real version of the executable file)


amd-jnovotny:
This isn't how our Sphinx works. This would break documentation.

ghost:
I checked - Sphinx documentation is generated.
Even in Windows environment.

amd-jnovotny:
I'll check with the tools team, who are typically the folks to update the docs CI. It's not typically the way we do things, so we probably don't want a non-standard approach here. But I'll see what they say.

We could still update the README with the note about installing doxygen in any event.

amd-jnovotny:
I confirmed that the changes to the requirements file do not align with our workflow, so we can't accept those changes. Feel free to file a new PR with the other changes. The team would have to review and approve the clang command changes.

PR #3830, opened 2025-06-23
[Tests] Convert LRN  from ctest to gtest

Vsevolod1983:


scerzh (approved):
LGTM

test/gtest/lrn.cpp
```diff
@@ -269,53 +263,76 @@ struct verify_lrn_bwd
     }
 };
 
-template <class T>
-struct lrn_driver : test_driver
+inline auto GenCases(bool limit = false)
```
TristanZerrAMD, line 268: for my own curiosity, does a modern C++ optimizer do anything different if the inline keyword is omitted?

test/gtest/lrn.cpp
```diff
@@ -325,17 +342,77 @@ struct lrn_driver : test_driver
...
+using GPU_Lrn_FP16 = LrnCommon<half_float::half>;
+
+TEST_P(GPU_Lrn_FP32, TestFloat) { this->Run(); }
+TEST_P(GPU_Lrn_FP16, TestFloat16) { this->Run(); }
+
+INSTANTIATE_TEST_SUITE_P(Smoke, GPU_Lrn_FP32, GetCasesSmoke());
+INSTANTIATE_TEST_SUITE_P(Full, GPU_Lrn_FP32, GetCasesFull());
+
+INSTANTIATE_TEST_SUITE_P(Smoke, GPU_Lrn_FP16, GetCasesSmoke());
+INSTANTIATE_TEST_SUITE_P(Full, GPU_Lrn_FP16, GetCasesFull());
```
TristanZerrAMD, line 418: It might make the code a bit easier to understand to just call GenCases() directly?  There would be some memory duplication with the static constants removed, though not sure it makes much of a difference since this is just memory for setup code, not the actual runtime test case execution?

TristanZerrAMD (approved):
Looks good, keep it coming!  Just some minor ideas, no need to change anything.

test/gtest/lrn.cpp
```diff
@@ -23,20 +23,15 @@
...
-#include "test.hpp"
-#include "verify.hpp"
-#include "get_handle.hpp"
-#include "tensor_holder.hpp"
-#include <miopen/miopen.h>
-#include <miopen/tensor.hpp>
-#include <miopen/stringutils.hpp>
+
+#include "gtest_common.hpp"
+#include <tensor_util.hpp>
```
randyspauldingamd, line 28: List needed headers:
#include "network_data.hpp"

I didn't do a full scan, so please check for any others.

test/gtest/lrn.cpp
```diff
@@ -325,17 +342,77 @@ struct lrn_driver : test_driver
...
 
-// To address compiler issue for bfloat1 type in SWDEV-202752
-// creating explicit instance of lrn_driver with bfloat16 with noop
-template <>
-struct lrn_driver<bfloat16> : test_driver
-{
-    lrn_driver() {}
-    void run() { std::cout << "bfloat16 is not supported in lrn" << std::endl; };
+    // we need cpu data for backward pass later, so return it from this function
+    void VerifyLrnFoward(const miopen::LRNDescriptor& plrnDesc, const tensor<T>& pinput)
```
randyspauldingamd: typo: VerifyLrnForward

test/gtest/lrn.cpp
```diff
@@ -23,20 +23,15 @@
...
-#include <limits>
-#include <iostream>
+
+namespace {
+
+using TestCase =
+    std::tuple<std::vector<int>, unsigned int, double, double, double, miopenLRNMode_t>;
 
 template <class T>
 struct verify_lrn_foward
```
randyspauldingamd: typo: verify_lrn_forward

test/gtest/lrn.cpp
```diff
@@ -325,17 +342,77 @@ struct lrn_driver : test_driver
...
+
+        double threshold = std::numeric_limits<T>::epsilon() * tolerance;
+        double error     = miopen::rms_range(cpu, gpu);
+
+        if(error > threshold)
+        {
+            direction.fail();
+        }
+
+        ASSERT_LE(error, threshold) << "n: " << n << std::endl
```
randyspauldingamd, line 387: Use EXPECT_LE. Otherwise if fwd fails, ASSERT_LE exits the function, the cpu results are not saved, and then it segfaults in bwd. (I verified this using failure injection)

test/gtest/lrn.cpp
```diff
@@ -325,17 +342,77 @@ struct lrn_driver : test_driver
...
+    }
+
+    template <class TDirection>
+    void CompareResults(const TDirection& direction, bool saveCpuResults = false)
+    {
+        const tensor<T> cpu = direction.cpu();
+        const tensor<T> gpu = direction.gpu();
+
+        // taken from the original test
+        double tolerance = 80;
```
randyspauldingamd: This tolerance is very conservative. I ran the tests 3 times on gfx942 and the maximum needed tolerance was 1.274 for forward and 5.180 for backward. I did not see any variation between the 3 runs, but round up to 1.5 and 6.0 respectively, just in case.

test/gtest/lrn.cpp
```diff
@@ -135,7 +130,7 @@ struct verify_lrn_foward
         return out;
     }
 
-    void fail(int) const
+    void fail() const
     {
         std::cout << "verify_lrn_foward" << std::endl;
```
randyspauldingamd: typo: verify_lrn_forward

randyspauldingamd (changes requested):

test/gtest/lrn.cpp
Vsevolod1983: This is for performance mostly. Test cases will be generated only once (one time for Full and one time for Smoke) and gets saved into static variables instead of being generated several times. Just saves some time.

test/gtest/lrn.cpp
Vsevolod1983: Good catch, I will just swap ASSERT macro and setting cpu_tensor to have it works

test/gtest/lrn.cpp
Vsevolod1983: Done

test/gtest/lrn.cpp
Vsevolod1983: I think it might be inlined( if compiler decides to do so) even without inline keyword specified. But let's keep it as an additional hint for it.

test/gtest/lrn.cpp
randyspauldingamd: Discussed, and not needed

test/gtest/lrn.cpp
randyspauldingamd: That works too! I'm curious why we use ASSERT* so much. Usually EXPECT* is preferred, so the most test clauses run and we get maximum information. The runtime benefit from using ASSERT is negligible. Are there other reasons I'm not aware of?

TristanZerrAMD: I'm curious about this one too.  In the couple places I worked previously with GTest, we only used EXPECT.

randyspauldingamd (approved):
LGTM

test/gtest/lrn.cpp
Vsevolod1983: ASSERT stops execution of the current function immediately, right? Probably sometimes it makes sense, we detected an error and immediately stopping the test. I'm wondering why ASSERT does not work as an exception to stop the current test but just quit the current function only.

randyspauldingamd: Correct. Gtest says they prefer that all test clauses are in the top-level TEST* function, in which case "stop the current function" would equal "stop the test." I imagine the reason they don't stop the entire test is to make sure cleanup runs. It seems pretty difficult to achieve that without adding a bunch more syntax.

PR #3833, opened 2025-06-24
Fix NCHW precision issues for fused CBA/CA solvers

SamuelReeder:
This pull request achieves support of NCHW & NCDHW layout for fused CBA/CA solvers. The prior issue was that the workspace was not entirely implemented, and there was a `nullptr` in place of it. When a convolution with supposedly NCHW tensors was performed, the convolution was actually being performed on NHWC tensors, because the tensors were never transformed. Therefore, the computation lacked precision as it was using misarranged data. 

I used variants of the following driver commands to test my implementation:
```bash
./bin/MIOpenDriver CBAInferbfp16 --spatial_dim 2 --batchsize 4 --in_channels 8 --out_channels 16 --group_count 4 --in_h 20 --in_w 20 --fil_h 3 --fil_w 3 --conv_stride_h 2 --conv_stride_w 2 --pad_mode same --dilation_h 1 --dilation_w 1 --in_layout NCHW --fil_layout NCHW --out_layout NCHW --alpha 1 --beta 1 --activMode 7 -F 5 -t 1

./bin/MIOpenDriver CBAInferbfp16 --spatial_dim 3 --batchsize 4 --in_channels 8 --out_channels 16 --group_count 4 --in_d 10 --in_h 20 --in_w 20 --fil_d 3 --fil_h 3 --fil_w 3 --conv_stride_d 2 --conv_stride_h 2 --conv_stride_w 2 --pad_mode same --dilation_d 1 --dilation_h 1 --dilation_w 1 --in_layout NCDHW --fil_layout NCDHW --out_layout NCDHW --alpha 1 --beta 1 --activMode 7 -F 5 -t 1
```


src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -154,6 +154,19 @@ struct CKArgs
         C  = C1 / G; // Number of input Channel per group
         K  = K1 / G; // Number of output Channel per group
 
+        if(problem.IsLayoutNHWC())
+        {
+            auto miopen_in_strides  = problem.GetIn().GetStrides();
```
adickin-amd: Im wondering if it makes more sense to set NHWC like we do NCHW so its consistent?  Its nice this is refactored but I wonder if this makes things more confusing than  just assigning the shapes like you did below.

src/fusion.cpp
```diff
@@ -73,10 +73,10 @@ miopenStatus_t ConvBiasActivFusion(const Handle& handle,
                                    const TensorDescriptor& yDesc,
                                    Data_t y)
 {
-    assert(workspace == nullptr);
-    assert(workspaceSizeInBytes == 0);
-    std::ignore = workspace;
-    std::ignore = workspaceSizeInBytes;
+    // assert(workspace == nullptr);
```
adickin-amd: can we remove this?

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
SamuelReeder: Yes that should be fine.... I don't think there's any functional reason we set NHWC strides that way as opposed to doing it manually.

include/miopen/miopen.h
```diff
@@ -3902,6 +3902,31 @@ miopenExecuteFusionPlan(const miopenHandle_t handle,
...
+ * @param input            Source data tensor  (input)
+ * @param outputDesc       Decriptor of the output tensor (input)
+ * @param output           Destination data tensor  (output)
+ * @param args             An argument object of the fused kernel (input)
+ * @param workspace        A pointer to an intermediate workspace (input)
+ * @param workspaceSize Size of the memory in bytes pointed to by workSpace above (input)
+ * @return           miopenStatus_t
+ */
+MIOPEN_EXPORT miopenStatus_t
+miopenExecuteFusionPlanWithWorkspace(const miopenHandle_t handle,
```
BrianHarrisonAMD: I think we usually call these _v2.
And we likely would want to add a deprecated notice to the original since it's now potentially incorrect to use.

driver/CBAInferFusion_driver.hpp
```diff
@@ -1173,16 +1212,29 @@ void CBAInferFusionDriver<Tgpu, Tref>::runGPUFusedConvBiasInference()
         std::cerr << "ConvBiasInference plan not supported." << std::endl;
     }
 
+    size_t workspace_size = 0;
+    miopenFusionPlanGetWorkSpaceSize(
+        GetHandle(), fusePlanDesc, &workspace_size, miopenConvolutionFwdAlgoImplicitGEMM);
```
BrianHarrisonAMD, line 1208: Odd this API existed, but setting it wasn't possible?

test/gtest/conv_activ_infer.cpp
```diff
@@ -223,7 +223,7 @@ INSTANTIATE_TEST_SUITE_P(
     GPU_ConvGrpActivInfer3D_BFP16,
     testing::Combine(testing::Values(miopenActivationCLAMP),
                      testing::ValuesIn(GroupConvTestConfig<3>::GetConfigs<Direction::Forward>()),
-                     testing::Values(miopenTensorNDHWC /*, miopenTensorNCDHW*/),
+                     testing::Values(miopenTensorNDHWC, miopenTensorNCDHW),
                      testing::Values(0.5f),
                      testing::Values(1.0f),
                      testing::Values(0.5f)));
```
BrianHarrisonAMD: You and @adickin-amd need to coordinate since he added a bunch of new tests that need NCHW enabled on.

driver/CBAInferFusion_driver.hpp
SamuelReeder: I can't find any such function, so I don't think so. The old fusions passed the workspace directly in the execution API call too and got it via that call beforehand.

SamuelReeder:
Adam's PR will be merged into develop first and then he will merge develop into here. At that point can finalize reviews and merge.

SamuelReeder:
I fixed `test/cba_inference.cpp`--which seems outdated and was trivially succeeding because no shapes were applicable, but the applicability check changes for the solvers were making it fail.

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
```diff
@@ -179,6 +169,10 @@ struct CKArgs
             out_lens = {G, N, K, Do, Ho, Wo};
             wei_lens = {G, K, C, Z, Y, X};
 
+            in_strides  = {C, Di * Hi * Wi * G * C, 1, Hi * Wi * G * C, Wi * G * C, G * C};
+            out_strides = {K, Do * Ho * Wo * G * K, 1, Ho * Wo * G * K, Wo * G * K, G * K};
+            wei_strides = {K * Z * Y * X * C, Z * Y * X * C, 1, Y * X * C, X * C, C};
```
BrianHarrisonAMD, line 174: We should be using the strides from the problem I think.
Why did this need to be changed?

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
```diff
@@ -206,6 +200,10 @@ struct CKArgs
             out_lens = {G, N, K, Ho, Wo};
             wei_lens = {G, K, C, Y, X};
 
+            in_strides  = {C, Hi * Wi * G * C, 1, Wi * G * C, G * C};
+            out_strides = {K, Ho * Wo * G * K, 1, Wo * G * K, G * K};
+            wei_strides = {K * Y * X * C, Y * X * C, 1, X * C, C};
```
BrianHarrisonAMD, line 205: Same here.

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -166,20 +166,15 @@ struct CKArgs
...
-            auto miopen_in_strides  = problem.GetIn().GetStrides();
-            auto miopen_out_strides = problem.GetOut().GetStrides();
-            auto miopen_wei_strides = problem.GetWeights().GetStrides();
-            miopen_in_strides.insert(miopen_in_strides.begin(), C);
-            miopen_out_strides.insert(miopen_out_strides.begin(), K);
-            miopen_wei_strides.insert(miopen_wei_strides.begin(), K * miopen_wei_strides[0]);
-            std::copy(miopen_in_strides.begin(), miopen_in_strides.end(), in_strides.begin());
-            std::copy(miopen_out_strides.begin(), miopen_out_strides.end(), out_strides.begin());
-            std::copy(miopen_wei_strides.begin(), miopen_wei_strides.end(), wei_strides.begin());
+            bias_strides = {K, 0, 1, 0, 0, 0};
```
BrianHarrisonAMD, line 177: Same here, shouldn't we use problem strides?

src/solver/conv_ck_igemm_grp_fwd_bias_activ_fused.cpp
```diff
@@ -203,23 +198,15 @@ struct CKArgs
...
-            out_lens     = {G, N, K, Ho, Wo};
-            wei_lens     = {G, K, C, Y, X};
-            bias_strides = {K, 0, 1, 0, 0};
+            in_lens  = {G, N, C, Hi, Wi};
+            out_lens = {G, N, K, Ho, Wo};
+            wei_lens = {G, K, C, Y, X};
+
+            in_strides  = {C, Hi * Wi * G * C, 1, Wi * G * C, G * C};
+            out_strides = {K, Ho * Wo * G * K, 1, Wo * G * K, G * K};
+            wei_strides = {K * Y * X * C, Y * X * C, 1, X * C, C};
```
BrianHarrisonAMD, line 207: Same here.

include/miopen/miopen.h
BrianHarrisonAMD, line 1: Did we update the MIOpenDriver command logging to match all these changes? 

We need to check `MIOPEN_ENABLE_LOGGING_CMD=1` is updated to match all the new driver changes that have been done, along with the new layouts we support.

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
SamuelReeder: I don't think they worked for NCHW layout, but I'll try again to verify. Maybe I changed other things since making that change. It doesn't make sense why they wouldn't work.

include/miopen/miopen.h
SamuelReeder: I'll look into that too. Now I realize what you meant.

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
BrianHarrisonAMD: If they don't work, then we have a bug, since the problem strides are what was provided from the API, or were calculated if they aren't provided.

include/miopen/miopen.h
BrianHarrisonAMD: Might need to coordinate this across all the PRs you have for CBA.
Not sure which one has all the changes, but seemed like here makes sense.

The driver specific changes to change the arg letter for example might need additional changes (not sure).

src/solver/conv_ck_igemm_grp_fwd_activ_fused.cpp
SamuelReeder: I checked and it unfortunately does not work. The problem calculates NCHW strides in some cases, however, CK requires packed NHWC strides regardless of the tensor layout because of the transpose. That's why calculating NHWC strides manually for either layout works, and getting strides from problem for NHWC works, but for NCHW alone the computed problem strides do not work for CK. 

Most CK solvers I checked have a conditional for calculating the strides either way, but I chose to manually calculate NHWC strides universally because Adam pointed out the alternative is repetitive.

include/miopen/miopen.h
SamuelReeder: There's nothing new that this PR introduces that needs to be fixed. It already logs NHWC args and omits NCHW layout args because they are default. However, the command logging for fusion already omits some important args that should be added at some point. Including activation, alpha, beta, and gamma. I'm thinking its best to do this separately?

I changed the logging for the arg letter in the driver fixes PR.

BrianHarrisonAMD (approved):

PR #3838, opened 2025-06-24
Resolve issue with gfx90a transpose

BrianHarrisonAMD:
In some gfx90a environments we are now seeing the same issue that lead to originally disabling these transpose kernels. 
For now we will limit the transpose kernel instances in all asics down to the reduced set.


randyspauldingamd (approved):
lgtm

adickin-amd (approved):

JonathanLichtnerAMD (approved):
LGTM

cderb (approved):

PR #3841, opened 2025-06-25
[NFC] Remove static reduction

averinevg:
Static reduction has been disabled by default for a long time and can be safely removed without any performance impact.


BrianHarrisonAMD (approved):
LGTM

PR #3844, opened 2025-06-26
Fix CK Inline build to pull include directories correctly

adickin-amd:
Previous inline builds only worked when the include files were in /opt/rocm already. The inline build could fail when new headers or types were added.  These fixes make it so that the correct CK includes get used when building Ck inline.


BrianHarrisonAMD (approved):
LGTM

JonathanLichtnerAMD (approved):

randyspauldingamd (changes requested):
I'd like to add a quick fix where the current version doesn't currently recognize CK being commented out in requirements.txt.

adickin-amd:
Closing this PR, The new PR will be this one: https://github.com/ROCm/MIOpen/pull/3853

adickin-amd:
closing, see above comment

PR #3845, opened 2025-06-27
Fix CBA driver random number generation and type casting

SamuelReeder:
`conv_driver.hpp` lacks the explicit out host data cast to `Tgpu` type, but I believe it is necessary for `CBAInferFusion_driver.hpp`. Moreover, the conv output must necessarily be cast to bfp16 before bias to align with CK.


driver/CBAInferFusion_driver.hpp
```diff
@@ -1346,6 +1349,15 @@ int CBAInferFusionDriver<Tgpu, Tref>::RunForwardCPU()
...
         runCPUActivFwdInference();
     }
+
+    if(sizeof(Tgpu) <= sizeof(Tref))
+    {
+        for(size_t i = 0; i < out_host.size(); ++i)
+        {
+            out_host[i] = static_cast<Tref>(static_cast<Tgpu>(out_host[i]));
+        }
+    }
```
BrianHarrisonAMD: Should probably just check if the types aren't the same, and then do the cast.

Also, are we sure the static cast is doing the proper RNE for bfloat16 conversions?
Should double check it's calling the expected conversion inside the bfloat16 class (I am paranoid).

SamuelReeder: If `MIOPEN_USE_RNE_BFLOAT16==1`, it does RNE for bfloat16 conversion. Looks to be on by default too.

SamuelReeder:
Changed the `--fusion_mode` command letter to `-J` (as in join) from `-F`.  Was considering `P` (as in plan) among others, but most conflict with some other command from `conv_driver.hpp`, which is best to avoid for consistency.

driver/CBAInferFusion_driver.hpp
```diff
@@ -1356,14 +1378,16 @@ int CBAInferFusionDriver<Tgpu, Tref>::VerifyForward()
...
     int match = miopenInferVerify(out.size(), out_host.data(), out.data(), allowedEps);
     if(match == 0)
     {
-        std::cout << "Forward Activation FAILED" << std::endl;
+        std::cout << "Forward Activation FAILED (error: " << error << ")" << std::endl;
         return EC_VerifyFwd;
     }
 
-    std::cout << "Forward Activation Verifies on CPU and GPU" << std::endl;
+    std::cout << "Forward Activation Verifies on CPU and GPU (error: " << error << ")" << std::endl;
```
BrianHarrisonAMD: The error printing here is different than the way miopenInferVerify is calculating actually.
I think we should unify these, and make sure the printed error matches.
Otherwise it might be a bit confusing to track down when the different ways of verifying don't match.

BrianHarrisonAMD: Also, reviewing that miopenInferVerify it seems like the rms_range version might be better suited.

SamuelReeder: Ok so keep it for now? conv_driver.hpp does it the same--but not to excuse the inconsistent code :)

BrianHarrisonAMD: No I think we need to use it instead of miopenInferVerify for determining if it's verified or not.
Right now it uses miopenInferVerify, but prints the error value from rms_range.
I think it should print the value from rms_range, and use it for determining if it's valid.

I think conv_driver should be doing it the same, is it not?

BrianHarrisonAMD: Okay you spooked me, I think conv_driver is doing what we want(ish):

```
auto error        = is_fwd_run_failed ? std::numeric_limits<double>::max()
                                          : (isInt8 ? miopen::rms_range(outhost.data, out_int8)
                                                    : miopen::rms_range(outhost.data, out.GetVector()));

    auto tolerance = GetDefaultTolerance();
    // iGemm's deviation is higher than other algorithms.
    // The reason is most likely different order of computations.
    if(is_fwd_igemm)
        tolerance = tolerance * 10;

    if(!std::isfinite(error) || error > tolerance)
    {
        std::cout << "Forward Convolution FAILED: " << error << " > " << tolerance << std::endl;
        return EC_VerifyFwd;
    }

    std::cout << "Forward Convolution Verifies OK on " << (UseGPUReference() ? "GPU" : "CPU")
              << " reference (" << error << " < " << tolerance << ')' << std::endl;

```

SamuelReeder: Oh yes, that looks better.

BrianHarrisonAMD (approved):

PR #3846, opened 2025-06-27
Fix layout calculation logic to match for Enum & String

BrianHarrisonAMD:
The convolution output logic was using the string layout from input to determine strides, and then was requesting the input enum when constructing the output tensor.

Since these are calculated differently you can get different results which will cause problems.

Unifying the two to instead use the string layout will resolve mismatches, and also allows us to use the provided y layout instead of relying on the x layout enum.

issue could be replicated with the following:

`./bin/MIOpenDriver convfp16 -n 8 -c 1 -H 5 -W 5 -k 32 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC -m conv -g 1 -F 1 -t 1`

### DIFF
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -547,15 +547,10 @@ const std::optional<miopenTensorLayout_t>& TensorDescriptor::GetLayoutEnum() con
             if(tensorLayout)
                 return tensorLayout;
 
-            const auto known_layouts = {std::make_pair("NCHW", miopenTensorNCHW),
-                                        std::make_pair("NHWC", miopenTensorNHWC),
-                                        std::make_pair("NCDHW", miopenTensorNCDHW),
-                                        std::make_pair("NDHWC", miopenTensorNDHWC),
-                                        std::make_pair("CHWN", miopenTensorCHWN)};
-            for(const auto& [layout_str, layout_enum] : known_layouts)
+            auto layout = GetLayout_str();
+            if(IsPossibleLayout4D5D(layout))
             {
-                if(this->IsPossibleLayout4D5D(layout_str))
-                    return layout_enum;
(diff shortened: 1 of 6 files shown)


src/tensor.cpp
```diff
@@ -763,6 +758,41 @@ std::string TensorDescriptor::GetLayout(std::string storage_layout) const
...
+    }
+    else if(layout_str == "NCDHW")
+    {
+        return miopenTensorNCDHW;
+    }
+    else
+    {
+        MIOPEN_THROW("We only support NCHWc4, NCHWc8, CHWNc4, CHWNc8, NCHW, NHWC, NDHWC, NCDHW "
+                    "vectorized tensor layout.");
+        return default_layout;
```
BrianHarrisonAMD: Don't need return.

test/utils/gtest_formating_checks.py
```diff
@@ -40,6 +40,8 @@
     "../../test/gtest/layout_transpose.cpp",
     "../../test/gtest/reduce_custom_fp32.cpp",
     "../../test/gtest/unary_tensor_ops.cpp",
+    "../../test/gtest/cba_infer.cpp",
+    "../../test/gtest/conv_activ_infer.cpp",
```
scerzh: A PR for this issue is under review: #3850

averinevg (approved):

scerzh (approved):

test/utils/gtest_formating_checks.py
BrianHarrisonAMD: Ill remove this change.

PR #3849, opened 2025-06-30
Fix BF16 intrinsic names

scerzh:
`bf16` buffer intrinsics are missing in the current version of compiler and `i16` versions are to be used instead.

To reproduce the issue the latest staging-compiler docker image is needed.

Install requirements (inside MIOpen's source dir)
```bash
CXX=/opt/rocm/llvm/bin/clang++ cget install -f ./dev-requirements.txt --prefix /root/MIOpen/install_dir
```
Configure and build
```bash
rm -rf build
mkdir build
CXX=/opt/rocm/llvm/bin/clang++
cmake -DCMAKE_BUILD_TYPE=Release -DMIOPEN_BACKEND=HIP -DMIOPEN_TEST_DISCRETE=OFF -DCMAKE_PREFIX_PATH='/opt/rocm/;/opt/rocm/hip;/root/MIOpen/install_dir' -S . -B build
cmake --build build -j64 -t miopen_gtest
```
Run the tests
```bash
./build/bin/miopen_gtest --gtest_filter=*GPU_UnitTestConvSolverHipImplicitGemm*BFP16*
```
You should see the errors similar to the following:
```test/plain
Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.raw.ptr.buffer.load.v2i16
ptr @llvm.amdgcn.raw.ptr.buffer.load.v2bf16
Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.raw.ptr.buffer.load.v4i16
ptr @llvm.amdgcn.raw.ptr.buffer.load.v4bf16
Intrinsic name not mangled correctly for type arguments! Should be: llvm.amdgcn.raw.ptr.buffer.store.i16
ptr @llvm.amdgcn.raw.ptr.buffer.store.bf16
```

This PR fixes the described issue.

### DIFF
--- a/src/include/miopen/solver/static_ck_common.hpp
+++ b/src/include/miopen/solver/static_ck_common.hpp
@@ -46,9 +46,6 @@ MIOPEN_DECLARE_ENV_VAR_BOOL(MIOPEN_DEBUG_STATIC_CK_BLOCK_SYNC_LDS_WITHOUT_SYNC_V
 #define WORKAROUND_MIOPEN_ISSUE_557 1
 #define WORKAROUND_SWDEV_413051 1
 
-// LLVM buffer intrinsics llvm.amdgcn.buffer.* have been removed in HIP 6.4
-#define WORKAROUND_SWDEV_498660 (HIP_PACKAGE_VERSION_FLAT >= 6004000000)
-
 namespace miopen {
 namespace solver {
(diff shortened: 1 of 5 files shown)


shurale-nkn (approved):
LGTM

